PR #1063, opened 2020-11-04
[WIP] Initial work on supporting DecimalType

sperlingxx:
This PR is about to enable `DecimalType` in spark GPU runtime.

Due to a lot of unready dependencies in cuDF, the primary intention of this PR is to import `DecimalType`  as a supported type under GPU runtime. Along with the support on supported type system, we enabled the creation of `GpuColumnVector` and `GpuScalar` with `DecimalType`.

In terms of tests, we test creating decimal as `GpuScalar` and `GpuColumnVector`, with some expressions on decimal-type column vectors (such as: GpuIsNull). 

Here is the summary of progress:

* Enable decimal type for valid RapidsMeta in GpuOverrides (according to latest supportedType mechanism).
* support decimal for GpuColumnVector (with test cases in DecimalUnitTest.scala)
* support decimal for GpuScalar (with test cases in DecimalUnitTest.scala)
* decimal converters for GpuRowToColumnarExec (with test cases in GpuBatchUtilsSuite.scala)
* decimal columnar copy for HostColumnarToGpu (with test cases in DecimalUnitTest.scala)
* test cases for decimal in GpuCoalesceBatchesSuite (Some cases fails because contiguous split of decimal column has yet supported by cuDF. I've created a PR to solve this problem.)
* Creating decimal column vector from scalar will be supported after this PR merged (which is close to).
* Test `GpuProject` with decimal columns.
* Add unit tests for binary operators with decimal data.

### DIFF
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
@@ -39,6 +39,24 @@ import java.util.List;
  */
 public class GpuColumnVector extends GpuColumnVectorBase {
 
+  private static HostColumnVector.DataType convertFrom(DataType spark, boolean nullable) {
+    if (spark instanceof ArrayType) {
+      ArrayType arrayType = (ArrayType) spark;
+      return new HostColumnVector.ListType(nullable,
+          convertFrom(arrayType.elementType(), arrayType.containsNull()));
+    } else if (spark instanceof MapType) {
+      MapType mapType = (MapType) spark;
+      return new HostColumnVector.ListType(nullable,
+          new HostColumnVector.StructType(false, Arrays.asList(
+              convertFrom(mapType.keyType(), false),
+              convertFrom(mapType.valueType(), mapType.valueContainsNull())
+          )));
+    } else {
+      // Only works for basic types
+      return new HostColumnVector.BasicType(nullable, getRapidsType(spark));
+    }
+  }
+
   public static final class GpuColumnarBatchBuilder implements AutoCloseable {
     private final ai.rapids.cudf.HostColumnVector.ColumnBuilder[] builders;
     private final StructField[] fields;
@@ -60,17 +78,9 @@ public class GpuColumnVector extends GpuColumnVectorBase {
       try {
         for (int i = 0; i < len; i++) {
           StructField field = fields[i];
-          if (field.dataType() instanceof MapType) {
-            builders[i] = new ai.rapids.cudf.HostColumnVector.ColumnBuilder(new HostColumnVector.ListType(true,
-                new HostColumnVector.StructType(true, Arrays.asList(
-                    new HostColumnVector.BasicType(true, DType.STRING),
-                    new HostColumnVector.BasicType(true, DType.STRING)))), rows);
-          } else {
-            DType type = getRapidsType(field);
-            builders[i] = new ai.rapids.cudf.HostColumnVector.ColumnBuilder(new HostColumnVector.BasicType(true, type), rows);
-          }
-          success = true;
+          builders[i] = new HostColumnVector.ColumnBuilder(convertFrom(field.dataType(), field.nullable()), rows);
         }
+        success = true;
       } finally {
         if (!success) {
           for (ai.rapids.cudf.HostColumnVector.ColumnBuilder b: builders) {
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
@@ -331,6 +331,14 @@ public class GpuColumnVector extends GpuColumnVectorBase {
    */
   private static <T> boolean typeConversionAllowed(ColumnViewAccess<T> cv, DataType colType) {
     DType dt = cv.getDataType();
+    if (dt.isDecimalType()) {
+      if (!(colType instanceof DecimalType)) {
+        return false;
+      }
+      // check for overflow
+      int maxPrecision = dt.isBackedByLong() ? DType.DECIMAL64_MAX_PRECISION : DType.DECIMAL32_MAX_PRECISION;
+      return ((DecimalType) colType).precision() <= maxPrecision;
+    }
     if (!dt.isNestedType()) {
       return getRapidsType(colType).equals(dt);
     }
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
@@ -163,9 +163,8 @@ public class GpuColumnVector extends GpuColumnVectorBase {
       return DType.STRING;
     } else if (type instanceof DecimalType) {
       DecimalType decType = (DecimalType) type;
-      if (decType.precision() <= DType.DECIMAL32_MAX_PRECISION) {
-        return DType.create(DType.DTypeEnum.DECIMAL32, -decType.scale());
-      } else if (decType.precision() <= DType.DECIMAL64_MAX_PRECISION) {
+      // Currently, maps all DecimalTypes to DType.DECIMAL64 to reduce the complexity.
+      if (decType.precision() <= DType.DECIMAL64_MAX_PRECISION) {
         return DType.create(DType.DTypeEnum.DECIMAL64, -decType.scale());
       } else {
         return null;
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
@@ -177,10 +177,9 @@ public class GpuColumnVector extends GpuColumnVectorBase {
       DecimalType dt = (DecimalType) type;
       if (dt.precision() > DType.DECIMAL64_MAX_PRECISION) {
         return null;
-      } else if (dt.precision() > DType.DECIMAL32_MAX_PRECISION) {
-        return DType.create(DType.DTypeEnum.DECIMAL64, -dt.scale());
       } else {
-        return DType.create(DType.DTypeEnum.DECIMAL32, -dt.scale());
+        // Map all DecimalType to DECIMAL64, in case of underlying DType transaction.
+        return DType.create(DType.DTypeEnum.DECIMAL64, -dt.scale());
       }
     }
     return null;
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
@@ -161,6 +161,15 @@ public class GpuColumnVector extends GpuColumnVectorBase {
       return DType.TIMESTAMP_MICROSECONDS;
     } else if (type instanceof StringType) {
       return DType.STRING;
+    } else if (type instanceof DecimalType) {
+      DecimalType decType = (DecimalType) type;
+      if (decType.precision() <= DType.DECIMAL32_MAX_PRECISION) {
+        return DType.create(DType.DTypeEnum.DECIMAL32, -decType.scale());
+      } else if (decType.precision() <= DType.DECIMAL64_MAX_PRECISION) {
+        return DType.create(DType.DTypeEnum.DECIMAL64, -decType.scale());
+      } else {
+        return null;
+      }
     }
     return null;
   }
@@ -180,6 +189,48 @@ public class GpuColumnVector extends GpuColumnVectorBase {
       throw new IllegalArgumentException(type + " is not supported for GPU processing yet.");
     }
     return result;
+  }
+
+  static DataType getSparkType(DType type) {
+    switch (type.getTypeId()) {
+      case BOOL8:
+        return DataTypes.BooleanType;
+      case INT8:
+        return DataTypes.ByteType;
+      case INT16:
+        return DataTypes.ShortType;
+      case INT32:
+        return DataTypes.IntegerType;
+      case INT64:
+        return DataTypes.LongType;
+      case FLOAT32:
+        return DataTypes.FloatType;
+      case FLOAT64:
+        return DataTypes.DoubleType;
+      case TIMESTAMP_DAYS:
+        return DataTypes.DateType;
+      case TIMESTAMP_MICROSECONDS:
+        return DataTypes.TimestampType;
+      case STRING:
+        return DataTypes.StringType;
+      case DECIMAL32:
+        return new DecimalType(DType.DECIMAL32_MAX_PRECISION, -type.getScale());
+      case DECIMAL64:
+        return new DecimalType(DType.DECIMAL64_MAX_PRECISION, -type.getScale());
+      default:
+        throw new IllegalArgumentException(type + " is not supported by spark yet.");
+    }
+  }
+
+  protected static <T> DataType getSparkTypeFrom(ColumnViewAccess<T> access) {
+    DType type = access.getDataType();
+    if (type == DType.LIST) {
+      try (ColumnViewAccess<T> child = access.getChildColumnViewAccess(0)) {
+        return new ArrayType(getSparkTypeFrom(child), true);
+      }
+    } else {
+      return getSparkType(type);
+    }
   }
 
   /**
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
@@ -161,14 +161,6 @@ public class GpuColumnVector extends GpuColumnVectorBase {
       return DType.TIMESTAMP_MICROSECONDS;
     } else if (type instanceof StringType) {
       return DType.STRING;
-    } else if (type instanceof DecimalType) {
-      DecimalType decType = (DecimalType) type;
-      // Currently, maps all DecimalTypes to DType.DECIMAL64 to reduce the complexity.
-      if (decType.precision() <= DType.DECIMAL64_MAX_PRECISION) {
-        return DType.create(DType.DTypeEnum.DECIMAL64, -decType.scale());
-      } else {
-        return null;
-      }
     }
     return null;
   }
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
@@ -161,6 +161,17 @@ public class GpuColumnVector extends GpuColumnVectorBase {
       return DType.TIMESTAMP_MICROSECONDS;
     } else if (type instanceof StringType) {
       return DType.STRING;
+    } else if (type instanceof DecimalType) {
+      // Decimal supportable check has been conducted in the GPU plan overriding stage.
+      // So, we don't have to handle decimal-supportable problem at here.
+      DecimalType dt = (DecimalType) type;
+      if (dt.precision() > DType.DECIMAL64_MAX_PRECISION) {
+        return null;
+      } else if (dt.precision() > DType.DECIMAL32_MAX_PRECISION) {
+        return DType.create(DType.DTypeEnum.DECIMAL64, -dt.scale());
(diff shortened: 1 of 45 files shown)


sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
```diff
@@ -158,6 +158,15 @@ private static final DType toRapidsOrNull(DataType type) {
       return DType.TIMESTAMP_MICROSECONDS;
     } else if (type instanceof StringType) {
       return DType.STRING;
+    } else if (type instanceof DecimalType) {
+      DecimalType decType = (DecimalType) type;
+      if (decType.precision() <= DType.DECIMAL32_MAX_PRECISION) {
```
revans2, line 163: Spark is kind of split on how decimals are stored internally.  For `UnsafeRow` only supports a `long` value or a binary representation for `BigDecimal`.

https://github.com/apache/spark/blob/0b557b329046c66ee67a8c94c5bb95ffbe50e135/sql/catalyst/src/main/java/org/apache/spark/sql/catalyst/expressions/UnsafeRow.java#L382-L395

For the build in `ColumnVector` implementation they distinguish between int and long backed storage.

https://github.com/apache/spark/blob/0b557b329046c66ee67a8c94c5bb95ffbe50e135/sql/core/src/main/java/org/apache/spark/sql/execution/vectorized/WritableColumnVector.java#L361-L375

If we only support DECIMAL64 then it would potentially reduce complexity by a lot. But if we also support DECIMAL32 we could save on GPU memory.

The main reason I ask is because I just merged in some code that speeds up the transfer of fixed width data types.  I have not tested it with decimal in any way, but there is code in there that makes some assumptions about what the size of the data types are and I think it is wrong right now.  Should be simple to fix it, but I wanted to call this out and be sure we are all on the same page.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -450,6 +450,7 @@ object GpuOverrides {
       case DateType => true
       case TimestampType => ZoneId.systemDefault().normalized() == GpuOverrides.UTC_TIMEZONE_ID
       case StringType => true
+      case dt: DecimalType if dt.precision <= ai.rapids.cudf.DType.DECIMAL64_MAX_PRECISION => true
```
revans2, line 453: Do we have a list of all of the operators we currently support that also support Decimal in Spark? We have to be sure that if we are allowing decimal for all operators, like this is doing, that we have disallowed it everywhere that we do not have automated tests for.

The only tests I see in this patch are unit tests for Scalars.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
```diff
@@ -201,6 +210,10 @@ static final DataType getSparkType(DType type) {
         return DataTypes.TimestampType;
       case STRING:
         return DataTypes.StringType;
+      case DECIMAL32:
+        return new DecimalType(DType.DECIMAL32_MAX_PRECISION, -type.getScale());
```
jlowe, line 214: Do we need to track the effective precision of the value as it progresses through operations in order to be compatible with how Spark operates on `Decimal`?  For example, take this simple Spark shell session:
```scala
scala> val x = Decimal(5.43)
x: org.apache.spark.sql.types.Decimal = 5.43

scala> (x.precision, x.scale)
res2: (Int, Int) = (3,2)

scala> val y = x * x
y: org.apache.spark.sql.types.Decimal = 29.4849

scala> (y.precision, y.scale)
res3: (Int, Int) = (6,4)
```

If we perform the same operation with `DECIMAL32` we will lose track of the precision, and when we cast back to a SparkType instead of getting `DecimalType(6,4)` as we should, we'll get `DecimalType(9,4)` instead.

revans2: Spark should already be doing that for us, which is another reason why I don't think we want to rely on this API much longer.  We are in the process of moving away from it for nested types and I think it would be best to just not support it for decimal at all.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
sperlingxx: Maybe we can start from only supporting DECIMAL64 for reducing complexity?

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
sperlingxx: Got it. I'll remove decimal parts.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: We will be able to support binary operators for decimal type only if we enable `BinaryOperable` with decimal  on cuDF java layer.
In terms of unary operators (abs, ceil, floor, round, …) and other advanced operators(such as: aggregation operators), we have to wait the support in libcudf level. This refers to https://github.com/rapidsai/cudf/issues/3556.

sperlingxx: I've rewrited `isSupportedType` part. In latest version, general method `isSupportedType` remain unchanged. And a decimal specialized method `isSupportedTypeWithDecimal` was added for GpuExpressions (operators) who support Deimcal. For now, only very limited ones are decimal-enabled, such as: GpuLiteral, GpuAlias. But this collection will be expanded only if  decimal-related dependencies are ready.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
sperlingxx: @revans2 @jlowe  The problem is that `GpuColumnVector.from(cudf.ColumnVector cudfCv)` relies on the method `getSparkType`. If `getSparkType` doesn't support DecimalTypes, we won't be able to create decimal-typed `GpuColumnVector` from `cudf.ColumnVector` through the method `GpuColumnVector.from(cudf.ColumnVector cudfCv)`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: I found it is not easy to add DecimalType into supported type while keeping compatible with existed RapidsMetas. I tried to solve this problem with two actions:
1. Add method `isSupportedTypeWithDecimal` for RapidMeta who overrides the `areAllSupportedTypes` interface, such as: `Literal`, `Cast`.
2. Refactor `GpuOverrides.areAllSupportedTypes`, which receives RapidsMeta itself to determine whether it is decimal supportable. This is for RapidsMeta without  interface overriding of `areAllSupportedTypes`.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
revans2: I am in the process of updating the plugin so `GpuColumnVector.from(cudf.ColumnVector cudfCv)` goes away and is replaced with `GpuColumnVector.from(cudf.ColumnVector cudfCv, DataType)`, similar to what I have been doing for `Table` to `ColumnarBatch` conversion. I have a patch working, just need to finish testing it and a bit more cleanup.  Because no one has taken a look at https://github.com/NVIDIA/spark-rapids/pull/1078 yet, I'll likely just update it instead of making a separate PR.

The goal eventually is to have `getSparkType` go away entirely because we cannot support it when there is not a 1:1 mapping between cudf types and spark types.

sperlingxx: I've refined this PR on the basis of the patch #1078. I appended decimal type check into `typeConversionAllowed`.

tests/src/test/scala/com/nvidia/spark/rapids/unit/DecimalUnitTest.scala
```diff
@@ -0,0 +1,180 @@
...
+    // assertion error throws while running `typeConversionAllowed` check
+    assertThrows[AssertionError] {
+      withResource(GpuColumnVector.from(ColumnVector.decimalFromLongs(0, 1L),
+        DecimalType(DType.DECIMAL64_MAX_PRECISION + 1, 0))) { _ => }
+    }
+    assertThrows[AssertionError] {
+      withResource(GpuColumnVector.from(ColumnVector.decimalFromInts(0, 1),
+        DecimalType(DType.DECIMAL32_MAX_PRECISION + 1, 0))) { _ => }
+    }
+    // TODO: support fromScalar(cudf.ColumnVector cv, int rows) for fixed-point decimal in cuDF
```
revans2, line 116: This is going to be a big problem.  We need this for a project and other operations to work.

sperlingxx: I will support this feature ASAP. And I really want to know what other missing essentials for this PR. Thanks!

revans2: What we really have to have/test is 

1) Shuffle decimal types including Contiguous Split, UCX Headers Updated, JCudfSerialzation.  I think all of these have been updated, but not sure if they have been tested yet.
2) CoalesceBatch for decimal types.  This includes Contiguous Split and Table.concat.  Again I think these are in place, but it needs to be tested.
3) ColumnarToRow and RowToColumnar.  I think that is covered here for the most part, but it is not tested.  I would like to have a follow on issue to try and "fix" some of the issues with the fast fixed width transfers, which I am happy to handle.
4) `fromScalar` support. This is something that we are missing but I didn't realize how important it is until recently.  We have a lot of Expressions that require this so we can take a constant value and turn it into a column, which happens regularly.

sperlingxx: Hi @revans2 , thanks for outlining all essential work for basic decimal supporting. I am trying to address them these days, but I found it is not easy to accomplish all these targets at one time. 
Here is a summary of current progress:
* Enable decimal type for valid `RapidsMeta` in `GpuOverrides` (according to latest supportedType mechanism).
* support decimal for `GpuColumnVector` (with test cases in DecimalUnitTest.scala)
* support decimal for `GpuScalar`  (with test cases in DecimalUnitTest.scala)
* decimal converters for `GpuRowToColumnarExec` (with test cases in GpuBatchUtilsSuite.scala)
* decimal columnar copy for `HostColumnarToGpu` (with test cases in DecimalUnitTest.scala)
* test cases for decimal in `GpuCoalesceBatchesSuite` (Some cases fails because contiguous split of decimal column has yet supported by cuDF. I've created a [PR](https://github.com/rapidsai/cudf/pull/6762) to solve this problem.)
* Creating decimal column vector from scalar will be supported after this [PR](https://github.com/rapidsai/cudf/pull/6723) merged (which is close to).

Could we get this PR merged at first, in order not to block other work relying on this (such as: binary operation support) ?  Then, we address unsolved issues in other PRs. Perhaps we need to create an issue to summarize the progress of all these items?

revans2: I am fine with breaking things up, but I really want to see something that is self-contained and tested.

If we drop support for decimal in literal values then we don't need `fromScalar` to work yet.  But any code that we do put in I want to see at least minimal automated tests that it is working.  I trust that you have run tests manually, but there is a lot that is changing right now with nested types etc, and I want to be sure that we have tests to verify that someone else didn't break something.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -503,6 +503,14 @@ object GpuOverrides {
     }
   }
 
+  /**
+   * A walkaround method to include DecimalType for expressions who supports Decimal.
+   */
+  def isSupportedTypeWithDecimal(dataType: DataType): Boolean = dataType match {
```
revans2, line 509: Should this be deleted?

sperlingxx: My bad. It's removed.

tests/src/test/scala/com/nvidia/spark/rapids/unit/DecimalUnitTest.scala
sperlingxx: Sorry, I am afraid that I don't get what the "minimal automated tests" should be, besides existed unit tests for decimal. Does it refer to tests with mixed data types, such as `GpuSinglePartitioningSuite`? If it does, I am working on them. I've just put a [PR](https://github.com/rapidsai/cudf/pull/6770) to enable creating decimal column with `Table.TestBuilder` .

sperlingxx: I've just added tests for _binaryOps_ and _project_.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
revans2: I am fine either way, and it is not a big deal. I was just thinking out loud.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
```diff
@@ -289,6 +289,32 @@ private object GpuRowToColumnConverter {
...
+      (if (precision > ai.rapids.cudf.DType.DECIMAL32_MAX_PRECISION) 8 else 4) + VALIDITY
+    }
+  }
+
+  private class NotNullDecimalConverter(precision: Int, scale: Int) extends TypeConverter {
+    override def append(
+        row: SpecializedGetters,
+        column: Int,
+        builder: ai.rapids.cudf.HostColumnVector.ColumnBuilder): Double = {
+      builder.append(row.getDecimal(column, precision, scale).toJavaBigDecimal)
```
revans2, line 312: I would prefer to see us use `toUnscaledLong` to avoid any extra object creation, it is also what we are going to ultimately store the data as.  This might mean that we need different classes for DECIMAL64 and DECIMAL32 too.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
```diff
@@ -289,6 +289,32 @@ private object GpuRowToColumnConverter {
...
 
+  private class DecimalConverter(precision: Int, scale: Int) extends TypeConverter {
+    override def append(
+        row: SpecializedGetters,
+        column: Int,
+        builder: ai.rapids.cudf.HostColumnVector.ColumnBuilder): Double = {
+      if (row.isNullAt(column)) {
+        builder.appendNull()
+      } else {
+        new NotNullDecimalConverter(precision, scale).append(row, column, builder)
```
revans2, line 300: This is on the data path.  I would like us to avoid object creation if at all possible to speed up the data path.  Please make a static method instead, or use inheritance, which I think is less ideal.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
```diff
@@ -289,6 +289,32 @@ private object GpuRowToColumnConverter {
...
+  }
+
+  private class NotNullDecimalConverter(precision: Int, scale: Int) extends TypeConverter {
+    override def append(
+        row: SpecializedGetters,
+        column: Int,
+        builder: ai.rapids.cudf.HostColumnVector.ColumnBuilder): Double = {
+      builder.append(row.getDecimal(column, precision, scale).toJavaBigDecimal)
+      // Infer the storage type via precision, because we can't access DType of builder.
+      if (precision > ai.rapids.cudf.DType.DECIMAL32_MAX_PRECISION) 8 else 4
```
revans2, line 314: Here too it would be great to avoid conditionals in the data path.  I would prefer it if we either passed in the size or had separate implementations for DECIMAL32 and DECIMAL64

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -434,6 +434,11 @@ abstract class SparkPlanMeta[INPUT <: SparkPlan](plan: INPUT,
   override val childParts: Seq[PartMeta[_]] = Seq.empty
   override val childDataWriteCmds: Seq[DataWritingCommandMeta[_]] = Seq.empty
 
+  // We assume that all common plans are decimal supportable by default, considering
+  // whether decimal allowable is mainly determined in expression-level.
+  override def isSupportedType(t: DataType): Boolean =
+    GpuOverrides.isSupportedType(t, allowDecimal = true)
```
revans2, line 440: Do we have tests that verify that we can support decimal for all top level spark operations?   Have we tested join, expand, generate, filter, project, union, window, sort, or hash agregate?  What about all of the arrow python UDF code where we go to/from arrow?

I think it would be much better if we split this big PR up into smaller pieces and put each piece in separately with corresponding tests to show that it works, and we only add decimal to the allow list for those things that we know it works for because we have tested it.  If you want me to help with this I am happy to do it.  I am already in the middle of doing it for Lists I am going to add in structs, maps, binary, null type and finally calendar interval based off of how much time I have and priorities.  Some of these we will only be able to do very basic things with, but that should be enough to unblock others for using them for more complicated processing.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -144,6 +144,8 @@ object GpuDivModLike {
       case DType.INT64 => Scalar.fromLong(0L)
       case DType.FLOAT32 => Scalar.fromFloat(0f)
       case DType.FLOAT64 => Scalar.fromDouble(0)
+      case dt if dt.isDecimalType && dt.isBackedByInt => Scalar.fromDecimal(0, 0)
```
revans2, line 147: I'm not sure this is going to work in all cases.  I think this might be another place were we have some tech debt to pay off and need to pass in a DataType instead of a DType.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -144,6 +144,8 @@ object GpuDivModLike {
...
+      case dt if dt.isDecimalType && dt.isBackedByLong => Scalar.fromDecimal(0, 0L)
```
revans2, line 148: For the div/mod to work properly we also need to update `isScalarZero`, or we are going to miss a divide by zero case in decimal.

tests/src/test/scala/com/nvidia/spark/rapids/DecimalBinaryOpSuite.scala
```diff
@@ -0,0 +1,212 @@
...
+
+import org.apache.spark.sql.rapids.{GpuAdd, GpuEqualTo, GpuGreaterThan, GpuGreaterThanOrEqual, GpuLessThan, GpuLessThanOrEqual, GpuMultiply, GpuSubtract}
+import org.apache.spark.sql.types.{DataType, DataTypes, Decimal, DecimalType, StructType}
+
+/*
+  unsupported operators are as below:
+  - GpuEqualNullSafe
+  - GpuDivModLike (GpuDivide/GpuIntegralDivide/GpuPmod/GpuRemainder)
+*/
+class DecimalBinaryOpSuite extends GpuExpressionTestSuite {
```
revans2, line 29: I personally would prefer to see us update the python tests.  They tend to be more complete.  These tests are not covering the case where you need to take a literal and turn it into a column because https://github.com/rapidsai/cudf/pull/6723 has not been merged in yet.

This is another reason to break down this into smaller pieces.  I cannot let us put in supporting for binary ops without the above PR being merged in and tested.  If we break this apart we can get some of the functionality in, and then we can look at what we need to do to get the rest of it in.

PR #1064, opened 2020-11-04
Use array instead of List for from(Table, DataType)

revans2:
Two changes that were both merged in recently conflicted with one another.  One changed an API that the other used.  This fixes the build.


revans2:
build

kuhushukla (approved):

jlowe (approved):

PR #1066, opened 2020-11-04
More fixes for conversion to ColumnarBatch

revans2:
This is getting to be a rather large change and there is a lot more on the way, so I am putting in some of the changes now and will keep working on more of them as this is reviewed.

The issue is that there is a many to one mapping of Spark Types to CUDF types. So we can no longer take a `cudf.Table` and turn it into a `ColumnarBatch`, or a `cudf.ColumnVector` and turn it into a `GpuColumnVector`.

The is going through API by API to try and replace them with versions that take the Spark DataType.

### DIFF
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuCompressedColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuCompressedColumnVector.java
@@ -21,7 +21,12 @@ import ai.rapids.cudf.DeviceMemoryBuffer;
 import com.nvidia.spark.rapids.format.ColumnMeta;
 import com.nvidia.spark.rapids.format.SubBufferMeta;
 import com.nvidia.spark.rapids.format.TableMeta;
+import org.apache.spark.sql.types.ArrayType;
+import org.apache.spark.sql.types.BinaryType;
 import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.MapType;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
 import org.apache.spark.sql.vectorized.ColumnVector;
 import org.apache.spark.sql.vectorized.ColumnarBatch;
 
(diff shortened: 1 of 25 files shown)


revans2:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuCompressedColumnVector.java
```diff
@@ -109,6 +111,44 @@ public static ColumnarBatch from(DeviceMemoryBuffer compressedBuffer, TableMeta
...
+    ColumnMeta columnMeta = new ColumnMeta();
+    int numColumns = tableMeta.columnMetasLength();
+    assert numColumns == colTypes.length : "Size mismatch on types";
+    ColumnVector[] columns = new ColumnVector[numColumns];
+    try {
+      for (int i = 0; i < numColumns; ++i) {
+        tableMeta.columnMetas(columnMeta, i);
+        DataType type = colTypes[i];
+        // TODO we need a good way to verify that we can map between the DType(s) we have in
+        //  columnMeta and the types passed in.
```
jlowe, line 136: Can't we call `GpuColumnVector.getRapidsType` and verify it matches the DType in the `ColumnMeta`?

jlowe (approved):
Just a comment about a TODO comment that was added, but otherwise looks OK to me.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuCompressedColumnVector.java
revans2: That does not work for nested types. I could do some checks and do it for non-nested types, but I will write something to make this work properly.

revans2:
build

revans2:

jlowe (dismissed):

revans2:
build

revans2:
build

jlowe (approved):

PR #1069, opened 2020-11-05
Accelerate the data transfer between JVM and Python for the plan 'GpuWindowInPandasExec'

firestarman:
This PR aims at accelerating the data transfer between JVM and Python for the plan 'GpuWindowInPandasExec' by leveraging the cuDF. Instead of going from cuDF to rows to arrow, we can go from cuDF to arrow directly and vice versa, to avoid transformations between row and arrow.

This PR includes
- Removing the `GpuWindowInPandasExec ` in shims since it does not extends from `WindowExecBase `any more.
- Processing the incoming batches for the window function execution on Python side. 
  - Split the big incoming batches into small batches where each contains only one group data,
  - Calculate the window frame bounds on each group batch. Now it is OK to send the batches to the Python side one by one.
- Adding more integration tests for windowing.


firestarman:
build

firestarman:
Mark as WIP to verify the actual integration tests first.

firestarman:
build

firestarman:
rebased onto the top of branch-0.3.

firestarman:
build

firestarman:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala
```diff
@@ -0,0 +1,570 @@
...
+                splitBatches.head
+              }
+            }
+          }
+        } else {
+          // Only one group, return it directly
+          batch
+        }
+        // Shall we release the semaphore here ?
+        GpuSemaphore.releaseIfNecessary(TaskContext.get())
```
revans2, line 165: We only want to release it if the data is off of the GPU.  This is why we have a callback in the cudf API that writes the data out over the wire so you know when it is done copying the data to the CPU before it starts to write it out over the wire.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    // Collect all valid window expressions and group them by their frame.
+    windowExpression.foreach { e =>
+      e.foreach {
+        case e @ GpuWindowExpression(function, spec) =>
+          val frame = spec.frameSpecification.asInstanceOf[GpuSpecifiedWindowFrame]
+          function match {
+            case GpuAggregateExpression(_, _, _, _, _) => collect("AGGREGATE", frame, e)
+            case _: GpuAggregateWindowFunction => collect("AGGREGATE", frame, e)
```
revans2, line 266: Do we actually support these functions here?

revans2:
I am a little curious why we don't need the shim layers any more.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -190,16 +192,27 @@ case class GpuWindowExpression(windowFunction: Expression, windowSpec: GpuWindow
...
   private lazy val boundRowProjectList = windowSpec.partitionSpec ++
-      windowFunc.windowInputProjection
+      windowInputProjection
   private lazy val boundRangeProjectList = windowSpec.partitionSpec ++
       windowSpec.orderSpec.map(_.child.asInstanceOf[GpuExpression]) ++
-      windowFunc.windowInputProjection
+      windowInputProjection
 
   override def columnarEval(cb: ColumnarBatch) : Any = {
+    if (windowFunc == null) {
```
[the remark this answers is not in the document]
firestarman: @mythrocks Do you have any suggestion for this ?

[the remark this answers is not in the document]
firestarman: Per my understanding, if it is a Python UDF, it should not go here, since Spark will create a separate physical plan for Python UDF.
Here is just a double check, i think.

firestarman:
> I am a little curious why we don't need the shim layers any more.

It is the class `WindowExecBase` is different between spark 310 and 300, from which our current implementation extends.

In this PR, the `GpuWindowInPandasExec` does not extends from  `WindowExecBase` any more, so no need shim layers.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala
firestarman: For CPU version, this logic is also used by `WindowExec` besides `WindowInPandasExec`.
For GPU version currently, is it only used by `GpuWindowInPandasExec `.
We may support it in our `GpuWindowExec` (@mythrocks pls correct me if wrong), but not in `GpuWindowInPandasExec `, so i think it is ok to remove this if you prefer.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala
firestarman: Got it, thanks, i think here we should release the semaphore, since the data has been off of the GPU already.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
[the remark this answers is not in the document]
firestarman: i think i found a better way, updated.

firestarman:
build

firestarman:
rebased

firestarman:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala
revans2: Sorry no.  The batch being returned (resultBatch) is still on the GPU so you cannot release the semaphore until it is off the GPU. You also do not want to hold the lock while doing any I/O which is why I said that we added in a callback in the cudf API.

https://github.com/NVIDIA/spark-rapids/blob/d865883ee35034f119e4e014ad1e89d354688907/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuArrowEvalPythonExec.scala#L426

Is the place where the semaphore is released properly and this should be removed.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala
revans2: OK So I read the original Spark code and this is not anywhere in that code. They do not support anything but a PythonUDF

https://github.com/apache/spark/blob/1eb236b9360a000afc30424341698fe26ee96d0f/sql/core/src/main/scala/org/apache/spark/sql/execution/python/WindowInPandasExec.scala#L206

We should do the same and delete all of the extra code, and possibly throw an exception if we see anything else.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala
firestarman: Shall i also remove this line for semaphore acquiring? https://github.com/NVIDIA/spark-rapids/blob/d865883ee35034f119e4e014ad1e89d354688907/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala#L117 

I release this semaphore here because i acquire one before executing the computation as above.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala
firestarman: Yes, we have this too, it is here https://github.com/NVIDIA/spark-rapids/blob/d865883ee35034f119e4e014ad1e89d354688907/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala#L422 

The code about collecting the functions is similar with `windowFrameExpressionFactoryPairs ` in  `WindowExecBase` https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/window/WindowExecBase.scala#L118 ,
which is used to filter the Python UDF as you mentioned.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala
revans2: You are right you don't need to acquire the semaphore either.  The semaphore is about GPU memory, not computation.  You grab it when you are going to put something on the GPU and you release it when you are off the GPU or everything on the GPU can be spilled.  Typically you don't need to worry about this because there are only a few places where this happens, but this is one of those cases.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala
firestarman: I move this logic into the `GpuWindowInPandasExec`, then our implementation does not need to extends from `WindowExecBase` any more.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala
firestarman: ok, removed, thanks.

firestarman:
build

revans2 (commented):
The changes look good my only concern right now is documentation.  We are moveing from an implementation that supports all window operations to a version that does not, but is faster for the ones that it does.  This is off by default, but if someone tries to use this with a range based query it will fail.  For regular window functions we fall back to the CPU, but we cannot do that here.  I am OK with that, but we need to update the documentation so it is clear to a customer that this experimental feature is broken in this one area.

firestarman:
> The changes look good my only concern right now is documentation. We are moveing from an implementation that supports all window operations to a version that does not, but is faster for the ones that it does. This is off by default, but if someone tries to use this with a range based query it will fail. For regular window functions we fall back to the CPU, but we cannot do that here. I am OK with that, but we need to update the documentation so it is clear to a customer that this experimental feature is broken in this one area.

Hi, for `RangeFrame`, it will fall back to the CPU as well, here is the check https://github.com/NVIDIA/spark-rapids/blob/7e6c3d180263539941f86dd158615ad6d163e00e/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExec.scala#L72 , and i will update the doc accordingly.

  


>For regular window functions we fall back to the CPU, but we cannot do that here.

I do not fully get it, Could you explain more ? 
Since as far as I know, regular window functions such as `RowNumber`, `Rank ` are executed in the plan `GpuWindowExec`, while `GpuWindowInPandasExec` is dedicated to Pandas (Python) UDF over window. Seems we do not need to consider the regular window functions in this plan.

firestarman:
rebased

firestarman:
build

firestarman:
updated the config doc

firestarman:
build

revans2 (approved):

revans2:
> Sorry I do not fully get it, Could you explain more ?

Actually you understood the code better than I did. I thought we were not falling back to the CPU for range queries on python UDFS, but I was wrong, and you updated the documentation so it all looks good now.

PR #1074, opened 2020-11-05
Improve AQE transition logic

andygrove:

The `insertColumnarToGpu` method inserts a `HostColumnarToGpu` transition for plans that are columnar but not on the GPU. This logic was not working correctly for query stages when AQE is enabled because we don't (and cannot) replace query stages with a GPU version. They simply wrap a GPU plan. We had logic to later remove these `HostColumnarToGpu` transitions that were incorrectly inserted.

This PR fixes `insertColumnarToGpu` so that it does not insert unnecessary transitions for GPU query stages.


andygrove:
build

andygrove:
build

Issue #1073, opened 2020-11-05
[BUG] java.lang.RuntimeException: BinaryExpressions must override either eval or nullSafeEval


martinstuder:
**What is your question?**

I'm running into the following exception and I'm wondering whether this is an indication of an unsupported binary expression type (I'm running rapids-4-spark v0.2.0) or whether this is a potential indication of a bug:
```
Py4JJavaError: An error occurred while calling o29050.checkpoint.
: org.apache.spark.SparkException: Job aborted due to stage failure: Task 8 in stage 1630.0 failed 4 times, most recent failure: Lost task 8.3 in stage 1630.0 (TID 331268, 10.50.24.17, executor 6): java.lang.RuntimeException: BinaryExpressions must override either eval or nullSafeEval
	at scala.sys.package$.error(package.scala:30)
	at org.apache.spark.sql.catalyst.expressions.BinaryExpression.nullSafeEval(Expression.scala:621)
	at com.nvidia.spark.rapids.CudfBinaryOperator.com$nvidia$spark$rapids$GpuBinaryExpression$$super$nullSafeEval(GpuExpressions.scala:236)
	at com.nvidia.spark.rapids.GpuBinaryExpression.columnarEval(GpuExpressions.scala:187)
	at com.nvidia.spark.rapids.GpuBinaryExpression.columnarEval$(GpuExpressions.scala:164)
	at com.nvidia.spark.rapids.CudfBinaryOperator.columnarEval(GpuExpressions.scala:236)
	at com.nvidia.spark.rapids.RapidsPluginImplicits$ReallyAGpuExpression.columnarEval(implicits.scala:35)
	at com.nvidia.spark.rapids.GpuBinaryExpression.columnarEval(GpuExpressions.scala:169)
	at com.nvidia.spark.rapids.GpuBinaryExpression.columnarEval$(GpuExpressions.scala:164)
	at com.nvidia.spark.rapids.CudfBinaryOperator.columnarEval(GpuExpressions.scala:236)
	at com.nvidia.spark.rapids.RapidsPluginImplicits$ReallyAGpuExpression.columnarEval(implicits.scala:35)
	at com.nvidia.spark.rapids.GpuFilter$.apply(basicPhysicalOperators.scala:119)
	at com.nvidia.spark.rapids.GpuFilter$.$anonfun$apply$1(basicPhysicalOperators.scala:105)
	at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
[...]

PR #1078, opened 2020-11-06
Another round of fixes for mapping of DataType to DType

revans2:
Getting closer, but there is still a lot more to do.

One thing to note is that for part of this I changed `Gpu{Unary,Binary,Trinary}Expression` to return a `ColumnVector` instead of a `GpuColumnVector` this let me put the conversion in a single location and also fixed some issues with some `UnaryExpressions` that couldn't know the right spark type to return because they were going to be cast to the final type shortly afterwards.

Actually with the last push this should completely fix the issues with not doing the conversion correctly in all cases.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuHashPartitioning.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuHashPartitioning.scala
@@ -89,6 +89,7 @@ case class GpuHashPartitioning(expressions: Seq[Expression], numPartitions: Int)
     try {
       gpuKeyColumns = getGpuKeyColumns(batch)
       gpuDataColumns = getGpuDataColumns(batch)
+      val sparkTypes = gpuDataColumns.map(_.dataType())
 
       val (keys, dataIndexes, table) = dedupe(gpuKeyColumns, gpuDataColumns)
       // Don't need the batch any more table has all we need in it.
@@ -102,8 +103,10 @@ case class GpuHashPartitioning(expressions: Seq[Expression], numPartitions: Int)
       val partedTable = table.onColumns(keys: _*).hashPartition(numPartitions)
       table.close()
       val parts = partedTable.getPartitions
-      val columns = dataIndexes.map { idx =>
-        GpuColumnVector.from(partedTable.getColumn(idx).incRefCount())
+      val columns = dataIndexes.zip(sparkTypes).map { pair =>
+        val idx = pair._1
+        val sparkType = pair._2
+        GpuColumnVector.from(partedTable.getColumn(idx).incRefCount(), sparkType)
       }
       partedTable.close()
       (parts, columns)
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuHashPartitioning.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuHashPartitioning.scala
@@ -103,9 +103,7 @@ case class GpuHashPartitioning(expressions: Seq[Expression], numPartitions: Int)
       val partedTable = table.onColumns(keys: _*).hashPartition(numPartitions)
       table.close()
       val parts = partedTable.getPartitions
-      val columns = dataIndexes.zip(sparkTypes).map { pair =>
-        val idx = pair._1
-        val sparkType = pair._2
+      val columns = dataIndexes.zip(sparkTypes).map { case (idx, sparkType) =>
         GpuColumnVector.from(partedTable.getColumn(idx).incRefCount(), sparkType)
       }
       partedTable.close()
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRoundRobinPartitioning.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRoundRobinPartitioning.scala
@@ -43,19 +43,17 @@ case class GpuRoundRobinPartitioning(numPartitions: Int)
     val sparkTypes = GpuColumnVector.extractTypes(batch)
     withResource(GpuColumnVector.from(batch)) { table =>
       if (numPartitions == 1) {
-        val columns = (0 until table.getNumberOfColumns).zip(sparkTypes).map { pair =>
-          val idx = pair._1
-          val sparkType = pair._2
-          GpuColumnVector.from(table.getColumn(idx).incRefCount(), sparkType)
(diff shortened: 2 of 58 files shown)


revans2:
build

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuHashPartitioning.scala
```diff
@@ -102,8 +103,10 @@ case class GpuHashPartitioning(expressions: Seq[Expression], numPartitions: Int)
       val partedTable = table.onColumns(keys: _*).hashPartition(numPartitions)
       table.close()
       val parts = partedTable.getPartitions
-      val columns = dataIndexes.map { idx =>
-        GpuColumnVector.from(partedTable.getColumn(idx).incRefCount())
+      val columns = dataIndexes.zip(sparkTypes).map { pair =>
+        val idx = pair._1
+        val sparkType = pair._2
```
jlowe, line 108:
```suggestion
      val columns = dataIndexes.zip(sparkTypes).map { case (idx, sparkType) =>
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRoundRobinPartitioning.scala
```diff
@@ -40,24 +40,25 @@ case class GpuRoundRobinPartitioning(numPartitions: Int)
...
-    val table: Table = GpuColumnVector.from(batch)
-    try {
+    val sparkTypes = GpuColumnVector.extractTypes(batch)
+    withResource(GpuColumnVector.from(batch)) { table =>
       if (numPartitions == 1) {
-        val columns = (0 until table.getNumberOfColumns).map(
-          idx => GpuColumnVector.from(table.getColumn(idx).incRefCount())).toArray
+        val columns = (0 until table.getNumberOfColumns).zip(sparkTypes).map { pair =>
+          val idx = pair._1
+          val sparkType = pair._2
```
jlowe, line 48:
```suggestion
        val columns = (0 until table.getNumberOfColumns).zip(sparkTypes).map { case (idx, sparkType) =>
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRoundRobinPartitioning.scala
```diff
@@ -40,24 +40,25 @@ case class GpuRoundRobinPartitioning(numPartitions: Int)
...
       }
-      val partedTable = table.roundRobinPartition(numPartitions, getStartPartition)
-      try {
+      withResource(table.roundRobinPartition(numPartitions, getStartPartition)) { partedTable =>
         val parts = partedTable.getPartitions
-        val columns = (0 until partedTable.getNumberOfColumns.toInt).map(
-          idx => GpuColumnVector.from(partedTable.getColumn(idx).incRefCount())).toArray
+        val columns = (0 until partedTable.getNumberOfColumns.toInt).zip(sparkTypes).map { pair =>
+          val idx = pair._1
+          val sparkType = pair._2
```
jlowe, line 57:
```suggestion
        val columns = (0 until partedTable.getNumberOfColumns.toInt).zip(sparkTypes).map {
        case (idx, sparkType) =>
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -83,14 +83,16 @@ trait GpuBaseLimitExec extends LimitExec with GpuExec {
           try {
             if (numColumns > 0) {
               table = GpuColumnVector.from(batch)
-              (0 until numColumns).foreach(i => {
+              (0 until numColumns).zip(output).foreach{ pair =>
+                val i = pair._1
+                val sparkType = pair._2.dataType
```
jlowe, line 88:
```suggestion
              (0 until numColumns).zip(output).foreach{ case (i, outAttr) =>
                val sparkType = outAttr.dataType
```

tests/src/test/scala/com/nvidia/spark/rapids/GpuPartitioningSuite.scala
```diff
@@ -41,7 +41,12 @@ class GpuPartitioningSuite extends FunSuite with Arm {
 
   private def buildSubBatch(batch: ColumnarBatch, startRow: Int, endRow: Int): ColumnarBatch = {
     val columns = GpuColumnVector.extractBases(batch)
-    val sliced = columns.safeMap(c => GpuColumnVector.from(c.subVector(startRow, endRow)))
+    val types = GpuColumnVector.extractTypes(batch)
+    val sliced = columns.zip(types).map { pair =>
+      val c = pair._1
+      val t = pair._2
```
jlowe, line 47:
```suggestion
    val sliced = columns.zip(types).map { case (c, t) =>
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/nullExpressions.scala
```diff
@@ -91,11 +85,13 @@ case class GpuCoalesce(children: Seq[Expression]) extends GpuExpression with
       })
 
       if (runningResult != null) {
-        runningResult.incRefCount()
+        GpuColumnVector.from(runningResult.incRefCount(), dataType)
       } else if (runningScalar != null) {
-        GpuScalar.extract(runningScalar)
+        GpuColumnVector.from(runningScalar, batch.numRows(), dataType)
```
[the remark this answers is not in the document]
revans2: I thought it might be related to #1073 but after thinking about it more I am going to "fix" it in another way.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -56,7 +53,7 @@ case class GpuUnaryPositive(child: Expression) extends GpuUnaryExpression
 
   override def sql: String = s"(+ ${child.sql})"
 
-  override def doColumnar(input: GpuColumnVector) : GpuColumnVector = input
+  override def doColumnar(input: GpuColumnVector) : ColumnVector = input.getBase.incRefCount()
```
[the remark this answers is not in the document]
revans2: Sorry that was a bug.  This code never gets called in practice. `UnaryPositive` always becomes a noop and is filtered out of the plan.  We have it here just because.  The contract for `doColumnar` in this case is that the caller will close input, so you don't have to worry about it.

revans2:
build

jlowe (approved):

Issue #1048, opened 2020-10-30
[FEA] Benchmark runner should upload summary file and Spark event log to s3


andygrove:
**Is your feature request related to a problem? Please describe.**
When running the benchmarks in k8s, the summary JSON file and Spark event log get produced in the Spark driver Docker container, making it difficult to script uploading these files to s3 for later analysis.

**Describe the solution you'd like**
I would like some additional command-line arguments on the benchmark runner so that I can provide an S3 endpoint and credentials for uploading the summary file to s3.

**Describe alternatives you've considered**
None

**Additional context**
N/A

andygrove:
The benchmark runner can now upload the summary file. It does not upload the event log so we should open a new issue for that if and when it is needed.

PR #1081, opened 2020-11-06
Add option to upload benchmark summary JSON file

andygrove:

Adds option to upload benchmark summary JSON file to any supported hadoop filesystem, such as s3.

It requires the appropriate `spark.hadoop.fs` configuration options to be set, and that credentials are available at runtime, such as by setting environment variables for AWS keys.

This partially addresses https://github.com/NVIDIA/spark-rapids/issues/1048

### DIFF
--- a/integration_tests/src/main/python/benchmark.py
+++ b/integration_tests/src/main/python/benchmark.py
@@ -91,8 +91,6 @@ def main():
                         help='Whether to call System.gc between iterations')
     parser.add_argument('--upload-uri', required=False,
                         help='Upload URI for summary output')
-    parser.add_argument('--upload-path', required=False,
-                        help='Upload path for summary output')
 
     args = parser.parse_args()
 
@@ -132,9 +130,6 @@ def main():
 
             if args.upload_uri is not None:
                 cmd.append("--upload-uri " + args.upload_uri)
-
-            if args.upload_path is not None:
-                cmd.append("--upload-path " + args.upload_path)
 
             if args.iterations is None:
                 cmd.append("--iterations 1")
--- a/integration_tests/src/main/python/benchmark.py
+++ b/integration_tests/src/main/python/benchmark.py
@@ -89,8 +89,6 @@ def main():
                         help='The number of iterations to run (defaults to 1)')
     parser.add_argument('--gc-between-runs', required=False, action='store_true',
                         help='Whether to call System.gc between iterations')
-    parser.add_argument('--s3-endpoint-url', required=False,
-                        help='S3 endpoint URL (optional)')
     parser.add_argument('--upload-uri', required=False,
                         help='Upload URI for summary output')
     parser.add_argument('--upload-path', required=False,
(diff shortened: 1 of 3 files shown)


andygrove:
@jlowe Could you review when you have a chance? This is the re-implementation of file upload to use Hadoop directly rather than the AWS dependency.

integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/BenchmarkRunner.scala
```diff
@@ -95,6 +100,32 @@ object BenchmarkRunner {
...
+
+              // because we are using the hadoop library directly, we need to convert the spark
+              // hadoop configuration settings into plain hadoop settings
+              val hadoopConf = new Configuration()
+              val sparkHadoopPrefix = "spark.hadoop."
+              spark.sqlContext.getAllConfs
+                  .filter(_._1.startsWith(sparkHadoopPrefix + "fs"))
+                  .foreach { entry =>
+                    hadoopConf.set(entry._1.substring(sparkHadoopPrefix.length), entry._2)
+                  }
```
jlowe, line 119:
```suggestion
              val hadoopConf = spark.sparkContext.hadoopConfiguration
```

integration_tests/src/main/python/benchmark.py
```diff
@@ -89,6 +89,10 @@ def main():
                         help='The number of iterations to run (defaults to 1)')
     parser.add_argument('--gc-between-runs', required=False, action='store_true',
                         help='Whether to call System.gc between iterations')
+    parser.add_argument('--upload-uri', required=False,
+                        help='Upload URI for summary output')
+    parser.add_argument('--upload-path', required=False,
+                        help='Upload path for summary output')
```
jlowe, line 95: Why do we need to specify the URI and path separately?  It will be very confusing/surprising if they specify a path without a URI and it silently does nothing.

IMO the URI should be the path.  For example, `hdfs://somenamenode/path/to/where/reports/go` or `s3://somebucket/path/to/reports` or `file:/some/local/reports/dir`.  I'm not seeing the benefit of having a URI and a path when a URI can be a fully qualified path.

andygrove: Thanks @jlowe I have made that change and tested it out locally. That is definitely simpler.

andygrove:
build

integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/BenchmarkRunner.scala
```diff
@@ -95,6 +100,24 @@ object BenchmarkRunner {
...
+          case Success(report) =>
+            if (conf.uploadUri.isSupplied) {
+              println(s"Uploading ${report.filename} to " +
+                  s"${conf.uploadUri()}/${report.filename}")
+
+              val hadoopConf = spark.sparkContext.hadoopConfiguration
+              val fs = FileSystem.newInstance(new URI(conf.uploadUri()), hadoopConf)
+              fs.copyFromLocalFile(
+                new Path(report.filename),
+                new Path(s"${conf.uploadUri()}/${report.filename}"))
```
jlowe, line 115: Probably best to use the existing `Path(parent, child)` constructor here to handle cases where the parent ends in a path terminator already.
```suggestion
                new Path(conf.uploadUri(), report.filename))
```

jlowe (commented):
This will need to be updated to latest branch-0.3 to get the premerge CI to pass.

andygrove:
build

jlowe (approved):

PR #1022, opened 2020-10-26
Run parallelism pytests on spark local mode

NvTimLiu:
Refer to https://github.com/NVIDIA/spark-rapids/pull/966

Shot at making integration tests run faster by running parallelism pytests for spark local mode.

Move spark local mode conf of pytest parallelism into confMap

Add maven parameter '-Dpytest.RAP_TEST_PARALLEL=xx' to override the parallelism

PR #1086, opened 2020-11-07
Speed up test performance using pytest-xdist

revans2:
This covers the same work that was done in #1022 but finishes it.

It now will also autodetect the parallelism by default so we can ideally run as fast as possible in the normal case.

It also falls back to slower processing if pytest-xdist cannot be found.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -52,16 +52,33 @@ default version runs again Spark 3.0.0, to run against other version use one of 
 Integration tests are stored in the [integration_tests](../integration_tests/README.md) directory.
 There are two frameworks used for testing. One is based off of pytest and pyspark in the 
 `src/main/python` directory. These tests will run as a part of the build if you have the environment
-variable `SPARK_HOME` set.  If you have` SPARK_CONF_DIR` also set the tests will try to use
-whatever cluster you have configured.
+variable `SPARK_HOME` set.
 
-To run the tests separate from the build go to the `integration_tests` directory and submit
-`run_tests.py` through `spark-submit`.  Be sure to include the necessary jars for the RAPIDS
-plugin either with `spark-submit` or with the cluster when it is 
+By default the tests try to use the python packages `pytest-xdist` and `findspark` to oversubscribe
+your GPU and run the tests in spark local mode. This can speed up theses tests significantly as all
+of the tests that run by default process relatively small amounts of data. Be careful because if
+you have `SPARK_CONF_DIR` also set the tests will try to use whatever cluster you have configured.
+If you do want to run the tests in parallel on an existing cluster it is recommended that you set
+`-Dpytest.TEST_PARALLEL` to the one less than the number of worker applications that will be
+running on the cluster.  This is because `pytest-xdist` will launch one control application that
+is not included in that number. All it does is farm out work to the other applications, but because
+it needs to know about the spark cluster to determine which tests to run and how it still shows up
+as a spark application.
+
+To run the tests separate from the build go to the `integration_tests` directory. You can submit
+`runtests.py` through `spark-submit`, but if you want to run the tests in parallel with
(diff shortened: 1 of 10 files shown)


revans2:
build

revans2:
build

integration_tests/run_pyspark_from_build.sh
```diff
@@ -38,16 +38,55 @@ else
     then
         TEST_TAGS="-m $TEST_TAGS"
     fi
+    if [[ "${TEST_PARALLEL}" == "" ]];
+    then
+        # For now just assume that we are going to use the GPU on the system with the most memory
+        # and then divide it up into 3GiB chunks
+        # we subtract one for the main controlling process that will still
+        # launch an application, but it will never be used
+        TEST_PARALLEL=`nvidia-smi --query-gpu=memory.total --format=csv,noheader | awk '{if (MAX < $1){ MAX = $1}} END {print int(MAX / (2.3 * 1024)) - 1}'`
```
pxLi, line 47: This works for our pre-merge pipeline. Just curious do we plan to support parallelism on multiple GPUs?

revans2: I don't have any plans right now. I am sure we could come up with a way.  The big issue is trying to tell different GPUs apart. I have 2 Nvidia GPUs in my desktop. One is for graphics and another is for compute. Cudf will not run on the graphics GPU at all so if I tried to launch anything on it, I would end up with problems. I also know that some people only have one graphics card that does both, and may not have all of the RAM available for compute. So I am thinking that we should switch this over from total memory to free memory to handle that case. If we can come up with a way to tell what the compute capability is for each GPU then we should be able to hack up how we launch pytest to do that. I'm just not sure how often we will run into that situation in practice.

revans2:
build

docs/testing.md
```diff
@@ -52,17 +52,34 @@ default version runs again Spark 3.0.0, to run against other version use one of
...
-variable `SPARK_HOME` set.  If you have` SPARK_CONF_DIR` also set the tests will try to use
-whatever cluster you have configured.
-
-To run the tests separate from the build go to the `integration_tests` directory and submit
-`run_tests.py` through `spark-submit`.  Be sure to include the necessary jars for the RAPIDS
-plugin either with `spark-submit` or with the cluster when it is 
+variable `SPARK_HOME` set.
+
+By default the tests try to use the python packages `pytest-xdist` and `findspark` to oversubscribe
+your GPU and run the tests in spark local mode. This can speed up theses tests significantly as all
```
tgravescs, line 58:
```suggestion
your GPU and run the tests in spark local mode. This can speed up these tests significantly as all
```

docs/testing.md
```diff
@@ -52,17 +52,34 @@ default version runs again Spark 3.0.0, to run against other version use one of
...
+of the tests that run by default process relatively small amounts of data. Be careful because if
+you have `SPARK_CONF_DIR` also set the tests will try to use whatever cluster you have configured.
+If you do want to run the tests in parallel on an existing cluster it is recommended that you set
+`-Dpytest.TEST_PARALLEL` to the one less than the number of worker applications that will be
```
tgravescs, line 62:
```suggestion
`-Dpytest.TEST_PARALLEL` to one less than the number of worker applications that will be
```

tgravescs (commented):
so if I'm running the tests through maven and want to override the parallelism, I set TEST_PARALLEL=1 in the env?

revans2:
> so if I'm running the tests through maven and want to override the parallelism, I set TEST_PARALLEL=1 in the env?

You would want to add `-Dpytest.TEST_PARALLEL=1` to your command line. If you are running through maven I think it is better to do it with maven.

revans2:
build

tgravescs (approved):

PR #1090, opened 2020-11-09
Fix the integration build

revans2:
Not sure why the integration build is failing, and honestly I don't totally know why it worked before, because I could not make it work.  This changes the PYTHONPATH to point to the full path to the file because we are running in local mode that works.  If we were running on YARN or something that what existed before should work.


jlowe (approved):

tgravescs (approved):

jlowe:
build

GaryShen2008 (approved):
Let's fix it by this change first.

PR #1062, opened 2020-11-04
Get cudf/spark dependency from the correct .m2 dir

NvTimLiu:
Bug: https://github.com/NVIDIA/spark-rapids/issues/1031

'WORKSPACE' & 'M2DIR' vars are needed for shims to gen the correct cudf/spark dependency info in shims.

Below error in 'spark*-info.properties' is due to unset of 'WORKSPACE' & 'M2DIR':
    build/dependency-info.sh: line 30: /jenkins/printJarVersion.sh: No such file or directory
    build/dependency-info.sh: line 33: /jenkins/printJarVersion.sh: No such file or directory

To fix the error, we set the default values for them in 'build/dependency-info.sh':
    'M2DIR=$HOME/.m2/repository'
    'WORKSPACE=../..'

We also need to explicitly set the correct 'M2DIR' path, in case we change it by '-Dmaven.repo.local=$M2DIR'.
Already updated Jenkins scripts to set the correct 'M2DIR'.

Signed-off-by: Tim Liu <timl@nvidia.com>

PR #1095, opened 2020-11-11
Dependency info

NvTimLiu:
List the latest SNAPSHOT jar file in local maven repo

    Get cudf/spark dependency from the correct .m2 dir (#1062)

    * Get cudf/spark dependency from the correct .m2 dir

    'WORKSPACE' & 'M2DIR' vars are needed for shims to gen the correct cudf/spark dependency info in shims.

    Below error in 'spark*-info.properties' is due to unset of 'WORKSPACE' & 'M2DIR':
        build/dependency-info.sh: line 30: /jenkins/printJarVersion.sh: No such file or directory
        build/dependency-info.sh: line 33: /jenkins/printJarVersion.sh: No such file or directory

    To fix the error, we set the default values for them in 'build/dependency-info.sh':
        'M2DIR=$HOME/.m2/repository'
        'WORKSPACE=../..'

    We also need to explicitly set the correct 'M2DIR' path, in case we change it by '-Dmaven.repo.local=$M2DIR'.
    Already updated Jenkins scripts to set the correct 'M2DIR'.

### DIFF
--- a/build/dependency-info.sh
+++ b/build/dependency-info.sh
@@ -23,14 +23,18 @@
 #   SPARK_VER - The version of spark
 
 # Parse cudf and spark dependency versions
+set -e
 
 CUDF_VER=$1
 CUDA_CLASSIFIER=$2
 SERVER_ID=snapshots
-${WORKSPACE}/jenkins/printJarVersion.sh "cudf_version" "${HOME}/.m2/repository/ai/rapids/cudf/${CUDF_VER}" "cudf-${CUDF_VER}" "-${CUDA_CLASSIFIER}.jar" $SERVER_ID
+# set defualt values for 'M2DIR' & 'WORKSPACE' so that shims can get the correct cudf/spark dependnecy
+M2DIR=${M2DIR:-"$HOME/.m2/repository"}
+WORKSPACE=${WORKSPACE:-"../.."}
+${WORKSPACE}/jenkins/printJarVersion.sh "cudf_version" "${M2DIR}/ai/rapids/cudf/${CUDF_VER}" "cudf-${CUDF_VER}" "-${CUDA_CLASSIFIER}.jar" $SERVER_ID
 
 SPARK_VER=$3
-SPARK_SQL_VER=`${WORKSPACE}/jenkins/printJarVersion.sh "spark_version" "${HOME}/.m2/repository/org/apache/spark/spark-sql_2.12/${SPARK_VER}" "spark-sql_2.12-${SPARK_VER}" ".jar" $SERVER_ID`
+SPARK_SQL_VER=`${WORKSPACE}/jenkins/printJarVersion.sh "spark_version" "${M2DIR}/org/apache/spark/spark-sql_2.12/${SPARK_VER}" "spark-sql_2.12-${SPARK_VER}" ".jar" $SERVER_ID`
 
 # Split spark version from spark-sql_2.12 jar filename
 echo ${SPARK_SQL_VER/"-sql_2.12"/}
--- a/build/dependency-info.sh
+++ b/build/dependency-info.sh
@@ -30,7 +30,8 @@ CUDA_CLASSIFIER=$2
 SERVER_ID=snapshots
 # set defualt values for 'M2DIR' & 'WORKSPACE' so that shims can get the correct cudf/spark dependnecy
 M2DIR=${M2DIR:-"$HOME/.m2/repository"}
-WORKSPACE=${WORKSPACE:-"../.."}
+MY_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+WORKSPACE=${WORKSPACE:-"$MY_PATH/.."}
 ${WORKSPACE}/jenkins/printJarVersion.sh "cudf_version" "${M2DIR}/ai/rapids/cudf/${CUDF_VER}" "cudf-${CUDF_VER}" "-${CUDA_CLASSIFIER}.jar" $SERVER_ID
 
 SPARK_VER=$3
--- a/jenkins/printJarVersion.sh
+++ b/jenkins/printJarVersion.sh
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+set -e
 
 function print_ver(){
     TAG=$1
--- a/jenkins/printJarVersion.sh
+++ b/jenkins/printJarVersion.sh
@@ -24,10 +24,14 @@ function print_ver(){
     SERVER_ID=$5
     
     if [[ "$VERSION" == *"-SNAPSHOT" ]]; then
-        PREFIX=${VERSION%-SNAPSHOT}
-        # List the latest SNAPSHOT jar file in the maven repo
-        FILE_NAME=`ls -t $REPO/$PREFIX-[0-9]*$SUFFIX | head -1 | xargs basename`
-        echo $TAG=$FILE_NAME
+        # only run in Jenkins build
+        if [ -n "$JENKINS_URL" ]; then
+            echo $TAG=$VERSION
+        else
+            PREFIX=${VERSION%-SNAPSHOT}
+            # List the latest SNAPSHOT jar file in the maven repo
+            echo $TAG=`ls -t $REPO/$PREFIX-[0-9]*$SUFFIX | head -1 | xargs basename`
+        fi
     else
         echo $TAG=$VERSION$SUFFIX
     fi
--- a/jenkins/printJarVersion.sh
+++ b/jenkins/printJarVersion.sh
@@ -23,16 +23,12 @@ function print_ver(){
     SUFFIX=$4
     SERVER_ID=$5
     
-    if [[ "$VERSION" == *"-SNAPSHOT" ]]; then
-        # Collect snapshot dependency info only in Jenkins build, 
-        # In dev build, print 'SNAPSHOT' tag without time stamp, e.g.: cudf-0.17-SNAPSHOT.jar
-        if [ -n "$JENKINS_URL" ]; then
-            echo $TAG=$VERSION
-        else
-            PREFIX=${VERSION%-SNAPSHOT}
-            # List the latest SNAPSHOT jar file in the maven repo
-            echo $TAG=`ls -t $REPO/$PREFIX-[0-9]*$SUFFIX | head -1 | xargs basename`
-        fi
+    # Collect snapshot dependency info only in Jenkins build
+    # In dev build, print 'SNAPSHOT' tag without time stamp, e.g.: cudf-0.17-SNAPSHOT.jar
+    if [[ "$VERSION" == *"-SNAPSHOT" && -n "$JENKINS_URL" ]]; then
+        PREFIX=${VERSION%-SNAPSHOT}
+        # List the latest SNAPSHOT jar file in the maven repo
+        echo $TAG=`ls -t $REPO/$PREFIX-[0-9]*$SUFFIX | head -1 | xargs basename`
     else
         echo $TAG=$VERSION$SUFFIX
     fi
--- a/jenkins/printJarVersion.sh
+++ b/jenkins/printJarVersion.sh
@@ -24,7 +24,8 @@ function print_ver(){
     SERVER_ID=$5
     
     if [[ "$VERSION" == *"-SNAPSHOT" ]]; then
-        # only run in Jenkins build
+        # Collect snapshot dependency info only in Jenkins build, 
+        # In dev build, print 'SNAPSHOT' tag without time stamp, e.g.: cudf-0.17-SNAPSHOT.jar
         if [ -n "$JENKINS_URL" ]; then
             echo $TAG=$VERSION
         else
--- a/jenkins/printJarVersion.sh
+++ b/jenkins/printJarVersion.sh
@@ -25,9 +25,9 @@ function print_ver(){
     
     if [[ "$VERSION" == *"-SNAPSHOT" ]]; then
         PREFIX=${VERSION%-SNAPSHOT}
-        TIMESTAMP=`grep -oP '(?<=timestamp>)[^<]+' < $REPO/maven-metadata-$SERVER_ID.xml`
-        BUILD_NUM=`grep -oP '(?<=buildNumber>)[^<]+' < $REPO/maven-metadata-$SERVER_ID.xml`
-        echo $TAG=$PREFIX-$TIMESTAMP-$BUILD_NUM$SUFFIX
+        # List the latest SNAPSHOT jar file in the maven repo
+        FILE_NAME=`ls -t $REPO/$PREFIX-[0-9]*$SUFFIX | head -1 | xargs basename`
+        echo $TAG=$FILE_NAME
     else
         echo $TAG=$VERSION$SUFFIX
     fi
--- a/pom.xml
+++ b/pom.xml
@@ -536,8 +536,8 @@
                             <target>
                                 <mkdir dir="${project.build.directory}/extra-resources"/>
                                 <mkdir dir="${project.build.directory}/tmp"/>
-                                <exec executable="bash" output="${project.build.directory}/extra-resources/rapids4spark-version-info.properties">
-                                    <arg value="${project.basedir}/../build/build-info"/>
+                                <exec executable="bash" failonerror="true" output="${project.build.directory}/extra-resources/rapids4spark-version-info.properties">
+                                    <arg value="${user.dir}/build/build-info"/>
                                     <arg value="${project.version}"/>
                                     <arg value="${cudf.version}"/>
                                 </exec>
--- a/pom.xml
+++ b/pom.xml
@@ -161,6 +161,7 @@
         <pytest.TEST_TAGS>not qarun</pytest.TEST_TAGS>
         <pytest.TEST_PARALLEL></pytest.TEST_PARALLEL>
         <rat.consoleOutput>false</rat.consoleOutput>
+        <root.dir>${project.build.directory}</root.dir>
         <slf4j.version>1.7.30</slf4j.version>
         <spark300.version>3.0.0</spark300.version>
         <spark300emr.version>3.0.0-amzn</spark300emr.version>
@@ -537,7 +538,7 @@
                                 <mkdir dir="${project.build.directory}/extra-resources"/>
                                 <mkdir dir="${project.build.directory}/tmp"/>
                                 <exec executable="bash" failonerror="true" output="${project.build.directory}/extra-resources/rapids4spark-version-info.properties">
-                                    <arg value="${user.dir}/build/build-info"/>
+                                    <arg value="${root.dir}/build/build-info"/>
                                     <arg value="${project.version}"/>
                                     <arg value="${cudf.version}"/>
                                 </exec>
--- a/pom.xml
+++ b/pom.xml
@@ -161,7 +161,6 @@
         <pytest.TEST_TAGS>not qarun</pytest.TEST_TAGS>
         <pytest.TEST_PARALLEL></pytest.TEST_PARALLEL>
         <rat.consoleOutput>false</rat.consoleOutput>
-        <root.dir>${project.build.directory}</root.dir>
         <slf4j.version>1.7.30</slf4j.version>
         <spark300.version>3.0.0</spark300.version>
         <spark300emr.version>3.0.0-amzn</spark300emr.version>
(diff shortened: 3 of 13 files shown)


NvTimLiu:
build

NvTimLiu:
@jlowe @revans2 
It is not a good way to parse the correct dependency info from maven-metadata-[REPO_ID].xml file, because there are a list of  maven-metadata-[REPO_ID].xml
Mvn will cmp the timestamp among these `.xml` file and determine whether need to update SNAPSHTO jars.

I buildt spark-rapids from Jenkins maven `package -s jenkins/settings`,  the REPO_ID：https://github.com/NVIDIA/spark-rapids/blob/branch-0.3/jenkins/settings.xml#L49

@jlowe Your case REPO_ID is `snapshots-repo` by `mvn package`, defined here https://github.com/NVIDIA/spark-rapids/blob/branch-0.3/pom.xml#L579

So I suppose we can just find out which dependency jar file is the latest one in SNAPSHOT folder, as below change shows.
https://github.com/NVIDIA/spark-rapids/pull/1095/commits/05cd80c9c7b640a2c0eb59f3baf4a45de6e5ddc8

Could you please help to review? Thanks!

build/dependency-info.sh
```diff
@@ -23,14 +23,18 @@
...
 # Parse cudf and spark dependency versions
+set -e
 
 CUDF_VER=$1
 CUDA_CLASSIFIER=$2
 SERVER_ID=snapshots
-${WORKSPACE}/jenkins/printJarVersion.sh "cudf_version" "${HOME}/.m2/repository/ai/rapids/cudf/${CUDF_VER}" "cudf-${CUDF_VER}" "-${CUDA_CLASSIFIER}.jar" $SERVER_ID
+# set defualt values for 'M2DIR' & 'WORKSPACE' so that shims can get the correct cudf/spark dependnecy
+M2DIR=${M2DIR:-"$HOME/.m2/repository"}
+WORKSPACE=${WORKSPACE:-"../.."}
```
revans2, line 33: How do you know what the current working directory is?  This could be called from any directory. You should get the path of this script and then set a path relative to it.  Even better if you can make it an absolute path.

jenkins/printJarVersion.sh
```diff
@@ -24,9 +25,9 @@ function print_ver(){
     
     if [[ "$VERSION" == *"-SNAPSHOT" ]]; then
         PREFIX=${VERSION%-SNAPSHOT}
-        TIMESTAMP=`grep -oP '(?<=timestamp>)[^<]+' < $REPO/maven-metadata-$SERVER_ID.xml`
-        BUILD_NUM=`grep -oP '(?<=buildNumber>)[^<]+' < $REPO/maven-metadata-$SERVER_ID.xml`
-        echo $TAG=$PREFIX-$TIMESTAMP-$BUILD_NUM$SUFFIX
+        # List the latest SNAPSHOT jar file in the maven repo
+        FILE_NAME=`ls -t $REPO/$PREFIX-[0-9]*$SUFFIX | head -1 | xargs basename`
```
jlowe, line 29: Have you tested this when there are no timestamped versions of the cudf snapshot jar?  (e.g.: developer always builds cudf from source themselves.)

It's annoying that we're jumping through all these hoops to get what Maven already knows.  It would be much better if we could get Maven itself to emit this information.  Unfortunately it doesn't look like Maven has any easy way to do this.  I did find a Maven extension that can record the dependency info at https://github.com/khmarbaise/deployment-recorder-extension but it looks like it only records events when an artifact is deployed, not resolved, so I think it would only record the information we need if the cudf artifact was downloaded during that particular build.

build/dependency-info.sh
NvTimLiu: @revans2  Good suggest, thanks a lot!

jenkins/printJarVersion.sh
NvTimLiu: @jlowe Unfortunately it doesn't look like Maven has any easy way to do this.  --- exactly.
I know there is a file `_remote.repositories` which has all the record downloaded from remote as below. Still we need to parse this file to get the latest time stamp:

cudf-0.17-20201110.090227-19-cuda10-2.jar>apache.snapshots=
cudf-0.17-20201110.070212-17-cuda10-2.jar>apache.snapshots=
cudf-0.17-20201103.102820-14.pom>apache.snapshots=
cudf-0.17-20201103.102820-14-cuda10-1.jar>temp=
cudf-0.17-20201110.090227-19-cuda10-1.jar>apache.snapshots=
cudf-0.17-20201110.090005-20-cuda10-1.jar>snapshots-repo=
cudf-0.17-20201103.102820-14-cuda10-1.jar>apache.snapshots=

NvTimLiu: @jlowe Have you tested this when there are no timestamped versions of the cudf snapshot jar? (e.g.: developer always builds cudf from source themselves.) --- Do you mean that we install cudf dependency into the .m2 dir? 
I did not test this case. Maybe there is a problem. In case we installed cudf into .m2, instead of getting from remote, then the latest jars not from remote. 

So the real dependency should be the local install one? Or we still think that the remote one the the correct dependency?

GaryShen2008: Do we really need to gather the dependency version in local building?
How about we do this printJarVersion.sh only when our nightly or release building?
It can avoid the developer building and make the life easy.

NvTimLiu: Just test that when we install cudf locally by mvn install cudf-0.17*.jar, it only overwrite the cudf-0.17-SNAPSHOT-cuda*.jar file, along with updating the `maven-metadata-local.xml` file

revans2: I'm not sure gathering that info is important, but at a minimum we should test it so we know it does not crash. I did a little bit of that and it looks like it will work properly, but I did not actually install the patch and run a build.  I think that is what @jlowe wants because that is what was missed last time and why we had to revert the patch.

NvTimLiu: So if there is only cudf-0.17-SNAPSHOT-cuda*.jar file, this scripts will fail, too.

revans2: If that is true then we cannot check this in.  We need to detect that.

jlowe: > Do we really need to gather the dependency version in local building?

I believe @tgravescs would like this to work in the off chance that a developer hands a custom build to a user and that user then files a report.  It would be nice to at least indicate in some way the cudf snapshot jar was not an official snapshot from the snapshots repo (e.g.: artifact is just cudf-0.17-SNAPSHOT-cuda10-1.jar without any timestamp, indicating it was a local build).

> ... at a minimum we should test it so we know it does not crash. ... I think that is what @jlowe wants

Exactly.  Putting something useful in the properties for a local developer build would be great, but at a minimum this should not crash the build, local or otherwise, when everything is working normally.

NvTimLiu: Ok, I'll check to only fail the Jenkins build, and NOT fail the local build.
@tgravescs Need we set `cudfVer=cudf-0.17-SNAPSHOT-cuda10-1.jar` dependency info for the local build, if we install the dev built cudf jars?

GaryShen2008: I think we can know the local building just from the dependency info is missing. And the cudf version should be included in pom file.

jlowe: Note that there can be some snapshot files intermixed with local build files in the maven repository.  For example from my local repo:
```
$ ls -lrt ~/.m2/repository/ai/rapids/cudf/0.17-SNAPSHOT
total 613400
-rw-rw-r-- 1 jlowe jlowe     23109 Oct 28 11:53 cudf-0.17-20201028.162547-8.pom
-rw-rw-r-- 1 jlowe jlowe        40 Oct 28 11:53 cudf-0.17-20201028.162547-8.pom.sha1
-rw-rw-r-- 1 jlowe jlowe         0 Oct 28 11:53 cudf-0.17-20201028.162547-8-cuda10-1.jar.part.lock
-rw-rw-r-- 1 jlowe jlowe  39518965 Oct 28 11:53 cudf-0.17-20201028.162547-8-cuda10-1.jar.part
-rw-rw-r-- 1 jlowe jlowe     23109 Nov  5 08:23 cudf-0.17-20201105.090410-17.pom
-rw-rw-r-- 1 jlowe jlowe        40 Nov  5 08:23 cudf-0.17-20201105.090410-17.pom.sha1
-rw-rw-r-- 1 jlowe jlowe 214535283 Nov  5 08:24 cudf-0.17-20201105.090410-17-cuda10-1.jar
-rw-rw-r-- 1 jlowe jlowe        40 Nov  5 08:24 cudf-0.17-20201105.090410-17-cuda10-1.jar.sha1
-rw-rw-r-- 1 jlowe jlowe      1406 Nov 11 08:32 maven-metadata-snapshots-repo.xml
-rw-rw-r-- 1 jlowe jlowe        40 Nov 11 08:32 maven-metadata-snapshots-repo.xml.sha1
-rw-rw-r-- 1 jlowe jlowe       315 Nov 11 08:32 resolver-status.properties
-rw-rw-r-- 1 jlowe jlowe     23203 Nov 11 08:32 cudf-0.17-20201111.093609-21.pom
-rw-rw-r-- 1 jlowe jlowe        40 Nov 11 08:32 cudf-0.17-20201111.093609-21.pom.sha1
-rw-rw-r-- 1 jlowe jlowe 242708552 Nov 11 08:35 cudf-0.17-20201111.093609-21-cuda10-1.jar
-rw-rw-r-- 1 jlowe jlowe        40 Nov 11 08:35 cudf-0.17-20201111.093609-21-cuda10-1.jar.sha1
-rw-rw-r-- 1 jlowe jlowe     23203 Nov 11 10:30 cudf-0.17-SNAPSHOT.pom
-rw-rw-r-- 1 jlowe jlowe    356008 Nov 11 15:33 cudf-0.17-SNAPSHOT-tests.jar
-rw-rw-r-- 1 jlowe jlowe 130840131 Nov 11 15:33 cudf-0.17-SNAPSHOT-cuda10-1.jar
-rw-rw-r-- 1 jlowe jlowe       492 Nov 11 15:33 _remote.repositories
-rw-rw-r-- 1 jlowe jlowe       940 Nov 11 15:33 maven-metadata-local.xml
```

Note that there are timestamped snapshots here, but the jar I'm using is not one of them.  If a timestamped snapshot jar is found it should be compared against the "main" jar path (cudf-0.17-SNAPSHOT-cuad10-1.jar in the above example) to know that it indeed is the one being used.

NvTimLiu: @jlowe The "main" jar timestamp is stored in the `maven-metadata-local.xml` file, if we lacally run `maven install`. So when running local build, we only need check timestamps  between in  `maven-metadata-local.xml`(if exists) and in timestamped jars(cudf-0.17-20201111.093609-21-cuda10-1.jar)
-rw-rw-r-- 1 jlowe jlowe       940 Nov 11 15:33 maven-metadata-local.xml

NvTimLiu: > > Do we really need to gather the dependency version in local building?
> 
> I believe @tgravescs would like this to work in the off chance that a developer hands a custom build to a user and that user then files a report. It would be nice to at least indicate in some way the cudf snapshot jar was not an official snapshot from the snapshots repo (e.g.: artifact is just cudf-0.17-SNAPSHOT-cuda10-1.jar without any timestamp, indicating it was a local build).
> 
> > ... at a minimum we should test it so we know it does not crash. ... I think that is what @jlowe wants
> 
> Exactly. Putting something useful in the properties for a local developer build would be great, but at a minimum this should not crash the build, local or otherwise, when everything is working normally.

@jlowe @revans2 @tgravescs @GaryShen2008 
Let's divide it into 2 parts
    We first make sure that: in case of dependency info fetch error, only fails the Jenkins build, and NOT fail the local build. --- this is more import, I think.
    Then let's check if we need to parse the exact local build dependency info, and make another PR to do it. --- This is of low priority, we can optimize it later.

What's your thought?

GaryShen2008: Agree.  We collect dependency-info for nightly and release build, don't collect in local build first, so that local build can pass whatever the environment is.
To collect dependency-info for the local build seems more complicated, and it's low priority.

NvTimLiu:
build

pom.xml
```diff
@@ -536,8 +536,8 @@
                             <target>
                                 <mkdir dir="${project.build.directory}/extra-resources"/>
                                 <mkdir dir="${project.build.directory}/tmp"/>
-                                <exec executable="bash" output="${project.build.directory}/extra-resources/rapids4spark-version-info.properties">
-                                    <arg value="${project.basedir}/../build/build-info"/>
+                                <exec executable="bash" failonerror="true" output="${project.build.directory}/extra-resources/rapids4spark-version-info.properties">
+                                    <arg value="${user.dir}/build/build-info"/>
```
jlowe, line 540: `user.dir` seems inapproprite here.  Shouldn't it be `${project.basedir}/build/build-info` instead?  Maven guarantees `project.basedir` evaluates to the directory where pom.xml resides which is what we want.  `user.dir` is less controllable.

jenkins/printJarVersion.sh
```diff
@@ -24,12 +25,14 @@ function print_ver(){
...
+        # List the latest SNAPSHOT jar file in the maven repo
+        echo $TAG=`ls -t $REPO/$PREFIX-[0-9]*$SUFFIX | head -1 | xargs basename`
     else
         echo $TAG=$VERSION$SUFFIX
     fi
 }
 
-print_ver $1 $2 $3 $4 $5
+# only run in Jenkins build
+if [ -n "$JENKINS_URL" ]; then
```
jlowe, line 36: Rather than skipping **all** properties for a dev build, how about we add this condition to the existing SNAPSHOT check?  Then we'll get properties either way, but we'll know it's a dev build because the snapshot won't be timestamped.

jenkins/printJarVersion.sh
jlowe: I'm fine with punting on the tricky SNAPSHOT timestamp handling for dev builds for now, but I think it's trivial for us to provide _some_ kind of info for dev builds.  We just need to move the `JENKINS_URL` check to the `SNAPSHOT` check in the script, and then we'll only try to do snapshot timestamp analysis on jenkins builds.  Dev builds will print whatever cudf version was used without timestamp info for snapshots (i.e.: it would just be something like cudf-0.17-SNAPSHOT-cuda10-1 in a dev build).

NvTimLiu:
build

pom.xml
```diff
@@ -161,6 +161,7 @@
         <pytest.TEST_TAGS>not qarun</pytest.TEST_TAGS>
         <pytest.TEST_PARALLEL></pytest.TEST_PARALLEL>
         <rat.consoleOutput>false</rat.consoleOutput>
+        <root.dir>${project.build.directory}</root.dir>
```
jlowe:
```suggestion
        <root.dir>${project.basedir}</root.dir>
```

jenkins/printJarVersion.sh
```diff
@@ -23,10 +24,15 @@ function print_ver(){
     SERVER_ID=$5
     
     if [[ "$VERSION" == *"-SNAPSHOT" ]]; then
-        PREFIX=${VERSION%-SNAPSHOT}
-        TIMESTAMP=`grep -oP '(?<=timestamp>)[^<]+' < $REPO/maven-metadata-$SERVER_ID.xml`
-        BUILD_NUM=`grep -oP '(?<=buildNumber>)[^<]+' < $REPO/maven-metadata-$SERVER_ID.xml`
-        echo $TAG=$PREFIX-$TIMESTAMP-$BUILD_NUM$SUFFIX
+        # Collect snapshot dependency info only in Jenkins build, 
+        # In dev build, print 'SNAPSHOT' tag without time stamp, e.g.: cudf-0.17-SNAPSHOT.jar
+        if [ -n "$JENKINS_URL" ]; then
```
jlowe, line 29: Isn't this backwards?  Don't we want a timestamp when `JENKINS_URL` is set?  I suggest simplifying the logic to something like:
```
    if [[ "$VERSION" == *"-SNAPSHOT" && -n "$JENKINS_URL" ]]; then
        # Collect snapshot dependency info only in Jenkins build, 
        PREFIX=${VERSION%-SNAPSHOT}
        # List the latest SNAPSHOT jar file in the maven repo
        echo $TAG=`ls -t $REPO/$PREFIX-[0-9]*$SUFFIX | head -1 | xargs basename`
    else
        echo $TAG=$VERSION$SUFFIX
    fi
```
Then we'll just print the jar filename i a dev build but do the timestamp shenanigans when we have Jenkins.

NvTimLiu: Agree!

NvTimLiu: build

jlowe (dismissed):

revans2 (approved):

revans2:
The build failed not sure why it didn't update the PR.

https://blossom.nvidia.com/sw-gpu-spark-jenkins/job/rapids_premerge-github/391/

```
09:34:32  [INFO] ------------------------------------------------------------------------
09:34:32  [INFO] Reactor Summary:
09:34:32  [INFO] 
09:34:32  [INFO] RAPIDS Accelerator for Apache Spark Root Project ... FAILURE [  3.511 s]
09:34:32  [INFO] RAPIDS Accelerator for Apache Spark SQL Plugin ..... SKIPPED
09:34:32  [INFO] RAPIDS Accelerator for Apache Spark Shuffle Plugin . SKIPPED
09:34:32  [INFO] RAPIDS Accelerator for Apache Spark SQL Plugin Shims SKIPPED
09:34:32  [INFO] RAPIDS Accelerator for Apache Spark SQL Plugin Spark 3.0.0 Shim SKIPPED
09:34:32  [INFO] RAPIDS Accelerator for Apache Spark SQL Plugin Spark 3.0.0 EMR Shim SKIPPED
09:34:32  [INFO] RAPIDS Accelerator for Apache Spark SQL Plugin Spark 3.0.1 Shim SKIPPED
09:34:32  [INFO] RAPIDS Accelerator for Apache Spark SQL Plugin Spark 3.0.1 EMR Shim SKIPPED
09:34:32  [INFO] RAPIDS Accelerator for Apache Spark SQL Plugin Shim Aggregator SKIPPED
09:34:32  [INFO] RAPIDS Accelerator for Apache Spark Scala UDF Plugin SKIPPED
09:34:32  [INFO] RAPIDS Accelerator for Apache Spark Distribution ... SKIPPED
09:34:32  [INFO] RAPIDS Accelerator for Apache Spark Tests .......... SKIPPED
09:34:32  [INFO] rapids-4-spark-integration-tests_2.12 .............. SKIPPED
09:34:32  [INFO] rapids-4-spark-api-validation ...................... SKIPPED
09:34:32  [INFO] ------------------------------------------------------------------------
09:34:32  [INFO] BUILD FAILURE
09:34:32  [INFO] ------------------------------------------------------------------------
09:34:32  [INFO] Total time: 3.803 s
09:34:32  [INFO] Finished at: 2020-11-17T15:34:28+00:00
09:34:32  [INFO] Final Memory: 23M/494M
09:34:32  [INFO] ------------------------------------------------------------------------
09:34:32  [ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:1.8:run (default) on project rapids-4-spark-parent: An Ant BuildException has occured: exec returned: 127
09:34:32  [ERROR] around Ant part ...<exec output="/home/jenkins/agent/workspace/jenkins-rapids_premerge-github-391/target/extra-resources/rapids4spark-version-info.properties" failonerror="true" executable="bash">... @ 6:180 in /home/jenkins/agent/workspace/jenkins-rapids_premerge-github-391/target/antrun/build-main.xml
09:34:32  [ERROR] -> [Help 1]
09:34:32  [ERROR] 
09:34:32  [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
09:34:32  [ERROR] Re-run Maven using the -X switch to enable full debug logging.
09:34:32  [ERROR] 
09:34:32  [ERROR] For more information about the errors and possible solutions, please read the following articles:
09:34:32  [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```

NvTimLiu:
build

NvTimLiu:
rebase PR source to fix merge conflicts

NvTimLiu:
build

jlowe (approved):

PR #1097, opened 2020-11-11
Cleanup some instances of excess closure serialization

jlowe:
I ran across an instance in `GpuCoalesceBatches` where the code was referencing `child.output` within the `mapPartitions` call which requires the entire `child` instance to be serialized.  That essentially serializes the entire Catalyst plan up to that point.

This removes many instances of `child.output` being referenced in `mapPartitions` along with some cleanup of other member variables being accessed within the `mapPartitions` closure, requiring the entire object to be serialized.


jlowe:
Note that I believe there are still many instances where class constructor arguments are being referenced within `mapPartitions` closures and probably requiring the entire class to be serialized.  I suggest tackling those cases in a followup issue if that's agreeable.

jlowe:
build

tgravescs (approved):

PR #1099, opened 2020-11-11
Update how we deal with type checking

revans2:
After looking at how complex nested type checking is going to be combined with decimal type checking in #1063 @jlowe and I decided that we needed a more complete solution.  This is hopefully that solution.  It is a little bit ugly but it provides a fairly clean and consistent way to cover all of our existing checks and also updates it for decimal, Array, Struct, and Map support.

It does not prevent users from overriding the type check and doing it all themselves, but that should be very rare instead of common like it is now.


revans2:
build

jlowe (approved):

revans2:
@sperlingxx does this change look good to you?

kuhushukla (approved):

sperlingxx:
> @sperlingxx does this change look good to you?

LGTM!

Issue #688, opened 2020-09-08
[FEA] udf compiler should be auto-appended to `spark.sql.extensions`


abellina:
Currently, in addition to setting `spark.rapids.sql.udfCompiler.enabled=true`, you also need to add it to `spark.sql.extensions`. As suggested here (https://github.com/NVIDIA/spark-rapids/pull/687#discussion_r485230754), we should try and make that automatic since then users can toggle using the setting only.

PR #1102, opened 2020-11-12
Auto-register UDF extention when main plugin is set

wjxiz1992:

To deal with https://github.com/NVIDIA/spark-rapids/issues/688.
With this change, the udf-compiler extension will be registered automatically which has the same behavior as `SQLExecPlugin`.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -290,7 +290,7 @@ Casting from string to timestamp currently has the following limitations.
  Only timezone 'Z' (UTC) is supported. Casting unsupported formats will result in null values. 
  
 ## UDF to Catalyst Expressions
-To speedup the process of UDF, spark-rapids introduces a udf-compiler extension to translate UDFs to Catalyst expressions. This compiler will be injected automatically to spark extensions by setting `spark.plugins=com.nvidia.spark.SQLPlugin` and is disabled by default.
+To speedup the process of UDF, spark-rapids introduces a udf-compiler extension to translate UDFs to Catalyst expressions. 
 
 To enable this operation on the GPU, set
 [`spark.rapids.sql.udfCompiler.enabled`](configs.md#sql.udfCompiler.enabled) to `true`.
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -290,10 +290,10 @@ Casting from string to timestamp currently has the following limitations.
  Only timezone 'Z' (UTC) is supported. Casting unsupported formats will result in null values. 
  
 ## UDF to Catalyst Expressions
-To speedup the process of UDF, spark-rapids introduces a udf-compiler extension to translate UDFs to Catalyst expressions.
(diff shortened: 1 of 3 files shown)


revans2 (approved):
Looks good to me

revans2:
build

abellina (approved):

abellina:
We should also update documentation:

- https://github.com/NVIDIA/spark-rapids/blob/16e80917795a232cfb05f92d8f9296ef0369f3d4/docs/compatibility.md
- https://github.com/NVIDIA/spark-rapids/blob/34d52d43ca6f465df97b539390c4c13a7a2d6a5a/udf-compiler/README.md

Could happen in a different PR IMHO.

abellina (changes requested):
Could we remove the extensions modification from the UDF tests?

https://github.com/NVIDIA/spark-rapids/blob/16e80917795a232cfb05f92d8f9296ef0369f3d4/udf-compiler/src/test/scala/com/nvidia/spark/OpcodeSuite.scala#L37

wjxiz1992:
> Could we remove the extensions modification from the UDF tests?
> 
> https://github.com/NVIDIA/spark-rapids/blob/16e80917795a232cfb05f92d8f9296ef0369f3d4/udf-compiler/src/test/scala/com/nvidia/spark/OpcodeSuite.scala#L37

Removing current extension requires adding main sql plugin config. But that will fail the OpcodeSuite. 
Can we keep the UDF test suites independent and leave the extension there.

revans2:
I would like to see the documentation updated as a part of this PR, just so it is not forgotten.

docs/compatibility.md
```diff
@@ -290,10 +290,10 @@ Casting from string to timestamp currently has the following limitations.
  Only timezone 'Z' (UTC) is supported. Casting unsupported formats will result in null values. 
  
 ## UDF to Catalyst Expressions
-To speedup the process of UDF, spark-rapids introduces a udf-compiler extension to translate UDFs to Catalyst expressions.
+To speedup the process of UDF, spark-rapids introduces a udf-compiler extension to translate UDFs to Catalyst expressions. This compiler will be injected automatically to spark extensions by setting `spark.plugins=com.nvidia.spark.SQLPlugin` and is disabled by default.
```
revans2, line 293: nit: I don't think we need this line.  I don't think the user cares about the internal details of how we implemented it, just how to enable it, which is covered by the change below.

revans2 (approved):

docs/compatibility.md
wjxiz1992: Right, reasonable to me. Updated.

revans2 (approved):

revans2:
build

abellina:
> Removing current extension requires adding main sql plugin config. But that will fail the OpcodeSuite.
Can we keep the UDF test suites independent and leave the extension there?

Sounds good, sorry I missed that.

abellina (approved):

PR #1104, opened 2020-11-12
Support lists to/from the GPU

revans2:
This is blocked on https://github.com/rapidsai/cudf/pull/6749 going in.  After it is in this will support moving lists to/from the GPU., even nested lists.

There is also some refactoring of the code associated with this to try and make it cleaner/simpler to test/support.

I will likely still add in some tests for existing array operators to be sure that they are working correctly.

### DIFF
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -69,16 +69,16 @@ else
     then
         # With xdist 0 and 1 are the same parallelsm but
         # 0 is more effecient
-        TEST_PARALLEL=""
+        TEST_PARALLEL_OPTS=""
         MEMORY_FRACTION='1'
     else
         MEMORY_FRACTION=`python -c "print(1/($TEST_PARALLEL + 1))"`
-        TEST_PARALLEL="-n $TEST_PARALLEL"
+        TEST_PARALLEL_OPTS="-n $TEST_PARALLEL"
     fi
     RUN_DIR="$SCRIPTPATH"/target/run_dir
     mkdir -p "$RUN_DIR"
     cd "$RUN_DIR"
-    if [[ "${TEST_PARALLEL}" != "" ]];
+    if [[ "${TEST_PARALLEL_OPTS}" != "" ]];
     then
         export PYSP_TEST_spark_driver_extraClassPath="${ALL_JARS// /:}"
         export PYSP_TEST_spark_driver_extraJavaOptions="-ea -Duser.timezone=GMT $COVERAGE_SUBMIT_FLAGS"
@@ -91,7 +91,7 @@ else
 
         python \
           "$SCRIPTPATH"/runtests.py --rootdir "$SCRIPTPATH" "$SCRIPTPATH"/src/main/python \
-          $TEST_PARALLEL \
+          $TEST_PARALLEL_OPTS \
           -v -rfExXs "$TEST_TAGS" \
           --std_input_path="$SCRIPTPATH"/src/test/resources/ \
           "$TEST_ARGS" \
@@ -110,6 +110,5 @@ else
           "$TEST_ARGS" \
           $RUN_TEST_PARAMS \
           "$@"
-
     fi
 fi
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -53,10 +53,8 @@ else
     if python -c 'import findspark';
     then
         echo "FOUND findspark"
-        FIND_SPARK=1
     else
         TEST_PARALLEL=0
-        FIND_SPARK=0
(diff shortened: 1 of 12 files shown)


integration_tests/run_pyspark_from_build.sh
```diff
@@ -80,7 +78,7 @@ else
     RUN_DIR="$SCRIPTPATH"/target/run_dir
     mkdir -p "$RUN_DIR"
     cd "$RUN_DIR"
-    if [[ "${FIND_SPARK}" == "1" ]];
+    if [[ "${TEST_PARALLEL}" != "" ]];
```
revans2, line 81: This is here just to make debugging simpler. I found out that with xdist enabled the spark output would not show up on stdout/stderr.  So if you set the parallelism to 1 we run in a way that will output the data.

jlowe: Shouldn't this be:
```suggestion
    if [[ "${TEST_PARALLEL}" != "0" ]];
```
in case someone set it explicitly (e.g.: TEST_PARALLEL=4)

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
```diff
@@ -103,188 +85,247 @@ private object GpuRowToColumnConverter {
...
+    val structBuilder = builder.getChild(0)
+    val keyBuilder = structBuilder.getChild(0)
+    val valueBuilder = structBuilder.getChild(1)
+    for (i <- 0 until numElements) {
+      ret += keyConverter.append(srcKeys, i, keyBuilder)
+      ret += valueConverter.append(srcValues, i, valueBuilder)
+      structBuilder.endStruct()
+    }
+    builder.endList()
+    ret + 4
```
kuhushukla, line 304: Some comments here will make it easy for newbies to follow along

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
```diff
@@ -103,188 +85,247 @@ private object GpuRowToColumnConverter {
...
-        val values = new ArrayBuffer[AnyRef]()
-        values += new String(srcKeys.getUTF8String(i).getBytes)
-        values += new String(srcValues.getUTF8String(i).getBytes)
-        listOfData += new StructData(values.asJava)
+        column: Int, builder: ai.rapids.cudf.HostColumnVector.ColumnBuilder): Double = {
+      if (row.isNullAt(column)) {
+        builder.appendNull()
+        4.125
+      } else {
+        mapConvert(keyConverter, valueConverter, row, column, builder) + 0.125
```
kuhushukla, line 316: Some comments here will make it easy for newbies to follow along

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
```diff
@@ -103,188 +85,247 @@ private object GpuRowToColumnConverter {
...
     override def append(row: SpecializedGetters,
-      column: Int, builder: ai.rapids.cudf.HostColumnVector.ColumnBuilder): Long =
+      column: Int, builder: ai.rapids.cudf.HostColumnVector.ColumnBuilder): Double =
       if (row.isNullAt(column)) {
         builder.appendNull()
-        0
+        4.125 // Offset and validity
       } else {
-        NotNullStringConverter.append(row, column, builder)
+        NotNullStringConverter.append(row, column, builder) + 4.125
```
jlowe, line 270: Is this supposed to be 0.125 instead of 4.125?  NotNullStringConverter is already accounting for the offset.

jlowe (commented):
Not super comfortable with the summing of doubles to compute byte counts.  Seems like we're asking to be bitten by an unfortunate truncation because the computed size is epsilon under the actual value due to floating point error.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
revans2: I assumed you wanted some help on the return values, but if you want other comments let me know and I'll add them.

kuhushukla: latest change looks good to me.

jlowe (approved):

revans2:
build

revans2:
The dependency has is now in the SNAPSHOT build for cudf so this is no longer WIP

kuhushukla (approved):

PR #1106, opened 2020-11-12
Improve mechanism for expected exceptions in tests

andygrove:

Replace `testExpectedExceptionStartsWith` with a more flexible `testExpectedException` that accepts a function to determine if an exception is the expected one or not, so we are not limited to just looking at `startsWith`.

I needed this in another PR that I am working on.


jlowe (approved):

jlowe:
build

PR #1112, opened 2020-11-13
Remove bad config from pytest setup

revans2:
Thanks @NvTimLiu for finding this.

We should not be forcing the tests to only run in local mode.


revans2:
build

tgravescs (approved):
yep, my bad, I should have caught that in review

Issue #1098, opened 2020-11-11
[BUG] unix_timestamp on GPU returns invalid data for bad input


andygrove:
**Describe the bug**

The `unix_timestamp` function should either return null values or throw an exception for values that cannot be parsed given the specified format (depending on the value of `spark.sql.legacy.timeParserPolicy`).

However, the function currently returns incorrect data.

**Steps/Code to reproduce bug**

```scala
testSparkResultsAreEqual("Test unix_timestamp mix of valid and invalid inputs",
  spark => {
    import spark.implicits._
    Seq(
      "31/12/1999",
      "31-12-1999",
      "12/31/1999",
      "12-31-1999",
      "1999-12-31",
      "1999/12/31",
    ).toDF("c0")
  },
  conf()) {
  df => df.withColumn("c1", unix_timestamp(col("c0"), "yyyy-MM-dd"))
}
```

```
Running on the GPU and on the CPU did not match 
CPU: ArrayBuffer([31/12/1999,null], [31-12-1999,null], [12/31/1999,null], [12-31-1999,null], [1999-12-31,946598400], [1999/12/31,null])

GPU: ArrayBuffer([31/12/1999,-61206796800], [31-12-1999,-61206796800], [12/31/1999,-61806412800], [12-31-1999,-61806412800], [1999-12-31,946598400], [1999/12/31,946598400])
``` 

**Expected behavior**
Behavior should match Spark on CPU.

**Environment details (please complete the following information)**
N/A

**Additional context**
N/A

andygrove:
cuDF now has an `is_timestamp` function that we could use to filter out invalid values.

Issue #1111, opened 2020-11-13
[FEA] support `spark.sql.legacy.timeParserPolicy` when parsing CSV files


andygrove:
**Is your feature request related to a problem? Please describe.**
We should  support `spark.sql.legacy.timeParserPolicy` when parsing CSV files

sameerz:
Needs additional support for CSV parsing in cudf.

Issue #1094, opened 2020-11-10
[BUG] unix_timestamp on GPU returns invalid data for special dates


andygrove:
**Describe the bug**

The `unix_timestamp` function on GPU does not correctly handle the special date strings "now", "epoch", "yesterday", "today", "tomorrow".

**Steps/Code to reproduce bug**

Test:

```scala
  testSparkResultsAreEqual("Test unix_timestamp special dates - default",
    spark => {
      import spark.implicits._
      Seq("now", "epoch", "yesterday", "today", "tomorrow").toDF("c0")
    },
    conf()) {
    df => df.withColumn("c1", unix_timestamp(col("c0"), "yyyy-MM-dd"))
  }
```

Results:

```
CPU: ArrayBuffer([now,1605050295], [epoch,0], [yesterday,1604880000], [today,1604966400], [tomorrow,1605052800])

GPU: ArrayBuffer([now,-62193657600], [epoch,-62193657600], [yesterday,-62193657600], [today,-62193657600], [tomorrow,-62193657600])
```

**Expected behavior**
GPU and CPU results should match.

**Environment details (please complete the following information)**
N/A

**Additional context**
N/A

andygrove:
Resolved by https://github.com/NVIDIA/spark-rapids/pull/1113

Issue #50, opened 2020-05-29
[FEA] support `spark.sql.legacy.timeParserPolicy`


revans2:
**Is your feature request related to a problem? Please describe.**
When parsing dates and times it would be good if we could also follow the `spark.sql.legacy.timeParserPolicy` config.

sameerz:
Trace through where this config is used in Spark and if the plugin cannot match the same functionality, fall back to the CPU.

andygrove:
The default value for `spark.sql.legacy.timeParserPolicy' is `EXCEPTION` in which case Spark throws an exception if any of the following functions are unable to parse data using the specified pattern, and suggests that the conversion may work with `LEGACY`. If the config is set to `CORRECTED` then the conversion will return `null` instead of throwing an exception.

- unix_timestamp
- from_unixtime
- from_utc_timstamp
- to_unix_timestamp
- to_utc_timestamp
- to_date
- to_timestamp
- date_format

I propose that we follow the same behavior but fall back to CPU for `LEGACY` for these functions until we have a reason to add support for specific legacy formats that are no longer supported in Spark 3.0 and later. If we do end up doing that we can then just fall back to CPU for legacy formats that we do not support.

andygrove:
Resolved by https://github.com/NVIDIA/spark-rapids/pull/1113 for functions, and I filed a follow on https://github.com/NVIDIA/spark-rapids/issues/1111 for handling this for CSV reads

nsinghal20:
> The default value for `spark.sql.legacy.timeParserPolicy` is `EXCEPTION` in which case Spark throws an exception if any of the following functions are unable to parse data using the specified pattern, and suggests that the conversion may work with `LEGACY`. If the config is set to `CORRECTED` then the conversion will return `null` instead of throwing an exception.
> 
> * unix_timestamp
> * from_unixtime
> * from_utc_timstamp
> * to_unix_timestamp
> * to_utc_timestamp
> * to_date
> * to_timestamp
> * date_format
> 
> I propose that we follow the same behavior but fall back to CPU for `LEGACY` for these functions until we have a reason to add support for specific legacy formats that are no longer supported in Spark 3.0 and later. If we do end up doing that we can then just fall back to CPU for legacy formats that we do not support.

This answer seems satisfying to my problem but can anyone suggest an example timestamp / data for which it may be throwing error , my date is 2024-12-10  13:22:22  but still it throws exception

revans2:
[...]

PR #1113, opened 2020-11-13
Support unix_timestamp on GPU for subset of formats

andygrove:

This PR makes improvements to `unix_timestamp` so that a small number of date/timestamp formats can now be supported on GPU without requiring that `incompatibleOps=true` because they are now compatible with Spark (with the exception of https://github.com/NVIDIA/spark-rapids/issues/1117).

Specific changes in this PR:

- `unix_timestamp` and `to_unix_timestamp` no longer return bad data for invalid inputs when using one of the compatible formats because it first validates the contents of a column using `is_timestamp` and will either throw an exception or convert invalid inputs to null, depending on the value specified for `legacyTimeParserPolicy`.
- A new configuration option `spark.rapids.sql.incompatibleDateFormats.enabled` has been added.
- `unix_timestamp` now supports the special dates `epoch`, `now`, `today`, `yesterday`, and `tomorrow`.
- `unix_timestamp` will fall back to CPU if `legacyTimeParserPolicy` is `LEGACY`, or if the supplied format is not supported on GPU and `spark.rapids.sql.incompatibleDateFormats.enabled=false`.
- As a side effect of these changes, `to_date` is now supported on GPU because it gets translated to an expression that is a combination of `CAST` and `unix_timestamp`, using one of the compatible formats.

This PR closes https://github.com/NVIDIA/spark-rapids/issues/1098, https://github.com/NVIDIA/spark-rapids/issues/1094, and largely addresses https://github.com/NVIDIA/spark-rapids/issues/50. I filed a new issue https://github.com/NVIDIA/spark-rapids/issues/1111 for the remainder of the work to support `legacyTimeParserPolicy`.


andygrove:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -293,11 +296,24 @@ abstract class UnixTimeExprMeta[A <: BinaryExpression with TimeZoneAwareExpressi
...
-        if (rightLit.isDefined) {
-          strfFormat = DateUtils.toStrf(rightLit.get)
-        } else {
-          willNotWorkOnGpu("format has to be a string literal")
+        extractStringLit(expr.right) match {
+          case Some(rightLit) =>
+            if (GpuOverrides.getTimeParserPolicy == LegacyTimeParserPolicy) {
+              willNotWorkOnGpu("legacyTimeParserPolicy is LEGACY")
+            } else {
+              val gpuSupportedFormats = Seq(
```
revans2, line 304: nit: would it be better if this was a singleton?  Not a big deal right now, just curious about it in the future when/if we support more formats.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -338,7 +362,93 @@ abstract class GpuToTimestamp
   override def doColumnar(lhs: GpuColumnVector, rhs: Scalar): ColumnVector = {
     val tmp = if (lhs.dataType == StringType) {
       // rhs is ignored we already parsed the format
-      lhs.getBase.asTimestampMicroseconds(strfFormat)
+
+      val DAY_MICROS = 24 * 60 * 60 * 1000000L
+      val specialDates = GpuCast.calculateSpecialDates
```
revans2, line 367: nit: Would it be cleaner to have the special date calculations be in a date/time utility instead of GpuCast?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -338,7 +362,93 @@ abstract class GpuToTimestamp
...
+          .map {
+            case (name, days) => (name, days * DAY_MICROS)
+          }
+
+      def daysScalar(name: String): Scalar = {
+        Scalar.timestampFromLong(DType.TIMESTAMP_MICROSECONDS, specialDates(name))
+      }
+
+      def daysEqual(name: String): ColumnVector = {
+        lhs.getBase.equalTo(Scalar.fromString(name))
```
revans2, line 377: Scalars have to be closed when done being used.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -338,7 +362,93 @@ abstract class GpuToTimestamp
...
+                withResource(daysEqual(GpuCast.TOMORROW)) { isTomorrow =>
+                  withResource(lhs.getBase.isNull) { isNull =>
+                    val canBeConverted = isTimestamp.or(isEpoch.or(isNow.or(isToday.or(
+                      isYesterday.or(isTomorrow.or(isNull))))))
+
+                    // throw error if legacyTimeParserPolicy is EXCEPTION
+                    if (timeParserPolicy == ExceptionTimeParserPolicy) {
+                      withResource(Scalar.fromBool(false)) { falseScalar =>
+                        if (canBeConverted.hasNulls || canBeConverted.contains(falseScalar)) {
+                          throw new RuntimeException(
```
revans2, line 412: Is this the same type of exception that Spark is throwing? It is minor but would be nice to do if possible.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -338,7 +362,93 @@ abstract class GpuToTimestamp
...
+          withResource(daysEqual(GpuCast.NOW)) { isNow =>
+            withResource(daysEqual(GpuCast.TODAY)) { isToday =>
+              withResource(daysEqual(GpuCast.YESTERDAY)) { isYesterday =>
+                withResource(daysEqual(GpuCast.TOMORROW)) { isTomorrow =>
+                  withResource(lhs.getBase.isNull) { isNull =>
+                    val canBeConverted = isTimestamp.or(isEpoch.or(isNow.or(isToday.or(
+                      isYesterday.or(isTomorrow.or(isNull))))))
+
+                    // throw error if legacyTimeParserPolicy is EXCEPTION
+                    if (timeParserPolicy == ExceptionTimeParserPolicy) {
```
revans2, line 409: I think we can move this checkup and save a lot of computation in some cases.  `isTimestamp` is the only thing that is common between this check and the actual conversion. It would be nice to only do the check when necessary and then close all of the temp columns before starting on the actual conversion.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -293,11 +296,24 @@ abstract class UnixTimeExprMeta[A <: BinaryExpression with TimeZoneAwareExpressi
...
-        } else {
-          willNotWorkOnGpu("format has to be a string literal")
+        extractStringLit(expr.right) match {
+          case Some(rightLit) =>
+            if (GpuOverrides.getTimeParserPolicy == LegacyTimeParserPolicy) {
+              willNotWorkOnGpu("legacyTimeParserPolicy is LEGACY")
+            } else {
+              val gpuSupportedFormats = Seq(
+                "yyyy-MM-dd",
+                "yyyy-MM-dd HH:mm:ss"
```
revans2, line 306: Is this a regression in some way?  Did we only support these before?  If not then it would be good to find a way to expand this to something much closer to what we did support before.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
andygrove: The boolean vectors `isTimestamp`, `isEpoch`, `isNow`, etc are used later on when performing the actual conversion. However, I am wondering if we should introduce a new config option to override all of this and force the expression to run onto GPU to preserve what we were able to do before with incompatibleOps, at least until we have support for a wider range of formats.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
andygrove: This is a regression in terms of what will run on GPU but is also an improvement because we are now compatible with Spark for these formats instead of returning bad data for inputs that don't match. I can certainly expand the list of supported formats and maybe make this more general so that we support any fixed-width format. The tricky one is supporting timestamp with millis because cuDF is much stricter than Spark, which will accept partial millis. I will spend more time looking at options here.

revans2: I am fine with doing that in a follow-on issue.  We are fixing a bug after all.  I mostly wanted to be sure that I understood it correctly and that we don't drop it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
revans2: My real concern was about extra work and extra memory usage.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
andygrove: I have now changed the behavior so that the expression runs on GPU if a compatible format is being used, or if incompatibleOps is true, so this is no longer a regression.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -293,11 +296,20 @@ abstract class UnixTimeExprMeta[A <: BinaryExpression with TimeZoneAwareExpressi
...
-        val rightLit = extractStringLit(expr.right)
-        if (rightLit.isDefined) {
-          strfFormat = DateUtils.toStrf(rightLit.get)
-        } else {
-          willNotWorkOnGpu("format has to be a string literal")
+        extractStringLit(expr.right) match {
+          case Some(rightLit) =>
+            sparkFormat = rightLit
+            if (GpuOverrides.getTimeParserPolicy == LegacyTimeParserPolicy) {
+              willNotWorkOnGpu("legacyTimeParserPolicy is LEGACY")
```
revans2, line 303: nit: I think we want to explain that LEGACY is not support, but this is really minor because it is implied by the context.

```
willNotWorkOnGpu("a legacyTimeParserPolicy of LEGACY is not supported")
```

andygrove:
Thanks for the review @revans2. I have now addressed the remaining issues.

revans2 (approved):
looks great

revans2:
build

andygrove:
The PySpark tests failed. I will either need to modify the config to enable incompatible formats or add more formats to the list of compatible formats.

revans2:
> The PySpark tests failed. I will either need to modify the config to enable incompatible formats or add more formats to the list of compatible formats.

There are only two formats that are being tested,

'yyyy/MM' and 'yyyy/MM/dd'

One of the tests is marked as xfail because we allow invalid dates to be inserted.

The tests are already marked with `incompat` because `to_unix_timestamp` is not compatible so I am fine with either fix you suggested.  I personally would like to see more formats being marked as compatible.  Perhaps a disallow list in the parser would be better so we can support everything except for the one variable width one we know of.

revans2 (approved):

integration_tests/src/main/python/date_time_test.py
```diff
@@ -190,14 +186,12 @@ def test_to_unix_timestamp_improved(data_gen):
         (StringGen('[0-9]{4}/[01][12]/[0-2][1-8]'),'yyyy/MM/dd'),
         (ConvertGen(DateGen(nullable=False), lambda d: d.strftime('%Y/%m').zfill(7), data_type=StringType()), 'yyyy/MM')]
 
-@incompat
 @pytest.mark.parametrize('data_gen,date_form', str_date_and_format_gen, ids=idfn)
 def test_string_to_unix_timestamp(data_gen, date_form):
-    print("date: " + date_form)
+    conf = {"spark.rapids.sql.improvedTimeOps.enabled": "true"}
```
andygrove, line 191: The test fails if I don't add this config and I don't understand why yet. I am debugging this still.

andygrove:
build

andygrove:
@revans This is ready for review now. I had misunderstood the correct behavior for `legacyTimeParserPolicy=EXCEPTION`. We should not throw exceptions for values that we fail to parse for a given format unless those inputs would have been valid in Spark versions prior to 3.0.0. This is not true of any of the formats that we currently support on GPU.

andygrove:
build

andygrove:
It looks like the behavior has changed in Spark 3.1.0 and `unix_timestamp` gets translated to `cast(gettimestamp(c0#20108, yyyy-MM-dd, Some(UTC)) as date)` which we do not currently support on GPU, which is why the tests are failing against 3.1.0 currently.

revans2 (commented):
The code looks fine, but if we know that there are failures in 3.1.0 then we should at least file a follow-on issue for them and either skip the tests for spark 3.1.0+ or mark them as xfail

revans2 (approved):

andygrove:
build

PR #1116, opened 2020-11-13
Optimize normal shuffle by coalescing smaller batches on host

jlowe:
This optimizes the normal (legacy) Spark shuffle when using the plugin by coalescing smaller shuffle partitions on the host before copying the data to the GPU.  This reduces the number of GPU copies and allows tasks to perform this coalesce in parallel during shuffle without holding the GPU semaphore as it did before.  The GPU semaphore is only acquired when the coalesce happens which for many queries is sufficient to coalesce the entire shuffle input on the CPU.

A new `ShuffleCoalesceExec` node will appear in query plans immediately after GpuColumnarExchangeExec nodes when using normal shuffle.  This replaces the GpuCoalesceBatches node in most cases, as the coalesce is occurring on the host specific to the serialized shuffle tables rather than a general table concatenation on the GPU implemented by GpuCoalesceBatches.


jlowe:
@andygrove I tried to update the code to properly handle what AQE will do to shuffles, but I'm not very familiar with that area.  It would be great if you could double-check the code.

jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarBatchSerializer.scala
```diff
@@ -245,3 +216,38 @@ private class GpuColumnarBatchSerializerInstance(
...
+class SerializedTableColumn(
+    val header: SerializedTableHeader,
+    val hostBuffer: HostMemoryBuffer) extends GpuColumnVectorBase(NullType) {
+  override def close(): Unit = {
+    if (hostBuffer != null) {
+      hostBuffer.close()
+    }
+  }
+
+  override def hasNull: Boolean = throw new IllegalStateException("should not be called")
```
abellina, line 234: minor nit, these could be `UnsupportedOperationException`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -261,6 +260,14 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
       p.withNewChildren(p.children.map(c => insertCoalesce(c, shouldDisable)))
   }
 
+  /** Inserts a shuffle coalesce after every shuffle */
+  private def insertShuffleCoalesce(plan: SparkPlan): SparkPlan = plan match {
```
abellina, line 264: This one says "after every shuffle", but it really is only after every shuffle, if the RAPIDS shuffle isn't enabled, and the coalesce bit happens on the host, do you think something like this might make it clearer?

```suggestion
  private def insertShuffleCoalesceOnHost(plan: SparkPlan): SparkPlan = plan match {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShuffleCoalesceExec.scala
```diff
@@ -0,0 +1,207 @@
...
+
+  override protected def doExecute(): RDD[InternalRow] = {
+    throw new IllegalStateException("ROW BASED PROCESSING IS NOT SUPPORTED")
+  }
+
+  override def doExecuteColumnar(): RDD[ColumnarBatch] = {
+    val metricsMap = metrics
+    // Keep the batch size below a 2GB limit so that no nested type child column can exceed
+    // the cudf 2^31 row count limit
+    val targetBatchByteSize = Math.min(rapidsConf.gpuTargetBatchSizeBytes, Integer.MAX_VALUE)
```
abellina, line 64: seems like this protection should be on the config itself. Does it make sense for the target batch size to be above MAX_VALUE?

abellina (changes requested):
Looking more at `ShuffleCoalesceExec`, so I may have a second pass.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarBatchSerializer.scala
jlowe: I used `IllegalStateException` because calling this means the code got into an illegal state.  It should never be called.  We technically _could_ support it, but we don't ever want to.

abellina: Ok that makes sense.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
jlowe: Personally I don't want to flag everywhere in the code that `ShuffleCoalesceExec` is really `ShuffleCoalesceOnHost`.  If we feel strongly enough that we need to emphasize the host part of it then let's just change the name of the query plan node.

abellina: This is definitely in the nit category. The comment says "after every shuffle" but isn't always the case, so a comment tweak may be more appropriate in this scenario.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShuffleCoalesceExec.scala
jlowe: Sure, I'll update the config's upper limit across the board.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShuffleCoalesceExec.scala
```diff
@@ -0,0 +1,207 @@
...
+
+    outputBatchesMetric += 1
+    outputRowsMetric += batch.numRows
+
+    // update the stats for the next batch in progress
+    numTablesInBatch = serializedTables.size
+    batchByteSize = 0
+    numRowsInBatch = 0
+    if (numTablesInBatch > 0) {
+      require(numTablesInBatch == 1)
```
abellina, line 182: Ok this one confused me a bit, but I think I understand. `serializedTables.size - numTablesInBatch <= 1`, given the logic in `bufferNextBatch`. Should we add a string message to `require` in case this relationship changes? Or maybe a comment.

abellina (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -55,9 +55,9 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
       GpuRowToColumnarExec(optimizeAdaptiveTransitions(r2c.child, Some(r2c)), goal)
 
     case ColumnarToRowExec(GpuBringBackToHost(
-        GpuCoalesceBatches(e: GpuShuffleExchangeExecBase, _))) if parent.isEmpty =>
+        ShuffleCoalesceExec(e: GpuShuffleExchangeExecBase, _))) if parent.isEmpty =>
```
andygrove, line 58: Since `ShuffleCoalesceExec` is only inserted for legacy shuffle, don't we still need to handle the `GpuCoalesceBatches` case as well, for the RapidsShuffle case?

I am going to pull this PR locally and run some benchmarks with AQE and legacy shuffle but it would be good to do the same with UCX as well.

jlowe: Yes, nice catch!  This needs to insert the correct coalesce depending on the shuffle being used.

andygrove: Ok, so one way to handle this would be to duplicate the case statements, although we could simplify this if there is a common base class or trait.

```scala
    case ColumnarToRowExec(GpuBringBackToHost(
        ShuffleCoalesceExec(e: GpuShuffleExchangeExecBase, _))) if parent.isEmpty =>
      e.withNewChildren(e.children.map(c => optimizeAdaptiveTransitions(c, Some(e))))
    case ColumnarToRowExec(GpuBringBackToHost(
        GpuCoalesceBatches(e: GpuShuffleExchangeExecBase, _))) if parent.isEmpty =>
      e.withNewChildren(e.children.map(c => optimizeAdaptiveTransitions(c, Some(e))))
```

and then

```scala
            // Directly wrap shuffle query stage with coalesce batches operator
            if (GpuShuffleEnv.isRapidsShuffleEnabled) {
              GpuCoalesceBatches(s, TargetSize(conf.gpuTargetBatchSizeBytes))
            } else {
              ShuffleCoalesceExec(s, conf)
            }
```

jlowe:
build

revans2 (commented):
Looks good to me, but I would like to see some more documentation because we now have more coupling between different parts of the code.

jlowe:
@abellina @andygrove @revans2 I think this is ready for another look.

jlowe:
build

revans2 (approved):

abellina:
Looking to build this and give it a whirl with UCX, will chime in if anything shows up.

abellina:
Ok I tried this with adaptive on and UCX on, and I see the coalesce node getting added after a custom shuffle reader. Also tried it with adaptive off, not seeing any issues.

abellina (approved):

andygrove:
The AQE changes look good to me.

andygrove (approved):

abellina:
There's one test failure you probably already saw, I think we just need to change the comparison on https://github.com/NVIDIA/spark-rapids/blob/branch-0.3/tests/src/test/scala/com/nvidia/spark/rapids/AdaptiveQueryExecSuite.scala#L176 to `ShuffleCoalesceExec`

jlowe:
build

jlowe:
I've updated the AQE join test to check for both forms of `ShuffleCoalesceExec` that can occur as inputs to a shuffled hash join.

abellina:
build

abellina (approved):

PR #1120, opened 2020-11-13
Add in support for filtering ArrayType

revans2:
This adds in support for filtering ArrayType columns.

One big change kind of related change is that ExprBaseMeta now checks both input and output types.  A lot of the code assumed that it was doing this already, but it was not.  This made some checks a lot simpler for operators that support only specific inputs, not outputs.

I also dropped the default max size of generated arrays in the tests because it was starting to take a while to generate some of the columns.

### DIFF
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -94,6 +94,7 @@ else
           $TEST_PARALLEL_OPTS \
           -v -rfExXs "$TEST_TAGS" \
           --std_input_path="$SCRIPTPATH"/src/test/resources/ \
+          --color=yes \
           "$TEST_ARGS" \
           $RUN_TEST_PARAMS \
           "$@"
@@ -107,6 +108,7 @@ else
           "$SCRIPTPATH"/runtests.py --rootdir "$SCRIPTPATH" "$SCRIPTPATH"/src/main/python \
           -v -rfExXs "$TEST_TAGS" \
           --std_input_path="$SCRIPTPATH"/src/test/resources/ \
+          --color=yes \
           "$TEST_ARGS" \
           $RUN_TEST_PARAMS \
           "$@"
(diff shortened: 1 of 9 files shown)


revans2:
build

integration_tests/run_pyspark_from_build.sh
```diff
@@ -94,6 +94,7 @@ else
           $TEST_PARALLEL_OPTS \
           -v -rfExXs "$TEST_TAGS" \
           --std_input_path="$SCRIPTPATH"/src/test/resources/ \
+          --color=yes \
```
revans2, line 97: not critical, just nice when building so I can see failures more easily.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -115,10 +114,6 @@ class GpuGetMapValueMeta(
     if (!isStringLit(expr.key)) {
       willNotWorkOnGpu("Only String literal keys are supported")
     }
-    expr.child.dataType match {
```
revans2, line 118: This is now covered by the `isSupportedType` below.

jlowe (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -679,20 +679,33 @@ abstract class BaseExprMeta[INPUT <: Expression](
...
         if (!ignoreUnsetDataTypes) {
           willNotWorkOnGpu(s"expression ${expr.getClass.getSimpleName} $expr " +
             s" does not have a corresponding dataType.")
         }
     }
+    val inputDataTypes = childExprs.map { expr =>
+      try {
+        expr.dataType
+      } catch {
+        case _ : java.lang.UnsupportedOperationException => null
```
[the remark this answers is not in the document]
revans2: This is something in Spark code that we cannot control.  Expression has an API to get the `dataType` and for some reason  one API decided to throw instead of returning something.  We could try and detect it with reflection and go on, but I think this is cleaner.

PR #1121, opened 2020-11-13
Shuffle/better error handling

abellina:
Mostly on the client side, this removes the global try/catch in `handleOp` and instead pipes errors into the `handleError` function in the fetch handler (shuffle iterator). This, in turn, creates `FetchFailed` exceptions that Spark will use when the task fails.

Removes a log-then-throw anti-pattern in the server. The server still has a global catch-all that logs then forgets an exception. The main issue on the server is there isn't really a good sink for the errors. I can add a follow up issue to add some logic to handle this, but would like to hear what people think. I think the best approach is a side-band message that the client can be waiting for, causing it to cancel any pending receives.

On the server side I was not safeClosing the `BufferSendStates` in one case, so I fixed that. This in turn will release buffers to the catalog, and frees send bounce buffers.

Follow up from: https://github.com/NVIDIA/spark-rapids/pull/994

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleServer.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleServer.scala
@@ -282,14 +282,7 @@ class RapidsShuffleServer(transport: RapidsShuffleTransport,
         throw new IllegalStateException(s"Error occurred while while handling metadata $tx")
       } else {
         logDebug(s"Received metadata request: $tx => $metaRequest")
-        try {
-          handleMetadataRequest(metaRequest)
-        } catch {
-          case e: Throwable => {
-            logError(s"Exception while handling metadata request from $tx: ", e)
-            throw e
-          }
-        }
+        handleMetadataRequest(metaRequest)
       }
     } finally {
       logDebug(s"Metadata request handled in ${TransportUtils.timeDiffMs(start)} ms")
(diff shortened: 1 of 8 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleServer.scala
```diff
@@ -366,74 +356,71 @@ class RapidsShuffleServer(transport: RapidsShuffleTransport,
...
-          }
-        }
-      })
+            } finally {
+              bufferTx.close()
+            })
+      }
+    } catch {
+      case t: Throwable =>
+        bufferSendStates.safeClose(t)
```
jlowe, line 422: Can probably use `closeOnExcept(bufferSendStates)` for the `try` block rather than needing an explicit `catch` here.

abellina:
build

abellina:
@jlowe this should be ready for another look.

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleIteratorSuite.scala
```diff
@@ -87,6 +87,49 @@ class RapidsShuffleIteratorSuite extends RapidsShuffleTestHelper {
...
+      mockCatalog,
+      123))
+
+    val ac = ArgumentCaptor.forClass(classOf[RapidsShuffleFetchHandler])
+    when(mockTransport.makeClient(any(), any())).thenReturn(client)
+    doNothing().when(client).doFetch(any(), ac.capture(), any())
+    cl.start()
+
+    val handler = ac.getValue.asInstanceOf[RapidsShuffleFetchHandler]
+    val causeException = new RuntimeException("Exception to suppress")
```
jlowe, line 109: Nit: 
```suggestion
    val causeException = new RuntimeException("some exception")
```

jlowe (approved):
Looks good,  just a small nit in the test.

abellina:
@jlowe nice catch, done!

abellina:
build

jlowe (approved):

PR #1124, opened 2020-11-15
Add shim layers for GpuWindowInPandasExec.

firestarman:
Databricks Spark has different requirements in node `WindowInPandasExec` from Apache Spark.

On Databricks, the parameter `windowExpression` passed in contains not only the window expressions, but also the input expressions from child.  Meanwhile it is nondeterministic, which means it may include either all of the child input expressions, or some of them or none of them, depending on which are required by the first following action.

What's more,  `windowExpression` is expected to be used as a project list for output, so you should project your result batches based on this 'list'.

This PR is also adding a Python daemon module since Databricks changes the API `worker`.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuPythonHelper.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuPythonHelper.scala
@@ -77,10 +77,10 @@ object GpuPythonHelper extends Logging {
     }
   }
 
-  def isPythonOnGpuEnabled(sqlConf: SQLConf): Boolean = {
+  def isPythonOnGpuEnabled(sqlConf: SQLConf, name: String = "spark"): Boolean = {
     val pythonEnabled = new RapidsConf(sqlConf).get(PYTHON_GPU_ENABLED)
     if (pythonEnabled) {
-      checkPythonConfigs(sparkConf)
+      checkPythonConfigs(sparkConf, name)
     }
     pythonEnabled
   }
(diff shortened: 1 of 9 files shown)


firestarman:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuPythonHelper.scala
```diff
@@ -100,11 +100,19 @@ object GpuPythonHelper extends Logging {
...
+    ("databricks", ("rapids.daemon_databricks", "rapids.worker"))
+  )
+
   // Check the related conf(s) to launch our rapids daemon or worker for
   // the GPU initialization when python on gpu enabled.
   // - python worker module if useDaemon is false, otherwise
   // - python daemon module.
-  private[sql] def checkPythonConfigs(conf: SparkConf): Unit = synchronized {
+  private[sql] def checkPythonConfigs(conf: SparkConf, name: String): Unit = synchronized {
+    val defaultDaemonWorker = mapDefaultPythonModules(name)
```
revans2, line 114: nit:
```
val (daemonModule, workerModule) =  mapDefaultPythonModules(name)
```

revans2 (approved):
Just a nit. But I think it really shows why we need to move away from trying to replace the python code ourselves and have an official python API to do the initialization that we want to do.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuPythonHelper.scala
firestarman: Will update this in the following PR to avoid approving it again.

tgravescs:
@firestarman  this didn't update the spark301db shim so this failed with databricks 7.3, please fix

PR #1129, opened 2020-11-16
Fix resource leaks with new shuffle optimization

jlowe:
#1116 accidentally dropped a fix for leaking Netty shuffle buffers, and that can result in warning messages like the following:
```
20/11/16 16:13:58 ERROR ResourceLeakDetector: LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
```

In addition the `ShuffleCoalesceIterator` can buffer `HostMemoryBuffer` objects yet was not installing a task completion listener to guarantee the buffers are closed even if the iterator is not drained (e.g.: in a limit query).


jlowe:
build

abellina (approved):
I tested this diff, lgtm.

revans2 (approved):

PR #1133, opened 2020-11-16
Refactor batch coalesce to be based solely on batch data size

jlowe:
#1116 placed a 2GB limit on the GPU batch target size which will prevent us from trying to coalesce batches that can exceed the 2GB element limit for any single column.  This PR refactors batch coalesce so it is based solely on data size rather than needing to also track string sizes.  This significantly simplifies the coalesce logic and also removes the need for the plugin to traverse the column hierarchy to calculate individual column sizes when the batch is single-buffer based (e.g.: compressed batch or `contiguous_split` batch).

Once cudf's `ColumnVector#getDeviceMemorySize` is updated to account for nested columns then even normal GPU batches only will need the plugin to traverse the top-level columns and not need to check for string/list types.

Note that the original batch coalesce code checked for batches containing `RapidsHostColumnVector`, but those column types should never be seen by a batch coalesce.  These types of batches only exist specifically between a partition and the shuffle, and there should never be a coalesce in-between.  This also removes the implicit size fallback if the column type is ignored.


jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
```diff
@@ -325,20 +295,10 @@ abstract class AbstractGpuCoalesceIterator(
               saveOnDeck(cb)
             } else if (wouldBeBytes > goal.targetSizeBytes && numBytes > 0) {
               saveOnDeck(cb)
```
revans2, line 297: nit: Might be nice to have a comment here about why we don't need to worry about overflow for strings/etc.  Might also be nice to have an assertion some where, perhaps in `TagetSize` itself to verify we did not make a mistake and go over the goal.  Just trying to be defensive in this.

revans2 (approved):
Looks good and I would be fine with this going in, but I would like to see us be a bit more defensive.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
jlowe: We cannot put in an assert on `TargetSize` since `RequireSingleBatch` forces us to coalesce above the 2GB goal if necessary.  If we cannot rely on the cudf C++ exception to cover the case where we exceed the cudf row limit then we'll need to continue checking individual column lengths.

jlowe: I should clarify that we can put an assert on `TargetSize`, but the code has to handle the case where we're above 2GB due to `RequireSingleBatch`.

revans2: Yes, I know which is why I wanted the comments and the assertion.  The comments are to explain to someone in the future that we are relying on CUDF concat on `RequireSingleBatch` to do the right thing and blow up if any of the child columns overflow, and that we are limiting the `TargetSize` to 2GB so there is no possibility of an overflow in that case. All of this is a nit anyways I am just trying to be defensive and if we have an assumption try and enforce that assumption and error out as early as possible.

jlowe: I added a `TargetSize` check and some comments.

jlowe:
build

jlowe:
I checked with @nvdbaranec on cudf's concatenate and apparently only one of the two possible code paths is properly checking for a row count overflow.  I'm marking this as a draft PR until a fix for that goes in, since this is removing the explicit string limit check in the plugin and relies on cudf::concatenate to catch this.

jlowe:
This is blocked pending https://github.com/rapidsai/cudf/pull/6809

jlowe:
The cudf dependency has been merged to cudf 0.17, so this is finally ready to go.

jlowe:
build

abellina (approved):

Issue #1096, opened 2020-11-11
[FEA] Implement parquet CreateDataSourceTableAsSelectCommand


tgravescs:
**Is your feature request related to a problem? Please describe.**
Currently we don't support spark org.apache.spark.sql.execution.command.CreateDataSourceTableAsSelectCommand.  We should add support for this for parquet.

Note there are other execs similar to this like CreateTable, CreateTableCommand, those will not be covered here.

PR #1134, opened 2020-11-17
Support saveAsTable for writing orc and parquet

tgravescs:
This adds support for saveAsTable and create table from select sql statements.

This is basically just metadata operations and then calling into the existing GpuInsertIntoHadoopFsRelationCommand.

I ended up copying the Spark DataSource and making a GpuDatasource version, that is a slightly modified version. One of the main changes is we pass in the provider and the fileformat because we already have to figure that out in the GpuOverrides CreateDataSourceTableAsSelectCommandMeta. 

Other than that I split the parquet and orc tests into read and write files. I added a bunch of different write tests

I had to add a shim function because https://issues.apache.org/jira/browse/SPARK-32431 went into Spark 3.1 and that changed the call into checkColumnNameDuplication.

Also note that Spark hasn't enabled any of the datasource v2 writers, it always falls back to the v1 version at this point.

fixes https://github.com/NVIDIA/spark-rapids/issues/1096

### DIFF
--- a/integration_tests/src/main/python/asserts.py
+++ b/integration_tests/src/main/python/asserts.py
@@ -238,6 +238,42 @@ def assert_gpu_and_cpu_writes_are_equal_iterator(write_func, read_func, base_pat
     """
     _assert_gpu_and_cpu_writes_are_equal(write_func, read_func, base_path, False, conf=conf)
 
+def assert_gpu_fallback_write(write_func,
+        read_func,
+        base_path,
+        cpu_fallback_class_name,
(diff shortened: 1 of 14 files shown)


tgravescs:
build

revans2 (approved):
For the code that was copied and pasted from spark I mostly checked to see if it matched what spark had.  I am a little concerned with the amount of code that we have copied over, but it should be fine for now.

jlowe (approved):

PR #1137, opened 2020-11-17
Let GpuWindowInPandas declare ArrayType supported.

firestarman:
This PR is to let plan  `GpuWindowInPandasExec`  and two related expressions declare array type supported, since cuDF has already supported reading an array type column from a stream in arrow format and putting it into a cuDF column vector. Then we can return array type data from Pandas UDF and receive array type data at JVM side in this plan. 

The new added IT tests only cover the two cases mentioned above. Other cases, such as array type data exsits in the input, are not verfied yet.

Also remove the single batch requirement for output, since it takes much more time when there are plenty of groups.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -1505,8 +1505,21 @@ object GpuOverrides {
             childExprs.map(_.convertToGpu()),
             a.evalType, a.udfDeterministic, a.resultId)
 
-        override def isSupportedType(t: DataType): Boolean =
-          GpuOverrides.isSupportedType(t, allowArray = true, allowNesting = true)
+        override def tagExprForGpu(): Unit = {
+          // Allow array type data only in output for the top Python UDF. Because seems
+          // there is something wrong in writing array type data by cuDF arrow IPC writer.
+          wrapped.children
+            .filter(_.dataType.isInstanceOf[ArrayType])
+            .foreach(e => willNotWorkOnGpu(s"array type data($e) as the input of Python UDF is" +
+              " not supported on GPU"))
+        }
+
+        // Allow only one level array type data which has been verified now.
+        // Will update this after verifying the nested array type.
+        override def isSupportedType(t: DataType): Boolean = t match {
+          case ArrayType(elementType, _) => super.isSupportedType(elementType)
+          case _ => super.isSupportedType(t)
+        }
       }
     ),
     expr[Rand](
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -1516,10 +1516,8 @@ object GpuOverrides {
 
         // Allow only one level array type data which has been verified now.
         // Will update this after verifying the nested array type.
-        override def isSupportedType(t: DataType): Boolean = t match {
-          case ArrayType(elementType, _) => super.isSupportedType(elementType)
-          case _ => super.isSupportedType(t)
-        }
+        override def isSupportedType(t: DataType): Boolean =
+          GpuOverrides.isSupportedType(t, allowArray = true)
       }
     ),
     expr[Rand](
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -1504,6 +1504,9 @@ object GpuOverrides {
           GpuPythonUDF(a.name, a.func, a.dataType,
             childExprs.map(_.convertToGpu()),
             a.evalType, a.udfDeterministic, a.resultId)
+
+        override def isSupportedType(t: DataType): Boolean =
+          GpuOverrides.isSupportedType(t, allowArray = true, allowNesting = true)
       }
     ),
     expr[Rand](
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
@@ -166,6 +166,9 @@ class GpuWindowExpressionMeta(
       childExprs.head.convertToGpu(),
       childExprs(1).convertToGpu().asInstanceOf[GpuWindowSpecDefinition]
     )
+
+  override def isSupportedType(t: DataType): Boolean =
+    GpuOverrides.isSupportedType(t, allowArray = true, allowNesting = true)
 }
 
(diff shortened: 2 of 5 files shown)


firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -166,6 +166,9 @@ class GpuWindowExpressionMeta(
       childExprs.head.convertToGpu(),
       childExprs(1).convertToGpu().asInstanceOf[GpuWindowSpecDefinition]
     )
+
+  override def isSupportedType(t: DataType): Boolean =
```
revans2, line 170: Do we need to add some extra checks into `tagExprForGpu` so that we only allow it in the cases we care about/tested?  Specifically I mean for PythonUDFs. We are probably okay for now because we did not allow arrays for the regular `GpuWindowExec`, but it might be nice to put in a comment about this.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1460,6 +1460,9 @@ object GpuOverrides {
           GpuPythonUDF(a.name, a.func, a.dataType,
             childExprs.map(_.convertToGpu()),
             a.evalType, a.udfDeterministic, a.resultId)
+
+        override def isSupportedType(t: DataType): Boolean =
+          GpuOverrides.isSupportedType(t, allowArray = true, allowNesting = true)
```
revans2, line 1465: We have not tested array inputs or nested arrays.  I would like to see some basic tests for those too, or else we need to add some checks so that we don't convert plans with them in it if we cannot support them yet.

revans2 (commented):
Overall this looks really good.  I would just like to see a little bit more documentation to try and help someone else out in the future and some more tests to show that we actually can support what we have enabled.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
firestarman: Yes we need. Updated to allow array type only for Python UDF.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
firestarman: Since there is something wrong in writing array data by cudf arrow IPC writer, i chose to add some checks.
I have tried excluding all the possible cases which are enabled but not intended. If any missing, i will take it as a follow.

firestarman:
build

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
firestarman: Yes, we do, i am working on this

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1504,6 +1504,22 @@ object GpuOverrides {
           GpuPythonUDF(a.name, a.func, a.dataType,
             childExprs.map(_.convertToGpu()),
             a.evalType, a.udfDeterministic, a.resultId)
+
+        override def tagExprForGpu(): Unit = {
+          // Allow array type data only in output for the top Python UDF. Because seems
+          // there is something wrong in writing array type data by cuDF arrow IPC writer.
```
revans2, line 1510: nit: Could you file an issue for this?  Ideally we would file the issue and include a link to the issue here so it is clear when we could fix it.  But it is okay to check this in first if it will take some time to create a reproducible use case.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1504,6 +1504,22 @@ object GpuOverrides {
...
+          wrapped.children
+            .filter(_.dataType.isInstanceOf[ArrayType])
+            .foreach(e => willNotWorkOnGpu(s"array type data($e) as the input of Python UDF is" +
+              " not supported on GPU"))
+        }
+
+        // Allow only one level array type data which has been verified now.
+        // Will update this after verifying the nested array type.
+        override def isSupportedType(t: DataType): Boolean = t match {
```
revans2, line 1519: nit: I think you can achieve the same thing with

```
GpuOverrides.isSupportedType(allowArray=true, allowNesting=false)
```

`allowNesting` is false by default so it is probably not needed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -166,6 +173,11 @@ class GpuWindowExpressionMeta(
       childExprs.head.convertToGpu(),
       childExprs(1).convertToGpu().asInstanceOf[GpuWindowSpecDefinition]
     )
+
+  // Allow array type only for Python UDF, also add an extra check in
+  // `tagExprForGpu` to make sure this.
+  override def isSupportedType(t: DataType): Boolean =
+    GpuOverrides.isSupportedType(t, allowArray = true, allowNesting = true)
```
revans2, line 180: We have not tested nested arrays in the output i.e.  `ArrayType(ArrayType(LongType))` I would prefer that we don't explicitly allow it without a test for it.

revans2 (commented):
Overall it looks really good with just a few nits.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
firestarman: Here it is https://github.com/NVIDIA/spark-rapids/issues/1171

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
firestarman: Oh ye, updated. Looks like i did not fully understand this API.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
firestarman: Nice catch.
I meant one level array. Updated. Looks like i did not fully understand this API `GpuOverrides.isSupportedType`.

firestarman:
rebased

firestarman:
updated

firestarman:
build

firestarman:
build

firestarman:
rebased again

firestarman:
build

revans2 (approved):

PR #1138, opened 2020-11-17
[REVIEW] Support decimal type for GpuProjectExec

sperlingxx:
The target of current PR is to support decimal type for `GpuProjectExec`.  There are several prerequisites for this target:
* support decimal data in `GpuColumnVector`
* support decimal data in `GpuScalar`
* decimal data transmission between cpu and gpu 
   * `GpuRowToColumnarExec`
   * `GpuColumnarToRowExec`
   * `HostColumnarToGpu`

The PR also enables decimal type in some basic expressions (Literal/Alias/AttributeReference/IsNull/IsNotNull).

In addition, `DECIMAL64` is the only supported `DType.ID` for decimal column vectors and scalars in terms of spark-rapids. This restriction is to ensure one to one mapping relation between `DataType` and `DType`, in case of DType transaction caused by precision promotion.

### DIFF
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
@@ -17,6 +17,7 @@
 
 package com.nvidia.spark.rapids;
 
+import ai.rapids.cudf.DType;
 import ai.rapids.cudf.HostColumnVectorCore;
 
 import org.apache.spark.sql.types.ArrayType;
@@ -30,6 +31,8 @@ import org.apache.spark.sql.vectorized.ColumnarArray;
 import org.apache.spark.sql.vectorized.ColumnarMap;
 import org.apache.spark.unsafe.types.UTF8String;
 
+import java.math.BigDecimal;
+import java.math.RoundingMode;
 
 /**
  * A GPU accelerated version of the Spark ColumnVector.
@@ -159,7 +162,10 @@ public class RapidsHostColumnVectorCore extends ColumnVector {
 
   @Override
   public final Decimal getDecimal(int rowId, int precision, int scale) {
-    throw new IllegalStateException("The decimal type is currently not supported by rapids cudf");
+    assert precision <= DType.DECIMAL64_MAX_PRECISION : "Assert " + precision + " <= DECIMAL64_MAX_PRECISION(" + DType.DECIMAL64_MAX_PRECISION + ")";
+    assert cudfCv.getType().getTypeId() == DType.DTypeEnum.DECIMAL64: "Assert DType to be DECIMAL64";
+    assert scale == -cudfCv.getType().getScale() : "Assert fetch decimal with its original scale";
+    return Decimal.createUnsafe(cudfCv.getLong(rowId), precision, scale);
   }
 
   @Override
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -648,7 +648,7 @@ object GpuOverrides {
         }
 
         override def isSupportedType(t: DataType): Boolean =
-          GpuOverrides.isSupportedType(t, allowCalendarInterval = true)
+          GpuOverrides.isSupportedType(t, allowCalendarInterval = true, allowDecimal = true)
       }),
     expr[Signum](
       "Returns -1.0, 0.0 or 1.0 as expr is negative, 0 or positive",
@@ -663,7 +663,8 @@ object GpuOverrides {
             allowMaps = true,
             allowArray = true,
             allowStruct = true,
-            allowNesting = true)
+            allowNesting = true,
+            allowDecimal = true)
 
         override def convertToGpu(child: Expression): GpuExpression =
           GpuAlias(child, a.name)(a.exprId, a.qualifier, a.explicitMetadata)
@@ -676,7 +677,8 @@ object GpuOverrides {
             allowMaps = true,
             allowArray = true,
             allowStruct = true,
-            allowNesting = true)
+            allowNesting = true,
+            allowDecimal = true)
 
         // This is the only NOOP operator.  It goes away when things are bound
         override def convertToGpu(): Expression = att
@@ -873,7 +875,8 @@ object GpuOverrides {
             allowMaps = true,
             allowArray = true,
             allowStruct = true,
-            allowNesting = true)
+            allowNesting = true,
+            allowDecimal = true)
 
         override def convertToGpu(child: Expression): GpuExpression = GpuIsNull(child)
       }),
@@ -885,7 +888,8 @@ object GpuOverrides {
             allowMaps = true,
             allowArray = true,
             allowStruct = true,
-            allowNesting = true)
+            allowNesting = true,
+            allowDecimal = true)
 
         override def convertToGpu(child: Expression): GpuExpression = GpuIsNotNull(child)
       }),
@@ -1894,7 +1898,8 @@ object GpuOverrides {
               allowMaps = true,
               allowArray = true,
               allowStruct = true,
-              allowNesting = true)
+              allowNesting = true,
+              allowDecimal = true)
 
           override def convertToGpu(): GpuExec =
             GpuProjectExec(childExprs.map(_.convertToGpu()), childPlans(0).convertIfNeeded())
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
@@ -111,9 +111,10 @@ private object GpuRowToColumnConverter {
         StructConverter(st.fields.map(getConverterFor))
       case (st: StructType, false) =>
         NotNullStructConverter(st.fields.map(getConverterFor))
-      // NOT SUPPORTED YET
-      // case dt: DecimalType => new DecimalConverter(dt)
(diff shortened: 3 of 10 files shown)


firestarman:
build

sperlingxx:
build

sperlingxx:
rebuild

revans2 (commented):
This code change is not going to provide any functionality to the end user in practice, and makes it very difficult to add in more functionality until we can convert a decimal Scalar into a ColumnVector.

We cannot support literal decimal values in a GpuProjectExec, GpuFilterExec or other similar processing until we can turn a `Scalar` into a `ColumnVector`. Similarly all binary and unary operators in Spark also rely on this functionality.  So that means we cannot enable any of them without it, if we say that we support decimal literals.

Without support for a GpuProjectExec or a GpuFilterExec there is no reason to support moving the data to or from the GPU because there will never be a situation where the data will do that.

For me I would much rather see us not try to support Literal initially and instead add in support for GpuProjectExec and GpuFilterExec, with tests, because then we could have a decimal that is a part of the query, but is not directly processed.

i.e.
```
select a_int, b_decimal from bar where a_int >= 100
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -598,7 +598,7 @@ object GpuOverrides {
         }
 
         override def isSupportedType(t: DataType): Boolean =
-          GpuOverrides.isSupportedType(t, allowCalendarInterval = true)
+          GpuOverrides.isSupportedType(t, allowCalendarInterval = true, allowDecimal = true)
```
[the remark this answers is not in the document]
revans2: Never mind https://github.com/rapidsai/cudf/pull/6723 was merged in.  So all we really have to worry about is tests for those use cases.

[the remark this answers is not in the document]
sperlingxx: The test case of "decimal scalar to column vector"  is enabled since jar package of cuDF 0.17-SNAPSHOT updated to latest.

sperlingxx:
build

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
```diff
@@ -117,6 +117,23 @@ object HostColumnarToGpu {
         for (i <- 0 until rows) {
           b.appendUTF8String(cv.getUTF8String(i).getBytes)
         }
+      case (dt, nullable) if dt.isDecimalType =>
+        // TODO: cuDF should support appending unscaledValue for decimal
```
revans2, line 121: Aren't we doing that for the row to columnar transition?

revans2 (dismissed):
Looks good, just one question/possible improvement.

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
sperlingxx: Oh, it is a mistake. I've fixed it.

firestarman:
build

revans2 (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
```diff
@@ -113,8 +115,24 @@ object HostColumnarToGpu {
...
+            } else {
+              b.append(cv.getDecimal(i, DType.DECIMAL64_MAX_PRECISION, dt.scale).toUnscaledLong)
+            }
+          }
+        } else {
+          for (i <- 0 until rows) {
+            b.append(cv.getDecimal(i, DType.DECIMAL64_MAX_PRECISION, dt.scale).toUnscaledLong)
+          }
+        }
+      case (t, n) =>
```
abellina, line 134:
```suggestion
      case (t, _) =>
```

abellina (commented):
nit, one change that was reverted from a prior revision.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
```diff
@@ -113,8 +115,24 @@ object HostColumnarToGpu {
...
+        throw new UnsupportedOperationException(s"Converting to GPU for ${t} is not currently " +
           s"supported")
```
abellina, line 136: also, not related to this diff, but while you are at it:
```suggestion
          "supported")
```

sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
```diff
@@ -159,7 +161,9 @@ public final ColumnarMap getMap(int ordinal) {
 
   @Override
   public final Decimal getDecimal(int rowId, int precision, int scale) {
-    throw new IllegalStateException("The decimal type is currently not supported by rapids cudf");
+    BigDecimal bigDec = cudfCv.getBigDecimal(rowId).setScale(scale, RoundingMode.UNNECESSARY);
+    assert bigDec.precision() <= precision : "Assert " + bigDec.precision() + " <= " + precision;
+    return Decimal.apply(bigDec);
```
jlowe, line 166: This is performance sensitive, since it's called by the columnar to row conversion.  As such it would be great to avoid the full `BigDecimal` treatment if possible.  AFAICT the scale being requested here should always match the underlying scale being used.  Should we do something like the following to avoid `BigDecimal` in what I believe will be the common (only?) case in practice?
```suggestion
    if (precision <= DType.DECIMAL64_MAX_PRECISION && scale == cudf.getType().getScale()) {
      assert cudfCv.getType().getTypeId() == DTypeEnum.DECIMAL64;
      return Decimal.createUnsafe(cudfCv.getLong(rowId), precision, scale);
    }
    BigDecimal bigDec = cudfCv.getBigDecimal(rowId).setScale(scale, RoundingMode.UNNECESSARY);
    assert bigDec.precision() <= precision : "Assert " + bigDec.precision() + " <= " + precision;
    return Decimal.apply(bigDec);
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
```diff
@@ -480,27 +481,38 @@ private object GpuRowToColumnConverter {
...
+    }
+
+    override def getNullSize: Double = 8 + VALIDITY
+  }
+
+  private class NotNullDecimalConverter(precision: Int, scale: Int) extends TypeConverter {
+    override def append(
+      row: SpecializedGetters,
+      column: Int,
+      builder: ai.rapids.cudf.HostColumnVector.ColumnBuilder): Double = {
```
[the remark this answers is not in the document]
revans2: Yes or it would not compile.  The Double is needed because we are estimating the size of the output data in bytes, but validity is not a full byte, so it is a bit of a hack work around.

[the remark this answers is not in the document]
sperlingxx: The return type is defined in its parent class `TypeConverter`.  I am not sure why it should be `Double`.

sperlingxx:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
revans2: I am fairly confident that the scale/precision will always match. So could we just turn it into assertions instead of an if/else? I would like to avoid conditionals if possible too.

```
assert precision <= DType.DECIMAL64_MAX_PRECISION : ...
assert cudf.getType().getTYpeId() == DTypeEnum.DECIMAL64 : ....
assert scale == cudf.getType().getScale() : ....
return Decimal.createUnsafe(cudfCv.getLong(rowId), precision, scale);
```

revans2: Also I forgot to add that once this goes in and we have integration test support for decimal I will update the fast path to support decimal too.

jlowe: Sounds great.  I'm all for removing the `BigDecimal` fallback code if we're confident that will never be necessary.  Actually I think we can also remove the `assert cudf.getType().getTYpeId() == DTypeEnum.DECIMAL64` I proposed since `cudfCv.getLong` will itself assert that the type is backed by a long.

revans2:
The test failures look unrelated.

```
ansi_cast timestamps to long *** FAILED ***
org.apache.spark.sql.AnalysisException: cannot resolve 'CAST(`c0` AS BIGINT)' due to data type mismatch: cannot cast timestamp to bigint
```

Looks like something changed in spark 3.0.2.  I'll file an issue for it.

firestarman:
build

sperlingxx:
build

jlowe:
Please do not squash your commits while a PR is being reviewed.  It makes it much more difficult to determine what has changed since the last time it was reviewed.  If you need to upmerge, merge into your branch instead of rebasing.  It doesn't require a force-push and often won't dismiss prior approvals.  We squash when committing a PR, so I see no reason to squash during review.

jlowe (approved):

revans2 (approved):

revans2:
I would prefer to get this in sooner than later so we can unblock some other decimal work and we can hopefully start working on integration tests too, so I am going to merge this now.

PR #1152, opened 2020-11-18
Add spark301db shim layer for WindowInPandas.

firestarman:
Spark300db was removed, so need to add it back to spark301db.

Also adding some metrics.

### DIFF
--- a/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
+++ b/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
@@ -133,7 +133,7 @@ class Spark310Shims extends Spark301Shims {
   }
 
   override def getExecs: Map[Class[_ <: SparkPlan], ExecRule[_ <: SparkPlan]] = {
-    Seq(
+    super.getExecs ++ Seq(
       GpuOverrides.exec[FileSourceScanExec](
         "Reading data from files, often from Hive tables",
         (fsse, conf, p, r) => new SparkPlanMeta[FileSourceScanExec](fsse, conf, p, r) {
@@ -196,7 +196,7 @@ class Spark310Shims extends Spark301Shims {
       GpuOverrides.exec[ShuffledHashJoinExec](
         "Implementation of join using hashed shuffled data",
         (join, conf, p, r) => new GpuShuffledHashJoinMeta(join, conf, p, r))
-    ).map(r => (r.getClassFor.asSubclass(classOf[SparkPlan]), r)).toMap
+    ).map(r => (r.getClassFor.asSubclass(classOf[SparkPlan]), r))
   }
 
   override def getScans: Map[Class[_ <: Scan], ScanRule[_ <: Scan]] = Seq(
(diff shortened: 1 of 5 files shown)


firestarman:
build

firestarman:
If for databricks, do we need to add the shim layer for each spark version ? Such as spark300db, spark301db, or later spark310db ?

revans2:
spark300db is in the process of being removed, but we should look into spark310db just to see if the version from spark301db will work or not.

firestarman:
OK, 
And for this PR, I verified it by running a local Jenkins job over Databricks spark301, and it passed. If no other concerns, it is ready to merge, i think.

revans2:
I think we are missing support for regular 3.1.0 spark and possibly 3.0.1 I am seeing failures in the nightly tests.

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
```diff
@@ -133,7 +133,7 @@ class Spark310Shims extends Spark301Shims {
   }
 
   override def getExecs: Map[Class[_ <: SparkPlan], ExecRule[_ <: SparkPlan]] = {
-    Seq(
+    super.getExecs ++ Seq(
```
firestarman, line 136: Fix the failures of the regular 3.1.0 spark nightly.

firestarman:
rebased

firestarman:
build

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
firestarman, line 136: Fix the faliures about window tests for the regular 310 spark.

firestarman (commented):
> I think we are missing support for regular 3.1.0 spark and possibly 3.0.1 I am seeing failures in the nightly tests.

Yes.

Spark 300 and 301 nightly are passed. https://blossom.nvidia.com/sw-gpu-spark-jenkins/job/rapids_integration-0.3-github/ 

The new added shim layer in this PR should fix the failures about window tests in databricks 301 spark.

And the change as below should fix the window tests failures for regular 310 spark.

revans2 (approved):

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
revans2: @tgravescs do you have any issues with this?  If we get breaking changes between 301 and 310 we might need to change this, but I wanted to check with you that you don't see any other issues before merging.

jlowe (approved):
Merging this to fix the nightly tests.  @tgravescs we can address concerns about pulling in the map of execs from Spark301Shims in a separate PR if that's not desired long-term.

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
tgravescs: it seems fine until something changes that breaks it.

Issue #1079, opened 2020-11-06
[BUG]The Scala UDF function cannot invoke the UDF compiler when it's passed to "explode"


wjxiz1992:
**Describe the bug**
When a UDF is passed to explode() function like `explode(myUdf)`, the UDF compiler will not be used to try to compile this `myUdf`.

**Root cause**
The udf plugin didn't try to replace a udf is this udf is a child plan of `Project`

**Steps/Code to reproduce bug**

```
    val myudf: (String) => Array[String] = a => {
      a.split(",")
    }
    val u = makeUdf(myudf)
    val dataset = List("first,second").toDF("x").repartition(1)
    var result = dataset.withColumn("new", explode(u(col("x"))))
    result.explain(true)
```

**Expected behavior**
I should be able to see the compiled Catalyst expressions like "split(x....) AS new" after "Analyzed Logical Plan".

**Environment details (please complete the following information)**
 - Environment location: Local in an IDEA debug environment
 - Spark configuration settings related to the issue :     
```
    .set("spark.sql.extensions", "com.nvidia.spark.udf.Plugin")
    .set("spark.rapids.sql.udfCompiler.enabled", "true")
    .set("spark.rapids.sql.test.enabled", "true")
```

wjxiz1992:
With the fix from@abellina, the project child of `explode` can be compiled now.
```
diff --git a/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala b/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
index 679d126b9..98164e70e 100644
--- a/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
+++ b/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
@@ -82,7 +82,7 @@ case class LogicalPlanRules() extends Rule[LogicalPlan] with Logging {
       plan match {
         case project: Project =>
           Project(project.projectList.map(e => attemptToReplaceExpression(plan, e))
-              .asInstanceOf[Seq[NamedExpression]], project.child)
+              .asInstanceOf[Seq[NamedExpression]], apply(project.child))
         case x => {
           x.transformExpressions(replacePartialFunc(plan))
         }
```
The real problem and correct description for this issue turns to be "child of Project cannot be replaced".  
@abellina Do you think we can close this issue with your fix and create another issue for `GpuExplode` ? It's for the sql plugin side, not a UDF problem.

PR #1153, opened 2020-11-18
Scala UDF will compile children expressions in Project

wjxiz1992:

Co-authored-by: Alessandro Bellina <abellina@nvidia.com>

This PR aims to resolve https://github.com/NVIDIA/spark-rapids/issues/1079.

The current fix lets the compiler compile Logical Plan recursively.
Originally, any udf that is a child of Project, will not be compiled.

### DIFF
--- a/udf-compiler/src/test/scala/com/nvidia/spark/OpcodeSuite.scala
+++ b/udf-compiler/src/test/scala/com/nvidia/spark/OpcodeSuite.scala
@@ -2333,29 +2333,30 @@ class OpcodeSuite extends FunSuite {
     }
 
     val u = makeUdf((x: String, y: String, z: Boolean) => {
-        var r = new mutable.ArrayBuffer[String]()
-        r = r :+ x
-        if (!cond(y)) {
-          r = r :+ y
+      var r = new mutable.ArrayBuffer[String]()
+      r = r :+ x
+      if (!cond(y)) {
+        r = r :+ y
 
-          if (z) {
-            r = r :+ transform(y)
-          }
-        }
         if (z) {
-          r = r :+ transform(x)
+          r = r :+ transform(y)
         }
-        r.distinct.toArray
-      })
+      }
+      if (z) {
+        r = r :+ transform(x)
+      }
+      r.distinct.toArray
+    })
 
     val dataset = List(("######hello", null),
-                       ("world", "######hello"),
-                       ("", "@@@@target")).toDF("x", "y")
+      ("world", "######hello"),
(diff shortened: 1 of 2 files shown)


wjxiz1992:
Hi @abellina, as we discussed offline, the current fix can solve the `explode` issue, but not perfect yet. Can you guide me here for a better solution?

udf-compiler/src/test/scala/com/nvidia/spark/OpcodeSuite.scala
```diff
@@ -2356,5 +2356,26 @@ class OpcodeSuite extends FunSuite {
                    ("world", "######hello", Array("world", "######hello", "@@@@hello")),
                    ("", "@@@@target", Array("", "@@@@target", "######target", null))).toDF
     checkEquiv(result, ref)
+
```
abellina, line 2359: @wjxiz1992 there's a missing `}` here.

abellina:
The physical plan for the explode test is:
```
== Physical Plan ==
Generate explode(split(x#5, ,, -1)), [x#5], false, [new#8]
+- Exchange RoundRobinPartitioning(1), false, [id=#9]
   +- LocalTableScan [x#5]
```

which looks correct to me.

The physical plan for the `flatten` test isn't. Do you expect this to work for an expression in a `flatten`? If so, this approach seems to fail there:

```
== Physical Plan ==
*(1) Project [x#5, flatten(UDF(x#5)) AS new#10]
+- Exchange RoundRobinPartitioning(1), false, [id=#9]
   +- LocalTableScan [x#5]
```

abellina:
Looking at this more it seems the `flatten` case is being looked at, but the UDF doesn't compile:

```
org.apache.spark.SparkException: Unsupported type for new:scala.collection.mutable.ArrayOps$ofRefProject [x#5, flatten(UDF(x#5)) AS new#10]
```

It looks to be a test issue.

jlowe:
Is this critical for the 0.3 release?  We are wrapping up the 0.3 release very soon, so this probably should be retargeted to branch-0.4.

jlowe (changes requested):
Marking this as changes needed since it is still targeting branch-0.3 despite the PR project being set to Release 0.4.

wjxiz1992:
Rebased on branch-0.4.
Hi @abellina can you take another look at this? the Flatten test case has been removed.

jlowe:
build

jlowe (approved):
Thanks for updating to 0.4.  Looks OK to me, but it would be good for @abellina to review as well before committing.

abellina (approved):

Issue #1110, opened 2020-11-13
[BUG] Special date "now" has wrong value on GPU


andygrove:
**Describe the bug**
Spark has a number of special dates such as "now", "today", "yesterday", "tomorrow", where "today" is meant to be midnight on the current day and "now" is meant to be the current timestamp. It looks like we are currently using the same value for both "now" and "today" which is incorrect.

**Steps/Code to reproduce bug**
We need to improve the tests. We will likely need to write custom tests for "now" because CPU and GPU runs will naturally produce different results because they run at different times.

**Expected behavior**
Behavior should match Spark.

PR #1159, opened 2020-11-18
fix bug where 'now' had same value as 'today' for timestamps

andygrove:

When parsing dates, the special values `now` and `today` have the same value, but for timestamps, `now` should reflect the current time and `today` should reflect the start of the day.

I also moved more date/time handling from GpuCast to DateUtils.

This closes https://github.com/NVIDIA/spark-rapids/issues/1110

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/DateUtils.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/DateUtils.scala
@@ -16,7 +16,11 @@
 
 package com.nvidia.spark.rapids
 
+import java.time.ZoneId
+
 import scala.collection.mutable.ListBuffer
+
+import org.apache.spark.sql.catalyst.util.DateTimeUtils
 
 /**
  * Class for helper functions for Date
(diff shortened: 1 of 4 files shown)


andygrove:
build

jlowe:
test failures seem related

andygrove:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DateUtils.scala
```diff
@@ -42,6 +46,47 @@ object DateUtils {
...
   val ONE_DAY_MICROSECONDS = 86400000000L
 
+  val EPOCH = "epoch"
+  val NOW = "now"
+  val TODAY = "today"
+  val YESTERDAY = "yesterday"
+  val TOMORROW = "tomorrow"
+
+  def specialDatesDays: Map[String, Int] = {
+    val today = DateTimeUtils.currentDate(ZoneId.of("UTC"))
```
jlowe, line 56: I realize this isn't specific to this change, but should we be using a hardcoded UTC timezone here?  I realize it works because we only support UTC for timestamps, but what if the user is using dates (not timestamps) in a different timezone than UTC?

andygrove: That's a good question. I've been reading a blog post about time zone handling in Spark 3.0.0 [1] and it does state that `now` should use the session time zone, so it does seem better to remove the explicit UTC timezone. I will try this out.

[1] https://databricks.com/blog/2020/07/22/a-comprehensive-look-at-dates-and-timestamps-in-apache-spark-3-0.html

jlowe (approved):

jlowe:
I expect this needs to be upmerged to branch-0.3 in order for CI to pass.

andygrove:
build

Issue #724, opened 2020-09-10
[BUG] PySpark test_broadcast_nested_loop_join_special_case intermittent failure


andygrove:
**Describe the bug**
This test failed seemingly randomly on one of my PRs. It worked on a second run.

```
[2020-09-10T16:53:48.414Z] =================================== FAILURES ===================================
[2020-09-10T16:53:48.414Z] _____________ test_broadcast_nested_loop_join_special_case[String] _____________
[2020-09-10T16:53:48.414Z] 
[2020-09-10T16:53:48.414Z] data_gen = String
[2020-09-10T16:53:48.414Z] 
[2020-09-10T16:53:48.414Z]     @ignore_order(local=True)
[2020-09-10T16:53:48.414Z]     @pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
[2020-09-10T16:53:48.414Z]     def test_broadcast_nested_loop_join_special_case(data_gen):
[2020-09-10T16:53:48.414Z]         def do_join(spark):
[2020-09-10T16:53:48.414Z]             left, right = create_df(spark, data_gen, 50, 25)
[2020-09-10T16:53:48.414Z]             return left.crossJoin(broadcast(right)).selectExpr('COUNT(*)')
[2020-09-10T16:53:48.414Z] >       assert_gpu_and_cpu_are_equal_collect(do_join, conf={'spark.rapids.sql.exec.BroadcastNestedLoopJoinExec': 'true'})
[2020-09-10T16:53:48.414Z] 
[2020-09-10T16:53:48.414Z] ../../src/main/python/join_test.py:106: 
[2020-09-10T16:53:48.414Z] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[2020-09-10T16:53:48.414Z] ../../src/main/python/asserts.py:296: in assert_gpu_and_cpu_are_equal_collect
[2020-09-10T16:53:48.414Z]     _assert_gpu_and_cpu_are_equal(func, True, conf=conf)
[2020-09-10T16:53:48.414Z] ../../src/main/python/asserts.py:288: in _assert_gpu_and_cpu_are_equal
[2020-09-10T16:53:48.414Z]     assert_equal(from_cpu, from_gpu)
[2020-09-10T16:53:48.414Z] ../../src/main/python/asserts.py:86: in assert_equal
[...]

PR #1165, opened 2020-11-19
Fix race condition in SerializeConcatHostBuffersDeserializeBatch

andygrove:

The following code had a race condition. The object was created with numRows = 0 and then updated. Another thread could access the object before the row count was set.

```scala
this.batchInternal = new ColumnarBatch(new Array[ColumnVector](0))
batchInternal.setNumRows(numRows.toInt)
```

I also changed `readObject` to return an object rather than set the shared object as a side-effect.

This closes https://github.com/NVIDIA/spark-rapids/issues/724

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
@@ -50,7 +50,7 @@ class SerializeConcatHostBuffersDeserializeBatch(
   extends Serializable with Arm with AutoCloseable {
   @transient private val headers = data.map(_.header)
   @transient private val buffers = data.map(_.buffer)
-  @transient private var batchInternal: ColumnarBatch = null
+  @transient @volatile private var batchInternal: ColumnarBatch = _
 
   def batch: ColumnarBatch = {
     if (batchInternal == null) {
@@ -60,7 +60,7 @@ class SerializeConcatHostBuffersDeserializeBatch(
       writeObject(oout)
       val barr = out.toByteArray
       val oin = new ObjectInputStream(new ByteArrayInputStream(barr))
-      readObject(oin)
+      batchInternal = readObject(oin)
     }
     batchInternal
   }
@@ -81,7 +81,7 @@ class SerializeConcatHostBuffersDeserializeBatch(
     }
   }
 
-  private def readObject(in: ObjectInputStream): Unit = {
+  private def readObject(in: ObjectInputStream): ColumnarBatch = {
     val range = new NvtxRange("DeserializeBatch", NvtxColor.PURPLE)
     try {
       val tableInfo: JCudfSerialization.TableAndRowCountPair =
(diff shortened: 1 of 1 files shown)


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
```diff
@@ -90,8 +90,9 @@ class SerializeConcatHostBuffersDeserializeBatch(
         val table = tableInfo.getTable
         if (table == null) {
           val numRows = tableInfo.getNumRows
-          this.batchInternal = new ColumnarBatch(new Array[ColumnVector](0))
-          batchInternal.setNumRows(numRows.toInt)
+          val columnarBatch = new ColumnarBatch(new Array[ColumnVector](0))
+          columnarBatch.setNumRows(numRows.toInt)
+          this.batchInternal = columnarBatch
```
jlowe, line 95:
```suggestion
          this.batchInternal = new ColumnarBatch(new Array[ColumnVector](0), numRows.toInt)
```

jlowe:
@andygrove can you elaborate more on the scenario you think is happening that triggers the race?

jlowe:
Curious, what does `AtomicReference` bring to the table that `volatile` doesn't solve?  It's using `volatile` underneath, and we're not leveraging the lazy set/get functionality.

andygrove:
I didn't realize that AtomicReference still used volatile underneath. I thought it was more efficient as a result but if that is not the case then I agree there is no point. I will revert.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
```diff
@@ -50,17 +50,20 @@ class SerializeConcatHostBuffersDeserializeBatch(
...
   @transient private val headers = data.map(_.header)
   @transient private val buffers = data.map(_.buffer)
-  @transient private var batchInternal: ColumnarBatch = null
+  @transient private val batchInternalRef: AtomicReference[ColumnarBatch] =
+    new AtomicReference[ColumnarBatch]()
 
   def batch: ColumnarBatch = {
+    var batchInternal = batchInternalRef.get()
     if (batchInternal == null) {
       // TODO we should come up with a better way for this to happen directly...
```
jlowe, line 59: Do we want locking here to prevent multiple threads from wasting time creating duplicate objects?  I'm wondering if things get simpler if we just synchronize the relevant methods.

andygrove: That's a good idea. I will make that change.

jlowe:
Note that GpuCartesianProductExec has its own serializer which looks similar, not sure if we want a similar change there as well.

andygrove:
> @andygrove can you elaborate more on the scenario you think is happening that triggers the race?

I'm not sure I fully understand the access pattern yet, but my process here was:

- Make the issue very reproducible by running the broacast nested loop join test in a loop at the same time as running a heavy process to slow my desktop down. It consistently fails pretty fast (< 5 minutes)
- Added debug logging and saw broadcast batches with row count of zero instead of 25, sometimes
- I saw that we were creating the batch and making it available before setting the row count, and confirmed that changing that meant I could no longer make the test fail

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
```diff
@@ -50,21 +50,23 @@ class SerializeConcatHostBuffersDeserializeBatch(
...
       val barr = out.toByteArray
       val oin = new ObjectInputStream(new ByteArrayInputStream(barr))
-      readObject(oin)
+      batchInternalRef = readObject(oin)
     }
-    batchInternal
+    batchInternalRef
   }
 
+  def batchInternal = this.synchronized { batchInternalRef }
```
jlowe, line 68: There's no point in synchronizing access to a volatile member.  There's no harm in leaving `batchInternal` a `volatile` and just synchronize access to the `batch` method so two of them don't try to set it at the same time.  All other references are just reads.

jlowe (approved):

revans2 (approved):

andygrove:
build

PR #1166, opened 2020-11-19
Fix the cast tests for 3.1.0+

### LINKED ISSUE #1164: [BUG] ansi_cast tests are failing in 3.1.0
**Describe the bug**
When running cast tests against 3.1.0 we now get failures about not being able to cast between numbers and timestamps.  We probably just need to enable a flag to keep the tests running in the short term.

revans2:
This fixes #1164 

@andygrove please let me know if there is a better way to do this.


revans2:
build

andygrove (approved):

PR #1173, opened 2020-11-20
Support relational operators for decimal type

nartal1:
Support relational operators for decimal type. Added integration tests to verify corner cases.

### DIFF
--- a/integration_tests/src/main/python/cmp_test.py
+++ b/integration_tests/src/main/python/cmp_test.py
@@ -21,7 +21,7 @@ from spark_session import with_cpu_session
 from pyspark.sql.types import *
 import pyspark.sql.functions as f
 
-@pytest.mark.parametrize('data_gen', eq_gens, ids=idfn)
+@pytest.mark.parametrize('data_gen', eq_gens_with_decimal_gen, ids=idfn)
 def test_eq(data_gen):
     (s1, s2) = gen_scalars(data_gen, 2, force_no_nulls=True)
     data_type = data_gen.data_type
@@ -31,7 +31,7 @@ def test_eq(data_gen):
                 s2 == f.col('b'),
                 f.lit(None).cast(data_type) == f.col('a'),
                 f.col('b') == f.lit(None).cast(data_type),
-                f.col('a') == f.col('b')))
+                f.col('a') == f.col('b')), conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': 'true'})
 
 @pytest.mark.parametrize('data_gen', eq_gens, ids=idfn)
 def test_eq_ns(data_gen):
@@ -45,7 +45,7 @@ def test_eq_ns(data_gen):
                 f.col('b').eqNullSafe(f.lit(None).cast(data_type)),
                 f.col('a').eqNullSafe(f.col('b'))))
 
-@pytest.mark.parametrize('data_gen', eq_gens, ids=idfn)
+@pytest.mark.parametrize('data_gen', eq_gens_with_decimal_gen, ids=idfn)
 def test_ne(data_gen):
     (s1, s2) = gen_scalars(data_gen, 2, force_no_nulls=True)
     data_type = data_gen.data_type
@@ -55,11 +55,9 @@ def test_ne(data_gen):
                 s2 != f.col('b'),
                 f.lit(None).cast(data_type) != f.col('a'),
                 f.col('b') != f.lit(None).cast(data_type),
-                f.col('a') != f.col('b')))
+                f.col('a') != f.col('b')), conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': 'true'})
 
-@incompat
-@pytest.mark.allow_non_gpu('ProjectExec', 'Alias')
-@pytest.mark.parametrize('data_gen', orderable_gens_with_decimal_gen, ids=idfn)
+@pytest.mark.parametrize('data_gen', orderable_gens, ids=idfn)
 def test_lt(data_gen):
     (s1, s2) = gen_scalars(data_gen, 2, force_no_nulls=True)
     data_type = data_gen.data_type
@@ -69,7 +67,7 @@ def test_lt(data_gen):
                 s2 < f.col('b'),
                 f.lit(None).cast(data_type) < f.col('a'),
                 f.col('b') < f.lit(None).cast(data_type),
-                f.col('a') < f.col('b')))
+                f.col('a') < f.col('b')), conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': 'true'})
 
 @pytest.mark.parametrize('data_gen', orderable_gens, ids=idfn)
 def test_lte(data_gen):
@@ -81,7 +79,7 @@ def test_lte(data_gen):
                 s2 <= f.col('b'),
                 f.lit(None).cast(data_type) <= f.col('a'),
                 f.col('b') <= f.lit(None).cast(data_type),
-                f.col('a') <= f.col('b')))
+                f.col('a') <= f.col('b')), conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': 'true'})
 
 @pytest.mark.parametrize('data_gen', orderable_gens, ids=idfn)
 def test_gt(data_gen):
@@ -93,7 +91,7 @@ def test_gt(data_gen):
                 s2 > f.col('b'),
                 f.lit(None).cast(data_type) > f.col('a'),
                 f.col('b') > f.lit(None).cast(data_type),
-                f.col('a') > f.col('b')))
+                f.col('a') > f.col('b')), conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': 'true'})
 
 @pytest.mark.parametrize('data_gen', orderable_gens, ids=idfn)
 def test_gte(data_gen):
@@ -105,7 +103,7 @@ def test_gte(data_gen):
                 s2 >= f.col('b'),
                 f.lit(None).cast(data_type) >= f.col('a'),
                 f.col('b') >= f.lit(None).cast(data_type),
-                f.col('a') >= f.col('b')))
+                f.col('a') >= f.col('b')), conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': 'true'})
 
 @pytest.mark.parametrize('data_gen', eq_gens + array_gens_sample, ids=idfn)
 def test_isnull(data_gen):
--- a/integration_tests/src/main/python/cmp_test.py
+++ b/integration_tests/src/main/python/cmp_test.py
@@ -31,7 +31,7 @@ def test_eq(data_gen):
                 s2 == f.col('b'),
                 f.lit(None).cast(data_type) == f.col('a'),
                 f.col('b') == f.lit(None).cast(data_type),
-                f.col('a') == f.col('b')), conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': 'true'})
+                f.col('a') == f.col('b')), conf=allow_negative_scale_of_decimal_conf)
 
 @pytest.mark.parametrize('data_gen', eq_gens, ids=idfn)
 def test_eq_ns(data_gen):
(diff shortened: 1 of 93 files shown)


integration_tests/src/main/python/data_gen.py
```diff
@@ -208,6 +209,15 @@ def __init__(self, nullable=True, min_val = INT_MIN, max_val = INT_MAX,
...
         self._start(rand, lambda : rand.randint(self._min_val, self._max_val))
 
+
+class DecimalGen(DataGen):
+    """Generate Decimals, with some built in corner cases."""
+    def __init__(self, nullable=True):
+        super().__init__(DecimalType(7,3), nullable=nullable)
+
+    def start(self, rand):
+        self._start(rand, lambda : Decimal(str(round(random.uniform(2000, 4000), 2))))
```
revans2, line 219: I really only see one good choice when generating random data for Decimal.  We need to use `sre_yield` like with the StringGen.  We need to create a pattern that will work.  Essentially we need to have a pattern like

```
"-?[0-9]{1," + str(precision-scale) + "}.[0-9]{0," + str(scale) + "}"
```

Note that by default Spark does not support a negative scale (That is controlled by `park.sql.legacy.allowNegativeScaleOfDecimal`) but we do need to support it for testing.  So the above pattern only works for positive scale.  For negative scale we would need something more like.

```
"-?[0-9]{1,"+str(precision) + "}e" + str(-scale)
```

Note I have not tested these, this is just off the top of my head.

integration_tests/src/main/python/data_gen.py
```diff
@@ -685,6 +697,8 @@ def gen_scalars_for_sql(data_gen, count, seed=0, force_no_nulls=False):
 # a selection of generators that should be orderable (sortable and compareable)
 orderable_gens = [byte_gen, short_gen, int_gen, long_gen, float_gen, double_gen,
         string_gen, boolean_gen, date_gen, timestamp_gen]
+orderable_gens_with_decimal_gen = [byte_gen, short_gen, int_gen, long_gen, float_gen, double_gen,
```
revans2, line 700: nit:  It is a lot cleaner to do

```
orderable_gens_with_decimal_gen = orderable_gens + [decimal_gen]
```

Ideally I'd like to see a comment that `orderable_gens_with_decimal_gen` would go away once all we can add decimal types to `orderable_gens` directly.

integration_tests/src/main/python/data_gen.py
```diff
@@ -668,8 +678,10 @@ def gen_scalars_for_sql(data_gen, count, seed=0, force_no_nulls=False):
 boolean_gen = BooleanGen()
 date_gen = DateGen()
 timestamp_gen = TimestampGen()
+decimal_gen = DecimalGen()
```
revans2, line 681: We are going to need more than just one.  We really are going to need a few different sets of decimal_gens. A default set that work with `park.sql.legacy.allowNegativeScaleOfDecimal` disabled, and others that work when it is enabled.

integration_tests/src/main/python/cmp_test.py
```diff
@@ -57,7 +57,9 @@ def test_ne(data_gen):
                 f.col('b') != f.lit(None).cast(data_type),
                 f.col('a') != f.col('b')))
 
-@pytest.mark.parametrize('data_gen', orderable_gens, ids=idfn)
+@incompat
+@pytest.mark.allow_non_gpu('ProjectExec', 'Alias')
```
revans2, line 61: These should be changed now.  It is not incompat and Project and Alias are on the GPU in all cases.

nartal1: Done

integration_tests/src/main/python/data_gen.py
nartal1: Added for default case, positive and negative scale. Thanks!

integration_tests/src/main/python/data_gen.py
nartal1: Removed this as all support for all operators which take in `orderable_gens` is done. Have added for `eq_gens` as there is no decimal type support for `EqualNullSafe`

integration_tests/src/main/python/data_gen.py
```diff
@@ -208,6 +209,15 @@ def __init__(self, nullable=True, min_val = INT_MIN, max_val = INT_MAX,
     def start(self, rand):
         self._start(rand, lambda : rand.randint(self._min_val, self._max_val))
 
+
+class DecimalGen(DataGen):
+    """Generate Decimals, with some built in corner cases."""
+    def __init__(self, nullable=True):
```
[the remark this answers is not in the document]
nartal1: Your PR covered this.

integration_tests/src/main/python/data_gen.py
nartal1: Your PR covered this.

nartal1:
build

integration_tests/src/main/python/cmp_test.py
```diff
@@ -31,7 +31,7 @@ def test_eq(data_gen):
                 s2 == f.col('b'),
                 f.lit(None).cast(data_type) == f.col('a'),
                 f.col('b') == f.lit(None).cast(data_type),
-                f.col('a') == f.col('b')))
+                f.col('a') == f.col('b')), conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': 'true'})
```
revans2, line 34: If this is going to be common for a lot of decimal tests can we make it something in data_gen that can be used by multiple tests?

integration_tests/src/main/python/data_gen.py
```diff
@@ -708,13 +745,16 @@ def gen_scalars_for_sql(data_gen, count, seed=0, force_no_nulls=False):
 # TODO add in some array generators to this once that is supported for sorting
 # a selection of generators that should be orderable (sortable and compareable)
 orderable_gens = [byte_gen, short_gen, int_gen, long_gen, float_gen, double_gen,
-        string_gen, boolean_gen, date_gen, timestamp_gen, null_gen]
+        string_gen, boolean_gen, date_gen, timestamp_gen, null_gen] + decimal_gens
```
revans2, line 748: Do the sort tests work with this change?

integration_tests/src/main/python/data_gen.py
```diff
@@ -208,6 +209,37 @@ def __init__(self, nullable=True, min_val = INT_MIN, max_val = INT_MAX,
     def start(self, rand):
         self._start(rand, lambda : rand.randint(self._min_val, self._max_val))
 
+
+#Maximum number of decimal digits a Long can represent is 18.
+DECIMAL_MIN = -(1 << 59)
+DECIMAL_MAX = (1 << 59)  #approximate
```
revans2, line 215: I don't think these work the way we want them to because the maximum and minimum values are dependent on the precision and scale.  These values are invalid for some combinations of these.

I would really prefer to have us generate a string that represents the correct values, using something like the pattern below.

```
Decimal(('9'*precision) + 'e' + str(-scale)), Decimal('-' + ('9' * precision) + 'e' + str(-scale))
```

integration_tests/src/main/python/data_gen.py
```diff
@@ -208,6 +209,37 @@ def __init__(self, nullable=True, min_val = INT_MIN, max_val = INT_MAX,
...
+    def __init__(self, precision=None, scale=None, nullable=True, special_cases=[]):
+        if precision is None:
+            precision = 18
+            scale = 0
+            special_cases = [Decimal('-1'), Decimal('0'), Decimal('1'), Decimal(DECIMAL_MIN), Decimal(DECIMAL_MAX)]
+        super().__init__(DecimalType(precision, scale), nullable=nullable, special_cases=special_cases)
+        self._scale = scale
+        self._precision = precision
+        if (scale > 0):
+            pattern = "[0-9]{1,"+ str(precision - scale) + "}\.[0-9]{0," + str(scale) + "}"
```
revans2, line 227: I made a mistake here.  If scale > precision, then this pattern is not going to work.  I think we might just want to use the other pattern all of the time.

integration_tests/src/main/python/data_gen.py
```diff
@@ -208,6 +209,37 @@ def __init__(self, nullable=True, min_val = INT_MIN, max_val = INT_MAX,
...
+#Maximum number of decimal digits a Long can represent is 18.
+DECIMAL_MIN = -(1 << 59)
+DECIMAL_MAX = (1 << 59)  #approximate
+class DecimalGen(DataGen):
+    """Generate Decimals, with some built in corner cases."""
+    def __init__(self, precision=None, scale=None, nullable=True, special_cases=[]):
+        if precision is None:
+            precision = 18
+            scale = 0
+            special_cases = [Decimal('-1'), Decimal('0'), Decimal('1'), Decimal(DECIMAL_MIN), Decimal(DECIMAL_MAX)]
```
revans2, line 222: Even -1 and 1 do not work all of the time.  a precision of 1 and scale of -3 would be an example where this would not work. It can only hold multiples of 100 up to 900.

nartal1: Sorry, I don't understand above statement.  precision of 1 and scale of -3 is `1000`  and I am not getting why it can hold multiples of 100. Could you please explain. I tried setting precision and scale as mentioned above and it passed. So I am missing something.

integration_tests/src/main/python/data_gen.py
nartal1: done

integration_tests/src/main/python/data_gen.py
nartal1: done

integration_tests/src/main/python/data_gen.py
nartal1: It doesn't affect sort tests as they don't use `orderable_gens` instead use `orderable_gen_classes`

integration_tests/src/main/python/cmp_test.py
nartal1: Done

integration_tests/src/main/python/data_gen.py
nartal1: It doesn't affect sort tests as they don't use `orderable_gens` instead use `orderable_gen_classes`

integration_tests/src/main/python/data_gen.py
revans2:
```
>>> df = spark.createDataFrame(data=[(Decimal('-1')), (Decimal('0'))], schema=DecimalType(7, -3))
>>> df.show()
+-----+
|value|
+-----+
| 0E+3|
| 0E+3|
+-----+
```

The details of -2 vs -3 scale don't matter, what matters is -1 becomes 0 for any scale that is negative.  For a Decimal where the precision is larger than the scale this actually becomes an error'

```
>>> df = spark.createDataFrame(data=[(Decimal('-1')), (Decimal('0'))], schema=DecimalType(7, 7))
>>> df.show()
20/11/25 12:52:57 ERROR Executor: Exception in task 0.0 in stage 32.0 (TID 125)
java.lang.ArithmeticException: Decimal precision 8 exceeds max precision 7
	at org.apache.spark.sql.types.Decimal.set(Decimal.scala:122)
	at org.apache.spark.sql.types.Decimal$.apply(Decimal.scala:571)
```

We cannot blindly add in -1 as a special case.

nartal1: Thanks for the exmaples. I have removed `-1` and `1` from special cases. Added another generator where scale and precision are same.

integration_tests/src/main/python/data_gen.py
```diff
@@ -208,6 +209,33 @@ def __init__(self, nullable=True, min_val = INT_MIN, max_val = INT_MAX,
     def start(self, rand):
         self._start(rand, lambda : rand.randint(self._min_val, self._max_val))
 
+class DecimalGen(DataGen):
+    """Generate Decimals, with some built in corner cases."""
+    def __init__(self, precision=None, scale=None, nullable=True, special_cases=[]):
+        if precision is None:
+            #Maximum number of decimal digits a Long can represent is 18
+            precision = 18
```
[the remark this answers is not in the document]
nartal1: Agreed. Selecting correct precision and scale for each test would be good.

sperlingxx:
build

nartal1:
build

nartal1:
build

nartal1:
@revans2 I think I have addressed remaining review comments. Please take another look.

revans2 (approved):

PR #1174, opened 2020-11-20
Fix race condition in SerializeConcatHostBuffersDeserializeBatch

andygrove:

This PR fixes a race condition where a batch was created with a zero row count and then updated with the actual row count, but other threads were able to use this batch when it was in the intermediate state.

This is a smaller change than the PR yesterday that attempted to fix this, but also introduced a regression.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
@@ -90,8 +90,7 @@ class SerializeConcatHostBuffersDeserializeBatch(
         val table = tableInfo.getTable
         if (table == null) {
           val numRows = tableInfo.getNumRows
-          this.batchInternal = new ColumnarBatch(new Array[ColumnVector](0))
-          batchInternal.setNumRows(numRows.toInt)
+          this.batchInternal = new ColumnarBatch(new Array[ColumnVector](0), numRows.toInt)
         } else {
           val colDataTypes = in.readObject().asInstanceOf[Array[DataType]]
           // This is read as part of the broadcast join so we expect it to leak.


andygrove:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
```diff
@@ -90,8 +90,7 @@ class SerializeConcatHostBuffersDeserializeBatch(
         val table = tableInfo.getTable
         if (table == null) {
           val numRows = tableInfo.getNumRows
-          this.batchInternal = new ColumnarBatch(new Array[ColumnVector](0))
-          batchInternal.setNumRows(numRows.toInt)
+          this.batchInternal = new ColumnarBatch(new Array[ColumnVector](0), numRows.toInt)
```
abellina, line 93: seems like if there's a race to this line in the code we could have a race to `batchInternal` being constructed twice?

andygrove: The specific issue here was that we created the object with a zero row count, and then set the row count, so for a whlie the object had bad data and other threads were using it in that state.

Yes, we still could potentially be creating the object multiple times which is performance issue but is safe functionally. This is what we tried to address in the original PR with some sychronization but it caused a regression somehow.

jlowe: In this specific case it's a degenerate columnar batch so it's not like we're leaking off heap resources.  However if we're racing through this code and it's _not_ a degenerate batch then we could leak the GPU resources.  I think we need some sort of "I lost the race" logic at least minimally in the `else` code below.  I also think we need to understand more what threads and streams are involved.  Did we regress because two threads were trying to pull from the same input stream?

andygrove: This is the stack trace from the failed tests and it certainly seems like it could have been caused by two threads pulling from the same input stream.

```
20:30:18  Caused by: java.io.StreamCorruptedException: unexpected block data
20:30:18  	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1663)
20:30:18  	at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2365)
20:30:18  	at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2289)
20:30:18  	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2147)
20:30:18  	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1646)
20:30:18  	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:482)
20:30:18  	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:440)
20:30:18  	at org.apache.spark.serializer.JavaDeserializationStream.readObject(JavaSerializer.scala:76)
```

abellina: I do not know enough about the access pattern though. I guess if access to `batchInternal` could happen from other threads, should we not synchronize these methods? For example, `numRows` will return different things depending on whether `batchInternal` is null or not.

andygrove: It looks like `numRows` has a fast path for when the batch doesn't need to be materialized? I'm also not familiar with the access patterns here though.

jlowe: Sounds like we need more investigation regarding the scenario when multiple threads are passing through here.  Leaking GPU resources is annoying, but producing the wrong row count is far worse.  Any objections to putting this in now and filing a followup issue?  Or do we think we can resolve the race to create a GPU-based columnar batch quickly?

revans2: +1

abellina (approved):

jlowe (approved):

PR #1176, opened 2020-11-20
Add in support for null type

revans2:
This adds in null type support for almost all operators. The only operators that are missing is python UDF support and window operations.  I didn't consider these critical right now.  Window support is likely to work, but there have been issues with nulls in the past.  I am not sure what the conversion to arrow would look like so I avoided it in the first take on this.

### DIFF
--- a/shims/spark301db/src/main/scala/com/nvidia/spark/rapids/shims/spark301db/GpuShuffledHashJoinExec.scala
+++ b/shims/spark301db/src/main/scala/com/nvidia/spark/rapids/shims/spark301db/GpuShuffledHashJoinExec.scala
@@ -18,8 +18,8 @@ package com.nvidia.spark.rapids.shims.spark301db
 
 import com.nvidia.spark.rapids._
 import com.nvidia.spark.rapids.GpuMetricNames._
-
 import org.apache.spark.TaskContext
+
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.Expression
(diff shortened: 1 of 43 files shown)


revans2:
build

shims/spark301db/src/main/scala/com/nvidia/spark/rapids/shims/spark301db/GpuShuffledHashJoinExec.scala
```diff
@@ -18,8 +18,8 @@ package com.nvidia.spark.rapids.shims.spark301db
 
 import com.nvidia.spark.rapids._
 import com.nvidia.spark.rapids.GpuMetricNames._
-
 import org.apache.spark.TaskContext
```
jlowe, line 21: Style nit: Not sure why this import was separated out from the other `org.apache.spark` imports.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -990,18 +995,30 @@ object GpuOverrides {
     expr[Coalesce] (
       "Returns the first non-null argument if exists. Otherwise, null",
       (a, conf, p, r) => new ExprMeta[Coalesce](a, conf, p, r) {
+        override def isSupportedType(t: DataType): Boolean =
+          GpuOverrides.isSupportedType(t,
+            allowNull = true)
```
jlowe, line 1000: This is very common in the expressions.  Worth creating an `ExprMetaHandlesNull` to cut down on the boilerplate?  I'm not stuck on it.

revans2: Decimal is coming which will not overlap 100% with these.  I thought the main point of GpuOverrides.isSupportedType was to let us do this an item at a time.  Once things calm down and we have tested everything fully, then we would drop isNull as being configurable from isSupportedType.

revans2:
build

jlowe (approved):

PR #1180, opened 2020-11-20
Remove batches if they are received after the iterator detects that t…

### LINKED ISSUE #1177: [BUG] leaks possible in the rapids shuffle if batches are received after the task completes
See comment below.

Old explanation:

> When there is a limit query, tasks complete early and blocks that are being requested in flight could be leaked.

Update:

In looking through my logs I only see cases where we detected this in failure conditions, where the task was closed forcefully. I now think Spark is doing the right thing, even in limit cases, and draining iterators. 

---

The short term solution is to make the `RapidsShuffleIterator` reject batches as they show up, which would make the `RapidsShuffleClient` close them on receipt. This can easily be done, but it's wasteful because we are going to be waiting for IO and holding up resources on the peer also when there is no compute to be done for these batches.

A longer term solution will be tracked separately.

abellina:
Closes #1177

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleClient.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleClient.scala
@@ -43,8 +43,10 @@ trait RapidsShuffleFetchHandler {
   /**
    * Called when a buffer is received and has been handed off to the catalog.
    * @param bufferId - a tracked shuffle buffer id
+   * @return a boolean that lets the caller know the batch was accepted (true), or
+   *         rejected (false), in which case the caller should dispose of the batch.
    */
-  def batchReceived(bufferId: ShuffleReceivedBufferId): Unit
+  def batchReceived(bufferId: ShuffleReceivedBufferId): Boolean
 
   /**
    * Called when the transport layer is not able to handle a fetch error for metadata
@@ -450,11 +452,20 @@ class RapidsShuffleClient(
 
           val stats = tx.getStats
 
+          val toDelete = new ArrayBuffer[ShuffleReceivedBufferId]()
+
           // hand buffer off to the catalog
           buffMetas.foreach { consumed: ConsumedBatchFromBounceBuffer =>
             val bId = track(consumed.contigBuffer, consumed.meta)
-            consumed.handler.batchReceived(bId.asInstanceOf[ShuffleReceivedBufferId])
+            if (!consumed.handler.batchReceived(bId)) {
+              toDelete.append(bId)
+            }
             transport.doneBytesInFlight(consumed.contigBuffer.getLength)
+          }
+
+          if (toDelete.nonEmpty) {
+            logDebug(s"Received ${toDelete.size} batches after task completed. Freeing.")
+            toDelete.foreach(id => catalog.removeBuffer(id))
           }
 
           logDebug(s"Received buffer size ${stats.receiveSize} in" +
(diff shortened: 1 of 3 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleClient.scala
```diff
@@ -450,13 +452,22 @@ class RapidsShuffleClient(
...
           val stats = tx.getStats
 
+          val toDelete = new ArrayBuffer[ShuffleReceivedBufferId]()
+
           // hand buffer off to the catalog
           buffMetas.foreach { consumed: ConsumedBatchFromBounceBuffer =>
             val bId = track(consumed.contigBuffer, consumed.meta)
-            consumed.handler.batchReceived(bId.asInstanceOf[ShuffleReceivedBufferId])
+            if (!consumed.handler.batchReceived(bId)) {
+              toDelete.append(bId)
```
jlowe, line 461: Why collect them instead of deleting them as they are handled?  Seems like we can ditch the `ArrayBuffer` completely and track the number of times we threw away in a local variable, performing a debug log if that variable is non-zero at the end.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleIterator.scala
```diff
@@ -198,46 +202,52 @@ class RapidsShuffleIterator(
...
-                    s"total batches resolved $clientResolvedBatches.")
-                } else {
-                  logDebug(s"Task: $taskAttemptId Client $blockManagerId is " +
-                    s"NOT done fetching batches. Total batches expected $clientExpectedBatches, " +
-                    s"total batches resolved $clientResolvedBatches.")
+              if (taskComplete) {
+                false
+              } else {
+                batchesInFlight = batchesInFlight - 1
+                val nvtxRange = new NvtxRange(s"BATCH RECEIVED", NvtxColor.DARK_GREEN)
```
jlowe, line 228: use `withResource` here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleIterator.scala
```diff
@@ -198,46 +202,52 @@ class RapidsShuffleIterator(
...
+                try {
+                  if (markedAsDone) {
```
jlowe, line 230: Is `markedAsDone` redundant with `taskComplete`?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleClient.scala
abellina: will do, that's better!

abellina (commented):
I made the suggested changes, but I am having some environment issues to reproduce this in the wild. Leaving this in Draft mode until I can double check everything still works.

jlowe (approved):

abellina:
@jlowe Note that I no longer believe this is a limit issue. I updated https://github.com/NVIDIA/spark-rapids/issues/1177. Sorry for the confusion. This change triggers in error conditions for a fact, but I am now doubting whether it is useful in other conditions.

abellina:
Updated https://github.com/NVIDIA/spark-rapids/issues/1177 again. Was able to reproduce, main thing that tricked me was the coalesce batches. Not sure if that is expected, where you have a shuffle -> coalesce -> local limit X. Should coalesce only coalesce to X? Either way, it's definitely seen if coalesce is adjusted to produce small batches, such that local limit finds what it needs early.

abellina:
build

jlowe:
build

PR #1182, opened 2020-11-20
Close buffer catalog on device manager shutdown

jlowe:
This fixes a leak seen during unit tests because the catalog stores are not being closed when the device manager is shut down.


jlowe:
build

pxLi:
build

abellina (approved):

Issue #463, opened 2020-07-29
[FEA] Support NESTED_SCHEMA_PRUNING_ENABLED for parquet and ORC


revans2:
Once we support reading structs from Parquet and/or ORC we should look at also supporting SQLConf.NESTED_SCHEMA_PRUNING_ENABLED which is on by default.

> Prune nested fields from a logical relation's output which are unnecessary in
> satisfying a query. This optimization allows columnar file format readers to avoid
> reading unnecessary nested column data. Currently Parquet and ORC are the
> data sources that implement this optimization.

From what I am reading we probably can make this work totally in the plugin by deciding what to push down to cudf to parse and properly updating the metadata.  If that is not true then we will need to work with cudf on this.

This is related to #425 and #234

jlowe:
Marking this P1 because without basic support for this, libcudf could load more columns than Spark is expecting.  At that point the schemas don't match and we could end up misinterpreting the input data and corrupting the query output.

revans2:
Turns out for Parquet we get this for free.  I will be putting up a PR with just tests in it, but for now I am going to drop Parquet from this feature request.  It looks like `ParquetReadSupport.clipParquetSchema` is doing this for us already.

revans2:
I put this back into NeedsTriage so we can re-evaluate it in terms of just ORC support.

firestarman:
This depends on https://github.com/rapidsai/cudf/issues/8910

GaryShen2008:
Retarget this feature to 21.10.

PR #1188, opened 2020-11-23
Add in tests for parquet nested pruning support

revans2:
This is related to #463 It just adds in tests to show that we are doing the right thing for pruning/read schema.

Note that for these tests I explicitly set the read schema. If we want to have it be implicit through the query I can do that too.


revans2:
build

jlowe (approved):

revans2:
build

revans2:
Jenkins got confused some how so I canceled that build and started it over again

Issue #971, opened 2020-10-16
[FEA] Add documentation for metrics added by the plugin


jlowe:
**Is your feature request related to a problem? Please describe.**
#952 indicates there is some confusion on the meaning behind some of the SQL metrics added by the plugin.

**Describe the solution you'd like**
There should be documentation somewhere in the docs/ directory that lists the new metrics added by the plugin to various SQL nodes and what is measured by those metrics.

jlowe:
Fixed by #3681.

PR #1194, opened 2020-11-24
Drop redundant total time metric from scan

### LINKED ISSUE #972: [BUG] total time metric is redundant with scan time
**Describe the bug**
As indicated in #952 the `total time` metric for GPU scan nodes is not very useful since there's already a Spark standard `scan time` metric.  More confusingly, `total time` is almost always going to be slightly smaller than `scan time` implying it isn't really a total.

Run a query that uses `GpuFIleSourceScanExec` and examine the metrics on the SQL UI.

**Expected behavior**
We probably should just remove the total time from the node.  If we keep both for some reason, total time should not be less than scan time as it is today.

revans2:
This fixes #972 

I am happy to add in documentation for the metrics too if that is what we want #971

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBatchScanExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBatchScanExec.scala
@@ -22,7 +22,7 @@ import scala.collection.JavaConverters._
 import scala.math.max
 
 import ai.rapids.cudf
-import ai.rapids.cudf.{HostMemoryBuffer, NvtxColor, Table}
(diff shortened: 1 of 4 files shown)


revans2:
build

tgravescs (approved):

tgravescs:
it would be nice to have documentation for the metrics but I would think that would be across the board not just for scan so a separate issue.

jlowe (approved):

jlowe:
build

revans2:
build

revans2:
build

revans2:
build

revans2:
build

revans2:
build

sameerz:
build

PR #1196, opened 2020-11-24
Rename ShuffleCoalesceExec to GpuShuffleCoalesceExec

jlowe:
@tgravescs pointed out that `ShuffleCoalesceExec` can appear to be something that isn't running on the GPU.  While this is technically true since it is coalescing on the host before putting the results on the GPU, it can be confusing to users who think there is something than could be on the GPU but currently isn't.  Since `ShuffleCoalesceExec` is all about coalescing shuffles intended for the GPU, this renames ShuffleCoalesceExec to GpuShuffleCoalesceExec to reduce that confusion.


tgravescs:
build

tgravescs (approved):

abellina (approved):

jlowe:
build

Issue #1184, opened 2020-11-23
[BUG] test_window_aggregate_udf_array_from_python fails on databricks 3.0.1


tgravescs:
**Describe the bug**
test_window_aggregate_udf_array_from_python test is failing on databricks.

11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[No_Partition-Byte][IGNORE_ORDER]
11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[No_Partition-Short][IGNORE_ORDER]
11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[No_Partition-Integer][IGNORE_ORDER]
11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded-Byte][IGNORE_ORDER]
11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded-Short][IGNORE_ORDER]
11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded-Integer][IGNORE_ORDER]
11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded_Following-Byte][IGNORE_ORDER]
11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded_Following-Short][IGNORE_ORDER]
11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded_Following-Integer][IGNORE_ORDER]
11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded_Preceding-Byte][IGNORE_ORDER]
11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded_Preceding-Short][IGNORE_ORDER]
11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded_Preceding-Integer][IGNORE_ORDER]
11:37:59  FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Lower_Upper-Byte][IGNORE_ORDER]
[...]

PR #1201, opened 2020-11-25
Replace only window expressions on databricks.

firestarman:
On Databricks, when pulling only one element from an array type column, Databricks will put the
`GetArrayItem` into the `GpuWindowInPandas`, instead of generating a new `ProjectExec`.

This PR is to fix the issue https://github.com/NVIDIA/spark-rapids/issues/1184 by replacing only window expressions instead of the whole expression tree to avoid removing the original parent expression unexpectedly,  e.g. `GetArrayItem` or `Cast ` or others.  


firestarman:
build

GaryShen2008 (approved):

PR #1202, opened 2020-11-25
Fix a bug with the support for java.lang.StringBuilder.append.

seanprime7:
As java.lang.StringBuilder.append is a mutable operation, when this
method is called with a string builder object, all the references of
this object needs to be updated in locals and stack.

### DIFF
--- a/udf-compiler/src/test/scala/com/nvidia/spark/OpcodeSuite.scala
+++ b/udf-compiler/src/test/scala/com/nvidia/spark/OpcodeSuite.scala
@@ -1432,6 +1432,29 @@ class OpcodeSuite extends FunSuite {
   }
 
   // Tests for string ops
+  test("java lang string builder test - append") {
+    // We do not support java.lang.StringBuilder officially in the udf compiler,
+    // but string + string in Scala generates some code with
+    // java.lang.StringBuilder. For that reason, we have some tests with
+    // java.lang.StringBuilder.
+    val myudf: (String, String, Boolean) => String = (a,b,c) => {
+      val sb = new java.lang.StringBuilder()
+      if (c) {
+        sb.append(a)
+        sb.toString + " true"
+      } else {
+        sb.append(b)
+        sb.toString + " false"
+      }
+    }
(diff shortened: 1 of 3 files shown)


udf-compiler/src/test/scala/com/nvidia/spark/OpcodeSuite.scala
```diff
@@ -1432,6 +1432,29 @@ class OpcodeSuite extends FunSuite {
...
+    val myudf: (String, String, Boolean) => String = (a,b,c) => {
+      val sb = new java.lang.StringBuilder()
+      if (c) {
+        sb.append(a)
+        sb.toString + " true"
+      } else {
+        sb.append(b)
+        sb.toString + " false"
+      }
+    }
```
abellina, line 1449: could we add a test that goes through the case where we append multiple times?

seanprime7: Updated the test case to have multiple appends.

abellina:
build

seanprime7:
@abellina I have added more `+`s to the test case as well.
I think you need to start the build again.

abellina:
build

jlowe:
CI failed due to timeout. Note that this will need to upmerge to latest on branch-0.3 to avoid an issue testing with the latest Spark 3.1.0-SNAPSHOT.

seanprime7:
> CI failed due to timeout. Note that this will need to upmerge to latest on branch-0.3 to avoid an issue testing with the latest Spark 3.1.0-SNAPSHOT.

It seems that the tests ran successfully but it timed out after that.

seanprime7:
Rebased the PR to the latest on branch-0.3

abellina:
build

wjxiz1992:
Build failed due to String cast to Timestamp test case. Will build when it's resolved

wjxiz1992:
build

wjxiz1992 (approved):

PR #1207, opened 2020-11-25
Spark SQL hash function using murmur3

rwlee:


jlowe:
@rwlee @sameerz Are we targeting this for 0.3 or 0.4?

jlowe:
> DO NOT MERGE

I'm converting this to a Draft PR so it's clear this is not ready to merge.

rwlee:
Ready for review, requires https://github.com/rapidsai/cudf/pull/7024

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1841,6 +1841,17 @@ object GpuOverrides {
         override def tagExprForGpu(): Unit = {}
         override def convertToGpu(child: Seq[Expression]): GpuExpression = GpuConcat(child)
       }),
+    expr[Murmur3Hash] (
+      "MD5 hash operator",
+      (a, conf, p, r) => new ExprMeta[Murmur3Hash](a, conf, p, r) {
```
revans2, line 1846: This need to be upmerged.  Type checking has changed.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/HashFunctions.scala
```diff
@@ -34,3 +39,44 @@ case class GpuMd5(child: Expression)
...
+
+case class GpuMurmur3Hash(child: Seq[Expression]) extends GpuExpression {
+  override def dataType: DataType = IntegerType
+
+  override def toString: String = s"hash($child)"
+  def nullable: Boolean = children.exists(_.nullable)
+  def children: Seq[Expression] = child
+
+  def columnarEval(batch: ColumnarBatch): Any = {
+    var nullStrScalar: Scalar = null
```
revans2, line 51: This is not used

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/HashFunctions.scala
```diff
@@ -34,3 +39,44 @@ case class GpuMd5(child: Expression)
...
+    val rows = batch.numRows()
+    val childEvals: ArrayBuffer[Any] = new ArrayBuffer[Any](children.length)
```
revans2, line 53: `childEvals` is populated before `columns` is, but `columns` is the only one that is closed at the end. If we got an exception while processing we would leak anything in `childEvals` that is not in `columns` yet.

I think it would be cleaner to just not have childEvals and loop through children instead so we avid the problem entirely.

```
children.foreach { expr =>
    expr.columnarEval(batch) match { ...
    }
}
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/HashFunctions.scala
```diff
@@ -34,3 +39,44 @@ case class GpuMd5(child: Expression)
...
+    val columns: ArrayBuffer[ColumnVector] = new ArrayBuffer[ColumnVector]()
+    try {
+      children.foreach(childEvals += _.columnarEval(batch))
+      childEvals.foreach {
+        case vector: GpuColumnVector =>
+          columns += vector.getBase
+        case col => if (col != null) {
+          withResource(GpuScalar.from(col)) {
+            scalarValue => col match {
```
revans2, line 62: Why not just use the CUDF APIs directly?

```
columns += ColumnVector.fromScalar(scalarValue, rows)
```

revans2:
Marked it as draft again.  Even if it is ready to review we cannot run this or check it in until the dependency is in.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/HashFunctions.scala
rwlee: When expanding from a scalar, the function requires the `DataType` be supplied. This seems to be the easiest way to extract the type, is there an easier way to extract the type from the `GpuScalar` or the `columnarEval`.

rwlee:
> Marked it as draft again. Even if it is ready to review we cannot run this or check it in until the dependency is in.

https://github.com/rapidsai/cudf/pull/7024 merged in this afternoon.

rwlee:
build

rwlee:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/HashFunctions.scala
```diff
@@ -34,3 +39,33 @@ case class GpuMd5(child: Expression)
...
+case class GpuMurmur3Hash(child: Seq[Expression]) extends GpuExpression {
+  override def dataType: DataType = IntegerType
+
+  override def toString: String = s"hash($child)"
+  def nullable: Boolean = children.exists(_.nullable)
+  def children: Seq[Expression] = child
+
+  def columnarEval(batch: ColumnarBatch): Any = {
+    val rows = batch.numRows()
+    val childEvals: ArrayBuffer[Any] = new ArrayBuffer[Any](children.length)
```
revans2, line 52: nit: This is not used any more.

revans2 (approved):
This looks great, just one minor nit that I would prefer to have fixed, but I am OK with not fixing it too.

rwlee:
build

revans2:
build

revans2 (approved):

PR #1208, opened 2020-11-26
[BUG] Fix GHSL2020-239 [skip ci]

pxLi:

fix for GHSL2020-239 which found that authorized user could add bash injection into our workflow w/ specific branch name.
The vulnerable step was required by blossom team, so we just keep it.

Tested on my forked repo.


pxLi:
build

GaryShen2008 (approved):

PR #1213, opened 2020-11-30
Skip casting StringType to TimestampType for Spark 310

razajafri:
This PR fixes the build issue because of a change in Spark 310 with how they are handling ANSI cast

There also seemed to be a typo in a previous check-in, the tests are passing regardless of the order of the arguments, but I have fixed the order as I thought it should be. PTAL @revans2


pxLi:
build

pxLi (approved):

GaryShen2008 (approved):

PR #1217, opened 2020-11-30
Aggregate reductions in Complete mode should use updateExpressions

### LINKED ISSUE #898: [BUG] count reductions are failing on databricks because lack for Complete support
**Describe the bug**
In databricks the tests for `test_generic_reductions` are failing.  I traced this down to databricks using the `Complete` aggregation type, which our code does not properly support. 

Run `test_generic_reductions` on databricks with the latest branch-0.3 code.

**Expected behavior**
Test should pass

This looks like it should be somewhat simple to fix in a stop gap, but I am not 100% sure that we are doing the right thing in all cases with this. Ideally we check all possible aggregation types instead of just checking for a single one, but that may be difficult.

The reason this works for `count(*)` is because it is turned into `count(1)` and we do a SUM reduction on it, so it comes out the same.

kuhushukla:
Fixes #898 
Signed-off-by: Kuhu Shukla <kuhus@nvidia.com>

This fixes the reduction aggregates to work in complete mode. We already had some of this covered for when we have grouping keys. This change merely extends that change to reductions.

### DIFF
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -371,7 +371,6 @@ non_nan_all_basic_gens = [byte_gen, short_gen, int_gen, long_gen,
         string_gen, boolean_gen, date_gen, timestamp_gen]
 
 
-@pytest.mark.xfail(is_databricks_runtime(), reason='https://github.com/NVIDIA/spark-rapids/issues/898')
 @pytest.mark.parametrize('data_gen', non_nan_all_basic_gens, ids=idfn)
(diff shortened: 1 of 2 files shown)


kuhushukla:
build

kuhushukla:
I additionally ran some hash agg integration tests on DB cluster and they pass after this change.

abellina (approved):

tgravescs (approved):

tgravescs:
test fail due to issue in 3.1.0 with HiveStringType

kuhushukla:
build

PR #1220, opened 2020-11-30
Support replacing ORC format when Hive is configured

### LINKED ISSUE #1216: [BUG] Failure to recognize ORC file format when loaded via Hive
**Describe the bug**
When loading an ORC file from a Hive store the plugin fails to recognize it as a supported file format.  Enabling the explain config results in a message like this:
```
!Exec <FileSourceScanExec> cannot run on GPU because unsupported file format: org.apache.spark.sql.hive.orc.OrcFileFormat
```

Set `spark.sql.orc.impl=hive` and try to load an ORC file.

**Expected behavior**
ORC files should be supported even if the Hive implementation was configured.

jlowe:
Fixes #1216.

Adds checks for Spark's Hive implementation class for `OrcFileFormat` which allows replacing ORC writes with GPU-accelerated versions even when `spark.sql.orc.impl=hive`.


tgravescs (approved):

jlowe:
build

jlowe:
build

PR #1232, opened 2020-12-02
Move SortOrder creation to shims

jlowe:
Apache Spark 3.1 changed `SortOrder` in https://issues.apache.org/jira/browse/SPARK-33503 so that the `sameOrderExpressions` field is no longer a `Set` but is now a `Seq`.  This requires moving `SortOrder` creation to the shims, otherwise when running against Spark 3.1 the plugin will try to call a `SortOrder` constructor that does not exist.

### DIFF
--- a/shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
+++ b/shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
@@ -415,4 +415,17 @@ class Spark300Shims extends SparkShims {
       resolver: Resolver): Unit = {
     GpuSchemaUtils.checkColumnNameDuplication(schema, colType, resolver)
   }
+
+  override def sortOrderChildren(s: SortOrder): Seq[Expression] = {
+    (s.sameOrderExpressions + s.child).toSeq
+  }
+
+  override def sortOrder(
+      child: Expression,
+      direction: SortDirection,
+      nullOrdering: NullOrdering): SortOrder = SortOrder(child, direction, nullOrdering, Set.empty)
+
+  override def copySortOrderWithNewChild(s: SortOrder, child: Expression): SortOrder = {
+    s.copy(child = child)
+  }
 }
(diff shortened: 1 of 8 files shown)


jlowe:
build

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
```diff
@@ -298,4 +298,15 @@ class Spark310Shims extends Spark301Shims {
     val shuffleOrigin = cpuShuffle.map(_.shuffleOrigin).getOrElse(ENSURE_REQUIREMENTS)
     GpuShuffleExchangeExec(outputPartitioning, child, shuffleOrigin)
   }
+
+  override def sortOrderChildren(s: SortOrder): Seq[Expression] = s.children
+
+  override def sortOrder(
```
mythrocks, line 304: I think I've missed why the three argument `sortOrder()` method isn't in the `SparkShims` base class.

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -415,4 +415,17 @@ class Spark300Shims extends SparkShims {
...
+  override def sortOrderChildren(s: SortOrder): Seq[Expression] = {
+    (s.sameOrderExpressions + s.child).toSeq
+  }
+
+  override def sortOrder(
+      child: Expression,
+      direction: SortDirection,
+      nullOrdering: NullOrdering): SortOrder = SortOrder(child, direction, nullOrdering, Set.empty)
+
+  override def copySortOrderWithNewChild(s: SortOrder, child: Expression): SortOrder = {
```
mythrocks, line 428: I might have missed how this implementation isn't identical to that in `Spark310Shims`. If it is indeed identical, why did this copy functionality need to be shimmed?

jlowe: It's identical source code but compiles differently.  Source vs. binary compatibility.

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
jlowe: That's the method that each specific shim instance needs to override.  The two-argument version is just syntactic sugar to call the three argument version which must be implemented in the shims.  If the three argument version were to be placed in the SparkShims class then we wouldn't need to shim it at all.  :smile:

mythrocks (approved):

tgravescs (approved):

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
mythrocks: I was referring to the fact that the code seemed identical across the derived Shims classes. But I see that your reply to my other comment applies here as well.

PR #1237, opened 2020-12-02
In transport, limit pending transfer requests to fit within a bounce

abellina:
buffer

There are many better ways of handling the transport loop, but I have identified that this change is better, performance wise in some scenarios.

For the good parts, the change makes it so we don't over-pack a request. We may have  more bytes for a bounce-buffer backed request to handle, but having many lengths can be bad. Without this change, we would not send good receive requests that could go in separate bounce buffers (and simultaneously).

The bad parts are the fact that the code "puts back" requests into the hash priority queue (this causes `siftUp` to be called, which is `log(N)`). There are better ways of handling this problem, the best one we can think is having different queues, one per client, still sorted by request length. These changes would be too big to put in branch-0.3.


abellina:
build

abellina:
build

jlowe (approved):
Please file a followup issue to track cleaning up the requests handling.  It probably doesn't make sense to use a priority queue, or at least tracking all requests for all servers in one queue.

abellina:
Thanks @jlowe https://github.com/NVIDIA/spark-rapids/issues/1239

PR #1240, opened 2020-12-02
Fix loading from ORC file with no column names

### LINKED ISSUE #961: [BUG] ORC predicate pushdown should work with case-insensitive analysis
commit 6c88d7c1259ea9fe89f5c8190c683bba506d528e
Author: Liang-Chi Hsieh <viirya@gmail.com>
Date:   Tue Aug 25 04:42:39 2020 +0000

    [SPARK-32646][SQL][3.0][TEST-HADOOP2.7][TEST-HIVE1.2] ORC predicate pushdown should work with case-insensitive analysis

### LINKED ISSUE #962: [BUG] Loading columns from an ORC file without column names returns no data
commit b745041f698120be21ab889706880e976a599fdb
Author: SaurabhChawla <saurabhc@qubole.com>
Date:   Thu Jul 16 13:11:47 2020 +0000

    [SPARK-32234][SQL] Spark sql commands are failing on selecting the orc tables

jlowe:
Fixes #961
Fixes #962

Updates the GPU ORC scan to handle ORC files with no column names.  When no column names are detected, it generates a mapping of ORC file schema fields to the expected Spark read schema.  This mapping, if present, is applied during stripe building so we should still only load the columns referenced by Spark rather than loading all and filtering afterwards.

Because we need to check for the column names being missing relatively early, this refactors the interface between the partition reader factory and partition reader to have the factory build the ORC reader and pass it rather than have the partition reader build it.

### DIFF
--- a/integration_tests/src/main/python/orc_test.py
+++ b/integration_tests/src/main/python/orc_test.py
@@ -19,7 +19,7 @@ from datetime import date, datetime, timezone
 from data_gen import *
 from marks import *
 from pyspark.sql.types import *
-from spark_session import with_cpu_session, with_spark_session, is_before_spark_310
+from spark_session import with_cpu_session, with_spark_session, is_before_spark_310, is_spark_300
 
 def read_orc_df(data_path):
     return lambda spark : spark.read.orc(data_path)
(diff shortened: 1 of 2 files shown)


jlowe:
build

sameerz:
build

tgravescs:
do we have tests that cover spark.sql.caseSensitive ?

jlowe:
> do we have tests that cover spark.sql.caseSensitive ?

Apparently not, although I think that should be handled in a followup issue.  #961 actually has nothing to do with case sensitive analysis as pointed out in https://github.com/NVIDIA/spark-rapids/issues/961#issuecomment-725627980.

Adding tests for caseSensitive would need to cover Parquet as well as the setting is not specific to ORC.

jlowe:
Integration test is failing because Apache Spark 3.0.0 is crashing on the ORC file it generated without the plugin.  I'll need to exclude this test from running on Apache Spark 3.0.0.  It passes for me locally on Spark 3.0.1.

jlowe:
build

tgravescs (approved):

jlowe:
I filed #1255 to track adding case sensitive integration tests for Parquet and ORC.

Issue #37, opened 2020-05-28
[FEA] round and bround SQL functions


revans2:
**Is your feature request related to a problem? Please describe.**
It would be great to support the round and bround SQL functions.
https://github.com/rapidsai/cudf/issues/3790 was filed to implement them in cudf.

tgravescs:
It looks like the cudf side has been implemented by https://github.com/rapidsai/cudf/pull/6685

nartal1:
Yes, I was planning to start on cudf Java work on this once I am done with binary operations for decimal type.

PR #1244, opened 2020-12-02
Support round and bround SQL functions 

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/37


nartal1:
There is a bug in this which I am still figuring out. All test cases do not pass. 
If the scale is 0, the value is incremented by 1(which shouldn't happen). 
If I have something like below:

```
df1 = spark.createDataFrame(data=[(Decimal('3977631')), (Decimal('8291028')), (Decimal('8291025'))], schema=DecimalType(7,0))
ret = df2.selectExpr('round(value, 0)')

CPU result: (Decimal('3977631')), (Decimal('8291028')), (Decimal('8291025'))
GPU result: (Decimal('3977632')), (Decimal('8291029')), (Decimal('8291026'))
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
```diff
@@ -148,6 +149,35 @@ trait CudfUnaryExpression extends GpuUnaryExpression {
...
+          withResource(GpuScalar.from(r, right.dataType)) { scalar =>
+            GpuColumnVector.from(doColumnar(l, scalar), dataType)
+          }
+        case _ => null
+      }
+    } finally {
+      if (lhs.isInstanceOf[AutoCloseable]) {
+        lhs.asInstanceOf[AutoCloseable].close()
+      }
+      if (rhs.isInstanceOf[AutoCloseable]) {
```
andygrove, line 174: `rhs` could be null at this point so needs a null check here. Perhaps it would be better to introduce something similar to the `withResource` method that can work with `Any`'s that are possibly also `AutoCloseable`?

revans2: I added on https://github.com/NVIDIA/spark-rapids/blob/cce1ec1ed401bcd7465f5bef9ab28ed38aa5a9a8/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Arm.scala#L52-L62

andygrove: It turns out we already have a `withResource` that can work here. 

```scala
withResourceIfAllowed(left.columnarEval(batch)) { lhs =>
  withResourceIfAllowed(right.columnarEval(batch)) { rhs =>
    (lhs, rhs) match {
      case (l: GpuColumnVector, r) =>
        withResource(GpuScalar.from(r, right.dataType)) { scalar =>
          GpuColumnVector.from(doColumnar(l, scalar), dataType)
        }
      case _ => null
    }
  }
}
```

revans2: But further to the point why is this not a `GpuBinaryExpression`?  `BRound` and `Round` are both `BinaryExpression`s which would clean up this code.

nartal1: Thanks @andygrove  and @revans2 for taking a look.  Sorry missed this conversation. The reason I was not making it as `GpuBinaryExpression` is because there isn't an BinaryOp for round or BRound in JNI or in cudf binaryops.  And I was not sure what would be the overidden enum be for these operators. But this also isn't working for all cases. So I will try how to go about it as a `GpuBinaryExpression`

revans2: `GpuBinaryExpression` does not assume an enum but `CudfBinaryExpression` does.

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -183,6 +183,22 @@ def test_shift_right_unsigned(data_gen):
                 'shiftrightunsigned(a, cast(null as INT))',
                 'shiftrightunsigned(a, b)'))
 
+@pytest.mark.parametrize('data_gen', [decimal_gen_scale_precision], ids=idfn)
```
revans2, line 186: `round` and `bround` support all numeric types, not just decimal (float, int, byte, double, etc).  Do we have tests for these too?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
nartal1: Thanks! changed it to use `GpuBinaryExpression`. Will be in next patch.

integration_tests/src/main/python/arithmetic_ops_test.py
nartal1: Yes, will add tests for other numeric types.

nartal1:
Cleaned up the code to use `GpuBinaryExpression`. 
Need to add numerical tests and also verify decimal type for negative scale

nartal1:
> There is a bug in this which I am still figuring out. All test cases do not pass.
> If the scale is 0, the value is incremented by 1(which shouldn't happen).
> If I have something like below:
> 
> ```
> df1 = spark.createDataFrame(data=[(Decimal('3977631')), (Decimal('8291028')), (Decimal('8291025'))], schema=DecimalType(7,0))
> ret = df2.selectExpr('round(value, 0)')
> 
> CPU result: (Decimal('3977631')), (Decimal('8291028')), (Decimal('8291025'))
> GPU result: (Decimal('3977632')), (Decimal('8291029')), (Decimal('8291026'))
> ```

I couldn't figure out the issue here as the code seems straight forward. So, switched to see if i can reproduce it in Java code. This is reproducible in Java side as well. So have to debug it in cudf Java

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -349,6 +350,70 @@ abstract class CudfBinaryMathExpression(name: String) extends CudfBinaryExpressi
...
+  }
+  private lazy val _scale: Int = scaleV.asInstanceOf[Int]
+
+  override def inputTypes: Seq[AbstractDataType] = Seq(NumericType, IntegerType)
+
+  override def doColumnar(val0: GpuColumnVector, val1: Scalar): ColumnVector = {
+    val scaleVal=val1.getInt
+    val scale = dataType match {
+      case DecimalType.Fixed(p, s) => s
+      case ByteType | ShortType | IntegerType | LongType | FloatType | DoubleType => val1.getInt
```
nartal1, line 383: @revans2 Could you please suggest how we handle overflow when for each types.
For example(considering short type), pyspark results in `0` for the below round operation:  
```
>>> df2= spark.createDataFrame(data=[32562], schema=ShortType())
>>> ret = df2.selectExpr('round(value, -5)')
>>> ret.show()
+----------------+
|round(value, -5)|
+----------------+
|               0|
+----------------+

But we see different GPU result(-31072) as overflow results in undefined behavior in libcudf. 
Should we throw an exception whenever we sense an overflow for each type at this point ?

File: integration_tests/src/main/python/arithmetic_ops_test.py [hunk 4]
```diff
@@ -183,6 +183,26 @@ def test_shift_right_unsigned(data_gen):
...
+                'bround(a, 10)')),
+                conf=allow_negative_scale_of_decimal_conf)
+
+@approximate_float
+@pytest.mark.parametrize('data_gen', round_gens, ids=idfn)
+def test_decimal_round(data_gen):
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark: debug_df(unary_op_df(spark, data_gen).selectExpr(
+                'round(a)',
+                #'round(a, -2)',
```
nartal1, line 202: For FloatType seeing `inf`  in few rows in GPU result.  Looking into it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
revans2: The problem is completely in how we implement round/bround vs how spark does it, and I am not 100% sure how to make them sync up without a lot of work on the cudf side for these corner cases.

cudf tries to do the round on the native type, which can result in an overflow.  Spark will convert the native value to a decimal value (128-bits if needed), set the scale to do the rounding, and then convert the value back (with some special cases for NaN and Infinite in floating point).

https://github.com/apache/spark/blob/0626901bcbeebceb6937001e1f32934c71876210/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/mathExpressions.scala#L1220-L1249

There can be no overflow in those cases because all of the processing is happening on 128-bits.  For integer smaller than a long we could cast it to a long first, do the round/bround, and then cast it back.  But we would still end up with issues in long because of overflow.

Similar with float/double.

nartal1:
@revans2 I am waiting on another fix in cudf to get merged before starting the CI here. It would be great if you could please take another look and suggest if looks okay for this iteration. If further changes are required, I can make the changes and verify it locally.

nartal1:
build

nartal1:
build

nartal1:
build

integration_tests/src/main/python/arithmetic_ops_test.py
nartal1: Added tests for other numeric types.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1726,6 +1726,26 @@ object GpuOverrides {
 
         override def convertToGpu(child: Expression): GpuExpression = GpuAverage(child)
       }),
+    expr[BRound](
+      "Round an expression to d decimal places using HALF_EVEN rounding mode",
+      ExprChecks.binaryProjectNotLambda(
+        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all,
+        ("lhs", TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all),
```
revans2, line 1733: Could we make the names a little clearer like `value` and `scale`?

revans2, line 1733: Spark only supports numeric types for "lhs"/"value"  This should be the same as `TypeSig.numeric`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1726,6 +1726,26 @@ object GpuOverrides {
...
+        ("rhs", TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all)),
```
revans2, line 1734: Spark only supports integer (TypeSig.INT) as the "rhs"/"scale" and it must be a literal value.

```
("scale", TypeSig.lit(TypeEnum.INT), TypeSig.lit(TypeEnum.INT))
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1726,6 +1726,26 @@ object GpuOverrides {
...
+      (a, conf, p, r) => new BinaryExprMeta[BRound](a, conf, p, r) {
+        override def convertToGpu(lhs: Expression, rhs: Expression): GpuExpression =
+          GpuBRound(lhs, rhs)
+      }),
+    expr[Round](
+      "Round an expression to d decimal places using HALF_UP rounding mode",
+      ExprChecks.binaryProjectNotLambda(
+        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all,
+        ("lhs", TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all),
+        ("rhs", TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all)),
```
revans2, line 1744: Same comments here I would prefer to see scale as the name instead and TypeSig.INT is all that is supported here.

```
("scale", TypeSig.lit(TypeEnum.INT), TypeSig.lit(TypeEnum.INT))
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1726,6 +1726,26 @@ object GpuOverrides {
...
+      (a, conf, p, r) => new BinaryExprMeta[BRound](a, conf, p, r) {
+        override def convertToGpu(lhs: Expression, rhs: Expression): GpuExpression =
+          GpuBRound(lhs, rhs)
+      }),
+    expr[Round](
+      "Round an expression to d decimal places using HALF_UP rounding mode",
+      ExprChecks.binaryProjectNotLambda(
+        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all,
+        ("lhs", TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all),
```
revans2, line 1743: Same comments here I would prefer to see value as the name and only TypeSig.numeric is supported.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -369,6 +370,70 @@ abstract class CudfBinaryMathExpression(name: String) extends CudfBinaryExpressi
...
+  // by checking if scaleV == null as well.
+  private lazy val scaleV: Any = scale match {
+    case _: GpuExpression => scale.columnarEval(null)
+    case _ => scale.eval(EmptyRow)
+  }
+  private lazy val _scale: Int = scaleV.asInstanceOf[Int]
+
+  override def inputTypes: Seq[AbstractDataType] = Seq(NumericType, IntegerType)
+
+  override def doColumnar(val0: GpuColumnVector, val1: Scalar): ColumnVector = {
```
revans2, line 399: nit: Could we rename the inputs `val0` and `val1` are not that clear.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -369,6 +370,70 @@ abstract class CudfBinaryMathExpression(name: String) extends CudfBinaryExpressi
...
+    val scaleVal=val1.getInt
```
revans2, line 400: nit: we usually have spaces around the '='
```
val scaleVal = val1.getInt
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -369,6 +370,70 @@ abstract class CudfBinaryMathExpression(name: String) extends CudfBinaryExpressi
...
+    val scale = dataType match {
```
revans2, line 401: This is not needed at all.  We already calculated the scale several different ways already.  You can use either `scaleVal` or `_scale`. I personally think that `scaleVal` might be preferable, but either should be fine.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
revans2: Do we have tests for overflow to check if it is working correctly or are we going to mark the operators and incompatible until we can figure out a way to make it work properly?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
```diff
@@ -148,6 +148,10 @@ trait CudfUnaryExpression extends GpuUnaryExpression {
   override def doColumnar(input: GpuColumnVector): ColumnVector = input.getBase.unaryOp(unaryOp)
 }
 
+trait GpuRoundBase extends GpuBinaryExpression {
```
revans2, line 151: Unless we are going to include more code that uses `roundMode` in here I don't see any reason for having this trait.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
nartal1: Sorry, I was using `scaleVal` to get the scale for non-decimals. The reason i had `scale` here was to get the correct scale for `decimals`. We have updated the scale in `DataType` for `DecimalType` and that is what is retrieved here. I think we cannot use `_scale` as it is. Please let me know if it looks okay.

revans2: You are right I missed the part where the scale can depend on the child's scale too.

revans2 (approved):

revans2:
build

revans2:
@andygrove you reviewed this before are you okay with merging it?

andygrove (approved):
LGTM

PR #1256, opened 2020-12-03
Sonatype staging repo for cudf dependency

NvTimLiu:
Add dependency of cudf pre-release jars on Sonatype staging repo.
Sonatype staging repo is needed in the case:
    cudf and spark-rapids jars will to be released at the same time
    cudf jars are on the Sonatyp staging repo ready for release
    need to build spark-rapids with dependency of the staging cudf jars

### DIFF
--- a/jenkins/databricks/build.sh
+++ b/jenkins/databricks/build.sh
@@ -21,6 +21,7 @@ SPARKSRCTGZ=$1
 # version of Apache Spark we are building against
 BASE_SPARK_VERSION=$2
 BUILD_PROFILES=$3
+STAGING_REPO=$4
 
 echo "tgz is $SPARKSRCTGZ"
 echo "Base Spark version is $BASE_SPARK_VERSION"
@@ -47,8 +48,13 @@ CUDA_VERSION=`mvn help:evaluate -q -pl dist -Dexpression=cuda.version -DforceStd
 SPARK_VERSION_TO_INSTALL_DATABRICKS_JARS=$BASE_SPARK_VERSION-databricks
 RAPIDS_BUILT_JAR=rapids-4-spark_$SCALA_VERSION-$SPARK_PLUGIN_JAR_VERSION.jar
 
+BUILD_ARG=""
+if [ -n "$STAGING_REPO" ]; then
+    BUILD_ARG="-s $WORKSPACE/jenkins/settings.xml -P"'sonatype-staging,!artifactory,!mirror-apache-to-urm,!deploy-to-urm'
+fi
+
 echo "Scala version is: $SCALA_VERSION"
-mvn -B -P${BUILD_PROFILES} clean package -DskipTests || true
+mvn -B -P${BUILD_PROFILES} clean package -DskipTests $BUILD_ARG || true
 # export 'M2DIR' so that shims can get the correct cudf/spark dependnecy info
 export M2DIR=/home/ubuntu/.m2/repository
 CUDF_JAR=${M2DIR}/ai/rapids/cudf/${CUDF_VERSION}/cudf-${CUDF_VERSION}-${CUDA_VERSION}.jar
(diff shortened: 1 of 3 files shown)


NvTimLiu:
build

jenkins/settings.xml
```diff
@@ -50,6 +50,17 @@
...
         </repository>
+        <!-- Add Sonatype staging repo for buiding rapids against pre-release jars  -->
+        <repository>
+           <id>sonatype-staging</id>
+           <url>https://oss.sonatype.org/content/repositories/staging</url>
+           <releases>
+             <enabled>true</enabled>
+           </releases>
+           <snapshots>
+             <enabled>false</enabled>
```
jlowe, line 61: I'm confused.  If it's a pre-release jar, isn't that the definition of a snapshot?

GaryShen2008: This is not a snapshot jar, actually it's a pre-release jar, which is uploaded to sonatype staging repo.
Before we release cuDF before our plugin, so the cuDF jar should exist on public maven repo.
For 0.3, the cuDF jar will be built just one day before our plugin. So, I hope to release both after QA sign off with rapids plugin v0.3.

jlowe: Doesn't that mean we technically are not testing the build against the official cudf release in Maven central then?

jenkins/databricks/run-tests.py
```diff
@@ -64,6 +65,8 @@ def main():
           base_spark_pom_version = arg
       elif opt in ('-b', '--bulidprofiles'):
           build_profiles = arg
+      elif opt in ('-u', '--stagingrepo'):
+          staging_repo = arg
```
NvTimLiu, line 69: In Databricks nightly snapshot build, staging_repo=''
In Databricks release build, staging_repo='Sonatype'

jenkins/settings.xml
```diff
@@ -81,14 +87,31 @@
...
+           </snapshots>
+        </repository>
+      </repositories>
+     </profile>
   </profiles>
-  <activeProfiles>
```
NvTimLiu, line 89: Remove `activeProfiles`, seems maven -P!xxx not works for `activeProfiles` items
change to use:
     <activation>
        <activeByDefault>true</activeByDefault>
      </activation>

jenkins/databricks/build.sh
```diff
@@ -47,8 +48,13 @@ CUDA_VERSION=`mvn help:evaluate -q -pl dist -Dexpression=cuda.version -DforceStd
 SPARK_VERSION_TO_INSTALL_DATABRICKS_JARS=$BASE_SPARK_VERSION-databricks
 RAPIDS_BUILT_JAR=rapids-4-spark_$SCALA_VERSION-$SPARK_PLUGIN_JAR_VERSION.jar
 
+BUILD_ARG=""
+if [ -n "$STAGING_REPO" ]; then
+    BUILD_ARG="-s $WORKSPACE/jenkins/settings.xml -P"'sonatype-staging,!artifactory,!mirror-apache-to-urm,!deploy-to-urm'
```
NvTimLiu, line 53: deactive all URM repos by -P'!artifactory,!mirror-apache-to-urm,!deploy-to-urm'
URM repo is unaccessiable in Databricks

NvTimLiu:
build

tgravescs:
So once the real cudf version is released we would remove cudf from the sonatype staging and we would just pick up the released version, correct?

jenkins/databricks/run-tests.py
tgravescs: staging_repo arg here need an argument?  doesn't look like its used other then for a boolean which cuases us to set the build_arg. If that is the case perhaps use "u" instead of "u:" and set staging_repo here to something like "USE_STAGING" or True for the build.sh script to pick up and act on.

tgravescs:
I don't see changes to the Jenkinsfile so I assume you are doing that as separate release Jenkinsfile or something?

GaryShen2008:
As we discussed, we'll cancel this PR since we plan to release cuDF 0.17 before our release build of spark-rapids 0.3.

Issue #1179, opened 2020-11-20
[BUG] SerializeConcatHostBuffersDeserializeBatch may have thread issues


jlowe:
#1174 fixed an issue initializing the columnar batch row count on a degenerate batch that was row-only, but since that change is necessary to fix an issue it implies threads are trying to access the batch _while it is being created_.  The concern is if there could be multiple threads trying to create a normal GPU batch simultaneously and end up leaking all but the last one recorded.

It's also disconcerting that #1165 needed to be reverted, so we need to understand what the threading implications are here and make sure the serializer is doing the correct thing in light of multiple thread access.

andygrove:
Here is the stack trace that we saw in the integration tests that led to us reverting the previous PR.

```
java.io.StreamCorruptedException: unexpected block data
20:30:07  	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1663)
20:30:07  	at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2365)
20:30:07  	at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2289)
20:30:07  	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2147)
20:30:07  	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1646)
20:30:07  	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:482)
20:30:07  	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:440)
20:30:07  	at org.apache.spark.serializer.JavaDeserializationStream.readObject(JavaSerializer.scala:76)
20:30:07  	at org.apache.spark.broadcast.TorrentBroadcast$.$anonfun$unBlockifyObject$4(TorrentBroadcast.scala:328)
20:30:07  	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1377)
20:30:07  	at org.apache.spark.broadcast.TorrentBroadcast$.unBlockifyObject(TorrentBroadcast.scala:330)
20:30:07  	at org.apache.spark.broadcast.TorrentBroadcast.$anonfun$readBroadcastBlock$4(TorrentBroadcast.scala:249)
20:30:07  	at scala.Option.getOrElse(Option.scala:189)
20:30:07  	at org.apache.spark.broadcast.TorrentBroadcast.$anonfun$readBroadcastBlock$2(TorrentBroadcast.scala:223)
[...]

PR #1264, opened 2020-12-03
Fix race condition in batch creation

andygrove:

In `GpuBroadcastNestedLoopJoinExecBase#doExecuteColumnar` the following code computes the number of rows the join will output by multiplying each batch row count by the `buildCount` which is the row-count of the broadcast side of the relation.

```scala
val counts = streamed.executeColumnar().map(getRowCountAndClose)
GpuNoColumnCrossJoin.divideIntoBatches(
  counts.map(s => s * buildCount),
  targetSizeBytes,
  numOutputRows,
  numOutputBatches)
```

`buildCount` is a lazy val, defined as follows:

```scala
lazy val buildCount: Int = computeBuildRowCount(broadcastRelation, buildTime, buildDataSize)
```

The method called when instantiating `buildCount` contains the following code:

```scala
broadcastRelation.value.batch.numRows()
```

This reference to `batch` calls code that will create the batch if it does not already exist. 

```scala
def batch: ColumnarBatch = {
  if (batchInternal == null) {
    .... 
    batchInternal = readObject(...)
  }
  batchInternal
}

```

Through adding some debug code I was able to prove that multiple threads are following the code path from `counts.map(s => s * buildCount)` and trying to instantiate an instance of the `lazy val` at the same time (note that they all have a distinct `lazy val` and are not referencing a shared instance because this `lazy val` is local within the code block).

The result is that multiple batches get created and references are lost to all but one of them and this will cause leaks.

This PR fixes the issue by synchronizing the batch creation logic.

This closes https://github.com/NVIDIA/spark-rapids/issues/1179

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
@@ -60,7 +60,7 @@ class SerializeConcatHostBuffersDeserializeBatch(
       writeObject(oout)
       val barr = out.toByteArray
       val oin = new ObjectInputStream(new ByteArrayInputStream(barr))
-      batchInternal = readObject(oin)
+      readObject(oin)
     }
     batchInternal
   }
@@ -81,7 +81,7 @@ class SerializeConcatHostBuffersDeserializeBatch(
     }
   }
 
-  private def readObject(in: ObjectInputStream): ColumnarBatch = {
+  private def readObject(in: ObjectInputStream): Unit = {
     val range = new NvtxRange("DeserializeBatch", NvtxColor.PURPLE)
     try {
       val tableInfo: JCudfSerialization.TableAndRowCountPair =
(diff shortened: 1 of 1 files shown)


andygrove:
I am now wondering what is different between this and the PR we had to revert [1]. They are almost identical.

[1] https://github.com/NVIDIA/spark-rapids/pull/1165

jlowe:
> I am now wondering what is different between this and the PR we had to revert

I am assigned to #1179 and was just about to ask you what it was about #1165 that cause so much havok.

To me the race is clearly there, as we're directly calling `.batch` on the broadcast variable in task execution code, so if there's more than one concurrent task then more than one thread will try to create the broadcast batch instance at the same time.  What's a mystery is why adding locking in #1165 was bad -- all I heard is we had to revert it, not what it did or why.

andygrove:
I've updated the issue with an example stack trace that we saw in the integration test failures that led to this being reverted.

Sorry, I thought you knew I was looking into this. I mentioned at standup that I'd look at at this afternoon but there is more to it than I thought so probably best to hand it back to you. Let me know if I can help though.

jlowe (approved):

andygrove:
build

pxLi:
build

revans2:
build

PR #1265, opened 2020-12-03
Chunking input before writing a ParquetCachedBatch

razajafri:
This PR deals with breaking up the incoming batches before writing them to a ParquetCachedBatch to make sure the CachedBuffer never exceeds the Int max-value. 

I am still working on writing the tests to test the CPU writing. This PR so far only has tests for the `compressColumnarBatchWithParquet`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -319,7 +319,7 @@
             <dependency>
               <groupId>org.mockito</groupId>
               <artifactId>mockito-core</artifactId>
-              <version>2.28.2</version>
+              <version>3.6.0</version>
               <scope>test</scope>
             </dependency>
         </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -171,6 +171,7 @@
         <spark301.version>3.0.1</spark301.version>
         <spark302.version>3.0.2-SNAPSHOT</spark302.version>
         <spark310.version>3.1.0-SNAPSHOT</spark310.version>
+        <mockito.version>3.6.0</mockito.version>
     </properties>
 
     <dependencyManagement>
@@ -319,7 +320,7 @@
             <dependency>
               <groupId>org.mockito</groupId>
               <artifactId>mockito-core</artifactId>
-              <version>3.6.0</version>
+              <version>${mockito.version}</version>
               <scope>test</scope>
             </dependency>
         </dependencies>
--- a/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
+++ b/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
@@ -1249,8 +1249,7 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
         new RowToColumnarIterator(iter, structSchema, RequireSingleBatch, converters)
       })
       columnarBatchRdd.flatMap(cb => {
-        withResource(cb) { cb => compressColumnarBatchWithParquet(cb)
-        }
+        withResource(cb)(cb => compressColumnarBatchWithParquet(cb))
       })
     } else {
       val broadcastedHadoopConf = getBroadcastedHadoopConf(conf, parquetSchema)
--- a/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
+++ b/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
@@ -166,7 +166,7 @@ private class ByteArrayOutputFile(stream: ByteArrayOutputStream) extends OutputF
   override def defaultBlockSize(): Long = ByteArrayOutputFile.BLOCK_SIZE
 }
 
-private[rapids] class ParquetBufferConsumer(val numRows: Int) extends HostBufferConsumer with
+private class ParquetBufferConsumer(val numRows: Int) extends HostBufferConsumer with
   AutoCloseable {
   @transient private[this] val offHeapBuffers = mutable.Queue[(HostMemoryBuffer, Long)]()
   private var buffer: Array[Byte] = _
@@ -336,6 +336,8 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
   }
 
   val _2GB: Long = 2L * 1024 * 1024 * 1024
+  val APPROX_PAR_META_DATA = 10 * 1024 * 1024 // we are estimating 10MB
+  val BYTES_ALLOWED_PER_BATCH = _2GB - APPROX_PAR_META_DATA
 
   private[rapids] def compressColumnarBatchWithParquet(
      gpuCB: ColumnarBatch): List[ParquetCachedBatch] = {
@@ -344,7 +346,7 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
     val estimatedRowSize = scala.Range(0, gpuCB.numCols()).map { index =>
       gpuCB.column(index).dataType().defaultSize
     }.sum
-    val rowsInBatch = (_2GB / estimatedRowSize).toInt
+    val rowsInBatch = (BYTES_ALLOWED_PER_BATCH / estimatedRowSize).toInt
     val splitIndices = scala.Range(rowsInBatch, gpuCB.numRows(), rowsInBatch)
     val cudfTables = if (splitIndices.nonEmpty) {
       val splutVectors = scala.Range(0, gpuCB.numCols()).map { index =>
@@ -1050,14 +1052,14 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
               parquetOutputFormat.getRecordWriter(outputFile, sharedHadoopConf)
             }
             var totalSize = 0
-            // TODO: the following could still potentially produce a buffer more than 2g
-            while (rowIterator.hasNext && totalSize < _2GB) {
+            var leftOverRow: Option[InternalRow] = None
+            while (rowIterator.hasNext && totalSize < BYTES_ALLOWED_PER_BATCH) {
               rows += 1
               if (rows < 0) {
                 throw new IllegalStateException("CachedBatch doesn't support rows larger " +
                   "than Int.MaxValue")
               }
-              val row = rowIterator.next()
+              val row = leftOverRow.getOrElse(rowIterator.next())
               totalSize += {
                 row match {
                   case r: UnsafeRow =>
@@ -1069,7 +1071,11 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
                     }.sum
                 }
               }
-              recordWriter.write(null, row)
+              if (totalSize <= BYTES_ALLOWED_PER_BATCH) {
+                recordWriter.write(null, row)
+              } else {
+                leftOverRow = Some(row)
+              }
             }
             // passing null as context isn't used in this method
             recordWriter.close(null)
--- a/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
+++ b/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
@@ -351,7 +351,7 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
 
   private[rapids] def compressColumnarBatchWithParquet(
      gpuCB: ColumnarBatch): List[ParquetCachedBatch] = {
-    // NOTE: this doesn't take nulls into account so we could be over-estimating the acutal size
+    // NOTE: this doesn't take nulls into account so we could be over-estimating the actual size
     // but that's still better than under-estimating
     val estimatedRowSize = scala.Range(0, gpuCB.numCols()).map { index =>
       gpuCB.column(index).dataType().defaultSize
@@ -1331,4 +1331,3 @@ private object ParquetOutputFileFormat {
     memoryManager
   }
 }
-
--- a/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
+++ b/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
@@ -253,6 +253,16 @@ private case class CloseableColumnBatchIterator(iter: Iterator[ColumnarBatch]) e
  * This class assumes, the data is Columnar and the plugin is on
  */
 class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
+
+  private[rapids] def getProducer(
+     cbIter: Iterator[ColumnarBatch],
+     schema: Seq[Attribute],
+     conf: Configuration,
+     sqlConf: SQLConf) : CachedBatchIteratorProducer[ColumnarBatch] = {
+    new CachedBatchIteratorProducer[ColumnarBatch](cbIter, schema, conf, sqlConf)
+
+  }
+
   override def supportsColumnarInput(schema: Seq[Attribute]): Boolean = true
 
   override def supportsColumnarOutput(schema: StructType): Boolean = schema.fields.forall { f =>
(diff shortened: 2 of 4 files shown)


shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -253,6 +253,16 @@ private case class CloseableColumnBatchIterator(iter: Iterator[ColumnarBatch]) e
  * This class assumes, the data is Columnar and the plugin is on
  */
 class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
+
```
razajafri, line 256: This change was made to gain access to the Inner class, open to suggestions to make this better

razajafri:
This is pretty much done. It's still a WIP because the thing remaining is that the tests won't compile with Spark 3.0.0. I have used reflection for some tests that were simple but for others, especially the ones on the CPU, the tests looked really hard to read after I tried reflection.

@tgravescs following up on our conversation, I think we should think about alternatives like you said, maybe these tests can live with the shim310?

razajafri:
As decided in the sprint-planning I am removing the WIP tag and marking this for review. We will include tests for the CPU part when we have a framework to compile tests selectively

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -253,6 +253,16 @@ private case class CloseableColumnBatchIterator(iter: Iterator[ColumnarBatch]) e
...
+  private[rapids] def getProducer(
```
jlowe, line 257: Unused.

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -330,19 +340,58 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
-    withResource(GpuColumnVector.from(gpuCB)) { table =>
-      withResource(Table.writeParquetChunked(ParquetWriterOptions.DEFAULT, buffer)) { writer =>
-        writer.write(table)
+  val _2GB: Long = 2L * 1024 * 1024 * 1024
+  val APPROX_PAR_META_DATA: Int = 10 * 1024 * 1024 // we are estimating 10MB
+  val BYTES_ALLOWED_PER_BATCH: Long = _2GB - APPROX_PAR_META_DATA
+
+  private[rapids] def compressColumnarBatchWithParquet(
+     gpuCB: ColumnarBatch): List[ParquetCachedBatch] = {
+    // NOTE: this doesn't take nulls into account so we could be over-estimating the acutal size
```
jlowe, line 354:
```suggestion
    // NOTE: this doesn't take nulls into account so we could be over-estimating the actual size
```

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -330,19 +340,58 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
+    // but that's still better than under-estimating
+    val estimatedRowSize = scala.Range(0, gpuCB.numCols()).map { index =>
+      gpuCB.column(index).dataType().defaultSize
+    }.sum
+    val rowsInBatch = (BYTES_ALLOWED_PER_BATCH / estimatedRowSize).toInt
+    val splitIndices = scala.Range(rowsInBatch, gpuCB.numRows(), rowsInBatch)
+    val cudfTables = if (splitIndices.nonEmpty) {
+      val splutVectors = scala.Range(0, gpuCB.numCols()).map { index =>
```
jlowe, line 362: Was this supposed to be named `splitVectors`?

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -330,19 +340,58 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
+  private[rapids] def compressColumnarBatchWithParquet(
+     gpuCB: ColumnarBatch): List[ParquetCachedBatch] = {
+    // NOTE: this doesn't take nulls into account so we could be over-estimating the acutal size
+    // but that's still better than under-estimating
+    val estimatedRowSize = scala.Range(0, gpuCB.numCols()).map { index =>
+      gpuCB.column(index).dataType().defaultSize
+    }.sum
+    val rowsInBatch = (BYTES_ALLOWED_PER_BATCH / estimatedRowSize).toInt
+    val splitIndices = scala.Range(rowsInBatch, gpuCB.numRows(), rowsInBatch)
+    val cudfTables = if (splitIndices.nonEmpty) {
```
jlowe, line 361: This seems to be leaked if there's an exception.

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -330,19 +340,58 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
+      try {
+        // Splitting the table
+        // e.g. T0 = {col1, col2,...,coln} => split columns into 'm' cols =>
+        // T00= {splitCol1(0), splitCol2(0),...,splitColn(0)}
+        // T01= {splitCol1(1), splitCol2(1),...,splitColn(1)}
+        // ...
+        // T0m= {splitCol1(m), splitCol2(m),...,splitColn(m)}
+        for (i <- splutVectors(0).indices) yield
+          new Table({
+            for (j <- splutVectors.indices) yield splutVectors(j)(i)
```
jlowe, line 374: This has the drawback of creating a bunch of leakable resources up front. I believe this will be easier to maintain and read if we construct the one table we need to perform the write then close that table rather than creating them all up-front first.  The code that writes a single table can be refactored into a separate method so it's easy to call from this split case and the non-split case rather than needing to fall-through to common-code.

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -330,19 +340,58 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
+  val _2GB: Long = 2L * 1024 * 1024 * 1024
+  val APPROX_PAR_META_DATA: Int = 10 * 1024 * 1024 // we are estimating 10MB
+  val BYTES_ALLOWED_PER_BATCH: Long = _2GB - APPROX_PAR_META_DATA
+
+  private[rapids] def compressColumnarBatchWithParquet(
+     gpuCB: ColumnarBatch): List[ParquetCachedBatch] = {
+    // NOTE: this doesn't take nulls into account so we could be over-estimating the acutal size
+    // but that's still better than under-estimating
+    val estimatedRowSize = scala.Range(0, gpuCB.numCols()).map { index =>
+      gpuCB.column(index).dataType().defaultSize
```
jlowe, line 357: Since these are GPU vectors, wouldn't it be more accurate to estimate the row size based on the actual device memory size and dividing by the number of rows?  `ColumnVector.getDeviceMemorySize()`

tests/src/test/scala/com/nvidia/spark/rapids/ParquetWriterSuite.scala
```diff
@@ -118,4 +133,111 @@ class ParquetWriterSuite extends SparkQueryCompareTestSuite {
...
+      withResource(cudfCols) { _=>
+        val cb = new ColumnarBatch(gpuCols, ROWS)
+        whenSplitCalled(cb)
+        try {
+          val method = compressWithParquetMethod.getOrElse {
+            val classOfSerializer = Class.forName(
+              "com.nvidia.spark.rapids.shims.spark310.ParquetCachedBatchSerializer")
+            parquetSerializerInstance = Some(classOfSerializer.newInstance())
+            val compressWithParquet =
+              classOfSerializer.getMethod("compressColumnarBatchWithParquet",
```
jlowe, line 231: Why is this using reflection to call code we wrote?

jlowe (changes requested):

tests/src/test/scala/com/nvidia/spark/rapids/ParquetWriterSuite.scala
razajafri: Because the code won't compile against Spark 3.0 if we directly reference the `ParquetCachedBatchSerializer`. 

This is why we need a test framework with which we can ignore tests at compile time.

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
razajafri: Will be addressed in #1285

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
razajafri: Fair point. This will be addressed in #1286?

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
razajafri: This will be addressed in #1287

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
razajafri: This was me being creative, :)  but I can rename this as a follow-on #1288

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
razajafri: Can we address this as a part of #1289?

razajafri:
I have created follow-ons to address your comments @jlowe

tests/src/test/scala/com/nvidia/spark/rapids/ParquetWriterSuite.scala
jlowe: ah right the tests problem.  that too will be a followup.

jlowe (dismissed):

jlowe:
CI will probably fail without merging up to branch-0.3 which contains the fix for #1275

jlowe:
build

razajafri:
build

razajafri:
build

razajafri:
build

razajafri:
build

GaryShen2008:
build

razajafri:
build

razajafri:
build

razajafri:
Blossom is timing out. Kicking it off again

razajafri:
Build

razajafri:
Build

pom.xml
```diff
@@ -319,7 +319,7 @@
             <dependency>
               <groupId>org.mockito</groupId>
               <artifactId>mockito-core</artifactId>
-              <version>2.28.2</version>
+              <version>3.6.0</version>
```
revans2, line 322: nit: Is this change still needed?

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -330,19 +340,58 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
-      withResource(Table.writeParquetChunked(ParquetWriterOptions.DEFAULT, buffer)) { writer =>
-        writer.write(table)
+  val _2GB: Long = 2L * 1024 * 1024 * 1024
+  val APPROX_PAR_META_DATA: Int = 10 * 1024 * 1024 // we are estimating 10MB
+  val BYTES_ALLOWED_PER_BATCH: Long = _2GB - APPROX_PAR_META_DATA
+
+  private[rapids] def compressColumnarBatchWithParquet(
+     gpuCB: ColumnarBatch): List[ParquetCachedBatch] = {
+    // NOTE: this doesn't take nulls into account so we could be over-estimating the actual size
+    // but that's still better than under-estimating
```
revans2, line 355: But this could underestimate the size.  It makes some assumptions like strings are 10 bytes long, and that arrays hold a single value in them.

Why are we not getting the actual size of the batch and using it guess at the number of rows per batch.  It would still not be perfect, but better than this.

```
estimatedRowSize = GpuColumnVector.getTotalDeviceMemoryUsed(gpuCB) / gpuCB.getNumRows()
```

Might need to check for empty batches, but hopefully we don't have any at this point.

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -330,19 +340,58 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
+      } finally {
+        for (i <- splutVectors(0).indices)
+            for (j <- splutVectors.indices) splutVectors(j)(i).close()
+      }
+    } else {
+      Seq(GpuColumnVector.from(gpuCB))
+    }
+
+    val buffers = new ListBuffer[ParquetCachedBatch]
+    cudfTables.foreach { table =>
```
revans2, line 385: This will leak a number of things in the case of an exception.  I'm not sure it is worth the added complexity right now to fix it, but it would be good to at least put in a comment and file a follow on issue to look into it.

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -1162,10 +1248,8 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
       val columnarBatchRdd = input.mapPartitions(iter => {
         new RowToColumnarIterator(iter, structSchema, RequireSingleBatch, converters)
       })
-      columnarBatchRdd.map(cb => {
-        withResource(cb) { columnarBatch =>
-          val cachedBatch = compressColumnarBatchWithParquet(columnarBatch)
-          cachedBatch
+      columnarBatchRdd.flatMap(cb => {
+        withResource(cb) { cb => compressColumnarBatchWithParquet(cb)
```
revans2, line 1252: nit: It is probably cleaner to do it like the following.  At a minimum if the closure can fit on a single line it is ideal to use `()` instead of `{}` and to put it on a single line.

```
withResource(cb)(compressColumnarBatchWithParquet)
```

tests/pom.xml
```diff
@@ -104,7 +104,8 @@
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
-            <artifactId>mockito-core</artifactId>
+            <artifactId>mockito-inline</artifactId>
+            <version>3.6.0</version>
```
revans2, line 108: If we need a version in multiple places we should have the base pom set a property with that version so we can be consistent.  Also we should have the base POM do the setting of the versions in the dependencyManagement section.

tests/src/test/scala/com/nvidia/spark/rapids/ParquetWriterSuite.scala
```diff
@@ -118,4 +125,112 @@ class ParquetWriterSuite extends SparkQueryCompareTestSuite {
...
+          // noop
+        }
+      })
+
+    withResource(cudfCols) { _ =>
+      val cb = new ColumnarBatch(gpuCols, ROWS)
+      whenSplitCalled(cb)
+      try {
+        val method = compressWithParquetMethod.getOrElse {
+          val classOfSerializer = Class.forName(
```
revans2, line 221: I would prefer to have us file a follow on issue to have tests that compile/run depending on the version of the dependency instead of having us do reflection.

pom.xml
razajafri: yes, for mocking static methods in `Table'

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
razajafri: Thanks Bobby, an issue has already been filed for this. Please checkout #1285

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
razajafri: #1287 is filed

pom.xml
revans2: I really don't feel good about mocking static methods, but OK.

tests/src/test/scala/com/nvidia/spark/rapids/ParquetWriterSuite.scala
razajafri: Can we let this one stay? I would rather have some parts tested, I already have filed a follow-on for the test framework
#1290

revans2 (approved):

razajafri:
build

revans2 (approved):

Issue #1271, opened 2020-12-04
[BUG] CastOpSuite and AnsiCastOpSuite failing with ArithmeticException on Spark 3.1


jlowe:
**Describe the bug**
Testing against the latest Apache Spark 3.1 SNAPSHOT the CastOpSuite fails like this:
```
CastOpSuite:
- Test all supported casts with in-range values *** FAILED ***
  Cast from FloatType to IntegerType failed; ansi=true org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 1505.0 failed 1 times, most recent failure: Lost task 0.0 in stage 1505.0 (TID 1518) (10.28.9.126 executor driver): java.lang.ArithmeticException: Casting 2.14748365E9 to int causes overflow
  	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext(Unknown Source)
  	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
  	at org.apache.spark.sql.execution.WholeStageCodegenExec$$anon$1.hasNext(WholeStageCodegenExec.scala:755)
  	at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:458)
  	at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:458)
  	at org.apache.spark.util.random.SamplingUtils$.reservoirSampleAndCount(SamplingUtils.scala:41)
  	at org.apache.spark.RangePartitioner$.$anonfun$sketch$1(Partitioner.scala:306)
  	at org.apache.spark.RangePartitioner$.$anonfun$sketch$1$adapted(Partitioner.scala:304)
  	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsWithIndex$2(RDD.scala:915)
  	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsWithIndex$2$adapted(RDD.scala:915)
  	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
  	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:373)
  	at org.apache.spark.rdd.RDD.iterator(RDD.scala:337)
  	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:90)
  	at org.apache.spark.scheduler.Task.run(Task.scala:131)
[...]

PR #1275, opened 2020-12-04
Temporarily disable more CAST tests for Spark 3.1.0

andygrove:

Temporarily disable more CAST tests for Spark 3.1.0 due to recent changes causing test failures.

Follow-up issue is https://github.com/NVIDIA/spark-rapids/issues/1271

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
@@ -531,7 +531,6 @@ object CastOpSuite {
 
   def intsAsFloats(session: SparkSession): DataFrame = {
     import session.sqlContext.implicits._
-
     intValues.map(_.toFloat).toDF("c0")
   }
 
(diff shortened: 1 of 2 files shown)


andygrove:
build

tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
```diff
@@ -56,6 +56,7 @@ class CastOpSuite extends GpuExpressionTestSuite {
     case (TimestampType | DateType, _: NumericType) => true
     case (TimestampType | DateType, BooleanType) => true
     case (StringType, TimestampType) => true
+    case (FloatType, IntegerType) => true
```
jlowe, line 59: The disable I understand, but why was this change made?

andygrove: I updated issue https://github.com/NVIDIA/spark-rapids/issues/1271 explaining the Spark change and why it broke our tests

andygrove: Sorry, you are asking about this code change. This is a method that determines which CASTs to skip against 3.1.0 so this is just adding another one to that list.

andygrove:
```scala
  def should310SkipAnsiCast(from: DataType, to: DataType): Boolean = (from, to) match {
    case (_: NumericType, TimestampType | DateType) => true
    case (BooleanType, TimestampType | DateType) => true
    case (TimestampType | DateType, _: NumericType) => true
    case (TimestampType | DateType, BooleanType) => true
    case (StringType, TimestampType) => true
    case (FloatType, IntegerType) => true
    case _ => false
  }
```

jlowe (approved):

jlowe:
build

jlowe:
build

PR #1278, opened 2020-12-04
Add a config to disable decimal types by default

abellina:

@jlowe @revans2 I had to change this in more places than I would have liked. Marking this as draft for now as I am still running `verify`


jlowe:
Missing the `configs.md` doc change, but otherwise lgtm.

abellina:
So need to fix some tests that are looking for a gpu-only plan, `mvn verify` returned ;)

abellina:
The tests that have decimal need to be updated to turn on decimal in the tests. I am proposing just to turn on decimals for all tests, just forcing the RapidsConf setting to true for decimals. @jlowe suggested adding a test that verifies the default false value for the setting. @revans2 FYI, let me know if you want me to enable decimals for each test that actually needs it.

jlowe:
>  @jlowe suggested adding a test that verifies the default false value for the setting.

To clarify, I'd like to see a test that setting the config to disable decimal types actually does disable decimal types.

abellina:
> > @jlowe suggested adding a test that verifies the default false value for the setting.
> 
> To clarify, I'd like to see a test that setting the config to disable decimal types actually does disable decimal types.
Oh that's not what I did. Let me fix that.

abellina:
Should be better now @jlowe https://github.com/NVIDIA/spark-rapids/pull/1278/commits/ae55df31eccd66da34d99a8271dfa45f8688ca12

abellina:
build

jlowe (approved):

abellina:
build

abellina:
build

pxLi:
build

pxLi:
build

pxLi:
build

pxLi:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -457,6 +457,12 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(false)
 
+  val DECIMAL_TYPE_ENABLED = conf("spark.rapids.sql.decimalType.enabled")
+      .doc("Enable decimal types up to 64-bit precision. Note that not all GPU " +
```
revans2, line 461: nit: I'm not sure that this tells the whole story. I would prefer something like.

```
.doc("Enable decimal type support on the GPU.  Decimal support on the GPU is limited to less than 18 digits and " +
    "and is only supported by a small number of operations currently.  This can result in a lot of data movement " +
    "to and from the GPU, which can slow down processing in some cases.")
```

revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
abellina: let me change that

abellina:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
abellina: that's done @revans2

revans2 (approved):
My only comment now is that a docs change went in for type support.  We will need to update it with a comment about decimal being off by default.  You can do it here if you want or I can do a separate one.  Either way is fine with me.

abellina:
> My only comment now is that a docs change went in for type support. We will need to update it with a comment about decimal being off by default. You can do it here if you want or I can do a separate one. Either way is fine with me.

Looking at the doc I see at the top: https://github.com/NVIDIA/spark-rapids/blob/e555a0d375e4b4b2eb10e2e681d49e3e338d5144/docs/supported_ops.md#decimal, a good place for us to mention this. 

We also do qualify decimals in a number of other places. There are two instances of these caveats:
```
* as was state previously Decimal is only supported up to a precision of 18 and Timestamp is only supported in the UTC time zone.
```
So we can add "Decimals are off by default" here too I think. Let me know any of the above sounds reasonable, and I can put up a commit here.

revans2:
That is exactly what I was thinking.

revans2 (approved):

abellina:
build

abellina:
@revans2 you still OK with this?

revans2 (approved):

PR #1294, opened 2020-12-07
Support reading decimal columns from parquet files

sperlingxx:
This pull request is to enable reading decimal columns from parquet files via turning `allowDecimal` to `true`.  This pull request also provides test coverage for decimal reading.
But there exist some limits on decimal reading: we can only read decimal columns whose storage type is `INT32/64`. For now, cuDF doesn't support `FIXED_LENGTH_BYTE_ARRAY`. `INT32` will be read as `INT64` because we only support `DECIMAL64` in spark-rapids.


sperlingxx:
build

sperlingxx:
build

sperlingxx:
build

integration_tests/src/main/python/parquet_test.py
```diff
@@ -27,14 +27,20 @@ def read_parquet_df(data_path):
 def read_parquet_sql(data_path):
     return lambda spark : spark.sql('select * from parquet.`{}`'.format(data_path))
 
+
+# LIMIT: We have to use DECIMAL64(precision>8) to test decimal reading, because we only support DECIMAL64 in current.
```
revans2, line 31: Either we need to be able to support something that is read as DECIMAL32 or we cannot ship this.  Also because of performance issues we are shipping 0.3 with decimal disabled by default #1278. That is until we can get a simple cost based optimizer in place or we have enough support for decimal that it is generally worth it. As such if this is hard to fix I'm not sure we need to get this into the 0.3 release.

sperlingxx:
build

integration_tests/src/main/python/parquet_test.py
sperlingxx: Hi @revans2, I added a type conversion for DECIMAL32 columns who are read via `cudf.readParquet`. It seems to work, but I am not sure whether this is a good approach.

sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -197,6 +198,33 @@ object GpuParquetScanBase {
...
+    }
+    val ret = new Table(columns: _*)
+    // clean temporary column vectors produced by castTo
+    (0 until t.getNumberOfColumns).foreach {
+      case i if t.getColumn(i).getType.getTypeId == DType.DTypeEnum.DECIMAL32 =>
+        columns(i).close()
+      case _ =>
+    }
+    // clean original table
+    t.close()
```
revans2, line 225: All of this can cause leaks in the case of an exception. I think we can fix it with very little change though.

revans2 (commented):
@sameerz @jlowe @tgravescs I personally feel that this is getting to be way too late to put this into the 0.3 release. We really should be looking at bug fixes instead of new feature work.  Especially because we know that  https://github.com/rapidsai/cudf/issues/6909 is still an outstanding issue.  But if this is something we need to get in, then we need to at a minimum we need to throw an exception if we get back a data type we don't expect for a decimal column.

jlowe:
> I personally feel that this is getting to be way too late to put this into the 0.3 release. 

Agree.  I would like to see the legacy decimal encoding supported before this goes in, otherwise we're left in a situation where the plugin crashes a query that used to work without it.

sperlingxx:
build

sperlingxx:
build

jlowe:
@sperlingxx per the above discussion, I think this PR should be retargeted to branch-0.4.

sperlingxx:
> @sperlingxx per the above discussion, I think this PR should be retargeted to branch-0.4.

Yes! So, I labeled this pull request with `WIP`.

jlowe (changes requested):
> Yes! So, I labeled this pull request with WIP.

That's fine, adding a request to retarget this PR to branch-0.4 so it cannot be accidentally merged in the interim.

sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2201,9 +2202,20 @@ object GpuOverrides {
...
+        (TypeSig.commonCudfTypes + TypeSig.STRUCT + TypeSig.MAP + TypeSig.ARRAY +
+            TypeSig.DECIMAL).nested(),
         TypeSig.all),
       (p, conf, parent, r) => new SparkPlanMeta[BatchScanExec](p, conf, parent, r) {
+        override def tagPlanForGpu(): Unit = {
+          val hasDecimal = p.scan.readSchema().exists(_.dataType.isInstanceOf[DecimalType])
+          if (hasDecimal) {
+            if (!(p.scan.isInstanceOf[ParquetScan] && this.conf.decimalTypeEnabled)) {
+              this.willNotWorkOnGpu(
+                "Only supports reading decimal from parquet with DECIMAL_TYPE_ENABLED = true")
```
jlowe, line 2214: Users won't know what `DECIMAL_TYPE_ENABLED` is.  I don't think the decimal enable config is relevant here, and this should be simplified to something like, "Decimals are only supported by Parquet".

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -147,8 +147,9 @@ class Spark300Shims extends SparkShims {
       GpuOverrides.exec[FileSourceScanExec](
         "Reading data from files, often from Hive tables",
         ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.STRUCT + TypeSig.MAP +
-            TypeSig.ARRAY).nested(), TypeSig.all),
+            TypeSig.ARRAY + TypeSig.DECIMAL).nested(), TypeSig.all),
```
jlowe, line 150: What about other shims that are specifying scan overrides like Spark310Shims?

jlowe (changes requested):
Test failures are related, I'm guessing because the 3.1.0 shims were not updated.

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
sperlingxx: Yes, I missed it. And it leaded to the CI failure.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: Yes! I've removed the redundant `decimalTypeEnabled` config check.

sperlingxx:
build

integration_tests/src/main/python/parquet_test.py
```diff
@@ -27,23 +27,31 @@ def read_parquet_df(data_path):
...
 
 # test with original parquet file reader, the multi-file parallel reader for cloud, and coalesce file reader for
 # non-cloud
-original_parquet_file_reader_conf={'spark.rapids.sql.format.parquet.reader.type': 'PERFILE'}
-multithreaded_parquet_file_reader_conf={'spark.rapids.sql.format.parquet.reader.type': 'MULTITHREADED'}
-coalesce_parquet_file_reader_conf={'spark.rapids.sql.format.parquet.reader.type': 'COALESCING'}
+original_parquet_file_reader_conf = {'spark.rapids.sql.format.parquet.reader.type': 'PERFILE'}
+multithreaded_parquet_file_reader_conf = {'spark.rapids.sql.format.parquet.reader.type': 'MULTITHREADED'}
+coalesce_parquet_file_reader_conf = {'spark.rapids.sql.format.parquet.reader.type': 'COALESCING'}
+decimal_type_enable_conf = {'spark.rapids.sql.decimalType.enabled': 'true'}
```
revans2, line 52: This should be on by default right now in all of the python tests.

integration_tests/src/main/python/parquet_test.py
```diff
@@ -193,11 +201,28 @@ def test_ts_read_fails_datetime_legacy(gen, spark_tmp_path, ts_write, ts_rebase,
             lambda spark : readParquetCatchException(spark, data_path),
             conf=all_confs)
 
+
+# https://github.com/NVIDIA/spark-rapids/issues/1324
```
revans2, line 205: nit: I would prefer to have us describe what we are testing instead of just adding an issue number.  The issue number is fine for xfail because it provides tracking so I can see if the issue is fixed and we can go back to look if the test is now passing.  Here I'm not even sure we need a comment because the name of the test explains a lot of this.

integration_tests/src/main/python/parquet_test.py
```diff
@@ -193,11 +201,28 @@ def test_ts_read_fails_datetime_legacy(gen, spark_tmp_path, ts_write, ts_rebase,
...
+    assert_gpu_and_cpu_are_equal_collect(read_func(data_path), conf=all_confs)
+
+
 parquet_gens_legacy_list = [[byte_gen, short_gen, int_gen, long_gen, float_gen, double_gen,
     string_gen, boolean_gen, DateGen(start=date(1590, 1, 1)),
-    TimestampGen(start=datetime(1900, 1, 1, tzinfo=timezone.utc))],
-    pytest.param([timestamp_gen], marks=pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/133')),
-    pytest.param([date_gen], marks=pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/133'))]
+    TimestampGen(start=datetime(1900, 1, 1, tzinfo=timezone.utc))] + decimal_gens,
+                            pytest.param([timestamp_gen], marks=pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/133')),
```
revans2, line 224: nit: can we move the indentation back?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2201,9 +2202,19 @@ object GpuOverrides {
     exec[BatchScanExec](
       "The backend for most file input",
       ExecChecks(
-        (TypeSig.commonCudfTypes + TypeSig.STRUCT + TypeSig.MAP + TypeSig.ARRAY).nested(),
+        (TypeSig.commonCudfTypes + TypeSig.STRUCT + TypeSig.MAP + TypeSig.ARRAY +
+            TypeSig.DECIMAL).nested(),
         TypeSig.all),
       (p, conf, parent, r) => new SparkPlanMeta[BatchScanExec](p, conf, parent, r) {
+        override def tagPlanForGpu(): Unit = {
```
revans2, line 2209: This should not be needed because the Scan should tag itself as not supporting decimal.

sperlingxx: Yes, it is unnecessary. I've removed it.

integration_tests/src/main/python/parquet_test.py
sperlingxx: Done

integration_tests/src/main/python/parquet_test.py
sperlingxx: Deleted.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
jlowe: The problem with removing this code is that the scans **are not** checking for decimal types correctly.  For example, I just tried out this patch on an ORC file with decimals which should not run with the plugin because the ORC reader doesn't support decimals yet.  Note how it says it is going to run the ORC load on the GPU:
```
scala> val df = spark.read.orc("/home/jlowe/tmp/orc-decimals.orc")
df: org.apache.spark.sql.DataFrame = [c_0: decimal(18,0), c_1: decimal(7,3) ... 4 more fields]

scala> df.explain
20/12/18 16:17:04 WARN GpuOverrides: 
*Exec <FileSourceScanExec> could run on GPU

== Physical Plan ==
GpuColumnarToRow false
+- GpuFileGpuScan orc [c_0#54,c_1#55,c_2#56,c_3#57,c_4#58L,c_5#59] Batched: true, DataFilters: [], Format: ORC, Location: InMemoryFileIndex[file:/home/jlowe/tmp/orc-decimals.orc], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<c_0:decimal(18,0),c_1:decimal(7,3),c_2:decimal(10,10),c_3:decimal(15,12),c_4:bigint,c_5:fl...
```

I agree with @revans2 that this change really shouldn't be here, as the scans should be responsible for properly tagging whether they can load the types or not.  However in the past the scans all supported the same types and were relying on the parent exec type checking to check the types.  Now that the scans don't all support the same types, they need to handle the type check explicitly in the separate `tagSupport` methods of `Scan[CSVScan]`, `Scan[OrcScan]`, and `Scan[ParquetScan]`.

Please add the appropriate type checking code in those places along with some fallback tests on CSV and ORC that verify we're not trying to load these on the GPU when decimals are involved.  If we get this wrong it won't crash, and the data will be silently corrupted on load instead.

jlowe:
I also noticed that this PR didn't generate a new [docs/supported_ops.md](https://github.com/NVIDIA/spark-rapids/blob/branch-0.4/docs/supported_ops.md) as it should.  We're not currently generating the documentation for scans from the `ScanMeta`, but it is hardcoded in `SupportedOpsDocs.help` which should be updated accordingly.  After updating the `help` method, performing a usual `mvn verify` build will generate a new `supported_ops.md` file that should be added to this PR.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: Hi @jlowe, I've changed type checking codes of CSV and ORC reader to disable decimal reading. And corresponding test case is also provided.

jlowe (changes requested):
This is looking better, but the supported_ops.md still states at the bottom of the page that Parquet does not support reading the Decimal type.  The static table in TypeChecks needs to be updated to reflect that decimal is support for Parquet input and the supported_ops.md file regenerated by `mvn verify`.

sperlingxx:
build

sperlingxx:
> This is looking better, but the supported_ops.md still states at the bottom of the page that Parquet does not support reading the Decimal type. The static table in TypeChecks needs to be updated to reflect that decimal is support for Parquet input and the supported_ops.md file regenerated by `mvn verify`.

The static table has been updated.

jlowe (approved):

revans2 (approved):

Issue #1203, opened 2020-11-25
[BUG] tpcds query 51 fails with join error on Spark 3.1.0


revans2:
**Describe the bug**
I tried to run query 51 against spark 3.1.0-SNAPSHOT Scale factor 1 and I got 

```
pyspark.sql.utils.IllegalArgumentException: HashJoin should not take FullOuter as the JoinType with building left side
```

This appears to come from 

https://github.com/apache/spark/blob/7cf6a6f996e25754de13aa66badbe6d1d53efb36/sql/core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoin.scala#L73

Which was added as a part of https://issues.apache.org/jira/browse/SPARK-32383

This can be problematic for us because even though the spark join preserves ordering the cudf join does not, so we probably need to stop using HashJoin as a base for our joins entirely.

PR #1303, opened 2020-12-07
Add explicit outputOrdering for BHJ and SHJ in spark310 shim

andygrove:

This fixes the regression with running TPC-DS q51 with Spark 3.1.0 and closes https://github.com/NVIDIA/spark-rapids/issues/1203

### DIFF
--- a/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/GpuBroadcastHashJoinExec.scala
+++ b/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/GpuBroadcastHashJoinExec.scala
@@ -22,7 +22,7 @@ import com.nvidia.spark.rapids.shims.spark301.GpuBroadcastExchangeExec
 
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.catalyst.expressions.{Expression, SortOrder}
 import org.apache.spark.sql.catalyst.optimizer.{BuildLeft, BuildRight, BuildSide}
 import org.apache.spark.sql.catalyst.plans.JoinType
 import org.apache.spark.sql.catalyst.plans.physical.{BroadcastDistribution, Distribution, UnspecifiedDistribution}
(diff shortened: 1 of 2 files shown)


andygrove:
build

tgravescs:
title supposed to be 3.1.0?

andygrove:
> title supposed to be 3.1.0?

Yes, thanks.

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/GpuBroadcastHashJoinExec.scala
```diff
@@ -109,6 +109,10 @@ case class GpuBroadcastHashJoinExec(
     "joinTime" -> SQLMetrics.createNanoTimingMetric(sparkContext, "join time"),
     "filterTime" -> SQLMetrics.createNanoTimingMetric(sparkContext, "filter time"))
 
+  // Spark 3.1.0 started introducing ordering guarantees for hash joins but GPU
+  // hash joins do not provide ordering guarantees
+  override def outputOrdering: Seq[SortOrder] = Nil
```
jlowe, line 114: Is it OK for us to replace a node using stricter ordering guarantees with one that is less conformant?  I don't remember whether the ordering constraint pass has occurred before or after the plugin modifies the plan.

revans2: The ordering fixup happens after we replace the plan.

https://github.com/NVIDIA/spark-rapids/blob/4cb4f5bd01043828953762ebfeec5e6db8c57769/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala#L2278-L2279

jlowe: Thanks, @revans2!

jlowe (approved):

revans2 (commented):
This looks OK to me, but I was hoping for perhaps a larger change.  We currently have two copies of `GpuHashJoin` mostly because the parent `HashJoin` we get from Spark changed in 3.1.0.  I would almost rather that we remove inheriting from `HashJoin` and pull in what we need which should hopefully reduce the total code size and insulate us from these types of issues in the future.

andygrove:
OK, I can work on a follow up PR to remove the HashJoin inheritance. I am definitely in favor of this because our hash join operators extend SparkPlan via their base classes as well as via HashJoin and this caused some issues when i was working on this PR.

revans2:
> OK, I can work on a follow up PR to remove the HashJoin inheritance. I am definitely in favor of this because our hash join operators extend SparkPlan via their base classes as well as via HashJoin and this caused some issues when i was working on this PR.

+1

Issue #1308, opened 2020-12-07
[BUG] TPC-DS query 14a runs much slower on 0.3


jlowe:
**Describe the bug**
Compared to 0.2, TPC-DS query 14a is running over 2X slower at scale factor 1TB.

**Steps/Code to reproduce bug**
Run TPC-DS query 14a with decimals and scale factor 1TB with AQE=off and UCX=off against the plugin version 0.2 and 0.3

**Expected behavior**
0.3 performance should be at least as good as 0.2.

jlowe:
This was fixed by #1310

PR #1310, opened 2020-12-08
Fix canonicalization of GpuFileSourceScanExec, GpuShuffleCoalesceExec

jlowe:
At least part of the issue in #1308 is caused by parts of the query being executed multiple times redundantly because Spark is unable to determine that subsections of the query are identical.  This is caused by broken canonicalization where a GPU Exec node is not comparing correctly with a semantically-equivalent copy of itself.

In this case it's the use of a `RapidsConf` instance as a parameter of the case class.  `RapidsConf` doesn't have an `equals` method, so two instances will be considered different.  However I don't think adding an `equals` method is the real fix here, as the configs could be slightly different in ways unrelated to the semantics of the node's execution and therefore should still be considered equal with respect to that node's canonicalization.

In the case of `GpuShuffleCoalesceExec` it was easy, there was only one config value being used, so I replaced the conf parameter with the value from the conf.  For `GpuFileSourceScanExec` it is trickier since there are many config parameters being used, and it already has quite a few arguments.  In this case I opted to place the conf instance in a separate parameter list which is excluded from the automatically generated equals and hashcode comparison which is used by Spark plan canonicalization.  In our case, there are no RAPIDS config settings that change the semantics for what data is read and how it is presented, so we can safely ignore any RAPIDS configs for purposes of canonicalization.

### DIFF
--- a/shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
+++ b/shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
@@ -182,8 +182,7 @@ class Spark300Shims extends SparkShims {
               wrapped.optionalBucketSet,
               None,
               wrapped.dataFilters,
-              wrapped.tableIdentifier,
-              conf)
+              wrapped.tableIdentifier)(conf)
           }
         }),
       GpuOverrides.exec[SortMergeJoinExec](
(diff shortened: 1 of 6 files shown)


jlowe:
build

tgravescs (approved):

revans2 (approved):

jlowe:
This has a bug where an extra argument was added to `GpuFileSourceScanExec` but `otherCopyArgs` was not overridden.  It can fail queries where the node needs to be copied after it's inserted into the plan.  I'll post a followup PR.

Issue #1289, opened 2020-12-05
Unused code needs to be removed from ParquetCachedBatchSerializer


razajafri:
Unused.

_Originally posted by @jlowe in https://github.com/NVIDIA/spark-rapids/pull/1265#discussion_r536434190_

Issue #1288, opened 2020-12-05
Rename variable in ParquetCachedBatchSerializer to `splitVectors`


razajafri:
Was this supposed to be named `splitVectors`?

_Originally posted by @jlowe in https://github.com/NVIDIA/spark-rapids/pull/1265#discussion_r536440126_

Issue #1287, opened 2020-12-05
Fix possible leak in ParquetCachedBatchSerializer


razajafri:
This seems to be leaked if there's an exception.

_Originally posted by @jlowe in https://github.com/NVIDIA/spark-rapids/pull/1265#discussion_r536441720_

Issue #1286, opened 2020-12-05
ParquetCachedBatchSerializer allocates a bunch of Tables upfront. 


razajafri:
This has the drawback of creating a bunch of leakable resources up front. I believe this will be easier to maintain and read if we construct the one table we need to perform the write then close that table rather than creating them all up-front first.  The code that writes a single table can be refactored into a separate method so it's easy to call from this split case and the non-split case rather than needing to fall-through to common-code.

_Originally posted by @jlowe in https://github.com/NVIDIA/spark-rapids/pull/1265#discussion_r536442577_

kuhushukla:
May I suggest renaming the pr title to :
`ParquetCachedBatchSerializer allocates a bunch of Tables upfront. `

PR #1311, opened 2020-12-08
ParquetCachedBatchSerializer code cleanup

razajafri:
This PR addresses the follow-on items opened as a result of #1265. I have made 0.4 the target branch.

#1285
#1286
#1287
#1288
#1289

### DIFF
--- a/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
+++ b/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
@@ -340,8 +340,6 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
 
   private[rapids] def compressColumnarBatchWithParquet(
      gpuCB: ColumnarBatch): List[ParquetCachedBatch] = {
-    // NOTE: this doesn't take nulls into account so we could be over-estimating the actual size
-    // but that's still better than under-estimating
     val estimatedRowSize = scala.Range(0, gpuCB.numCols()).map { idx =>
       gpuCB.column(idx).asInstanceOf[GpuColumnVector].getBase.getDeviceMemorySize / gpuCB.numRows()
     }.sum
@@ -349,8 +347,20 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
     val splitIndices = scala.Range(rowsInBatch, gpuCB.numRows(), rowsInBatch)
     val buffers = new ListBuffer[ParquetCachedBatch]
     if (splitIndices.nonEmpty) {
-      val splitVectors = scala.Range(0, gpuCB.numCols()).map { index =>
-        gpuCB.column(index).asInstanceOf[GpuColumnVector].getBase.split(splitIndices: _*)
+      val splitVectors = new ListBuffer[Array[ColumnVector]]
+      for (index <- 0 until gpuCB.numCols()) {
+        try {
+          splitVectors +=
+            gpuCB.column(index).asInstanceOf[GpuColumnVector].getBase.split(splitIndices: _*)
+        } catch {
+          case _ => splitVectors.foreach { array =>
+            array.foreach { vector =>
+              if (vector != null) {
+                vector.close()
+              }
+            }
+          }
+        }
       }
       try {
         // Splitting the table
@@ -359,12 +369,16 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
         // T01= {splitCol1(1), splitCol2(1),...,splitColn(1)}
         // ...
         // T0m= {splitCol1(m), splitCol2(m),...,splitColn(m)}
-        for (i <- splitVectors(0).indices)
-          withResource(new Table({
-            for (j <- splitVectors.indices) yield splitVectors(j)(i)
-          }: _*)) { table =>
+        def makeTableForIndex(i: Int): Table = {
+          val columns = splitVectors.indices.map(j => splitVectors(j)(i))
+          new Table(columns :_*)
+        }
+
+        for (i <- splitVectors(0).indices) {
+          withResource(makeTableForIndex(i)) { table =>
             buffers += ParquetCachedBatch(writeTableToCachedBatch(table))
           }
+        }
       } finally {
         for (i <- splitVectors(0).indices)
           for (j <- splitVectors.indices) splitVectors(j)(i).close()
--- a/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
+++ b/shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
@@ -348,24 +348,12 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
     val buffers = new ListBuffer[ParquetCachedBatch]
     if (splitIndices.nonEmpty) {
       val splitVectors = new ListBuffer[Array[ColumnVector]]
-      for (index <- 0 until gpuCB.numCols()) {
-        try {
+      try {
+        for (index <- 0 until gpuCB.numCols()) {
           splitVectors +=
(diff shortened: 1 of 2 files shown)


razajafri:
build

razajafri:
build

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -353,13 +342,14 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
      gpuCB: ColumnarBatch): List[ParquetCachedBatch] = {
     // NOTE: this doesn't take nulls into account so we could be over-estimating the actual size
```
revans2, line 343: This comment is no longer accurate

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -353,13 +342,14 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
-    val estimatedRowSize = scala.Range(0, gpuCB.numCols()).map { index =>
-      gpuCB.column(index).dataType().defaultSize
+    val estimatedRowSize = scala.Range(0, gpuCB.numCols()).map { idx =>
+      gpuCB.column(idx).asInstanceOf[GpuColumnVector].getBase.getDeviceMemorySize / gpuCB.numRows()
     }.sum
     val rowsInBatch = (BYTES_ALLOWED_PER_BATCH / estimatedRowSize).toInt
     val splitIndices = scala.Range(rowsInBatch, gpuCB.numRows(), rowsInBatch)
-    val cudfTables = if (splitIndices.nonEmpty) {
-      val splutVectors = scala.Range(0, gpuCB.numCols()).map { index =>
+    val buffers = new ListBuffer[ParquetCachedBatch]
```
revans2, line 350: buffers needs a `closeOnExcept` or equivalent to not leak on excption.

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -353,13 +342,14 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
+    if (splitIndices.nonEmpty) {
+      val splitVectors = scala.Range(0, gpuCB.numCols()).map { index =>
```
revans2, line 352: If a failure happens in split, everything that has already succeeded for split will leak.  What is more because this is a map `splitVectors` is not actually materialized until you start to use it, and then I am not sure how much of it will be materialized at a time.

I would personally prefer to have a mutable data structure that we populate in a for loop, that has the equivalent of a `withResources` or a `closeOnExcept`.

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -369,31 +359,33 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
         // T01= {splitCol1(1), splitCol2(1),...,splitColn(1)}
         // ...
         // T0m= {splitCol1(m), splitCol2(m),...,splitColn(m)}
-        for (i <- splutVectors(0).indices) yield
-          new Table({
-            for (j <- splutVectors.indices) yield splutVectors(j)(i)
-          }: _*)
+        for (i <- splitVectors(0).indices)
+          withResource(new Table({
```
revans2, line 363: There is no leak here, but it is rather confusing.  Could we make either creating the table or pulling the columns out for the table into a separate function so it is cleaner?

```
def makeTableForIndex(i: Int): Table = {
  val columns = splitVectors.indices.map(j => splitVectors(j)(i).toArray
  new Table(columns :_*)
}

for (i <- splitVectors(0).indices) {
  withResource(makeTableForIndex(i)) { table =>
    buffers += ParquetCachedBatch(writeTableToCachedBatch(table))
  }
}
```

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
razajafri: hmm... not sure if I am missing something here. The batches have data on the Heap, so in case of an Exception, won't they be reclaimed by jvm?

revans2: Sorry I got confused by what needed to be closed and what did not.

razajafri:
build

razajafri:
@revans2 have I addressed all your concerns?

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
```diff
@@ -351,16 +340,27 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
+      for (index <- 0 until gpuCB.numCols()) {
+        try {
+          splitVectors +=
+            gpuCB.column(index).asInstanceOf[GpuColumnVector].getBase.split(splitIndices: _*)
+        } catch {
+          case _ => splitVectors.foreach { array =>
+            array.foreach { vector =>
+              if (vector != null) {
+                vector.close()
+              }
```
razajafri, line 360: I am not sure this is correct. I almost want to rethrow the exception here so the caller knows things went wrong. Or maybe throw another exception?

razajafri:
```
        try {
          splitVectors +=
            gpuCB.column(index).asInstanceOf[GpuColumnVector].getBase.split(splitIndices: _*)
        } catch {
          case e => splitVectors.foreach { array =>
            array.foreach { vector =>
              if (vector != null) {
                vector.close()
              }
            }
          }
            // the caller will want to know there was an exception but won't be able to close the
            // splitVectors
            throw e
        }
```

razajafri: I confused myself with the log throw anti-pattern. I have made the above mentioned fix

razajafri:
build

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/ParquetCachedBatchSerializer.scala
revans2: The code how you have it looks correct to me.  You should re-throw the exception in the case of an error. The indentation appears to be off but that is minor.

As an alternative because when you create a table it will increment the reference count for the columns passed in, you could just use a try/finally block for splitVectors.

```
val splitVectors =  new ListBuffer[Array[ColumnVector]]
try {
  for (index <- 0 until gpuCB.numCols()) {
    splitVectors +=
        gpuCB.column(index).asInstanceOf[GpuColumnVector].getBase.split(splitIndices: _*)
  }

  def makeTableForIndex(i: Int): Table = {
    val columns = splitVectors.indices.map(j => splitVectors(j)(i))
    new Table(columns :_*)
  }

  for (i <- splitVectors(0).indices) {
    withResource(makeTableForIndex(i)) { table =>
      buffers += ParquetCachedBatch(writeTableToCachedBatch(table))
    }
  }
} finally {
  // Not 100% sure the safeClose intrinsic works here
  splitVectors.foreach(array => array.safeClose())
}
```

razajafri:
@revans2 the code looks cleaner now. Thanks. Do you have anything else?

razajafri:
build

revans2 (approved):

jlowe (approved):

Issue #140, opened 2020-06-10
[BUG] Orc writer wrong for timestamps prior to 1970


revans2:
**Describe the bug**
When writing timestamps prior to 1970 and reading them back in using the CPU the results are wrong.

**Steps/Code to reproduce bug**
An integration test has been added for this.

ttnghia:
Is this still relevant? It seems we just had a fixed in https://github.com/rapidsai/cudf/issues/11525 but there is still https://github.com/rapidsai/cudf/issues/11691.

PR #1312, opened 2020-12-08
fix orc test

pxLi:
Signed-off-by: Peixin Li <peixinl@nvidia.com>

I cannot reproduce on ubuntu env. But I can easily reproduce this on integration tests w/ our integration centos image:
1. [test_read_round_trip-fail](https://blossom.nvidia.com/sw-gpu-spark-jenkins/view/Testing/job/peixin-orc_test/82/console)
2. [test_simple_partitioned_read-fail](https://blossom.nvidia.com/sw-gpu-spark-jenkins/view/Testing/job/peixin-orc_test/85/console)

If I increase the `time start` a bit larger than the epoch (1970.1.1, not the same as since that would bring the weighted special cases) , then the test works well. Not sure what's the root cause here. 

This PR does not need to be merged if anyone else has better context about this, related to https://github.com/NVIDIA/spark-rapids/issues/140

### DIFF
--- a/integration_tests/src/main/python/orc_test.py
+++ b/integration_tests/src/main/python/orc_test.py
@@ -39,7 +39,7 @@ def test_basic_read(std_input_path, name, read_func, v1_enabled_list, orc_impl):
 
 orc_gens_list = [[byte_gen, short_gen, int_gen, long_gen, float_gen, double_gen,
     string_gen, boolean_gen, DateGen(start=date(1590, 1, 1)),
-    TimestampGen(start=datetime(1590, 1, 1, tzinfo=timezone.utc))],
+    TimestampGen(start=datetime(1970, 1, 1, 0, 0, 1, 0, tzinfo=timezone.utc))],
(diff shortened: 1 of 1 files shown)


pxLi:
build

jlowe:
@nvdbaranec has been trying to reproduce this in an interactive environment without success.  @pxLi if you can reproduce this easily, it would be great if you could work with Dave to help setup a repro case for him to debug.  I don't want to sweep this bug under the rug, which is what this change wants to do.

pxLi:
> @nvdbaranec has been trying to reproduce this in an interactive environment without success. @pxLi if you can reproduce this easily, it would be great if you could work with Dave to help setup a repro case for him to debug. I don't want to sweep this bug under the rug, which is what this change wants to do.

Yeah, I provided him the env locally to reproduce yesterday. this PR is just a workaround which was intended to be taken as a real bug fix.

jlowe:
Based on @nvdbaranec's investigations, this appears to be an issue with the GMT timezone in the CentOS environment.  As such I think the approach in #1337 is more appropriate since we don't need to unnecessarily restrict the range of timestamps being tested.

pxLi:
sure, let me close this one

PR #1313, opened 2020-12-08
Support spilling to disk directly via cuFile/GDS

rongou:
This adds an option to spill to disk directly via GDS. Default is off and shouldn't have any effect on existing behavior.

Initial test on my local desktop shows that it performs similarly to spilling to host memory, with a single Samsung 970 EVO Plus 2TB NVMe M.2 drive. Need to do more benchmarking in real deployment environments.

Depends on https://github.com/rapidsai/cudf/pull/6940.


revans2 (commented):
My only comment on this is that either we need to hide the config, or we need to document it better and explain it is an experimental feature and point to more information about GDS if we want others to try it out.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
```diff
@@ -152,7 +152,7 @@ class BufferSendState(
             requestHandler.acquireShuffleBuffer(bufferId)) { rapidsBuffer =>
             //these are closed later, after we synchronize streams
             rapidsBuffer.storageTier match {
-              case StorageTier.DEVICE =>
+              case StorageTier.DEVICE | StorageTier.GDS =>
```
abellina, line 155: I think instead of doing this, we should have `BufferSendState` have a special case for GDS buffers, then have it call `CuFile.copyFileToDeviceBuffer` with a given offset (indirectly). That way we remove an extra copy and we don't need to allocate the full GDS buffer, in an already constrained device.

If you have a large GDS file this would allow you to, on the read path, to chunk those reads at most bounce buffer length also. Not sure if this has implications to GDS (several read requests, vs one large one, and what should the ideal request size be)

abellina: I think that for an initial implementation it may be fine the way it is, but we may OOM more often, in the disk spill case since now we need more GPU memory in the interim.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -0,0 +1,128 @@
...
+class RapidsGdsStore(
+    diskBlockManager: RapidsDiskBlockManager,
+    catalog: RapidsBufferCatalog = RapidsBufferCatalog.singleton)
+    extends RapidsBufferStore("gds", catalog) {
+  private[this] val sharedBufferFiles = new ConcurrentHashMap[RapidsBufferId, File]
+
+  override def createBuffer(
+      other: RapidsBuffer,
+      stream: Cuda.Stream): RapidsBufferBase = {
+    val otherBuffer = other.getMemoryBuffer
```
abellina, line 40: `withResource(other.getMemoryBuffer) { otherBuffer => `

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -0,0 +1,128 @@
...
+        case d: DeviceMemoryBuffer => d
+        case _ => throw new IllegalStateException("copying from buffer without device memory")
+      }
+      val id = other.id
+      val path = if (id.canShareDiskPaths) {
+        sharedBufferFiles.computeIfAbsent(id, _ => id.getDiskPath(diskBlockManager))
+      } else {
+        id.getDiskPath(diskBlockManager)
+      }
+      val fileOffset = if (id.canShareDiskPaths) {
```
abellina, line 52: maybe a comment here would help with `copyDeviceBufferToFile`'s last argument. Without going to cudf to look there, I am not sure why we need to fork here.

abellina (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -0,0 +1,128 @@
...
+
+    private def getDeviceBuffer: DeviceMemoryBuffer = synchronized {
+      if (deviceBuffer.isEmpty) {
+        val path = if (id.canShareDiskPaths) {
+          sharedBufferFiles.get(id)
+        } else {
+          id.getDiskPath(diskBlockManager)
+        }
+        val buffer = DeviceMemoryBuffer.allocate(size);
+        CuFile.copyFileToDeviceBuffer(buffer, path, fileOffset)
```
abellina, line 85: Could `CuFile.copyFileToDeviceBuffer` fail? If so we should wrap using `closeOnExcept`:
```
closeOnExcept(DeviceMemoryBuffer.allocate(size)) { buffer => 
   CuFile.copyFileToDeviceBuffer(buffer, path, fileOffset)
}
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
rongou: Yes probably need more tuning down the road.

rongou:
@revans2 added wording to the config to specify it's an experimental feature, and link to gds docs.

revans2 (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -0,0 +1,127 @@
...
+      fileOffset: Long,
+      size: Long,
+      meta: TableMeta,
+      spillPriority: Long) extends RapidsBufferBase(id, size, meta, spillPriority) {
+    private[this] var deviceBuffer: Option[DeviceMemoryBuffer] = None
+
+    override val storageTier: StorageTier = StorageTier.GDS
+
+    private def getDeviceBuffer: DeviceMemoryBuffer = synchronized {
+      if (deviceBuffer.isEmpty) {
```
jlowe, line 76: Trying to cache the GDS data in GPU memory via a GDS store buffer seems problematic to me.  What happens if GPU memory pressure pushes out all of the device memory buffers into the GDS store, then these buffers get referenced back into GPU memory?  When the next OOM occurs, none of these GDS buffers in GPU memory are spillable, as they're not tracked in the device memory store.

If we want to cache these buffers in GPU memory then I think we need to implement reverse-spill logic where these cached buffers are tracked in the device store.  Then when we try to spill from the device store we can prioritize "spilling" buffers that have already been spilled to GDS and trivially release them.  That is much more involved, so probably the first step is to copy the data back into a temporary, non-cached buffer.  Not great for handling spilled broadcast data, but should work OK for the typical case of spilled shuffle partitions.  We can implement the reverse-spill logic in a separate PR.

jlowe (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
rongou: Done.

jlowe (approved):

jlowe:
build

rongou:
build

pxLi:
build

pxLi:
build

abellina (approved):

Issue #750, opened 2020-09-11
[BUG] udf_cudf_test::test_with_column fails with IPC error 


firestarman:
**Describe the bug**
When running udf_cudf test with docker (centos7 cuda10.1), the operation with column (like plus one to one column) will throw 'OSError: Invalid IPC stream: negative continuation token'
**Steps/Code to reproduce bug**
1. enable test_with_column in udf_cudf_test.py
2. run the test case normally with pytest.
**Expected behavior**
The test will fail with exception:
'OSError: Invalid IPC stream: negative continuation token'
**Environment details (please complete the following information)**
 - Environment location: Docker (centos7 cuda 10.1) with standalone spark in it.
 - Spark configuration settings: need to enable udf support (spark.rapids.sql.python.gpu.enabled=true)
**Additional context**
All the settings of spark are already in test cases, you could use them directly.

sameerz:
Marking as P2, @GaryShen2008 please confirm whether this test is required for the pandas cudf functionality.

firestarman:
No more IPC error with the rapids jar from Sep, just hangs forever instead.

firestarman:
Re-open due to not root cause. See PR 1343 https://github.com/NVIDIA/spark-rapids/pull/1343

sameerz:
@firestarman can we move this to release 0.4?

firestarman:
> @firestarman can we move this to release 0.4?

Yes please.

firestarman:
I am closing this issue, since i found the root cause and[ the PR 1343](https://github.com/NVIDIA/spark-rapids/pull/1343) should be the best fix on the Spark plugin side.

The error is because GPU will send nothing about the arrow data to the python side when it is empty partition, but CPU still sends the `begin` and `end ` arrow packages to Python.
The [actual arrow writer inside a cuDF writer is created only in the method `write` ](https://github.com/rapidsai/cudf/blob/branch-0.18/java/src/main/native/src/TableJni.cpp#L230), but the method `write` will not be called when the partition data is empty. So nothing is sent to Python, even the `begin` and `end ` arrow package. Meanwhile the Python process expects at least the `begin` and `end ` arrow package, so it throws the IPC error due to negative continuation token.

PR #1235, opened 2020-12-02
Trigger reading and handling control data.

firestarman:
Trigger reading and handling control data by calling the API `hasNext` of the output iterator from GPU python runner after the queue becomes empty.

Because per the protocol, there is some control data after the stream data from python processes. When the queue becomes empty, we only read all the stream data, but ignore the control data. Seems we did not shut down the stream correctly even it works.

Signed-off-by: Firestarman <firestarmanllc@gmail.com>

PR #1343, opened 2020-12-09
Do not run UDFs when the partition is empty.

firestarman:
When the partition is empty, do not launch the Python runners and Python processes which are expensive and useless, and return the input iterator directly instead.

Filed this PR to fix the udf related failures in IT tests due to IPC error. This is not a fix for the root cause, but it still makes sense. 

It looks like the cuDF 'ArrowIPCTableWriter' does not handle the empty partiton case correctly. So the Python side reports the IPC error. 

The IPC error should have been there for a while, however we just ignored it due to not read the control data before. But now is exposed by the PR https://github.com/NVIDIA/spark-rapids/pull/1235. Anyway will continue investigating it.

Also reduce the data size for cudf tests to save running time.

### DIFF
--- a/integration_tests/src/main/python/udf_cudf_test.py
+++ b/integration_tests/src/main/python/udf_cudf_test.py
@@ -45,7 +45,7 @@ _conf = {
 
 small_data = [(1, 1.0), (1, 2.0), (2, 3.0), (2, 5.0), (2, 10.0)]
 
-large_data = list(map(lambda i: (i, i/1.0), range(1, 5000))) * 2
+large_data = list(map(lambda i: (i, i/1.0), range(1, 512))) * 2
 
 
 def _create_df(spark, data=large_data):
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuArrowEvalPythonExec.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuArrowEvalPythonExec.scala
@@ -613,54 +613,60 @@ case class GpuArrowEvalPythonExec(
         PythonWorkerSemaphore.acquireIfNecessary(context)
       }
 
-      val pyRunner = new GpuArrowPythonRunner(
-        pyFuncs,
-        evalType,
-        argOffsets,
-        pythonInputSchema,
-        timeZone,
-        runnerConf,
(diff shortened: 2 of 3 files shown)


firestarman:
build

integration_tests/src/main/python/udf_cudf_test.py
```diff
@@ -45,7 +45,7 @@
 
 small_data = [(1, 1.0), (1, 2.0), (2, 3.0), (2, 5.0), (2, 10.0)]
 
-large_data = list(map(lambda i: (i, i/1.0), range(1, 5000))) * 2
+large_data = list(map(lambda i: (i, i/1.0), range(1, 512))) * 2
```
revans2, line 48: This change appears to be unrelated to running UDFs with an empty partition.  Is there some reason I am missing?

revans2 (commented):
In regular Spark what does an empty iterator do?  Does it send to data to python and get no data back?

integration_tests/src/main/python/udf_cudf_test.py
firestarman: Updated the PR description for this change, I just explained it in the commit message only before. 
This change is to save the cudf tests execution time, especially for databricks, and it is unrelated to the IPC error.

firestarman:
> In regular Spark what does an empty iterator do? Does it send to data to python and get no data back?

Per the code [here](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/python/ArrowPythonRunner.scala#L88), no data will be sent to the Python side and I observed the UDF will not run for empty partitions.
But you can see the arrow writer still executes the `start `and `end `, similar with our GPU arrow writer's `constructor `and `close`. But ours seems did not send the right metadata for some reason, i am not sure. So the python side reads an illegal value.

revans2 (approved):
We might want a follow on issue to address the root cause, but this is fine for now.

firestarman:
> We might want a follow on issue to address the root cause, but this is fine for now.

yes, i have re-opened this issue https://github.com/NVIDIA/spark-rapids/issues/750

firestarman:
Closed the issue  https://github.com/NVIDIA/spark-rapids/issues/750 since I found  the root cause, pls go to the issue for details.

PR #1349, opened 2020-12-09
Add canonicalization tests

andygrove:
This PR updates the SparkQueryCompareTestSuite to plan each query twice and check if the canonicalized plan is the same or not both on CPU and GPU and fail if the behavior is not consistent between CPU and GPU.

I have skipped this check for 4 tests that are failing due to what looks like a bug in Spark, where the AttributeReference for the aggregate function result has different exprId values each time for SortAggregate but not for HashAggregate.

I filed https://issues.apache.org/jira/browse/SPARK-33744 for the Spark bug.


jlowe:
One potential issue we could have with this approach is when the provided `fun` is not idempotent, e.g.: it's doing a table create.

jlowe (approved):

andygrove:
build

andygrove:
build

revans2 (approved):

jlowe (approved):

PR #1350, opened 2020-12-09
fix GpuRangePartitioning canonicalization

andygrove:

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRangePartitioning.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRangePartitioning.scala
@@ -40,9 +40,10 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
 case class GpuRangePartitioning(
     gpuOrdering: Seq[SortOrder],
     numPartitions: Int,
-    part: GpuRangePartitioner,
-    schema: StructType)
+    schema: StructType)(val part: GpuRangePartitioner)
   extends GpuExpression with GpuPartitioning {
+
+  override def otherCopyArgs: Seq[AnyRef] = Seq(part)
 
   var rangeBounds: Array[InternalRow] = _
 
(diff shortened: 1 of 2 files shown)


jlowe (approved):

jlowe:
build

kuhushukla (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRangePartitioning.scala
```diff
@@ -40,10 +40,11 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
 case class GpuRangePartitioning(
     gpuOrdering: Seq[SortOrder],
     numPartitions: Int,
-    part: GpuRangePartitioner,
-    schema: StructType)
+    schema: StructType)(val part: GpuRangePartitioner)
```
revans2, line 43: Why even pass in GpuRangePartitioner?  It is not used anywhere but here.  There is no options to pass in anything else.  It does not get reused.  Why not make it a value withing the class?  Or we could even make it a trait that GpuRangePartitioning inherits from.

revans2: Just to be clear I am fine with this as is.  Just saw an opportunity to possibly clean it up even more.

jlowe: I was thinking the same thing but wasn't familiar enough with the range partitioning code to be sure.  Glad you were thinking the same thing.  I'm cool either way, as long as we're confident in the change to ship 0.3 ASAP afterwards.

revans2: Then lets merge this as is.  I don't want to hold up 0.3

jlowe: :+1: sounds good.  @andygrove would you mind filing the followup to clean this up in a later release?

andygrove: Filed as https://github.com/NVIDIA/spark-rapids/issues/1355

PR #1356, opened 2020-12-09
[auto-merge] branch-0.3 to branch-0.4 [skip ci] [bot]

nvauto:
auto-merge triggered by github actions on `branch-0.3` to create a PR keeping `branch-0.4` up-to-date. If this PR is unable to be merged due to conflicts, it will remain open until manually fix.

PR #1358, opened 2020-12-10
auto-merge: instant merge after creation [skip ci]

pxLi:

issue from an auto-merge PR: https://github.com/NVIDIA/spark-rapids/pull/1356#issuecomment-742175119

### DIFF
--- a/.github/workflows/auto-merge/automerge
+++ b/.github/workflows/auto-merge/automerge
@@ -115,7 +115,6 @@ def main():
     if term:
         sys.exit(0)
 
-    time.sleep(10)  # sleep 10s, then try auto-merge
     auto_merge(number, sha)
 
 


pxLi:
build

jlowe (approved):

jlowe:
I think removing the 10 second wait (which I never understood the purpose) is a reasonable workaround to reduce the window of opportunity, but the window is still there.  I think a better solution is changing the workflow to recognize the merge failed because the SHA changed because new commits were pushed so it can automatically loopback and retry.

So I'm OK with this going in to make it less likely to occur, but I think it could be improved to eliminate this type of error entirely.

pxLi:
> I think removing the 10 second wait (which I never understood the purpose) is a reasonable workaround to reduce the window of opportunity, but the window is still there. I think a better solution is changing the workflow to recognize the merge failed because the SHA changed because new commits were pushed so it can automatically loopback and retry.
> 
> So I'm OK with this going in to make it less likely to occur, but I think it could be improved to eliminate this type of error entirely.

yeah, totally agree. since the github API does not give the specific reason for us to identify detailed merge failures, so we may need to add some custom logic in our code. I just created https://github.com/NVIDIA/spark-rapids/issues/1363 for a future fix

PR #1361, opened 2020-12-10
Use the correct precision when reading spark columnar data.

### LINKED ISSUE #1348: [BUG] Host columnar decimal conversions are failing
As reported in #1279 and #1280 there are issues converting columnar host decimals to GPU decimals.  Those issues were "fixed" by disabling the decimal type by default, but the root cause needs to be investigated and fixed.

revans2:
I didn't add any tests, but I did manually run all of TPC-DS scale factor 1 on a decimal parquet data set, which had a lot of failures before this and now all of them pass.

this fixes #1348


revans2:
build

jlowe (approved):

PR #1367, opened 2020-12-10
Reuse gpu expression conversion rules when checking sort order

### LINKED ISSUE #1305: [FEA] Investigate potential GPU expressions of sort order
#1302 fixed an issue where `GpuSubstring` appeared as a child of `SortOrder` and not being canonicalized.  We need to investigate if there are other types of expressions that may need to be canonicalized properly besides the small list currently covered.

If there are many expressions we may need to rethink how to approach this issue so it isn't handled on a case-by-case basis.

revans2:
This fixes #1305 

In this patch instead of trying to bring everything back to the CPU to check if a sort order is satisfied, we selectively decide if we need to bring them both to the GPU to do it.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -401,42 +401,51 @@ object GpuOverrides {
     !regexList.exists(pattern => s.contains(pattern))
   }
 
-  private def canonicalizeToCpuForSortOrder(exp: Expression): Expression = exp match {
-    case g: GpuLiteral => Literal(g.value, g.dataType).canonicalized
-    case g: GpuKnownFloatingPointNormalized =>
-      KnownFloatingPointNormalized(canonicalizeToCpuForSortOrder(g.child)).canonicalized
-    case g: GpuNormalizeNaNAndZero =>
-      NormalizeNaNAndZero(canonicalizeToCpuForSortOrder(g.child)).canonicalized
-    case g: GpuAlias =>
-      ShimLoader.getSparkShims.alias(canonicalizeToCpuForSortOrder(g.child), g.name)(
-        g.exprId,
-        g.qualifier,
-        g.explicitMetadata)
-          .canonicalized
-    case g: GpuSubstring =>
-      Substring(
-        canonicalizeToCpuForSortOrder(g.str),
-        canonicalizeToCpuForSortOrder(g.pos),
-        canonicalizeToCpuForSortOrder(g.len))
-          .canonicalized
-    case o: GpuExpression =>
-      throw new IllegalStateException(s"${o.getClass} is not expected to be a part of a SortOrder")
-    case other => other.canonicalized
+  private def convertExprToGpuIfPossible(expr: Expression, conf: RapidsConf): Option[Expression] = {
+    val wrapped = wrapExpr(expr, conf, None)
+    wrapped.tagForGpu()
+    if (wrapped.canExprTreeBeReplaced) {
+      Some(wrapped.convertToGpu())
+    } else {
+      None
+    }
   }
 
-  private def gpuOrderingSemanticEquals(found: Expression, required: Expression): Boolean = {
-    found.deterministic && required.deterministic &&
-        canonicalizeToCpuForSortOrder(found) == canonicalizeToCpuForSortOrder(required)
+  private def gpuOrderingSemanticEquals(
+      found: Expression,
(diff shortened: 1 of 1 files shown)


revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -401,42 +401,51 @@ object GpuOverrides {
...
+  private def gpuOrderingSemanticEquals(
+      found: Expression,
+      required: Expression,
+      conf: RapidsConf): Boolean = {
+    if (found.deterministic && required.deterministic) {
+      val foundOnGpu = found.find(_.isInstanceOf[GpuExpression]).isDefined
+      val requiredOnGpu = required.find(_.isInstanceOf[GpuExpression]).isDefined
+      if (foundOnGpu == requiredOnGpu) {
+        found.canonicalized == required.canonicalized
+      } else if (foundOnGpu) {
```
jlowe, line 423: I think this would be simpler if we had the conversion do a best-effort conversion rather than an optional result.  That avoids the three cases that need to be handled, e.g.:
```scala
  if (found.deterministic && required.deterministic) {
    convertToGpuIfPossible(found, conf).canonicalized == convertToGpuIfPossible(required, conf).canonicalized
  }
```

and `convertToGpuIfPossible` returns an `Expression` that may or may not have been converted to a `GpuExpression`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -401,42 +401,51 @@ object GpuOverrides {
...
 
-  private def gpuOrderingSemanticEquals(found: Expression, required: Expression): Boolean = {
-    found.deterministic && required.deterministic &&
-        canonicalizeToCpuForSortOrder(found) == canonicalizeToCpuForSortOrder(required)
+  private def gpuOrderingSemanticEquals(
+      found: Expression,
+      required: Expression,
+      conf: RapidsConf): Boolean = {
+    if (found.deterministic && required.deterministic) {
+      val foundOnGpu = found.find(_.isInstanceOf[GpuExpression]).isDefined
```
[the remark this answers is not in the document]
revans2: I'm not sure what you mean by this. `AttributeReference` is not a `GpuExpression`, but we use it as a part of a `GpuExpression` tree.  I think `AttributeReference` should be the common case as the child of a `SortOrder`, but I was wrong on other things related to this so I could be wrong on this too.

Also found and required may be coming from different places.  The goal here is to make them the same, either all on the GPU or all off of the CPU so we can properly compare them.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
revans2: I am happy to make the change.  The reason I did it this way was to avoid converting things to the GPU if they were not needed.  It might be premature optimization so if you want me to change it I will.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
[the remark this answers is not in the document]
jlowe: Ah `AttributeReference` is an example where that's not true.  I think it used to be true when we were tagging everything with a GPU version, but that's no longer the case.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
jlowe: I was thinking it could still avoid it using the same `find` approach being done, just moving that into `convertToGpuIfPossible`, e.g.:
```scala
def convertToGpuIfPossible(expr: Expression): Expression = {
  if (expr.find(_.isInstanceOf[GpuExpression]).isDefined) {
    // already been converted
    expr
  } else {
    val wrapped = wrapExpr(expr, conf, None)
    wrapped.tagForGpu()
    if (wrapped.canExprTreeBeReplaced) {
      wrapped.convertToGpu()
    } else {
      expr
    }
  }
}
```

revans2: The one case that misses is if both are on the CPU already.  But it is probably not a big deal, so I'll make the change.

revans2:
build

jlowe (approved):

PR #1376, opened 2020-12-11
MetaUtils.getBatchFromMeta should return batches with GpuColumnVectorFromBuffer

jlowe:
When building a `ColumnarBatch` from a device buffer and shuffle metadata, `MetaUtils` is not returning a batch with `GpuColumnVectorFromBuffer` instances but rather normal `GpuColumnVector` instances.  When these batches are seen by the subsequent `GpuCoalesceBatches` node in the query plan, `GpuCoalesceBatches` mistakenly thinks the columns are not from a contiguous buffer and performs a redundant contiguous split on the batch.


jlowe:
build

revans2 (approved):

abellina (approved):

Issue #1381, opened 2020-12-11
[FEA] Track unsupported datetime patterns instead of tracking patterns that are supported. 


razajafri:
**Is your feature request related to a problem? Please describe.**
In datetimeExpressions.scala we currently have a list of datetime patterns that we support, which is by no means complete. It will be quite a lot of work listing all the possible combinations of patterns we support

**Describe the solution you'd like**
We should instead list patterns that we don't currently support, or even better would be to change the DateUtils.scala, which already has the framework in-place to do so

**Describe alternatives you've considered**
NA

**Additional context**
NA

PR #1382, opened 2020-12-11
Handle "MM[/-]dd" and "dd[/-]MM" datetime formats in UnixTimeExprMeta

### LINKED ISSUE #959: [FEA] Audit_3.0.1: This might affect our Casting string to timestamp
commit 4638402d4775acf0d7ff190ef5eabb6f2dc85af5
Author: Kent Yao <yaooqinn@hotmail.com>
Date:   Thu Jun 11 03:29:12 2020 +0000

    [SPARK-31939][SQL][TEST-JAVA11] Fix Parsing day of year when year field pattern is missing

razajafri:
This PR adds the code to handle the datetime pattern "MM/dd" or "dd/MM" and closes #959. I spoke with @andygrove and he agrees that we should instead track the formats that we don't support instead of having an exhaustive list of patterns that we do support. 

That item is being tracked by #1381

### DIFF
--- a/integration_tests/src/main/python/date_time_test.py
+++ b/integration_tests/src/main/python/date_time_test.py
@@ -185,6 +185,8 @@ def test_to_unix_timestamp_improved(data_gen):
 
 str_date_and_format_gen = [pytest.param(StringGen('[0-9]{4}/[01][0-9]'),'yyyy/MM', marks=pytest.mark.xfail(reason="cudf does no checks")),
         (StringGen('[0-9]{4}/[01][12]/[0-2][1-8]'),'yyyy/MM/dd'),
+        (StringGen('[01][12]/[0-2][1-8]'), 'MM/dd'),
+        (StringGen('[0-2][1-8]/[01][12]'), 'dd/MM'),
         (ConvertGen(DateGen(nullable=False), lambda d: d.strftime('%Y/%m').zfill(7), data_type=StringType()), 'yyyy/MM')]
 
 @pytest.mark.parametrize('data_gen,date_form', str_date_and_format_gen, ids=idfn)
@@ -196,5 +198,3 @@ def test_string_to_unix_timestamp(data_gen, date_form):
 def test_string_unix_timestamp(data_gen, date_form):
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : unary_op_df(spark, data_gen, seed=1).select(f.unix_timestamp(f.col('a'), date_form)))
-
-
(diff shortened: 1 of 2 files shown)


revans2:
The code looks OK, but can we make the name of the PR clearer for what it is doing?

razajafri:
> The code looks OK, but can we make the name of the PR clearer for what it is doing?

I have updated the PR title and description to make it more clearer. Does it look OK?

revans2:
looks good

integration_tests/src/main/python/date_time_test.py
```diff
@@ -185,6 +185,8 @@ def test_to_unix_timestamp_improved(data_gen):
 
 str_date_and_format_gen = [pytest.param(StringGen('[0-9]{4}/[01][0-9]'),'yyyy/MM', marks=pytest.mark.xfail(reason="cudf does no checks")),
         (StringGen('[0-9]{4}/[01][12]/[0-2][1-8]'),'yyyy/MM/dd'),
+        (StringGen('[01][12]/[0-2][1-8]'), 'MM/dd'),
+        (StringGen('[0-2][1-8]/[01][12]'), 'dd/MM'),
```
andygrove, line 189: It looks like these tests only generate valid inputs for the formats being tested? If this is the case, it would be good to test against some invalid inputs as well, either here or in the scala `ParseDateTimeSuite` where we have examples of this type of testing.

andygrove: I tested this locally and it looks good. I added the following tests to `ParseDateTimeSuite` to test these new formats against the existing test data.

```scala
testSparkResultsAreEqual("to_unix_timestamp parse dd/MM",
  timestampsAsStrings,
  new SparkConf().set(SQLConf.LEGACY_TIME_PARSER_POLICY.key, "CORRECTED")) {
  df => {
    df.createOrReplaceTempView("df")
    df.sqlContext.sql("SELECT c0, to_unix_timestamp(c0, 'dd/MM') FROM df")
  }
}

testSparkResultsAreEqual("to_unix_timestamp parse MM/dd",
  timestampsAsStrings,
  new SparkConf().set(SQLConf.LEGACY_TIME_PARSER_POLICY.key, "CORRECTED")) {
  df => {
    df.createOrReplaceTempView("df")
    df.sqlContext.sql("SELECT c0, to_unix_timestamp(c0, 'MM/dd') FROM df")
  }
}
```

andygrove (approved):

razajafri:
build

PR #1385, opened 2020-12-14
Window Pandas UDF: Support sending multiple groups in a single batch to Python.

firestarman:
This PR is to support sending multiple groups in a single batch to Python side by having the window bounds been relative to the head of the whole batch, instead of the head of its group.

Thanks to the Python implementation, we can do this without changes on the Python side. But it only works when all the windows are bounded, meaning we still need to split the batch into groups when there are unbounded windows. 

Doing this is likely to improve the data transfer performance for GPUs, especially when there are too many small groups.
But it is disabled by default because this works only for bounded windows. Besides it is not 100% compatible with CPU when the data sent to Python contains nulls. 
On the Python side, Pandas will cast integral types to float when there are nulls. So for GPUs all the groups in a batch will be float type, but for CPUs only the groups with nulls will be float type. If some integral values in a group without nulls are big enough, they will lose precision after being casted to float for GPUs, then they will be different from those for CPUs.

This PR also refactors some classes to reduce the duplicated code.

- Moved some basic Python UDF related classes to a new created file `GpuArrowPythonRunner` .
- Moved most transformation for windowing into the new created iterator `WindowingPythonIterator`.
- Added some additional metrics for Python UDF plans.


firestarman:
build

firestarman:
build

sameerz:
@firestarman @GaryShen2008 can we retarget this to 0.5?

GaryShen2008:
Moved to 0.5 target.

firestarman:
build

firestarman:
build

firestarman:
build

firestarman:
@mythrocks  @revans2  Could you help reivew this ?

PR #1390, opened 2020-12-14
Accelerated columnar to row/row to columnar for decimal

revans2:
This turns on accelerated DECIMAL64 transitions and adds in comments where we would need to update the code for DECIMAL32 support.

I ran through all of TPC-DS with decimal enabled to verify that all of the tests are still passing.

I did a quick measurement and this saved about 10 ms out of 1500 ms on TPC-DS Q1 Scale factor 1 with decimal support enabled.  Not a lot but enough to be helpful, but for this query it looks it was really only two small transitions that were improved.


revans2:
build

jlowe (approved):

PR #1391, opened 2020-12-14
Remove extra null join filtering because cudf is fast for this now.

revans2:
https://github.com/rapidsai/cudf/issues/6052 is closed and into cudf-0.18 so the extra filtering that we have been doing is not needed any more, so lets rip it back out.

I also took the opportunity to bring the join code in sync for the different shim layers, so there are some other minor changes associated with that.


revans2:
build

jlowe (approved):

abellina (approved):

Issue #1354, opened 2020-12-09
[DOC] Document RAPIDS UDF


jlowe:
**Report needed documentation**
#1351 tracks implementing support for UDFs that provide a RAPIDS execution path. This is a user-visible feature that requires clear documentation on how users can implement their own RAPIDS-accelerated UDFs.

**Describe the documentation you'd like**
There should be a document under the `docs/` directory that is dedicated to describing how UDFs and the RAPIDS Accelerator interact.  That document should either directly describe how to implement a RAPIDS-accelerated UDF or point to a separate, dedicated document for RAPIDS UDFs.  The document should contain pointers to the working examples of RAPIDS UDF implementations.

PR #1393, opened 2020-12-14
Support RAPIDS-accelerated HiveSimpleUDF

### LINKED ISSUE #1352: [FEA] Example of RAPIDS UDF using cudf Java APIs
#1351 tracks implementing support for UDFs that provide a RAPIDS execution path.  With that implementation should be working examples that can be used as a reference by users when implementing their own RAPIDS-accelerated UDFs.

A simple example Maven project can be added as a sub-project in the repo.  If the example is very useful in practice, we could consider adding it to the sql-plugin project directly so it is available for use out-of-the-box with the existing plugin setup instructions rather than needing to build and use an additional jar.

If desired we could move the example into a completely separate repository.  However the amount of code involved should be quite small and build quickly, and it would be easier to find and reference in the documentation if its part of the repository.

jlowe:
Closes #1352.

This adds support for Hive simple UDFs and allows replacing Catalyst `HiveSimpleUDF` nodes in query plans with GPU-accelerated versions iff the Hive UDF function implements the new `RapidsUDF` interface.

Since Spark's Hive support is an optional component of Spark distributions, a `NoClassDefFoundError` can result if the plugin blindly accesses `HiveSimpleUDF`.  Therefore the process for building the replacement rule is a bit tricker than usual to allow the plugin to function on Spark distributions sans Hive support.  This is tested in practice by _not_ adding the spark-hive dependency to the tests and integration-tests modules since the Scala tests in those modules do not test `GpuHiveSimpleUDF`.  An integration test was added to test this functionality, so Spark with Hive support is required to run the integration tests, and it is skipped if the Spark session does not support Hive.


jlowe:
build

jlowe:
build

tgravescs (approved):
small test change, otherwise looks good

integration_tests/src/main/python/rapids_udf_test.py
```diff
@@ -0,0 +1,37 @@
...
+        pytest.skip("The Spark session does not have Hive support")
+
+def test_hive_simple_udf():
+    with_spark_session(skip_if_no_hive)
+    data_gens = [["i", int_gen], ["s", StringGen('([^%]{0,1}(%[0-9A-F][0-9A-F]){0,1}){0,30}')]]
+    def evalfn(spark):
+        spark.sql("DROP TEMPORARY FUNCTION IF EXISTS urldecode")
+        spark.sql("CREATE TEMPORARY FUNCTION urldecode AS 'com.nvidia.spark.rapids.udf.URLDecode'")
+        df = gen_df(spark, data_gens)
+        df.write.mode("overwrite").parquet("/home/jlowe/tmp/hmmm.parquet")
```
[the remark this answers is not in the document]
jlowe: Nice catch!  This was some debug code I put in during testing that should have been removed.

jlowe:
build

tgravescs (approved):

sql-plugin/src/main/java/com/nvidia/spark/RapidsUDF.java
```diff
@@ -0,0 +1,36 @@
...
+   * Evaluate a user-defined function with RAPIDS cuDF columnar inputs
+   * producing a cuDF column as output.
+   * <p>
+   * Note that the inputs should NOT be closed by this method, as they will
+   * be closed by the caller. This method must close any intermediate cuDF
+   * results produced during the computation (e.g.: `Table`, `ColumnVector`
+   * or `Scalar` instances).
+   * @param args columnar inputs to the UDF that will be closed by the caller
+   *             and should not be closed within this method.
+   * @return columnar output from the user-defined function
```
revans2, line 33: Do we want/need to add anything about what happens if the type of the output does not match the expected UDF type?  We end up throwing an exception.  The main reason I call this out is because `UDF` is done through reflection so the type is backed in, but cudf is type erased so we don't get the same static feedback.

udf-examples/src/main/java/com/nvidia/spark/rapids/udf/URLDecode.java
```diff
@@ -0,0 +1,63 @@
...
+import ai.rapids.cudf.Scalar;
+import com.nvidia.spark.RapidsUDF;
+import org.apache.hadoop.hive.ql.exec.UDF;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+
+public class URLDecode extends UDF implements RapidsUDF {
+
+  public String evaluate(String s) {
```
revans2, line 30: Could we add some java docs here?  This is supposed to be an example so explaining the different parts would be helpful.

revans2 (commented):
Looks good, mostly want a bit more documentation, like can we get a README.md for udf-examples?  Do we want to update the docs directory to describe how this works?

revans2:
Oh one other thing.  Do we need to start thinking about how this is going to interact with the UDF compiler plugin? I know it does not work on HiveUDFs yet, but should we be thinking ahead and have the interfaces split off in a separate module so it can use the same detection methods to not try and compile/replace UDFs that have a gpu version already?

jlowe:
> can we get a README.md for udf-examples?

I was planning on tackling that in #1354.

jlowe:
@revans2 I added some more comments, let me know if this matches your expectations.

> should we be thinking ahead and have the interfaces split off in a separate module so it can use the same detection methods to not try and compile/replace UDFs that have a gpu version already?

Do you want to see that as part of this PR?  Since the UDF compiler doesn't handle Hive UDFs I'm tempted to file a followup to tackle this if and when needed.

revans2:
> Do you want to see that as part of this PR? Since the UDF compiler doesn't handle Hive UDFs I'm tempted to file a followup to tackle this if and when needed.

No just something to think about as we go forward. It should be hidden from any end user because they will depend on the plugin release and cudf not however we split things apart internally and then stitch them back together again.

revans2 (approved):

jlowe:
@revans2 I updated this to the new expression checks in branch-0.4.  I _think_ I picked the appropriate checks based on `PythonUDF` but it would be good to get confirmation.

jlowe:
build

sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/GpuHiveSimpleUDF.scala
```diff
@@ -0,0 +1,131 @@
...
+          GpuColumnVector.from(s, batch.numRows(), expr.dataType)
+        }
+    }
+  }
+
+  override def columnarEval(batch: ColumnarBatch): Any = {
+    withResource(children.safeMap(evalExpr(_, batch))) { exprResults =>
+      val funcInputs = exprResults.map(_.getBase()).toArray
+      withResource(new NvtxRange(nvtxRangeName, NvtxColor.PURPLE)) { _ =>
+        GpuColumnVector.from(function.evaluateColumnar(funcInputs: _*), dataType)
```
firestarman, line 72: Shall we need to check the length and type of the cuDF ColumnVector returned from `evaluateColumnar` ?
Suppose it should have the same length with the input columns, and can cast to the `dataType` .

jlowe: Yes, that's a good point.  I added both row checks and type conversion checks.  There was already a conversion check occurring in `GpuColumnVector.from` but it only has an effect if assertions are enabled.  Since this is calling user-provided code, I added a version that performs an unconditional check.  cc: @revans2 for his thoughts on this.

jlowe:
build

sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/GpuHiveSimpleUDF.scala
revans2: Table also will verify that the lengths of all of the columns are the same, but that might not happen until after the processing has completed. I agree checking these sooner than later is ideal, especially because it is user code that we will be dealing with.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
```diff
@@ -512,6 +512,21 @@ assert typeConversionAllowed(cudfCv, type) : "Type conversion is not allowed fro
...
 
+  /**
+   * Converts a cudf internal vector to a Spark compatible vector. No reference counts
+   * are incremented so you need to either close the returned value or the input value,
+   * but not both. This conversion performs an unconditional check that the types are
+   * convertible rather than an assertion check.
+   * @throws IllegalArgumentException if the type conversion check fails
+   */
+  public static GpuColumnVector fromChecked(ai.rapids.cudf.ColumnVector cudfCv, DataType type) {
+    if (!typeConversionAllowed(cudfCv, type)) {
```
revans2, line 523: Please update the javadocs on the `typeConversionAllowed` method.

```
  /**
   * This should only ever be called from an assertion.
   */
```

is no longer true.  The message we have right now is probably OK, but it might be worth refactoring `typeConversionAllowed` in the future so it can explain exactly why it is not allowing the conversion, to help with users debugging issues.

jlowe:
build

revans2 (commented):
Just a nit, but looks good

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
jlowe: Done.  I'll file a followup issue to track improving the message for users.

jlowe:
build

revans2 (approved):

PR #1323, opened 2020-12-08
More advanced type checking and documentation

revans2:
I know this is a huge change, and I am happy to put in some comments in the code if needed.  Please take a look.

PR #1396, opened 2020-12-15
[WIP] re-enable cast float to int tests with 3.1.0

andygrove:
We had previously disabled two tests for CAST float to int due to changes in 3.1.0 that caused these tests to fail.

This is the logic used starting with Spark 3.1.0 to determine the range of valid floats that can be cast to an int:

```scala
Math.floor(x) <= Int.MaxValue && Math.ceil(x) >= Int.MinValue
```

We can no longer cast `Int.MaxValue.toFloat` to an int so the tests are updated to use in-range values.

Closes https://github.com/NVIDIA/spark-rapids/issues/1271

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/AnsiCastOpSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/AnsiCastOpSuite.scala
@@ -478,7 +478,7 @@ class AnsiCastOpSuite extends GpuExpressionTestSuite {
   }
 
   testSparkResultsAreEqual("Write floats to int (values within range)", intsAsFloats,
-    sparkConf, assumeCondition = before3_1_0) {
+    sparkConf) {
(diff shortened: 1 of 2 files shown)


andygrove:
build

andygrove:
build

revans2:
FYI: As follow on work to #1323 I was going to try and move `GpuCast.canCast` into something that could be used to generate the docs (instead of having them hard coded). Because of that it would also require that we also encode what Spark supports for casting. This should let us query this for the tests too. Do you think that would be helpful?

andygrove:
@revans2 That sounds great. Because of this change in 3.1.0 there is now a small difference in the range of values we support versus Spark and it would be good to document that.

andygrove:
The changes in this PR were already merged as part of https://github.com/NVIDIA/spark-rapids/pull/1402 so I am closing this.

PR #1401, opened 2020-12-15
Accelerate the coalescing parquet reader when reading files from multiple partitioned folders

### LINKED ISSUE #1200: [FEA] Accelerate the scan speed for coalescing parquet reader when reading files from multiple partitioned folders
When reading parquet data from a "well" partitioned dataset, the GPU scan takes longer than CPU:
image
image

There're totally 10,200 files in my job but the log `Partition values for the next file ...` appeared 9700+ times. I'm not sure if the scan really benift from the "parallel" reader.
image

tgravescs:
fixes #1200

Accelerate the scan speed for coalescing parquet reader when reading files from multiple partitioned folders.

Previously whenever we hit a file that was in a different partition we split the batch so we could easily add the partition values.  This results in us having to do a lot of batches when there are a lot of partitioned files, which is not great for performance.
To fix that we can change it to combine files with different partitioning by keeping track of the partition values and which rows those values need to be applied to.  Then after we read the files we need to add those columns that are built based off the partition values and row counts. This works because we read the files in the same order as when we construct what goes into each batch.

in this PR I added the tracking of partition values and the corresponding number of rows, then after we read the files into the columnar batch we add all the partition values.  The partition values are constructed by doing 1 partition column at a time. For each column it generates the individual partition value columns for the number of rows necessary, then it concatenates all of those together, then it moves to the next partition column.  ie meaning if you have paths with multiple partitions ../key1=2/key2=foo/ , it does key1=X for all values of X first then it does key2=Y afterwards.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarPartitionReaderWithPartitionValues.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarPartitionReaderWithPartitionValues.scala
@@ -93,7 +93,7 @@ object ColumnarPartitionReaderWithPartitionValues extends Arm {
     }
   }
 
-  private def buildPartitionColumns(
+  def buildPartitionColumns(
       numRows: Int,
       partitionValues: Array[Scalar],
       sparkTypes: Array[DataType]): Array[GpuColumnVector] = {
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarPartitionReaderWithPartitionValues.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarPartitionReaderWithPartitionValues.scala
@@ -77,23 +77,33 @@ object ColumnarPartitionReaderWithPartitionValues extends Arm {
     var partitionColumns: Array[GpuColumnVector] = null
     try {
       partitionColumns = buildPartitionColumns(fileBatch.numRows, partitionValues, sparkTypes)
-      val fileBatchCols = (0 until fileBatch.numCols).map(fileBatch.column)
-      val resultCols = fileBatchCols ++ partitionColumns
-      val result = new ColumnarBatch(resultCols.toArray, fileBatch.numRows)
-      fileBatchCols.foreach(_.asInstanceOf[GpuColumnVector].incRefCount())
+      val result = addGpuColumVectorsToBatch(fileBatch, partitionColumns)
       partitionColumns = null
       result
     } finally {
-      if (fileBatch != null) {
-        fileBatch.close()
-      }
       if (partitionColumns != null) {
         partitionColumns.safeClose()
       }
     }
   }
 
-  def buildPartitionColumns(
+  def addGpuColumVectorsToBatch(
+      fileBatch: ColumnarBatch,
+      partitionColumns: Array[GpuColumnVector]): ColumnarBatch = {
+    try {
+      val fileBatchCols = (0 until fileBatch.numCols).map(fileBatch.column)
+      val resultCols = fileBatchCols ++ partitionColumns
+      val result = new ColumnarBatch(resultCols.toArray, fileBatch.numRows)
+      fileBatchCols.foreach(_.asInstanceOf[GpuColumnVector].incRefCount())
+      result
+    } finally {
+      if (fileBatch != null) {
+        fileBatch.close()
+      }
+    }
+  }
+
+  private def buildPartitionColumns(
       numRows: Int,
       partitionValues: Array[Scalar],
       sparkTypes: Array[DataType]): Array[GpuColumnVector] = {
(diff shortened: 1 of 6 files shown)


tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -990,23 +990,106 @@ class MultiFileParquetPartitionReader(
...
+      rowsPerPartition: Array[Long],
+      inPartitionValues: Array[InternalRow]): Array[GpuColumnVector] = {
+    val numCols = partitionSchema.fields.size
+    val allPartCols = new Array[GpuColumnVector](numCols)
+    // build the partitions vectors for all partitions within each column
+    // and concatenate those together then go to the next column
+    for ((field, colIndex) <- partitionSchema.fields.zipWithIndex) {
+      val dataType = field.dataType
+      val partitionColumns = new Array[GpuColumnVector](inPartitionValues.size)
+      try {
```
jlowe, line 1003: Can use `withResource(new Array[GpuColumnVector](inPartitionValues.size))` here

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -990,23 +990,106 @@ class MultiFileParquetPartitionReader(
...
+      allPartCols = buildAndConcatPartitionColumns(rowsPerPartition, inPartitionValues)
+      val finalCb =
+        ColumnarPartitionReaderWithPartitionValues.addGpuColumVectorsToBatch(cb, allPartCols)
+      succeeded = true
+      finalCb
+    } finally {
+      if (!succeeded) {
+        allPartCols.safeClose()
+      }
+    }
```
jlowe, line 1041:
```suggestion
    val numCols = partitionSchema.fields.size
    closeOnExcept(buildAndConcatPartitionColumns(rowsPerPartition, inPartitionValues)) {
      allPartCols =>
        ColumnarPartitionReaderWithPartitionValues.addGpuColumnVectorsToBatch(cb, allPartCols)
    }
```

jlowe (changes requested):

pxLi:
build

pxLi:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarPartitionReaderWithPartitionValues.scala
```diff
@@ -77,19 +77,29 @@ object ColumnarPartitionReaderWithPartitionValues extends Arm {
     var partitionColumns: Array[GpuColumnVector] = null
     try {
```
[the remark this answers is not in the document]
tgravescs: I think I'm handling this now, let me know if you see anything I missed.

tgravescs:
thanks Jason, updated.

tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarPartitionReaderWithPartitionValues.scala
```diff
@@ -74,25 +74,30 @@ object ColumnarPartitionReaderWithPartitionValues extends Arm {
...
-      fileBatchCols.foreach(_.asInstanceOf[GpuColumnVector].incRefCount())
-      partitionColumns = null
-      result
-    } finally {
-      if (fileBatch != null) {
-        fileBatch.close()
-      }
-      if (partitionColumns != null) {
-        partitionColumns.safeClose()
+    if (fileBatch != null ) {
```
jlowe, line 77: Nit: Is this null check necessary?  The original code evaluated `fileBatch.numRows` without checking.

jlowe (approved):
Just a small nit for what appears to be an unnecessary null check but otherwise looks good to me.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarPartitionReaderWithPartitionValues.scala
tgravescs: I don't think so, the original code had that check in the finally block so I left it. The only place I could see it possibly being null is when its called from above:

 ```
val fileBatch: ColumnarBatch = fileReader.get()
      ColumnarPartitionReaderWithPartitionValues.addPartitionValues(fileBatch,
        partitionValues, partValueTypes)
```

and the fileReader.get() would return null.  I'm fine with removing it, perhaps code changed along the way and that check was never cleaned up?

jlowe: > the original code had that check in the finally block so I left it.

That protects code in the `finally` block, but before we get there the code would have blindly dereferenced `fileBatch` first.  That means we wouldn't get an NPE while trying to cleanup from throwing an NPE. :smile:

I'm totally cool with leaving it in if we think there's a remote chance a `get()` could return null which is why I went ahead and approved it.  However I would expect `next()` to return false before `get()` is allowed to return `null`.

tgravescs:
build

jlowe (approved):

PR #1402, opened 2020-12-15
Better GPU Cast type checks

revans2:
This automates generating the docs for cast.  It also updates the tests in cast so that they go off of the types listed as supported.

I did take one of the changes from #1396 to make this work (Thanks @andygrove for that work)

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
@@ -664,14 +664,14 @@ object CastOpSuite {
         "2018-1random_text",
         "2018-11-08random_text",
         "2018-11-9random_text",
-        // `yyyy-[m]m-[d]dT*` in some version of Spark this is valid and in others it is not
+        // `yyyy-[m]m-[d]dT*` in Spark 3.1+ these no longer work for AnsiCast, but did before
         "2010-1-01T!@#$%",
         "2010-1-02T,",
         "2010-01-03T*",
         "2010-01-04TT",
         "2010-02-3T*",
         "2010-02-4TT",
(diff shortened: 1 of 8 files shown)


revans2:
build

tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
```diff
@@ -648,38 +630,56 @@ object CastOpSuite {
...
+        "2010-01-6 ",
+        "2010-01-6 T",
+        "2010-01-6 T\n",
+        "2010-01-6 T\n12:34:56.000111Z",
+        "2018random_text",
+        "2018-11random_text",
+        "2018-1random_text",
+        "2018-11-08random_text",
+        "2018-11-9random_text",
+        // `yyyy-[m]m-[d]dT*` in some version of Spark this is valid and in others it is not
```
andygrove, line 667: It looks like the only time we are excluding these values from the test is when `ansiEnabled=true` rather than excluding them for particular Spark versions?

revans2: Yes, version 3.1 + when ANSI is enabled they now cause an exception to be thrown.  Because the test this is for is "Test all supported casts with in-range values" I assumed that is fine. If you want me to allow different values for different versions I can.

andygrove: I think this change is good but the comment confused me so I just wanted to make sure that I understood this correctly.

revans2: I'll update the comment to explain what is happening.

andygrove (approved):

pxLi:
build

PR #1408, opened 2020-12-16
Support RAPIDS-accelerated HiveGenericUDF

jlowe:
This adds support for Hive generic UDFs and allows replacing Catalyst HiveGenericUDF nodes in query plans with GPU-accelerated versions iff the UDF function implements the `RapidsUDF` interface.

This also cleans up the Hive detection code to match how Apache Spark detects if spark-hive is available.


integration_tests/src/main/python/rapids_udf_test.py
```diff
@@ -32,4 +32,16 @@ def evalfn(spark):
     assert_gpu_and_cpu_are_equal_sql(
         evalfn,
         "hive_simple_udf_test_table",
-        "SELECT i, urldecode(s) FROM hive_simple_udf_test_table")
+        "SELECT urldecode(s) FROM hive_simple_udf_test_table")
```
revans2, line 35: Why drop the `i` from the select?

integration_tests/src/main/python/rapids_udf_test.py
```diff
@@ -32,4 +32,16 @@ def evalfn(spark):
...
+
+def test_hive_generic_udf():
+    with_spark_session(skip_if_no_hive)
+    data_gens = [["i", int_gen], ["s", StringGen('.{0,30}')]]
```
revans2, line 39: Why generate an i if we are not going to use it anywhere?

sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/GpuHiveSimpleUDF.scala
```diff
@@ -18,23 +18,22 @@ package org.apache.spark.sql.hive.rapids
...
 
 import org.apache.spark.sql.catalyst.expressions.{Expression, UserDefinedExpression}
 import org.apache.spark.sql.hive.HiveShim.HiveFunctionWrapper
-import org.apache.spark.sql.hive.HiveSimpleUDF
 import org.apache.spark.sql.types.DataType
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
-/** GPU-accelerated version of Spark's `HiveSimpleUDF` */
-case class GpuHiveSimpleUDF(
+abstract class GpuHiveUDFBase(
```
revans2, line 31: nit: Should we rename this file now that it is for more than just `GpuHiveSimpleUDF`?

udf-examples/src/main/java/com/nvidia/spark/rapids/udf/URLEncode.java
```diff
@@ -0,0 +1,111 @@
...
+ * This class demonstrates how to implement a Hive GenericUDF that also
+ * provides a RAPIDS implementation that can run on the GPU when the query
+ * is executed with the RAPIDS Accelerator for Apache Spark.
+ */
+public class URLEncode extends GenericUDF implements RapidsUDF {
+  private transient PrimitiveObjectInspectorConverter.TextConverter converter;
+  private final Text textResult = new Text();
+
+  /** Standard getDisplayString method for implementing GenericUDF */
+   @Override
```
revans2, line 46: nit: indentation appears to be off

revans2:
Really just some nits.  It looks good overall.

sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/GpuHiveSimpleUDF.scala
jlowe: I renamed it to `hiveUDFs` which is the same name used for a similar file in Apache Spark.

jlowe:
build

revans2 (approved):

PR #1409, opened 2020-12-16
Add option to use decimals in tpc-ds csv to parquet conversion

andygrove:

Add `useDecimalType` parameter to TPC-DS `csvToParquet` and `csvToOrc` so that files can be generated with decimal type instead of double.

I confirmed that the documentation does not need updating since this new parameter has a default value.

### DIFF
--- a/integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/tpcds/TpcdsLikeSpark.scala
+++ b/integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/tpcds/TpcdsLikeSpark.scala
@@ -21,7 +21,7 @@ import com.nvidia.spark.rapids.tests.tpcds.TpcdsLikeSpark.{csvToOrc, csvToParque
 import org.rogach.scallop.ScallopConf
 
 import org.apache.spark.sql.{DataFrame, DataFrameWriter, Row, SparkSession}
-import org.apache.spark.sql.types.{DateType, DoubleType, IntegerType, LongType, StringType, StructField, StructType}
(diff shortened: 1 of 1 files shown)


integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/tpcds/TpcdsLikeSpark.scala
```diff
@@ -146,573 +148,596 @@ object TpcdsLikeSpark {
       appendDat))
   }
 
-  def setupAllCSV(spark: SparkSession, basePath: String, appendDat: Boolean = true): Unit = {
-    tables.foreach(_.setupCSV(spark, basePath, appendDat))
+  def setupAllCSV(
+      spark: SparkSession,
+      basePath: String,
+      useDecimalType: Boolean = false,
```
jlowe, line 154: Why place the new parameter here?  Anyone who was invoking the old method now is passing a boolean value that they _thought_ meant `appendDat` but now means `useDecimalType` which seems unfortunate.  If we placed the new parameter at the end of the parameter list then it becomes backwards-compatible and no surprises.

jlowe (changes requested):

integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/tpcds/TpcdsLikeSpark.scala
andygrove: Good point. Thanks. I have updated this.

jlowe (approved):

jlowe:
build

PR #1411, opened 2020-12-16
Add udf-examples jar to databricks build

jlowe:
After #1393 a new udf-examples jar needs to be added in order for some of the integration tests to pass.  This was handled automatically in `integration_tests/run_pyspark_from_build.sh` but the databricks build does not use that script.

This updates the databricks build script to copy the udf-examples jar in addition to the sql-plugin jar.


jlowe:
@tgravescs are there init script changes necessary on the databricks cluster side as well?

tgravescs (approved):
changes looks fine. I ran the rapids_udf_test.py tests on databricks using the build.sh script here to set things up and they.

tgravescs:
build

PR #1413, opened 2020-12-16
Add config for cast float to integral types

andygrove:

As of Spark 3.1.0, we are no longer 100% compatible when casting floating point types to integral types due to changes in Spark. See docs in PR for more info.

Closes https://github.com/NVIDIA/spark-rapids/issues/1271


andygrove:
build

andygrove:
build

revans2:
I am rather confused by this. The issue is only relevant for Spark 3.1+ and also appears to only show up for ANSI cast, not the regular cast.

If that is true can we make the config check only happen on Spark 3.1+ AnsiCast?  I already moved AnsiCast to the shim layer for a similar issue so we should be able to subclass the meta to do this.

Also it would be great to document what we do support instead of just what Spark supports.  That way an end user can know if it is OK to turn it on or not.

andygrove:
Thanks @revans2 I have partially addressed this but have some questions. I moved the checks to the 310 shim but I did this by overriding CastExprMeta in place. Would it be better to extend the CastChecks instead?

Also, in this case it would be nice to be able to have the default setting differ depending on the Spark version being used but I don't think we have a mechanism for doing this. Do you think we should consider adding this?

docs/compatibility.md
```diff
@@ -286,7 +299,7 @@ The following formats/patterns are supported on the GPU. Timezone of UTC is assu
 | `"tomorrow"`          | Yes               |
 | `"yesterday"`         | Yes               |
 
-## String to Timestamp
+### String to Timestamp
```
andygrove, line 302: This is unrelated but I thought it was worth fixing while I was here.

revans2:
> Thanks @revans2 I have partially addressed this but have some questions. I moved the checks to the 310 shim but I did this by overriding CastExprMeta in place. Would it be better to extend the CastChecks instead?
> 
> Also, in this case it would be nice to be able to have the default setting differ depending on the Spark version being used but I don't think we have a mechanism for doing this. Do you think we should consider adding this?

I like the way this has been done.  No need to change `CastChecks`.

I have not thought about configs that are dependent on the version of Spark being used at all. If we think we are going to run into more compatibility issues as time goes on that we cannot work around then it might be interesting to explore it, but for now I don't see a lot of value in it, assuming that we have explained clearly that a given config is only used if the version is ....

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
```diff
@@ -93,6 +94,21 @@ class Spark310Shims extends Spark301Shims {
   }
 
   def exprs310: Map[Class[_ <: Expression], ExprRule[_ <: Expression]] = Seq(
+    GpuOverrides.expr[Cast](
```
revans2, line 97: We should probably move Cast from GpuOverrides into the 300 shim just for consistency.

andygrove:
build

revans2 (approved):

PR #1422, opened 2020-12-17
Log the number of concurrent gpu tasks allowed on Executor startup

### LINKED ISSUE #1375: [FEA] Add log statement for what the concurrentGpuTasks tasks is set to on executor startup
add a log statement to what concurrentGpuTasks is se to when we initialize the semaphore. so it is easier to debug. People are trying to dynamically set it but it can't if executor has already started.  This would be really helpful in debugging.

tgravescs:
fixes #1375

Add a log message for easier debugging because this setting can't be configured dynamically and we have seen people try and then get confused.


tgravescs:
build

jlowe (approved):

PR #1424, opened 2020-12-17
Make the multi-threaded parquet reader the default since coalescing doesn't handle partitioned files well

### LINKED ISSUE #1366: Seeing performance differences of multi-threaded/coalesce/perfile Parquet reader type for a single file
We ran into a case where a single Parquet file loaded an order of magnitude slower (800ms vs 80ms) comparing 0.3 snapshot vs 0.2 of the plugin. 

But when we set: `spark.rapids.sql.format.parquet.reader.type=MULTITHREADED` or `spark.rapids.sql.format.parquet.reader.type=COALESCING` we see consistent or better results <= 80ms in 0.3. 

This seems interesting because it's a single 128MB parquet file, and there's a single batch that comes out of it. So it looks like the reader has some overheads if left by default `AUTO`. Note that the file was being read from `hdfs://`. 

The file in question was one of the Parquet parts of the `web_sales` table for TPCDS at 1TB, and the query where we saw significant effect was q88. The file has decimals in it, and the test projected out the integers s.t. the gpu parquet scan would be enabled:

```
val test = spark.read.parquet("hdfs://.../web_sales/ws_sold_date_sk=2450816")
test.select("ws_web_page_sk", "ws_order_number").write.mode("overwrite").parquet("test")
```

tgravescs:
Make the multi-threaded parquet reader the default since coalescing doesn't handle partitioned files well.

fixes #1366

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
@@ -544,13 +544,16 @@ object RapidsConf {
       "See spark.rapids.sql.format.parquet.multiThreadedRead.numThreads and " +
       "spark.rapids.sql.format.parquet.multiThreadedRead.maxNumFilesParallel to control " +
       "the number of threads and amount of memory used. " +
-      "By default this is set to AUTO so we select the reader we think is best. This will " +
+      "This can be set to AUTO to select the reader we think is best. This will " +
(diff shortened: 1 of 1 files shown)


tgravescs:
build

tgravescs:
Note this is intentionally targetted at 0.3, the 0.4 changes #1200 went into 0.4 and we will have to revert this change there.

revans2 (approved):

abellina (approved):

jlowe (approved):

Issue #1290, opened 2020-12-05
Build a Test framework to skip tests that can't be compiled


razajafri:
Why is this using reflection to call code we wrote?

_Originally posted by @jlowe in https://github.com/NVIDIA/spark-rapids/pull/1265#discussion_r536450868_

jlowe:
Scala tests now build with shim framework which solves this.

PR #1430, opened 2020-12-19
Add project for testing code that requires Spark 3.1.0 or later

razajafri:
This PR deals with cases where tests need classes that are in the upcoming version of Spark. It addresses the concern in  #1290 but isn't a test framework. 

This module will not be built by default, it's only kicked off with `spark310tests` profile


pom.xml
```diff
@@ -136,6 +136,9 @@
         </profile>
         <profile>
             <id>spark310tests</id>
+            <modules>
+                <module>tests-next-spark</module>
```
razajafri, line 140: I named it so its not tied to 3.1.0 only. If you can come up with a better name please let me know

sameerz:
@razajafri is this related to or does it close issue #1290 ?

razajafri:
@sameerz though not a `framework` it addresses the story.

tests-next-spark/README.md
```diff
@@ -0,0 +1,19 @@
+---
+layout: page
+title: Testing
+nav_order: 1
+---
+# RAPIDS Accelerator for Testing against the upcoming version of Apache Spark
+
+While writing unit-tests, we can run into situations where we depend on classes that are not
+ available in the default version of Spark (currently 3.0.0). In such a scenario we put
```
jlowe, line 9: I'd refrain from putting in a version in here which is immediately out of date (3.0.1 is actually the "current" version).  Instead I suggest focusing on the snapshot nature of it, e.g.:
```suggestion
While writing unit-tests, we can run into situations where we depend on classes that are only
available in the SNAPSHOT version of the upcoming next version of Spark. In such a scenario we put
```

tests-next-spark/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+    <version>0.4.0-SNAPSHOT</version>
+
+    <properties>
+        <spark.test.version>3.1.0-SNAPSHOT</spark.test.version>
+    </properties>
+    <profiles>
+        <profile>
+            <id>spark310tests</id>
+        </profile>
+    </profiles>
```
jlowe, line 38: This profile is not used by this POM.  It's used in the parent pom to include this one.
```suggestion
```

tests/pom.xml
```diff
@@ -104,7 +104,7 @@
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
-            <artifactId>mockito-inline</artifactId>
+            <artifactId>mockito-core</artifactId>
             <version>${mockito.version}</version>
```
jlowe, line 108: The version should be specified by the `dependencyManagement` section of the parent pom, so it doesn't need to be specified here.
```suggestion
```

jlowe (changes requested):

pom.xml
revans2: I personally would prefer to have the name of the version in it, something like `tests-spark310-plus` to indicate that this is for spark 3.1.0 and above.  The only time we could get away with a `tests-next` type package name would be if we have one and only one set of tests that are specific to a version.  And even then I don't see much advantage to not being specific in the name.

pom.xml
```diff
@@ -530,6 +533,7 @@
                         <exclude>docs/demo/**/*.ipynb</exclude>
                         <exclude>docs/demo/**/*.zpln</exclude>
                         <exclude>**/src/main/resources/META-INF/services/*</exclude>
+                        <exclude>tests-next-spark/target/**/*</exclude>
```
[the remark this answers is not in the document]
revans2: What I have found is that rat disables checking of the target directory for all active projects.  If the project is not active (i.e. it was disabled by a profile) the rat will check it and likely fail if it was not cleaned up in between.  I have been bitten by this a few times.  I personally am fine with this for all sub-projects that we enable/disable, but I would almost rather have it be `**/target/**/*` along with an explanation as to why we are doing this (because it is not obvious). Or just remove the rule because it is not hard to delete a directory when you make a mistake.

pom.xml
razajafri: Any objections to `tests-spark310+`

pom.xml
```diff
@@ -530,6 +533,10 @@
                         <exclude>docs/demo/**/*.ipynb</exclude>
                         <exclude>docs/demo/**/*.zpln</exclude>
                         <exclude>**/src/main/resources/META-INF/services/*</exclude>
+                        <!-- Apache Rat excludes target folder for projects that are included by
+                        default. The reason we are explicitly excluding the target folder is because
+                        the tests-spark310+ project isn't included by default -->
```
jlowe, line 538: Recommend making this comment not specific to the test-spark310+ project, as it's really applicable to any conditionally-included project.
```suggestion
                        default, but there are some projects that are conditionally included.  -->
```

pom.xml
```diff
@@ -530,6 +533,10 @@
...
+                        <exclude>**/target/**/*</exclude>
```
jlowe, line 539: Nit: I believe **/* is redundant, should be able to simplify this to `**/target/**`.

revans2 (commented):
Looks OK to me but I'll hold off on approving until Jason has a look

jlowe (approved):

revans2 (approved):

razajafri:
build

PR #1438, opened 2020-12-31
Cast from numeric types to decimal type

sperlingxx:
This pull request is to support casting from numeric types to decimal type under GPU runtime, which is one of important feature of TPC-DS decimal support plan.

Here are limits of current implementation:
1. Has restriction on range of Float/Double values, because storage capacity of GPU decimal is restricted by INT64.
2. There may exists tiny difference between CPU results and GPU results when casting from floats/doubles. In CPU runtime, floats/doubles are firstly converted to strings, then converting to decimals from strings. We can't do the same thing under GPU runtime in current.


sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -681,10 +681,10 @@ class CastChecks extends ExprChecks {
   val booleanChecks: TypeSig = integral + fp + BOOLEAN + TIMESTAMP + STRING
   val sparkBooleanSig: TypeSig = numeric + BOOLEAN + TIMESTAMP + STRING
 
-  val integralChecks: TypeSig = integral + fp + BOOLEAN + TIMESTAMP + STRING + BINARY
+  val integralChecks: TypeSig = numeric + BOOLEAN + TIMESTAMP + STRING + BINARY
```
revans2, line 684: You have an assert in the cast code indicating that it only works for ANSI casts, but here you have updated all casts to support decimal. We cannot merge this patch in like this as it will throw an exception instead of falling back to the CPU.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -372,6 +372,59 @@ case class GpuCast(
...
+
+      case (FloatType | DoubleType, dt: DecimalType) =>
+        assert(ansiMode, "GpuCastToDecimal can only run under ansiMode")
+
+        // Approach to minimize difference between CPUCast and GPUCast:
+        // step 1. cast input to FLOAT64 (if necessary)
+        // step 2. cast FLOAT64 to container DECIMAL (who keeps one more digit for rounding)
+        // step 3. perform HALF_UP rounding on container DECIMAL
+        // Corner case: If target scale reaches DECIMAL64_MAX_PRECISION, container DECIMAL can not
+        // be created because of precision overflow. In this case, we perform casting op directly.
```
revans2, line 408: We need to document this in the compatibility document so users understand that our cast is not 100% perfect. We also need a follow on issue to figure out how to make it 100% correct.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -372,6 +372,59 @@ case class GpuCast(
...
+        withResource(input.getBase.castTo(DType.FLOAT64)) { double =>
+          val scale = (dt.scale + 1) min DType.DECIMAL64_MAX_PRECISION
+          // SafeFactor here is to adapt the tolerance caused by FLOAT32 to FLOAT64 conversion.
+          // For example, 1e16f will become 1.000000003e16 after casting to double type.
+          val safeFactor = 1.0001
```
revans2, line 413: How exactly was the `safeFactor` determined?

sperlingxx: The value `1.0001` here is not a strictly computed number. Theoretically, it should be the smallest figure which can cover the tolerance caused by FLOAT32 to FLOAT64 conversion.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
sperlingxx: Compatibility document has been updated. And I will create the issue in soon.

sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
sperlingxx: I've updated `CastCheck` to support `ansiMode` check for numeric to decimal conversion.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -372,6 +386,59 @@ case class GpuCast(
       case (ShortType | IntegerType | LongType | ByteType | StringType, BinaryType) =>
         input.getBase.asByteList(true)
 
+      case (IntegerType | LongType, dt: DecimalType) =>
```
revans2, line 389: What about byte and short which are also listed as supported for ANSI cast in the docs?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
revans2: What would it take to actually calculate this? Because it is not good enough.

If I want to cast the double `999.999` to `Decimal(5, 2)` I would get an error back in Spark, but with this I would likely get an overflow when I do the half up rounding.

This is because `1.0001 * 10 ^ 3` = `1000.1` and `999.999` fits in the range supplied here. (assuming that my comment about `scale` vs `dt.scale` is adjusted for.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -717,6 +721,18 @@ class CastChecks extends ExprChecks {
...
+  protected def addDecimalSig(sig: TypeSig): TypeSig = {
+    val ansiModeValue = ansiMode.getOrElse(SparkSession.getActiveSession match {
+      case Some(session) => session.sessionState.conf.ansiEnabled
+      case None => false
+    })
+    if (ansiModeValue) {
+      sig + DECIMAL
+    } else {
+      sig
+    }
```
gerashegalov, line 733: nit: we can write this
```scala
    ansiMode.orElse(SparkSession.getActiveSession.map(_.sessionState.conf.ansiEnabled))
      .collect { case ansiModeValue if ansiModeValue => sig + DECIMAL }
      .getOrElse(sig)
```

sperlingxx: Thank you @gerashegalov! I've applied your suggestions. The functional programming style looks really cool!

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -372,6 +386,59 @@ case class GpuCast(
...
+        // step 3. perform HALF_UP rounding on container DECIMAL
+        // Corner case: If target scale reaches DECIMAL64_MAX_PRECISION, container DECIMAL can not
+        // be created because of precision overflow. In this case, we perform casting op directly.
+        withResource(input.getBase.castTo(DType.FLOAT64)) { double =>
+          val scale = (dt.scale + 1) min DType.DECIMAL64_MAX_PRECISION
+          // SafeFactor here is to adapt the tolerance caused by FLOAT32 to FLOAT64 conversion.
+          // For example, 1e16f will become 1.000000003e16 after casting to double type.
+          val safeFactor = 1.0001
+          assertValuesInRange(double,
+            Scalar.fromDouble(-safeFactor * math.pow(10, DType.DECIMAL64_MAX_PRECISION - scale)),
```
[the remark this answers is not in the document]
sperlingxx: I've rewritten bound checking part and corresponding tests. Currently, it will throw exception correctly for half up rounding cases, like: casting the double 999.999 to Decimal(5, 2).

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -372,6 +386,59 @@ case class GpuCast(
...
+        // Check whether max precision of target decimal exceeds DType.DECIMAL64_MAX_PRECISION,
+        // so the sanity range is relied on the scale of target DecimalType.
+        assertValuesInRange(input.getBase,
+          Scalar.fromDouble(-math.pow(10, DType.DECIMAL64_MAX_PRECISION - dt.scale)),
+          Scalar.fromDouble(math.pow(10, DType.DECIMAL64_MAX_PRECISION - dt.scale)))
+
+        if (dt.scale < 0) {
+          // Rounding is essential when scale is negative,
+          // so we apply HALF_UP rounding manually to keep align with CpuCast.
+          withResource(input.getBase.castTo(DType.create(DType.DTypeEnum.DECIMAL64, 0))) { zero =>
```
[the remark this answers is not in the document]
sperlingxx: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
sperlingxx: Yes, this approach doesn't look good. I've rewritten it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -372,6 +372,59 @@ case class GpuCast(
       case (ShortType | IntegerType | LongType | ByteType | StringType, BinaryType) =>
         input.getBase.asByteList(true)
 
+      case (IntegerType | LongType, dt: DecimalType) =>
+        assert(ansiMode, "GpuCastToDecimal can only run under ansiMode")
```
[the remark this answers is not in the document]
sperlingxx: Previously, I am not sure how to accomplish non-ansi casts. Referring `IfElse` support work, I think we can use `IfElse` to replace overflow values with nulls. Is it a reasonable implementation?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -38,6 +38,20 @@ class CastExprMeta[INPUT <: CastBase](
   val toType = cast.dataType
 
   override def tagExprForGpu(): Unit = {
+    if (!conf.isCastLongToDecimalEnabled && toType.isInstanceOf[DecimalType] &&
```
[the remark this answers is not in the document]
sperlingxx: Removed.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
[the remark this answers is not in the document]
revans2: Yes that is what we do in the rest of the code.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -400,11 +491,58 @@ case class GpuCast(
...
+  private def replaceOutOfRangeValues(values: ColumnVector,
+    minValue: => Scalar,
+    maxValue: => Scalar,
+    replaceValue: => Scalar,
+    inclusiveMin: Boolean = true,
+    inclusiveMax: Boolean = true): ColumnVector = {
+
+    withResource(minValue) { minValue =>
+      withResource(maxValue) { maxValue =>
+        val minPredicate = inclusiveMin match {
```
revans2, line 528: nit: technically if we get an exception when calculating `maxPredicate`, `minPredicate` will leak.  Not a huge deal but it might be nice to fix it.

```
def calcMaxPred = inclusiveMax match {
  case true => values.greaterThan(maxValue)
  case false => values.greaterOrEqualTo(maxValue)
}

def calcMinPred = inclusiveMin match {
  case true => values.lessThan(minValue)
  case false => values.lessOrEqualsTo(minValue)
}

val rangePredicate = withResource(calcMinPred) { minPredicate =>
    withResource(calcMaxPred) { maxPredicate =>
      ...
    }
}
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -400,11 +491,58 @@ case class GpuCast(
...
+        case false => values.greaterOrEqualTo(maxValue)
+      })
+    }
+  }
+
+  /**
+   * Detects outlier values of a column given with specific range, and replaces them with
+   * a inputted substitution value.
+   *
+   * @param values ColumnVector
```
revans2, line 512: nit: ColumnVector is not much of an explanation and does not add anything more than would be auto generated from the signature.

revans2 (dismissed):
Just a few nits.  Overall it looks really good

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
sperlingxx: Fixed

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
sperlingxx: Fixed

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
```diff
@@ -111,17 +111,17 @@ class Spark310Shims extends Spark301Shims {
...
-      new CastChecks() {
+      new CastChecks {
         import TypeSig._
         // nullChecks are the same
 
         override val booleanChecks: TypeSig = integral + fp + BOOLEAN + STRING
         override val sparkBooleanSig: TypeSig = numeric + BOOLEAN + STRING
 
-        override val integralChecks: TypeSig = integral + fp + BOOLEAN + STRING
+        override val integralChecks: TypeSig = numeric + BOOLEAN + TIMESTAMP + STRING + BINARY
```
revans2, line 121: Sorry this is a really minor nit. Spark does not support TIMESTAMP or BINARY (as you can see on the line below), so us claiming to support it is technically wrong.  This should not impact the generated docs or the correctness of the code in any way.

It is also an issue for fpChecks with TIMESTAMP.

The only reason to fix it is to avoid confusion.

revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
revans2: Sorry this is still missing byte.  Is that expected?

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
sperlingxx: It's been fixed. My bad :)

shims/spark310/src/main/scala/com/nvidia/spark/rapids/shims/spark310/Spark310Shims.scala
```diff
@@ -111,17 +111,17 @@ class Spark310Shims extends Spark301Shims {
...
-      new CastChecks() {
+      new CastChecks {
         import TypeSig._
         // nullChecks are the same
 
         override val booleanChecks: TypeSig = integral + fp + BOOLEAN + STRING
         override val sparkBooleanSig: TypeSig = numeric + BOOLEAN + STRING
 
-        override val integralChecks: TypeSig = integral + fp + BOOLEAN + STRING
+        override val integralChecks: TypeSig = numeric + STRING + STRING
```
revans2, line 121: `BOOLEAN + STRING` not `STRING + STRING`

sperlingxx: Fixed.

sperlingxx:
build

sperlingxx:
build

pxLi:
build

pxLi:
build

revans2 (approved):

Issue #1423, opened 2020-12-17
[BUG] Mortgage ETL sample failed with spark.sql.adaptive enabled on AWS EMR 6.2 


mgzhao:
**Describe the bug**

Mortgage ETL sample failed with spark.sql.adaptive enabled on AWS EMR 6.2 with 0.3 release.
With  "spark.sql.adaptive.enabled":"false",  the same ETL sample on EMR PySpark Notebook will run and give us end-to-end number. 

With AQE on we get the following error message from the notebook
```
An error was encountered:
An error occurred while calling o757.parquet.
: org.apache.spark.SparkException: Job aborted.
    at org.apache.spark.sql.rapids.GpuFileFormatWriter$.write(GpuFileFormatWriter.scala:250)
    at org.apache.spark.sql.rapids.GpuInsertIntoHadoopFsRelationCommand.runColumnar(GpuInsertIntoHadoopFsRelationCommand.scala:166)
    at com.nvidia.spark.rapids.GpuDataWritingCommandExec.sideEffectResult$lzycompute(GpuDataWritingCommandExec.scala:61)
    at com.nvidia.spark.rapids.GpuDataWritingCommandExec.sideEffectResult(GpuDataWritingCommandExec.scala:60)
    at com.nvidia.spark.rapids.GpuDataWritingCommandExec.doExecuteColumnar(GpuDataWritingCommandExec.scala:84)
    at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeColumnar$1(SparkPlan.scala:207)
    at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeQuery$1(SparkPlan.scala:218)
    at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
    at org.apache.spark.sql.execution.SparkPlan.executeQuery(SparkPlan.scala:215)
    at org.apache.spark.sql.execution.SparkPlan.executeColumnar(SparkPlan.scala:203)
    at com.nvidia.spark.rapids.GpuColumnarToRowExecParent.doExecute(GpuColumnarToRowExec.scala:301)
    at org.apache.spark.sql.execution.SparkPlan.$anonfun$execute$1(SparkPlan.scala:180)
    at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeQuery$1(SparkPlan.scala:218)
    at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
[...]

PR #1439, opened 2021-01-04
when AQE enabled we fail to fix up exchanges properly and EMR

tgravescs:
This is part of issue in https://github.com/NVIDIA/spark-rapids/issues/1423

The issue here is in the way that AQE does its subqueries at exchange boundaries and when we re-analyze the subqueries we only see the children of the exchange.  This can results in our fixUpExchangeOverhead marking the plan different initially from when the subquery is actually ran and we end up having one part of the join coming from CPU exchange and one part coming from GPU exchange and we blow up.

To handle this case we have to make sure we initially mark the plan the same as AQE will on its subqueries. This means in the fixUpExchangeOverhead when AQE is enabled we only look a the children in deciding if the exchange will run on the GPU or not.

there are more details in the issue.  

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
@@ -543,14 +543,16 @@ abstract class SparkPlanMeta[INPUT <: SparkPlan](plan: INPUT,
     }
   }
 
-  // This has to handle adaptive execution and the fact that it splits and
-  // runs subqueries at shuffle boundaries. When we analyze the subquery
-  // we don't know what the other side of the join is so we have to ensure it
-  // gets marked properly the first time through. It also only sees see the
-  // children when doing the subquery so if any child can't be replaced
-  // when AQE is on we make sure to mark the exchange as can't be replaced.
-  // Otherwise we could get a mismatch like
+  // For adaptive execution we have to ensure we mark everything properly
+  // the first time through and that has to match what happens when AQE
+  // splits things up and does the subquery analysis at the shuffle boundaries.
+  // If the AQE subquery analysis changes the plan from what is originally
+  // marked we can end up with mismatches like happened in:
   // https://github.com/NVIDIA/spark-rapids/issues/1423
+  // AQE splits subqueries at shuffle boundaries which means that it only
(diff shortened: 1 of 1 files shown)


revans2 (dismissed):
This looks good to me but I would like to hear from @andygrove to see if there is something I missed about AQE for this.

tgravescs:
sorry let me update the comment here I have a typo and I had a more detailed version

tgravescs:
build

tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -543,11 +543,23 @@ abstract class SparkPlanMeta[INPUT <: SparkPlan](plan: INPUT,
...
+  // child can't be replaced.
   private def fixUpExchangeOverhead(): Unit = {
     childPlans.foreach(_.fixUpExchangeOverhead())
     if (wrapped.isInstanceOf[ShuffleExchangeExec] &&
-      (parent.filter(_.canThisBeReplaced).isEmpty &&
-        childPlans.filter(_.canThisBeReplaced).isEmpty)) {
+      ((parent.filter(_.canThisBeReplaced).isEmpty &&
+        childPlans.filter(_.canThisBeReplaced).isEmpty) ||
+        (plan.conf.adaptiveExecutionEnabled &&
+          childPlans.filter(_.canThisBeReplaced).isEmpty))) {
```
andygrove, line 562: nit: it might be simpler to change this condition to:

```scala
childPlans.filter(_.canThisBeReplaced).isEmpty && 
  (plan.conf.adaptiveExecutionEnabled || 
  parent.filter(_.canThisBeReplaced).isEmpty)
```

LGTM to me though

andygrove (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
tgravescs: I'll update and retest

tgravescs:
build

andygrove (approved):

revans2 (approved):

PR #1452, opened 2021-01-05
Support DecimalType for CaseWhen

razajafri:
Turned Decimal Type on for CaseWhen and modified test to include the DecimalType

### DIFF
--- a/integration_tests/src/main/python/data_gen.py
+++ b/integration_tests/src/main/python/data_gen.py
@@ -684,6 +684,8 @@ def to_cast_string(spark_type):
         return 'TIMESTAMP'
     elif isinstance(spark_type, StringType):
         return 'STRING'
+    elif isinstance(spark_type, DecimalType):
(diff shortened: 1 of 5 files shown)


integration_tests/src/main/python/data_gen.py
```diff
@@ -684,6 +684,8 @@ def to_cast_string(spark_type):
         return 'TIMESTAMP'
     elif isinstance(spark_type, StringType):
         return 'STRING'
+    elif isinstance(spark_type, DecimalType):
+        return 'DECIMAL'
```
jlowe, line 688: Shouldn't this include the precision and scale in the cast type?

razajafri: yes, I should've noticed that we are using `cast` not `decimal` function. made the change

razajafri:
build

jlowe (approved):

PR #1453, opened 2021-01-05
DecimalType support for IfElse and Coalesce

razajafri:
Turned on the DecimalType support for IfElse and Coalesce and added DecimalGen in tests for both ops


revans2 (approved):

revans2:
build

PR #1455, opened 2021-01-06
Upgrade Scala Maven Plugin to 4.3.0

gerashegalov:
Fixes the issue with `JAVA_HOME` resolution described in https://github.com/davidB/scala-maven-plugin/issues/221.
Encountered when building using IntelliJ's maven plugin

Signed-off-by: Gera Shegalov <gshegalov@nvidia.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -176,6 +176,7 @@
         <spark302.version>3.0.2-SNAPSHOT</spark302.version>
         <spark310.version>3.1.0-SNAPSHOT</spark310.version>
         <mockito.version>3.6.0</mockito.version>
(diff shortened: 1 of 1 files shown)


revans2 (approved):

revans2:
build

tgravescs:
I don't think it will affect us much since our build is pretty fast, but the maven plugin changed the way it figures out if it needs to rebuild something. The Spark build time increased a lot now that they moved up as well.  Unless of course they have fixed it, they were supposed to be working on alternate solution, don't remember the details off the top of my head but could dig up if needed.

PR #1458, opened 2021-01-06
Add in support for struct and named_struct

### LINKED ISSUE #78: [FEA] Support CreateNamedStruct function
Once we support structs #17 we should look at supporting the `named_struct` function.

revans2:
This fixes #78 

It lacks support for nested types as values, but I can add that in if we need it.  The only issue would be that I cannot support literal values for these because we don't support scalars for nested types yet.

This is blocked until https://github.com/rapidsai/cudf/pull/7084 is merged in.

### DIFF
--- a/integration_tests/src/main/python/struct_test.py
+++ b/integration_tests/src/main/python/struct_test.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020-2021, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/integration_tests/src/main/python/struct_test.py
+++ b/integration_tests/src/main/python/struct_test.py
@@ -31,3 +31,11 @@ def test_struct_get_item(data_gen):
                 'a.first',
                 'a.second',
                 'a.third'))
+
+@pytest.mark.parametrize('data_gen', all_basic_gens + [decimal_gen_default, decimal_gen_scale_precision], ids=idfn)
+def test_make_struct(data_gen):
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : binary_op_df(spark, data_gen).selectExpr(
+                'struct(a, b)',
+                'named_struct("foo", b, "bar", 5, "end", a)'))
+
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
@@ -22,7 +22,7 @@ import java.time.ZoneId
 
 import ai.rapids.cudf.DType
 
-import org.apache.spark.sql.catalyst.expressions.{CaseWhen, Expression, UnaryExpression, WindowSpecDefinition}
+import org.apache.spark.sql.catalyst.expressions.{CaseWhen, CreateNamedStruct, Expression, UnaryExpression, WindowSpecDefinition}
 import org.apache.spark.sql.execution.SparkPlan
 import org.apache.spark.sql.types._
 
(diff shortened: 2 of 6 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -22,7 +22,7 @@ import java.time.ZoneId
 
 import ai.rapids.cudf.DType
 
-import org.apache.spark.sql.catalyst.expressions.{CaseWhen, Expression, UnaryExpression, WindowSpecDefinition}
```
jlowe: 2021 copyrights on this file

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  override def sql: String = getTagValue(FUNC_ALIAS).map { alias =>
+    val childrenSQL = children.indices.filter(_ % 2 == 1).map(children(_).sql).mkString(", ")
+    s"$alias($childrenSQL)"
+  }.getOrElse(super.sql)
+
+  override def columnarEval(batch: ColumnarBatch): Any = {
+    // The names are only used for the type. Here we really just care about the data
+    val columns: Array[ColumnVector] = new Array(valExprs.size)
+    try {
```
jlowe, line 85: Suggest using `withResource` here.

integration_tests/src/main/python/struct_test.py
```diff
@@ -31,3 +31,11 @@ def test_struct_get_item(data_gen):
                 'a.first',
```
jlowe, line 31: 2021 copyrights

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
revans2: That does not work for Arrays because it changes the type to be a Seq and I cannot modify it after that.

integration_tests/src/main/python/struct_test.py
revans2: done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
jlowe: It should work, as there was a specialization added explicitly for the `Array` case.  The pattern with an Array is being used in [GpuParquetScan](https://github.com/NVIDIA/spark-rapids/blob/branch-0.4/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala#L1034) and [MetaUtils](https://github.com/NVIDIA/spark-rapids/blob/branch-0.4/sql-plugin/src/main/scala/com/nvidia/spark/rapids/MetaUtils.scala#L228), for example.

revans2: I agree, and I think I tried it once, but ran into some compile issues because now the array could be covered by both functions, and my scala skills were just not strong enough to figure out how to make it work properly. I can try again if you want me to.

revans2: But I could always add in a different name that would probably make it still work.  I'll give it a shot

jlowe: It's not a big deal, the `try` block works so don't waste too much time on it.  I am curious why it didn't work in this instance.  What's being done here doesn't seem that different than other uses of `withResource(new Array.[..` in the code base.

revans2: Sorry it looks like someone fixed it for me in Arm. Thanks @tgravescs  So I will update it

jlowe (approved):

revans2:
build

PR #1472, opened 2021-01-07
Add an example RAPIDS-accelerated Hive UDF using native code

### LINKED ISSUE #1353: [FEA] Example of RAPIDS UDF using custom GPU code
#1351 tracks implementing support for UDFs that provide a RAPIDS execution path. With that implementation should be working examples that can be used as a reference by users when implementing their own RAPIDS-accelerated UDFs.

We will need to discuss whether this is more appropriate to be a separate repository or part of the current repository.  Since custom GPU code needs to be built as part of the working example, there's a lot more involved in setting up the native build environment that is not needed for building any other code in this repository.  Therefore it may be more appropriate to implement it in a separate repository.  If it is implemented in the current repository there needs to be a Maven profile to control whether it is built.

jlowe:
Closes #1353

This adds an example of a Hive UDF that leverages native CUDA code to implement a RAPIDS-accelerated version.  A Dockerfile is provided to show how to setup an environment for compiling the native UDF code against libcudf and the version of cub, thrust, and libcudacxx used by libcudf.

An integration test is provided, but the user must supply the `--rapids_udf_example_native` flag in order to enable the test, as most nightly builds will not bother to perform the extra steps necessary to build the native code required by the new example.

Documentation that covers this native example and implementing RAPIDS-acclerated Hive UDFs in general will be covered in #1354.

### DIFF
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020-2021, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -52,4 +52,8 @@ def pytest_addoption(parser):
     )
     parser.addoption(
         "--cudf_udf", action='store_true', default=False, help="if true enable cudf_udf test"
+    )
+    parser.addoption(
+        "--rapids_udf_example_native", action='store_true', default=False,
+        help="if true enable tests for RAPIDS UDF examples with native code"
     )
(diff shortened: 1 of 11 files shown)


jlowe:
build

tgravescs (approved):
overall looks fine to me, but would be good to get someone more familiar with c++/cuda side to review.
Do we want to put the cuda version in the name of the jar or library?
Do we want to add README here to explain using docker file and use profile udf-native-examples or is that in the other PR?

tgravescs:
oops never mind, you already said docs later

jlowe:
> Do we want to put the cuda version in the name of the jar or library?

The native code statically links the CUDA runtime, so we don't need to put the CUDA version in the jar name.

Issue #1326, opened 2020-12-08
[FEA] Add in Broadcast support for decimal values


revans2:
**Is your feature request related to a problem? Please describe.**
To support other operations we need to be able to send decimal types from one task to another (Like join #1325)

The code should already support this, but we need to enable it and test.

revans2:
I didn't think about how to test this properly.  Broadcast really needs Joins working, because broadcast is not used without joins right now.  So I will do Shuffle separately and keep this just for Broadcast.

Issue #1325, opened 2020-12-08
[FEA] Add in join support for decimal


revans2:
**Is your feature request related to a problem? Please describe.**
CUDF supports decimal joins, we should update our join code/tests to also support decimals in joins.

We primarily want hash join support, but cross join support would be great too.

revans2:
This was fixed by #1475 but not liked properly

PR #1475, opened 2021-01-08
 Join support for DecimalType

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/1326 and https://github.com/NVIDIA/spark-rapids/issues/1325

### DIFF
--- a/integration_tests/src/main/python/join_test.py
+++ b/integration_tests/src/main/python/join_test.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020-2021, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/integration_tests/src/main/python/join_test.py
+++ b/integration_tests/src/main/python/join_test.py
@@ -23,7 +23,9 @@ from spark_session import with_spark_session, is_before_spark_310
 all_gen = [StringGen(), ByteGen(), ShortGen(), IntegerGen(), LongGen(),
            BooleanGen(), DateGen(), TimestampGen(), null_gen,
            pytest.param(FloatGen(), marks=[incompat]),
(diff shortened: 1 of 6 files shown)


integration_tests/src/main/python/join_test.py
```diff
@@ -23,7 +23,9 @@
 all_gen = [StringGen(), ByteGen(), ShortGen(), IntegerGen(), LongGen(),
```
razajafri, line 23: Its better to use the `all_gen` in `data_gen.py` and add the stuff missing

revans2: So I really hate `all_gen` being in data_gen.py because it is not really generating everything (all). I let it through before so I didn't say anything when it was reused, but if you do use it I would want it to be renamed so it is clear what is in it.

revans2 (approved):

revans2:
build

integration_tests/src/main/python/join_test.py
nartal1: `all_gen` in `data_gen.py` is missing these 2 gens: `null_gen, decimal_gen_neg_scale` . 
Please let me know if below would be acceptable name if we want to rename it for better readability? 

1. `all_gen_minus_null_gen_and_decimal_gen_neg_scale` OR  `all_gen_excluding_null_gen_and_decimal_gen_neg_scale` (problem I see with this is we may have to rename it again if we add any other gen in the future which is not included in this variable).
Suggestions are welcome on variable name :) .

Or would you prefer to keep `all_gen` local to the test files and remove it from `data_gen.py` ?(I can create follow on PR for that as it involves changing other files).

razajafri: I can't think of a name suitable. @jlowe is better at naming things may be he has something

The right thing to do in my mind would be to add `everything` in all and let tests create local versions of a list if they need something different. Considering it might break tests if you add everything in `all_gen` lets create a follow-on for this as a lower priority task

PR #1476, opened 2021-01-08
DecimalType support for Aggregate Count

### LINKED ISSUE #1334: [FEA] Add support for count aggregate on decimal
cudf does not support this yet, but when they do we should add in support for this.

razajafri:
Turned on the Decimal Type for Count on and added test

This fixes #1334

### DIFF
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -418,3 +418,10 @@ def test_arithmetic_reductions(data_gen):
                 'avg(a)'),
             conf = _no_nans_float_conf)
 
+@ignore_order
+@pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
+@pytest.mark.parametrize('count_func', [f.count, f.countDistinct], ids=idfn)
+def test_agg_count(data_gen, count_func):
+    assert_gpu_and_cpu_are_equal_collect(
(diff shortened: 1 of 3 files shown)


jlowe (approved):
lgtm.  I was expecting a change to an existing test rather than a new test, but I couldn't find a test that just did a count without combining it with other aggregates and/or filters.

revans2 (approved):

revans2:
build

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -418,3 +418,10 @@ def test_arithmetic_reductions(data_gen):
                 'avg(a)'),
             conf = _no_nans_float_conf)
 
+@ignore_order
+@pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
+@pytest.mark.parametrize('count_func', [f.count, f.countDistinct], ids=idfn)
```
revans2, line 423: Using a function in here with idfn is producing different names each time the test runs and is causing the parallel runs of the tests to fail. If you drop the `ids=idfn` it fixes the problem

revans2 (changes requested):

revans2 (approved):

revans2:
build

revans2 (approved):

PR #1477, opened 2021-01-08
Allow structs and arrays to pass through for Shuffle and Sort 

kuhushukla:
This change allows structs and arrays for shuffle and sort for pass through as i/p and o/p. The change is fairly simple so I am not sure if more is expected from this change or if I missed anything else. Please guide as needed. I saw some analysis exception with legacy setting for negative scale when trying to sort arrays with decimal type which I am not fully aware on where we stand in terms of support so I excluded those in that particular test.


integration_tests/src/main/python/data_gen.py
```diff
@@ -772,6 +772,7 @@ def gen_scalars_for_sql(data_gen, count, seed=0, force_no_nulls=False):
 boolean_gens = [boolean_gen]
 
 single_level_array_gens = [ArrayGen(sub_gen) for sub_gen in all_basic_gens + decimal_gens + [null_gen]]
+single_level_array_gens_non_decimal = [ArrayGen(sub_gen) for sub_gen in all_basic_gens + [null_gen]]
```
revans2, line 775: Why not include decimal?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2321,7 +2321,8 @@ object GpuOverrides {
       }),
     exec[ShuffleExchangeExec](
       "The backend for most data being exchanged between processes",
-      ExecChecks(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all),
+      ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.ARRAY +
```
revans2, line 2324: Technically this allows arrays of arrays, arrays of structs, and decimals in there too.  I don't have a problem with that so long as we have tested it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2372,7 +2373,8 @@ object GpuOverrides {
       (agg, conf, p, r) => new GpuSortAggregateMeta(agg, conf, p, r)),
     exec[SortExec](
       "The backend for the sort operator",
-      ExecChecks(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all),
+      ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.ARRAY +
+        TypeSig.STRUCT).nested(), TypeSig.all),
```
revans2, line 2377: Technically this allows arrays of arrays, arrays of structs, and decimals in there too. I don't have a problem with that so long as we have tested it.

We might want a comment here about `SortOrder` will restrict what we actually allow to be sorted by and other types are only shuffled.

revans2 (commented):
Is there a reason we don't want to allow for maps too? as they are just an array of structs (assuming that we have tested that it works)

integration_tests/src/main/python/data_gen.py
kuhushukla: > I saw some analysis exception with legacy setting for negative scale when trying to sort arrays with decimal type which I am not fully aware on where we stand in terms of support so I excluded those in that particular test.
>

I will check it out further

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
kuhushukla: I will add more tests.

kuhushukla:
build

revans2 (dismissed):
Looks good, but needs an upmerge (sorry conflicts with something I just merged in)

revans2 (dismissed):

kuhushukla:
build

revans2 (approved):

PR #1484, opened 2021-01-09
add doc for nsight systems bundled with cuda toolkit

rongou:

This is messier than I thought. There is a shell script in the cuda bin directory, but it doesn't work. In CUDA 11.2 they renamed the binary to `nsys-ui`. Anyway, I only listed the 3 versions we currently support.

@sameerz @gerashegalov

### DIFF
--- a/docs/dev/README.md
+++ b/docs/dev/README.md
@@ -246,6 +246,12 @@ added to the profile via NVTX ranges.  See
 [the NVTX profiling guide](nvtx_profiling.md) for additional information on
 setting up the build for profiling and adding NVTX ranges.
 
+> Note: The Nsight Systems GUI is installed as part of the CUDA Toolkit, but is
+> under a slightly different name depending on the CUDA version:
+> * `/usr/local/cuda-10.1/nsight-systems-2019.3.7.5/Host-x86_64/nsight-sys`
+> * `/usr/local/cuda-10.2/nsight-systems-2019.5.2/host-linux-x64/nsight-sys`
+> * `/usr/local/cuda-11.0/nsight-systems-2020.3.2/host-linux-x64/nsight-sys`
+
 ## Code Coverage
 
 We use [jacoco](https://www.jacoco.org/jacoco/trunk/doc/) for code coverage because it lets
--- a/docs/dev/README.md
+++ b/docs/dev/README.md
@@ -246,11 +246,13 @@ added to the profile via NVTX ranges.  See
 [the NVTX profiling guide](nvtx_profiling.md) for additional information on
 setting up the build for profiling and adding NVTX ranges.
 
-> Note: The Nsight Systems GUI is installed as part of the CUDA Toolkit, but is
-> under a slightly different name depending on the CUDA version:
-> * `/usr/local/cuda-10.1/nsight-systems-2019.3.7.5/Host-x86_64/nsight-sys`
-> * `/usr/local/cuda-10.2/nsight-systems-2019.5.2/host-linux-x64/nsight-sys`
-> * `/usr/local/cuda-11.0/nsight-systems-2020.3.2/host-linux-x64/nsight-sys`
+Note: Nsight Systems can be downloaded from its
+[product page](https://developer.nvidia.com/nsight-systems) directly.
+It is also installed as part of the CUDA Toolkit. To launch the GUI, run the following
+executable, depending on your CUDA Toolkit version:
+* `/usr/local/cuda-10.1/nsight-systems-2019.3.7.5/Host-x86_64/nsight-sys`
+* `/usr/local/cuda-10.2/nsight-systems-2019.5.2/host-linux-x64/nsight-sys`
+* `/usr/local/cuda-11.0/nsight-systems-2020.3.2/host-linux-x64/nsight-sys`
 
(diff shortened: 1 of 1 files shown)


sameerz (commented):
Is this completely up to date, or are these examples?  That is, https://developer.nvidia.com/gameworksdownload#?dn=nsight-systems-2020-5-1-85 appears to show a more recent version of nsight.  When I installed https://developer.nvidia.com/rdp/assets/nsight-systems-2020-5h-linux-deb-installer on Ubuntu 18.04 it installed in /opt/nvidia/nsight-systems/

rongou:
These are the bundled versions of Nsight Systems in CUDA toolkits. You can also install them separately.

docs/dev/README.md
```diff
@@ -246,6 +246,12 @@ added to the profile via NVTX ranges.  See
 [the NVTX profiling guide](nvtx_profiling.md) for additional information on
 setting up the build for profiling and adding NVTX ranges.
 
+> Note: The Nsight Systems GUI is installed as part of the CUDA Toolkit, but is
```
jlowe, line 249: Curious, why is this using a quoted block?

rongou: I guess it's just a side note in case someone wants to use the bundled version instead of downloading the standalone one.

jlowe: That's what I thought is covered by the "Note:" part.  Having it rendered as a quoted block makes it look like this was ripped from some other document which is a bit weird.  We're not quoting anything here.

rongou: Ah I see what you mean. I think I was after the visual effect so it's less intrusive. I'll remove the quoted block.

rongou: Removed the quoted block and reworded it a bit. PTAL

jlowe (approved):

jlowe:
build

docs/dev/README.md
```diff
@@ -246,6 +246,14 @@ added to the profile via NVTX ranges.  See
...
 setting up the build for profiling and adding NVTX ranges.
 
+Note: Nsight Systems can be downloaded from its
+[product page](https://developer.nvidia.com/nsight-systems) directly.
+It is also installed as part of the CUDA Toolkit. To launch the GUI, run the following
+executable, depending on your CUDA Toolkit version:
+* `/usr/local/cuda-10.1/nsight-systems-2019.3.7.5/Host-x86_64/nsight-sys`
+* `/usr/local/cuda-10.2/nsight-systems-2019.5.2/host-linux-x64/nsight-sys`
+* `/usr/local/cuda-11.0/nsight-systems-2020.3.2/host-linux-x64/nsight-sys`
+
```
gerashegalov, line 256: with CUDA 11.2 I have on Ubuntu 18.4 I have:
```bash
$ which nsight-sys 
/usr/local/cuda-11.2/bin/nsight-sys
```

However, when I call it I get 
```
$ /usr/local/cuda-11.2/bin/nsight-sys
Error: Nsight Systems 2020.4.3 hasn't been installed with CUDA Toolkit 11.2
```

I wonder how much we should advertise this in the doc or need to add additional [caveats](https://forums.developer.nvidia.com/t/unable-to-install-nsys-nsight-system-cli/160704).

rongou: Right, the wrapper script doesn't work, that's why I listed the executables. They also changed the name to `nsys-ui` for cuda 11.2. Like I said, it's kinda a mess.

gerashegalov: ack, nsys-ui in my install simply calls nsight-sys script in the same dir internally
```bash
$ cat /usr/local/cuda-11.2/bin/nsys-ui
#!/bin/bash
DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
exec "$DIR/nsight-sys"
```
It turns out that I actually don't have /usr/local/cuda-11.*/nsight-systems dir in my install
```
~$ find /usr/local/cuda-11.2/ -name nsight\*
/usr/local/cuda-11.2/bin/nsight-sys
/usr/local/cuda-11.2/bin/nsight_ee_plugins_manage.sh
/usr/local/cuda-11.2/nsightee_plugins
```

rongou: That's strange. It should have been installed as part of the CUDA Toolkit: https://docs.nvidia.com/nsight-systems/InstallationGuide/index.html

rongou:
build

pxLi:
build

docs/dev/README.md
jrhemstad: I strongly suggest encouraging people to always use the latest Nsight Systems rather than using what is bundled in the toolkit. You can miss out on a lot of important features and bug fixes by using what's in the toolkit. It's easy to download via `apt` or from the download site

rongou: Good point. Changed the wording to recommend direct download.

gerashegalov (approved):
:ship:

abellina (approved):

jlowe (approved):

jlowe:
build

sameerz:
build

rongou:
build

pxLi:
build

Issue #1487, opened 2021-01-11
[FEA] Change spark 3.1.0 to 3.1.1


tgravescs:
**Is your feature request related to a problem? Please describe.**
Spark messed up the release of 3.1.0 so it going straight to 3.1.1, we need to update our shim layer to support.

PR #1488, opened 2021-01-11
Rename Spark 3.1.0 shim to Spark 3.1.1 to match community

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/1487

Rename the shim to match the community versions. There won't be a 3.1.0 so no reason to keep that shim. I left the 3.1.0-SNAPSHOT as supported in the service provider list for now in case anyone was using those snapshots.


tgravescs:
build

jlowe (approved):

revans2 (approved):
My only concern is that `ParquetMaterializer` was moved, but it is not in a package that has the shim name in it, so if it changes again it might cause issues in the future, but this did nothing to cause that issue or fix it so it is fine.

tgravescs:
@razajafri any reason the ParquetMaterializer wasn't in a shim specific package?  if not we should file an issue and put it in one. It can still be in the spark package for example we have: spark311/src/main/scala/org/apache/spark/sql/rapids/shims/spark311/GpuColumnarToRowTransitionExec.scala

razajafri:
When I created the file I added a `rapids` to the end of the package name to make it obvious that it's a rapids file. I can file an issue to add `shims.spark311` to the end of the existing package. Will that work?

tgravescs:
yes please match the other files in the shim with shims.spark311

PR #1489, opened 2021-01-11
Adds in some support for the array sql function

revans2:
This depends on https://github.com/rapidsai/cudf/pull/7112 being merged in first. I did this mostly so I could run some tests to verify 7112 which is needed for another project.  My main concern with this is that if we have too many columns we can blow the size limit on the nested array fairly quickly. I can fix this with a little more work if others are concerned about it, otherwise I will just file a follow on issue for it.

I didn't put in support for nested types because I didn't want to spend the time testing them and figuring out if they work properly when there is no scalar support for them yet. I'll also include this in a follow on issue.

### DIFF
--- a/integration_tests/src/main/python/array_test.py
+++ b/integration_tests/src/main/python/array_test.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020-2021, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/integration_tests/src/main/python/array_test.py
+++ b/integration_tests/src/main/python/array_test.py
@@ -46,3 +46,12 @@ def test_nested_array_index(data_gen):
                 'a[1]',
                 'a[3]',
                 'a[50]'))
+
+
+@pytest.mark.parametrize('data_gen', all_basic_gens + [decimal_gen_default, decimal_gen_scale_precision], ids=idfn)
+def test_make_array(data_gen):
+    (s1, s2) = gen_scalars_for_sql(data_gen, 2, force_no_nulls=not isinstance(data_gen, NullGen))
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : binary_op_df(spark, data_gen).selectExpr(
+                'array(a, b)',
+                'array(b, a, null, {}, {})'.format(s1, s2)))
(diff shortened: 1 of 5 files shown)


revans2:
build

integration_tests/src/main/python/array_test.py
```diff
@@ -46,3 +46,12 @@ def test_nested_array_index(data_gen):
                 'a[1]',
```
jlowe, line 46: 2021 copyrights

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -18,14 +18,71 @@ package org.apache.spark.sql.rapids
...
+    withResource(new Array[ColumnVector](children.size)) { columns =>
+      val numRows = batch.numRows()
+      children.indices.foreach { index =>
+        children(index).columnarEval(batch) match {
+          case cv: GpuColumnVector =>
+            columns(index) = cv.getBase
+          case other =>
+            val dt = dataType.elementType
+            withResource(GpuScalar.from(other, dt)) { scalar =>
+              columns(index) = ColumnVector.fromScalar(scalar, numRows)
```
jlowe, line 75: Should there be a TODO comment and/or followup issue to avoid manifesting splatted scalar columns for each list entry when libcudf supports nested scalar types which could be splatted directly?

jlowe (commented):
A few small comments but otherwise lgtm.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
revans2: Spark will already take care of this for us.  `array(1, 2, 3)` is already folded into a scalar array, which then we have to fall back to the CPU for because we cannot splat it.  This code is only for the mixed case where we have a mixture of columns and scalar values.  Unfortunately the underlying CUDF API also does not support scalars so we could not even save memory and avoid splatting the individual scalar values, not even in a detailed API.  I am happy to file a follow on issue if that is something that we feel would be good to try and support.

revans2:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
jlowe: > array(1, 2, 3) is already folded into a scalar array, which then we have to fall back to the CPU for because we cannot splat it.

Ouch, that's unfortunate.  Even more cause to seek nested type scalar support in libcudf.

> I am happy to file a follow on issue if that is something that we feel would be good to try and support.

It's probably not worth it yet.  We're already splatting scalars into columns everywhere else.  Let's see if that ends up being an issue in practice, and we already have a separate issue to track the general splatting of scalars to columns in expressions.

jlowe (approved):

revans2:
build

Issue #1333, opened 2020-12-08
[FEA] Support window operations on Decimal


revans2:
**Is your feature request related to a problem? Please describe.**
This is still blocked on CUDF adding in support for decimal window operations.  The window operations currently supported include: 
  - [x] lead / lag
  - [x] count
  - [x] sum rapidsai/cudf#7117
  - [x] row number
  - [x] max / min
  - [x] collect list

mythrocks:
CUDF Decimal support in window functions:
`min`, `max`, `count`, `lead`, `lag`: https://github.com/rapidsai/cudf/pull/7037
`row_number`: https://github.com/rapidsai/cudf/pull/7061

sperlingxx:
What does "collect list" point to?

sperlingxx:
I did some tests with cuDF java tests. I found that all window operations required by current issue can run on decimal columns except `sum`. 
`sum` on decimal is still unsupported by cuDF because rollingWindow with sum operation is yet supported on decimal:  https://github.com/rapidsai/cudf/blob/branch-0.18/cpp/src/rolling/rolling_detail.hpp#L54

revans2:
Thanks, then we should split this up into at least 2 issues.  The first is for what we can support today. The second should be a follow on to implement sum, and then we can make sure that there is an issue in CUDF to track it.  I asked @razajafri to take lead on making sure that all of the issues we need are in cudf so please coordinate on that.

The final issue is for collect_list which @mythrocks is working on a general cudf implementation for (we do not currently support it for any window operations).

razajafri:
rapidsai/cudf#7061 and rapidsai/cudf#7037 are both merged in and can be supported by the plugin. I will create an issue for `sum` in cudf. As for `collect_list` @mythrocks is working on it.  I spoke with him and he will create an issue in cudf to track that work

razajafri:
I have opened a cudf issue to track `sum` in a window operation rapidsai/cudf#7117

revans2:
Lead, Lag, Max, and Min were merged in as a part of #1512 
Count was merged in as a part of #1476 but it was a mistake because it was intended to only go in for regular aggregate operations so tests for it were added as a part of #1512

Both sum and row_number are are merged into cudf so we should be able to get those merged in shortly.

sperlingxx:
[...]

PR #1512, opened 2021-01-14
Support window operations on Decimal

sperlingxx:
This pull request is to address issue #1333. With this pull request, most of window operations required by TPC-DS can run on decimal type data except `sum` and `collect_list`.


sperlingxx:
build

integration_tests/src/main/python/window_function_test.py
```diff
@@ -68,11 +68,44 @@ def test_window_aggs_for_rows(data_gen):
...
+    ('a', RepeatSeqGen(LongGen(nullable=(True, 10.0)), length=20)),
+    ('b', DecimalGen(precision=18, scale=10, nullable=True)),
+    ('c', IntegerGen())]
+
+
+@ignore_order
+@pytest.mark.parametrize('data_gen', [_grpkey_longs_with_decimals,
+                                      _grpkey_longs_with_nullable_decimals], ids=idfn)
+def test_window_aggs_for_rows_on_decimal(data_gen):
+    # TODO: test sum aggregation when underlying cuDF implementation is ready
```
revans2, line 86: nit: If there is a follow on issue for this it would be great to put the issue number here too.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -722,13 +722,16 @@ object GpuOverrides {
...
         Seq(ParamCheck("windowFunction",
-          TypeSig.commonCudfTypes + TypeSig.ARRAY.nested(TypeSig.commonCudfTypes),
+          TypeSig.commonCudfTypes + TypeSig.DECIMAL +
+            TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL),
           TypeSig.all),
-          ParamCheck("windowSpec", TypeSig.CALENDAR + TypeSig.NULL + TypeSig.integral,
-            TypeSig.numericAndInterval))),
+          ParamCheck("windowSpec",
+            TypeSig.CALENDAR + TypeSig.NULL + TypeSig.integral + TypeSig.DECIMAL,
+            TypeSig.numericAndInterval + TypeSig.DECIMAL))),
```
revans2, line 734: nit: `numericAndInterval` already covers `DECIMAL` so this change is not needed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1660,8 +1663,9 @@ object GpuOverrides {
     expr[Max](
       "Max aggregate operator",
       ExprChecks.fullAgg(
-        TypeSig.commonCudfTypes + TypeSig.NULL, TypeSig.orderable,
-        Seq(ParamCheck("input", TypeSig.commonCudfTypes + TypeSig.NULL, TypeSig.orderable))),
+        TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL, TypeSig.orderable,
```
revans2, line 1666: If we are going to add DECIMAL for MAX in all cases we need some tests to verify that it will work for groupby and reduction use cases as well.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1677,8 +1681,9 @@ object GpuOverrides {
     expr[Min](
       "Min aggregate operator",
       ExprChecks.fullAgg(
-        TypeSig.commonCudfTypes + TypeSig.NULL, TypeSig.orderable,
-        Seq(ParamCheck("input", TypeSig.commonCudfTypes + TypeSig.NULL, TypeSig.orderable))),
+        TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL, TypeSig.orderable,
```
revans2, line 1684: If we are going to add DECIMAL for MIN in all cases we need some tests to verify that it will work for groupby and reduction use cases as well.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -643,7 +643,7 @@ object CaseWhenCheck extends ExprChecks {
  * This is specific to WidowSpec, because it does not follow the typical parameter convention.
  */
 object WindowSpecCheck extends ExprChecks {
-  val check: TypeSig = TypeSig.commonCudfTypes
+  val check: TypeSig = TypeSig.commonCudfTypes + TypeSig.DECIMAL
```
revans2, line 646: This says that we should be able to partition by decimal columns and order by them as well, but we have no tests for that.  Could we update/add a few tests to verify that we are covering this use case?

revans2 (commented):
Looks good mostly nits and a bit more testing.

sperlingxx:
build

integration_tests/src/main/python/window_function_test.py
sperlingxx: Issue link has been added.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: Fixed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: The TypeChecks of `Max` and `Min` has been refined. In the updated version, `DECIMAL` TypeSig is only supported in `WindowAggExprContext`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
sperlingxx: I think existing tests has already covered `partition` and `order`.
 In unit test,  `windowAggregationTesterForDecimal` method will cast column for partition and column for order to decimal before window aggregation.
In integration test, `test_multi_types_window_aggs_for_rows_lead_lag` method covers `LEAD`/`LAG` operations, which includes decimal type in both `part_and_order_gens` and `data_gens`.

sperlingxx:
build

revans2 (approved):

PR #1531, opened 2021-01-15
Decimal Support for writing Parquet

razajafri:
This PR adds support for writing Decimal types to Parquet file. 

There is an issue in cudf at the time of writing this PR (rapidsai/cudf#7152) i.e. Decimals with precision < 10 are not able to be read back using Spark-cpu. 

This depends on rapidsai/cudf#7153

### DIFF
--- a/integration_tests/src/main/python/parquet_write_test.py
+++ b/integration_tests/src/main/python/parquet_write_test.py
@@ -20,6 +20,7 @@ from data_gen import *
 from marks import *
 from pyspark.sql.types import *
 from spark_session import with_cpu_session, with_gpu_session, is_before_spark_310
+import random
 
 # test with original parquet file reader, the multi-file parallel reader for cloud, and coalesce file reader for
 # non-cloud
@@ -32,9 +33,22 @@ reader_opt_confs = [original_parquet_file_reader_conf, multithreaded_parquet_fil
 writer_confs={'spark.sql.legacy.parquet.datetimeRebaseModeInWrite': 'CORRECTED',
               'spark.sql.legacy.parquet.int96RebaseModeInWrite': 'CORRECTED'}
 
+# https://github.com/rapidsai/cudf/issues/7152
+# The above bug in cudf causes problem writing Decimals with precision < 10
+random.seed(23423)
 parquet_write_gens_list = [
-        [byte_gen, short_gen, int_gen, long_gen, float_gen, double_gen,
-            string_gen, boolean_gen, date_gen, timestamp_gen]]
+    [byte_gen, short_gen, int_gen, long_gen, float_gen, double_gen,
+     string_gen, boolean_gen, date_gen, timestamp_gen,
+     DecimalGen(precision=10, scale=random.randint(1,10)),
+     DecimalGen(precision=11, scale=random.randint(1,11)),
+     DecimalGen(precision=12, scale=random.randint(1,12)),
+     DecimalGen(precision=13, scale=random.randint(1,13)),
+     DecimalGen(precision=14, scale=random.randint(1,14)),
+     DecimalGen(precision=15, scale=random.randint(1,15)),
+     DecimalGen(precision=16, scale=random.randint(1,16)),
+     DecimalGen(precision=17, scale=random.randint(1,17)),
+     DecimalGen(precision=18, scale=random.randint(1,18))]
+]
 
 parquet_ts_write_options = ['INT96', 'TIMESTAMP_MICROS', 'TIMESTAMP_MILLIS']
 
@@ -42,6 +56,7 @@ parquet_ts_write_options = ['INT96', 'TIMESTAMP_MICROS', 'TIMESTAMP_MILLIS']
 @pytest.mark.parametrize('reader_confs', reader_opt_confs)
 @pytest.mark.parametrize('v1_enabled_list', ["", "parquet"])
 @pytest.mark.parametrize('ts_type', parquet_ts_write_options)
+@allow_non_gpu("CoalesceExec")
 def test_write_round_trip(spark_tmp_path, parquet_gens, v1_enabled_list, ts_type, reader_confs):
     gen_list = [('_c' + str(i), gen) for i, gen in enumerate(parquet_gens)]
     data_path = spark_tmp_path + '/PARQUET_DATA'
@@ -112,6 +127,7 @@ def test_compress_write_round_trip(spark_tmp_path, compress, v1_enabled_list, re
 
 @pytest.mark.parametrize('parquet_gens', parquet_write_gens_list, ids=idfn)
 @pytest.mark.parametrize('ts_type', parquet_ts_write_options)
+@allow_non_gpu("CoalesceExec")
 def test_write_save_table(spark_tmp_path, parquet_gens, ts_type, spark_tmp_table_factory):
     gen_list = [('_c' + str(i), gen) for i, gen in enumerate(parquet_gens)]
     data_path = spark_tmp_path + '/PARQUET_DATA'
--- a/integration_tests/src/main/python/parquet_write_test.py
+++ b/integration_tests/src/main/python/parquet_write_test.py
@@ -47,7 +47,7 @@ parquet_ts_write_options = ['INT96', 'TIMESTAMP_MICROS', 'TIMESTAMP_MILLIS']
 @pytest.mark.parametrize('reader_confs', reader_opt_confs)
 @pytest.mark.parametrize('v1_enabled_list', ["", "parquet"])
 @pytest.mark.parametrize('ts_type', parquet_ts_write_options)
-def test_parquet_write_round_trip(spark_tmp_path, parquet_gens, v1_enabled_list, ts_type,
+def test_write_round_trip(spark_tmp_path, parquet_gens, v1_enabled_list, ts_type,
                                   reader_confs):
     gen_list = [('_c' + str(i), gen) for i, gen in enumerate(parquet_gens)]
     data_path = spark_tmp_path + '/PARQUET_DATA'
@@ -118,7 +118,6 @@ def test_compress_write_round_trip(spark_tmp_path, compress, v1_enabled_list, re
 
 @pytest.mark.parametrize('parquet_gens', parquet_write_gens_list, ids=idfn)
 @pytest.mark.parametrize('ts_type', parquet_ts_write_options)
-@allow_non_gpu("CoalesceExec")
 def test_write_save_table(spark_tmp_path, parquet_gens, ts_type, spark_tmp_table_factory):
     gen_list = [('_c' + str(i), gen) for i, gen in enumerate(parquet_gens)]
     data_path = spark_tmp_path + '/PARQUET_DATA'
--- a/integration_tests/src/main/python/parquet_write_test.py
+++ b/integration_tests/src/main/python/parquet_write_test.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020-2021, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuOrcFileFormat.scala
```diff
@@ -44,7 +44,13 @@ object GpuOrcFileFormat extends Logging {
 
   def tagGpuSupport(meta: RapidsMeta[_, _, _],
                     spark: SparkSession,
-                    options: Map[String, String]): Option[GpuOrcFileFormat] = {
+                    options: Map[String, String],
+                    schema: StructType): Option[GpuOrcFileFormat] = {
+
+
+    if(!schema.forall(field => GpuOverrides.isSupportedType(field.dataType))) {
```
tgravescs, line 51: isn't this already checked with the rules:
```
ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.STRUCT + TypeSig.MAP +
            TypeSig.ARRAY + TypeSig.DECIMAL).nested(), TypeSig.all),
```

integration_tests/src/main/python/parquet_write_test.py
```diff
@@ -32,16 +33,30 @@
 writer_confs={'spark.sql.legacy.parquet.datetimeRebaseModeInWrite': 'CORRECTED',
               'spark.sql.legacy.parquet.int96RebaseModeInWrite': 'CORRECTED'}
 
+# https://github.com/rapidsai/cudf/issues/7152
```
tgravescs, line 36: we could write the tests and just xfail them and we should file a followup issue to track it.

integration_tests/src/main/python/parquet_write_test.py
```diff
@@ -32,16 +33,30 @@
...
+     DecimalGen(precision=18, scale=random.randint(1,18))]
+]
 
 parquet_ts_write_options = ['INT96', 'TIMESTAMP_MICROS', 'TIMESTAMP_MILLIS']
 
 @pytest.mark.parametrize('parquet_gens', parquet_write_gens_list, ids=idfn)
 @pytest.mark.parametrize('reader_confs', reader_opt_confs)
 @pytest.mark.parametrize('v1_enabled_list', ["", "parquet"])
 @pytest.mark.parametrize('ts_type', parquet_ts_write_options)
+@allow_non_gpu("CoalesceExec")
```
tgravescs, line 59: why is this?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -41,6 +41,11 @@ object GpuParquetFileFormat {
       spark: SparkSession,
       options: Map[String, String],
       schema: StructType): Option[GpuParquetFileFormat] = {
+
+    if(!schema.forall(field => GpuOverrides.isSupportedType(field.dataType, allowDecimal = true))) {
```
tgravescs, line 45: nit space after if

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuOrcFileFormat.scala
tgravescs, line 51: nit space after if

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuOrcFileFormat.scala
```diff
@@ -44,7 +44,13 @@ object GpuOrcFileFormat extends Logging {
...
+      meta.willNotWorkOnGpu("Not all datatypes are supported")
```
tgravescs, line 52: in the very least print the data types here or tell us which one isn't supported

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2270,7 +2270,8 @@ object GpuOverrides {
       }),
     exec[DataWritingCommandExec](
       "Writing data",
-      ExecChecks(TypeSig.commonCudfTypes, TypeSig.all),
+      ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.STRUCT + TypeSig.MAP +
+        TypeSig.ARRAY + TypeSig.DECIMAL).nested(), TypeSig.all),
```
tgravescs, line 2274: what is the difference in this check and the file format specific ones?  is the schema at that point different than the plan output here?

integration_tests/src/main/python/parquet_write_test.py
razajafri: hmm... ok. I will see if I can do that and still use our data_gens

integration_tests/src/main/python/parquet_write_test.py
razajafri: Because `CoalesceExec` doesn't support `Decimals` yet

razajafri: Do we want to rather write a separate test for Decimals so we are at least testing coalesce with other types? we already have other tests testing coalesce though

tgravescs: yes, I don't want it to be ok if other types have that

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
razajafri: This check is useless in a way for `DataWritingCommandExec` as we are passing it for all datatypes and we will handle each type separately in the specific formats. 

Schema will be the same but we don't need to make the checks based on schema in most of the Execs therefore we are doing it inside the formats

jlowe: It's not useless, correct?  We need to pass this check before we pass the check for the scans, and this generates the documentation that we support decimal for this exec (with the caveat that it only works for some formats).  Speaking of caveats, this should use the `withPsNote` functionality in TypeChecks so we add the caveat that decimal type may not be supported by the specific output format.

jlowe:
Converting this to draft since it depends on a cudf change that is still pending.

jlowe (changes requested):
This needs to update the static documentation in TypeChecks to update the Input/Output table to specify Parquet supports decimals.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
razajafri: Yes, sorry, not useless at all. 

Great point for using `withPsNote`. I will make the change

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -1312,7 +1312,7 @@ object SupportedOpsDocs {
     println("<td>S</td>") // DATE
     println("<td>S</td>") // TIMESTAMP
     println("<td>S</td>") // STRING
-    println("<td><b>NS</b></td>") // DECIMAL
+    println("<td>S</td>") // DECIMAL
```
jlowe, line 1315: supported_ops.md was not regenerated after this change.

jlowe (changes requested):

integration_tests/src/main/python/parquet_write_test.py
```diff
@@ -33,16 +34,21 @@
...
+                 marks=pytest.mark.allow_non_gpu("CoalesceExec"))]
 
 parquet_ts_write_options = ['INT96', 'TIMESTAMP_MICROS', 'TIMESTAMP_MILLIS']
 
 @pytest.mark.parametrize('parquet_gens', parquet_write_gens_list, ids=idfn)
 @pytest.mark.parametrize('reader_confs', reader_opt_confs)
 @pytest.mark.parametrize('v1_enabled_list', ["", "parquet"])
 @pytest.mark.parametrize('ts_type', parquet_ts_write_options)
-def test_write_round_trip(spark_tmp_path, parquet_gens, v1_enabled_list, ts_type, reader_confs):
+def test_parquet_write_round_trip(spark_tmp_path, parquet_gens, v1_enabled_list, ts_type,
```
tgravescs, line 50: do we really need parquet in the name when its in parquet_write_test?  just make it wrap inputs

integration_tests/src/main/python/parquet_write_test.py
```diff
@@ -112,6 +118,7 @@ def test_compress_write_round_trip(spark_tmp_path, compress, v1_enabled_list, re
 
 @pytest.mark.parametrize('parquet_gens', parquet_write_gens_list, ids=idfn)
 @pytest.mark.parametrize('ts_type', parquet_ts_write_options)
+@allow_non_gpu("CoalesceExec")
```
tgravescs, line 121: assume this isn't needed if in parquet_write_gens_list?

integration_tests/src/main/python/parquet_write_test.py
razajafri: Changed it. I made the change to be able to execute the test by name otherwise it would run the orc test as well.

tgravescs: ok that is fine. you can also specify the test file. src/main/python/parquet_write_test.py -k test_write_round_trip

razajafri: Thanks, that's really helpful for future.

razajafri:
@jlowe I have converted this from a draft as the cudf issue is merged. Appreciate the review. Can you PTAL?

jlowe (dismissed):

jlowe:
build

jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -41,6 +41,13 @@ object GpuParquetFileFormat {
       spark: SparkSession,
       options: Map[String, String],
       schema: StructType): Option[GpuParquetFileFormat] = {
+
+    val unSupportedTypes =
+      schema.filter(field => !GpuOverrides.isSupportedType(field.dataType, allowDecimal = true))
+    if (!unSupportedTypes.isEmpty) {
```
gerashegalov, line 47: consider `if (unSupportedTypes.nonEmpty)`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -247,31 +254,45 @@ class GpuParquetWriter(
...
+        }
+      })
 
     super.write(newBatch, statsTrackers)
   }
   override val tableWriter: TableWriter = {
+    def precisionsList(t: DataType): List[Int] = {
+      t match {
+        case d: DecimalType => List(d.precision)
+        case s: StructType => s.flatMap(f => precisionsList(f.dataType)).toList
```
gerashegalov, line 284: we could save the conversion `toList` if the return type for `precisionsList` were `Seq[Int]`

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuOrcFileFormat.scala
```diff
@@ -44,7 +44,13 @@ object GpuOrcFileFormat extends Logging {
 
   def tagGpuSupport(meta: RapidsMeta[_, _, _],
                     spark: SparkSession,
-                    options: Map[String, String]): Option[GpuOrcFileFormat] = {
+                    options: Map[String, String],
+                    schema: StructType): Option[GpuOrcFileFormat] = {
+
+    val unSupportedTypes = schema.filter(field => !GpuOverrides.isSupportedType(field.dataType))
```
gerashegalov, line 50: looks the same as in GpuParquetFileFormat.scala, consider a shared utils.

razajafri: Its not exactly the same. The predicate being tested is different. If you still think we can benefit from refactoring this I can do it

razajafri:
@gerashegalov Thanks for the review I have incorporated most of your comments in the PR. The only thing that I haven't done is the utility method.

razajafri:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuOrcFileFormat.scala
gerashegalov: the difference is parameterizable but it's not that big of a deal given it's just a few lines, up to you

gerashegalov (approved):
LGTM

Issue #1330, opened 2020-12-08
[FEA] Support Decimal Casts


revans2:
**Is your feature request related to a problem? Please describe.**
Cudf supports casting decimal to/from double, float, int, long, etc.  We should support these too.  We need to be careful to test double/float especially because there can be round off issues that we need to test to see if they are correct.

sameerz:
Tagging @sperlingxx on this issue.

sameerz:
PR #1438 addresses casting with the exception of decimal to decimal casts, we need one more PR to address decimal to decimal casting.

PR #1532, opened 2021-01-15
Support casting from decimal to decimal

sperlingxx:
Current pull request is created to address the remaining part of #1330, casting from decimal  to another decimal type. 

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
@@ -380,84 +380,13 @@ case class GpuCast(
         input.getBase.asByteList(true)
 
       case (ShortType | IntegerType | LongType, dt: DecimalType) =>
-
-        // Use INT64 bounds instead of FLOAT64 bounds, which enables precise comparison.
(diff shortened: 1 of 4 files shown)


sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -917,4 +846,127 @@ case class GpuCast(
...
+        }
+      }
+    }
+  }
+
+  private def castDecimalToDecimal(input: ColumnVector,
+      from: DecimalType,
+      to: DecimalType): ColumnVector = {
+
+    val checkedInput = if (to.scale <= from.scale) {
```
revans2, line 936: Can we add some comments here, because it is was not clear at all to me that we were doing the cast in two phases, where we first check the for overflows and then do the casting.

revans2 (approved):
The code and tests look great. Just a minor nit on some extra comments I would like to see, but they are not critical.

sperlingxx:
build

revans2 (approved):

Issue #1510, opened 2021-01-13
[BUG] IllegalArgumentException during shuffle


rongou:
**Describe the bug**
When running TPC-DS benchmarks with the shuffle manager enabled (but no UCX), getting an exception
```console
java.lang.IllegalArgumentException: requirement failed:
Attempting to handle non-rapids enabled blocks from BlockManagerId(8, 127.0.0.1, 37359, None)
```

**Steps/Code to reproduce bug**
Here is the command line:
```console
/raid/spark-team/rou/spark/bin/spark-shell\
 --master spark://127.0.0.1:7077\
 --conf spark.serializer=org.apache.spark.serializer.KryoSerializer\
 --conf spark.kryo.registrator=com.nvidia.spark.rapids.GpuKryoRegistrator\
 --conf spark.locality.wait=0s\
 --conf spark.sql.files.maxPartitionBytes=1g\
 --conf spark.sql.shuffle.partitions=96\
 --conf spark.shuffle.manager=com.nvidia.spark.rapids.spark301.RapidsShuffleManager\
 --conf spark.shuffle.service.enabled=false\
 --conf spark.plugins=com.nvidia.spark.SQLPlugin\
 --conf spark.rapids.sql.concurrentGpuTasks=1\
 --conf spark.rapids.memory.pinnedPool.size=8G\
 --conf spark.rapids.sql.batchSizeBytes=1g\
 --conf spark.rapids.memory.gpu.direct.storage.spill.enabled=false\
 --conf spark.driver.memory=10G\
 --conf spark.driver.extraJavaOptions=-Dai.rapids.cudf.nvtx.enabled=false\
 --conf spark.executor.extraClassPath=/raid/spark-team/rou/rapids/cudf.jar:\
/raid/spark-team/rou/rapids/rapids-4-spark.jar\
 --conf spark.executor.extraJavaOptions=-Dai.rapids.cudf.nvtx.enabled=false\
 --conf spark.executor.instances=16\
 --conf spark.executor.cores=6\
 --conf spark.executor.memory=64G\
 --conf spark.executor.resource.gpu.amount=1\
 --conf spark.task.cpus=1\
 --conf spark.task.resource.gpu.amount=0.1666\
 --jars /raid/spark-team/rou/rapids/cudf.jar,\
/raid/spark-team/rou/rapids/rapids-4-spark.jar,\
/raid/spark-team/rou/rapids/rapids-4-spark-integration-tests.jar
```
and the query piped in:
[...]

PR #1543, opened 2021-01-17
Shuffle/transport enabled by default

abellina:
Closes https://github.com/NVIDIA/spark-rapids/issues/1510.

In the default, case `spark.rapids.shuffle.transport.enabled = true`, we will now see:

```
WARN RapidsShuffleManager: Rapids Shuffle Plugin enabled. Transport enabled (remote fetches will use com.nvidia.spark.rapids.shuffle.ucx.UCXShuffleTransport)
```
Expectations are that UCX must be installed, as that is our only transport.

When: `spark.rapids.shuffle.transport.enabled = false`, users will see the following message:

```
WARN RapidsShuffleManager: Rapids Shuffle Plugin enabled. Transport disabled (local cached blocks only).
```

I changed the exception message that will happen in case we have the transport disabled AND we end up needing to fetch remote blocks to:

```
java.lang.IllegalArgumentException: requirement failed: Attempting to handle non-rapids enabled blocks from BlockManagerId(3, 127.0.0.1, 43376, None). The shuffle transport is disabled. Please set spark.rapids.shuffle.transport.enabled=true to enable fetching remote blocks.
```


abellina:
build

abellina:
build

abellina:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/RapidsCachingReader.scala
```diff
@@ -106,8 +106,16 @@ class RapidsCachingReader[K, C](
           require(
             blockManagerId.topologyInfo.isDefined &&
               blockManagerId.topologyInfo.get
-                .startsWith(s"${RapidsShuffleTransport.BLOCK_MANAGER_ID_TOPO_PREFIX}="),
-            s"Attempting to handle non-rapids enabled blocks from $blockManagerId")
+                .startsWith(s"${RapidsShuffleTransport.BLOCK_MANAGER_ID_TOPO_PREFIX}="), {
+              val enabledHint = if (!rapidsConf.shuffleTransportEnabled) {
```
jlowe, line 110: I'm still confused why this is even a setting at all.  UCX is lazily loaded, so in the single-executor case the code doesn't try to do a remote shuffle and therefore doesn't load UCX.  What is the use-case where this property makes sense?

abellina: If we don't have this flag, then we will load UCX at block _write_ time. We are going to want to know then what the port for UCX is, so this creates the UCX class and tries to load those libraries.

I could split up having this port created, and actually initiating the library. But this is the reason for the flag. If it is set to false, we skip the "getting the port" step, and so our blocks don't carry this information when we commit map status.

jlowe (approved):

abellina:
build

pxLi:
build

pxLi:
build

pxLi:
build

PR #1546, opened 2021-01-18
Cleanup shutdown logging for UCX shuffle

### LINKED ISSUE #1155: [BUG] on shutdown don't print `Socket closed` exception when shutting down UCX.scala
On shutdown you'll likely see the exception below logged. 

We should not log this if we are shutting down, and if we are alive and this happens, we should find ways of blowing up, since something has gone wrong.

```
java.net.SocketException: Socket closed
        at java.net.PlainSocketImpl.socketAccept(Native Method)
        at java.net.AbstractPlainSocketImpl.accept(AbstractPlainSocketImpl.java:409)
        at java.net.ServerSocket.implAccept(ServerSocket.java:560)
        at java.net.ServerSocket.accept(ServerSocket.java:528)
        at com.nvidia.spark.rapids.shuffle.ucx.UCX.$anonfun$startManagementPort$4(UCX.scala:217)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

abellina:
In this PR we clean up the logging in UCX scala such that we don't log a backtrace in the benign case that the management socket, which is constantly in `accept()`, is closed due to a normal tear down. 

Java will throw a `SocketException` in this case, to unblock the `accept()` call, and the application must distinguish this exception from other potential exceptions during the call.

This PR also cleans up code around an extra lock we had, and makes  the `initialized` volatile.

Closes #1155

### DIFF
--- a/shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
+++ b/shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2021, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -69,7 +69,7 @@ class UCX(executorId: Int, usingWakeupFeature: Boolean = true) extends AutoClose
 
   private var worker: UcpWorker = _
   private val endpoints = new ConcurrentHashMap[Long, UcpEndpoint]()
-  private var initialized = false
+  @volatile private var initialized = false
 
   // a peer tag identifies an incoming connection uniquely
   private val peerTag = new AtomicLong(0) // peer tags
@@ -148,17 +148,15 @@ class UCX(executorId: Int, usingWakeupFeature: Boolean = true) extends AutoClose
 
       while(initialized) {
         try {
-          if (initialized) {
-            worker.progress()
-            // else worker.progress returned 0
-            if (usingWakeupFeature) {
-              drainWorker()
-              val sleepRange = new NvtxRange("UCX Sleeping", NvtxColor.PURPLE)
-              try {
-                worker.waitForEvents()
-              } finally {
-                sleepRange.close()
-              }
+          worker.progress()
+          // else worker.progress returned 0
+          if (usingWakeupFeature) {
+            drainWorker()
+            val sleepRange = new NvtxRange("UCX Sleeping", NvtxColor.PURPLE)
+            try {
+              worker.waitForEvents()
+            } finally {
+              sleepRange.close()
             }
           }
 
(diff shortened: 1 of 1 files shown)


jlowe:
@abellina could you elaborate more on what's being fixed here in the PR description?

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -69,7 +69,8 @@ class UCX(executorId: Int, usingWakeupFeature: Boolean = true) extends AutoClose
 
   private var worker: UcpWorker = _
   private val endpoints = new ConcurrentHashMap[Long, UcpEndpoint]()
-  private var initialized = false
+  @volatile private var initialized = false
+  private[this] val initializedMonitor = new Object
```
jlowe, line 73: It would be useful to have some comments explaining why a separate monitor object is required rather than locing the `UCX` instance directly.

abellina: It isn't required to have a separate monitor. Locking the instance should be fine here, let me rework.

jlowe:
Would be good to update the PR headline as it's used as-is in the changelog.  As it is now, users aren't going to know what it's talking about.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
abellina: done.

abellina:
@jlowe let me know if this looks ok, I'll rebase it if so (and at least sign off the last commit, or squash). It looks like I confused the sign-off bot.

jlowe (dismissed):

abellina:
build

pxLi:
build

pxLi:
build

pxLi:
build

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -223,10 +221,15 @@ class UCX(executorId: Int, usingWakeupFeature: Boolean = true) extends AutoClose
...
-                if (!initialized) {
-                  logWarning(s"UCX management socket closing", e)
+                if (initialized) {
+                  // This will cause the `SparkUncaughtExceptionHandler` to get invoked
+                  // and it will shut down the executor (as it should).
+                  throw e
                 } else {
-                  logError(s"Got exception while waiting for a UCX management connection", e)
+                  // initialized = false means we are shutting down,
+                  // the socket will throw SocketException in this case
```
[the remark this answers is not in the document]
abellina: you got it! https://github.com/NVIDIA/spark-rapids/pull/1546/commits/502fb1b859f4cec4948435b2cdb88ae094fe7ae1

abellina:
build

abellina:
@gerashegalov let me know if this looks OK to you. Thanks!

gerashegalov (approved):
LGTM

PR #1548, opened 2021-01-19
Support executing collect_list on GPU with windowing.

firestarman:
This PR is to support running `collect_list ` on GPU with windowing.

This PR depends on the PR https://github.com/rapidsai/cudf/pull/7189 .

- [x] Add integration tests.


firestarman:
Mark as draft since it depends on the support from cuDF native.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -529,3 +529,42 @@ abstract class GpuLastBase(child: Expression)
   override lazy val deterministic: Boolean = false
   override def toString: String = s"gpulast($child)${if (ignoreNulls) " ignore nulls"}"
 }
+
+/**
+ * Collects and returns a list of non-unique elements.
+ *
+ * FIXME Not sure whether GPU version requires the two offset parameters. Keep it here first.
```
mythrocks, line 536: What are the two offset parameters?
The signature of the GPU `collect()` window function currently takes only a child expression (column).

firestarman: Yes, the cuDF `collect()` takes no parameters other than child expression.
This signature is copied from CPU version here .https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/collect.scala#L102
```scala
case class CollectList(
    child: Expression,
    mutableAggBufferOffset: Int = 0,
    inputAggBufferOffset: Int = 0) extends Collect[mutable.ArrayBuffer[Any]] {
```
I am fine to remove them if not required by GPU version.

mythrocks: Thanks for the clarification, @firestarman. Let us keep the signatures identical for now, lest they be flagged during an audit for interface compatibility. It might be alright not to use them for the moment.

firestarman:
rebased

firestarman:
build

firestarman:
build

firestarman:
rebased

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2170,6 +2171,19 @@ object GpuOverrides {
       (a, conf, p, r) => new UnaryExprMeta[MakeDecimal](a, conf, p, r) {
         override def convertToGpu(child: Expression): GpuExpression =
           GpuMakeDecimal(child, a.precision, a.scale, a.nullOnOverflow)
+      }),
+    expr[CollectList](
+      "Collect a list of elements, now only supported by windowing.",
+      /* It should be 'fullAgg' eventually but now only support windowing, so 'windowOnly' */
+      ExprChecks.windowOnly(
```
revans2, line 2178: We need to change this slightly because windowOnly is really meant for operations like lead and lag that will only ever work on window operations.  We need something new that will indicate that Spark supports this for more than just window operations, but aggregates and reductions too.

integration_tests/src/main/python/window_function_test.py
```diff
@@ -213,3 +213,53 @@ def test_window_aggs_for_ranges_of_dates(data_gen):
         '       range between 1 preceding and 1 following) as sum_c_asc '
         'from window_agg_table'
     )
+
+
+'''
+  Spark will drop nulls when collecting, but seems GPU does not, so exceptions come up.
```
revans2, line 219: We cannot merge this in as is. We missed a fundamental requirement when implementing this for cudf.  The best we can do for now is check if the column is nullable or not.  But Spark does not always calculate this properly, because it is being cautious. So we should just wait on this.

revans2 (changes requested):

integration_tests/src/main/python/window_function_test.py
revans2: I spoke with @sameerz about this and I think we can work around this for the 0.4 release.

We can check the input column to see if it is nullable or not.  If it is not nullable we can replace the collect list expression.  If it is nullable then we can check a config to see if we force it to run the collect list anyways.  Or we can just have collect_list off by default and let users enable it if they want this functionality before it is 100% correct.

I would like us to modify the tests with this so that in the short term we have a specific test for collect_list separate from the others so we don't stop testing nulls to enable this.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
firestarman: Good suggestion. 
I added a new check named `aggNotGroupByOrReduction` for the `collect_list` .

integration_tests/src/main/python/window_function_test.py
firestarman: Thanks for review, and I chose the later one to disable it by default.

Sorry I don't fully get your suggestion on how to modify the integration tests, could you pls explain more ?

firestarman:
build

integration_tests/src/main/python/window_function_test.py
```diff
@@ -213,3 +213,54 @@ def test_window_aggs_for_ranges_of_dates(data_gen):
...
+E      Caused by: java.lang.AssertionError:  value at 350 is null
+E          	at ai.rapids.cudf.HostColumnVectorCore.assertsForGet(HostColumnVectorCore.java:228)
+E         	at ai.rapids.cudf.HostColumnVectorCore.getInt(HostColumnVectorCore.java:254)
+E         	at com.nvidia.spark.rapids.RapidsHostColumnVectorCore.getInt(RapidsHostColumnVectorCore.java:109)
+E         	at org.apache.spark.sql.vectorized.ColumnarArray.getInt(ColumnarArray.java:128)
+
+  Now set nullable to false to pass the tests, once native supports dropping nulls, will set it to true.
+'''
+collect_data_gen = [
```
revans2, line 228: Can we please add the a link to the follow on issue to fix collect_list so we don't collect nulls? That way we can update the tests to allow for nulls again?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -199,13 +199,15 @@ case class GpuWindowExpression(windowFunction: Expression, windowSpec: GpuWindow
...
-    } else {
-      GpuColumnVector.from(aggColumn, windowFunc.dataType)
+    // For nested type, do not cast
+    aggColumn.getType match {
+      case dType if dType.isNestedType =>
+        GpuColumnVector.from(aggColumn, windowFunc.dataType)
+      case _ =>
+        val expectedType = GpuColumnVector.getNonNestedRapidsType(windowFunc.dataType)
+        withResource(aggColumn) { aggColumn =>
+          GpuColumnVector.from(aggColumn.castTo(expectedType), windowFunc.dataType)
```
revans2, line 209: It might be worth adding in a comment that if the types are the same then the cast just increments the reference count. Just so someone does not try to "fix" it in the future.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -996,6 +996,18 @@ object ExprChecks {
...
+   * An aggregation check where window operations are supported by the plugin, but Spark
+   * also supports group by and reduction on these.
+   * This is now really for 'collect_list' which is only supported by windowing.
+   */
+  def aggNotGroupByOrReduction(
+      outputCheck: TypeSig,
+      sparkOutputSig: TypeSig,
+      paramCheck: Seq[ParamCheck] = Seq.empty,
+      repeatingParamCheck: Option[RepeatingParamCheck] = None): ExprChecks =
+    windowOnly(outputCheck, sparkOutputSig, paramCheck, repeatingParamCheck)
```
revans2, line 1010: This is still incorrect.  We need something more like the following.  Note I have not tested this.

```
  def aggNotGroupByOrReduction(
      outputCheck: TypeSig,
      sparkOutputSig: TypeSig,
      paramCheck: Seq[ParamCheck] = Seq.empty,
      repeatingParamCheck: Option[RepeatingParamCheck] = None): ExprChecks = {
    val notWindowParamCheck = paramCheck.map { pc =>
      ParamCheck(pc.name, TypeSig.none, pc.spark)
    }
    val notWindowRepeat = repeatingParamCheck.map { pc =>
      RepeatingParamCheck(pc.name, TypeSig.none, pc.spark)
    }
    ExprChecksImpl(Map(
      (GroupByAggExprContext,
          ContextChecks(TypeSig.none, sparkOutputSig, notWindowParamCheck, notWindowRepeat)),
      (ReductionAggExprContext,
          ContextChecks(TypeSig.none, sparkOutputSig, notWindowParamCheck, notWindowRepeat)),
      (WindowAggExprContext,
          ContextChecks(outputCheck, sparkOutputSig, paramCheck, repeatingParamCheck))))
}
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -529,3 +529,42 @@ abstract class GpuLastBase(child: Expression)
...
+  // WINDOW FUNCTION
+  override val windowInputProjection: Seq[Expression] = Seq(child)
+  override def windowAggregation(inputs: Seq[(ColumnVector, Int)]): AggregationOnColumn =
+    Aggregation.collect().onColumn(inputs.head._2)
+
+  // Declarative aggregate. But for now 'CollectList' does not support it.
+  // The members as below should NOT be used yet, ensured by the
+  // "TypeCheck.aggNotGroupByOrReduction" when trying to override the expression.
+  private lazy val cudfList = AttributeReference("collect_list", dataType)()
+  override val initialValues: Seq[GpuExpression] = Seq.empty
```
revans2, line 564: Can we throw an exception here instead of Seq.empty?  I just want to be sure that we don't accidentally corrupt data if an API is called that we do not expect.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
revans2, line 536: Could you update the FIXME to indicate that the offsets are not used, but are here for compatibility with the CPU version and automated checks.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2170,7 +2171,22 @@ object GpuOverrides {
...
+      ExprChecks.aggNotGroupByOrReduction(
+        TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.STRUCT),
+        TypeSig.ARRAY.nested(TypeSig.all),
+        Seq(ParamCheck("input",
+          (TypeSig.commonCudfTypes + TypeSig.DECIMAL).nested() + TypeSig.STRUCT,
+          TypeSig.all))),
+      (c, conf, p, r) => new ExprMeta[CollectList](c, conf, p, r) {
+        override def convertToGpu(): GpuExpression = GpuCollectList(
+          childExprs.head.convertToGpu(), c.mutableAggBufferOffset, c.inputAggBufferOffset)
+      }).disabledByDefault("for now the GPU collects null values to a list, but Spark does not." +
```
revans2, line 2188: @sameerz I just wanted to check with you that this is an acceptable short term solution.  That we disable the operator and the user will have to explicitly set `spark.rapids.sql.expression.CollectList=true` to use it.

sameerz: Yes this works for now.

integration_tests/src/main/python/window_function_test.py
firestarman: Sure.
The issue is here https://github.com/NVIDIA/spark-rapids/issues/1638 . 
Added two tests for nullabe and non-nullable columns separately, and skipped the former one with the tracking issue above.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
firestarman: Good suggestion, updated.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
firestarman: Got it, updated, and verified it, and it works.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
firestarman: Good suggestion, updated.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
firestarman: Done

firestarman:
cudf nightly with PR https://github.com/rapidsai/cudf/pull/7189 always fail, so the prebuilt will fail too.

firestarman:
> cudf nightly with PR [rapidsai/cudf#7189](https://github.com/rapidsai/cudf/pull/7189) always fail, so the pre-merge check will fail too.

The fix is in review https://github.com/rapidsai/cudf/pull/7266

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2264,6 +2265,21 @@ object GpuOverrides {
...
       }),
+    expr[CollectList](
+      "Collect a list of elements, now only supported by windowing.",
+      /* It should be 'fullAgg' eventually but now only support windowing,
+         so 'aggNotGroupByOrReduction' */
+      ExprChecks.aggNotGroupByOrReduction(
+        TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.STRUCT),
+        TypeSig.ARRAY.nested(TypeSig.all),
+        Seq(ParamCheck("input",
+          (TypeSig.commonCudfTypes + TypeSig.DECIMAL).nested() + TypeSig.STRUCT,
```
revans2, line 2276: again I think longer is clearer, but in this case because there is only one nested type I think it is okay.  So we don't support doing a collect list on a struct that also contains a struct?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2604,7 +2620,9 @@ object GpuOverrides {
       (expand, conf, p, r) => new GpuExpandExecMeta(expand, conf, p, r)),
     exec[WindowExec](
       "Window-operator backend",
-      ExecChecks(TypeSig.commonCudfTypes + TypeSig.DECIMAL, TypeSig.all),
+      ExecChecks(
+        (TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.STRUCT).nested() + TypeSig.ARRAY,
```
revans2, line 2624: again I would prefer more verbose instead of less

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -534,3 +534,44 @@ abstract class GpuLastBase(child: Expression)
...
+ * with the CPU version and automated checks.
+ */
+case class GpuCollectList(child: Expression,
+                          mutableAggBufferOffset: Int = 0,
+                          inputAggBufferOffset: Int = 0)
+  extends GpuDeclarativeAggregate with GpuAggregateWindowFunction {
+
+  def this(child: Expression) = this(child, 0, 0)
+
+  override lazy val deterministic: Boolean = false
```
revans2, line 551: nit: Could we add a comment similar to what is in the Spark code for this.

```
  // Both `CollectList` and `CollectSet` are non-deterministic since their results depend on the
  // actual order of input rows.
```

revans2 (commented):
Looking good, I ma really just down to nits at this point.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
mythrocks: This should be supported. If it's not working, it might be a bug in `libcudf`. I'd be happy to investigate.

firestarman:
cudf side is ready now

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
firestarman: Thanks for review. Added it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
firestarman: Got it, updated.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
firestarman: Updated it, and i think it may support but I have not verified that, so do not cover it here. I am also confirming this with Mithun, waiting for his feedback.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2264,6 +2265,21 @@ object GpuOverrides {
         override def convertToGpu(child: Expression): GpuExpression =
           GpuMakeDecimal(child, a.precision, a.scale, a.nullOnOverflow)
       }),
+    expr[CollectList](
+      "Collect a list of elements, now only supported by windowing.",
+      /* It should be 'fullAgg' eventually but now only support windowing,
```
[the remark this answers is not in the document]
firestarman: done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1684,11 +1683,13 @@ object GpuOverrides {
...
       "Aggregate expression",
       ExprChecks.fullAgg(
-        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL,
+        (TypeSig.commonCudfTypes + TypeSig.DECIMAL).nested() + TypeSig.NULL +
+          TypeSig.ARRAY.nested(TypeSig.STRUCT),
         TypeSig.all,
         Seq(ParamCheck(
           "aggFunc",
-          TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL,
+          (TypeSig.commonCudfTypes + TypeSig.DECIMAL).nested() + TypeSig.NULL +
```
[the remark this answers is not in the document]
firestarman: Got it, updated.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1684,11 +1683,13 @@ object GpuOverrides {
     expr[AggregateExpression](
       "Aggregate expression",
       ExprChecks.fullAgg(
-        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL,
+        (TypeSig.commonCudfTypes + TypeSig.DECIMAL).nested() + TypeSig.NULL +
+          TypeSig.ARRAY.nested(TypeSig.STRUCT),
```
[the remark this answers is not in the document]
firestarman: Got it, updated.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -774,12 +774,11 @@ object GpuOverrides {
...
         "\"window\") of rows",
       ExprChecks.windowOnly(
-        TypeSig.commonCudfTypes + TypeSig.DECIMAL +
-          TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL),
+        (TypeSig.commonCudfTypes + TypeSig.DECIMAL).nested + TypeSig.ARRAY.nested(TypeSig.STRUCT),
         TypeSig.all,
         Seq(ParamCheck("windowFunction",
-          TypeSig.commonCudfTypes + TypeSig.DECIMAL +
-            TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL),
+          (TypeSig.commonCudfTypes + TypeSig.DECIMAL).nested +
```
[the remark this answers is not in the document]
firestarman: Got it, updated.

integration_tests/src/main/python/window_function_test.py
```diff
@@ -223,3 +223,66 @@ def test_window_aggs_for_ranges_of_dates(data_gen):
...
+      collect_list(c_decimal) over
+        (partition by a order by b,c_int rows between UNBOUNDED preceding and CURRENT ROW) as collect_decimal,
+      collect_list(c_struct) over
+        (partition by a order by b,c_int rows between UNBOUNDED preceding and CURRENT ROW) as collect_struct
+    from window_collect_table
+  '''
+
+# SortExec does not support array type, so sort the result locally.
+@ignore_order(local=True)
+@pytest.mark.skip("https://github.com/NVIDIA/spark-rapids/issues/1638")
```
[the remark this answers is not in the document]
firestarman: done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -774,12 +774,11 @@ object GpuOverrides {
       "Calculates a return value for every input row of a table based on a group (or " +
         "\"window\") of rows",
       ExprChecks.windowOnly(
-        TypeSig.commonCudfTypes + TypeSig.DECIMAL +
-          TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL),
+        (TypeSig.commonCudfTypes + TypeSig.DECIMAL).nested + TypeSig.ARRAY.nested(TypeSig.STRUCT),
```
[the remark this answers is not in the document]
firestarman: Updated it to use the more verbose format.

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
firestarman: if yes, i will add it in a following PR

revans2: Sounds good to me

revans2 (approved):
@mythrocks are you OK with a follow on PR to add in struct or struct support?

mythrocks:
> @mythrocks are you OK with a follow on PR to add in struct or struct support?

Sounds good.

PR #1551, opened 2021-01-19
RAPIDS-accelerated Hive UDFs support all types

jlowe:
Currently the type-checks for RAPIDS-accelerated Hive UDFs state that they cannot support some nested types such as array-of-structs, but this isn't the case.  From the plugin perspective, the input and output types are not examined but passed as-is into and out of the user-provided UDF.  It is up to the user's code to support the type or not, so the plugin should not preclude support of these types as they can be represented in a cudf `ColumnVector`.


jlowe:
build

sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/GpuHiveOverrides.scala
```diff
@@ -48,12 +48,9 @@ object GpuHiveOverrides {
       GpuOverrides.expr[HiveSimpleUDF](
         "Hive UDF, support requires the UDF to implement a RAPIDS-accelerated interface",
         ExprChecks.projectNotLambda(
-          TypeSig.commonCudfTypes + TypeSig.ARRAY.nested(TypeSig.commonCudfTypes),
           TypeSig.all,
```
revans2, line 51: We have never tested with any user defined types and I don't know what they might look like. At a minimum I think we want to exclude those until we have some kind of tests for them.

jlowe: I updated the types to include all but UDT.

jlowe:
build

pxLi:
build

firestarman:
LGTM

revans2 (approved):

Issue #1517, opened 2021-01-14
[FEA] Add tests for Datetimes with bad combination of year/month/day...


mythrocks:
Related to audit of https://github.com/apache/spark/commit/1034815519 and https://github.com/apache/spark/commit/b9f2f78de5.

Failure cases for year/month/day combinations have been modified in ANSI mode. We need to add tests to verify `ANSI_ENABLED` behaviour. We should also check that the tests added in this commit work correctly with the `spark-rapids` plugin.
Must also verify that datetime parsing fails with bad pattern strings.

sameerz:
@andygrove is there anything in https://github.com/apache/spark/commit/b9f2f78de5 that is more stringent but that we have on by default?

andygrove:
I will take a look at this tomorrow. It is also related to https://github.com/NVIDIA/spark-rapids/issues/1117

andygrove:
> @andygrove is there anything in [apache/spark@b9f2f78](https://github.com/apache/spark/commit/b9f2f78de5) that is more stringent but that we have on by default?

@sameerz Yes, the plugin currently returns `null` for the invalid dates instead of throwing an exception when ANSI mode is enabled

andygrove:
I opened a PR to fix this for CAST and filed https://github.com/NVIDIA/spark-rapids/issues/1556 to do the same for the string to timestamp functions since this is a different code path.

https://github.com/NVIDIA/spark-rapids/issues/1556

sameerz:
Closing this and tracking the follow up in #1556

PR #1555, opened 2021-01-20
Enable ANSI mode for CAST string to timestamp

andygrove:
This PR adds a check to the CAST from string to timestamp to detect if any non-null values were not converted successfully. This check only happens in ANSI mode and will throw a `DateTimeException` if there were values that could not be converted. This matches Spark's behavior.

Partially addresses https://github.com/NVIDIA/spark-rapids/issues/1517 - we have different code paths for `CAST` and other methods of converting string to timestamp (such as `to_date`, `to_timestamp`, `unix_timestamp`, and `to_unix_timestamp`). I filed https://github.com/NVIDIA/spark-rapids/issues/1556 for adding tests for these functions.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
@@ -791,10 +791,7 @@ case class GpuCast(
       // When ANSI mode is enabled, we need to throw an exception if any values could not be
       // converted
       if (ansiMode) {
-        val wasNotNull = withResource(input.isNull()) { wasNull =>
-          wasNull.not()
-        }
-        withResource(wasNotNull) { wasNotNull =>
+        withResource(input.isNotNull) { wasNotNull =>
           withResource(finalResult.isNull) { isNull =>
             withResource(wasNotNull.and(isNull)) { notConverted =>
               if (notConverted.any().getBoolean) {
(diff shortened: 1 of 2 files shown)


andygrove:
build

revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -784,8 +785,28 @@ case class GpuCast(
...
+      // converted
+      if (ansiMode) {
+        val wasNotNull = withResource(input.isNull()) { wasNull =>
+          wasNull.not()
+        }
+        withResource(wasNotNull) { wasNotNull =>
+          withResource(finalResult.isNull) { isNull =>
+            withResource(wasNotNull.and(isNull)) { notConverted =>
+              if (notConverted.any().getBoolean) {
+                throw new DateTimeException(
```
revans2, line 801: Sorry I approved this too early.  If something fails in `ansiMode`  finalResult is leaked, and because we throw in ansiMode on purpose at a minimum I would like to see this fixed with a closeOnExcept or something like that.

revans2 (changes requested):

revans2 (approved):

revans2:
build

gerashegalov (approved):
LGTM

PR #1560, opened 2021-01-21
support sum in window aggregation for decimal

sperlingxx:
This pull request is to enable sum as a window aggregation operator for decimal columns, which is one of tasks listed in #1333.

NOTE: Due to hard-code precision promotion in the result type of sum expression, we can only support sum with decimal data whose precision not greater than 8.


sperlingxx:
build

revans2 (approved):

PR #976, opened 2020-10-19
[WIP] decimal meta adjustment for GPU runtime

sperlingxx:
Catalyst analyzer will apply `promotePrecision` and `checkOverflow` on binary arithmetic expressions if data in both sides of expressions are decimal types (refer to [link](https://github.com/apache/spark/blob/branch-3.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/DecimalPrecision.scala#L94)).

Considering GPU_MAX_PRECISION of decimal type can only be up to 19 (half of MAX_PRECISION in spark),  we need to reapply precision and scale adjustment with GPU constraints. To achieve this, we introduce two new expression-rules `PromotePrecisionExprMeta` and `CheckOverflowExprMeta` into GpuOverrides.  

In order to keep align with decimal overflow controlling behavior of spark, we follow the same strategy as catalyst with GPU constraints replacement. But in spark catalyst runtime, overflow decimal value will be replaced by null. (Otherwise, overflow exception will be thrown.)  In cuDF runtime, overflow check for decimal operations is still missing.

PR #1561, opened 2021-01-21
Decimal support for add and subtract

rwlee:
Combines work from @sperlingxx #976 and @revans2 changes to resolve precision mismatches when running on the GPU. Merging this should close #976, as the necessary bits have already been merged or were copied over and modified for this PR.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/decimalExpressions.scala
```diff
@@ -16,10 +16,95 @@
...
+    parent: Option[RapidsMeta[_, _, _]],
+    rule: DataFromReplacementRule)
+  extends UnaryExprMeta[PromotePrecision](expr, conf, parent, rule) {
+  override def convertToGpu(child: Expression): GpuExpression =
+    GpuPromotePrecision(child)
+}
+
+object DecimalExpressions {
+  // Underlying storage type of decimal data in cuDF is int64_t, whose max capacity is 19.
+  val GPU_MAX_PRECISION: Int = 19
```
revans2, line 68: Despite this code never actually being used. CUDF thinks this value should be 18

https://github.com/rapidsai/cudf/blob/639049808913c7d5a7170141d784da6cc45da36d/java/src/main/java/ai/rapids/cudf/DType.java#L25

So which is right? and why does this not just pull from that value?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/decimalExpressions.scala
```diff
@@ -16,10 +16,95 @@
...
+    expr: PromotePrecision,
+    conf: RapidsConf,
+    parent: Option[RapidsMeta[_, _, _]],
+    rule: DataFromReplacementRule)
+  extends UnaryExprMeta[PromotePrecision](expr, conf, parent, rule) {
+  override def convertToGpu(child: Expression): GpuExpression =
+    GpuPromotePrecision(child)
+}
+
+object DecimalExpressions {
```
revans2, line 66: This entire object is never used.  Why do we need it?

tests/src/test/scala/com/nvidia/spark/rapids/unit/DecimalUnitTest.scala
```diff
@@ -21,8 +21,9 @@ import java.math.RoundingMode
 import scala.util.Random
 
 import ai.rapids.cudf.{ColumnVector, DType, HostColumnVector}
-import com.nvidia.spark.rapids.{GpuAlias, GpuBatchScanExec, GpuColumnVector, GpuIsNotNull, GpuIsNull, GpuLiteral, GpuOverrides, GpuScalar, GpuUnitTests, HostColumnarToGpu, RapidsConf}
+import com.nvidia.spark.rapids._
 
+import org.scalatest.Matchers.convertToAnyShouldWrapper
 import org.apache.spark.SparkConf
```
revans2, line 27: Could we totally revert all of the changes to this file, as it looks like they are not needed.

revans2 (commented):
Because I was involved with writing some of the parts of the PR I think it is best if others approve/review it too.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/decimalExpressions.scala
rwlee: I don't think it's needed anymore, all references were stripped out with the simplification of the `PromotePrecision` operation.

revans2 (approved):

revans2:
build

jlowe (approved):

PR #1562, opened 2021-01-21
add alluxio support

wbo4958:
add spark.rapids.alluxio.pathsToReplace to specify how to replace file paths
to alluxio schema, then leverage the power of alluxio's data archestration to
accelerate file reading.


shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -158,8 +160,15 @@ class Spark300Shims extends SparkShims {
           override def convertToGpu(): GpuExec = {
             val sparkSession = wrapped.relation.sparkSession
             val options = wrapped.relation.options
+
+            val location = getLocationWithAlluxioReplace(
+              conf,
+              wrapped.relation,
+              wrapped.partitionFilters,
+              wrapped.dataFilters)
```
tgravescs, line 168: what happens if the user runs a query with input_file_name()? does it show original or alluxio:// paths? I think it uses alluxio one but want to verify and we need to decide if that is what we want. its probably something we just document.

Something like:

      ```
   lambda spark : spark.read.parquet(data_path)\
                    .filter(f.col('a') > 0)\
                    .selectExpr('a',
                        'input_file_name()',
                        'input_file_block_start()',
                        'input_file_block_length()'),
```

File: shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala [hunk 2]
```diff
@@ -436,4 +445,68 @@ class Spark300Shims extends SparkShims {
...
+        val ruleSplit = rule.split("->")
+        if (ruleSplit.size == 2) {
+          replaceMap += ruleSplit(0).trim -> ruleSplit(1).trim
+        } else {
+          logWarning("Wrong set for spark.rapids.alluxio.pathsToReplace " + rule)
+        }
+      })
+
+      if (replaceMap.size > 0) {
+        val listFiles: Seq[PartitionDirectory] = relation.location.listFiles(
```
tgravescs, line 470: you don't need the type specified here, can we rename to partitionDirs

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -436,4 +445,68 @@ class Spark300Shims extends SparkShims {
...
+          partitionFilters, dataFilters)
```
tgravescs, line 471: this might need to take into account isDynamicPruningFilter -> see how its done in GpuFileSourceScanExec.

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -436,4 +445,68 @@ class Spark300Shims extends SparkShims {
...
+            new Path(
+              f.toString.replaceFirst(matchedSet.head, replaceMap(matchedSet.head)))
+          } else {
+            f
+          }
+        }
+
+        val inputFiles: Seq[Path] = listFiles.flatMap(partitionDir => {
+          partitionDir.files.map(f => replaceFunc(f.getPath))
+        }).toSet.toSeq
```
tgravescs, line 485: why do we do toSet.ToSeq?  it should be a Seq already.

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -436,4 +445,68 @@ class Spark300Shims extends SparkShims {
...
+
+        // get the leaf dir of inputFiles
+        val leafDirs = inputFiles.map(_.getParent).toSet.toSeq
```
tgravescs, line 488: similar here, why toSet.ToSeq?  Also looking at the inferPartitioning in spark its filtering out non data files:

```
  // We use leaf dirs containing data files to discover the schema.
      val leafDirs = leafDirToChildrenFiles.filter { case (_, files) =>
        files.exists(f => isDataPath(f.getPath))
      }.keys.toSeq
```

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -436,4 +445,68 @@ class Spark300Shims extends SparkShims {
...
+
+        val finalRootPaths = relation.location.rootPaths.map(replaceFunc)
+
+        val partitionSpec = RapidsPartitioningUtils.inferPartitioning(
```
tgravescs, line 492: have we tested this on non-partitioned files?  ie no partitioning in the path

sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/RapidsPartitioningUtils.scala
```diff
@@ -0,0 +1,56 @@
...
+   */
+  def inferPartitioning(
+                         sparkSession: SparkSession,
+                         leafDirs: Seq[Path],
+                         basePaths: Set[Path],
+                         parameters: Map[String, String],
+                         userSpecifiedSchema: Option[StructType]): PartitionSpec = {
+
+    val caseInsensitiveOptions = CaseInsensitiveMap(parameters)
+    val timeZoneId = caseInsensitiveOptions.get(DateTimeUtils.TIMEZONE_OPTION)
```
tgravescs, line 44: in the spark version it also checks recursiveFileLookup, need to understand that.

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -436,4 +445,68 @@ class Spark300Shims extends SparkShims {
...
+          relation.sparkSession,
+          leafDirs,
+          finalRootPaths.toSet,
+          relation.options,
+          Option(relation.dataSchema))
+
+        new InMemoryFileIndex(
```
tgravescs, line 499: so I believe this will submit another job to spark to do the listing when its over 32 files, can you confirm? How long does that job take?  Not sure there is anything we can do or want to do about it, but just wondering

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -436,4 +445,68 @@ class Spark300Shims extends SparkShims {
...
+          logWarning("Wrong set for spark.rapids.alluxio.pathsToReplace " + rule)
+        }
+      })
+
+      if (replaceMap.size > 0) {
+        val listFiles: Seq[PartitionDirectory] = relation.location.listFiles(
+          partitionFilters, dataFilters)
+
+        val replaceFunc = (f: Path) => {
+          val matchedSet = replaceMap.keySet.filter(reg => f.toString.startsWith(reg))
```
tgravescs, line 474: this unfortunately seems very expensive, how long does this take with thousands of files?
only do f.toString once in local var.

tgravescs (commented):
what all testing have we done with this?  We definitely need some tests for this. integration tests might be harder to actually setup alluxio so at least some unit tests with different conditions and manual testing. Let make sure it works with different scheme (s3, hdfs, local, etc), partitioned and not partitioned, ideally with different FileIndex's coming in.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
```diff
@@ -21,14 +21,14 @@ import org.apache.spark.sql.{SparkSession, SparkSessionExtensions}
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.analysis.Resolver
 import org.apache.spark.sql.catalyst.encoders.ExpressionEncoder
-import org.apache.spark.sql.catalyst.expressions.{Alias, Expression, ExprId, NullOrdering, SortDirection, SortOrder}
+import org.apache.spark.sql.catalyst.expressions.{Alias, ExprId, Expression, NullOrdering, SortDirection, SortOrder}
```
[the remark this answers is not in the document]
wbo4958: Done

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -17,11 +17,11 @@
 package com.nvidia.spark.rapids.shims.spark300
 
 import java.time.ZoneId
-
```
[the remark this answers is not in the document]
wbo4958: Done

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -158,8 +160,15 @@ class Spark300Shims extends SparkShims {
           override def convertToGpu(): GpuExec = {
             val sparkSession = wrapped.relation.sparkSession
             val options = wrapped.relation.options
+
```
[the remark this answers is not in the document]
wbo4958: I have not got time to investigate v2. let's add a task to track this

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -436,4 +445,68 @@ class Spark300Shims extends SparkShims {
...
+        }
+      })
+
+      if (replaceMap.size > 0) {
+        val listFiles: Seq[PartitionDirectory] = relation.location.listFiles(
+          partitionFilters, dataFilters)
+
+        val replaceFunc = (f: Path) => {
+          val matchedSet = replaceMap.keySet.filter(reg => f.toString.startsWith(reg))
+          if (matchedSet.size > 0) {
```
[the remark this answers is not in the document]
wbo4958: Done

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
wbo4958: it shows alluxio:// paths, I just verified

revans2:
I am a little confused why we need this at all.  Can't the user just replace the paths on their own after they set up Alluxio? Are there plans to auto-deploy Alluxio with the plugin and make it transparent to the end user? Wouldn't letting the user change things on their own that make it better for them because then they can test that it is doing the right thing everywhere, even when not using the accelerator and adjust to how the files paths may have changed? Also is it expected that we only replace paths automatically for read nodes moved to the GPU?

tgravescs:
yes the user can change their paths, this is to allow them to not have to change their paths and it happen transparently.  Yes it only works when using the GPU readers

tgravescs:
@wbo4958  you need to signoff on a commit -> see https://github.com/NVIDIA/spark-rapids/blob/branch-0.4/CONTRIBUTING.md#sign-your-work

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
tgravescs: we need to document this.  can you please file an issue to add documentation for the entire feature and include this

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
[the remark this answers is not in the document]
tgravescs: not seeing a change here?

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
tgravescs: why do we have replications of same file? was this copied from somewhere in Spark?  what if user specified a directory twice because they wanted to read it twice?

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
tgravescs: answered above, but you didn't answer the isDataPath question?

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
tgravescs: so can you confirm it did run a separate job and that took < 1s?

sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/rapids/GpuPartitioningUtils.scala
```diff
@@ -0,0 +1,67 @@
...
+package org.apache.spark.sql.execution.datasources.rapids
+
+import org.apache.hadoop.fs.Path
+
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.util.{CaseInsensitiveMap, DateTimeUtils}
+import org.apache.spark.sql.execution.datasources.{PartitioningUtils, PartitionSpec}
+import org.apache.spark.sql.types.StructType
+
+
```
tgravescs, line 26: remove extra newline

sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/rapids/GpuPartitioningUtils.scala
```diff
@@ -0,0 +1,67 @@
...
+   * Mainly derived from PartitioningAwareFileIndex.inferPartitioning
+   * @param sparkSession
+   * @param leafDirs leaf directory paths
+   * @param basePaths Contains a set of paths that are considered as the base dirs of the
+   *                  input datasets. The partitioning discovery logic will make sure it
+   *                  will stop when it reaches any base path
+   * @param parameters a set of options to control partition discovery
+   * @param userSpecifiedSchema an optional user specified schema that will be use to provide
+   *                            types for the discovered partitions
+   * @return
```
tgravescs, line 39: fill in return

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
```diff
@@ -163,4 +166,40 @@ trait SparkShims {
...
+   * `something`.
+   * Case 3) `spark.read.parquet("/path/")`: the base path will be `/path/`, and the returned
+   * DataFrame will have the column of `something`.
+   *
+   * Users also can override the basePath by setting `basePath` in the options to pass the new base
+   * path to the data source.
+   * For example, `spark.read.option("basePath", "/path/").parquet("/path/something=true/")`,
+   * and the returned DataFrame will have the column of `something`.
+   */
+  def getBasePaths(
```
tgravescs, line 196: why does this need to be in shim, I only see it called from spark300shim?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
```diff
@@ -163,4 +166,40 @@ trait SparkShims {
...
+    hadoopConf: Configuration,
+    basePathOption: Option[Path],
+    rootPaths: Seq[Path],
+    leafFiles: Seq[Path]): Set[Path]
+
+  def replacePartitionDirectoryFiles(
```
tgravescs, line 202: I'm confused by this, I get there is a databricks one since ParittionDirectory but I don't see anything calling that version. the only place replacePartitionDirectoryFiles is called is from Spark300Shims. replaceWithAlluxioPathIfNeeded which is going to call into its local version of replacePartitionDirectoryFiles not the databricks version

sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/RapidsPartitioningUtils.scala
wbo4958: added the logic

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
wbo4958: Done

sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/rapids/GpuPartitioningUtils.scala
wbo4958: Done

sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/rapids/GpuPartitioningUtils.scala
wbo4958: Done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
wbo4958: Spark301dbShims -> replaceWithAlluxioPathIfNeeded(in Spark300Shims) -> replacePartitionDirectoryFiles (overrided in Spark301dbShims)

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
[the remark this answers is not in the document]
wbo4958: Done

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
wbo4958: Done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
wbo4958: Done

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
wbo4958: filed [an issue ](https://github.com/NVIDIA/spark-rapids/issues/1574) to track the document things

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
wbo4958: Yes, confirmed, I added some log to metric the function replaceWithAlluxioPathIfNeeded for the query based on 4152 files and the first run took longer (1.876, 6.26, 1.222, 1.171)s, but the 2nd run only took about 1s (0.049, 1.162, 0.845, 0.748)s.

tgravescs:
build

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
tgravescs: I'm assuming those numbers are the time of the extra job. I would expect the string replacement to take the same amount of time each run.  But I'm guessing the difference is in the actual listing.

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -440,4 +451,156 @@ class Spark300Shims extends SparkShims {
...
+      //   gs:/bar -> alluxio://0.1.2.3:19998/bar,
+      //   /baz -> alluxio://0.1.2.3:19998/baz }
+      val replaceMapOption = alluxioPathsReplace.map(rules => {
+        rules.map(rule => {
+          val split = rule.split("->")
+          if (split.size == 2) {
+            split(0).trim -> split(1).trim
+          } else {
+            throw new IllegalArgumentException(
+              "Wrong setting for spark.rapids.alluxio.pathsToReplace: " + rule)
```
tgravescs, line 476: Perhaps say Invalid setting for ${RapidsConf.ALLUXIO_PATHS_REPLACE.key}

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -440,4 +451,156 @@ class Spark300Shims extends SparkShims {
...
+          partitionFilters.filterNot(isDynamicPruningFilter), dataFilters)
+
+        // replacement func to check if the file path is prefixed with the string user configured
+        // if yes, replace it
+        val replaceFunc = (f: Path) => {
+          val pathStr = f.toString
+          val matchedSet = replaceMap.keySet.filter(reg => pathStr.startsWith(reg))
+          if (matchedSet.size > 1) {
+            throw new IllegalArgumentException(
+              "Found " + matchedSet.size + " same rules of spark.rapids.alluxio.pathsToReplace.")
```
tgravescs, line 496: use ${RapidsConf.ALLUXIO_PATHS_REPLACE.key} and add more text saying must only match 1

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -440,4 +451,156 @@ class Spark300Shims extends SparkShims {
...
+   * `something`.
+   * Case 3) `spark.read.parquet("/path/")`: the base path will be `/path/`, and the returned
+   * DataFrame will have the column of `something`.
+   *
+   * Users also can override the basePath by setting `basePath` in the options to pass the new base
+   * path to the data source.
+   * For example, `spark.read.option("basePath", "/path/").parquet("/path/something=true/")`,
+   * and the returned DataFrame will have the column of `something`.
+   */
+  // mainly copied from PartitioningAwareFileIndex.basePaths
```
tgravescs, line 571: add this into the description above as the last line rather then having separate // comment

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -440,4 +451,156 @@ class Spark300Shims extends SparkShims {
...
+        val inputFiles: Seq[Path] = partitionDirs.flatMap(partitionDir => {
+          replacePartitionDirectoryFiles(partitionDir, replaceFunc)
+        })
+
+        def isDataPath(path: Path): Boolean = {
+          val name = path.getName
+          !((name.startsWith("_") && !name.contains("=")) || name.startsWith("."))
+        }
+        // filter out non-data path and get unique leaf dirs of inputFiles
+        val leafDirs = inputFiles.filter(isDataPath).map(_.getParent).toSet.toSeq
```
tgravescs, line 514: it would be nice to put the isDataPath logic in inferPartitioning to keep the code close to the Spark code as it wuld be easier to find changes.

tgravescs:
Please don't rebase or force push when in the middle of a review unless absolutely necessary. Its much harder for the reviewed see what you changed between revisions

sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/rapids/GpuPartitioningUtils.scala
```diff
@@ -53,6 +57,18 @@ object GpuPartitioningUtils {
       val timeZoneId = caseInsensitiveOptions.get(DateTimeUtils.TIMEZONE_OPTION)
         .getOrElse(sparkSession.sessionState.conf.sessionLocalTimeZone)
 
+      // filter out non-data path and get unique leaf dirs of inputFiles
+      val leafDirs: Seq[Path] = leafFiles.filter(isDataPath).map(_.getParent).toSet.toSeq
+
+
```
tgravescs, line 63: remove extra newline

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -491,6 +491,7 @@ class Spark300Shims extends SparkShims {
           val pathStr = f.toString
           val matchedSet = replaceMap.keySet.filter(reg => pathStr.startsWith(reg))
           if (matchedSet.size > 1) {
+            // never reach here since replaceMap is a Map
```
tgravescs, line 494: what does this mean?  map can have size 0, do you mean by this point there will be something in it?

wbo4958: I mean the condition of (matchedSet.size>1) will not happen, since there is no replication in replaceMap.keySet.

wbo4958: eg. s3:/root->alluxio:/root,s3:/root->alluxio:/bar is allowed, and the last one (s3:/root) will override the previous one

tgravescs: ok

tgravescs:
build

tgravescs:
YOu need to fix the scala style violation:

18:29:06  error file=/home/jenkins/agent/workspace/jenkins-rapids_premerge-github-830/shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala message=There should be no empty line separating imports in the same group. line=24 column=0

wbo4958:
> YOu need to fix the scala style violation:
> 
> 18:29:06 error file=/home/jenkins/agent/workspace/jenkins-rapids_premerge-github-830/shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala message=There should be no empty line separating imports in the same group. line=24 column=0

Thx for this info, I will fix it.

pxLi:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -777,6 +777,19 @@ object RapidsConf {
...
 
+  // ALLUXIO CONFIGS
+
+  val ALLUXIO_PATHS_REPLACE = conf("spark.rapids.alluxio.pathsToReplace")
+    .doc("List of paths to be replaced with corresponding alluxio scheme. Eg, when configure" +
+      "is set to \"s3:/foo->alluxio://0.1.2.3:19998/foo,gcs:/bar->alluxio://0.1.2.3:19998/bar\", " +
+      "which means:  " +
+      "     s3:/foo/a.csv will be replaced to alluxio://0.1.2.3:19998/foo/a.csv and " +
+      "     gcs:/bar/b.csv will be replaced to alluxio://0.1.2.3:19998/bar/b.csv")
+    .internal()
```
tgravescs, line 788: we expect users to set it so this shouldn't be internal.

tgravescs (commented):
sorry I missed a minor thing with the config otherwise looks good

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
wbo4958: Thx Tom

tgravescs (approved):

Issue #1372, opened 2020-12-11
[FEA] Add support for array_contains


kuhushukla:
**Is your feature request related to a problem? Please describe.**
Once the following cudf issues are done, we can add the plugin side changes for array_contains operator. 
   - [x] https://github.com/rapidsai/cudf/issues/6983
   - [x] https://github.com/rapidsai/cudf/issues/6944

sameerz:
@kuhushukla assigning you on this issue for the plugin side work.

PR #1565, opened 2021-01-21
Add GPU accelerated array_contains operator

kuhushukla:
Closes https://github.com/NVIDIA/spark-rapids/issues/1372

Depends on https://github.com/rapidsai/cudf/pull/7125 and https://github.com/rapidsai/cudf/pull/7039
This Draft/WIP PR adds support for `array_contains` on the gpu. I have skipped some input check methods from the Spark implementation since right now we only support basic typed lists in LHS, it would be nice to get reviews on that decision, refer to (https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/collectionOperations.scala#L1120) . I hope I have got the Expr checks right given we want to support only basic typed lists. Appreciate any corrections if needed. This patch also needed the `hasNans` check since we have incompat behavior with Spark. Marking this as draft since we have moving pieces.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1918,6 +1918,27 @@ object GpuOverrides {
         override def convertToGpu(): GpuExpression =
           GpuCreateNamedStruct(childExprs.map(_.convertToGpu()))
       }),
+    expr[ArrayContains](
+      "Returns a boolean if the array contains the passed in key",
+      ExprChecks.binaryProjectNotLambda(
+        TypeSig.BOOLEAN,
+        TypeSig.all,
```
revans2, line 1925: Does spark really output anything except boolean?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1918,6 +1918,27 @@ object GpuOverrides {
...
+        ("array", TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.NULL),
+          TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.NULL)),
```
revans2, line 1927: If spark supports an array of anything for the key then it needs to also support an array of anything for "array"

integration_tests/src/main/python/array_test.py
```diff
@@ -80,4 +80,22 @@ def test_orderby_array_of_structs(data_gen):
...
-        'select array_table.a, array_table.a[0].child0 as first_val from array_table order by first_val')
\ No newline at end of file
+        'select array_table.a, array_table.a[0].child0 as first_val from array_table order by first_val')
+
+
+@pytest.mark.parametrize('data_gen', [ArrayGen(sub_gen) for sub_gen in [byte_gen, short_gen, int_gen, long_gen,
+                                                                        FloatGen(no_nans=True), DoubleGen(no_nans=True),
+                                                                        string_gen, boolean_gen, date_gen, timestamp_gen]], ids=idfn)
+def test_array_contains(data_gen):
+    assert_gpu_and_cpu_are_equal_sql(
```
revans2, line 90: We can generate scalar values. 

https://github.com/NVIDIA/spark-rapids/blob/d706e3ec92882adc5c05d94ba14b1fab9178d837/integration_tests/src/main/python/data_gen.py#L616-L637

Might need to switch from SQL to a dataframe though, but it should not be too bad.

```
@pytest.mark.parametrize('data_gen', byte_gen, short_gen, int_gen, long_gen,
                                                                        FloatGen(no_nans=True), DoubleGen(no_nans=True),
                                                                        string_gen, boolean_gen, date_gen, timestamp_gen], ids=idfn)
def test_array_contains(data_gen):
  arr_gen = ArrayGen(data_gen)
  lit = gen_scalar(data_gen)
  assert_gpu_and_cpu_are_equal(lambda spark: two_col_df(spark, arr_gen, data_gen).select(array_contains(col('a'), lit), array_contains(col('a'), col('b')), ...)
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1918,6 +1918,27 @@ object GpuOverrides {
         override def convertToGpu(): GpuExpression =
           GpuCreateNamedStruct(childExprs.map(_.convertToGpu()))
       }),
+    expr[ArrayContains](
+      "Returns a boolean if the array contains the passed in key",
+      ExprChecks.binaryProjectNotLambda(
+        TypeSig.BOOLEAN,
+        TypeSig.all,
+        ("array", TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.NULL),
```
revans2, line 1926: Why are we not testing for decimal support.  This is a new operator and there is no reason we should not start off with decimal support.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -184,4 +185,38 @@ case class GpuGetMapValue(child: Expression, key: Expression)
...
 
   override def right: Expression = key
+}
+
+case class GpuArrayContains(left: Expression, right: Expression)
+  extends GpuBinaryExpression with ImplicitCastInputTypes with NullIntolerant {
+
+  override def dataType: DataType = BooleanType
+
+  override def inputTypes: Seq[AbstractDataType] = {
```
revans2, line 195: I'm not sure that we need this or ImplicitCastInputTypes because that all happens in the logical plan, which this will never be a part of.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -184,4 +185,38 @@ case class GpuGetMapValue(child: Expression, key: Expression)
...
+
+  override def nullable: Boolean = {
+    left.nullable || right.nullable || left.dataType.asInstanceOf[ArrayType].containsNull
+  }
+
+  override def doColumnar(lhs: GpuColumnVector, rhs: Scalar): ColumnVector =
+    lhs.getBase.listContains(rhs)
+
+  override def doColumnar(numRows: Int, lhs: Scalar, rhs: Scalar): ColumnVector =
+    throw new IllegalStateException("This is not supported yet")
```
revans2, line 213: The we need to be explicit in the type checking that we do not support literal arrays.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
kuhushukla: I don't think decimal is supported for array contains on cudf side.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
kuhushukla: My understanding of the checks is fairly stunted and I am really happy to receive this feedback as it helps me understand. Will update.

integration_tests/src/main/python/array_test.py
```diff
@@ -80,4 +81,15 @@ def test_orderby_array_of_structs(data_gen):
...
+
+
+@pytest.mark.parametrize('data_gen', [byte_gen, short_gen, int_gen, long_gen,
+                                      FloatGen(no_nans=True), DoubleGen(no_nans=True),
+                                      string_gen, boolean_gen, date_gen, timestamp_gen], ids=idfn)
+def test_array_contains(data_gen):
+    arr_gen = ArrayGen(data_gen)
+    lit = gen_scalar(data_gen, force_no_nulls=True)
+    assert_gpu_and_cpu_are_equal_collect(lambda spark: debug_df(two_col_df(
+        spark, arr_gen, data_gen).select(array_contains(col('a'), lit.cast(data_gen.data_type)),
```
revans2, line 94: Because we are doing random number generation, it is likely that all of these will come back as false.  If we want to do something like.

```
array_contains(col('a'), col('a')[5])
```

too that would be good.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2476,7 +2500,7 @@ object GpuOverrides {
       "The backend for hash based aggregations",
       ExecChecks(
         (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.MAP)
-            .nested(TypeSig.STRING),
+            .nested(TypeSig.STRING) + TypeSig.ARRAY,
         TypeSig.all),
```
kuhushukla, line 2504: @revans2 this is an additional change per observations from a sample _user_ query I ran for array_contains. I want an array to be a valid input but not be nested in any way and that is what I am going for here. Let me know if you think I should do it separately or differently.

integration_tests/src/main/python/array_test.py
kuhushukla: Added another `array_contains` clause similar to the original change.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1992,6 +1992,30 @@ object GpuOverrides {
...
+          TypeSig.ARRAY.nested(TypeSig.all)),
+        ("key", TypeSig.commonCudfTypes, TypeSig.all)),
+      (in, conf, p, r) => new BinaryExprMeta[ArrayContains](in, conf, p, r) {
+        override def tagExprForGpu(): Unit = {
+          if (extractLit(in.left).isDefined) {
+            willNotWorkOnGpu("Literal arrays are not supported for array_contains")
+          }
+          if (conf.hasNans && (in.children.map(a => a.dataType).contains(DoubleType) ||
+            in.children.map(a => a.dataType).contains(FloatType))) {
+            willNotWorkOnGpu("Array Contains on floating point columns that can contain NaNs " +
```
revans2, line 2010: Just curious is this true even if what we are looking for is a literal that is not NaN?  That feels like a common case and we might be able to allow it to work.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
revans2: The way this is set up it will only work for an array of Strings. Is that what you wanted?

revans2 (commented):
Looking better just a few questions/nits.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
kuhushukla: Oh! Is that so? The 'ARRAY` is outside the nested call where we say we only support strings so I thought that would not be the case. 
For arrays of basic types but not decimals and other nested types , would this
```
      ExecChecks(
        (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.MAP)
            .nested(TypeSig.STRING) + TypeSig.ARRAY.nested(TypeSig.commonCudfTypes),
        TypeSig.all)
```
make more sense? I want to support arrays of basic types + arrays of strings aka `commonCudfTypes`

kuhushukla: I have made some changes here that might be more relevant based on the docs they seem to generate. Also, tested on the sample query to make sure that works. Appreciate any thoughts on this @revans2

revans2: My only concern now is that it looks like we are restricting the checks in a way that they do not support.

You only get top level types and nested types.  You cannot say that a Map of Strings or an Array of Doubles is OK, but nothing else.  You would end up with a Map of Strings, or Doubles OR and Array of Strings or Doubles.  If we run into a situation where we have to have something like that we can look at adding it in.  But until then I think it  would be best to just have it look more like.
```
(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.MAP + TypeSig.ARRAY)
  .nested(TypeSig.commonCudfTypes + TypeSig.NULL)
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
kuhushukla: Sorry for the delay, I am running some tests and fighting Spark a bit with the plans with literals but I do have a fix. Will post asap.

kuhushukla: As an update , if we look at : 
```
array_contains(col('a'), col('a')[5]))
```
in the pytests, does not exercise the scalar for rhs route, I am going to work on getting a scalar that actually exists in the lists.

kuhushukla:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
kuhushukla: Thanks Bobby. Makes sense, I was hoping to keep existing one as is and tack this on. Using the suggested approach.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
kuhushukla: I have added a test to check for a literal from the array that is not a nan to see if we get the right results. It needed a `collect()` and I appreciate any improvements if needed.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -24,7 +24,7 @@ import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.analysis.TypeCheckResult
```
jlowe, line 24: 2021 copyrights

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1992,6 +1992,49 @@ object GpuOverrides {
...
+        override def tagExprForGpu(): Unit = {
+          // do not support literal arrays as lhs
+          if (extractLit(in.left).isDefined) {
+            willNotWorkOnGpu("Literal arrays are not supported for array_contains")
+          }
+
+          val rhsVal = extractLit(in.right)
+          in.children.map(a => a.dataType).foreach {
+            case dataType@(DoubleType | FloatType) => if (rhsVal.isDefined) {
+              val isNanVal = dataType match {
```
jlowe, line 2013: Style nit:
```suggestion
            case dataType@(DoubleType | FloatType) =>
            if (rhsVal.isDefined) {
              val isNanVal = dataType match {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1992,6 +1992,49 @@ object GpuOverrides {
...
+        ("key", TypeSig.commonCudfTypes, TypeSig.all)),
+      (in, conf, p, r) => new BinaryExprMeta[ArrayContains](in, conf, p, r) {
+        override def tagExprForGpu(): Unit = {
+          // do not support literal arrays as lhs
+          if (extractLit(in.left).isDefined) {
+            willNotWorkOnGpu("Literal arrays are not supported for array_contains")
+          }
+
+          val rhsVal = extractLit(in.right)
+          in.children.map(a => a.dataType).foreach {
```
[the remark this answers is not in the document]
kuhushukla: Thanks for the review Bobby, I was trying to be _extra_ careful when looking at all children datatypes. I have incorporated the change and it looks much better.
As for 
> Also I think the current logic has some bugs in it, 
>
If you have an instance of the bugs you are alluding to, I can probably act on it and test the code some more for cases I might have missed. We already know casting arrays is not supported on the GPU and decimals won't work for some of the plans `array_contains` can generate and I have not added negative tests for those. Appreciate the review.

revans2 (approved):

integration_tests/src/main/python/array_test.py
```diff
@@ -80,4 +81,27 @@ def test_orderby_array_of_structs(data_gen):
...
+    arr_gen = ArrayGen(data_gen)
+    lit = gen_scalar(data_gen, force_no_nulls=True)
+    assert_gpu_and_cpu_are_equal_collect(lambda spark: two_col_df(
+        spark, arr_gen, data_gen).select(array_contains(col('a'), lit.cast(data_gen.data_type)),
+                                         array_contains(col('a'), col('b')),
+                                         array_contains(col('a'), col('a')[5])), no_nans_conf)
+
+
+@pytest.mark.parametrize('data_gen', [double_gen], ids=idfn)
+def test_array_contains_for_nans(data_gen):
```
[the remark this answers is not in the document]
kuhushukla: Done.

kuhushukla:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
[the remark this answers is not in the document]
revans2: No logic bug, I just got confused by it

jlowe (approved):

PR #1570, opened 2021-01-22
Branch 0.4 doc cleanup

sameerz:
Formatting changes: limiting line length, converting DOS to unix file format, and added a newline in `TypeChecks.scala` and rebuilt supported_ops.md.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -11,11 +11,11 @@ Apache Spark caches what is used to build the output of the `explain()` function
 knowledge about configs, so it may return results that are not up to date with the current config
 settings. This is true of all configs in Spark. If you changed
 `spark.sql.autoBroadcastJoinThreshold` after running `explain()` on a `DataFrame`, the resulting
-query would not change to reflect that config and still show a `SortMergeJoin` even though the
-new config might have changed to be a `BroadcastHashJoin` instead. When actually running something
-like with `collect`, `show` or `write` a new `DataFrame` is constructed causing spark to re-plan the
-query. This is why `spark.rapids.sql.enabled` is still respected when running, even if explain
-shows stale results.
+query would not change to reflect that config and still show a `SortMergeJoin` even though the new
+config might have changed to be a `BroadcastHashJoin` instead. When actually running something like
+with `collect`, `show` or `write` a new `DataFrame` is constructed causing spark to re-plan the
+query. This is why `spark.rapids.sql.enabled` is still respected when running, even if explain shows
+stale results.
 
 ### What versions of Apache Spark does the RAPIDS Accelerator for Apache Spark support?
 
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -13,7 +13,7 @@ settings. This is true of all configs in Spark. If you changed
 `spark.sql.autoBroadcastJoinThreshold` after running `explain()` on a `DataFrame`, the resulting
 query would not change to reflect that config and still show a `SortMergeJoin` even though the new
 config might have changed to be a `BroadcastHashJoin` instead. When actually running something like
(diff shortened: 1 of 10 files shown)


sameerz:
build

docs/FAQ.md
```diff
@@ -11,11 +11,11 @@ Apache Spark caches what is used to build the output of the `explain()` function
...
 settings. This is true of all configs in Spark. If you changed
 `spark.sql.autoBroadcastJoinThreshold` after running `explain()` on a `DataFrame`, the resulting
-query would not change to reflect that config and still show a `SortMergeJoin` even though the
-new config might have changed to be a `BroadcastHashJoin` instead. When actually running something
-like with `collect`, `show` or `write` a new `DataFrame` is constructed causing spark to re-plan the
-query. This is why `spark.rapids.sql.enabled` is still respected when running, even if explain
-shows stale results.
+query would not change to reflect that config and still show a `SortMergeJoin` even though the new
+config might have changed to be a `BroadcastHashJoin` instead. When actually running something like
+with `collect`, `show` or `write` a new `DataFrame` is constructed causing spark to re-plan the
```
jlowe, line 16:
```suggestion
with `collect`, `show` or `write` a new `DataFrame` is constructed causing Spark to re-plan the
```

docs/compatibility.md
```diff
@@ -436,5 +448,5 @@ When translating UDFs to Catalyst expressions, the supported UDF functions are l
 |                          | x.toArray                                                |
 |                          | lhs += rhs                                               |
 |                          | lhs :+ rhs                                               |
-| Method call              | Only if the method being called <ol><li>consists of operations supported by the UDF compiler, and</li><li>is one of the folllowing:<ul><li>a final method, or</li><li>a method in a final class, or</li><li>a method in a final object</li></ul></li></ol> |
```
jlowe, line 439: The intent of the original code was to produce a bulleted list which makes it much easier to read.  The new version no longer does that.  Was the old formatting causing problems in something other than Markdown?

jlowe (changes requested):
There's a lot of unnecessary change in this PR.,  Many lines with reasonable line lengths and normal line endings were reformatted in trivial ways that have no impact on the rendered output.  Not something that needs to change for this PR, but it would be good to avoid for future doc updates.

If there's a formatting standard/tool everyone should be using then the contributing/dev docs should be updated.

docs/compatibility.md
sameerz: The formatting does not work in github pages markdown: https://nvidia.github.io/spark-rapids/docs/compatibility.html

docs/FAQ.md
sameerz: The intent with the line length changes was to make line lengths 100 (similar to the scala style recommendation https://github.com/databricks/scala-style-guide#line-length ).

docs/compatibility.md
sameerz: Screen Shot 2021-01-22 at 9 08 53 AM

sameerz:
build

docs/FAQ.md
jlowe: Understood, but paragraphs that were not over the 100 character limit were also reformatted unnecessarily.  See the changes to the first paragraph in compatibility.md as one example.  If there's a reformatting tool you're using for this and everyone is supposed to use it, we should document it (and audit/automate it as much as possible) otherwise we're destined for another "let's rewrite the docs for formatting" PR after a while.

docs/compatibility.md
jlowe: Sigh.  Github pages/Jekyll is why we can't have nice things.

jlowe (approved):

Issue #1576, opened 2021-01-25
[BUG] loading SPARK-32639 example parquet file triggers a JVM crash 


gerashegalov:
**Describe the bug**
Replaying the scenario from #1463 [(SPARK-32639)](https://issues.apache.org/jira/browse/SPARK-32639) in GPU-enabled Scala Spark or pyspark shell results in SIGSEGV in the executor call path 

```
C  [cudf_io5390148209696334138.so+0x170830]  cudf::io::detail::parquet::reader::impl::decode_page_data(hostdevice_vector<cudf::io::parquet::gpu::ColumnChunkDesc>&, hostdevice_vector<cudf::io::parquet::gpu::PageInfo>&, hostdevice_vec
tor<cudf::io::parquet::gpu::PageNestingInfo>&, unsigned long, unsigned long, rmm::cuda_stream_view)+0x4b0

Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
j  ai.rapids.cudf.Table.readParquet([Ljava/lang/String;Ljava/lang/String;JJIZ)[J+0
j  ai.rapids.cudf.Table.readParquet(Lai/rapids/cudf/ParquetOptions;Lai/rapids/cudf/HostMemoryBuffer;JJ)Lai/rapids/cudf/Table;+122
j  com.nvidia.spark.rapids.MultiFileParquetPartitionReader.$anonfun$readToTable$1(Lai/rapids/cudf/ParquetOptions;Lai/rapids/cudf/HostMemoryBuffer;JLcom/nvidia/spark/rapids/NvtxWithMetrics;)Lai/rapids/cudf/Table;+4
j  com.nvidia.spark.rapids.MultiFileParquetPartitionReader$$Lambda$2380.apply(Ljava/lang/Object;)Ljava/lang/Object;+16
j  com.nvidia.spark.rapids.Arm.withResource(Ljava/lang/AutoCloseable;Lscala/Function1;)Ljava/lang/Object;+2
```

**Steps/Code to reproduce bug**
1. Start (py)Spark REPL with RAPIDS plugin and GPU enabled   

2. load the file from attached to SPARK-32639 
```python
spark.read.schema('value MAP<STRUCT<first:STRING, middle:STRING, last:STRING>, STRING>').parquet("/home/gshegalov/gits/spark-rapids/integration_tests/src/test/resources/SPARK-32639/000.snappy.parquet").take(1)
```

**Expected behavior**
[...]

PR #1578, opened 2021-01-25
Test to track RAPIDS-side issues re SPARK-32639

### LINKED ISSUE #1463: [FEA] Add Parquet GroupType mapkey field test
[SPARK-32639](https://issues.apache.org/jira/browse/SPARK-32639) fixed a Parquet parsing issue when handling a particular form of map encoding in Parquet.  A sample file is attached to that JIRA.  It would be good to add a plugin unit test that verifies the plugin can load this file properly as well.

gerashegalov:
This PR closes #1463. The test documents the scenario that fails on Spark (CPU) prior to Spark 3.1.0. 

It results in an executor JVM crash on GPU #1576. Since xfail does not handle the process crash gracefully and fails pytest, skip is used instead.


gerashegalov:
build

integration_tests/src/main/python/parquet_test.py
```diff
@@ -14,7 +14,7 @@
```
jlowe, line 14: Need to add 2021 copyrights

jlowe (changes requested):

jlowe (approved):

jlowe:
build

pxLi:
build

PR #1579, opened 2021-01-25
Support TakeOrderedAndProject

### LINKED ISSUE #103: [FEA] GPU version of TakeOrderedAndProject
Because limit is not always deterministic it is typically combined with an order by.  In theses cases a `TakeOrderedAndProject` exec is inserted into the plan.  It would be great to have a GPU enabled version of this to speed up these types of queries and to reduce the amount of data that goes across the PCIe bus.  Inherently this must look at all of the data to do its job, and if we don't have a GPU version of this we have to stream possibly huge amounts of data back from the GPU to the CPU so it can throw out most of it.

Conceptually it should be something like.

```
for each batch:
  combine batch with previous result if it exists
  sort combined result
  save top N results from combined batch
end
return final top N results
```

### LINKED ISSUE #1575: [DOC] Document differences between CPU and GPU plans
**Report needed documentation**
When comparing CPU and GPU query plans, usually there is a straightforward 1-to-1 mapping of CPU operation to GPU operation.  However there are some exceptions that can confuse users:

- Lack of whole stage codegen
- sort-merge-join becoming shuffled hash join
- coalesce batch nodes
- others?

**Describe the documentation you'd like**
It would be nice if the FAQ had a section describing the significant differences between the mapping of CPU to GPU plans and why those differences exist.

revans2:
This fixes #1575 and fixes #103

I did not test topN with nested types going along for the ride (not being sorted on, but in the table that is being sorted), so I didn't turn them on.  They should work, but it was not a requirement.

My initial performance tests showed a little bit of a performance gain over not supporting take ordered and project, but it was not a huge amount of data so it is hard to tell.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -17,6 +17,33 @@ with `collect`, `show` or `write` a new `DataFrame` is constructed causing Spark
 query. This is why `spark.rapids.sql.enabled` is still respected when running, even if explain shows
 stale results.
 
+### Why does the plan for the GPU query look different from the CPU query?
+
+Typically, there is a one to one mapping between CPU stages in a plan and GPU stages.  There are a
+few places where this is not the case.
+
+* `WholeStageCodeGen` - The GPU plan typically does not do code generation, and does not support generating code for an
+entire stage in the plan. Code generation is typically used to reduce the cost of processing data
+one row at a time. The GPU plan processes the data in a columnar format, so the costs are different.
+
+* ColumnarToRow and RowToColumnar Transitions - The CPU version of Spark plans typically process data in a row based format. The main exception to
+this is reading some kinds of columnar data, like parquet. When transitioning between the CPU and
+the GPU require transitions between row and columnar formatted data.
+
+* `GpuCoalesceBatches` and `GpuShuffleCoalesce` - Processing data on the GPU often scales sublinerarly. That means doubling the data does takes less
+than half the time. Because of this we want to process larger batches of data when possible. These
+operators will try to build larger batches of data to process.
+
+* `SortMergeJoin` - The RAPIDS accelerator does not support sort merge joins yet. For now we
+translate sort merge joins into shuffled hash joins. Because of this there are times when
+Sorts may be removed or other sorts added to meet the ordering requirements of the query.
+
+* `TakeOrderedAndProject` - The `TakeOrderedAndProject` operator will take the top N entries in each task,
+shuffle the results to a single executor and then take the top N results from that. The GPU
+plan often has more metrics than the CPU versions do, and when we tried to combine all of these
+operations into a single stage the metrics where confusing to understand what was happening. Instead
+we split the single stage up into multiple smaller parts so the metrics are clearer.
+
 ### What versions of Apache Spark does the RAPIDS Accelerator for Apache Spark support?
 
 The RAPIDS Accelerator for Apache Spark requires version 3.0.0 or 3.0.1 of Apache Spark. Because the
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -22,27 +22,31 @@ stale results.
 Typically, there is a one to one mapping between CPU stages in a plan and GPU stages.  There are a
 few places where this is not the case.
 
-* `WholeStageCodeGen` - The GPU plan typically does not do code generation, and does not support generating code for an
-entire stage in the plan. Code generation is typically used to reduce the cost of processing data
-one row at a time. The GPU plan processes the data in a columnar format, so the costs are different.
+* `WholeStageCodeGen` - The GPU plan typically does not do code generation, and does not support 
+  generating code for an entire stage in the plan. Code generation reduces the cost of processing 
+  data one row at a time. The GPU plan processes the data in a columnar format, so the costs 
+  of processing a batch is amortized over the entire batch of data and code generation is not
+  needed.
 
-* ColumnarToRow and RowToColumnar Transitions - The CPU version of Spark plans typically process data in a row based format. The main exception to
-this is reading some kinds of columnar data, like parquet. When transitioning between the CPU and
-the GPU require transitions between row and columnar formatted data.
+* `ColumnarToRow` and `RowToColumnar` transitions - The CPU version of Spark plans typically process 
+  data in a row based format. The main exception to this is reading some kinds of columnar data, 
+  like Parquet. Transitioning between the CPU and the GPU also requires transitioning between row
+  and columnar formatted data.
 
-* `GpuCoalesceBatches` and `GpuShuffleCoalesce` - Processing data on the GPU often scales sublinerarly. That means doubling the data does takes less
-than half the time. Because of this we want to process larger batches of data when possible. These
-operators will try to build larger batches of data to process.
+* `GpuCoalesceBatches` and `GpuShuffleCoalesce` - Processing data on the GPU scales 
+  sublinearly. That means doubling the data does often takes less than half the time. Because of
+  this we want to process larger batches of data when possible. These operators will try to combine
+  smaller batches of data into fewer, larger batches to process more efficiently.
 
-* `SortMergeJoin` - The RAPIDS accelerator does not support sort merge joins yet. For now we
-translate sort merge joins into shuffled hash joins. Because of this there are times when
-Sorts may be removed or other sorts added to meet the ordering requirements of the query.
+* `SortMergeJoin` - The RAPIDS accelerator does not support sort merge joins yet. For now, we 
+  translate sort merge joins into shuffled hash joins. Because of this there are times when sorts 
+  may be removed or other sorts added to meet the ordering requirements of the query.
 
-* `TakeOrderedAndProject` - The `TakeOrderedAndProject` operator will take the top N entries in each task,
-shuffle the results to a single executor and then take the top N results from that. The GPU
-plan often has more metrics than the CPU versions do, and when we tried to combine all of these
-operations into a single stage the metrics where confusing to understand what was happening. Instead
-we split the single stage up into multiple smaller parts so the metrics are clearer.
+* `TakeOrderedAndProject` - The `TakeOrderedAndProject` operator will take the top N entries in 
+  each task, shuffle the results to a single executor and then take the top N results from that. 
+  The GPU plan often has more metrics than the CPU versions do, and when we tried to combine all of 
+  these operations into a single stage the metrics were confusing to understand. Instead, we split 
+  the single stage up into multiple smaller parts, so the metrics are clearer.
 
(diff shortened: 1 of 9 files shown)


revans2:
build

docs/FAQ.md
```diff
@@ -17,6 +17,33 @@ with `collect`, `show` or `write` a new `DataFrame` is constructed causing Spark
...
 stale results.
 
+### Why does the plan for the GPU query look different from the CPU query?
+
+Typically, there is a one to one mapping between CPU stages in a plan and GPU stages.  There are a
+few places where this is not the case.
+
+* `WholeStageCodeGen` - The GPU plan typically does not do code generation, and does not support generating code for an
+entire stage in the plan. Code generation is typically used to reduce the cost of processing data
+one row at a time. The GPU plan processes the data in a columnar format, so the costs are different.
```
jlowe, line 27: Would be good to elaborate here, e.g.: mentioning GPU data is processed in batches of rows, and batches often contain many thousands of rows.

docs/FAQ.md
```diff
@@ -17,6 +17,33 @@ with `collect`, `show` or `write` a new `DataFrame` is constructed causing Spark
...
+
+* ColumnarToRow and RowToColumnar Transitions - The CPU version of Spark plans typically process data in a row based format. The main exception to
+this is reading some kinds of columnar data, like parquet. When transitioning between the CPU and
+the GPU require transitions between row and columnar formatted data.
```
jlowe, line 31:
```suggestion
this is reading some kinds of columnar data, like Parquet. Transitioning between the CPU and
the GPU requires transitioning between row and columnar formatted data.
```

docs/FAQ.md
```diff
@@ -17,6 +17,33 @@ with `collect`, `show` or `write` a new `DataFrame` is constructed causing Spark
...
+
+* `GpuCoalesceBatches` and `GpuShuffleCoalesce` - Processing data on the GPU often scales sublinerarly. That means doubling the data does takes less
```
jlowe, line 33:
```suggestion
* `GpuCoalesceBatches` and `GpuShuffleCoalesce` - Processing data on the GPU scales sublinearly. That means doubling the data often takes less
```

docs/FAQ.md
```diff
@@ -17,6 +17,33 @@ with `collect`, `show` or `write` a new `DataFrame` is constructed causing Spark
...
+than half the time. Because of this we want to process larger batches of data when possible. These
+operators will try to build larger batches of data to process.
```
jlowe, line 35:
```suggestion
operators collect smaller batches of data into fewer, larger batches to process.
```

docs/FAQ.md
```diff
@@ -17,6 +17,33 @@ with `collect`, `show` or `write` a new `DataFrame` is constructed causing Spark
...
+
+* `SortMergeJoin` - The RAPIDS accelerator does not support sort merge joins yet. For now we
+translate sort merge joins into shuffled hash joins. Because of this there are times when
+Sorts may be removed or other sorts added to meet the ordering requirements of the query.
```
jlowe, line 39:
```suggestion
sorts may be removed or other sorts added to meet the ordering requirements of the query.
```

integration_tests/src/main/python/data_gen.py
```diff
@@ -221,13 +221,13 @@ def __init__(self, precision=None, scale=None, nullable=True, special_cases=[]):
         DECIMAL_MIN = Decimal('-' + ('9' * precision) + 'e' + str(-scale))
```
jlowe, line 221: 2021 copyrights

integration_tests/src/main/python/sort_test.py
```diff
@@ -34,6 +34,14 @@ def test_single_orderby(data_gen, order):
             lambda spark : unary_op_df(spark, data_gen).orderBy(order),
```
jlowe, line 34: 2021 copyrights

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SpillableColumnarBatch.scala
```diff
@@ -44,6 +44,8 @@ trait SpillableColumnarBatch extends AutoCloseable {
    *       with decompressing the data if necessary.
```
jlowe, line 44: 2021 copyrights

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SpillableColumnarBatch.scala
```diff
@@ -82,6 +85,11 @@ class SpillableColumnarBatchImpl (id: TempSpillBufferId,
    */
   def spillId: TempSpillBufferId = id
 
+  override lazy val sizeInBytes: Long =
+    withResource(RapidsBufferCatalog.acquireBuffer(id)) { buff =>
```
jlowe, line 89: Note it is not necessary to go through the buffer acquistion logic here.  The buffer is immutable in the catalog, and since we're not trying to access the actual data but only the metadata we don't need to acquire.  Might be worth adding a `RapidsBufferCatalog.getBufferSize(id)` method but not necessary as part of this PR.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -19,15 +19,19 @@ package com.nvidia.spark.rapids
 import scala.collection.mutable.ArrayBuffer
```
jlowe, line 19: 2021 copyrights

jlowe (changes requested):

docs/FAQ.md
```diff
@@ -17,6 +17,33 @@ with `collect`, `show` or `write` a new `DataFrame` is constructed causing Spark
 query. This is why `spark.rapids.sql.enabled` is still respected when running, even if explain shows
 stale results.
 
+### Why does the plan for the GPU query look different from the CPU query?
+
+Typically, there is a one to one mapping between CPU stages in a plan and GPU stages.  There are a
+few places where this is not the case.
+
+* `WholeStageCodeGen` - The GPU plan typically does not do code generation, and does not support generating code for an
```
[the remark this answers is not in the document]
sameerz: I'd like to keep the line lengths to 100 characters for ease of editing, where possible.

docs/FAQ.md
```diff
@@ -17,6 +17,33 @@ with `collect`, `show` or `write` a new `DataFrame` is constructed causing Spark
...
+
+* `SortMergeJoin` - The RAPIDS accelerator does not support sort merge joins yet. For now we
+translate sort merge joins into shuffled hash joins. Because of this there are times when
+Sorts may be removed or other sorts added to meet the ordering requirements of the query.
+
+* `TakeOrderedAndProject` - The `TakeOrderedAndProject` operator will take the top N entries in each task,
+shuffle the results to a single executor and then take the top N results from that. The GPU
+plan often has more metrics than the CPU versions do, and when we tried to combine all of these
+operations into a single stage the metrics where confusing to understand what was happening. Instead
+we split the single stage up into multiple smaller parts so the metrics are clearer.
```
sameerz, line 45:
```suggestion
plan often has more metrics than the CPU versions do, and when we tried to combine all of these
operations into a single stage the metrics were confusing to understand. Instead we split the single 
stage into multiple smaller parts so the metrics are clearer.
```

docs/FAQ.md
```diff
@@ -17,6 +17,33 @@ with `collect`, `show` or `write` a new `DataFrame` is constructed causing Spark
...
+### Why does the plan for the GPU query look different from the CPU query?
+
+Typically, there is a one to one mapping between CPU stages in a plan and GPU stages.  There are a
+few places where this is not the case.
+
+* `WholeStageCodeGen` - The GPU plan typically does not do code generation, and does not support generating code for an
+entire stage in the plan. Code generation is typically used to reduce the cost of processing data
+one row at a time. The GPU plan processes the data in a columnar format, so the costs are different.
+
+* ColumnarToRow and RowToColumnar Transitions - The CPU version of Spark plans typically process data in a row based format. The main exception to
```
sameerz, line 29: Should ColumnarToRow and RowToColumnar be backquoted, like, for example, `WholeStageCodeGen` is in the previous section?  

```suggestion
* `ColumnarToRow` and `RowToColumnar` transitions - The CPU version of Spark plans typically process data in a row based format. The main exception to
```

docs/FAQ.md
sameerz: Would it be helpful to add a sentence to indicate why processing on the GPU scales sublinearly?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SpillableColumnarBatch.scala
revans2: That was a larger change than I really wanted to do, which is why I left it this way. Yes it is just meta data, but where the meta data is stored is in the RapidsBuffer. If I remember correctly that was because we initially thought about compressing the data when we spilled to host memory or disk. I didn't want to make a design change without verifying that it would be an OK thing to do.

docs/FAQ.md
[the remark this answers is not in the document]
revans2: I can see us trying to limit ourselves to 100 characters most of the time, I have real problems from a change management perspective when we try to make all of the lines as close to 100 characters as possible, like most of this file is.  When you make a single small change to a line that forces it over 100 characters instantly you feel like you need to reformat the rest of the paragraph to make all of the lines as close to 100 characters as possible.  This means git sees a much bigger change and trying to merge conflicting changes becomes a real pain.  Especially because all of those changes have no impact to what the final rendered result looks like.

revans2:
I addressed the review comments and updated the metrics a bit. I was seeing much linger times than I expected for `GpuTopN` total time and I found that I had `iter.next()` included in the total time, which resulted in also including the time it took for the upstream batch in some cases to be built.

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SpillableColumnarBatch.scala
jlowe: Yes we originally thought we could mutate the buffer silently behind the scenes, but we've since moved to a setup where the `RapidsBufferId` describes an immutable piece of data.  That means compressing the data requires re-registering the buffer with a new buffer ID (and corresponding metadata).  We leaned towards this route because there were cases during shuffle where it would be problematic and surprising if a buffer contents mutated.

I'll file a followup to add a buffer size accessor to the catalog that doesn't require acquiring the buffer.

jlowe (approved):

jlowe:
build

revans2:
@sameerz please take another look

sameerz (approved):

PR #1581, opened 2021-01-25
Support Decimal type for CollectLimitExec

razajafri:
I also added a python test for convention and consistency


integration_tests/src/main/python/collect_limit_test.py
```diff
@@ -0,0 +1,26 @@
...
+import pytest
+
+from data_gen import *
+from asserts import assert_gpu_and_cpu_are_equal_collect
+
+allow_collect_conf={"spark.rapids.sql.exec.CollectLimitExec": "true"}
+
+@pytest.mark.parametrize('data_gen', all_gen)
+def test_collect_limit(data_gen):
+    assert_gpu_and_cpu_are_equal_collect(lambda spark: unary_op_df(spark, data_gen).limit(10),
```
jlowe, line 24: Do we have issues with this test when there are multiple shuffle partitions?  It doesn't seem deterministic as written.

razajafri: I ran the test a couple of times locally without a problem. Let me add a `repartition` before the limit.

razajafri: So, the moment I add a `repartition` it starts to fail like you suspected. If I add `orderBy` or `sort`, `CollectLimitExec` is not executed at all. So, I have now made the limit = length of df, let me know if this is good. 

Any other suggestions to make the test better will be appreciated

revans2: This type of a test is going to be non-deterministic.  If we fixed the GPU version to ramp up like the CPU version does, then it would be much more likely to be deterministic, but until then I am fine with us just not having an automated test. We didn't have one before. The code is disabled by default. Lets just manually test it.

razajafri: what do you think about adding a scala test instead?

razajafri:
This was tested manually. @revans2 @jlowe PTAL

revans2 (approved):

jlowe (approved):

razajafri:
build

jlowe:
build

Issue #1572, opened 2021-01-22
[FEA] Add in decimal support for pmod, remainder and divide


revans2:
**Is your feature request related to a problem? Please describe.**
This is a follow on issue to #1329 rapidsai/cudf#7132 is not predicted to be done until the [0.19 release of cudf](https://github.com/rapidsai/cudf/issues/7132#issuecomment-763221946). As such I split these off.

 - [ ] pmod - cudf issue: https://github.com/rapidsai/cudf/issues/7132
 - [ ] remainder - cudf issue: https://github.com/rapidsai/cudf/issues/7132
 - [x] divide - - cudf issue: https://github.com/rapidsai/cudf/issues/7132

revans2:
Divide went in early to cudf.  It is not a complete solution because like multiply we cannot support the full range of precision values.

sameerz:
`mod` / `pmod` support depends on https://github.com/rapidsai/cudf/pull/10179

PR #1592, opened 2021-01-26
Add in support for Decimal divide

revans2:
This is blocked waiting on https://github.com/rapidsai/cudf/pull/7198 to be merged in.

This is similar to multiply where we cannot support the full range of scale/precision that spark can, not just because of 64-bit limitations.

This is a part of #1572

This also fixes two bugs that I found.

One were GpuCheckOverflow was not rounding when it should and the other was for DivModLike where the null output was getting the wrong type in some cases (especially with decimal values)

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -1627,7 +1627,7 @@ object GpuOverrides {
           // max of the two input non-scale portions + the new scale. Then it will do the divide,
           // which the rules for it are a little complex, but lie about it
           // in the return type of the Divide operator. Then in CheckOverflow it will reset the
-          // scale and check the precision so that they know it fits in final desired result which
+          // scale and check the precision so that they know it fits in final desired result.
           // We would like to avoid all of this if possible because having a temporary intermediate
           // value that can have a scale quite a bit larger than the final result reduces the
           // maximum precision that we could support, as we don't have unlimited precision. But
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -1410,7 +1410,7 @@ object GpuOverrides {
       "Multiplication",
       ExprChecks.binaryProjectNotLambda(TypeSig.numeric +
           TypeSig.psNote(TypeEnum.DECIMAL, "Because of Spark's inner workings the full range " +
-              "of decimal precision (even for 64-bit value) is not supported."),
+              "of decimal precision (even for 64-bit values) is not supported."),
         TypeSig.numeric,
         ("lhs", TypeSig.numeric, TypeSig.numeric),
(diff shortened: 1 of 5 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1614,10 +1614,42 @@ object GpuOverrides {
...
+        ("lhs", TypeSig.DOUBLE + TypeSig.DECIMAL, TypeSig.DOUBLE + TypeSig.DECIMAL),
+        ("rhs", TypeSig.DOUBLE + TypeSig.DECIMAL, TypeSig.DOUBLE + TypeSig.DECIMAL)),
       (a, conf, p, r) => new BinaryExprMeta[Divide](a, conf, p, r) {
+        override def tagExprForGpu(): Unit = {
+          // Division of Decimal types is a little odd. Spark will cast the inputs
+          // to a common wider value where scale is max of the two input scales, and precision is
+          // max of the two input non-scale portions + the new scale. Then it will do the divide,
+          // which the rules for it are a little complex, but lie about it
+          // in the return type of the Divide operator. Then in CheckOverflow it will reset the
+          // scale and check the precision so that they know it fits in final desired result which
```
jlowe, line 1630:
```suggestion
          // scale and check the precision so that they know it fits in final desired result.
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1614,10 +1614,42 @@ object GpuOverrides {
...
+          // right now we actually increase the scale (and corresponding precision) to get an extra
+          // decimal place so we can round it in GpuCheckOverflow
+          (childExprs.head.dataType, childExprs(1).dataType) match {
+            case (l: DecimalType, r: DecimalType) =>
+              val intermediateResult = GpuDivideUtil.decimalDataType(l, r)
+              if (intermediateResult.precision > DType.DECIMAL64_MAX_PRECISION) {
+                willNotWorkOnGpu("The actual output precision of the divide is too large" +
+                    s" to fit on the GPU $intermediateResult")
+              }
+            case _ => // NOOP
```
jlowe, line 1649: Do we not have to worry about other type combinations with Decimal because Catalyst will have already injected any necessary casting to the inputs of this operator, e.g.: dividing a DecimalType by IntType?

revans2: Spark only supports Doubles and Decimals and it makes sure that both inputs match each other.  This is why I updated the integration tests for divide to only take doubles and decimal values, because there is no real reason to test casting a byte to a double too.

jlowe (dismissed):

revans2:
The cudf dependency was merged in this morning. I will try to take this out of draft once CI finishes getting everything ready for the build

jlowe:
build

jlowe:
@revans2 with the recent cudf snapshot published, is this now ready to go?

revans2:
build

revans2:
@jlowe There were merge conflicts that I had to take care of. It should be good to go now.

jlowe (approved):

PR #1593, opened 2021-01-26
Avoid unnecessary Table instances after contiguous split

jlowe:
This depends on https://github.com/rapidsai/cudf/pull/7127.

This change avoids manifesting cudf `Table` and `ColumnVector` instances after a partition via contiguous split, as these buffers are likely to be shipped off to other nodes over the network rather than used as a columnar batch input in the same process.  Instantiating all of the tables and column objects for many splits can take a significant amount of time per task.

This also changes the table metadata used for UCX shuffle to use the new libcudf opaque metadata generated by the `contiguous_split` and `pack` methods, eliminating a lot of `ColumnMeta` flatbuffer classes.


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/RapidsShuffleInternalManagerBase.scala
```diff
@@ -88,6 +88,8 @@ class RapidsCachingWriter[K, V](
   private val uncompressedMetric: SQLMetric = metrics("dataSize")
 
   override def write(records: Iterator[Product2[K, V]]): Unit = {
+    // NOTE: This MUST NOT CLOSE the incoming batches because they are
```
abellina, line 91: :+1:

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBuffer.scala
```diff
@@ -139,24 +138,23 @@ trait RapidsBuffer extends AutoCloseable {
...
-    (0 until meta.columnMetasLength).foreach { i =>
-      meta.columnMetas(columnMeta, i)
-      assert(columnMeta.childrenLength == 0, "child columns are not yet supported")
-      val dtype = DType.fromNative(columnMeta.dtypeId(), columnMeta.dtypeScale())
-      columns(i) = GpuColumnVector.from(new ai.rapids.cudf.ColumnVector(
-        dtype, rowCount, nullCount, null, null, null),
-        sparkTypes(i))
+    val packedMeta = meta.packedMetaAsByteBuffer()
+    if (packedMeta != null) {
+      withResource(DeviceMemoryBuffer.allocate(0)) { deviceBuffer =>
```
abellina, line 149: I am curious if there is an advantage to having `DeviceMemoryBuffer` be non-null in this case. I assume it makes some things easier => we can slice it/add ref count, but then it's extra objects. I am not saying there's anything wrong, just mostly wondering in the mostly degenerate case (some kind of `count`), whether we are better off not having the `DeviceMemoryBuffers` around.

abellina (commented):
I have a single pass, and I am not seeing anything odd. It is nice to see all the columnar-meta code go away in favor of the packed format. I'll do a second pass today.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuCompressedColumnVector.java
```diff
@@ -16,151 +16,66 @@
...
-import org.apache.spark.sql.types.MapType;
-import org.apache.spark.sql.types.StructField;
-import org.apache.spark.sql.types.StructType;
 import org.apache.spark.sql.vectorized.ColumnVector;
 import org.apache.spark.sql.vectorized.ColumnarBatch;
 
+import static org.apache.spark.sql.types.DataTypes.NullType;
+
 /**
  * A GPU column vector that has been compressed. The columnar data within cannot
```
revans2, line 27: I'm not sure that this is accurate.  It really takes the place of an entire table.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CopyCompressionCodec.scala
```diff
@@ -76,20 +75,20 @@ class BatchedCopyCompressor(maxBatchMemory: Long, stream: Cuda.Stream)
...
-          ct.getTable,
-          inBuffer,
+          ct,
           CodecType.COPY,
           outBuffer.getLength)
-        stream.sync()
         CompressedTable(outBuffer.getLength, meta, outBuffer)
       }
     }
+    closeOnExcept(result) { _ => stream.sync() }
```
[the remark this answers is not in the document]
revans2: It is because stream.sync can fail (often catching errors from kernels that were running), and it would be best not to leak if it did.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVectorFromBuffer.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2021, NVIDIA CORPORATION.
```
revans2, line 2: Can we try and find a way to get rid of one of `GpuColumnVectorFromBuffer` or `GpuPackedTableColumn`? If feels odd that we would have both when they are almost exactly the same thing.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDeviceMemoryStore.scala
```diff
@@ -40,23 +40,64 @@ class RapidsDeviceMemoryStore(catalog: RapidsBufferCatalog = RapidsBufferCatalog
...
+      Some(table),
+      contigBuffer,
+      initialSpillPriority)
+    try {
+      logDebug(s"Adding table for: [id=$id, size=${buffer.size}, " +
+          s"meta_id=${buffer.meta.bufferMeta.id}, meta_size=${buffer.meta.bufferMeta.size}]")
+      addBuffer(buffer)
+    } catch {
+      case t: Throwable =>
+        buffer.free()
```
revans2, line 65: why not `closeOnExecpt`?

tests/src/test/scala/com/nvidia/spark/rapids/GpuPartitioningSuite.scala
```diff
@@ -42,8 +42,18 @@ class GpuPartitioningSuite extends FunSuite with Arm {
     }
   }
 
+  private def extractBases(batch: ColumnarBatch): Array[ColumnVector] = {
+    if (GpuPackedTableColumn.isBatchPacked(batch)) {
+      val packedColumn = batch.column(0).asInstanceOf[GpuPackedTableColumn]
+      val table = packedColumn.getContiguousTable.getTable
```
revans2, line 48: nit: This feels a little off.  We just grabbed the table and pulled out the columns. It might be good to have a comment here about how packedColumn still owns the table and the reference counts have not changed for it.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuCompressedColumnVector.java
```diff
@@ -16,151 +16,66 @@
...
    */
-  public static ColumnarBatch from(CompressedTable compressedTable, DataType[] colTypes) {
-    return from(compressedTable.buffer(), compressedTable.meta(), colTypes);
+  public static ColumnarBatch from(CompressedTable compressedTable) {
+    return from(compressedTable.buffer(), compressedTable.meta());
   }
 
   public static boolean isBatchCompressed(ColumnarBatch batch) {
-    if (batch.numCols() == 0) {
+    if (batch.numCols() != 1) {
```
gerashegalov, line 47: consider making this method a one-liner 
```suggestion
    return batch.numCols() == 1 && batch.column(0) instanceof GpuCompressedColumnVector
```

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuPackedTableColumn.java
```diff
@@ -0,0 +1,88 @@
...
+  private final ContiguousTable contigTable;
+
+  public static ColumnarBatch from(ContiguousTable contigTable) {
+    ColumnVector column = new GpuPackedTableColumn(contigTable);
+    return new ColumnarBatch(new ColumnVector[] { column }, (int) contigTable.getRowCount());
+  }
+
+  /** Returns true if this columnar batch uses a packed table */
+  public static boolean isBatchPacked(ColumnarBatch batch) {
+    if (batch.numCols() != 1) {
```
gerashegalov, line 49: similar comment to `isBatchCompressed`

sql-plugin/src/main/format/ShuffleMetadataResponse.fbs
```diff
@@ -16,46 +16,16 @@ include "ShuffleCommon.fbs";
...
   buffer_meta: BufferMeta;
 
-  /// metadata for each column in the table buffer
-  column_metas: [ColumnMeta];
-
   /// number of rows in the table
   row_count: long;
+
+  /// opaque metadata describing the packed table schema and data layout
+  packed_meta: [byte];
```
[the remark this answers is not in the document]
jlowe: I haven't explicitly examined the payload size.  Each column is encoded as a C++ struct in the metadata, and [eyeballing the code](https://github.com/rapidsai/cudf/pull/7096/files#diff-e7e0c690336137aefd4c8099055fb51b39f5f32106034287f651b5fb8442412cR27-R61) I'm guessing it's around 40 bytes per column, nested or otherwise, in the table.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBuffer.scala
jlowe: IIRC I think it's because I couldn't generate the packed metadata without a valid device buffer.

> mostly wondering in the mostly degenerate case (some kind of count), whether we are better off not having the DeviceMemoryBuffers around

For the `count` case we won't be creating the degenerate device buffer because that's a case where there are zero columns, only rows, and libcudf doesn't support tables without columns.  This code path would only be used for the case where there's a table with columns but no rows.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVectorFromBuffer.java
jlowe: I struggled with this at first as well, but they each serve an important and distinct function.  The sole purpose of `GpuPackedTableColumn` is to convey a buffer of data to the next stage without manifesting a `Table` or any `ColumnVector` instances, specifically so we can send these buffers to the GPU shuffle.

`GpuColumnVectorFromBuffer` is a real GPU `ColumnarBatch` and can be operated on as if it were a normal `ColumnarBatch` containing `GpuColumnVector`, just that it came from a contiguous table.  That requires manifesting the table and column instances to create it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDeviceMemoryStore.scala
jlowe: Because `close` and `free` are semantically different.  Calling `close` on a freshly created buffer will throw an "already closed" exception since it was never acquired to "open" it.  `free` will release the resources if the refcount is zero, and that's what we need to do in this case.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CopyCompressionCodec.scala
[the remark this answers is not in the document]
abellina: ok I think I missed we now need to close `meta` also, not just the `DeviceMemoryBuffer`

jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDeviceMemoryStore.scala
```diff
@@ -40,23 +40,64 @@ class RapidsDeviceMemoryStore(catalog: RapidsBufferCatalog = RapidsBufferCatalog
...
+      initialSpillPriority: Long): Unit = {
+    val contigBuffer = contigTable.getBuffer
+    val size = contigBuffer.getLength
+    val meta = MetaUtils.buildTableMeta(id.tableId, contigTable)
+    contigBuffer.incRefCount()
+    val buffer = new RapidsDeviceMemoryBuffer(
+      id,
+      size,
+      meta,
+      None,
```
abellina, line 91: why should this be `None` again? I am just trying to follow that.

jlowe: We don't want to manifest the table, just store the underlying table buffer.  The main change here is to stop assuming it's cheap to get the `Table` instance from a `ContiguousSplit` instance.

gerashegalov (approved):
LGTM

abellina (approved):

revans2 (approved):

PR #1597, opened 2021-01-27
Documentation for Parquet serializer

razajafri:
This PR adds docs for cache serializer. Please critique the content of the documentation. 

As far as the location of this doc lives, we can decide if we want it to live in a standalone document or here. I am agnostic to it.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -930,6 +930,45 @@ object RapidsConf {
...
+      |  at `StorageLevel.scala` in Spark.
+      |
+      |  Starting in Spark 3.1.0 users can add their own cache serializer, if they desire, by
+      |  setting the `spark.sql.cache.serializer` configuration. This is a static configuration
+      |  that is set once for the duration of a Spark application which means that you can only set the conf
+      |  before starting a Spark application and cannot be changed for that application's Spark
+      |  session.
+      |
+      |  RAPIDS Accelerator for Apache Spark version 0.4+ has a serializer `com.nvidia.spark.rapids.shims.spark311.ParquetCachedBatchSerializer`
+      |  that is optimized to run on the GPU and uses parquet to compress data before caching it.
```
sameerz, line 954:
```suggestion
      |  that is optimized to run on the GPU and uses Parquet to compress data before caching it.
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -930,6 +930,45 @@ object RapidsConf {
...
+      |  ParquetCachedBatchSerializer can be used independent of what the value of
+      |  `spark.rapids.sql.enabled` is. If it is set to true then the parquet compression will run
+      |  on GPU if possible, and importantly
```
sameerz, line 957:
```suggestion
      |  ParquetCachedBatchSerializer can be used independent of the value of
      |  `spark.rapids.sql.enabled`. If it is set to true then the Parquet compression will run
      |  on the GPU if possible, and importantly
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -930,6 +930,45 @@ object RapidsConf {
...
+      |  `spark.sql.inMemoryColumnarStorage.enableVectorizedReader` will not be honored as the GPU
+      |  data is always read in as columnar. If `spark.rapids.sql.enabled` is set to false
+      |  the cached objects will still be compressed on the CPU as a part of the caching process.
+      |  Also note that parquet doesn't support CalendarIntervalType or NullType out of the box, but
```
sameerz, line 961:
```suggestion
      |  Also note that Parquet doesn't support CalendarIntervalType or NullType out of the box, but
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -930,6 +930,45 @@ object RapidsConf {
...
+      |  ParquetCachedBatchSerializer does by decomposing intervals to struct containing the
+      |  months, days and microseconds and NullType to Int column containing nulls.
+      |
+      |  To use this serializer please run Spark
+      |  ```
+      |  spark-shell --conf spark.sql.cache.serializer=com.nvidia.spark.rapids.shims.spark311.ParquetCachedBatchSerializer"
```
sameerz, line 967: Will this documentation need to be updated for each version of Spark going forward?  

com.nvidia.spark.rapids.shims.spark311.ParquetCachedBatchSerializer

razajafri: hmm... I will change it so we don't have to keep updating it

revans2: It is about compatibility and we are potentially going to have to keep updating it.

We eventually might need some kind of a chart to indicate the version of the serializer that is supposed to be used with a given version of Spark. But in the short term we can explain the version requirements.

This is very similar to the ShuffleManager implementations that we have.  The reason for this is because the API that a ShuffleManager or a CachedBatchSerializer has to implement is not stable (marked as a DeveloperApi or even worse in the case of ShuffleManager it is private to spark).

https://github.com/apache/spark/blob/5718d64f3104f7a24a9d4b619748bcca03031c48/sql/core/src/main/scala/org/apache/spark/sql/columnar/CachedBatchSerializer.scala#L46-L47

https://github.com/apache/spark/blob/5718d64f3104f7a24a9d4b619748bcca03031c48/core/src/main/scala/org/apache/spark/shuffle/ShuffleManager.scala#L34

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -930,6 +930,45 @@ object RapidsConf {
       |  --conf spark.executorEnv.LIBCUDF_KERNEL_CACHE_PATH="/tmp/cudf-$USER"
       |  ```
       |""".stripMargin)
+
+      printSectionHeader("RAPIDS-Plugin Cache Serializer")
+      // scalastyle:off line.size.limit
+      println("""
+      |  Apache Spark provides an important feature to cache intermediate data and provide
```
revans2, line 937: I really hate having so much non-auto generated info in here.  Even the JIT kernel cache path feels a bit like it is pushing things.

What is the plan for documenting things like this in the future?
1. Caching
2. Alluxio
3. ShuffleManager (not really documented anywhere except in on-prem getting started)
4. UDFs which have their own file, but there is really no documentation for the UDF compiler

We really should be consistent, and I understand if some features are going to be more important to users than others, but if a user has to look in 5 different places to see what we offer I think it is much worse then a long side panel.  Especially if we offer huge amount of functionality that has to be enabled/disabled separately.

If we want it all here then I would like to see us find some way to template the auto generated docs instead of putting it all in a file that is already close to a thousand lines.  I personally would prefer to split it up and perhaps have some sub-directories for distinctly different features. Possibly even with SQL being one of them, because RapidsConf really should probably be more like RapidsSQLConf.

razajafri: @sameerz can we have a discussion on this? @revans2 brings up good points about the configs.md being really long. We could think about splitting it up into multiple files which could go under sub-menus on the left panel?

docs/configs.md
```diff
@@ -325,3 +325,40 @@ Name | Description | Default Value | Notes
...
+  `spark.sql.inMemoryColumnarStorage.enableVectorizedReader` will not be honored as the GPU
+  data is always read in as columnar. If `spark.rapids.sql.enabled` is set to false
+  the cached objects will still be compressed on the CPU as a part of the caching process.
+  Also note that Parquet doesn't support CalendarIntervalType or NullType out of the box, but
+  ParquetCachedBatchSerializer does by decomposing intervals to struct containing the
+  months, days and microseconds and NullType to Int column containing nulls.
+
+  To use this serializer please run Spark
+  ```
+  spark-shell --conf spark.sql.cache.serializer=<fully qualified name of ParquetCachedBatchSerializer>"
```
gerashegalov, line 361: why not spell out the fully qualified name explicitly?

razajafri: I wanted to avoid having to change the doc every time there is a new release of Spark. I am open to changing it back if others also think this is confusing.

revans2: But then where does the user go to get the name of it?

razajafri: good point. reverting

razajafri:
build

razajafri:
@sameerz @revans2 is this good to go?

revans2 (approved):

PR #1599, opened 2021-01-27
Add division by zero tests for Spark 3.1 behavior

### LINKED ISSUE #1464: [BUG] Respect ANSI mode for division-by-zero error [SPARK-33008]
**Describe the bug**
In ANSI mode, when a division by zero occurs performing a divide-like operation (Divide, IntegralDivide, Remainder or Pmod), we are returning an incorrect value. Instead, we should throw an exception, as stated in the SQL standard.

Divide an integer column by zero in a DataFrame with ANSI mode enabled.

gerashegalov:
Adds tests reproducing SPARK-33008 on CPU and GPU. Contributes to #1464 


integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -498,3 +500,40 @@ def test_greatest(data_gen):
             lambda spark : gen_df(spark, gen).select(
                 f.greatest(*command_args)), conf=allow_negative_scale_of_decimal_conf)
 
+
+def _spark_33008_datagen(spark_session: SparkSession) -> DataFrame:
```
jlowe, line 504: I'd like to see a more immediately descriptive name here rather than the indirect description via an Apache Spark JIRA ID.  Some code readers won't know that this is an Apache JIRA ID, and for those that do, they won't know what SPARK-33008 means without needing to look it up.  Ideally we should save both groups the frustration and describe the purpose of this dataset directly in its name.

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -498,3 +500,40 @@ def test_greatest(data_gen):
...
+
+
+def _assert_div_by_zero(session_func, test_div_func, conf):
+    with pytest.raises(Py4JJavaError) as py4jError:
+        session_func(test_div_func, conf)
+    assert 'java.lang.ArithmeticException: divide by zero' in str(py4jError.value.java_exception)
+
+@pytest.mark.parametrize('from_disk', [False, True])
+@pytest.mark.parametrize('ansi_mode', [False, True])
+def test_spark_33008(spark_tmp_path, from_disk, ansi_mode):
```
jlowe, line 518: Same jira-id-as-name issue here.

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -498,3 +500,40 @@ def test_greatest(data_gen):
...
+    if ansi_mode and is_before_spark_310():
+        pytest.xfail('https://github.com/apache/spark/pull/29882')
+
+    ansi_conf = {'spark.sql.ansi.enabled': ansi_mode}
+
+    if from_disk:
```
jlowe, line 524: Why is there an option to round-trip through Parquet for this test?  I'm not sure how that's supposed to impact divide-by-zero testing.  If it's necessary, why not ORC, CSV, etc?  If it's not necessary then let's remove it and save on testing time.

jlowe (changes requested):
Commits need to be signed off per [the contributing document](https://github.com/NVIDIA/spark-rapids/blob/9efcce3b11f604ed1fb35a6eb6be41f60268f881/CONTRIBUTING.md#sign-your-work).

revans2:
> I have updated the PR description to reflect what it does so its apparent right away without having to look up the spark issue

`doesn't return incorrect results` feels really odd to me. Why not ensure divide by 0 returns correct results?

jlowe:
> I have updated the PR description

That headline is very long for a summary or changelog entry, I'd suggest something much shorter, and arguably the Spark JIRA doesn't need to be part of it.  For example, "Add division by zero tests for Spark 3.1 behavior" or something similar.

razajafri:
> > I have updated the PR description
> 
> That headline is very long for a summary or changelog entry, I'd suggest something much shorter, and arguably the Spark JIRA doesn't need to be part of it. For example, "Add division by zero tests for Spark 3.1 behavior" or something similar.

That makes more sense. I have updated

integration_tests/src/main/python/arithmetic_ops_test.py
gerashegalov: It was a more of a speculation on my part that it may play a role, so I wanted to make sure that both cases are covered. 

So far through the test I learned that there is no difference w.r.t to that. It helps me learn the code pieces exercised in both cases. I'll remove it after proving it does not add any value

gerashegalov:
build

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -500,3 +501,32 @@ def test_greatest(data_gen):
             lambda spark : gen_df(spark, gen).select(
                 f.greatest(*command_args)), conf=allow_negative_scale_of_decimal_conf)
 
+def _assert_div_by_zero(session_func, test_div_func, conf):
```
revans2, line 504: Is it worth making something like this generic in asserts? I could see us wanting to use it in other situations.

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -500,3 +501,32 @@ def test_greatest(data_gen):
...
+def test_div_by_zero(ansi_mode, exp_type):
+    if ansi_mode == 'ansi':
+        if is_before_spark_310() or exp_type != 'bothLiterals':
+            pytest.xfail('https://github.com/apache/spark/pull/29882')
+
+    ansi_conf = {'spark.sql.ansi.enabled': ansi_mode == 'ansi'}
+    data_gen = lambda spark: two_col_df(spark, IntegerGen(), IntegerGen(min_val=0, max_val=0), length=1)
+
+    if exp_type == 'bothLiterals':
+        div_by_zero_func = lambda spark: data_gen(spark).select(f.lit(1) / f.lit(0))
```
revans2, line 520: I don't think we need to test 'bothLiterals'.  In Spark when both parts of an operation are literal values it runs that code as a part of optimizations and will replace it with the a new literal that is the result of the operation before our code even has a chance to run.  So you are only testing that Spark is behaving correctly, not our code.

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -500,3 +501,32 @@ def test_greatest(data_gen):
...
+    with pytest.raises(Py4JJavaError) as py4jError:
+        session_func(test_div_func, conf)
+    assert 'java.lang.ArithmeticException: divide by zero' in str(py4jError.value.java_exception)
+
+@pytest.mark.parametrize('ansi_mode', ['nonAnsi', 'ansi'])
+@pytest.mark.parametrize('exp_type', ['bothLiterals', 'justZeroLiteral', 'noLiterals'])
+def test_div_by_zero(ansi_mode, exp_type):
+    if ansi_mode == 'ansi':
+        if is_before_spark_310() or exp_type != 'bothLiterals':
+            pytest.xfail('https://github.com/apache/spark/pull/29882')
```
jlowe, line 514: The `exp_type != 'bothLiterals'` logic doesn't make sense to me given the xfail is pointing at a Spark Github issue that is marked as fixed in 3.1.0.  If the latter check doesn't work because of a plugin issue instead of a Spark issue, we should have a different xfail reason pointing at an unresolved plugin issue.  As it is, it's pointing to an issue that is fixed in Spark 3.1.0, so it's not clear why we expect this to fail if we're using Spark 3.1.0.

jlowe (changes requested):

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -500,3 +501,32 @@ def test_greatest(data_gen):
...
+    if exp_type == 'bothLiterals':
+        div_by_zero_func = lambda spark: data_gen(spark).select(f.lit(1) / f.lit(0))
+    elif exp_type == 'justZeroLiteral':
+        div_by_zero_func = lambda spark: data_gen(spark).select(f.col('a') / f.lit(0))
+    else:
+        div_by_zero_func = lambda spark: data_gen(spark).select(f.col('a') / f.col('b'))
+
+    if ansi_mode == 'ansi':
+        div_by_zero_func_collect = lambda spark: div_by_zero_func(spark).collect()
+        _assert_div_by_zero(with_cpu_session, div_by_zero_func_collect, ansi_conf)
```
revans2, line 528: I'm not sure we need to test that the CPU version does what we expect it to do. If we had something that would adjust our behavior to match what the CPU does, that might be interesting, but if we are not doing that, then I would just want to check that we do what we expect.

integration_tests/src/main/python/arithmetic_ops_test.py
gerashegalov: `exp_type != bothLiterals` is documenting that we are inheriting the fix from Spark during constant folding

integration_tests/src/main/python/arithmetic_ops_test.py
gerashegalov: will do

integration_tests/src/main/python/arithmetic_ops_test.py
gerashegalov: yes that documents that we are covered with the checks in constant folding

revans2: I am fine with it, but it is inconsistent with all of the other tests that we have done.

integration_tests/src/main/python/arithmetic_ops_test.py
gerashegalov: this is in line how we assert getting equivalent results for non exception cases via `assert_gpu_and_cpu_are_equal_collect`. Per your suggestion, I can move it to the generic layer and have something like `assert_gpu_and_cpu_same_error(func, conf, errorMsg)`

gerashegalov:
Current status for Spark 3.1.1:
```pytest
../../src/main/python/arithmetic_ops_test.py::test_div_by_zero[bothLiterals-nonAnsi] PASSED [ 16%]
../../src/main/python/arithmetic_ops_test.py::test_div_by_zero[bothLiterals-ansi] PASSED [ 33%]
../../src/main/python/arithmetic_ops_test.py::test_div_by_zero[justZeroLiteral-nonAnsi] PASSED [ 50%]
../../src/main/python/arithmetic_ops_test.py::test_div_by_zero[justZeroLiteral-ansi] XFAIL [ 66%]
../../src/main/python/arithmetic_ops_test.py::test_div_by_zero[noLiterals-nonAnsi] PASSED [ 83%]
../../src/main/python/arithmetic_ops_test.py::test_div_by_zero[noLiterals-ansi] XFAIL [100%]

XFAIL ../../src/main/python/arithmetic_ops_test.py::test_div_by_zero[justZeroLiteral-ansi]
  reason: https://github.com/apache/spark/pull/29882
XFAIL ../../src/main/python/arithmetic_ops_test.py::test_div_by_zero[noLiterals-ansi]
  reason: https://github.com/apache/spark/pull/29882
```

Next PR will fix them

integration_tests/src/main/python/arithmetic_ops_test.py
jlowe: I still don't get why this isn't just `is_before_spark_310`.  IIUC we can't get valid CPU results before Spark 3.1.0, and the plugin should pass this test for Spark >= 3.1.0, so why do we want or need the other check?  Apologies if I'm missing the basics here.

revans2 (approved):
It looks OK to me now, but I want to be sure that @jlowe is OK with it too

jlowe (approved):

jlowe (approved):

gerashegalov:
build

gerashegalov:
Adds tests reproducing SPARK-33008 on CPU and GPU. Contributes to #1464 

gerashegalov:
Adds tests reproducing SPARK-33008 on CPU and GPU. Contributes to #1464 

PR #1602, opened 2021-01-27
Increase the Databricks cluster's timeout [skip ci]

NvTimLiu:
Increase the Databricks cluster's timeout

As more test cases are added, the timeout of Databricks cluster needs to be increased  to run all the tests.


NvTimLiu:
oops, changed other files, drop it

Issue #1573, opened 2021-01-22
[BUG] TPCDS Q78 produces a slightly different result for SF1 with decimals


revans2:
**Describe the bug**
When running TPCDS-Q78 in the integration tests against scale factor 1 data with decimals the ratio value comes back different than expected.

We get back a `0.57` vs a `0.58` for ratio on line 66. I'll try to debug this more and see what I can come up with.

**Steps/Code to reproduce bug**
For now there is a slight bug in the code and you need to apply 

```diff
diff --git a/integration_tests/src/main/python/conftest.py b/integration_tests/src/main/python/conftest.py
index 8339f0bb..54a3bcdb 100644
--- a/integration_tests/src/main/python/conftest.py
+++ b/integration_tests/src/main/python/conftest.py
@@ -371,7 +371,7 @@ class TpcdsRunner:
     }
     if not self.tpcds_format in formats:
         raise RuntimeError("{} is not a supported tpcds input type".format(self.tpcds_format))
-    formats.get(self.tpcds_format)(jvm_session, self.tpcds_path, True)
+    formats.get(self.tpcds_format)(jvm_session, self.tpcds_path, True, True)
 
   def do_test_query(self, query):
     spark = get_spark_i_know_what_i_am_doing()
```

**Expected behavior**
The test should pass

revans2:
It is only ratio that is showing issues, but it is showing them in more than one place.

revans2:
It looks like round is doing the wrong thing for HALF_UP rounding. I expanded the input to it and it looks like `round(0.575, 2)` and `round(1.025, 2)` are doing the wrong thing on the GPU.  It is an explicit corner case that we should pay more attention to in the tests. I'll try and track this down to see what data types this impacts.

revans2:
I am going to keep pushing, but it looks like cudf is not doing the correct thing for rounding on some floating point values. Both float and double are impacted, but it looks like they are not impacted the same by similar results.

```
scala> Seq(1.025, 0.575, 2.005, 2.015, 2.025).toDF("f").repartition(1).selectExpr("round(CAST(f AS FLOAT), 2)").collect
21/01/22 17:34:27 WARN GpuOverrides: 
*Exec <ProjectExec> could run on GPU
  *Expression <Alias> round(cast(f#80 as float), 2) AS round(CAST(f AS FLOAT), 2)#82 could run on GPU
    *Expression <Round> round(cast(f#80 as float), 2) could run on GPU
[...]

PR #1609, opened 2021-01-27
Disable float round/bround by default

revans2:
This is related to #1573. It does not fix the underlying problem, but it disables the problematic operators by default.

I decided to reuse the incompat config for these instead of creating a new config. I can change that if we feel it would be better.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -29,12 +29,21 @@ not true for sorting. For all versions of the plugin `-0.0` == `0.0` for sorting
 
 ## Floating Point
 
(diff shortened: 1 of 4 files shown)


revans2:
build

jlowe (dismissed):
The existing incompat flag seems appropriate here.

revans2:
I just realized I didn't include the config to enable them in the explain message so I will do that before we can merge this in.

revans2:
build

revans2:

jlowe (approved):

PR #1610, opened 2021-01-27
Limit length of spark app name in BenchmarkRunner

andygrove:
Only include query name in spark app name when a single query is being run.

I needed this to resolve an issue when running automated benchmarks in Kubernetes, where the long app name exceeded the allowed length for pod names.


andygrove:
build

nartal1 (approved):

Issue #1421, opened 2020-12-17
[FEA] Integration tests should have a flag to avoid silently skipping some tests


jlowe:
**Is your feature request related to a problem? Please describe.**
Some integration tests are silently skipped because the environment isn't setup right, like avoiding testing timestamps when the timezone isn't UTC or I/O tests when the I/O path parameter isn't set.  The problem with this approach is we may run integration tests and not realize these tests are always getting silently skipped due to a problem setting up the environment correctly.  Large chunks of functionality could go untested because nobody notices we're skipping the tests.

**Describe the solution you'd like**
The integration tests should support a command-line parameter that indicates some tests that would normally be silently skipped should instead fail because the environment is not properly setup.

sameerz:
Once this is done, all builds need to be updated to enable all tests.

sameerz:
This issue has a dependency on unifying test cases for all types of builds.  If this does not get done by the end of the Feb 26 milestone, this issue will carry forward as a P1 to release 0.5.

sameerz:
Depends on issue #1568

PR #1611, opened 2021-01-27
Add in flag for integration tests to not skip required tests

revans2:
This addresses the first part of #1421 

I am going to see if pre-commit tests pass first and then look into updating other test instances.

### DIFF
--- a/integration_tests/pytest.ini
+++ b/integration_tests/pytest.ini
@@ -1,4 +1,4 @@
-; Copyright (c) 2020, NVIDIA CORPORATION.
+; Copyright (c) 2020-2021, NVIDIA CORPORATION.
 ;
 ; Licensed under the Apache License, Version 2.0 (the "License");
 ; you may not use this file except in compliance with the License.
--- a/integration_tests/pytest.ini
+++ b/integration_tests/pytest.ini
@@ -21,4 +21,4 @@ markers =
     limit(num_rows): Limit the number of rows that will be check in a result
     qarun: Mark qa test
     cudf_udf: Mark udf cudf test
-
+    rapids_udf_example_native: test UDFs that require custom cuda compilation
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -66,6 +66,12 @@ else
         echo "xdist not installed cannot run tests in parallel"
     fi
 
+    ACCEPTANCE_PARAM=""
+    if [[ "${TEST_ACCEPTANCE}" != "" ]];
+    then
+        ACCEPTANCE_PARAM="--acceptance_tests"
+    fi
+
     if [[ ${TEST_PARALLEL} -lt 2 ]];
     then
         # With xdist 0 and 1 are the same parallelsm but
@@ -96,6 +102,7 @@ else
           -v -rfExXs "$TEST_TAGS" \
           --std_input_path="$SCRIPTPATH"/src/test/resources/ \
           --color=yes \
+          $ACCEPTANCE_PARAM \
           "$TEST_ARGS" \
           $RUN_TEST_PARAMS \
           "$@"
@@ -110,6 +117,7 @@ else
           -v -rfExXs "$TEST_TAGS" \
           --std_input_path="$SCRIPTPATH"/src/test/resources/ \
           --color=yes \
+          $ACCEPTANCE_PARAM \
           "$TEST_ARGS" \
           $RUN_TEST_PARAMS \
           "$@"
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020-2021, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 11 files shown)


revans2:
build

integration_tests/run_pyspark_from_build.sh
```diff
@@ -66,6 +66,12 @@ else
         echo "xdist not installed cannot run tests in parallel"
```
jlowe, line 66: 2021 copyrights

integration_tests/run_pyspark_from_build.sh
```diff
@@ -66,6 +66,12 @@ else
...
     fi
 
+    ACCEPTANCE_PARAM=""
+    if [[ "${TEST_ACCEPTANCE}" != "" ]];
```
jlowe, line 70: So this variable simply has to exist and be non-empty?  Behavior might be surprising if it was set to `0`, `NO`, etc.  Should this be checking for `true` or `True` instead of any non-empty val?

integration_tests/src/main/python/rapids_udf_test.py
```diff
@@ -19,10 +19,14 @@
...
 def skip_if_no_hive(spark):
     if spark.conf.get("spark.sql.catalogImplementation") != "hive":
-        pytest.skip("The Spark session does not have Hive support")
+        if is_acceptance():
+            assert False, "The Spark session does not have Hive support during acceptance tests"
+        else:
+            pytest.skip("The Spark session does not have Hive support")
 
 def load_udf_or_skip_test(spark, udfname, udfclass):
     spark.sql("DROP TEMPORARY FUNCTION IF EXISTS {}".format(udfname))
```
jlowe, line 32: Eventually we will also want to fix the `AnalysisException` just below here for at least the non-native code example UDFs to ensure they run.  Doesn't have to be part of this PR.

integration_tests/src/main/python/udf_cudf_test.py
```diff
@@ -14,15 +14,21 @@
```
jlowe, line 14: 2021 copyrights

integration_tests/src/main/python/udf_test.py
```diff
@@ -14,15 +14,21 @@
```
jlowe, line 14: 2021 copyrights

revans2:
@jlowe the pre-commit build is failing on the native UDFs not being configured to run.  I think that is expected because we didn't want to require hive to build/run for these tests?  If not I can try and fix it. If so do we want to adjust this so we say what type of tests are being run instead (AKA precommit, nightly, developer)?

jlowe:
> I think that is expected because we didn't want to require hive to build/run for these tests?

No, pretty sure Hive is being leveraged by other integration tests during precommit runs.

There are two flavors of Hive UDF example tests: Java and native.  The latter requires building in an environment where libcudf was also built which is time consuming and IMO unnecessary for a precommit build.  Note that without the native example test we'll still be exercising the plugin code around Hive UDFs via the other tests that don't require a native code environment to build the example code (e.g.: `URLDecode` and `URLEncode` tests which are not tied to the `rapids_udf_example_native` tests option).

> If so do we want to adjust this so we say what type of tests are being run instead (AKA precommit, nightly, developer)?

Might be useful for tests that are too involved or time consuming to warrant running during precommit (like the Hive UDF native code example test).  I don't think it's required for this PR, we could track tackling the native UDF test in a followup.

revans2:
build

revans2:
@jlowe I have updated the tests to know what kind of tests are being run.

integration_tests/pytest.ini
```diff
@@ -21,4 +21,4 @@ markers =
     limit(num_rows): Limit the number of rows that will be check in a result
```
jlowe, line 21: 2021 copyrights

integration_tests/src/main/python/conftest.py
```diff
@@ -62,6 +62,29 @@ def is_databricks_runtime():
...
     return runtime_env() == "emr"
 
+_is_nightly_run = False
+_is_precommit_run = False
+
+def is_nightly_run():
+    return _is_nightly_run
+
+def is_at_least_precommit_run():
+    return _is_ngihtly_run or _is_precommit_run
```
jlowe, line 72:
```suggestion
    return _is_nightly_run or _is_precommit_run
```

integration_tests/src/main/python/conftest.py
```diff
@@ -62,6 +62,29 @@ def is_databricks_runtime():
...
+
+def require_nightly_run_tests(description):
```
jlowe, line 74: Nit: IMO this function name doesn't convey that it will skip the test.  Maybe because I'm thinking it acts like `require` from Scala.  Maybe `skip_unless_nightly_tests` would be more explicit?

integration_tests/src/main/python/conftest.py
```diff
@@ -62,6 +62,29 @@ def is_databricks_runtime():
...
+    if (_is_nightly_run):
+        raise AssertionError(description + ' during nightly test run')
+    else: 
+        pytest.skip(description)
+
+def require_precommit_run_tests(description):
```
jlowe, line 80: Similar name confusion here.

revans2:
build

jlowe (changes requested):

revans2:
build

jlowe (approved):

PR #1615, opened 2021-01-28
Add in window support for average

revans2:
This adds in support for average as a window operation.

In reality I was asked to see if average worked with decimal values in CUDF for window operations, and it does not. But, Spark will do similar things for averages on decimal values that fit in 64 bits that it does for hashed aggregates (get the unscaled long out and operate on that). So we can now support some decimal queries with average over a window.

### DIFF
--- a/integration_tests/src/main/python/window_function_test.py
+++ b/integration_tests/src/main/python/window_function_test.py
@@ -82,9 +82,12 @@ def test_window_aggs_for_rows(data_gen):
         '   (partition by a order by b,c rows between UNBOUNDED preceding and UNBOUNDED following) as count_1, '
         ' count(c) over '
         '   (partition by a order by b,c rows between UNBOUNDED preceding and UNBOUNDED following) as count_c, '
+        ' avg(c) over '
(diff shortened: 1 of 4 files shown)


revans2:
build

integration_tests/src/main/python/window_function_test.py
```diff
@@ -82,9 +82,12 @@ def test_window_aggs_for_rows(data_gen):
         '   (partition by a order by b,c rows between UNBOUNDED preceding and UNBOUNDED following) as count_1, '
```
jlowe, line 82: 2021 copyrights

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -387,7 +387,8 @@ case class GpuCount(children: Seq[Expression]) extends GpuDeclarativeAggregate
     Aggregation.count(false).onColumn(inputs.head._2)
```
jlowe, line 387: 2021 copyrights

revans2:
build

jlowe (approved):

PR #1616, opened 2021-01-28
Cost-based optimizer

andygrove:

This PR implements a simple cost-based optimizer that will attempt to avoid transitioning from CPU to GPU when there is no benefit to running on GPU.

There are two rules implemented:

1. Do not move from CPU to GPU for an operator that has no benefit on GPU, such as a projection that just changes the order of some attributes
2. Review sections of the plan that are on GPU and move them back to the CPU if the overall cost does no longer justifies being on the GPU.

There are two aspects to the work here - a cost model, and a mechanism for applying the cost model. We may want to experiment with different cost models or even consider allowing users to provide their own cost models tuned to their workloads.

## Explain output

There is a new indicator char of `$` to show which operators/expressions have been forced onto CPU by the cost-based optimizer:

```
$Exec <SortExec> cannot run on GPU because Removed by cost-based optimizer
  $Expression <SortOrder> more_strings_2#58 ASC NULLS FIRST cannot run on GPU because Removed by cost-based optimizer
    $Expression <AttributeReference> more_strings_2#58 cannot run on GPU because Removed by cost-based optimizer
  !Exec <ShuffleExchangeExec> cannot run on GPU because Columnar exchange without columnar children is inefficient
```

This is followed by a summary of the optimizations that were applied:

```
It is not worth moving to the GPU just for operator: Exchange hashpartitioning(more_strings_2#58, 200), true, [id=#89]
It is not worth moving to the GPU just for operator: Project [more_strings_2#58, more_strings_1#44]
It is no longer worth keeping this section on GPU; gpuCost=5.1000000000000005, cpuCost=5.0:
SortMergeJoin [strings#40], [strings#54], Inner, (cast(more_strings_2#58 as timestamp) < cast(more_strings_1#44 as timestamp))
  Sort [strings#54 ASC NULLS FIRST], false, 0
```

This is still WIP but getting closer.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -0,0 +1,153 @@
...
+    val childCosts = plan.childPlans.map(optimize)
+    val childCpuCosts = childCosts.map(_._1)
+    val childGpuCosts = childCosts.map(_._2)
+
+    // get the CPU and GPU cost of this operator
+    val (operatorCpuCost, operatorGpuCost) = costModel.applyCost(plan)
+
+    // calculate total (this operator + children)
+    val totalCpuCost = operatorCpuCost + childCpuCosts.sum
+    val totalGpuCost = operatorGpuCost + childGpuCosts.sum
```
revans2, line 46: My main concern here is that this looks like it is written really only for AQE, but that is not documented anywhere.

The total cost looks like the total cost for the entire query (which would produce incorrect results without AQE splitting the query up into smaller parts)

The ShuffleExchangeExec being an explicit cutoff point, and if the cost is not better for the GPU compared to the CPU then everything in the entire query (or at least the part that we are looking at for AQE) will then be put onto the CPU. I'm not sure that is what we want, especially in the context of some parts of the plan having to be on the CPU for what ever reason.

andygrove: I've removed the shuffle-specific logic for now and will add it back once I'm farther along with testing that part.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -0,0 +1,139 @@
...
+  /**
+   * Walk the plan and determine CPU and GPU costs for each operator and then make decisions
+   * about whether operators should run on CPU or GPU.
+   */
+  def optimize(plan: SparkPlanMeta[_]): (Double, Double) = {
+
+    // get the CPU and GPU cost of the child plan(s)
+    val childCosts = plan.childPlans.map(optimize)
+    val childCpuCosts = childCosts.map(_._1)
+    val childGpuCosts = childCosts.map(_._2)
```
gerashegalov, line 38:
```suggestion
    val (childCpuCosts, childGpuCosts) = plan.childPlans.map(optimize).unzip
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
gerashegalov: Do we have a notion for situations as follows:  an operator's 
- GPU cost is say 1 sec, 
- CPU cost is 10 sec 

whereas the wait time for GPU is 20sec, and so to reduce the user-perceived latency we should still utilize CPU?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
andygrove: Thanks @gerashegalov I didn't know about `unzip`!

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
andygrove: > Do we have a notion for situations as follows: an operator's
> 
>     * GPU cost is say 1 sec,
> 
>     * CPU cost is 10 sec
> 
> 
> whereas the wait time for GPU is 20sec, and so to reduce the user-perceived latency we should still utilize CPU?

No, nothing that advanced yet. I imagine we would do this as a follow on issue once we have a basic mechanism in place.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -0,0 +1,185 @@
...
+    var totalGpuCost = operatorGpuCost + childGpuCosts.sum
+
+    // determine how many transitions between CPU and GPU are taking place
+    val numTransitions = plan.childPlans
+      .count(_.canThisBeReplaced != plan.canThisBeReplaced)
+
+    if (numTransitions > 0) {
+      if (plan.canThisBeReplaced) {
+        // transition from CPU to GPU
+        val transitionCost = numTransitions * costModel.transitionToGpuCost(plan)
```
revans2, line 53: I think we need to be clearer on what `transitionToGpuCost` is actually doing.  Right now it is mostly a place holder, but in the future it could look at the schema to determine the cost of transitioning.  We might also want to guess at the size of the data that will transfer. To me it feels like we would want to do something like.

```
plan.childPlans.filter(_.canThisBeReplaced != plan.canThisBeReplaces).map(costModel.transitionToGpuCost(_)).sum
```

So that the transition cost is looking at the output of the plan, so it knows what to transition instead of some vague thing about what the input might be.  Or not pass in the plan at all if we just want it to be a single value and we can change things when the models get better.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -0,0 +1,185 @@
...
+    } else {
+      (totalCpuCost, totalGpuCost)
+    }
+  }
+
+  def forceOnCpuRecursively(plan: SparkPlanMeta[_], reason: String): Unit = {
+    if (plan.canThisBeReplaced) {
+      plan.willNotWorkOnGpu(reason)
+    }
+    plan.childPlans.foreach(plan => forceOnCpuRecursively(plan, reason))
```
revans2, line 85: We need this to stop before it hits the end or if the last stage should not be on the GPU the entire query will be on the CPU.

andygrove: I'm not sure if I have completely addressed your concern here, but I changed it to only replace sections of the plan if the final operator could have been replaced.

andygrove: I see the limitations of this approach now that I am testing with TPC-DS and  I am rewriting this part. The current approach was too naive and we want to selectively force sections of the plan back onto CPU, not everything below a certain point.

revans2: Right now we put everything back onto the CPU until we hit something that is already on the CPU.  I think that works fine for most of the stupid things we do now (like put something on the GPU just to rearrange columns).  But I don't think it will handle cases where something is more expensive on the GPU than the CPU and we may want to exclude just a part of the plan (instead of all of it).  But I think we can tackle that problem when we see it in practice.

andygrove:
@revans2 @jlowe This PR is still WIP but I have updated the description to show the direction I am going in with the EXPLAIN output and it would be good to get some feedback on this before I go too far.

jlowe:
I like the new explain output, but I'm wondering if it could quickly get very verbose on large plans.  We may want the ability to show the fact that the CBO removed an operation from the GPU but suppress the gory details that come out later.  On a large plan, there could easily be a ton of "after" output by the CBO that will be a chore to line up with the original.  Seems like it would be easy to extend the existing `spark.rapids.sql.explain` config to have another setting value.  We already do ALL, NOT_ON_GPU, maybe some CBO-specific levels?

andygrove:
@revans2 @jlowe This is ready for another review. The main changes since the last reviews are:

- The unit tests register a listener with `GpuOverrides` so that they can verify that the expected optimizations were applied. The optimizations are represented by case classes for the two types of optimization that we currently implement.
- There is a new setting `ALL_CBO` for `spark.rapids.sql.explain` that enables the CBO explain output. I'm not crazy about the name and `ALL` no longer means everything, which is slightly confusing.

revans2:
>     * There is a new setting `ALL_CBO` for `spark.rapids.sql.explain` that enables the CBO explain output. I'm not crazy about the name and `ALL` no longer means everything, which is slightly confusing.

Could we rename things then and have `ALL` be the same as `ALL_CBO` is now and introduce a new one that removes CBO details from `ALL`? Perhaps `ALL_NO_CBO`. If w want to have CBO details with other levels of explain, then we should have a separate config for printing CBO details.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -845,8 +845,9 @@ object RapidsConf {
 
   val EXPLAIN = conf("spark.rapids.sql.explain")
     .doc("Explain why some parts of a query were not placed on a GPU or not. Possible " +
-      "values are ALL: print everything, NONE: print nothing, NOT_ON_GPU: print only parts of " +
-      "a query that did not go on the GPU")
+      "values are ALL_NO_CBO: print everything except cost-based optimizations, ALL: print " +
+        "everything including cost-based optimizations, NONE: print nothing, NOT_ON_GPU: print " +
```
jlowe, line 849: Seeing this I'm wondering whether it's worth it to have something like NOT_ON_GPU_NO_CBO which indicates we may want to control CBO output orthogonally via a separate config as @revans2 mentioned.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -878,7 +879,45 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(true)
 
-  val USE_ARROW_OPT = conf("spark.rapids.arrowCopyOptimizationEnabled")
+  val CBO_ENABLED = conf("spark.rapids.sql.cbo.enabled")
```
jlowe, line 882: Nit: Since eventually these configs will become user-visible I'd like to avoid acronyms or odd terms the user may not be familiar with.  I'd suggest just using "optimizer" instead of "cbo", as the user may not know what "cbo" means and the fact that the optimizer is cost-based is arguably an implementation detail.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -494,6 +519,21 @@ abstract class SparkPlanMeta[INPUT <: SparkPlan](plan: INPUT,
   override val childParts: Seq[PartMeta[_]] = Seq.empty
   override val childDataWriteCmds: Seq[DataWritingCommandMeta[_]] = Seq.empty
 
+  var cpuCost: Double = 0
+  var gpuCost: Double = 0
```
jlowe, line 523: I'm not sure there's a point to having getters and setters if these variables aren't private.  Personally I'm OK if we want to go the simple route and just expose them for now, we can always place them behind accessors later since this isn't part of a public API.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -943,3 +983,43 @@ final class RuleNotFoundExprMeta[INPUT <: Expression](
   override def convertToGpu(): GpuExpression =
     throw new IllegalStateException("Cannot be converted to GPU")
 }
+
+sealed abstract class Optimization
```
jlowe, line 987: This is already a pretty large file, so we may want to move optimizer-related rules/classes to a separate file, especially since it's expected to grow over time.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
andygrove: I've reverted the changes to the `EXPLAIN` config and introduced a separate `spark.rapids.sql.optimizer.explain`. I think this makes more sense for now because all of the optimizer configs are internal so it seems unusual to update the documented `EXPLAIN` config to reference them.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1189,10 +1236,41 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
 
   lazy val getCloudSchemes: Option[Seq[String]] = get(CLOUD_SCHEMES)
 
+  lazy val optimizerEnabled: Boolean = get(OPTIMIZER_ENABLED)
+
+  lazy val optimizerExplain: String = get(OPTIMIZER_EXPLAIN)
+
+  lazy val optimizerDebugEnabled: Boolean = get(OPTIMIZER_DEBUG_ENABLED)
```
jlowe, line 1243: This config appears to be unused?  Also I would expect any debug logging to be done with the `logDebug` method from the `Logging` interface, otherwise we're going to get a plethora of debug logging enable configs for stuff over time.

andygrove: Yes, I was using this at one point. I've removed it now.

jlowe (approved):
I think this is a decent starting point worth checkpointing, but it would be good to hear from @revans2 before merging.

revans2 (approved):
I agree that it is time to checkpoint this.  I think the next steps really need to be towards getting it to the point that we can turn it on by default so we stop doing stupid things (and document it).  After that we can really dig into what it would take to have it make smart choices.

andygrove:
build

andygrove:
build

andygrove:
The tests in this PR depend on `nullableStringsDf` which was recently removed, causing the previous build failure. I have added it back.

andygrove:
@jlowe @revans2 I had to merge latest from branch-0.5 and re-instate a method in `SparkQueryCompareTestSuite` to fix CI.

jlowe (approved):

Issue #1620, opened 2021-01-29
[FEA] Automate integration tests for Arrow HostColumnToGpu functionality


tgravescs:
**Is your feature request related to a problem? Please describe.**
as part of https://github.com/NVIDIA/spark-rapids/issues/1072 we are adding ability to read Arrow data form a datasource v2 and then use cudf optimize copy to copy arrow data to GPU.

there is an integration test for it but it requires an additional jar for the data source v2. We need to see if we can make that jar easily accessible so the test runs automatically.

PR #1622, opened 2021-01-29
Support faster copy for a custom DataSource V2 which supplies Arrow data

tgravescs:
This provides support for more efficiently copying data to the GPU when a datasource V2 source provides the data as an ArrowColumnVector.  The CUDF side of this has already been merged. https://github.com/rapidsai/cudf/pull/7222

This currently only supports primitive types and strings. Decimal types and nested types are not supported. It will fallback to the regular copy code if it sees one of those types not supported.

The integration test require an extra jar which contains a datasource v2 which supplies ArrowColumnVector. I'm looking into pulling that code in and how best to automate those tests, filed https://github.com/NVIDIA/spark-rapids/issues/1620 to track that.  

You will notice I added a new AccessibleArrowColumnVector. This is because the Spark ArrowColumnVector doesn't have the Arrow ValueVector publicly accessible.  I use reflection in here to get a hold of that, but another option is just for user to use AccessibleArrowColumnVector.  I would like peoples feedback on whether to keep that or not?

I had to add a shim layer for the Arrow code because Spark 3.1.0 changed the arrow version and ArrowBuf class is now different.


tgravescs:
build

tgravescs:
build

tgravescs:
upmerging

tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
```diff
@@ -16,16 +16,73 @@
...
 import org.apache.spark.sql.types._
-import org.apache.spark.sql.vectorized.{ColumnarBatch, ColumnVector}
+import org.apache.spark.sql.vectorized.{ArrowColumnVector, ColumnarBatch, ColumnVector}
+import org.apache.spark.sql.vectorized.rapids.AccessibleArrowColumnVector
+
+object HostColumnarToGpu extends Logging {
+
+  // use reflection to get value vector from ArrowColumnVector
+  private def getArrowValueVector(cv: ColumnVector): ValueVector = {
+    val arrowCV = cv.asInstanceOf[ArrowColumnVector]
```
revans2, line 38: Is it worth moving the reflection APIs to lazy values in `HostColumnarToGpu` so we only have to calculate the reflection values once?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
```diff
@@ -16,16 +16,73 @@
...
+    valVecField.setAccessible(true)
+    valVecField.get(accessor).asInstanceOf[ValueVector]
+  }
+
+  def arrowColumnarCopy(
+      cv: ColumnVector,
+      ab: ai.rapids.cudf.ArrowColumnBuilder,
+      nullable: Boolean,
+      rows: Int): Unit = {
+    val valVector = if (cv.isInstanceOf[ArrowColumnVector]) {
```
revans2, line 58: nit: Why not use a match instead of is/else?  It feels more scala like to do it that way.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
```diff
@@ -174,33 +232,53 @@ class HostToGpuCoalesceIterator(iter: Iterator[ColumnarBatch],
...
+    // as we won't actually copy any data and we can't tell what type of data it is without
+    // having a column
+    if (useArrowCopyOpt && batch.numCols() > 0 &&
+      arrowTypesSupported(schema) &&
+      (batch.column(0).isInstanceOf[ArrowColumnVector] ||
+        batch.column(0).isInstanceOf[AccessibleArrowColumnVector])) {
+      logInfo("Using GpuArrowColumnarBatchBuilder")
+      batchBuilder = new GpuColumnVector.GpuArrowColumnarBatchBuilder(schema, batchRowLimit, batch)
+    } else {
+      logInfo("Using GpuColumnarBatchBuilder")
```
revans2, line 272: Should these be debug?

tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
```diff
@@ -16,16 +16,84 @@
...
+      case v: ArrowColumnVector =>
+        try {
+          getArrowValueVector(v)
+        } catch {
+          case e: Exception =>
+            throw new IllegalStateException("Trying to read from a ArrowColumnVector but can't " +
+              "access its Arrow ValueVector", e)
+        }
+      case av: AccessibleArrowColumnVector =>
+        // val arrowVec = av.asInstanceOf[AccessibleArrowColumnVector]
```
revans2, line 78: nit: I think this can be removed.

tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
tgravescs: oops missed that

revans2 (approved):

PR #1634, opened 2021-01-29
Fix databricks build since Arrow code added

tgravescs:
Databricks for some reason isn't picking up the Arrow dependency from Spark so we need to explicitly add it for it build.  After that I tested it and it works fine for reading datasource v2 Arrow data.

I also fixed some tabs being in the file.


tgravescs:
build

shims/spark301db/pom.xml
```diff
@@ -23,7 +23,7 @@
         <groupId>com.nvidia</groupId>
```
jlowe, line 23: 2021 copyrights

tgravescs:
build

jlowe (approved):

Issue #1635, opened 2021-01-29
[FEA] RAPIDS accelerated Java UDF


jlowe:
**Is your feature request related to a problem? Please describe.**
This is closely related to #1594 since they both are implemented in Spark via the `ScalaUDF` Catalyst class.  Java UDFs however use a slightly different registration API and are implemented in terms of Spark's `UDF` interface.  Spark wraps the object call in a lambda method which makes extracting the original user class tricky.

**Describe the solution you'd like**
Ideally the RAPIDS Accelerator plugin would be able to automatically extract the user class from the Java UDF wrapped in a lambda method and detect whether the user's code implements the `RapidsUDF` interface for executing the UDF on the GPU.

PR #1636, opened 2021-01-29
RAPIDS accelerated Spark Scala UDF support

### LINKED ISSUE #1594: [FEA] RAPIDS accelerated ScalaUDF
#1393 added support for users supplying an alternate RAPIDS implementation of a Hive UDF, and that works well for queries that are already in SQL.  However if the query is written against the DataFrame API and using a Scala UDF (seen in the Catalyst plan as `ScalaUDF`) then there isn't an option to provide a RAPIDS alternative implementation in the UDF.  Spark wraps the user's code in at least one layer of lambda functions, so it's tricky to get access to the user's original class that implements their UDF code to see if it implements the `RapidsUDF` interface.

Ideally the RAPIDS Accelerator plugin would be able to automatically identify whether the user code behind a `ScalaUDF` instance implements the `RapidsUDF` interface and therefore has a RAPIDS-accelerated implementation.  Due to the use of wrapping lambdas, we may need to examine the bytecode and 'peel off' the lambda layers (if this is possible).

We could provide a separate method, specific to the RAPIDS Accelerator plugin, where users could register their UDFs, but that has two main drawbacks:
- It requires the users to modify their code to additionally register the UDF
- We'd still have to match up the code found in a `ScalaUDF` instance with the code registered via the separate interface.

jlowe:
Closes #1594.

This implements a GPU version of `ScalaUDF` which is used to track Spark Scala UDFs.  Note that this class is also used for Java UDFs which are NOT supported by this change due to the obscuring of the user's class by a lambda wrapper for that case.  Adding support for Spark Java UDFs is tracked by #1635.

A working example of a RAPIDS accelerated Spark Scala UDF is also provided, which required adding the Scala version to the udf-examples jar (and everywhere it was referenced).  A unit test was added to exercise it.  It was not implemented as a Python test as was done for the Hive UDFs because PySpark does not support registering Scala UDFs (it uses the Java UDF interface instead).

The RAPIDS accelerated UDF documentation has also been updated to reflect the new functionality.

### DIFF
--- a/integration_tests/pom.xml
+++ b/integration_tests/pom.xml
@@ -107,7 +107,7 @@
         </dependency>
         <dependency>
             <groupId>com.nvidia</groupId>
-            <artifactId>rapids-4-spark-udf-examples</artifactId>
+            <artifactId>rapids-4-spark-udf-examples_${scala.binary.version}</artifactId>
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -252,10 +252,10 @@ can throw at it.
 The RAPIDS Accelerator provides the following solutions for running
 user-defined functions on the GPU:
 
-#### RAPIDS-Accelerated UDFs
+#### RAPIDS Accelerated UDFs
 
-UDFs can provide a RAPIDS-accelerated implementation which allows the RAPIDS Accelerator to perform
-the operation on the GPU.  See the [RAPIDS-accelerated UDF documentation](additional-functionality/rapids-udfs.md)
+UDFs can provide a RAPIDS accelerated implementation which allows the RAPIDS Accelerator to perform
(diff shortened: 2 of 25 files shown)


jlowe:
build

integration_tests/pom.xml
```diff
@@ -107,7 +107,7 @@
         </dependency>
         <dependency>
             <groupId>com.nvidia</groupId>
-            <artifactId>rapids-4-spark-udf-examples</artifactId>
+            <artifactId>rapids-4-spark-udf-examples_${scala.binary.version}</artifactId>
```
tgravescs, line 110: did we have an issue to update the rest of IT builds?

tgravescs (approved):

integration_tests/pom.xml
jlowe: I tried to catch all the builds that are in this repo.  Just filed #1640 to track build scripts in other repos.

revans2 (approved):

PR #1637, opened 2021-01-30
Fail DivModLike on zero divisor in ANSI mode

### LINKED ISSUE #1464: [BUG] Respect ANSI mode for division-by-zero error [SPARK-33008]

gerashegalov:
This PR closes #1464 

### DIFF
--- a/integration_tests/src/main/python/arithmetic_ops_test.py
+++ b/integration_tests/src/main/python/arithmetic_ops_test.py
@@ -500,18 +500,6 @@ def test_greatest(data_gen):
             lambda spark : gen_df(spark, gen).select(
                 f.greatest(*command_args)), conf=allow_negative_scale_of_decimal_conf)
 
-@pytest.mark.parametrize('ansi_mode, expr', [
-    pytest.param('ansi', '1/0', marks=pytest.mark.xfail(condition=is_before_spark_310(),
-                                                        reason='https://github.com/apache/spark/pull/29882')),
-    pytest.param('ansi', 'a/0', marks=pytest.mark.xfail(condition=is_before_spark_310(),
-                                                        reason='https://github.com/apache/spark/pull/29882')),
-    pytest.param('ansi', 'a/b', marks=pytest.mark.xfail(condition=is_before_spark_310(),
-                                                        reason='https://github.com/apache/spark/pull/29882')),
-    ('nonAnsi', '1/0'),
-    ('nonAnsi', 'a/0'),
-    ('nonAnsi', 'a/b')
-])
-
 
 def _test_div_by_zero(ansi_mode, expr):
     ansi_conf = {'spark.sql.ansi.enabled': ansi_mode == 'ansi'}
--- a/integration_tests/src/main/python/arithmetic_ops_test.py
+++ b/integration_tests/src/main/python/arithmetic_ops_test.py
@@ -500,24 +500,21 @@ def test_greatest(data_gen):
             lambda spark : gen_df(spark, gen).select(
                 f.greatest(*command_args)), conf=allow_negative_scale_of_decimal_conf)
 
-@pytest.mark.parametrize('ansi_mode', ['nonAnsi', 'ansi'])
-@pytest.mark.parametrize('exp_type', ['bothLiterals', 'justZeroLiteral', 'noLiterals'])
-def test_div_by_zero(ansi_mode, exp_type):
-    if ansi_mode == 'ansi':
-        if is_before_spark_310():
-            pytest.xfail('https://github.com/apache/spark/pull/29882')
-        elif exp_type != 'bothLiterals':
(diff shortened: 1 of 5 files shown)


gerashegalov:
build

gerashegalov:
build

gerashegalov:
do we need bug-compatibility for Spark 3.1-?

tgravescs:
can you clarify, 3.1 has a bug or they fixed something?

jlowe:
What's up with all the commits in this PR?  Did you end up rebasing commits in your branch that are already committed to branch-0.4?

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -500,29 +500,25 @@ def test_greatest(data_gen):
...
-        elif exp_type != 'bothLiterals':
-            pytest.xfail('https://github.com/NVIDIA/spark-rapids/issues/1464')
-
+@pytest.mark.parametrize('ansi_mode, expr', [
+    pytest.param('ansi', '1/0', marks=pytest.mark.xfail(condition=is_before_spark_310(),
+                                                        reason='https://github.com/apache/spark/pull/29882')),
+    pytest.param('ansi', 'a/0', marks=pytest.mark.xfail(condition=is_before_spark_310(),
+                                                        reason='https://github.com/apache/spark/pull/29882')),
+    pytest.param('ansi', 'a/b', marks=pytest.mark.xfail(condition=is_before_spark_310(),
+                                                        reason='https://github.com/NVIDIA/spark-rapids/issues/1464')),
```
jlowe, line 509: If this PR closes #1464 then why are we referencing it in new code here?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -23,6 +23,7 @@ import com.nvidia.spark.rapids.RapidsPluginImplicits._
 import org.apache.spark.sql.catalyst.analysis.{TypeCheckResult, TypeCoercion}
```
jlowe, line 23: 2021 copyrights

jlowe (changes requested):

andygrove:
The `GpuDivModLike` changes LGTM. I will review again once the other issues are resolved.

gerashegalov:
> can you clarify, 3.1 has a bug or they fixed something?

@tgravescs  3.1 fixed the incorrect handling of div-by-zero in SPARK-33008. So the question:
when the next version of the plugin runs with Spark < 3.1, should we have the same pre-fix behavior using NULL ignoring div-by-zero. Then I'd handle my fix via the shim layer.

integration_tests/src/main/python/arithmetic_ops_test.py
gerashegalov: > What's up with all the commits in this PR? Did you end up rebasing commits in your branch that are already committed to branch-0.4?

not knowingly or intentionally, I'll recreate the branch to get rid of it.

gerashegalov:
build

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -500,29 +500,25 @@ def test_greatest(data_gen):
...
-@pytest.mark.parametrize('exp_type', ['bothLiterals', 'justZeroLiteral', 'noLiterals'])
-def test_div_by_zero(ansi_mode, exp_type):
-    if ansi_mode == 'ansi':
-        if is_before_spark_310():
-            pytest.xfail('https://github.com/apache/spark/pull/29882')
-        elif exp_type != 'bothLiterals':
-            pytest.xfail('https://github.com/NVIDIA/spark-rapids/issues/1464')
-
+@pytest.mark.parametrize('ansi_mode, expr', [
+    pytest.param('ansi', '1/0', marks=pytest.mark.xfail(condition=is_before_spark_310(),
```
jlowe, line 504: Nit: Should we move the condition to within the test?  It would be less verbose than repeating it three times here.  If we could annotate the test only once with this condition that would work too.

gerashegalov: I changed this because my first attempt was mixing up reasons when done in the test, and generally did not adhere to https://docs.pytest.org/en/latest/skipping.html#skip-xfail-with-parametrize. But I can split the test cases to simplify.

jlowe (approved):

gerashegalov:
build

andygrove (approved):

tgravescs (approved):

PR #1639, opened 2021-02-01
support GpuScalarSubquery

sperlingxx:
This pull request is to provide GPU override of ScalarSubquery.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -31,6 +31,7 @@ import org.apache.spark.sql.catalyst.plans.physical._
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.connector.read.Scan
 import org.apache.spark.sql.execution._
+import org.apache.spark.sql.execution.ScalarSubquery
 import org.apache.spark.sql.execution.adaptive.{AdaptiveSparkPlanExec, BroadcastQueryStageExec, CustomShuffleReaderExec, ShuffleQueryStageExec}
 import org.apache.spark.sql.execution.aggregate.{HashAggregateExec, SortAggregateExec}
 import org.apache.spark.sql.execution.command.{CreateDataSourceTableAsSelectCommand, DataWritingCommand, DataWritingCommandExec, ExecutedCommandExec}
@@ -2164,13 +2165,25 @@ object GpuOverrides {
         override def convertToGpu(child: Expression): GpuExpression = GpuUnscaledValue(child)
       }),
     expr[MakeDecimal](
-      "Create a Decimal from an unscaled long value form some aggregation optimizations",
+      "Create a Decimal from an unscaled long value for some aggregation optimizations",
       ExprChecks.unaryProject(TypeSig.DECIMAL, TypeSig.DECIMAL,
         TypeSig.LONG, TypeSig.LONG),
       (a, conf, p, r) => new UnaryExprMeta[MakeDecimal](a, conf, p, r) {
         override def convertToGpu(child: Expression): GpuExpression =
           GpuMakeDecimal(child, a.precision, a.scale, a.nullOnOverflow)
-      })
+      }),
+    expr[ScalarSubquery](
+      "Subquery that will return only one row and one column",
+      ExprChecks.projectOnly(
+        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL,
+        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL,
+        Nil, None),
+      (a, conf, p, r) => new ExprMeta[ScalarSubquery](a, conf, p, r) {
+        override def convertToGpu(): GpuExpression = {
+          GpuScalarSubquery(GpuSubqueryExec(a.plan.name, a.plan.child), a.exprId)
+        }
+      }
+    )
   ).map(r => (r.getClassFor.asSubclass(classOf[Expression]), r)).toMap
 
   val expressions: Map[Class[_ <: Expression], ExprRule[_ <: Expression]] =
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -2179,9 +2179,7 @@ object GpuOverrides {
         TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL,
         Nil, None),
       (a, conf, p, r) => new ExprMeta[ScalarSubquery](a, conf, p, r) {
-        override def convertToGpu(): GpuExpression = {
-          GpuScalarSubquery(GpuSubqueryExec(a.plan.name, a.plan.child), a.exprId)
-        }
+        override def convertToGpu(): GpuExpression = GpuScalarSubquery(a.plan, a.exprId)
       }
     )
   ).map(r => (r.getClassFor.asSubclass(classOf[Expression]), r)).toMap
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalarSubquery.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalarSubquery.scala
@@ -35,7 +35,7 @@ case class GpuScalarSubquery(
   extends ExecSubqueryExpression with GpuExpression {
 
   override def dataType: DataType = plan.schema.fields.head.dataType
-  override def children: Seq[Expression] = Nil
+  override def children: Seq[Expression] = Seq.empty
   override def nullable: Boolean = true
   override def toString: String = plan.simpleString(SQLConf.get.maxToStringFields)
   override def withNewPlan(query: BaseSubqueryExec): GpuScalarSubquery = copy(plan = query)
@@ -53,11 +53,10 @@ case class GpuScalarSubquery(
     val rows = plan.executeCollect()
     if (rows.length > 1) {
       sys.error(s"more than one row returned by a subquery used as an expression:\n$plan")
-    }
-    if (rows.length == 1) {
-      assert(rows(0).numFields == 1,
-        s"Expects 1 field, but got ${rows(0).numFields}; something went wrong in analysis")
-      result = rows(0).get(0, dataType)
+    } else if (rows.length == 1) {
+      assert(rows.head.numFields == 1,
+        s"Expects 1 field, but got ${rows.head.numFields}; something went wrong in analysis")
+      result = rows.head.get(0, dataType)
     } else {
       // If there is no rows returned, the result should be null.
       result = null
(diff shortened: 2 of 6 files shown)


sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalarSubquery.scala
```diff
@@ -0,0 +1,88 @@
...
+  override def withNewPlan(query: BaseSubqueryExec): GpuScalarSubquery = copy(plan = query)
+
+  override def semanticEquals(other: Expression): Boolean = other match {
+    case s: GpuScalarSubquery => plan.sameResult(s.plan)
+    case _ => false
+  }
+  private var result: Any = _
+
+  override def updateResult(): Unit = {
+    val subPlanRDD = plan.child.execute()
```
revans2, line 46: Why are we doing this so differently from the Spark version?  It looks like we really wanted to replace the `BaseSubqueryExec`, and to work around replacing it we had to jump through a number of hoops making the code diverge from the original Spark code.  I really would prefer to just do it the same way that Spark did. I don't know 100% what types of `BaseSubqueryExec` can be passed into this, but if it ends up being a `SubqueryBroadcastExec` or a `ReusedSubqueryExec` this code will fail.

revans2 (changes requested):

sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalarSubquery.scala
sperlingxx: Yes, I checked catalyst codes and found that `ReusedSubqueryExec`  may also be passed into `ScalarSubquery`. The intention of replacing original Spark code is to avoid `executeCollect` which seems inefficient. I realized this replacement seems unnecessary. So, I rewrote `updateResult` with the exact Spark code.

revans2 (approved):
This looks good.  It  would be nice to see some integration tests with it too, but that is not critical.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2217,13 +2218,23 @@ object GpuOverrides {
...
           GpuMakeDecimal(child, a.precision, a.scale, a.nullOnOverflow)
       }),
+    expr[ScalarSubquery](
+      "Subquery that will return only one row and one column",
+      ExprChecks.projectOnly(
+        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL,
+        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL,
+        Nil, None),
+      (a, conf, p, r) => new ExprMeta[ScalarSubquery](a, conf, p, r) {
+        override def convertToGpu(): GpuExpression = GpuScalarSubquery(a.plan, a.exprId)
```
revans2, line 2235: Sorry I forgot to ask how does `a.plan` get converted to the GPU? Do your test show that it ends up on the GPU?

tests/src/test/scala/com/nvidia/spark/rapids/ScalarSubquerySuite.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.rapids.GpuScalarSubquery
+
+class ScalarSubquerySuite extends SparkQueryCompareTestSuite {
+
+  private def checkExecPlan(plan: SparkPlan): Unit = {
+    if (!plan.conf.getAllConfs(RapidsConf.SQL_ENABLED.key).toBoolean) return
+    plan.find(_.expressions.exists(e => e.find(_.isInstanceOf[GpuScalarSubquery]).nonEmpty)) match {
+      case Some(plan) =>
+        val subqueryExec = plan.expressions.collectFirst {
```
gerashegalov, line 28: subqueryExec is not used

gerashegalov: It seems all we do in this test boils down to an assert like this
```scala
    val subqueryExec = plan
      .map(_.expressions)
      .collectFirst { case gpuScalarSubquery: GpuScalarSubquery => gpuScalarSubquery }
      .map(_.plan)

  require(subqueryExec.isDefined, "Could not find GpuScalarSubquery expression among plans")

```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: `a.plan` here is `BaseSubqueryExec(child: subqueryExec)`. 
I think the child plan will go through GpuOverriding rules as an [independent physical plan](https://github.com/apache/spark/blob/branch-3.1/sql/core/src/main/scala/org/apache/spark/sql/execution/subquery.scala#L178) before contained with `BaseSubqueryExec`.
And the child is not necessarily on the GPU if I am not wrong. I think the avdantage of `GpuScalarSubquery` is making the outside plan (usually FilterExec) who owns scalar_subquery expression GPU-enabled if there is no other fallback reason.

sperlingxx:
> This looks good. It would be nice to see some integration tests with it too, but that is not critical.

I added a simple integration test for this.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalarSubquery.scala
```diff
@@ -0,0 +1,72 @@
...
+ * This placeholder is to make ScalarSubquery working as a GPUExpression to cooperate
+ * other GPU overrides.
+ */
+case class GpuScalarSubquery(
+    plan: BaseSubqueryExec,
+    exprId: ExprId)
+  extends ExecSubqueryExpression with GpuExpression {
+
+  override def dataType: DataType = plan.schema.fields.head.dataType
+  override def children: Seq[Expression] = Nil
```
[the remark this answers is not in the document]
sperlingxx: I changed it. Thanks for suggestion.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalarSubquery.scala
```diff
@@ -0,0 +1,72 @@
...
+  @volatile private var result: Any = _
+  @volatile private var updated: Boolean = false
+
+  override def updateResult(): Unit = {
+    val rows = plan.executeCollect()
+    if (rows.length > 1) {
+      sys.error(s"more than one row returned by a subquery used as an expression:\n$plan")
+    }
+    if (rows.length == 1) {
+      assert(rows(0).numFields == 1,
```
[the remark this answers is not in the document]
sperlingxx: "do we need this to be an elidable assert?" We just follow the `ScalarSubquery` of Spark. Perhaps multiple columns is not a fatal exception compared to multiple rows.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalarSubquery.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  // the first column in first row from `query`.
+  @volatile private var result: Any = _
+  @volatile private var updated: Boolean = false
+
+  override def updateResult(): Unit = {
+    val rows = plan.executeCollect()
+    if (rows.length > 1) {
+      sys.error(s"more than one row returned by a subquery used as an expression:\n$plan")
+    }
```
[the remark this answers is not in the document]
sperlingxx: I changed it. Thanks for suggestion.

tests/src/test/scala/com/nvidia/spark/rapids/ScalarSubquerySuite.scala
sperlingxx: I have refined this part. I didn't realize that the variable `subqueryExec` is no longer used.

sperlingxx:
build

revans2 (approved):
Looks great

revans2:
@gerashegalov are you OK with the changes?

tests/src/test/scala/com/nvidia/spark/rapids/ScalarSubquerySuite.scala
```diff
@@ -0,0 +1,42 @@
...
+
+class ScalarSubquerySuite extends SparkQueryCompareTestSuite {
+
+  private def checkExecPlan(plan: SparkPlan): Unit = {
+    if (!plan.conf.getAllConfs(RapidsConf.SQL_ENABLED.key).toBoolean) return
+    plan.find(_.expressions.exists(e => e.find(_.isInstanceOf[ScalarSubquery]).nonEmpty)) match {
+      case Some(plan) =>
+        throw new AssertionError(s"Assume no (cpu)ScalarSubquery, but found in $plan")
+      case None =>
+    }
```
gerashegalov, line 29: I think we should still assert what ScalarSubquery was replaced with to convey the intention. 

```suggestion
   val rapidSqlEnabled = plan.conf.getAllConfs(RapidsConf.SQL_ENABLED.key).toBoolean

    val cpuScalar = plan.find(_.expressions.exists(
      _.collectFirst { case x: ScalarSubquery => x }.nonEmpty
    ))

    val gpuScalar = plan.find(_.expressions.exists(
      _.collectFirst { case x: GpuScalarSubquery => x }.nonEmpty
    ))

    assert(rapidSqlEnabled && cpuScalar.isEmpty && gpuScalar.nonEmpty
      || !rapidSqlEnabled && cpuScalar.nonEmpty && gpuScalar.isEmpty,
      s"Unexpected values $rapidSqlEnabled $cpuScalar $gpuScalar")
```

tests/src/test/scala/com/nvidia/spark/rapids/ScalarSubquerySuite.scala
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package com.nvidia.spark.rapids
+
+import org.apache.spark.sql.execution.{ScalarSubquery, SparkPlan}
+
+class ScalarSubquerySuite extends SparkQueryCompareTestSuite {
+
+  private def checkExecPlan(plan: SparkPlan): Unit = {
+    if (!plan.conf.getAllConfs(RapidsConf.SQL_ENABLED.key).toBoolean) return
```
gerashegalov, line 24: nit: we can assert the behavior when rapids is disabled

tests/src/test/scala/com/nvidia/spark/rapids/ScalarSubquerySuite.scala
```diff
@@ -0,0 +1,42 @@
...
+    plan.find(_.expressions.exists(e => e.find(_.isInstanceOf[ScalarSubquery]).nonEmpty)) match {
```
gerashegalov, line 25: nit: instanceOf asInstanceOf are an anti-pattern , pattern matching is preferable.

gerashegalov:
> @gerashegalov are you OK with the changes?

It's fine, I am just suggesting to make the test more crisp, but not critical.

gerashegalov (approved):
LGTM, some optional suggestions

Issue #1644, opened 2021-02-01
[BUG] test_window_aggregate_udf_array_from_python fails on databricks


tgravescs:
**Describe the bug**
we started to run more tests on databricks and we found that the udf_test.py test_window_aggregate_udf_array_from_python is failing because the script is running with enabling assertions.  Its a bit weird it worked without the assertions enabled as I would have thought the output would be different but maybe something is allow to work.  We need to investigate what is going on. Seems like a schema not matching column types issue:
```
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[No_Partition-Byte][IGNORE_ORDER]
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[No_Partition-Short][IGNORE_ORDER]
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[No_Partition-Integer][IGNORE_ORDER]
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded-Byte][IGNORE_ORDER]
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded-Short][IGNORE_ORDER]
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded-Integer][IGNORE_ORDER]
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded_Following-Byte][IGNORE_ORDER]
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded_Following-Short][IGNORE_ORDER]
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded_Following-Integer][IGNORE_ORDER]
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded_Preceding-Byte][IGNORE_ORDER]
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded_Preceding-Short][IGNORE_ORDER]
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Unbounded_Preceding-Integer][IGNORE_ORDER]
FAILED src/main/python/udf_test.py::test_window_aggregate_udf_array_from_python[Lower_Upper-Byte][IGNORE_ORDER]
[...]

PR #1549, opened 2021-01-19
Run Databricks IT with python-xdist parallel

NvTimLiu:
Enable the databricks IT tests to run in parallel, issue: https://github.com/NVIDIA/spark-rapids/issues/1499

Signed-off-by: Tim Liu <timl@nvidia.com>

PR #1645, opened 2021-02-01
Run Databricks IT with python-xdist parallel, includes test fixes and xfail

tgravescs:
We want databricks to run all the tests and run them in parallel.  When enabling that a few were failing.

tpch - we can't run with AQE due to known databricks issues with it. After that there were a few operators running on CPU due to differences in databricks so added code to handle that
test_window_aggregate_udf_array_from_python failing - xfail it for now and filed https://github.com/NVIDIA/spark-rapids/issues/1644 to root cause it.

Note this was started by Tim with https://github.com/NVIDIA/spark-rapids/pull/1549/files

### DIFF
--- a/integration_tests/pytest.ini
+++ b/integration_tests/pytest.ini
@@ -14,6 +14,7 @@
 
 [pytest]
 markers =
+    allow_non_gpu_databricks(any, [ops]): Allow the test to run anything not on the GPU in Databricks environment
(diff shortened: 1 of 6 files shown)


tgravescs:
build

tgravescs:
@NvTimLiu

jlowe (approved):

PR #1652, opened 2021-02-02
Fix typo in arrow optimized config name - spark.rapids.arrowCopyOptimizationEnabled

tgravescs:
have a typo in the config name that should be spark.rapids.arrowCopyOptimizationEnabled

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
@@ -832,7 +832,7 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(true)
 
-  val USE_ARROW_OPT = conf("spark.rapids.arrowCopyOptmizationEnabled")
+  val USE_ARROW_OPT = conf("spark.rapids.arrowCopyOptimizationEnabled")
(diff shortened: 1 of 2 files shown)


tgravescs:
build

andygrove (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -832,7 +832,7 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(true)
 
-  val USE_ARROW_OPT = conf("spark.rapids.arrowCopyOptmizationEnabled")
+  val USE_ARROW_OPT = conf("spark.rapids.arrowCopyOptimizationEnabled")
```
jlowe, line 835: Shouldn't there be a corresponding `configs.md` update that corresponds to this change?

tgravescs: its an internal config  - mostly a test config to easily switch implementations, but I also figured it was a backup if something was going wrong could turn off the arrow stuff.

jlowe: ah right, I missed this was marked internal.

jlowe (approved):

Issue #1650, opened 2021-02-02
[BUG] BenchmarkRunner does not include query name in JSON summary filename when running multiple queries


andygrove:
**Describe the bug**
When running BenchmarkRunner with multiple queries, it is not possible to specify a summary-file-prefix that contains the individual query names, making it time-consuming to differentiate between the generated files.

It would be good to have the ability to provide a file prefix that includes a variable for the query name. For example: `--summary-file-prefix my-${QUERY}-tests`.

PR #1663, opened 2021-02-03
BenchmarkRunner - Include query name in JSON summary filename

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/1650. 

### DIFF
--- a/integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/BenchmarkRunner.scala
+++ b/integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/BenchmarkRunner.scala
@@ -113,7 +113,7 @@ object BenchmarkRunner {
                 spark,
                 query,
                 conf.iterations(),
-                summaryFilePrefix = summaryFilePrefixWithQuery,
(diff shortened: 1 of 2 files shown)


integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/BenchmarkRunner.scala
```diff
@@ -78,6 +78,10 @@ object BenchmarkRunner {
         conf.query().foreach { query =>
 
           println(s"*** RUNNING ${bench.name()} QUERY $query")
+          val summaryFilePrefixWithQuery = conf.summaryFilePrefix.toOption match {
+            case Some(t) => Some(t + s"-$query")
+            case _ => None
+          }
```
andygrove, line 84: You can use a `map` operation here. Something like:

```scala
val summaryFilePrefixWithQuery = conf.summaryFilePrefix.map(prefix => s"$prefix-$query")
```

andygrove (approved):
I tested this locally and LGTM

andygrove:
build

PR #1664, opened 2021-02-03
Filter nulls for left semi and left anti join to work around cudf

### LINKED ISSUE #1643: [BUG] TPC-DS-Like q10, q35, and q69 - slow or hanging at leftSemiJoin
This PR (https://github.com/NVIDIA/spark-rapids/pull/1391) removed some filtering in the plugin to work around slowness in cudf while building/probing the hash table used in cudf (https://github.com/rapidsai/cudf/issues/6052). It looks like `leftSemiJoin` is following a different path and needs to be addressed separately. As of now branch-0.4 regressed from branch-0.3 since we removed the filter logic and `leftSemiJoin` appears to still need it.

The symptom is a CPU thread stuck at:

```
Executor task launch worker for task 12689	RUNNABLE	Lock(java.util.concurrent.ThreadPoolExecutor$Worker@2076710178})
ai.rapids.cudf.Table.leftSemiJoin(Native Method)
ai.rapids.cudf.Table.access$3700(Table.java:45)
ai.rapids.cudf.Table$TableOperation.leftSemiJoin(Table.java:2401)
org.apache.spark.sql.rapids.execution.GpuHashJoin.doJoinLeftRight(GpuHashJoin.scala:274)
```

This ticket is to track the branch-0.4 fix. We can either (hopefully) fix it in cudf-0.18, or add the workaround again just for `leftSemiJoin`. Need to figure out if other joins are also affected. 

@hyperbolic2346 FYI

abellina:
performance issues

Closes #1643

This is only targetting `leftSemi` and `leftAnti` because these are based on `cudf::left_semi_anti_join`, and this path was not sped up in 0.18 (specifically building the hash table for the join is slow with many nulls). 

This passes the tests locally, and I fixed some leaks I had but I need to run performance tests at scale, so I am posting it as draft for now.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+ * Copyright (c) 2020-2021, NVIDIA CORPORATION. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -268,6 +268,9 @@ trait GpuHashJoin extends GpuExec {
     }
   }
 
+  // This is a work around added in response to https://github.com/NVIDIA/spark-rapids/issues/1643.
+  // to deal with slowness arising from many nulls in the build-side of the join. The work around
+  // should be removed when https://github.com/rapidsai/cudf/issues/7300 is addressed.
   private[this] def filterNulls(table: Table, joinKeyIndices: Range, closeTable: Boolean): Table = {
     var mask: ai.rapids.cudf.ColumnVector = null
     try {
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
@@ -16,14 +16,13 @@
 package org.apache.spark.sql.rapids.execution
 
 import ai.rapids.cudf.{NvtxColor, Table}
-import com.nvidia.spark.rapids.{CoalesceGoal, GpuBindReferences, GpuBoundReference, GpuBuildLeft, GpuBuildRight, GpuBuildSide, GpuColumnVector, GpuExec, GpuExpression, GpuFilter, GpuIsNotNull, GpuProjectExec, NvtxWithMetrics, RapidsMeta, RequireSingleBatch}
+import com.nvidia.spark.rapids.{CoalesceGoal, GpuBindReferences, GpuBuildLeft, GpuBuildRight, GpuBuildSide, GpuColumnVector, GpuExec, GpuExpression, GpuFilter, GpuIsNotNull, GpuProjectExec, NvtxWithMetrics, RapidsMeta, RequireSingleBatch}
 
 import org.apache.spark.TaskContext
 import org.apache.spark.sql.catalyst.expressions.{Attribute, Expression}
 import org.apache.spark.sql.catalyst.plans.{ExistenceJoin, FullOuter, InnerLike, JoinType, LeftAnti, LeftExistence, LeftOuter, LeftSemi, RightOuter}
 import org.apache.spark.sql.execution.SparkPlan
 import org.apache.spark.sql.execution.metric.SQLMetric
-import org.apache.spark.sql.rapids.GpuAnd
 import org.apache.spark.sql.vectorized.{ColumnarBatch, ColumnVector}
 
 object GpuHashJoin {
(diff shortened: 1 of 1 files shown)


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -16,14 +16,13 @@
 package org.apache.spark.sql.rapids.execution
```
jlowe, line 16: 2021 copyrights

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -262,23 +268,76 @@ trait GpuHashJoin extends GpuExec {
...
+  private[this] def filterNulls(table: Table, joinKeyIndices: Range, closeTable: Boolean): Table = {
+    var mask: ai.rapids.cudf.ColumnVector = null
+    try {
+      joinKeyIndices.indices.foreach { c =>
+        mask = withResource(table.getColumn(c).isNotNull) { nn =>
+          if (mask == null) {
+            nn.incRefCount()
+          } else {
+            withResource(mask) { _ =>
+              mask.and(nn)
```
jlowe, line 280: I think this needs to be a bitwise-OR here instead of bitwise-AND.  If any, not all, of the join key columns are null it shouldn't match.  It will still work with bitwise-AND but it can still lead to long collision chains on null columns and poor performance.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -262,23 +268,76 @@ trait GpuHashJoin extends GpuExec {
...
-          .innerJoin(rightTable.onColumns(joinKeyIndices: _*), false)
-      case LeftSemi => leftTable.onColumns(joinKeyIndices: _*)
-          .leftSemiJoin(rightTable.onColumns(joinKeyIndices: _*), false)
-      case LeftAnti => leftTable.onColumns(joinKeyIndices: _*)
-          .leftAntiJoin(rightTable.onColumns(joinKeyIndices: _*), false)
-      case FullOuter => leftTable.onColumns(joinKeyIndices: _*)
-          .fullJoin(rightTable.onColumns(joinKeyIndices: _*), false)
-      case _ => throw new NotImplementedError(s"Joint Type ${joinType.getClass} is not currently" +
-          s" supported")
+  private[this] def filterNulls(table: Table, joinKeyIndices: Range, closeTable: Boolean): Table = {
```
jlowe, line 271: It would be good to add a comment explaining what this does, why it's here, and link to the cudf issue that will allow it to be removed.

jlowe (changes requested):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
jlowe: Oh nevermind, this is the mask of values to preserve not to filter.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -232,16 +231,23 @@ trait GpuHashJoin extends GpuExec {
...
-        case GpuBuildRight => doJoinLeftRight(streamedTable, builtTable)
+    val joined =
+      withResource(new NvtxWithMetrics("hash join", NvtxColor.ORANGE, joinTime)) { _ =>
+        // `doJoinLeftRight` closes the right table if the last argument (`closeRightTable`)
+        // is true, but never closes the left table.
+        buildSide match {
+          case GpuBuildLeft =>
+            // tell `doJoinLeftRight` it is ok to close the `streamedTable`, this can help
+            // in order to close temporary/intermediary data after a filter in some scenarios.
+            doJoinLeftRight(builtTable, streamedTable, true)
```
jlowe, line 242: I'm not sure I understand the point of the close boolean argument.  It looks like the table, if specified to close, is always closed after the operation has fully completed, so why not have the caller perform the close?  It would be simpler to read and understand who is responsible for closing, and we could just wrap this in a `withResource(streamedTable)` to make sure streamedTable is always closed as it was in the original code.

abellina: >  if specified to close, is always closed after the operation has fully completed
Not always. `streamedTable` may be closed by the filter logic (https://github.com/NVIDIA/spark-rapids/pull/1664/files/358f81a81e0c5bab1df51f63738cd864faefe229#diff-753a0d906f1f2d01d8bb08ef13e5205b1710cf915bb0f5fcdd55c3b665d2900dR293). Otherwise, by the time the join executes, we have `streamedTable`, and a filtered version of it + the join. It was an attempt at not having as much in memory around.

We can do it the other way with the `withResource` block here in `doJoin` also. Let me know if the above makes more sense or not.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
abellina: sure thing, will add.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
jlowe: yeah, the desire to free this early is what leads to the messiness.  I'm fine with it since it shouldn't be here long. :smile:

jlowe (approved):

abellina:
build

revans2 (approved):

PR #1671, opened 2021-02-04
Allow metrics to be configurable by level

### LINKED ISSUE #1626: [BUG] TPC-DS-like query 77 at scale=3TB fails with maxResultSize exceeded error
Running q77 at 3TB fails with:
```
21/01/29 16:55:30 ERROR TaskSetManager: Total size of serialized results of 24771 tasks (1053.8 MiB) is bigger than spark.driver.maxResultSize (1024.0 MiB)
```
This issue is likely related to: https://github.com/NVIDIA/spark-rapids/issues/1284. But this is being run with the released 0.3 version of the plugin. 

Upping the `spark.driver.maxResultSize` does help, but we still seem to see 20MB of output per task, which is adding up quickly.

revans2:
This fixes #1626 

The goal of this is to reduce the number of metrics that we report by default and also to allow the user to disable most metrics if they hit a situation like the one in the issue being fixed.

I realize that this is a huge change, in terms of the files and lines that it touches.


revans2:
build

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuBroadcastHashJoinExec.scala
```diff
@@ -96,12 +93,15 @@ case class GpuBroadcastHashJoinExec(
...
     right: SparkPlan) extends BinaryExecNode with GpuHashJoin {
+  import GpuMetric._
 
-  override lazy val additionalMetrics: Map[String, SQLMetric] = Map(
-    "joinOutputRows" -> SQLMetrics.createMetric(sparkContext, "join output rows"),
-    "streamTime" -> SQLMetrics.createNanoTimingMetric(sparkContext, "stream time"),
-    "joinTime" -> SQLMetrics.createNanoTimingMetric(sparkContext, "join time"),
-    "filterTime" -> SQLMetrics.createNanoTimingMetric(sparkContext, "filter time"))
+  override lazy val outputRowsLevel: MetricsLevel = ESSENTIAL_LEVEL
+  override lazy val outputBatchesLevel: MetricsLevel = MODERATE_LEVEL
```
jlowe, line 99: Curious, why are these `lazy`?  Comment applies to many other places where this was copied.

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/GpuBroadcastHashJoinExec.scala
```diff
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *   http://www.apache.org/licenses/LICENSE-2.0
```
jlowe, line 8: Intentional change?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala
```diff
@@ -47,15 +67,75 @@ object GpuMetricNames {
...
+  val DESCRIPTION_SORT_TIME = "sort time"
+  val DESCRIPTION_AGG_TIME = "aggregation time"
+  val DESCRIPTION_JOIN_TIME = "join time"
+  val DESCRIPTION_JOIN_OUTPUT_ROWS = "join output rows"
+  val DESCRIPTION_FILTER_TIME = "filter time"
+  val DESCRIPTION_BUILD_DATA_SIZE = "build side size"
+  val DESCRIPTION_BUILD_TIME = "build time"
+  val DESCRIPTION_STREAM_TIME = "stream time"
+
+  // TODO these end up being ESSENTIAL_LEVEL do we really want that?
```
jlowe, line 81: Part of me thinks no GPU metric should be essential, only the Spark built-ins are, but I can see the argument that some GPU-specific metrics on some execs could be considered essential.  Curious what others think.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/NvtxWithMetrics.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, NVIDIA CORPORATION.
+ * Copyright (c) 2019,2021, NVIDIA CORPORATION.
```
jlowe, line 2: This file was changed in 2020, copyright should have been updated.
```suggestion
 * Copyright (c) 2019-2021 NVIDIA CORPORATION.
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -410,6 +429,21 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(false)
 
+  // METRICS
+
+  val METRICS_LEVEL = conf("spark.rapids.sql.metrics.level")
+      .doc("GPU plans can produce a lot more metrics than CPU plans do. In very large " +
+          "queries this can sometimes result in going over the max result size limit for the " +
+          "driver. Supported values include DEBUG: which will enable all metrics supported and " +
```
jlowe, line 437: Nit: Suggest removing the `:` from the values since it's not immediately clear whether it's part of the value.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -410,6 +429,21 @@ object RapidsConf {
...
+          "typically only needs to be enabled when debugging the plugin. MODERATE: which is the " +
+          "default value and should output enough metrics to understand how long each part of " +
```
jlowe, line 439: Nit: Don't think it's necessary to mention the default value since it appears both here and in the generated documentation next to this text.

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuBroadcastHashJoinExec.scala
```diff
@@ -1,11 +1,11 @@
 /*
- * Copyright (c) 2019-2020, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2021, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *   http://www.apache.org/licenses/LICENSE-2.0
```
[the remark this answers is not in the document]
revans2: I did it to match spark301 and spark301db versions of the same file. I picked wrong on which version to fix, so I will switch it.

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuBroadcastHashJoinExec.scala
revans2: I made it lazy so on the chance that we might want to calculate it later. If you want me to switch it over to just be a val I can.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala
revans2: I traced through a number of places for the metrics to see if they matched the Spark versions, and these I didn't find the time to do yet.  I'll look at them and see if I need to refactor it.

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuBroadcastHashJoinExec.scala
jlowe: Yeah these should all not be lazy.  If some exec in the future wants to calculate this dynamically then it can do a lazy override, but no need to apply this to all cases.  I'm also not sure why we would want a metric to sometimes be essential vs. moderate etc.

revans2: You cannot do a lazy override to a non-lazy value.  They have to match, but we can cross that bridge when we come to it.

revans2:
build

revans2:
@jlowe I think I have addressed all of your review comments. I also updated some of the code that made some metrics optional with nulls and instead turned them into NoopMetric instances.

jlowe (approved):

shims/spark301db/src/main/scala/com/nvidia/spark/rapids/shims/spark301db/GpuShuffledHashJoinExec.scala
```diff
@@ -85,13 +82,15 @@ case class GpuShuffledHashJoinExec(
...
-    "streamTime" -> SQLMetrics.createNanoTimingMetric(sparkContext, "stream time"),
-    "joinTime" -> SQLMetrics.createNanoTimingMetric(sparkContext, "join time"),
-    "joinOutputRows" -> SQLMetrics.createMetric(sparkContext, "join output rows"),
-    "filterTime" -> SQLMetrics.createNanoTimingMetric(sparkContext, "filter time"))
+  override val outputRowsLevel: MetricsLevel = ESSENTIAL_LEVEL
+  override val outputBatchesLevel: MetricsLevel = MODERATE_LEVEL
+  override lazy val additionalMetrics: Map[String, GpuMetric] = Map(
+    BUILD_DATA_SIZE -> createSizeMetric(ESSENTIAL_LEVEL, DESCRIPTION_BUILD_DATA_SIZE),
+    BUILD_TIME -> createNanoTimingMetric(ESSENTIAL_LEVEL, DESCRIPTION_BUILD_TIME),
+    STREAM_TIME -> createNanoTimingMetric(MODERATE_LEVEL, STREAM_TIME),
```
abellina, line 90: should this be `DESCRIPTION_STREAM_TIME` on the right?

abellina: same for `DESCRIPTION_JOIN_TIME` below?

abellina (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
```diff
@@ -387,14 +384,14 @@ case class HostColumnarToGpu(child: SparkPlan, goal: CoalesceGoal)
...
-    val numOutputBatches = longMetric(NUM_OUTPUT_BATCHES)
-    val collectTime = longMetric("collectTime")
-    val concatTime = longMetric("concatTime")
-    val totalTime = longMetric(TOTAL_TIME)
-    val peakDevMemory = longMetric("peakDevMemory")
+    val numInputRows = gpuLongMetric(NUM_INPUT_ROWS)
+    val numInputBatches = gpuLongMetric(NUM_INPUT_BATCHES)
+    val numOutputRows = gpuLongMetric(NUM_OUTPUT_ROWS)
+    val numOutputBatches = gpuLongMetric(NUM_OUTPUT_BATCHES)
+    val collectTime = gpuLongMetric("collectTime")
```
abellina, line 391: can we use `COLLECT_TIME` and `CONCAT_TIME` here?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileSourceScanExec.scala
```diff
@@ -342,36 +343,39 @@ case class GpuFileSourceScanExec(
...
-  } ++ staticMetrics ++ GpuMetricNames.buildGpuScanMetrics(sparkContext)
+  } ++ staticMetrics
 
   override protected def doExecute(): RDD[InternalRow] =
     throw new IllegalStateException(s"Row-based execution should not occur for $this")
 
   override protected def doExecuteColumnar(): RDD[ColumnarBatch] = {
-    val numOutputRows = longMetric("numOutputRows")
-    val scanTime = longMetric("scanTime")
+    val numOutputRows = gpuLongMetric("numOutputRows")
```
abellina, line 377: "numOutputRows" => `NUM_OUTPUT_ROWS`

revans2:
build

revans2:
Thanks for the review @abellina I think I have addressed all of your concerns

jlowe (approved):

abellina (approved):

PR #1676, opened 2021-02-05
BenchmarkRunner option to generate query plan diagrams in DOT format

andygrove:

This PR adds a `--generate-dot` flag to the benchmark runner so that it will generate query plan diagrams in DOT format. The generator is also updated to color-code operators as green or blue depending on whether they are on GPU or CPU, with transitions highlighted in red. 

The chart is also now inverted with the final operator at the bottom, to match Spark UI.

colorful2

### DIFF
--- a/integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/BenchmarkRunner.scala
+++ b/integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/BenchmarkRunner.scala
@@ -89,7 +89,8 @@ object BenchmarkRunner {
                   path,
                   iterations = conf.iterations(),
                   summaryFilePrefix = summaryFilePrefixWithQuery,
-                  gcBetweenRuns = conf.gcBetweenRuns())
+                  gcBetweenRuns = conf.gcBetweenRuns(),
+                  generateDotGraph = conf.generateDot())
               case "csv" =>
                 runner.writeCsv(
                   spark,
@@ -97,7 +98,8 @@ object BenchmarkRunner {
                   path,
                   iterations = conf.iterations(),
                   summaryFilePrefix = summaryFilePrefixWithQuery,
-                  gcBetweenRuns = conf.gcBetweenRuns())
+                  gcBetweenRuns = conf.gcBetweenRuns(),
+                  generateDotGraph = conf.generateDot())
               case "orc" =>
                 runner.writeOrc(
                   spark,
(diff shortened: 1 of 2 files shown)


integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/BenchmarkRunner.scala
```diff
@@ -89,23 +89,26 @@ object BenchmarkRunner {
                   path,
                   iterations = conf.iterations(),
                   summaryFilePrefix = summaryFilePrefixWithQuery,
-                  gcBetweenRuns = conf.gcBetweenRuns())
+                  gcBetweenRuns = conf.gcBetweenRuns(),
+                  generateDotGraph = conf.generateDot())
```
jlowe, line 93: Not a change that needs to be in this PR, but the number of options is getting to the point where we should consider using a builder pattern, building an options object and passing that.

integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/common/BenchUtils.scala
```diff
@@ -448,8 +477,17 @@ object BenchUtils {
...
+
+          (isGpuPlan(a), isGpuPlan(a.children(i))) match {
+            case (true, true) =>
+              w.println(s"""node$childId -> node$id [color="$nvGreen"];""")
+            case (false, false) =>
+              w.println(s"""node$childId -> node$id [color="$blue"];""")
+            case _ =>
+              // show emphasis on transitions between CPU and GPU
+              w.println(s"node$childId -> node$id [color=red, style=bold];")
+          }
```
jlowe, line 489:
```suggestion
          val style = (isGpuPlan(a), isGpuPlan(a.children(i))) match {
            case (true, true) => s"color=\"$nvGreen\""
            case (false, false) => s"color=\"$blue\""
            case _ =>
              // show emphasis on transitions between CPU and GPU
              "color=red, style=bold"
           }
           w.println(s"node$childId -> node$id [$style];")
```

jlowe (dismissed):
Minor comments that aren't mustfix, lgtm.

nartal1 (dismissed):

andygrove:
build

jlowe (dismissed):

nartal1 (dismissed):

andygrove:
build

andygrove:
build

jlowe (approved):

Issue #1141, opened 2020-11-17
[BUG] cudaErrorIllegalAddress when using PTDS and DEFAULT allocator (not ARENA)


abellina:
`ARENA` is the default allocator we run with in the plugin, and it works well with `PTDS`. I am debugging some OOM issues and went back to `DEFAULT` to see if I could reproduce there, but ran into other issues. 

If they should not be compatible, then we should disable `DEFAULT` in case we detect that `PTDS` is enabled.

This was TPCDS q6 @ SF10K using 2 DGX-2 machines. The error seems to be repeatable.

```
Caused by: ai.rapids.cudf.CudfException: CUDA error encountered at: /cudf17_nightly/cpp/src/io/utilities/hostdevice_vector.hpp:55: 700 cudaErrorIllegalAddress an illegal memory access was encountered
	at ai.rapids.cudf.Table.readParquet(Native Method)
	at ai.rapids.cudf.Table.readParquet(Table.java:675)
	at com.nvidia.spark.rapids.MultiFileParquetPartitionReader.$anonfun$readToTable$1(GpuParquetScan.scala:1043)
	at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
	at com.nvidia.spark.rapids.Arm.withResource$(Arm.scala:26)
	at com.nvidia.spark.rapids.FileParquetPartitionReaderBase.withResource(GpuParquetScan.scala:477)
	at com.nvidia.spark.rapids.MultiFileParquetPartitionReader.readToTable(GpuParquetScan.scala:1042)
	at com.nvidia.spark.rapids.MultiFileParquetPartitionReader.$anonfun$readBatch$1(GpuParquetScan.scala:1002)
	at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
	at com.nvidia.spark.rapids.Arm.withResource$(Arm.scala:26)
	at com.nvidia.spark.rapids.FileParquetPartitionReaderBase.withResource(GpuParquetScan.scala:477)
	at com.nvidia.spark.rapids.MultiFileParquetPartitionReader.readBatch(GpuParquetScan.scala:987)
	at com.nvidia.spark.rapids.MultiFileParquetPartitionReader.next(GpuParquetScan.scala:889)
	at org.apache.spark.sql.execution.datasources.v2.PartitionIterator.hasNext(DataSourceRDD.scala:79)
[...]

PR #1677, opened 2021-02-05
When PTDS is enabled, print warning if the allocator is not ARENA

abellina:

Add a disclaimer for #1141 (`cudaErrorillegalAddress` when PTDS and DEFAULT are used) for now while we root cause it.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala
```diff
@@ -204,6 +204,11 @@ object GpuDeviceManager extends Logging {
       if (conf.isPooledMemEnabled) {
         init = conf.rmmPool match {
           case c if "default".equalsIgnoreCase(c) =>
+            if (Cuda.isPtdsEnabled) {
+              logWarning("Configuring the DEFAULT allocator with a CUDF built for " +
+                  "Per-thread Default Stream (PTDS). This is known to be unstable! " +
```
jlowe, line 209:
```suggestion
                  "Per-Thread Default Stream (PTDS). This is known to be unstable! " +
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -352,7 +352,9 @@ object RapidsConf {
     .doc("Select the RMM pooling allocator to use. Valid values are \"DEFAULT\", \"ARENA\", and " +
       "\"NONE\". With \"DEFAULT\", `rmm::mr::pool_memory_resource` is used; with \"ARENA\", " +
       "`rmm::mr::arena_memory_resource` is used. If set to \"NONE\", pooling is disabled and RMM " +
-      "just passes through to CUDA memory allocation directly.")
+      "just passes through to CUDA memory allocation directly. Note: \"ARENA\" is the " +
+      "recommended pool allocator if CUDF is built with Per-thread Default Stream (PTDS), " +
```
jlowe, line 356:
```suggestion
      "recommended pool allocator if CUDF is built with Per-Thread Default Stream (PTDS), " +
```

sameerz:
build

abellina:
build

jlowe (approved):

PR #1678, opened 2021-02-05
Fix array/struct checks in Sort and HashAggregate and sorting tests in distributed mode

tgravescs:
I took this over from @kuhushukla. She found a couple of actual issues in Sort and HashAggregate where it wasn't checking types properly. Those fixes are here and then the rest is just with the test when running in distributed environment across multiple executors.
The issue here was just with the sort order being different because there was columns with duplicate keys.  To solve this I added a column that is just a unique integer so that you can sort on it and have the array and structs come along with it.

fixes https://github.com/NVIDIA/spark-rapids/issues/1541


tgravescs:
build

tgravescs:
reference https://github.com/NVIDIA/spark-rapids/pull/1653 was the original PR

tgravescs:
blackduck was having server maintenance, must be hitting that

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -44,6 +44,12 @@ class GpuSortMeta(
     if (GpuOverrides.isAnyStringLit(sort.sortOrder)) {
       willNotWorkOnGpu("string literal values are not supported in a sort")
     }
+    val sortOrderDataTypes = sort.sortOrder.map(_.dataType)
```
jlowe, line 47: Do we need a followup issue to leverage something like the TypeSig system so the limitations on sort keys, grouping keys, etc. are reflected in https://github.com/NVIDIA/spark-rapids/blob/branch-0.4/docs/supported_ops.md?

jlowe (approved):
I think we want a followup issue to get the type limitation documentation to be automatically generated for sort and grouping keys, but otherwise lgtm.

kuhushukla (approved):

sameerz:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
tgravescs: https://github.com/NVIDIA/spark-rapids/issues/1679

revans2: I am really confused by this change.  This should not be needed at all.  

https://github.com/NVIDIA/spark-rapids/blob/a830cffa4d305a8792605b2901ef32038d7bd576/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala#L1754

SortOrder already restricts these types so nested types will never get to this point.  @tgravescs did you not see that happen in practice?  If so then the type checking code is failing in some way. If not then #1679 really just needs a note in SortExec and the others that what columns can be sorted by is governed by SortOrder.

tgravescs: this was carry over from what Kuhu had found during various testing changes so I didn't see it myself.  Perhaps something else was going on there.  I can try to reproduce

revans2: I can try to reproduce it.  I am working on out of core sort and when I saw it on an upmerge I got scared and wanted to be sure that there wasn't a bug some other place in the code.  I'll do some tests manually.

revans2: I ran a simple test and the SortOrder rule is doing what I expected. When I remove the added checks.

```
spark.range(100).selectExpr("array(id, id+1, id-1) AS a", "id as b").orderBy(col("a")).write.mode("overwrite").parquet("./target/test")
```

```
  !Exec <SortExec> cannot run on GPU because not all expressions can be replaced
    !Expression <SortOrder> a#17 ASC NULLS FIRST cannot run on GPU because expression SortOrder a#17 ASC NULLS FIRST produces an unsupported type ArrayType(LongType,false); expression AttributeReference a#17 produces an unsupported type ArrayType(LongType,false)
      @Expression <AttributeReference> a#17 could run on GPU
    !Exec <ShuffleExchangeExec> cannot run on GPU because not all partitioning can be replaced
      !Partitioning <RangePartitioning> cannot run on GPU because not all expressions can be replaced
        !Expression <SortOrder> a#17 ASC NULLS FIRST cannot run on GPU because expression SortOrder a#17 ASC NULLS FIRST produces an unsupported type ArrayType(LongType,false); expression AttributeReference a#17 produces an unsupported type ArrayType(LongType,false)
          @Expression <AttributeReference> a#17 could run on GPU
```

But if I change the ordering to b instead of a I get almost all of it on the GPU.

```
!Exec <DataWritingCommandExec> cannot run on GPU because not all data writing commands can be replaced; unsupported data types in input: ArrayType(LongType,false)
  !Output <InsertIntoHadoopFsRelationCommand> cannot run on GPU because These types aren't supported for parquet List(StructField(a,ArrayType(LongType,false),false))
    *Expression <Alias> array(id#23L, (id#23L + 1), (id#23L - 1)) AS a#25 will run on GPU
      *Expression <CreateArray> array(id#23L, (id#23L + 1), (id#23L - 1)) will run on GPU
        *Expression <Add> (id#23L + 1) will run on GPU
        *Expression <Subtract> (id#23L - 1) will run on GPU
    *Expression <Alias> id#23L AS b#26L will run on GPU
    *Exec <RangeExec> will run on GPU
```

We don't have to rip this code out here. It is redundant. I just wanted to be sure that if there was a bug that we fixed the underlying bug or we replicated this code to other places that also take a SortOrder, like TakeOrderedAndProject

tgravescs: ok thanks for testing. I guess we can rip it out when its supported, should be obvious at that point.

Issue #1670, opened 2021-02-04
[BUG] Some TPC-DS queries fail with AQE when decimal types enabled


andygrove:
**Describe the bug**
I just tried running q4 and q11 with AQE and decimal types enabled, and they fail with:

```
Caused by: java.lang.IllegalStateException: operating on canonicalized plan
	at org.apache.spark.sql.execution.adaptive.CustomShuffleReaderExec.$anonfun$shuffleRDD$2(CustomShuffleReaderExec.scala:79)
```

**Steps/Code to reproduce bug**

I am running against Spark 3.0.1 using the benchmark runner Python script with the following template:

```bash
$SPARK_HOME/bin/spark-submit \
  --master local[*] \
  --driver-memory 64G \
  --jars $SPARK_RAPIDS_PLUGIN_JAR,$CUDF_JAR \
  --conf spark.executor.extraJavaOptions="-Dai.rapids.cudf.prefer-pinned=true -Dai.rapids.cudf.nvtx.enabled=true -Dai.rapids.spark.semaphore.enabled=true -Dai.rapids.spark.memory.gpu.rmm.init.task=false" \
  --conf spark.eventLog.enabled=true \
  --conf spark.eventLog.dir=./spark-event-logs \
  --conf spark.plugins=com.nvidia.spark.SQLPlugin \
  --conf spark.rapids.sql.enabled=true \
  --conf spark.rapids.sql.explain=ALL \
  --conf spark.rapids.sql.batchSizeBytes=50000000 \
  --conf spark.rapids.sql.reader.batchSizeBytes=50000000 \
  --conf spark.rapids.sql.concurrentGpuTasks=1 \
  --conf spark.rapids.sql.replaceSortMergeJoin.enabled=true \
  --conf spark.rapids.sql.variableFloatAgg.enabled=true \
  --conf spark.rapids.sql.castFloatToString.enabled=true \
  --conf spark.rapids.sql.decimalType.enabled=true \
  --conf spark.rapids.sql.hasNans=false \
  --conf spark.sql.adaptive.coalescePartitions.enabled=true \
  --conf spark.sql.adaptive.coalescePartitions.minPartitionNum=200 \
  --conf spark.sql.adaptive.skewJoin.skewedPartitionFactor=2 \
  --conf spark.sql.shuffle.partitions=400 \
  --conf spark.sql.optimizer.dynamicPartitionPruning.enabled=true \
[...]

PR #1685, opened 2021-02-08
Fix CustomShuffleReader replacement when decimal types enabled

sperlingxx:
This pull request is to fix the bug of AQE with decimal, which is reported in issue #1670.

### DIFF
--- a/docs/supported_ops.md
+++ b/docs/supported_ops.md
@@ -9,7 +9,7 @@ support all data types. The RAPIDS Accelerator for Apache Spark has further
 restrictions on what types are supported for processing. This tries
 to document what operations are supported and what data types each operation supports.
 Because Apache Spark is under active development too and this document was generated
-against version 3.0.0 of Spark. Most of this should still
+against version 3.0.1 of Spark. Most of this should still
 apply to other versions of Spark, but there may be slight changes.
 
 # General limitations
@@ -419,13 +419,13 @@ Accelerator supports are described below.
 <td>S</td>
 <td>S*</td>
 <td>S</td>
+<td>S*</td>
+<td>S</td>
 <td><b>NS</b></td>
 <td><b>NS</b></td>
+<td><em>PS* (missing nested BINARY, CALENDAR, MAP, UDT)</em></td>
 <td><b>NS</b></td>
-<td><b>NS</b></td>
-<td><b>NS</b></td>
-<td><b>NS</b></td>
-<td><b>NS</b></td>
+<td><em>PS* (missing nested BINARY, CALENDAR, MAP, UDT)</em></td>
 <td><b>NS</b></td>
 </tr>
 <tr>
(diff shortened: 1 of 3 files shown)


sperlingxx:
build

tests/src/test/scala/com/nvidia/spark/rapids/AdaptiveQueryExecSuite.scala
```diff
@@ -31,6 +31,7 @@ import org.apache.spark.sql.execution.joins.SortMergeJoinExec
 import org.apache.spark.sql.functions.{col, when}
```
jlowe, line 31: 2021 copyrights

jlowe (commented):
Copyrights need to be updated but otherwise lgtm.

revans2:
@sperlingxx Can this impact other types too?  We support shuffling nested types like arrays, maps, and structs, but GpuCustomShuffleReader does not.

andygrove:
Thanks @sperlingxx . I applied this change to my branch and ran quite a few queries with decimal + AQE enabled and it worked well.

abellina:
Is there a way to make this fail in a way that we can fallback to the cpu?

tests/src/test/scala/com/nvidia/spark/rapids/AdaptiveQueryExecSuite.scala
sperlingxx: Fixed.

sperlingxx:
> @sperlingxx Can this impact other types too? We support shuffling nested types like arrays, maps, and structs, but GpuCustomShuffleReader does not.

Hi @revans2, I added ArrayType and StructType as supported types. And MapType is not supported in `ShuffleExchangeExec`.

sperlingxx:
> Is there a way to make this fail in a way that we can fallback to the cpu?

I believe when we configure `spark.rapids.sql.decimalType.enabled` with false, it will fallback to cpu. The fallback strategy will fail at here because parent plan(CustomShuffleReaderExec) fallbacks but child plan does not.

sperlingxx:
build

docs/supported_ops.md
```diff
@@ -9,7 +9,7 @@ support all data types. The RAPIDS Accelerator for Apache Spark has further
 restrictions on what types are supported for processing. This tries
 to document what operations are supported and what data types each operation supports.
 Because Apache Spark is under active development too and this document was generated
-against version 3.0.0 of Spark. Most of this should still
+against version 3.0.1 of Spark. Most of this should still
```
revans2, line 12: The default is still to use 3.0.0 for building the docs.  What happened to change it to 3.0.1 when the build itself did not change?

sperlingxx: My bad. I ran `mvn verify` with `Dspark.version=3.0.1` to update the doc.

sperlingxx:
build

jlowe (approved):

PR #1688, opened 2021-02-08
[WIP] Allow repeated iterations of integration tests in the same Spark app

gerashegalov:

Reproduces #1522 typically at the rate of a few instances per hour
```bash
IT_MAX_ITERATIONS=-1 SPARK_HOME=~/spark-311 SPARK_SUBMIT_FLAGS='--master local-cluster[1,20,20000]' \
integration_tests/run_pyspark_from_build.sh -k 'test_select'
```


gerashegalov:
build

integration_tests/runtests.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
```
jlowe, line 13: 2021 copyrights

integration_tests/runtests.py
```diff
@@ -22,4 +23,16 @@
     #cProfile.run('main(sys.argv[1:])', 'test_profile')
     # arguments are the same as for pytest https://docs.pytest.org/en/latest/usage.html
     # or run pytest -h
-    sys.exit(main(sys.argv[1:]))
+    iteration = 0
+    maxIterEnvName = 'IT_MAX_ITERATIONS'
+    maxIterations = int(os.environ[maxIterEnvName]) if maxIterEnvName in os.environ else 1
```
jlowe, line 28:
```suggestion
    maxIterations = int(os.environ.get('IT_MAX_ITERATIONS', 1))
```

integration_tests/runtests.py
```diff
@@ -22,4 +23,16 @@
     #cProfile.run('main(sys.argv[1:])', 'test_profile')
     # arguments are the same as for pytest https://docs.pytest.org/en/latest/usage.html
     # or run pytest -h
-    sys.exit(main(sys.argv[1:]))
+    iteration = 0
+    maxIterEnvName = 'IT_MAX_ITERATIONS'
```
jlowe, line 27: Curious why this isn't a command-line argument instead of an environment variable?  We already use custom command-line arguments for other features, so curious why we wouldn't do the same here.  Should be trivial to search the arguments for this option, extract it, and pass the rest to pytest.

jlowe:
Also the comment for the commit in this PR is awful.  Please rebase and fix it -- it's worth it in the long run.

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/SparkShimServiceProvider.scala
```diff
@@ -22,7 +22,9 @@ object SparkShimServiceProvider {
   // keep 3.1.0 snapshot version for now until 3.1.1 is released
   val VERSION310 = SparkShimVersion(3, 1, 0)
   val VERSION = SparkShimVersion(3, 1, 1)
-  val VERSIONNAMES = Seq(s"$VERSION310-SNAPSHOT", s"$VERSION", s"$VERSION-SNAPSHOT")
+  val VERSION320 = SparkShimVersion(3, 2, 0)
+  val VERSIONNAMES: Seq[String] = Seq(VERSION310, VERSION, VERSION320)
```
revans2, line 26: I would like to see `3.2.0` added as a separate shim jar because once `3.1.1` is released (which should be soon) we will want to add `3.1.1` to the jars we include in the official output, but not `3.2.0-SNAPSHOT`

integration_tests/run_pyspark_from_build.sh
```diff
@@ -112,6 +112,7 @@ else
           --conf 'spark.executor.extraJavaOptions=-ea -Duser.timezone=UTC' \
           --conf 'spark.sql.session.timeZone=UTC' \
           --conf 'spark.sql.shuffle.partitions=12' \
+          --conf "spark.executor.extraClassPath=${ALL_JARS// /:}" \
```
revans2, line 115: Why do we need this in here? Is this for distributed mode? If so then you probably want to 
```
export PYSP_TEST_spark_executor_extraClassPath="${ALL_JARS// /:}"
```
instead so that the version that runs in parallel with pytest-xdist will pick up this change too.

integration_tests/runtests.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import sys
+import os
```
revans2, line 16: Have you looked into using https://pypi.org/project/pytest-repeat/ instead of modifying this script?

integration_tests/runtests.py
```diff
@@ -22,4 +23,16 @@
     #cProfile.run('main(sys.argv[1:])', 'test_profile')
     # arguments are the same as for pytest https://docs.pytest.org/en/latest/usage.html
     # or run pytest -h
-    sys.exit(main(sys.argv[1:]))
+    iteration = 0
```
revans2, line 26: In the documentation at https://github.com/NVIDIA/spark-rapids/blob/branch-0.4/integration_tests/README.md it says

```
The only reason we have a separate script is that spark-submit uses python if the file name ends with .py.
``` 

After this change it is no longer true.  If you decide not to use `pytest-repeat` could you please update the documentation to explain how this works and remove the comment about why we created this script.

integration_tests/run_pyspark_from_build.sh
gerashegalov: yes, it's for `local-cluster` or `standalone` backends

integration_tests/runtests.py
gerashegalov: no, I am not familiar with Python yet, thanks for the suggestion @revans2

integration_tests/run_pyspark_from_build.sh
```diff
@@ -85,14 +85,20 @@ else
     RUN_DIR="$SCRIPTPATH"/target/run_dir
     mkdir -p "$RUN_DIR"
     cd "$RUN_DIR"
+
+    export PYSP_TEST_spark_driver_extraJavaOptions="-ea -Duser.timezone=UTC $COVERAGE_SUBMIT_FLAGS"
+    export PYSP_TEST_spark_driver_extraClassPath="${ALL_JARS// /:}"
```
revans2, line 90: I think when running with spark-submit configs that impact the driver command line arguments have to be set on the command line of spark-submit because the driver is launched first and it launches the python process.  When using find-spark that is not the case. 

Did you test these changes with `TEST_PARALLEL=1`?

gerashegalov: Double-checked with the command in the description for `TEST_PARALLEL=1` and `TEST_PARALLEL=2`

integration_tests/runtests.py
gerashegalov: `pytest-repeat` works per se. It's very slow to start for larger repetition counts such as 1000. Interestingly, the way it inflates the test series from 1 2 3 to let us say count=3:  1 1 1 2 2 2 3 3 3 does not reproduce the particular problem as fast. 

But this peculiarity does not mean we have incorporate the script change given the functionality of pytest-repeat is more than sufficient in general, I can just keep it on a branch for reference.

gerashegalov:
This PR branch was the test to validate https://github.com/apache/spark/pull/31540 . It's not critical for 0.4. We can think if we need to salvage anything from this PR in the 0.5 timeframe.

sameerz:
@gerashegalov can you retarget this to branch-0.5?

gerashegalov:
The PR does not add much functionality other then defining the workload that was able to trigger NPE in #1522 fast. If it's worth it the test repetition logic from this PR can be suggested  as an option for pytest. Closing.

PR #1694, opened 2021-02-09
increase mem and limit parallelism for pre-merge

pxLi:

Increase memory request to avoid potential OOM kill for pre-merge pod run during the parallel pytest run
And limit parallelism to make sure if we change GPU type w/ larger GPU mem, it won't be OOM in the pod unless we manually change both GPU and adjust host memory


pxLi:
build

jlowe (approved):

revans2 (approved):
Looks like it only added a few mins to the build time, which is a lot better then crashing.

Issue #584, opened 2020-08-19
[DOC] Spark3.1.0 should not need the extraclasspath set for standalone more anymore


tgravescs:
For spark 3.1.0 we can update our docs to remove the need for the extraclasspath when using the plugin with Standalone Mode on the executor side. We should test this and then update our docs.

https://issues.apache.org/jira/browse/SPARK-32119

tgravescs:
tested thsi work on spark 3.1.1. rc2

PR #1699, opened 2021-02-09
Add docs about Spark 3.1 in standalone modes not needing extra class path

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/584


tgravescs:
build

tgravescs:
I was trying to note have to make a bunch of extra sections where its like here is example for 3.0 and here is one for 3.1.1 so decided on the note, if others have better ideas let me know.

jlowe (approved):

Issue #576, opened 2020-08-18
[FEATURE] Spark 3.1 orc nested predicate pushdown support


tgravescs:
**Describe the bug**
Spark 3.1 added nested predicate pushdown support in ORC - https://github.com/apache/spark/commit/7b6e1d5cec60dbb43c6fbc1528a207cbff604a5d

This removed a function we were using:
E                   Caused by: java.lang.NoSuchMethodError: org.apache.spark.sql.execution.datasources.orc.OrcFiltersBase.isSearchableType$(Lorg/apache/spark/sql/execution/datasources/orc/OrcFiltersBase;Lorg/apache/spark/sql/types/DataType;)Z

We should go through and add the same support or fix it to use the new function.

**Steps/Code to reproduce bug**
Run spark 3.1.0 integration tests:
FAILED integration_tests/src/main/python/orc_test.py::test_input_meta - py4j....

**Expected behavior**
tests pass

**Environment details (please complete the following information)**
 - Environment location: [Standalone, YARN, Kubernetes, Cloud(specify cloud provider)]
 - Spark configuration settings related to the issue

**Additional context**

tgravescs:
this got backported to 3.0.1 so we need to fix for 0.2 release

tgravescs:
it looks like changes in 3.0.1 and 3.1.0 are also different so we would need a shim layer. Temporarily I'm going to put in a change to just copy the function that got moved to private:

  - protected[sql] def isSearchableType(dataType: DataType) = dataType match {
  + private def isSearchableType(dataType: DataType) = dataType match {

And then we can look at this in more detail to support it fully.

tgravescs:
ok so leaving a P2 because we fixed the function and it works, this is only needed if we want to support nested predicate pushdown

sameerz:
Depends on https://github.com/rapidsai/cudf/issues/7640 and https://github.com/rapidsai/cudf/issues/7830

tgravescs:
our orc filter code was updated with https://github.com/NVIDIA/spark-rapids/pull/1982 and it looks like this should just work, so perhaps we just need to test it.

jlowe:
Closing as a duplicate of #463

PR #1701, opened 2021-02-10
remove xfail for orc test_input_meta for spark 3.1.0

tgravescs:

reference https://github.com/NVIDIA/spark-rapids/issues/576 where we actually made it work with a different PR, the thing left on that issue is to support predicate pushdown for nested types when required. This xfail wasn't removed before.


tgravescs:
build

jlowe (approved):

PR #1713, opened 2021-02-11
Call GpuDeviceManager.shutdown when the executor plugin is shutting down

abellina:

@gerashegalov  noticed the following host memory leak in unit tests:

```
21/02/11 15:56:42.217 Thread-7 ERROR MemoryCleaner: Leaked host buffer (ID: 1): 2021-02-11 21:55:41.0075 UTC: INC
java.lang.Thread.getStackTrace(Thread.java:1559)
ai.rapids.cudf.MemoryCleaner$RefCountDebugItem.<init>(MemoryCleaner.java:283)
ai.rapids.cudf.MemoryCleaner$Cleaner.addRef(MemoryCleaner.java:82)
ai.rapids.cudf.MemoryBuffer.incRefCount(MemoryBuffer.java:199)
ai.rapids.cudf.MemoryBuffer.<init>(MemoryBuffer.java:98)
ai.rapids.cudf.HostMemoryBuffer.<init>(HostMemoryBuffer.java:196)
ai.rapids.cudf.HostMemoryBuffer.<init>(HostMemoryBuffer.java:192)
ai.rapids.cudf.HostMemoryBuffer.allocate(HostMemoryBuffer.java:144)
com.nvidia.spark.rapids.RapidsHostMemoryStore.<init>(RapidsHostMemoryStore.scala:34)
com.nvidia.spark.rapids.RapidsBufferCatalog$.init(RapidsBufferCatalog.scala:139)
com.nvidia.spark.rapids.GpuDeviceManager$.initializeRmm(GpuDeviceManager.scala:262)
com.nvidia.spark.rapids.GpuDeviceManager$.initializeMemory(GpuDeviceManager.scala:292)
com.nvidia.spark.rapids.GpuDeviceManager$.initializeGpuAndMemory(GpuDeviceManager.scala:126)
```

This was because the `RapidsHostMemoryStore` (and its pool) were not being shutdown. The included here removes the leak from the tests.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2020, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2021, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
@@ -210,6 +210,7 @@ class RapidsExecutorPlugin extends ExecutorPlugin with Logging {
   override def shutdown(): Unit = {
     GpuSemaphore.shutdown()
     PythonWorkerSemaphore.shutdown()
+    GpuDeviceManager.shutdown()
   }
 }
 


sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -210,6 +210,7 @@ class RapidsExecutorPlugin extends ExecutorPlugin with Logging {
   override def shutdown(): Unit = {
```
jlowe, line 210: 2021 copyrights

abellina: :) coming up.

jlowe (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -210,6 +210,7 @@ class RapidsExecutorPlugin extends ExecutorPlugin with Logging {
...
     GpuSemaphore.shutdown()
     PythonWorkerSemaphore.shutdown()
+    GpuDeviceManager.shutdown()
```
gerashegalov, line 213: I wonder if we should make these classes (Auto)Closable? then we could use hadoop's IOUtils 
to `IOUtils.cleanup(null, GpuSemaphore,  PythonWorkerSemaphore, GpuDeviceManager)`

gerashegalov (approved):
LGTM

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
abellina: So we actually have `safeClose` already. `safeClose` will call `.close()` on every member of the collection, but then will throw at the end if there was a problem closing any member. I can certainly do this as part of this PR, the classes we care about are exactly these three.

jlowe: Agree not necessary for this PR.

abellina: Ok lets skip it for now, we can follow up with a cleanup PR.

abellina:
build

PR #1714, opened 2021-02-11
Add support for RAPIDS accelerated Java UDFs

### LINKED ISSUE #1635: [FEA] RAPIDS accelerated Java UDF

jlowe:
Fixes #1635.

This extends `GpuScalaUDF` to recognize RAPIDS accelerated Spark Java UDFs that are wrapped in a lambda function by Apache Spark by `UDFRegistration`.  This allows users to provide RAPIDS accelerated implementation of their Spark Java UDFs for use in Spark via `spark.udf.register` and PySpark via `spark.udf.registerJavaFunction`.

A working example of a RAPIDS accelerated Spark Scala UDF is also provided, which required adding the Scala version to the udf-examples jar (and everywhere it was referenced). A unit test was added to exercise it. It was not implemented as a Python test as was done for the Hive UDFs because PySpark does not support registering Scala UDFs (it uses the Java UDF interface instead).

The RAPIDS accelerated UDF documentation has also been updated to reflect the new functionality.

### DIFF
--- a/docs/additional-functionality/rapids-udfs.md
+++ b/docs/additional-functionality/rapids-udfs.md
@@ -33,16 +33,20 @@ on the GPU.
 The RAPIDS Accelerator only supports RAPIDS accelerated forms of the
 following UDF types:
 - Scala UDFs implementing a `Function` interface and registered via `SparkSession.udf.register`
+- Java UDFs implementing
+  [one of the `org.apache.spark.sql.api.java.UDF` interfaces](https://github.com/apache/spark/tree/branch-3.0/sql/core/src/main/java/org/apache/spark/sql/api/java)
+  and registered either via `SparkSession.udf.register` or
+  `spark.udf.registerJavaFunction` in PySpark
 - [Simple](https://github.com/apache/hive/blob/cb213d88304034393d68cc31a95be24f5aac62b6/ql/src/java/org/apache/hadoop/hive/ql/exec/UDF.java)
   or
   [Generic](https://github.com/apache/hive/blob/cb213d88304034393d68cc31a95be24f5aac62b6/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDF.java)
   Hive UDFs
 
 Other forms of Spark UDFs are not supported, such as:
-- Spark Java UDFs (i.e.: derived from `org.apache.spark.sql.api.java.UDF`* interfaces)
+- Scala or Java User-Defined Aggregate Functions
 - Hive Aggregate Function (UDAF)
 - Hive Tabular Function (UDTF)
-- Lambda functions
+- Lambda Functions
 
 ## Adding GPU Implementations to UDFs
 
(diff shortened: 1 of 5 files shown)


abellina (approved):

abellina:
build

docs/additional-functionality/rapids-udfs.md
```diff
@@ -33,16 +33,20 @@ on the GPU.
 The RAPIDS Accelerator only supports RAPIDS accelerated forms of the
 following UDF types:
 - Scala UDFs implementing a `Function` interface and registered via `SparkSession.udf.register`
+- Java UDFs implementing
+  [one of the `org.apache.spark.sql.api.java.UDF` interfaces](https://github.com/apache/spark/tree/branch-3.0/sql/core/src/main/java/org/apache/spark/sql/api/java)
```
tgravescs, line 37: referencing branch-3.0 here, is that what you want?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalaUDF.scala
```diff
@@ -45,24 +47,50 @@ object GpuScalaUDF {
...
+   * if it does. The lambda wrapper that Spark applies to Java UDFs will be inspected if necessary
+   * to locate the user's UDF instance.
+   */
+  private def getRapidsUDFInstance(function: AnyRef): Option[RapidsUDF] = {
+    function match {
+      case f: RapidsUDF => Some(f)
+      case f =>
+        try {
+          val clazz = f.getClass
+          val writeReplace = clazz.getDeclaredMethod("writeReplace")
```
tgravescs, line 80: this might be worth putting more detailed comments on what writeReplace method is and how this is working.

docs/additional-functionality/rapids-udfs.md
jlowe: Yes, in case Apache Spark decides to change it later or renames the `master` branch and the link ends up dangling.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalaUDF.scala
jlowe: I added some comments and also made the check more discriminating by checking the class name to see if it looks like a lambda instance before blindly invoking a `writeReplace` method if we find one.

jlowe:
build

tgravescs (approved):

PR #1717, opened 2021-02-12
Decimal32 support

razajafri:
This PR handles Decimal32 support. 

- Tests for division are ~~failing~~ all passing
- Support for nested types for reading legacy Decimals is ~~lacking~~ done
- ~~While copying vector back on the Host in `HostColumnarToGpu` I am getting the `BigDecimal` value, it could be optimized by getting the unscaledLong value like we were before this PR~~ This is now handled based on the precision
- ~~There might be issues with casting from D32 to D64 that I was seeing before but I don't see them now. It could be attributed to the scale value flipped in cudf, so I might have made a mistake while evaluating them~~ this was happening because I was confused between the scale values being opposite in cudf


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -124,6 +126,179 @@ object GpuCast {
...
+   * Asserts that all values in a column are within the specific range.
+   *
+   * @param values ColumnVector to be performed with range check
+   * @param minValue Named parameter for function to create Scalar representing range minimum value
+   * @param maxValue Named parameter for function to create Scalar representing range maximum value
+   * @param inclusiveMin Whether the min value is included in the valid range or not
+   * @param inclusiveMax Whether the max value is included in the valid range or not
+   * @throws IllegalStateException if any values in the column are not within the specified range
+   */
+  private def assertValuesInRange(values: ColumnVector,
```
razajafri, line 140: this method was moved from GpuCast class

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -124,6 +126,179 @@ object GpuCast {
...
+   * a inputted substitution value.
+   *
+   * @param values ColumnVector to be performed with range check
+   * @param minValue Named parameter for function to create Scalar representing range minimum value
+   * @param maxValue Named parameter for function to create Scalar representing range maximum value
+   * @param replaceValue Named parameter for function to create scalar to substitute outlier value
+   * @param inclusiveMin Whether the min value is included in the valid range or not
+   * @param inclusiveMax Whether the max value is included in the valid range or not
+   */
+  private def replaceOutOfRangeValues(values: ColumnVector,
```
razajafri, line 184: This method was moved from GpuCast class

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -124,6 +126,179 @@ object GpuCast {
...
+      // We do NOT use `Scalar.fromDecimal(-to.scale, math.pow(10, 18).toLong)` here, because
+      // cuDF binaryOperation on decimal will rescale right input to fit the left one.
+      // The rescaling may lead to overflow.
+      // absBound is the maximum value that the input column can have before being casted
+      val prec = to.precision + from.scale - to.scale
+
+      // if target precision is greater than and smaller than the max/min precision that can
+      // be held in the input, go ahead with the cast without further checking
+      if (isFrom32Bit && prec > Decimal.MAX_INT_DIGITS ||
+        !isFrom32Bit && prec > Decimal.MAX_LONG_DIGITS) {
```
razajafri, line 254: This should be commented out because cudf doesn't support Decimal types > 64-bit

sql-plugin/src/main/scala/com/nvidia/spark/rapids/decimalExpressions.scala
```diff
@@ -56,8 +56,12 @@ case class GpuUnscaledValue(child: Expression) extends GpuUnaryExpression {
   override def toString: String = s"UnscaledValue($child)"
 
   override protected def doColumnar(input: GpuColumnVector): ColumnVector = {
-    withResource(input.getBase.logicalCastTo(DType.INT64)) { view =>
-      view.copyToColumnVector()
+    if (input.getBase.getType.isDecimalType && input.getBase.getType.isBackedByInt) {
```
razajafri, line 59: checking isDecimalType isn't necessary I think

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
```diff
@@ -65,7 +65,7 @@ object GpuScalar {
     case DType.TIMESTAMP_DAYS => v.getInt
     case DType.TIMESTAMP_MICROSECONDS => v.getLong
     case DType.STRING => v.getJavaString
-    case dt: DType if dt.isDecimalType && dt.isBackedByLong => Decimal(v.getBigDecimal)
+    case dt: DType if dt.isDecimalType => Decimal(v.getBigDecimal)
```
razajafri, line 68: Might be able to optimize it in the int case?

sql-plugin/src/main/java/org/apache/spark/sql/catalyst/CudfUnsafeRow.java
```diff
@@ -239,7 +239,9 @@ public Decimal getDecimal(int ordinal, int precision, int scale) {
       return null;
     }
     // TODO when DECIMAL32 is supported a special case will need to be added here
```
revans2, line 241: Why is the TODO still here?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DecimalUtil.scala
```diff
@@ -0,0 +1,40 @@
...
+import ai.rapids.cudf.DType
+
+import org.apache.spark.sql.types.{DataType, Decimal, DecimalType}
+
+object DecimalUtil {
+
+  def createCudfDecimal(precision: Int, scale: Int): DType = {
+    if (precision <= Decimal.MAX_INT_DIGITS) {
+      DType.create(DType.DTypeEnum.DECIMAL32, -scale)
+    }
```
revans2, line 28: nit `} else {`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DecimalUtil.scala
```diff
@@ -0,0 +1,40 @@
...
+    else {
+      DType.create(DType.DTypeEnum.DECIMAL64, -scale)
+    }
+  }
+
+  private[rapids] def getDataTypeSize(dt: DataType): Int = {
```
revans2, line 34: We need docs for this, because it is not clear at all what this is doing.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -1055,43 +1059,89 @@ case class GpuCast(
...
-      // No need to promote precision unless target scale is larger than the source one,
-      // which indicates the cast is always valid when to.scale <= from.scale.
-      input.incRefCount()
-    } else {
+    val isFrom32Bit = DecimalType.is32BitDecimalType(from)
+    val isTo32Bit = DecimalType.is32BitDecimalType(to)
+    val cudfDecimal = DecimalUtil.createCudfDecimal(to.precision, to.scale)
+
+    def castCheckedDecimal(checkedInput: ColumnVector): ColumnVector = {
+      to.scale - from.scale match {
```
revans2, line 1070: nit: I'm not sure a match in this case is cleaner.  I personally think `to.scale == from.scale` is much easier to read than `to.scale - from.scale ... case 0`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -1055,43 +1059,89 @@ case class GpuCast(
...
+    }
+
+    def checkForOverflow: ColumnVector = {
       // Check whether there exists overflow during promoting precision or not.
       // We do NOT use `Scalar.fromDecimal(-to.scale, math.pow(10, 18).toLong)` here, because
       // cuDF binaryOperation on decimal will rescale right input to fit the left one.
       // The rescaling may lead to overflow.
-      val absBound = math.pow(10, DType.DECIMAL64_MAX_PRECISION + from.scale - to.scale).toLong
-      if (ansiMode) {
+      // absBound is the maximum value that the input column can have before being casted
```
revans2, line 1092: This comment feels really out of place, because absBound is not set until much later.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -1055,43 +1059,89 @@ case class GpuCast(
...
+      val prec = to.precision + from.scale - to.scale
```
revans2, line 1093: what is prec? it appears to be the intermediate precision that cudf will do internally.  If so prec is a fairly bad name for it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -1055,43 +1059,89 @@ case class GpuCast(
...
+
+      // if target precision is greater than and smaller than the max/min precision that can
```
revans2, line 1095: The condition checks are rather confusing and this comment does not help me understand them. I think trying to combine the min/max into a single sentence makes it unintelligible.  Could we try it a little differently?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -1055,43 +1059,89 @@ case class GpuCast(
...
+      // be held in the input, go ahead with the cast without further checking
+      // When we support 128 bit Decimals we should add a check for that
+      // if (isFrom32Bit && prec > Decimal.MAX_INT_DIGITS ||
+      // !isFrom32Bit && prec > Decimal.MAX_LONG_DIGITS)
+      if (isFrom32Bit && prec > Decimal.MAX_INT_DIGITS) {
```
revans2, line 1100: We we should not check for overflow on a 32 bit value if intermediate result would be larger than a 32 bit value can hold?  I'm not really following the logic of that, and would like to see some more comments explaining why that is.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -1055,43 +1059,89 @@ case class GpuCast(
...
+        return input.incRefCount()
```
revans2, line 1101: I really don't like the early return here. It is inside a nested method which makes it confusing from the code what point is being returned from what method. It might be best to either refactor the code so that `checkForOverflow` became something like `checkForDecimalOverflow` and was a stand alone method, or make it so we don't return early.

revans2 (commented):
I didn't have time to finish all of the review, but I thought I would get some of my comments in for today.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
razajafri: I agree, this code was riddled with confusion. I will try to refactor and make it more readable

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
razajafri: Its not the intermediate precision.  When moving from a smaller scale to a bigger scale (or 32-bit to 64-bit), the target type is able to hold much more values on the fraction side which leaves less room for the whole number. e.g.
```      
      999999.999 => from.scale = 3
      9999.99999 => to.scale = 5
```
this means that there could be values in the input that will not be able to fit in the target type, so we calculate the max number that should be in the input column before we can safely cast the values without overflowing. In the above example the source can have a maximum of 4 digits before decimal point and 3 after the decimal point. Therefore the values of absMax and absMin will be calculated based on the prec value.

revans2: Then put that in the comments.

razajafri:
@revans2 I have addressed your review comments. I have also added more comments in the cast method and simplified the math. PTAL

razajafri:
build

razajafri:
CI is failing because of the cudf dependency. Can I get more reviews on that please? https://github.com/rapidsai/cudf/pull/7417

razajafri:
@revans2 any more comments here? I was hoping to get this change in before the `TRUE_DIV` gets in

razajafri:
build

razajafri:
build

razajafri:
The test is failing intermittently. I will investigate this when I am back at work

razajafri:
build

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -716,7 +720,55 @@ abstract class FileParquetPartitionReaderBase(
         newColumns.safeClose()
       }
     } else {
-      inTable
+      inputTable
+    }
+  }
+
+  private def convertDecimal64ToDecimal32(cv: ColumnView, precisionList: Queue[Int]): ColumnView = {
```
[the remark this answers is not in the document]
razajafri: The problem is the List API in our Java bindings returns ColumnView so either I copy it out to a ColumnVector at every stage in case of a multi-nested list or just take a ColumnView as an input and do the dirty checks outside.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -680,25 +655,54 @@ abstract class FileParquetPartitionReaderBase(
...
-        withResource(inTable) { table =>
+        withResource(inputTable) { table =>
           var readAt = 0
           (0 until readDataSchema.length).foreach(writeAt => {
             val readField = readDataSchema(writeAt)
             if (areNamesEquiv(clippedGroups, readAt, readField.name, isSchemaCaseSensitive)) {
-              newColumns(writeAt) = table.getColumn(readAt).incRefCount()
+              val origCol = table.getColumn(readAt)
+              val col = if (typeCastingNeeded && precisionList.nonEmpty) {
+                val tmp = convertDecimal64ToDecimal32(origCol, precisionList)
```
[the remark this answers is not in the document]
revans2: I don't think you understood me. I realize that for efficiency we need to do some of these tricks. But the way it is written now anyone calling `convertDecimal64ToDecimal32` needs to know about this trick. I can see in the future when ORC, or CSV, or what ever needs supports decimal all of them will need to do something similar.  I would like a front door method that hides all of the ugliness.  Internally we have have a method that is doing the oddness, but the public facing API should be clean.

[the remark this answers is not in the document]
razajafri: It's a private method but I see your point, even for someone from this class calling it shouldn't need to know the internals. I will push an update after I figure out the failure

razajafri:
build

razajafri:
@revans2 I have filed a couple of follow-on issues. One for refactoring the binaryoperations [here](https://github.com/NVIDIA/spark-rapids/issues/2077) and the other one for making the creation of scalar simpler [here](https://github.com/NVIDIA/spark-rapids/issues/2076). PTAL and let me know if you have any more concerns

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DecimalUtil.scala
```diff
@@ -0,0 +1,43 @@
...
+
+package com.nvidia.spark.rapids
+
+import ai.rapids.cudf.DType
+
+import org.apache.spark.sql.types.{DataType, Decimal, DecimalType}
+
+object DecimalUtil {
+
+  def createCudfDecimal(precision: Int, scale: Int): DType = {
```
revans2, line 25: nit: can we have some comments here? I would like it clean that the input precision and scale should be what Spark expects and this will convert it into whatever CUDF expects.

revans2 (approved):
Just a nit left, and I am OK with this going in without it

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DecimalUtil.scala
razajafri: I will fix this in one of my PRs

Issue #1117, opened 2020-11-13
[BUG] CAST string to date returns wrong values for dates with out-of-range values


andygrove:
**Describe the bug**
When casting string to date, the GPU will convert dates with out-of-range values rather than treat them as invalid. For example, the string `2020-02-31` will be converted to `2020-03-02`.

**Steps/Code to reproduce bug**
Add `2020-02-31` to com.nvidia.spark.rapids.CastOpSuite#timestampsAsStrings

**Expected behavior**
Dates with out-of-range values should be treated as invalid.

andygrove:
I filed a feature request against cuDF: https://github.com/rapidsai/cudf/issues/6774

andygrove:
I spent some time working on this today. If we update `GpuCast` methods `convertDateOr`, `convertDateOrNull`, `convertTimestampOr`, and `convertTimestampOrNull` to call the `isTimestamp` method then it fixes this issue but also causes regressions in other tests that were previously passing, so this will need more investigation.

andygrove:
This was fixed in https://github.com/NVIDIA/spark-rapids/pull/1718

Issue #716, opened 2020-09-10
[BUG] Cast String to TimeStamp issues


tgravescs:
**Describe the bug**

We recently discovered that the config to turn off the string to timestamp cast was not being used.  We fixed that with https://github.com/NVIDIA/spark-rapids/pull/705.

But while investigating that it seemed like a few things might be off. the format we are using and is used in CUDF is "%Y-%m-%dT%H:%M:%SZ%f". Based on what we support it seems like the ms section should be %SZ.%f.  The pattern that we didn't match on but worked on the CPU was just: 2017-11-29 20:00:35. Ideally we would support this.

I think the regex's in this case are a pretty high performance overhead as well though so perhaps we should figure out different way to handle.

tgravescs:
Note I have seen other formats for strings - such as 2020-09-07T01:05:57.840+0000
  that we also don't support.

sameerz:
This is related to issue #987

andygrove:
This is also related to https://github.com/NVIDIA/spark-rapids/issues/1117 which I am currently working on

andygrove:
> Note I have seen other formats for strings - such as 2020-09-07T01:05:57.840+0000
> that we also don't support.

@tgravescs The main issue raised in this issue (using the wrong cuDF timestamp formats) was resolved by https://github.com/NVIDIA/spark-rapids/pull/1718 and there is now a separate issue for reducing the regex overhead (https://github.com/NVIDIA/spark-rapids/issues/1738).

We do not support the format `2020-09-07T01:05:57.840+0000` and this is documented. Should we file a separate issue for adding support?

tgravescs:
yes I think we should file one to say supported more formats

andygrove:
I filed https://github.com/NVIDIA/spark-rapids/issues/1748 for supporting additional formats.

PR #1718, opened 2021-02-12
CAST string to temporal type now calls isTimestamp

andygrove:

CAST string to temporal types now uses `isTimestamp` to detect valid inputs.

Closes https://github.com/NVIDIA/spark-rapids/issues/1117 and https://github.com/NVIDIA/spark-rapids/issues/716


andygrove:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -97,8 +97,10 @@ object GpuCast {
   private val TIMESTAMP_REGEX_YYYY = "\\A\\d{4}\\Z"
```
jlowe, line 97: 2021 copyrights

tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
```diff
@@ -943,6 +943,10 @@ object CastOpSuite {
         "2018-1random_text",
```
jlowe, line 943: 2021 copyrights

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -779,10 +804,11 @@ case class GpuCast(
 
       // convert dates that are in valid timestamp formats
       val converted =
-        convertTimestampOr(sanitizedInput, TIMESTAMP_REGEX_FULL, "%Y-%m-%dT%H:%M:%SZ%f",
-          convertTimestampOr(sanitizedInput, TIMESTAMP_REGEX_YYYY_MM_DD, "%Y-%m-%d",
-            convertTimestampOr(sanitizedInput, TIMESTAMP_REGEX_YYYY_MM, "%Y-%m",
-              convertTimestampOrNull(sanitizedInput, TIMESTAMP_REGEX_YYYY, "%Y"))))
+        convertTimestampOr(sanitizedInput, TIMESTAMP_REGEX_FULL_1, "%Y-%m-%d %H:%M:%S.%f",
+          convertTimestampOr(sanitizedInput, TIMESTAMP_REGEX_FULL_2, "%Y-%m-%dT%H:%M:%S.%f",
```
jlowe, line 808: Do we have to do an extra regex pass?  IIRC the regex passes are really expensive, and I'm curious how it compares to the cost of `isValidTimestamp` and `asTimestampMicrosceonds`.  I'm wondering if it would be overall cheaper to use the original regex and ifElse the two valid+conversion formats corresponding to that regex instead of adding another regex pass.

andygrove: Yes, that makes sense. I'll refactor this part to avoid the additional regex call.

andygrove: I pushed one commit that refactors this to use a single regex and then pushed another commit to remove the regex completely for this full timestamp case and tests pass with the current test values.

@revans2 would you have any concerns about removing regex use in this case?

revans2: Once we have sanitized the inputs the regular expressions are redundant and should be removed.

revans2: I think we can probably remove even more regular expressions from the code.

andygrove: I looked at this some more and remembered why we have the regex here. We need to make sure there is no additional text after the sanitized date/timestamp string. I added an additional test case to demonstrate this for timestamps and reverted the change to remove the regex for this case.

I do think could remove more of these regexes by checking the length of the sanitized strings (in conjunction with calling cuDF `isTimestamp`) rather than using regex and have filed an issue for this: https://github.com/NVIDIA/spark-rapids/issues/1738

jlowe (approved):

andygrove:
build

Issue #1642, opened 2021-02-01
[BUG] TPC-DS-like query 67 at scale=3TB fails with OOM


revans2:
This is with 8 executors (each with an A100 with 40GB) and 4 concurrent tasks (4 cores/exec)

There are other pieces that are failing with OOM (like the filter), but I believe this to join that got too big.

```
java.lang.OutOfMemoryError: Could not allocate native memory: std::bad_alloc: RMM failure at:/home/abellina/anaconda3/envs/cudf_dev_0.18_11/include/rmm/mr/device/detail/arena.hpp:382: Maximum pool size exceeded
        at ai.rapids.cudf.Table.concatenate(Native Method)
        at ai.rapids.cudf.Table.concatenate(Table.java:1194)
        at com.nvidia.spark.rapids.ConcatAndConsumeAll$.buildNonEmptyBatch(GpuCoalesceBatches.scala:55)
        at com.nvidia.spark.rapids.GpuCoalesceIterator.concatAllAndPutOnGPU(GpuCoalesceBatches.scala:540)
        at com.nvidia.spark.rapids.AbstractGpuCoalesceIterator.$anonfun$next$3(GpuCoalesceBatches.scala:325)
        at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
        at com.nvidia.spark.rapids.Arm.withResource$(Arm.scala:26)
        at com.nvidia.spark.rapids.AbstractGpuCoalesceIterator.withResource(GpuCoalesceBatches.scala:134)
        at com.nvidia.spark.rapids.AbstractGpuCoalesceIterator.$anonfun$next$1(GpuCoalesceBatches.scala:324)
        at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
        at com.nvidia.spark.rapids.Arm.withResource$(Arm.scala:26)
        at com.nvidia.spark.rapids.AbstractGpuCoalesceIterator.withResource(GpuCoalesceBatches.scala:134)
        at com.nvidia.spark.rapids.AbstractGpuCoalesceIterator.next(GpuCoalesceBatches.scala:257)
        at com.nvidia.spark.rapids.AbstractGpuCoalesceIterator.next(GpuCoalesceBatches.scala:134)
        at com.nvidia.spark.rapids.GpuColumnarBatchSorter$$anon$1.loadNextBatch(GpuSortExec.scala:147)
[...]

PR #1719, opened 2021-02-12
Add in out of core sort

### LINKED ISSUE #19: [FEA] Out of core sorts
Currently when sorting the plugin has to hold all of the data for a single partition in memory.  It would be great if we supported spilling some of that data to host memory/disk so if someone had a horribly skewed sort or something that the plugin could still work, even if it was not ideal.

revans2:
This depends on https://github.com/rapidsai/cudf/pull/7381

The code is functionally complete and does a lot of things.

It improves the memory utilization of non-out of core sorting by not gathering columns that will not be output.
It adds in support for spill metrics so sort can report how many bytes were spilled
It fixes a bug in hash aggregate that made first and last not output the correct answer if multiple batches came in.
and it adds in support for out of core sort.

I still need to do some documentation work for sort.
I have been running some benchmarks and out of core is noticeably slower than the other version, but it can sort massive amounts of data. (I was able to sort 100GB in a single task on a 16GB V100), The performance was also fairly awful, but it was spilling.

There are a few more optimizations I want to try and play around with, but I was also hoping for some feedback/ideas on how to improve the performance.

If the code is hard to follow please let me know so I can add in more comments.

I am really interested in ideas for how to split up the batches to combine later on, because this is key to the algorithms performance. Right now I split incoming batches up into ones that are about `target size/8`.  This means two things. First, if you want to sort data that is smaller than `target size/8` or with default configs that is `2GB/8` = `256MB` then no merge sort will be done. The overhead is in putting the data into the spill buffer and pulling it back out. I need to do some profiling on this and clean up but ideally it should match the original time. The second thing this means is that if we get less than 8 times the target size (16GB) of input data then we can likely sort all of the data in just 2 passes. One pass to sort it and store it, and the second pass to merge sort and output results.

If we could be more dynamic in this we could probably get better numbers in both cases, but I am not sure of a good algorithm to do that.

This addresses part of #1642 it should fix it until we start to support rank, but then we need to look at out of core window functions.

This fixes #19


revans2:
This should now be ready for review.  The dependency was merged into cudf 0.19 this morning, but will probably not be all the way through CI until tomorrow.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SortUtils.scala
```diff
@@ -46,4 +56,211 @@ object SortUtils {
...
+            BoundReference(index, so.dataType, so.nullable),
+            so.direction, so.nullOrdering)
+      }
+    }
+    (needsComputation.toArray, cudfOrdering.toArray, cpuOrdering.toArray)
+  }
+
+  lazy val computationSchema: Seq[Attribute] =
+    needsComputation.map { so =>
+      AttributeReference(s"SORT_TMP", so.dataType, so.nullable)()
```
jlowe, line 120: I'm not following the use of `SORT_TMP`.  A block comment at the top of the class describing at a high level what the algorithm is and how extra columns are used would help.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -16,254 +16,459 @@
...
-  override def convertToGpu(): GpuExec =
+  override def convertToGpu(): GpuExec = {
     GpuSortExec(childExprs.map(_.convertToGpu()).asInstanceOf[Seq[SortOrder]],
       sort.global,
-      childPlans(0).convertIfNeeded())
-
-  override def tagPlanForGpu(): Unit = {
-    if (GpuOverrides.isAnyStringLit(sort.sortOrder)) {
-      willNotWorkOnGpu("string literal values are not supported in a sort")
-    }
```
jlowe, line 46: Does this limitation no longer apply?  I didn't see where this check moved if it does.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -16,254 +16,459 @@
...
-  private val numSortCols = sortOrder.length
-  private val totalTimeMetric : Option[GpuMetric] = initMetric(TOTAL_TIME)
-  private val outputBatchesMetric : Option[GpuMetric] = initMetric(NUM_OUTPUT_BATCHES)
-  private val outputRowsMetric : Option[GpuMetric] = initMetric(NUM_OUTPUT_ROWS)
-
-  private def initMetric(metricName: String): Option[GpuMetric] = if (shouldUpdateMetrics) {
-    Some(exec.gpuLongMetric(metricName))
-  } else {
-    None
+case class OutOfCoreBatch(buffer: SpillableColumnarBatch, row: UnsafeRow) extends AutoCloseable {
```
jlowe, line 153: A one-liner comment explaining what this is for would help.  I get the `buffer` part, `row` is a little less clear when browsing the code.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -16,254 +16,459 @@
...
-    new Iterator[ColumnarBatch] {
-      var resultBatch: Option[ColumnarBatch] = None
+  def isEmpty: Boolean = pending.isEmpty
 
-      TaskContext.get().addTaskCompletionListener[Unit](_ => closeBatch())
+  override def close(): Unit = pending.forEach(_.close())
+}
 
-      private def closeBatch(): Unit = resultBatch.foreach(_.close())
+case class GpuOutOfCoreSortIterator(
```
jlowe, line 187: A block comment here explaining the high-level algorithm would help

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -16,254 +16,459 @@
...
         }
-      }
 
-      override def hasNext: Boolean = {
-        if (resultBatch.isDefined) {
-          true
+      withResource(sortedTbl.contiguousSplit(splitIndexes: _*)) { split =>
+        memUsed += split.map(_.getBuffer.getLength).sum
+        val stillPending = if (sortedOffset >= 0) {
+          withResource(new NvtxRange("Store Sorted Table", NvtxColor.DARK_GREEN)) { _ =>
```
jlowe, line 287: Is this block expected to take a significant amount of time?  I'm surprised to see an NVTX range here, as I'm seeing only some object creation and adding to a linked list.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -16,254 +16,459 @@
...
+import org.apache.spark.sql.catalyst.expressions.codegen.LazilyGeneratedOrdering
 import org.apache.spark.sql.catalyst.plans.physical.{Distribution, OrderedDistribution, Partitioning, UnspecifiedDistribution}
 import org.apache.spark.sql.execution.{SortExec, SparkPlan, UnaryExecNode}
-import org.apache.spark.sql.types.{ArrayType, DataType, MapType, StructType}
+import org.apache.spark.sql.rapids.execution.TrampolineUtil
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
+sealed trait SortExecType extends Serializable
+
+object OutOfCoreSort extends SortExecType
```
abellina, line 39: A comment on each of these types would be good. I.e. I am not 100% sure why we wouldn't always want `OutOfCoreSort` though I believe that's to do with unstable sort.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -16,254 +16,459 @@
...
+    val totalTime = gpuLongMetric(TOTAL_TIME)
+    val outputBatch = gpuLongMetric(NUM_OUTPUT_BATCHES)
+    val outputRows = gpuLongMetric(NUM_OUTPUT_ROWS)
+    val outOfCore = sortType == OutOfCoreSort
+    child.executeColumnar().mapPartitions { cbIter =>
+      if (outOfCore) {
+        val cpuOrd = new LazilyGeneratedOrdering(sorter.cpuOrdering)
+        val spillCallback = GpuMetric.makeSpillCallback(allMetrics)
+        val iter = GpuOutOfCoreSortIterator(cbIter, sorter, cpuOrd,
+          // bad things can happen if the target size is really small
```
abellina, line 118: are the bad things something documented later? :)

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
abellina: I think this range is mainly for visibility.

abellina (changes requested):
I am mostly checking with `GpuSortExec` now.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -16,254 +16,459 @@
...
 
   // sort performed is local within a given partition so will retain
   // child operator's partitioning
   override def outputPartitioning: Partitioning = child.outputPartitioning
 
   override def requiredChildDistribution: Seq[Distribution] =
-    if (global) OrderedDistribution(sparkSortOrder) :: Nil else UnspecifiedDistribution :: Nil
+    if (global) OrderedDistribution(sortOrder) :: Nil else UnspecifiedDistribution :: Nil
 
   // Eventually this might change, but for now we will produce a single batch, which is the same
```
abellina, line 81: comment needs updating after the change.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -16,254 +16,459 @@
...
-        if (batchIter.hasNext) {
-          if (singleBatchOnly && haveSortedBatch) {
-            throw new UnsupportedOperationException("Expected single batch to sort")
+  private final def splitAfterSortAndSave(sortedTbl: Table, sortedOffset: Int = -1): Unit = {
+    var memUsed: Long = GpuColumnVector.getTotalDeviceMemoryUsed(sortedTbl)
+    // We need to figure out how to split up the data into reasonable batches. We could try and do
+    // something really complicated and figure out how much data get per batch, but in practice
+    // we really only expect to see one or two batches worth of data come in, so lets optimize
+    // for that case and set the targetBatchSize to always be 1/8th the targetSize.
+    val targetBatchSize = targetSize / 8
```
abellina, line 236: In the poorly configured case, `targetSize` could be 16KB. So this means we'd do 2KB batches here?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -16,254 +16,459 @@
...
           }
-        } finally {
-          nvtxRange.close()
         }
-      }
 
-      override def hasNext: Boolean = {
-        if (resultBatch.isDefined) {
-          true
+      withResource(sortedTbl.contiguousSplit(splitIndexes: _*)) { split =>
```
abellina, line 284:
```suggestion
      withResource(sortedTbl.contiguousSplit(splitIndexes: _*)) { splits =>
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -16,254 +16,459 @@
...
+      // of an impact, but if we wanted to be exact we would look at those too.
+      peakMemory = Math.max(peakMemory, memUsed)
+      withResource(mergedBatch) { mergedBatch =>
+        // First we want figure out what is fully sorted from what is not
+        val sortSplitOffset = if (pending.isEmpty) {
+          // No need to split it
+          mergedBatch.numRows() - 1
+        } else {
+          // The data is only fully sorted if there is nothing pending that is smaller than it
+          // so get the nest smallest row that is pending
```
abellina, line 389:
```suggestion
          // so get the next smallest row that is pending
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -16,254 +16,459 @@
...
-  } else {
-    None
+case class OutOfCoreBatch(buffer: SpillableColumnarBatch, row: UnsafeRow) extends AutoCloseable {
+  override def close(): Unit = buffer.close()
+}
+
+class Pending(cpuOrd: LazilyGeneratedOrdering) extends AutoCloseable {
+  private val pending = new PriorityQueue[OutOfCoreBatch](new Comparator[OutOfCoreBatch]() {
+    override def compare(a: OutOfCoreBatch, b: OutOfCoreBatch): Int =
+      cpuOrd.compare(a.row, b.row)
```
abellina, line 160: Could `row` be renamed to something like `cutOff` or `lower`, if I understand this correctly?

docs/configs.md
```diff
@@ -78,6 +78,7 @@ Name | Description | Default Value
...
 <a name="sql.metrics.level"></a>spark.rapids.sql.metrics.level|GPU plans can produce a lot more metrics than CPU plans do. In very large queries this can sometimes result in going over the max result size limit for the driver. Supported values include DEBUG which will enable all metrics supported and typically only needs to be enabled when debugging the plugin. MODERATE which should output enough metrics to understand how long each part of the query is taking and how much data is going to each part of the query. ESSENTIAL which disables most metrics except those Apache Spark CPU plans will also report or their equivalents.|MODERATE
+<a name="sql.outOfCoreSort.enabled"></a>spark.rapids.sql.outOfCoreSort.enabled|enable or disable out of core sorting. Out of core sorting allows for sorting large skewed data sets, but is not a stable sort. Apache Spark does not guarantee a stable sort in most cases, except when coalescing all data to a single task and sorting there. In that case we will produce different results.|true
```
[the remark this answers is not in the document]
revans2: I personally think that "out of core" or "external memory" would be the correct term to use. Spillable is very Spark specific, but that is the audience we are going for so if you insist I will change it. The only reason this is configurable is because of that one small use case where the stable sort matters. I have not seen enough of a performance difference to justify turning it off for performance. 

Perhaps we should just call it stable vs unstable sort, and in the future we can add in an out of core stable sort implementation.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala
```diff
@@ -57,6 +57,9 @@ object GpuMetric {
   val BUILD_DATA_SIZE = "buildDataSize"
   val BUILD_TIME = "buildTime"
   val STREAM_TIME = "streamTime"
+  val SPILL_AMOUNT = "spillData"
```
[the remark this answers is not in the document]
revans2: Which do you want it to be?  Right now it is all data spilled to host memory or disk.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala
```diff
@@ -199,6 +223,12 @@ trait GpuExec extends SparkPlan with Arm {
 
   lazy val additionalMetrics: Map[String, GpuMetric] = Map.empty
 
+  protected def spillMetrics: Map[String, GpuMetric] = Map(
+    SPILL_AMOUNT -> createSizeMetric(ESSENTIAL_LEVEL, DESCRIPTION_SPILL_AMOUNT),
+    SPILL_AMOUNT_DISK -> createSizeMetric(MODERATE_LEVEL, DESCRIPTION_SPILL_AMOUNT_DISK),
+    SPILL_AMOUNT_HOST -> createSizeMetric(MODERATE_LEVEL, DESCRIPTION_SPILL_AMOUNT_HOST)
```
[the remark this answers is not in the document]
revans2: If the GPU OOM spill code already uses them, then wouldn't we be double counting because the the callback also happens because of a GPU OOM.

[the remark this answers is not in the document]
jlowe: Yeah, I was originally thinking we should just let the OOM spill code handle this, but that doesn't tie it to a specific exec node which we want here.

docs/configs.md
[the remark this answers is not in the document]
jlowe: > if you insist I will change it.

I'm not insisting, just pointing out that the term will not be familiar to many Spark users reading the documentation and it's not really defined anywhere.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala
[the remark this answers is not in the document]
jlowe: > it is all data spilled to host memory or disk.

That's fine, the description should be updated to say this or something similar so the user knows what is being measured.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBuffer.scala
```diff
@@ -72,6 +72,11 @@ trait RapidsBuffer extends AutoCloseable {
   /** The storage tier for this buffer */
   val storageTier: StorageTier
 
+  /**
+   * A callback function that can be used for metrics when a buffer is spilled
+   */
+  val spillCallback: (String, Long) => Unit
```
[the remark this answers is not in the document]
revans2: I started out with not putting defaults everywhere because I wanted to be sure that I had plumbed it through properly in all places with compile errors.  If you want me to move it to RapidsBufferId I can do it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
revans2: Yup it didn't apply a while ago, but we never got around to removing it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
revans2: I added it when debugging what was taking a lot of time, and never removed it.  It turns out that the time was being taken by the next range where we add things to pending, because pending will do a lazy compile of the comparator which I was trying to track down. I am happy to remove this one if we want to.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
revans2: Yes, but the only time that would really happen is unit tests.  We eventually might want to prevent people from shooting themselves in the foot, but for some reason I am feeling optimistic about human nature today, and I don't want to have to figure out how to make the tests that set it to 0 still work if we disallow setting it to anything smaller than say a few hundred MB.

revans2:
build

revans2:
@jlowe and @abellina I think I have addressed most of your review comments. There are still a few things left that I asked questions on. I think I will switch the configs over to talk about stable vs non-stable sorts instead of out of core. But I want to hold off to be sure that there are not other things found/decided on.

revans2:
build

revans2:
@jlowe @abellina I think I have addressed all of your review comments. Please take another look

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SortUtils.scala
jlowe: Would still like to see a block comment on the class describing the high-level algorithm being performed (i.e.: mentioning the fact that there are extra columns, how they're computed and used, maybe walk through an example, etc.)

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
jlowe: Unless there's a compelling reason for them to remain, I'd like to see the extremely small NVTX ranges like this one removed since they can distort profiles and probably aren't that useful in practice.  If these tiny ranges serve a useful purpose, may be worth commenting so someone doesn't come along later and remove them.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBuffer.scala
```diff
@@ -72,6 +79,19 @@ trait RapidsBuffer extends AutoCloseable {
...
+   * be used for metrics gathering in parts of the GPU plan that can spill. No GPU memory
+   * should ever be allocated from this callback, blocking in this function is strongly
+   * discouraged. It should be as light weight as possible. It takes three arguments
+   * <ul>
+   * <li><code>from</code> the storage tier the data is being spilled from.</li>
+   * <li><code>to</code> the storage tier the data is being spilled to.</li>
+   * <li><code>amount</code> the amount of data in bytes that is spilled.</li>
+   * </ul>
+   */
+  val spillCallback: (StorageTier, StorageTier, Long) => Unit
```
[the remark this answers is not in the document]
revans2: I don't default it here because the only time it needs to be defaulted is in the degenerate case and tests. I would rather be cautious and force someone adding a new buffer to decide if they need it or not, but with RapidsBufferBase it may be a moot point.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
abellina: Like with metrics, and log lines, should we have levels for nvtx ranges? This is a separate discussion but since we are talking about it, I can spawn off an issue if that's interesting.

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
revans2: If I am looking at NVTX everything is already debug level. The main reason to remove them is because they are so small all the time that showing them has no useful value. I just didn't remove them earlier when I was debugging, and Jason did the right thing to ask "why is this section of code taking long enough that we need to pay attention to it?" Because if it did take that long then we have some optimizing to do.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SortUtils.scala
```diff
@@ -46,4 +59,265 @@ object SortUtils {
...
+ * strings by their length instead of in lexicographical order. Because cudf does not support this
+ * directly we instead go through the SortOrder instances that are a part of this sorter and find
+ * the ones that require computation. We then do the sort in a few stages first we compute any
+ * needed columns from the SortOrder instances that require some computation, and add them to the
+ * original batch.  The method `appendProjectedColumns` does this. This then provides a number of
+ * methods that can be used to operate on a batch that has these new columns added to it. These
+ * include sorting, merge sorting, and finding bounds. These can be combined in various ways to
+ * do different algorithms. When you are done with these different operations you can drop the
+ * temporary columns that were added, just for computation, using `removeProjectedColumns`.
+ * Some times you may want to pull data back to the CPU and sort rows there too. we provide
```
jlowe, line 92:
```suggestion
 * Some times you may want to pull data back to the CPU and sort rows there too. We provide
```

jlowe (approved):
One small typo that doesn't really matter.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SortUtils.scala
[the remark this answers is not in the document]
abellina: @revans2 this one is a small nit, but I believe it was missed. Not a blocker.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
abellina: > for some reason I am feeling optimistic about human nature today

:)

revans2:
I think I got the final nits from everyone.  Thanks for the reviews

revans2:
build

jlowe (approved):

abellina (approved):

PR #1725, opened 2021-02-13
Branch 0.4 FAQ reorg

sameerz:
Reorganization of FAQ, fixed links, fixed capitalization of proper nouns.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -29,7 +29,7 @@ with most cloud service providers to set up testing and validation on their dist
 ### What CUDA versions are supported?
 
 CUDA 10.1, 10.2 and 11.0 are currently supported, but you need to download the cudf jar that 
-corresponds to the version you are using. Please look [here][download.md] for download 
+corresponds to the version you are using. Please look [here](download.md) for download 
 links for the latest release.
 
 ### What is the right hardware setup to run GPU accelerated Spark?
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -5,17 +5,76 @@ nav_order: 11
 ---
 # Frequently Asked Questions
 
-### Why does `explain()` show that the GPU will be used even after setting `spark.rapids.sql.enabled` to `false`?
+* TOC
+{:toc}
```
jlowe, line 9: Note that this causes an undesirable artifact when rendering with Markdown.  See https://github.com/NVIDIA/spark-rapids/blob/db7831a32c57df8cd91dbae7d88f49523401d7a8/docs/FAQ.md

revans2: I personally think that is fine.  We already have issue with a table at the beginning of each markdown file so I am not too concerned.

docs/FAQ.md
```diff
@@ -5,17 +5,76 @@ nav_order: 11
...
+
+### What CUDA versions are supported?
+
+CUDA 10.1, 10.2 and 11.0 are currently supported, but you need to download the cudf jar that 
+corresponds to the version you are using. Please look [here](download.md) for download 
+links for the latest release.
+
+### What is the right hardware setup to run GPU accelerated Spark?
+
+Reference architectures should be available around Q1 2021.
```
revans2, line 37: It is Q1 2020. Do we have an update on this?

revans2 (approved):
Just a nit that we might want to think about for a follow on if we don't have something to say right now.

docs/FAQ.md
sameerz: No update yet, we are working on the hardware architectures and those will be out late Q1 or early Q2.

sameerz:
build

jlowe (approved):

PR #1741, opened 2021-02-16
Reduce regex use in CAST

andygrove:
This replaces four uses of regex in CAST with checks for expected length instead.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -770,7 +769,7 @@ case class GpuCast(
     }
 
     /** This method does not close the `input` ColumnVector. */
-    def convertTimestampFullOr(
+    def convertFullTimestampOr(
         input: ColumnVector,
         orElse: ColumnVector): ColumnVector = {
```
jlowe, line 774: This has the same issue of failing to close `orElse`if some earlier processing throws.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -728,28 +733,22 @@ case class GpuCast(
...
         }
       }
     }
 
     /** This method does not close the `input` ColumnVector. */
-    def convertTimestampOr(
+    def convertVarLenTimestampOr(
         input: ColumnVector,
         regex: String,
         cudfFormat: String,
```
jlowe, line 754: Same `orElse` leaks on early throw here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -783,9 +782,15 @@ case class GpuCast(
           isTimestamp1.or(isTimestamp2)
         }
       }
+
       val isValidTimestamp = withResource(isCudfMatch) { isCudfMatch =>
-        withResource(input.matchesRe(TIMESTAMP_REGEX_FULL)) { isRegexMatch =>
-          isCudfMatch.and(isRegexMatch)
+        val isValidLength = withResource(Scalar.fromInt(27)) { requiredLen =>
```
jlowe, line 787: It's not clear where `27` is coming from.  I suggest naming this constant as a `val` and placing it next to the format string(s) that it applies to.

jlowe (changes requested):

jlowe (approved):

andygrove:
build

Issue #1157, opened 2020-11-18
[BUG] Fix regression supporting to_date on GPU with Spark 3.1.0


andygrove:
**Describe the bug**
Prior to Spark 3.1.0, `to_date` gets translated to `CAST(unix_timestamp(expr, format, timezone) AS date)` which we support on GPU, but since 3.1.0 it instead gets translated to `CAST(gettimestamp(expr, format, timezone) AS date)` which we do not support on GPU.

**Steps/Code to reproduce bug**
Remove the `execsAllowedNonGpu` from the relevant tests in com.nvidia.spark.rapids.ParseDateTimeSuite

**Expected behavior**
We should support to_date on GPU with 3.1.0 by supporting `gettimestamp` on GPU.

tgravescs:
reference https://github.com/apache/spark/commit/75d342858a71293b7c2cbbd5cec2eb417ed3852c
this was an optimiation which really changed it from 
Cast(SecondsToTimestamp(UnixTimestamp(left, format)), DateType)) to
Cast(GetTimestamp(left, format), DateType))

I'm moving this to a P1 as I thought we had a lot of queries using to_date, if not we should move back to p2

tgravescs:
I think we have everything we need to implement this.

Note we could knock back down to P2  since spark.rapids.sql.castStringToTimestamp.enabled is off by default.

PR #1742, opened 2021-02-17
Fix regression supporting to_date with Spark-3.1

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/1157 . Since `GetTimestamp` is private in Spark code, creating a spark package.  


nartal1:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -477,11 +477,18 @@ abstract class GpuToTimestamp
...
-          longMicroSecs.div(downScaleFactor)
+    // Return Timestamp value if dataType it is expecting is of TimestampType
+    if (dataType.equals(TimestampType)) {
+      tmp
+    }
+    else {
+      withResource(tmp) { r =>
+        // The type we are returning is a long not an actual timestamp
+        withResource(Scalar.fromInt(downScaleFactor)) { downScaleFactor =>
+          withResource(tmp.asLongs()) { longMicroSecs => {
```
andygrove, line 488: nit: This usage of `withResource` has an additional set of `{` and `}` after `=>` which is inconsistent with the other lines and could be omitted

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -477,11 +477,18 @@ abstract class GpuToTimestamp
...
+            longMicroSecs.div(downScaleFactor)
+          }
+          }
```
jlowe, line 491: Whenever braces line up like this, something is off.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -477,11 +477,18 @@ abstract class GpuToTimestamp
...
-    withResource(tmp) { r =>
-      // The type we are returning is a long not an actual timestamp
-      withResource(Scalar.fromInt(downScaleFactor)) { downScaleFactor =>
-        withResource(tmp.asLongs()) { longMicroSecs =>
-          longMicroSecs.div(downScaleFactor)
+    // Return Timestamp value if dataType it is expecting is of TimestampType
+    if (dataType.equals(TimestampType)) {
+      tmp
+    }
+    else {
```
jlowe, line 484:
```suggestion
    } else {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
nartal1: Thanks for the tip :) . Will watch out going forward.

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/expressions/rapids/Timestamp.scala
```diff
@@ -0,0 +1,41 @@
...
+
+
+package org.apache.spark.sql.catalyst.expressions.rapids
+
+import com.nvidia.spark.rapids.{ExprChecks, ExprRule, GpuExpression, GpuOverrides, TypeEnum, TypeSig}
+
+import org.apache.spark.sql.catalyst.expressions.{Expression, GetTimestamp}
+import org.apache.spark.sql.rapids.{GpuGetTimestamp, UnixTimeExprMeta}
+
+object GetTimeStamp {
```
revans2, line 25: nit: The only thing in Timestamp.scala is this one object, should we make it match?

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/expressions/rapids/Timestamp.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  def getExprs: Map[Class[_ <: Expression], ExprRule[_ <: Expression]] = Seq(
+    GpuOverrides.expr[GetTimestamp](
```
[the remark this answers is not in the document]
nartal1: I was not including in 3.0.0 as we were not seeing failure. You are right, I missed the point where `GetTimestamp` could be used in `ParseToDate` and `ParseToTimestamp` .  Thanks for catching this. I will update the PR to include this `expr` in `GpuOverrides`  rather than in Shim.

nartal1:
build

revans2 (approved):

jlowe (dismissed):

nartal1:
Thanks @revans2 and @jlowe for the review.
@andygrove Could you please take another look as you had reviewed it earlier. Thanks!

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -477,11 +477,16 @@ abstract class GpuToTimestamp
...
-    withResource(tmp) { r =>
-      // The type we are returning is a long not an actual timestamp
-      withResource(Scalar.fromInt(downScaleFactor)) { downScaleFactor =>
-        withResource(tmp.asLongs()) { longMicroSecs =>
-          longMicroSecs.div(downScaleFactor)
+    // Return Timestamp value if dataType it is expecting is of TimestampType
+    if (dataType.equals(TimestampType)) {
+      tmp
+    } else {
+      withResource(tmp) { r =>
```
andygrove, line 484: nit: I don't see this `r` variable being used anywhere although it looks like `tmp` is referenced instead. This isn't incorrect but makes the code harder to read IMO. Perhaps just rename `r` to `tmp`?

andygrove (approved):
LGTM with one minor nit comment

nartal1:
build

andygrove (approved):

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/expressions/rapids/Timestamp.scala
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.spark.sql.catalyst.expressions.{Expression, GetTimestamp}
+import org.apache.spark.sql.rapids.{GpuGetTimestamp, UnixTimeExprMeta}
+
+object TimeStamp {
+
+  def getExprs: Map[Class[_ <: Expression], ExprRule[_ <: Expression]] = Seq(
+    GpuOverrides.expr[GetTimestamp](
+      "Gets timestamps from strings using given pattern.",
+      ExprChecks.binaryProjectNotLambda(TypeSig.TIMESTAMP, TypeSig.TIMESTAMP,
+        ("timeExp", TypeSig.STRING, TypeSig.STRING),
```
revans2, line 31: Actually Spark support more than just strings for `timeExp`, they also support Date and timestamp as input.

```
  override def inputTypes: Seq[AbstractDataType] =
    Seq(TypeCollection(StringType, DateType, TimestampType), StringType)
```

revans2: @nartal1 you merged it too fast.  This needs to be fixed

nartal1: I am sorry @revans2, missed this.  Will put up a follow on PR now.

PR #1750, opened 2021-02-17
Detect task failures in benchmarks

andygrove:
This is a **breaking change** to the JSON reporting of query times.

Previously, the `queryTime` field would contain `-1` if the query failed and `>= 0` if the query was successful.

With these changes, `queryTime` will always be the elapsed time and there is a new `queryStatus` field which will have one of the following values: `Completed`, `CompletedWithTaskFailures`, or `Failed`.

### DIFF
--- a/integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/common/BenchUtils.scala
+++ b/integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/common/BenchUtils.scala
@@ -168,7 +168,6 @@ object BenchUtils {
 
     var df: DataFrame = null
     val queryTimes = new ListBuffer[Long]()
-
     for (i <- 0 until iterations) {
       spark.sparkContext.setJobDescription(s"Benchmark Run: query=$queryDescription; iteration=$i")
       
--- a/integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/common/BenchUtils.scala
+++ b/integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/common/BenchUtils.scala
@@ -27,7 +27,8 @@ import org.json4s.DefaultFormats
 import org.json4s.jackson.JsonMethods.parse
 import org.json4s.jackson.Serialization.writePretty
 
-import org.apache.spark.{SPARK_BUILD_USER, SPARK_VERSION}
+import org.apache.spark.{SPARK_BUILD_USER, SPARK_VERSION, Success, TaskEndReason}
+import org.apache.spark.scheduler.{SparkListener, SparkListenerTaskEnd}
 import org.apache.spark.sql.{DataFrame, Row, SaveMode, SparkSession}
 import org.apache.spark.sql.execution.{InputAdapter, QueryExecution, SparkPlan, WholeStageCodegenExec}
 import org.apache.spark.sql.execution.adaptive.{AdaptiveSparkPlanExec, QueryStageExec}
@@ -167,7 +168,11 @@ object BenchUtils {
 
     var df: DataFrame = null
     val queryTimes = new ListBuffer[Long]()
+
+
     for (i <- 0 until iterations) {
+      val taskFailureListener = new TaskFailureListener
+      spark.sparkContext.addSparkListener(taskFailureListener)
       spark.sparkContext.setJobDescription(s"Benchmark Run: query=$queryDescription; iteration=$i")
       
       // cause Spark to call unregisterShuffle
@@ -197,6 +202,12 @@ object BenchUtils {
             ensureValidColumnNames(df).write.mode(mode).options(options).orc(path)
           case WriteParquet(path, mode, options) =>
             ensureValidColumnNames(df).write.mode(mode).options(options).parquet(path)
+        }
+
+        if (taskFailureListener.taskFailures.nonEmpty) {
+          throw new RuntimeException(
+            s"${taskFailureListener.taskFailures.size} task failure(s) detected. " +
+                s"First failure: ${taskFailureListener.taskFailures.head}")
         }
 
         val end = System.nanoTime()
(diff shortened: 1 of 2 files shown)


abellina (dismissed):

abellina:
LGTM!

andygrove:
build

tgravescs:
I was actually wanting this today, so its great to see this. Is there a way we can just mark it special for this case - passed but with failures so probably affected overall time - or what does mark the query as failed mean? It probably gets complicated talking when you want to graph it. Does it fail but still report query Time?

abellina:
Maybe an option "--fail-on-task-error" defaulting to true?

andygrove:
> I was actually wanting this today, so its great to see this. Is there a way we can just mark it special for this case - passed but with failures so probably affected overall time - or what does mark the query as failed mean? It probably gets complicated talking when you want to graph it. Does it fail but still report query Time?

I'll look at this before merging this PR. It would be good to report time and status separately perhaps (status = Completed, Failed, CompletedWithTaskFailures)?

tgravescs:
I think separate makes sense

andygrove:
@abellina @tgravescs This is ready for another review.

abellina (approved):

andygrove:
build

integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/common/BenchUtils.scala
```diff
@@ -202,25 +210,37 @@ object BenchUtils {
...
         queryTimes.append(elapsed)
-        println(s"$logPrefix Iteration $i took $elapsed msec.")
+
+        val status = if (taskFailureListener.taskFailures.isEmpty) {
+          STATUS_COMPLETED
+        } else {
+          // add the first task failure to the list of exceptions in the report
+          exceptions.append(taskFailureListener.taskFailures.head.toString)
+          STATUS_COMPLETED_WITH_TASK_FAILURES
+        }
```
gerashegalov, line 220: nit:
```
val failureOpt = taskFailureListener.taskFailures.headOption
val status = failureOpt.map(_ =>  STATUS_COMPLETED_WITH_TASK_FAILURES).getOrElse(STATUS_COMPLETED)
failureOpt.foreach(exceptions.add(_.toString))
```

gerashegalov (dismissed):
LGTM

integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/common/BenchUtils.scala
```diff
@@ -202,25 +210,37 @@ object BenchUtils {
         val end = System.nanoTime()
         val elapsed = NANOSECONDS.toMillis(end - start)
         queryTimes.append(elapsed)
-        println(s"$logPrefix Iteration $i took $elapsed msec.")
+
+        val status = if (taskFailureListener.taskFailures.isEmpty) {
+          STATUS_COMPLETED
+        } else {
+          // add the first task failure to the list of exceptions in the report
+          exceptions.append(taskFailureListener.taskFailures.head.toString)
```
tgravescs, line 218: do we need to/want to change the suffix on the file at all?   I think adding this will make the file name show up with -failed right?

tgravescs: ie line 239 below

integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/common/BenchUtils.scala
andygrove: @gerashegalov I went ahead and implemented this suggestion since it is much cleaner. Could you re-approve, please.

andygrove:
build

integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/common/BenchUtils.scala
```diff
@@ -202,25 +210,35 @@ object BenchUtils {
         val end = System.nanoTime()
         val elapsed = NANOSECONDS.toMillis(end - start)
         queryTimes.append(elapsed)
-        println(s"$logPrefix Iteration $i took $elapsed msec.")
+
+        val failureOpt = taskFailureListener.taskFailures.headOption
+        val status = failureOpt.map(_ =>  STATUS_COMPLETED_WITH_TASK_FAILURES)
+            .getOrElse(STATUS_COMPLETED)
+        failureOpt.foreach(failure => exceptions.append(failure.toString))
```
tgravescs, line 217: sorry if this comment comes through twice, you changed as I was commenting:

do we need to/want to change the suffix on the file at all? I think adding this will make the file name show up with -failed right?  line 257 below

andygrove: That's a good point. Now that we are recording the status per run it probably doesn't make sense to have the convention of adding `-failed` to the filename. What do you think @abellina (I think you may have asked for this feature?).

andygrove: I went ahead and pushed a change to remove the `-failed` suffix.

andygrove:
build

tgravescs (approved):

integration_tests/src/main/scala/com/nvidia/spark/rapids/tests/common/BenchUtils.scala
tgravescs: I'm fine either way, if we keep it might want to change it to not be failed if query succeeded but had failures

gerashegalov (approved):

Issue #1736, opened 2021-02-16
Random seed changed to handle expression as seed in Spark 3.1


razajafri:
**Is your feature request related to a problem? Please describe.**
Spark 3.1 can now handle seed as an expression constant values

**Describe the solution you'd like**
We should match the behavior in GpuRand

**Additional context** 
For additional context, look at this [change](https://github.com/apache/spark/commit/9268392b95)

razajafri:
We can't support this change as we currently rely on the `org.apache.spark.sql.catalyst.optimizer.ConstantFolding` rule. If we exclude this rule, a lot of other things won't work in the plugin.

PR #1751, opened 2021-02-17
Exclude foldable expressions from GPU if constant folding is disabled

razajafri:
This PR adds documentation that we don't support excluding the rule `org.apache.spark.sql.catalyst.optimizer.ConstantFolding`

This came up as a result of trying to support #1736 which was because of a change in Spark 3.1

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -352,4 +352,10 @@ Casting from string to timestamp currently has the following limitations.
 - <a name="Footnote1"></a>[1] The timestamp portion must be complete in terms of hours, minutes, seconds, and
  milliseconds, with 2 digits each for hours, minutes, and seconds, and 6 digits for milliseconds. 
  Only timezone 'Z' (UTC) is supported. Casting unsupported formats will result in null values. 
- 
+
+### Constant Folding
+
+ConstantFolding is an operator optimization rule in Catalyst that replaces expressions that can
+be statically evaluated with their equivalent literal values. We don't support excluding the rule 
+`org.apache.spark.sql.catalyst.optimizer.ConstantFolding` as we rely on this rule internally 
+and excluding it will cause some of our assumptions to fail. 
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -356,6 +356,6 @@ Casting from string to timestamp currently has the following limitations.
 ### Constant Folding
 
 ConstantFolding is an operator optimization rule in Catalyst that replaces expressions that can
-be statically evaluated with their equivalent literal values. The RAPIDS Accelerator doesn't
-support excluding the rule `org.apache.spark.sql.catalyst.optimizer.ConstantFolding` as we rely on 
-this rule internally and excluding it may cause the query to produce undesired results
+be statically evaluated with their equivalent literal values. The RAPIDS Accelerator relies
+on constant folding and parts of the query will not be accelerated if 
+`org.apache.spark.sql.catalyst.optimizer.ConstantFolding` is excluded as a rule
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -356,6 +356,6 @@ Casting from string to timestamp currently has the following limitations.
 ### Constant Folding
 
 ConstantFolding is an operator optimization rule in Catalyst that replaces expressions that can
-be statically evaluated with their equivalent literal values. We don't support excluding the rule 
-`org.apache.spark.sql.catalyst.optimizer.ConstantFolding` as we rely on this rule internally 
-and excluding it will cause some of our assumptions to fail. 
+be statically evaluated with their equivalent literal values. The RAPIDS Accelerator doesn't
+support excluding the rule `org.apache.spark.sql.catalyst.optimizer.ConstantFolding` as we rely on 
+this rule internally and excluding it may cause the query to produce undesired results
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -358,4 +358,4 @@ Casting from string to timestamp currently has the following limitations.
 ConstantFolding is an operator optimization rule in Catalyst that replaces expressions that can
 be statically evaluated with their equivalent literal values. The RAPIDS Accelerator relies
 on constant folding and parts of the query will not be accelerated if 
-`org.apache.spark.sql.catalyst.optimizer.ConstantFolding` is excluded as a rule
+`org.apache.spark.sql.catalyst.optimizer.ConstantFolding` is excluded as a rule.
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
@@ -1028,6 +1028,10 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
     entry.get(conf)
   }
 
+  def getStringConf(key: String): Option[String] = {
+    conf.get(key)
+  }
+
   lazy val rapidsConfMap: util.Map[String, String] = conf.filterKeys(
     _.startsWith("spark.rapids.")).asJava
 
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
@@ -1028,10 +1028,6 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
     entry.get(conf)
   }
 
-  def getStringConf(key: String): Option[String] = {
-    conf.get(key)
-  }
-
   lazy val rapidsConfMap: util.Map[String, String] = conf.filterKeys(
     _.startsWith("spark.rapids.")).asJava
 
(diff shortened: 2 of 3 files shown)


razajafri:
build

docs/compatibility.md
```diff
@@ -352,4 +352,10 @@ Casting from string to timestamp currently has the following limitations.
 - <a name="Footnote1"></a>[1] The timestamp portion must be complete in terms of hours, minutes, seconds, and
  milliseconds, with 2 digits each for hours, minutes, and seconds, and 6 digits for milliseconds. 
  Only timezone 'Z' (UTC) is supported. Casting unsupported formats will result in null values. 
- 
+
+### Constant Folding
+
+ConstantFolding is an operator optimization rule in Catalyst that replaces expressions that can
+be statically evaluated with their equivalent literal values. We don't support excluding the rule 
+`org.apache.spark.sql.catalyst.optimizer.ConstantFolding` as we rely on this rule internally
```
jlowe, line 360: Rather than just squirreling this away in the corner of a doc most users will not read, why not also update the plugin to check this config setting when it starts to optimize a plan?  It could emit a warning and refuse to optimize the plan which will let the query run correctly, albeit completely on the CPU.

docs/compatibility.md
```diff
@@ -352,4 +352,10 @@ Casting from string to timestamp currently has the following limitations.
...
  milliseconds, with 2 digits each for hours, minutes, and seconds, and 6 digits for milliseconds. 
  Only timezone 'Z' (UTC) is supported. Casting unsupported formats will result in null values. 
- 
+
+### Constant Folding
+
+ConstantFolding is an operator optimization rule in Catalyst that replaces expressions that can
+be statically evaluated with their equivalent literal values. We don't support excluding the rule 
+`org.apache.spark.sql.catalyst.optimizer.ConstantFolding` as we rely on this rule internally 
+and excluding it will cause some of our assumptions to fail.
```
jlowe, line 361: I suggest rewording this to remove the assumptions part and simply say the query may execute incorrectly (or even crash if that's possible?)

docs/compatibility.md
```diff
@@ -352,4 +352,10 @@ Casting from string to timestamp currently has the following limitations.
 - <a name="Footnote1"></a>[1] The timestamp portion must be complete in terms of hours, minutes, seconds, and
  milliseconds, with 2 digits each for hours, minutes, and seconds, and 6 digits for milliseconds. 
  Only timezone 'Z' (UTC) is supported. Casting unsupported formats will result in null values. 
- 
+
+### Constant Folding
+
+ConstantFolding is an operator optimization rule in Catalyst that replaces expressions that can
+be statically evaluated with their equivalent literal values. We don't support excluding the rule
```
jlowe, line 359: Suggest changing "We" to "The RAPIDS Accelerator" in this paragraph.

docs/compatibility.md
revans2: The issue that would happen is that we get two literal values in say a GPU Add and we try to run it and it blows up. I agree that we probably want to disable the plugin, but at least output a warning if we see that ConstantFolding has been disabled.

As an alternative when replacing the plan with a GPU enabled version we could look for all foldable CPU operators, and if they are not on Literal values we CANNOT replace them.  Because then at least we could not crash, and we would honor what was asked by the end user, to disable constant folding.

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1028,6 +1028,10 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
     entry.get(conf)
   }
 
+  def getStringConf(key: String): Option[String] = {
+    conf.get(key)
+  }
```
jlowe, line 1033: I don't think we want to encourage code looking for confs in RapidsConf that aren't confs explicitly supported by it.  There are parts of the code that explicitly filters the Spark configs to `spark.rapids.*` before passing to code that will build a RapidsConf.

If we want to lookup a SQLConf key then we should use a SQLConf to do so.  `SQLConf.get` can be used to obtain the current active instance.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -244,6 +245,18 @@ abstract class RapidsMeta[INPUT <: BASE, BASE, OUTPUT <: BASE](
       }
     }
 
+    val isExp = wrapped.isInstanceOf[Expression]
```
jlowe, line 248: I think this makes more sense to be in `BaseExprMeta` where we know we have an expression.  IIUC all expressions should be using it.

docs/compatibility.md
```diff
@@ -352,4 +352,10 @@ Casting from string to timestamp currently has the following limitations.
...
  milliseconds, with 2 digits each for hours, minutes, and seconds, and 6 digits for milliseconds. 
  Only timezone 'Z' (UTC) is supported. Casting unsupported formats will result in null values. 
- 
+
+### Constant Folding
+
+ConstantFolding is an operator optimization rule in Catalyst that replaces expressions that can
+be statically evaluated with their equivalent literal values. The RAPIDS Accelerator doesn't
+support excluding the rule `org.apache.spark.sql.catalyst.optimizer.ConstantFolding` as we rely on 
+this rule internally and excluding it may cause the query to produce undesired results
```
jlowe, line 361: This is no longer true after the other changes in this file.  Instead this should mention parts of the query may not run on the GPU without constant folding optimization enabled.

jlowe (commented):
The PR headline should be updated to reflect this is no longer just about updating docs.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
revans2: Also we have to allow Literal values to pass through this OK.  Literal values are foldable, but we support them

razajafri: isn't that covered by `exp.foldable && !GpuOverrides.isLit(exp)`. So the condition is true only if its foldable and not a literal.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -244,6 +245,7 @@ abstract class RapidsMeta[INPUT <: BASE, BASE, OUTPUT <: BASE](
       }
     }
 
+
```
revans2, line 248: nit extra blank line

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -807,6 +809,16 @@ abstract class BaseExprMeta[INPUT <: Expression](
   }
 
   final override def tagSelfForGpu(): Unit = {
+    val isConstantFoldingOff =
+      SQLConf.get.optimizerExcludedRules
+        .contains("org.apache.spark.sql.catalyst.optimizer.ConstantFolding")
+    if (isConstantFoldingOff) {
+      val exp = wrapped.asInstanceOf[Expression]
```
revans2, line 816: is this cast needed?

razajafri: no, I removed and  tested against Spark 3.0.0

razajafri:
@jlowe @revans2 PTAL

docs/compatibility.md
```diff
@@ -352,4 +352,10 @@ Casting from string to timestamp currently has the following limitations.
...
  milliseconds, with 2 digits each for hours, minutes, and seconds, and 6 digits for milliseconds. 
  Only timezone 'Z' (UTC) is supported. Casting unsupported formats will result in null values. 
- 
+
+### Constant Folding
+
+ConstantFolding is an operator optimization rule in Catalyst that replaces expressions that can
+be statically evaluated with their equivalent literal values. The RAPIDS Accelerator relies
+on constant folding and parts of the query will not be accelerated if 
+`org.apache.spark.sql.catalyst.optimizer.ConstantFolding` is excluded as a rule
```
jlowe, line 361:
```suggestion
`org.apache.spark.sql.catalyst.optimizer.ConstantFolding` is excluded as a rule.
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
jlowe: A `RapidsConf` can be built from a `SQLConf` but _this is optional_.  That means one cannot assume that a `RapidsConf` contains `SQLConf` settings, and that's why getting arbitrary keys from a `RapidsConf` should not be supported.

jlowe (approved):
Looks good to me, but would be good to hear from @revans2 as well.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -807,6 +808,15 @@ abstract class BaseExprMeta[INPUT <: Expression](
   }
 
   final override def tagSelfForGpu(): Unit = {
+    val isConstantFoldingOff =
+      SQLConf.get.optimizerExcludedRules
+        .contains("org.apache.spark.sql.catalyst.optimizer.ConstantFolding")
+    if (isConstantFoldingOff) {
```
revans2, line 814: nit: This looks great, but why do we need to check the config at all? `wrapped.foldable` should be cheap to do and it not directly impacted by the `optimizerExcludedRules`.

revans2 (approved):

razajafri:
build

jlowe (approved):

PR #1768, opened 2021-02-19
Updating getting started Databricks docs

sameerz:
Updated Getting Started with Databricks based on virtual review feedback.  Updated Mortgage Notebook.  Apologies if the whole notebook changed, it got reformatted when I saved with Jupyter.  The changes are in the first cell of the notebook, providing USER_ID as a variable.  The changes are the same as the ones in the .md file.  Tested the notebook on Azure Databricks.


docs/get-started/getting-started-databricks.md
```diff
@@ -15,10 +15,14 @@ on NVIDIA GPUs on Databricks.
     * AWS: 7.3 LTS ML (GPU, Scala 2.12, Spark 3.0.1)
     * Azure: 7.3 LTS ML (GPU, Scala 2.12, Spark 3.0.1)
 
+Databricks 7.3 LTS ML runs CUDA 10.1 Update
+2[*](https://docs.databricks.com/release-notes/runtime/7.3ml.html#system-environment), and the
```
jlowe, line 19: The hyperlink on the asterisk is so small it will be easily missed.  Maybe make this more obvious?
```suggestion
[Databricks 7.3 LTS ML](https://docs.databricks.com/release-notes/runtime/7.3ml.html#system-environment)
runs CUDA 10.1 Update 2, and the
```

docs/get-started/getting-started-databricks.md
```diff
@@ -92,15 +96,17 @@ larger dataset if needed. You can find the links to the datasets at
 
 ```bash
 %sh
+
+USER_ID=<your_user_id>
```
jlowe, line 100:
```suggestion
```

docs/get-started/getting-started-databricks.md
```diff
@@ -92,15 +96,17 @@ larger dataset if needed. You can find the links to the datasets at
...
  
-wget http://rapidsai-data.s3-website.us-east-2.amazonaws.com/notebook-mortgage-data/mortgage_2000.tgz -P /Users/<your user id>/
+wget http://rapidsai-data.s3-website.us-east-2.amazonaws.com/notebook-mortgage-data/mortgage_2000.tgz -P /Users/${USER_ID}/
```
jlowe, line 102:
```suggestion
wget http://rapidsai-data.s3-website.us-east-2.amazonaws.com/notebook-mortgage-data/mortgage_2000.tgz -P /Users/${USER}/
```

docs/get-started/getting-started-databricks.md
```diff
@@ -92,15 +96,17 @@ larger dataset if needed. You can find the links to the datasets at
...
  
 mkdir -p /dbfs/FileStore/tables/mortgage
 mkdir -p /dbfs/FileStore/tables/mortgage_parquet_gpu/perf
 mkdir /dbfs/FileStore/tables/mortgage_parquet_gpu/acq
 mkdir /dbfs/FileStore/tables/mortgage_parquet_gpu/output
  
-tar xfvz /Users/<your user id>/mortgage_2000.tgz --directory /dbfs/FileStore/tables/mortgage
+tar xfvz /Users/${USER_ID}/mortgage_2000.tgz --directory /dbfs/FileStore/tables/mortgage
```
jlowe, line 109:
```suggestion
tar xfvz /Users/${USER}/mortgage_2000.tgz --directory /dbfs/FileStore/tables/mortgage
```

docs/get-started/getting-started-databricks.md
tgravescs: I'm a little confused by this line, how does this compare tot he prequisite section?    we link to their page so it is saying use those init scripts?

sameerz: During a previous virtual review (0.2 timeframe) a user tried a different version of the runtime, and ended up using the wrong cudf version (10.2 or 11.0) because they were not matching the CUDA version.  Even though we are not officially supporting other versions of Databricks, I added this text so as to make sure people were aware of that Databricks has a specific version of CUDA they support (10.1 as of 7.3 LTS ML).  

I can remove this if it is confusing.

docs/get-started/getting-started-databricks.md
sameerz: @jlowe I am fine changing `${USER_ID}` to `${USER}`, but curious why we should remove `USER=<your_user_id>` from the top of the script?

jlowe: Because `USER` is a standard enviornment variable, it does not need to be set.

docs/get-started/getting-started-databricks.md
tgravescs: I see, that makes sense, I was trying to read more into then needed.

docs/get-started/getting-started-databricks.md
sameerz: I ran in the Databricks notebook without setting $USER_ID and instead using the $USER variable.  Unfortunately in that environment the notebook is run as `root`.  In a shared cluster environment all users would be writing to the same /Users/root folder.  

```bash
%sh

echo $USER
```
`root`

jlowe: ah nevermind then, keep it as is with the separate `USER_ID` variable.  We shouldn't smash `USER` with a new value.

jlowe (approved):
Just a minor comment on the tiny hyperlink, not must-fix.

docs/get-started/getting-started-databricks.md
sameerz: Updated hyperlink per suggestion.

sameerz:
build

jlowe (approved):

PR #1779, opened 2021-02-19
Add support of DateType and TimestampType for GetTimestamp expression

nartal1:
This is follow on PR for the review comment: https://github.com/NVIDIA/spark-rapids/pull/1742#discussion_r579487836

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/expressions/rapids/Timestamp.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/expressions/rapids/Timestamp.scala
@@ -28,7 +28,9 @@ object TimeStamp {
     GpuOverrides.expr[GetTimestamp](
       "Gets timestamps from strings using given pattern.",
       ExprChecks.binaryProjectNotLambda(TypeSig.TIMESTAMP, TypeSig.TIMESTAMP,
-        ("timeExp", TypeSig.STRING, TypeSig.STRING),
+        ("timeExp",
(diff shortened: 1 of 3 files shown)


nartal1:
build

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/expressions/rapids/Timestamp.scala
```diff
@@ -28,7 +28,9 @@ object TimeStamp {
     GpuOverrides.expr[GetTimestamp](
       "Gets timestamps from strings using given pattern.",
       ExprChecks.binaryProjectNotLambda(TypeSig.TIMESTAMP, TypeSig.TIMESTAMP,
-        ("timeExp", TypeSig.STRING, TypeSig.STRING),
+        ("timeExp",
+            TypeSig.STRING + TypeSig.DATE + TypeSig.TIMESTAMP,
```
revans2, line 32: I would prefer to have us say we only support STRING unless we have tests to back up support for the other types as well.

nartal1:
@revans2 Please take a look. I have addressed the review comment https://github.com/NVIDIA/spark-rapids/pull/1742#discussion_r579487836.

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/expressions/rapids/Timestamp.scala
nartal1: Added tests for date and timestamp type. PTAL.

revans2 (approved):

revans2:
build

Issue #1764, opened 2021-02-18
[FEA] Better corner case testing for casting


revans2:
From https://github.com/rapidsai/cudf/issues/5225 it is obvious that there are some corner cases that cause issues with float/double tests.  Some of these are being fixed as a part of https://github.com/rapidsai/cudf/pull/7410 but our tests are really inadequate despite fuzzing.  Really the issue is that we are only testing with 100 values and that the values themselves will never produce invalid values that overflow. They also will never in practice hit a Double.MaxValue or Double.MinValue.  It would be nice to have some special tests to augment these.

sameerz:
Will leave open until rapidsai/cudf#7410 issue is fixed so we can add the remaining test cases.

PR #1781, opened 2021-02-21
Better float/double cases for casting tests

sperlingxx:
This pull request is to address issue #1764. And current PR also includes some code clean work .

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
@@ -231,11 +231,11 @@ class CastOpSuite extends GpuExpressionTestSuite {
     testCastToString[Long](DataTypes.LongType)
   }
 
-  ignore("cast float to string") {
+  test("cast float to string") {
     testCastToString[Float](DataTypes.FloatType, comparisonFunc = Some(compareStringifiedFloats))
   }
 
-  ignore("cast double to string") {
+  test("cast double to string") {
     testCastToString[Double](DataTypes.DoubleType, comparisonFunc = Some(compareStringifiedFloats))
   }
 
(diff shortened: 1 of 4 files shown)


sperlingxx:
build

sperlingxx:
build

tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
```diff
@@ -425,6 +383,14 @@ class CastOpSuite extends GpuExpressionTestSuite {
       col("c1").cast(FloatType))
   }
 
+  ignore("overflow double strings") {
+    testSparkResultsAreEqual("overflow double strings", badDoubleStringsDf,
```
revans2, line 387: `badDoubleStringsDf` is not as comprehensive as I would like it to be. The values in it are big enough to force an overflow, but there are still a number of corner cases where according to https://github.com/rapidsai/cudf/issues/5225 are not being covered. The patch to fix it https://github.com/rapidsai/cudf/pull/7410 is not in and not working 100% either in these same overflow cases. I am happy to have these go in as it, but I would like to see a follow on issue to cover the corner cases that are missing, and have a comment in `badDoubleStringsDf` pointing to it.  Would be good to look at the float one too and make sure it is also not suffering from the same issue.

sperlingxx:
build

tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
sperlingxx: Hi @revans2, I tried with many other similar (slightly overflow) cases. And I found GPU can produce correct results for almost all of them, except some special cases included in `badDoubleStringsDf`.
For float type, I failed to find any mismatching case.
For double type, I only found numbers within range [1.79769313486231581E308, 1.797693134862316E308) produce inconsistent results with spark.

sperlingxx:
build

revans2 (approved):

revans2:
@razajafri you filed the original bug about casting from string to double.  Could you take a look at this PR too?

tests/src/test/scala/com/nvidia/spark/rapids/SparkQueryCompareTestSuite.scala
```diff
@@ -1131,14 +1131,16 @@ trait SparkQueryCompareTestSuite extends FunSuite with Arm {
       "9.8e5").toDF("c0")
   }
 
-  def badFloatStringsDf(session: SparkSession): DataFrame = {
+  def invalidFloatStringsDf(session: SparkSession): DataFrame = {
     import session.sqlContext.implicits._
     Seq(("A", "null"), ("1.3", "43.54")).toDF("c0", "c1")
   }
 
   def badDoubleStringsDf(session: SparkSession): DataFrame = {
```
razajafri, line 1139: nit: for consistency may be this should be renamed to `invalidDoubleStringsDf`

razajafri:
> @razajafri you filed the original bug about casting from string to double. Could you take a look at this PR too?

Just one nit otherwise LGTM

PR #1784, opened 2021-02-22
spillable cache for GpuCartesianRDD

sperlingxx:
Use SpillableColumnarBatch to cache stream-side data in case of re-computation caused by nested-loop.


sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
```diff
@@ -141,27 +143,41 @@ class GpuCartesianRDD(
   override def compute(split: Partition, context: TaskContext):
   Iterator[ColumnarBatch] = {
     val currSplit = split.asInstanceOf[GpuCartesianPartition]
+
+    // create a buffer to cache stream-side data in a spillable manner
+    val spillBatchBuffer = mutable.ArrayBuffer[SpillableColumnarBatch]()
+    closeOnExcept(spillBatchBuffer) { buffer =>
+      rdd2.iterator(currSplit.s2, context).foreach { cb =>
+        // TODO: is it necessary to create a specific spill priorities for spillBatchBuffer?
```
revans2, line 151: The priority you set is fine. The issue is with avoiding spilling, and we may want to actually move to more dynamic spill priorities at some point.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
```diff
@@ -141,27 +143,41 @@ class GpuCartesianRDD(
...
+        buffer += SpillableColumnarBatch(
+          cb.getBatch, SpillPriorities.ACTIVE_ON_DECK_PRIORITY)
+      }
+    }
+
     rdd1.iterator(currSplit.s1, context).flatMap { lhs =>
       val table = withResource(lhs) { lhs =>
```
revans2, line 158: If we are doing this right we need to make `lhs` spillable too.  Because we are going to do the join, and return multiple values while holding on to it. This means we will likely have to modify BroadcastNestedLoopJoinExecBase as well to be able to deal with this. We might need to make it a functor or something like that so we can keep the old behavior for broadcast nested loop join until we can update the broadcast tables to also be spillable, which is coming.

revans2:
Also I am a little nervous about putting something like this into 0.4. It is for operators that are off by default, but I would feel better if we could retarget this for 0.5 so we have more time to test it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
```diff
@@ -141,27 +143,41 @@ class GpuCartesianRDD(
   override def compute(split: Partition, context: TaskContext):
   Iterator[ColumnarBatch] = {
     val currSplit = split.asInstanceOf[GpuCartesianPartition]
+
+    // create a buffer to cache stream-side data in a spillable manner
+    val spillBatchBuffer = mutable.ArrayBuffer[SpillableColumnarBatch]()
+    closeOnExcept(spillBatchBuffer) { buffer =>
+      rdd2.iterator(currSplit.s2, context).foreach { cb =>
```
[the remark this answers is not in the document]
jlowe: Yes, agreed.  I'd rather not add yet another loop through the data.

jlowe (changes requested):
> I would feel better if we could retarget this for 0.5 so we have more time to test it.

+1.  This should target the 0.5 release.

sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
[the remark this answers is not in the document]
sperlingxx: I've replaced this with a lazy populating implementation.

jlowe:
This is still targeting branch-0.4.  It needs to be retargeted to branch-0.5.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
```diff
@@ -141,15 +143,32 @@ class GpuCartesianRDD(
...
-      // Ideally instead of looping through and recomputing rdd2 for
-      // each batch in rdd1 we would instead cache rdd2 in a way that
-      // it could spill to disk so we can avoid re-computation
+
+      val streamIterator = if (index == 0) {
+        // lazily compute and cache stream-side data
+        rdd2.iterator(currSplit.s2, context).map { serializableBatch =>
+          closeOnExcept(spillBatchBuffer) { buffer =>
+            val batch = SpillableColumnarBatch(
+              serializableBatch.getBatch, SpillPriorities.ACTIVE_ON_DECK_PRIORITY)
```
jlowe, line 160: This is now missing the spillable callback argument that was added in #1719 which should be used to tie any spilling to spill metrics added to this exec node.

jlowe (changes requested):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
sperlingxx: Hi @jlowe, I created another PR #1878 for this improvement, because current is based on `branch-0.4` (not `branch-0.5`).

jlowe: In the future a new PR isn't necessary when rebasing.  You just need to retarget the PR base branch, as was already done for this PR, and then merge in the new base branch.

jlowe:
Superceded by #1878

Issue #1785, opened 2021-02-22
[BUG] Rapids pytest integration tests FAILED on Yarn cluster with unrecognized arguments: `--std_input_path=src/test/resources/`


NvTimLiu:
**Describe the bug**
Run rapids pytest IT on EGX Yarn cluster
Rapids integration tests FAILED on Yarn cluster with unrecognized arguments: `--std_input_path=src/test/resources/` as below:
    _cd integration_tests/
    spark-submit --master --deploy-mode cluster runtests.py -std_input_path=src/test/resources/_
    _LOG: http://spark-egx-03:8088/cluster/app/application_1603128018386_6400_
        ERROR: usage: runtests.py [options] [file_or_dir] [file_or_dir] [...]
        runtests.py: error: unrecognized arguments: --std_input_path=src/test/resources/

If we remove `--deploy-mode cluster`, the tests can PASS.
    _spark-submit --master runtests.py -std_input_path=src/test/resources/_
    _LOG: http://spark-egx-03:8088/cluster/app/application_1603128018386_6399_

        src/main/python/struct_test.py::test_orderby_struct[Struct(['first', String],['second', Array(String)],[  'third', Array(String)])] PASSED [ 95%]
        src/main/python/struct_test.py::test_orderby_struct_2[Struct(['first', String],['second', Array(String)]  ,['third', Array(String)])] PASSED [100%]
============================= 23 passed in 41.09s ==============================

****Detailed scripts: timl@spark-egx-03:/home/timl/yarn-IT/submit-yarn.sh****
#!/bin/bash
set +ex
export SPARK_CONF_DIR=/usr/hdp/current/spark2-client/conf/
export HADOOP_HOME=/usr/hdp/3.1.0.0-78/hadoop
export HADOOP_CONF_DIR=/usr/hdp/3.1.0.0-78/hadoop/conf
export SPARK_HOME=/home/timl/spark-3.0.0-bin-hadoop3.2
export PATH=$SPARK_HOME/bin:$SPARK_HOME/sbin:$PATH

export PYSPARK_PYTHON=/usr/bin/python3.6
export PYSPARK_DRIVER_PYTHON=/usr/bin/python3.6

OS_TYPE=ubuntu18
CUDA_NAME=cuda11.0
CUDA_DOCKER_TAG=${CUDA_NAME/./-}

CUDA_CLASSIFIER=${CUDA_NAME/./-}

FINAL=${CUDA_CLASSIFIER: -2}
[...]

Issue #1640, opened 2021-02-01
Update CI builds for new udf-examples jar name


jlowe:
After #1636 the udf-examples jar name changed.  All of the CI builds that are referencing the jar need to be updated.

tgravescs:
I'm not sure all the builds got updated with the jar in the first place so we need to check missing as well.

GaryShen2008:
I don't think we have any CI/CD to use the udf example right now.
We're trying to add the job for the query using urlDecode.
So, it may be no action item for now.

jlowe:
> Does integration test use udf example jar?

Any CI/CD using a Spark built with Hive support and using `run_pyspark_from_build.sh` will add the udf-examples jar and run the urldecode integration tests.  This includes premerge CI and the primary nightly integration which is already running successfully with the new jar name.  I'm not sure about some of the integration tests specific to cloud environments.

tgravescs:
I think we should update all the integration tests to use run_pyspark_from_build.sh if possible. This might mean changes to the script if we need to specify a different master. I didn't think that was supported.  This way any changes devs make should automatically be picked up.

revans2:
> This might mean changes to the script if we need to specify a different master. I didn't think that was supported.

The scripts were set up to avoid making assumptions about what it was running on.  If you set `SPARK_CONF_DIR` to the configs for the cluster you want to run on it will pick them up and use them.  If you are running on a truly distributed cluster you need to either disable the parallel test run, or make sure that you have enough resources in your cluster to run N+1 applications in parallel. In that case too we probably need to copy the standard input files to the default distributed file system so that the tests can find them.

NvTimLiu:
checking

NvTimLiu:
Cloud environments changed to refer to the correct udf-examples jar: 
https://gitlab-master.nvidia.com/nvspark/blossom-jenkins/-/blob/main/utils/download-jars.sh#L17
https://gitlab-master.nvidia.com/nvspark/blossom-jenkins/-/blob/main/databricks/pytests-on-databricks.sh#L38
https://gitlab-master.nvidia.com/nvspark/blossom-jenkins/-/blob/main/emr/pytests-on-emr.sh#L41
https://gitlab-master.nvidia.com/nvspark/blossom-jenkins/-/blob/main/dataproc/pytests-on-dataproc.sh#L47

NvTimLiu:
close as Clound environment changed to refer to the correct udf-examples jar

Issue #1568, opened 2021-01-22
[BUILD] cleanup unused jenkins files and scripts


tgravescs:
**Is your feature request related to a problem? Please describe.**
Directory:
https://github.com/NVIDIA/spark-rapids/tree/branch-0.4/jenkins

has jenkins files, docker files, and scripts. I believe many of those are not used anymore (like the databricks ones). We should remove any files from here that aren't used.

I'm not sure if one of the Docker files was an example for users. We do already have one for k8s here: https://github.com/NVIDIA/spark-rapids/tree/branch-0.4/docs/get-started. If we want it visible to users perhaps we should document and put it in similar location or create a separate directory for it.

tgravescs:
@GaryShen2008

NvTimLiu:
@tgravescs 
1, The below 2 Dockerfile are useless, we’ll remove them from GitHub/spark-rapids repo.
Dockerfile.integration.ubuntu16
Dockerfile.ubuntu16
2, The other Dockerfiles/Jenkinsfiles/scripts are essential or spark-rapids prem-merge and integration jobs.
3, We put these Dockerfiles/Jenkinsfiles/scripts in GitHub, so dev team can update them conveniently, instead of making them visible to external users.
    --- For example, if someone needs to install a python module for IT, he can change the Dockerfile-blossom.integration.centos7 directly on the GitHub repo.
So after clean useless Dockerfiles in item 1, need we to move the whole Jenkins dir, or some subdirs[databricks] or files out of spark-rapids Github int internal Gitlab repo?

tgravescs:
2) ok, I guess Jenkinsfile.databricks301nightly is still used here, but its inconsistent with the other nightly now so I would rather see it moved out to be consistent.  I do see some of the other scripts are used from here, is there a particular reason to leave them in this repo vs move out to where Jenkins files are?
3) The problem is that some of the scripts are here and some are in the other repo so Dev doesn't know where to update all the builds, which is confusing. For instance we wanted to add build argument, the scripts updated nightly but not all the IT tests.  It would be have a single scripts for running the IT tests for instance that others could call with parameters if needed.   I don't think it matters which repo this in as much as keeping it consistent and documenting where things are at.

pxLi:
> 1. ok, I guess Jenkinsfile.databricks301nightly is still used here, but its inconsistent with the other nightly now so I would rather see it moved out to be consistent.  I do see some of the other scripts are used from here.
[...]

PR #1787, opened 2021-02-22
support distributed file path in cloud environment

NvTimLiu:
LOCAL_JAR_PATH is set in the cloud environment to support alternate local jars NOT building from the source code


integration_tests/run_pyspark_from_build.sh
```diff
@@ -28,7 +28,7 @@ else
     # support alternate local jars NOT building from the source code
     if [ -d "$LOCAL_JAR_PATH" ]; then
         CUDF_JARS=$(echo "$LOCAL_JAR_PATH"/cudf-*.jar)
-        PLUGIN_JARS=$(echo "$LOCAL_JAR_PATH"/rapids-4-spark*.jar)
```
NvTimLiu, line 31: fix file name filter

integration_tests/run_pyspark_from_build.sh
```diff
@@ -28,7 +28,7 @@ else
...
+        PLUGIN_JARS=$(echo "$LOCAL_JAR_PATH"/rapids-4-spark_*.jar)
```
NvTimLiu, line 31: fix file name filter

integration_tests/run_pyspark_from_build.sh
```diff
@@ -115,15 +115,22 @@ else
           $RUN_TEST_PARAMS \
           "$@"
     else
+        ## 'REMOTE_RES_PATH' is set in the cloud environment to support distributed file path
+        if [ -n "$REMOTE_RES_PATH" ]; then
```
NvTimLiu, line 119: Here 'REMOTE_RES_PATH' could be hdfs:/xxx or s3a://xxx

integration_tests/run_pyspark_from_build.sh
```diff
@@ -115,15 +115,22 @@ else
...
+            PYTEST_RES_PARAMS="$REMOTE_RES_PATH/../runtests.py \
+              --std_input_path=$REMOTE_RES_PATH/src/test/resources/"
+        else
+            PYTEST_RES_PARAMS="$SCRIPTPATH/runtests.py --rootdir $SCRIPTPATH $SCRIPTPATH/src/main/python \
```
NvTimLiu, line 123: seems `--rootdir` option does not support hdfs:/xxx or s3a://xxx PATH, only support file path?

 --rootdir=ROOTDIR     Define root directory for tests. Can be relative path: 'root_dir', './root_dir',
                        'root_dir/another_dir/'; absolute path: '/home/user/root_dir'; path with
                        variables: '$HOME/root_dir'.

ERROR: Directory /mnt/yarn/usercache/hadoop/appcache/application_1613996222010_0005/container_1613996222010_0005_01_000001/hdfs:/mnt/integration_tests' not found. Check your '--rootdir' option.

NvTimLiu:
Related issue https://github.com/NVIDIA/spark-rapids/issues/1640

NvTimLiu:
Related issue:
https://github.com/NVIDIA/spark-rapids/issues/1421
https://github.com/NVIDIA/spark-rapids/issues/1568

NvTimLiu:
https://github.com/NVIDIA/spark-rapids/issues/1785

NvTimLiu:
build

NvTimLiu:
build

integration_tests/run_pyspark_from_build.sh
```diff
@@ -28,9 +28,11 @@ else
     # support alternate local jars NOT building from the source code
     if [ -d "$LOCAL_JAR_PATH" ]; then
         CUDF_JARS=$(echo "$LOCAL_JAR_PATH"/cudf-*.jar)
-        PLUGIN_JARS=$(echo "$LOCAL_JAR_PATH"/rapids-4-spark*.jar)
+        PLUGIN_JARS=$(echo "$LOCAL_JAR_PATH"/rapids-4-spark_*.jar)
         TEST_JARS=$(echo "$LOCAL_JAR_PATH"/rapids-4-spark-integration-tests*.jar)
         UDF_EXAMPLE_JARS=$(echo "$LOCAL_JAR_PATH"/rapids-4-spark-udf-examples*.jar)
+        ## reset SCRIPTPATH to support alternate scripts path on the cloud environment
+        SCRIPTPATH="$LOCAL_JAR_PATH/integration_tests"
```
NvTimLiu, line 35: reset SCRIPTPATH to support alternate scripts path on the cloud environment

NvTimLiu:
@revans2  Could you please help to review? thanks!

integration_tests/run_pyspark_from_build.sh
revans2: Why was this causing issues where it does not cause issues in other places? It would be good to add a comment explaining this because a few lines below it in the else clause.

integration_tests/run_pyspark_from_build.sh
revans2: But it is not the SCRIPTPATH any longer.  SCRIPTPATH is where this script resides and we are finding other things relative to that path.  I would prefer to have a new variable like INTEGRATION_TEST_ROOT or something that can be set to SCRIPTPATH in the common case, and then set here to what we want it to be.

revans2 (commented):
In addition I really think we need to update the document for this script to include LOCAL_JAR_PATH in integration_tests/README.md.

Before this change LOCAL_JAR_PATH was clear, it was the path to set where the dependency jars come from.   After this change it looks like we are forcing a very specific directory layout, which needs to be documented very clearly and we need to be sure none of our CI builds are using LOCAL_JAR_PATH differently, or we need to separate out how to set where to find the integration_tests directory from how to find the local jars.

I also am not sure how this fixes the issues you were seeing with the tests finding std_input_path

NvTimLiu:
Why was this causing issues where it does not cause issues in other places? It would be good to add a comment explaining this because a few lines below it in the else clause. --> better to add this, though it does not cause error currently.  `PLUGIN_JARS=$(echo "$LOCAL_JAR_PATH"/rapids-4-spark*.jar)` will find `rapids-4-spark_2.12-SNAPSHOT.jar, rapids-4-spark-integration-tests*.jar rapids-4-spark-udf-examples*.jar` . Here we only want `PLUGIN_JARS=rapids-4-spark_2.12-SNAPSHOT.jar`

revans2:
> Why was this causing issues where it does not cause issues in other places? It would be good to add a comment explaining this because a few lines below it in the else clause. --> better to add this, though it does not cause error currently. `PLUGIN_JARS=$(echo "$LOCAL_JAR_PATH"/rapids-4-spark*.jar)` will find `rapids-4-spark_2.12-SNAPSHOT.jar, rapids-4-spark-integration-tests*.jar rapids-4-spark-udf-examples*.jar` . Here we only want `PLUGIN_JARS=rapids-4-spark_2.12-SNAPSHOT.jar`

Then lets update it in both places so it is consistent

NvTimLiu:
But it is not the SCRIPTPATH any longer. --> Sorry, here we need NOT to change `SCRIPTPATH`  as `SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )" cd "$SCRIPTPATH"` makes sure SCRIPTPATH is the script path (we already put run_pyspark_from_build.sh under dir `integration_tests`). I'll revert it.

NvTimLiu:
I also am not sure how this fixes the issues you were seeing with the tests finding std_input_path --> As `--rootdir` can NOT work with HDFS path(means we can not use HDFS), and `--std_input_path` can NOT work with  relative path, so I put the `integration_tests` into all the yarn nodes under the `absolute` dir '$SCRIPTPATH', then the absolute PATH $SCRIPTPATH works for both  `--rootdir`  & `--std_input_path`

Why it did not work before? --> We only put the `integration_tests` onto the Yarn master node, then we transferred the it on to Yarn worker nodes by --file/--archive,  so `integration_tests` path was relative[did NOT work with --std_input_path].

NvTimLiu:
In addition I really think we need to update the document for this script to include LOCAL_JAR_PATH in integration_tests/README.md. --> Will update README.md

revans2:
This code change as it is now look fine, but it has very little to do with the title of this issue.

> support distributed file path in cloud environment

It also has almost nothing to do with any of the issues/bugs mentioned  in this PR 

* Update CI Builds for new udf-examples jar name #1640
* Integration tests should have a flag to avoid silently skipping some tests #1421
* cleanup unused jenkins files and scripts #1568
* Rapids pytest integration tests FAILED on Yarn cluster with unrecognized arguments: `--std_input_path=src/test/resources/` #1785

Before approving this I just want to be sure I understand how this fits into those other issues. Will there be follow on work, especially for `--std_input_path` as we discussed in #1785 or do you think this is enough to resolve it?

NvTimLiu:
> It also has almost nothing to do with any of the issues/bugs mentioned in this PR
> 
> * Update CI Builds for new udf-examples jar name #1640

Updated Jenkins scripts to download the latest udf-example jar with `_2.12` appended in the jar name.
> * Integration tests should have a flag to avoid silently skipping some tests #1421
All the Cloud environment changed to use `integration_tests/run_pyspark_from_build.sh`

`TEST_TYPE` is NOT set, need we config it TEST_TYPE=`nightly` for all the Clould environment IT pipeline? @jlowe @revans2 
Test result on EMR as below:

23:49:36  SKIPPED [1] ../../../../../../integration_tests/src/main/python/conftest.py:468: rapids_udf_example_native is not configured to run
23:49:36  SKIPPED [105] ../../../../../../integration_tests/src/main/python/conftest.py:452: TPC-DS not configured to run
23:49:36  SKIPPED [4] ../../../../../../integration_tests/src/main/python/conftest.py:381: TPCxBB not configured to run
23:49:36  SKIPPED [11] ../../../../../../integration_tests/src/main/python/conftest.py:461: cudf_udf not configured to run
23:49:36  4715 passed, 21 skipped,145 xfailed, 7 xpassed,104 warnings, in 5248.88s (1:27:28)
23:49:36  
23:49:36  End of LogType:stdout
23:49:36  ***********************************************************************

> * cleanup unused jenkins files and scripts #1568

PR ready, retarget to branch-0.5

> * Rapids pytest integration tests FAILED on Yarn cluster with unrecognized arguments: `--std_input_path=src/test/resources/` #1785
> 
> Before approving this I just want to be sure I understand how this fits into those other issues. Will there be follow on work, especially for `--std_input_path` as we discussed in #1785 or do you think this is enough to resolve it?

Yes, for `--std_input_path` issue, we put integration tests scripts to all of the distributed filesystem nodes, so the `$SCRIPTPATH` can be found locally.

> This code change as it is now look fine, but it has very little to do with the title of this issue.
> 
> > support distributed file path in cloud environment
> 
> It also has almost nothing to do with any of the issues/bugs mentioned in this PR
> 
> * Update CI Builds for new udf-examples jar name #1640
> * Integration tests should have a flag to avoid silently skipping some tests #1421
> * cleanup unused jenkins files and scripts #1568
> * Rapids pytest integration tests FAILED on Yarn cluster with unrecognized arguments: `--std_input_path=src/test/resources/` #1785
> 
> Before approving this I just want to be sure I understand how this fits into those other issues. Will there be follow on work, especially for `--std_input_path` as we discussed in #1785 or do you think this is enough to resolve it?

NvTimLiu:
Update the document for how to set 'LOCAL_JAR_PATH'

NvTimLiu:
build

revans2:
> Yes, for --std_input_path issue, we put integration tests scripts to all of the distributed filesystem nodes, so the $SCRIPTPATH can be found locally.

I really don't like that.  How is it less work to keep that data in sync on all of those nodes than it is to have a script to copy it to a distributed file system?  Or worse it is not kept in sync and when we change something with the data the tests start to fail.

Like I said I think this code is fine as is and if you want to merge it in we can, but I don't like the current plan for the input test files and to fix it I think we will need changes to these scripts.

NvTimLiu:
> I really don't like that. How is it less work to keep that data in sync on all of those nodes than it is to have a script to copy it to a distributed file system? Or worse it is not kept in sync and when we change something with the data the tests start to fail.
> 
> Like I said I think this code is fine as is and if you want to merge it in we can, but I don't like the current plan for the input test files and to fix it I think we will need changes to these scripts.

I agree with you, the way of syncing the data onto all the nodes' SCRIPTPATH dir is ugly. 
But the we need NOT to change the `un_pyspark_from_build.sh` using this way.


I'd prefer to sync `--std_input_path` onto the distributed filesystem, too.

As the `--rootdir` did NOT support distributed file path, to utilize the distribute file system for the `--std_input_path`, besides `$SCRIPTPATH`, we will need to introduce 2 extra ENV into [run_pyspark_from_build.sh ](https://github.com/NVIDIA/spark-rapids/blob/branch-0.5/integration_tests/run_pyspark_from_build.sh#L124--L126) ,  e.g. 

1. `DISTRIBUTEPATH=hdfs/s3a/gs/integration_tests/src/test/resources` for the `--std_input_path=$INPUTPATH`,  
2. `LOCAL_ROOTDIR=./integration_tests` for the `--rootdir=$LOCAL_ROOTDIR $LOCAL_ROOTDIR/src/main/python`


-          "$SCRIPTPATH"/runtests.py --rootdir "$SCRIPTPATH" "$SCRIPTPATH"/src/main/python \
+          "$SCRIPTPATH"/runtests.py --rootdir "$LOCAL_ROOTDIR" "$LOCAL_ROOTDIR"/src/main/python \
           -v -rfExXs "$TEST_TAGS" \
-          --std_input_path="$SCRIPTPATH"/src/test/resources/ \
+          --std_input_path="$DISTRIBUTEPATH"/src/test/resources/ \

Seems the extra too ENVs are introduced in the  run_pyspark_from_build.sh?

revans2:
I am fine with having extra environment variable to help control `run_pyspark_from_build.sh` having one that would override the default value for std_input_path sounds like a good change.  And the LOCAL_ROOT_DIR change also sounds fine.

NvTimLiu:
build

NvTimLiu:
build

NvTimLiu:
> I am fine with having extra environment variable to help control `run_pyspark_from_build.sh` having one that would override the default value for std_input_path sounds like a good change. And the LOCAL_ROOT_DIR change also sounds fine.

@revans2  Added `--rootdir LOCAL_ROOTDIR`  & `--std_input_path CLOUD_INPUTPATH`, could you please help to review. Thanks!

integration_tests/run_pyspark_from_build.sh
```diff
@@ -93,6 +93,12 @@ else
     RUN_DIR="$SCRIPTPATH"/target/run_dir
     mkdir -p "$RUN_DIR"
     cd "$RUN_DIR"
+
+    ## Under cloud environment, overwrite the '--rootdir' param to point to the working directory of each excutor
+    LOCAL_ROOTDIR=${LOCAL_ROOTDIR:-"$SCRIPTPATH"}
+    ## Under cloud environment, overwrite the '--std_input_path' param to point to the distributed file path
+    CLOUD_INPUTPATH=${CLOUD_INPUTPATH:-"$SCRIPTPATH"}
```
revans2, line 100: nit: this is not a **CLOUD**_INPUTPATH in all cases. I think it would be better to just call it `INPUT_PATH` i.e.

```
INPUTPATH=${INPUTPATH:-"$SCRIPTPATH"}
```

revans2 (commented):
just a minor nit

integration_tests/run_pyspark_from_build.sh
NvTimLiu: Sounds good, updated!

NvTimLiu:
build

revans2 (approved):

revans2:
@NvTimLiu could you please upmerge your branch? it looks like some changes to cudf went in and broke backwards compatibility.

NvTimLiu:
build

PR #1788, opened 2021-02-22
Spill metrics everywhere

revans2:
This enables spill metrics for all parts of the GPU plan (except UCX shuffle) that can spill. It also removes some old code there was there when cudf contiguous split did not work with nested types.

Right now it changes the API in a non-backwards compatible way to force all of the operators that can spill to report metrics on spilling. I am OK with keeping it backwards compatible for now, but eventually I want to make the change permanent.


revans2:
build

jlowe (approved):

PR #1796, opened 2021-02-23
Have most of range partitioning run on the GPU

### LINKED ISSUE #1727: [FEA] Put RangePartitioner data path on the GPU
The range partitioner is written to mostly use a copy of the corresponding Spark code.  To do this it first pulls all of the data back to the CPU so it can do [reservoir sampling](https://en.wikipedia.org/wiki/Reservoir_sampling) on the data as it arrives to get N example rows. These rows are then brought back to the driver where it decides on cutoff points for the data. In my testing for [out of core sort](https://github.com/NVIDIA/spark-rapids/pull/1719) I noticed that this part of the code was very slow. So slow that the CPU can often beat the GPU in some simple order-by queries without any tuning. 

I did a quick proof of concept to see if I could make the performance better by randomly selecting a gather map and keeping the data on the GPU as much as possible. Please note that this was a proof of concept where I have not tested it completely (there are bugs in it and it is not ready for prime time). It was mostly to see what kind of performance improvements we might see.

I then ran a query to see what kind of performance difference we might be able to expect.

```
spark.time(spark.read.parquet("./500M").orderBy(col("a").desc, col("b")).write.mode("overwrite").parquet("./target/output"))
```

Here "./500M" is a data set with 500 million rows in it. The schema is
```
root
 |-- a: double (nullable = true)
 |-- b: string (nullable = true)
```

I got the following results For `local[12]` on a 12 thread CPU with 64 GB heap 1GB input partition size and 200 shuffle partitions. For the GPU I had a concurrency of 1 and a batch size of 2GB (note that this was on top of my out of core patch, but it should have little impact here) All of the times are in ms.

| TYPE | RUN1 | RUN2 | RUN3 | RUN4 |
|-------|-------|-------|-------|-------|
|CPU|108614|106662|108207|106205|
|ORIG GPU|120724|114868|117392|115627|
|Proof Of Concept|47432|46793|43453|44346|

This means the new proof of concept is 2.56x faster than the current GPU code and 2.36x faster than the CPU code.  A lot of the reason for the slowness is because the GPU semaphore is not released after we pull the data back to the CPU. So I tuned it to have 2 concurrent GPU tasks, 1GB of batch size, and 12 shuffle partitions.  The CPU time here got slower so I will not include it.

| TYPE | RUN1 | RUN2 | RUN3 | RUN4 |
|-------|-------|-------|-------|-------|
|ORIG GPU (tuned)|65561|63010|61392|61907|
|Proof Of Concept (tuned)|41325|36222|34965|34000|

This helped a lot and now the original GPU code is 1.87x faster than it was before and 1.72x faster than the CPU code, but the proof of concept also got better and is 1.75x faster than the original GPU partitioning and 3.02x faster than the CPU code. I tried more parallelism but it had diminishing impact to the performance of both of them and fairly quickly on the now partitioning code I was limited by the performance of by disk (500MB/sec) to write out the result instead of the performance of the GPU.

### LINKED ISSUE #1730: [BUG] Range Partitioning can crash when processing is in the order-by
**Describe the bug**
If we are going to order by a column that does not exist, but is generated on the fly Rnge partitioning crashes.

I am adding some tests as a part of the out of core sort work, but the diff is here...

```diff
diff --git a/integration_tests/src/main/python/sort_test.py b/integration_tests/src/main/python/sort_test.py
index 7c1ebd2a..6cf0003c 100644
--- a/integration_tests/src/main/python/sort_test.py
+++ b/integration_tests/src/main/python/sort_test.py
@@ -67,3 +67,15 @@ orderable_gens_sort_without_neg_decimal = [n for n in orderable_gens_sort if not
 def test_multi_orderby_with_limit(data_gen):
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : binary_op_df(spark, data_gen).orderBy(f.col('a'), f.col('b').desc()).limit(100))
+
+# We are not trying all possabilities, just doing a few where the numbers should work out.
+@pytest.mark.parametrize('data_gen', [byte_gen, long_gen, float_gen], ids=idfn)
+def test_orderby_with_processing(data_gen):
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : unary_op_df(spark, data_gen).orderBy(f.lit(100) - f.col('a')))
+
+# We are not trying all possabilities, just doing a few where the numbers should work out.
+@pytest.mark.parametrize('data_gen', [byte_gen, long_gen, float_gen], ids=idfn)
+def test_orderby_with_processing_and_limit(data_gen):
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : unary_op_df(spark, data_gen).orderBy(f.lit(100) - f.col('a')).limit(100))
```

The first test fails.  The second test passes
```
E                   Caused by: java.lang.UnsupportedOperationException: Cannot evaluate expression: (100.0 - input[0, float, true])
E                   	at org.apache.spark.sql.catalyst.expressions.Unevaluable.eval(Expression.scala:304)
E                   	at org.apache.spark.sql.catalyst.expressions.Unevaluable.eval$(Expression.scala:303)
E                   	at com.nvidia.spark.rapids.CudfBinaryOperator.eval(GpuExpressions.scala:261)
E                   	at org.apache.spark.sql.catalyst.expressions.InterpretedUnsafeProjection.apply(InterpretedUnsafeProjection.scala:90)
E                   	at org.apache.spark.sql.catalyst.expressions.InterpretedUnsafeProjection.apply(InterpretedUnsafeProjection.scala:34)
E                   	at scala.collection.Iterator$$anon$10.next(Iterator.scala:459)
E                   	at scala.collection.Iterator$$anon$10.next(Iterator.scala:459)
E                   	at scala.collection.Iterator$$anon$10.next(Iterator.scala:459)
E                   	at com.nvidia.spark.rapids.SamplingUtils$.reservoirSampleAndCount(SamplingUtils.scala:47)
E                   	at com.nvidia.spark.rapids.GpuRangePartitioner.$anonfun$sketch$1(GpuRangePartitioner.scala:49)
E                   	at com.nvidia.spark.rapids.GpuRangePartitioner.$anonfun$sketch$1$adapted(GpuRangePartitioner.scala:46)
E                   	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsWithIndex$2(RDD.scala:915)
E                   	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsWithIndex$2$adapted(RDD.scala:915)
E                   	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
E                   	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:373)
E                   	at org.apache.spark.rdd.RDD.iterator(RDD.scala:337)
E                   	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:90)
E                   	at org.apache.spark.scheduler.Task.run(Task.scala:131)
E                   	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:497)
E                   	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1439)
E                   	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:500)
E                   	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
E                   	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
E                   	... 1 more
```

**Expected behavior**
We should be able to sort just like the CPU does.

**Additional context**
For a perminent fix we might want to look at combining this work with #1727.  In the short term we can probably disable this use case (as it has not shown up outside of testing) and fall back to the CPU.

revans2:
This fixes #1727 it also fixes the underlying issue of #1730, which before made us turn off range partitioning on the GPU in some cases.

Range partitioning needs to get cutoff values to partition the data on. It does this first by doing random sampling of rows from the input data.  Before that was not done on the GPU, but now it is.  After that the rows are collected back to the drive, sorted on the CPU and then cutoffs are found that are theoretically evenly spaced.  Before that sorting on the CPU could not work because the SortOrder had been replaced with GPU versions.  Now we use the GpuSorter class to materialize those columns and then rewrite the SortOrder to be able to still sort on the CPU.

As for performance numbers I the past a simple order-by for a large amount of data (> 18 GB) the CPU and GPU would often tie or the GPU would lose without any tuning.  Now I see the GPU being 3 to 4x faster than the CPU, and the bottleneck for me is that my disk cannot write out the shuffle data and final results fast enough. If it could I think we would be even faster.


revans2:
build

integration_tests/src/main/python/sort_test.py
```diff
@@ -81,3 +81,17 @@ def test_orderby_with_processing_and_limit(data_gen):
     assert_gpu_and_cpu_are_equal_collect(
             # avoid ambiguity in the order by statement for floating point by including a as a backup ordering column
             lambda spark : unary_op_df(spark, data_gen).orderBy(f.lit(100) - f.col('a'), f.col('a')).limit(100))
+
+# We are not trying all possibilities, just doing a few with numbers so the query works.
+@pytest.mark.parametrize('data_gen', [byte_gen, long_gen, float_gen], ids=idfn)
+def test_single_orderby_with_skew(data_gen):
+    # When doing range partitioning the upstream data is sampled to try and get the bounds for cutoffs.
+    # the data comes back as really skewed those partitions will be resampled for more data
+    # this is to try and trigger it to happen
```
jlowe, line 90:
```suggestion
    # If the data comes back as really skewed then those partitions will be resampled for more data.
    # This is to try and trigger it to happen.
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRangePartitioner.scala
```diff
@@ -17,43 +17,55 @@
 package com.nvidia.spark.rapids
```
jlowe, line 17: 2021 copyrights

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
```diff
@@ -48,6 +48,19 @@ private class GpuRowToColumnConverter(schema: StructType) extends Serializable {
     }
     bytes
   }
+
+  /**
+   * Convert an array of rows into a batch. Please not that this does not do bounds checking so
+   * keep the size of the batch small.
+   * @param rows the rows to convert.
```
jlowe, line 55: `schema` parameter is missing here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
```diff
@@ -48,6 +48,19 @@ private class GpuRowToColumnConverter(schema: StructType) extends Serializable {
     }
     bytes
   }
+
+  /**
+   * Convert an array of rows into a batch. Please not that this does not do bounds checking so
```
jlowe, line 53:
```suggestion
   * Convert an array of rows into a batch. Please note that this does not do bounds checking so
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SamplingUtils.scala
```diff
@@ -20,62 +22,218 @@ package com.nvidia.spark.rapids
...
+        read = true
+        withSortColumns
+      }
+    }).map(_.copy())
+    retIterator.toArray
+  }
+
+  /**
+   * Reservoir sampling implementation that also returns the input size.
+   * @param k the number of rows to randomly select.
```
jlowe, line 147: All the other parameters aren't documented here.

revans2:
build

revans2:
@jlowe I addressed your review comments please take another look

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SamplingUtils.scala
```diff
@@ -68,6 +68,10 @@ object SamplingUtils extends Arm {
   /**
    * Random sampling without replacement.
    * @param fraction the percentage of rows to randomly select
```
jlowe, line 70: `input` parameter is missing

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SamplingUtils.scala
jlowe: `input` parameter is missing

kuhushukla (dismissed):
Looking at the the majority of the changes, this looks good.

revans2:
build

jlowe (dismissed):

jlowe (dismissed):

revans2:
build

revans2:
build

revans2:
build

revans2:
@jlowe could you please take another look?

integration_tests/src/main/python/window_function_test.py
```diff
@@ -241,6 +241,7 @@ def _gen_data_for_collect(nullable=True):
             ['child_decimal', DecimalGen(precision=8, scale=3)]]))]
 
 
+
```
jlowe, line 244: Is this change intentional?

jlowe (approved):
Looks like there's an inadvertent whitespace change to a file that would otherwise be untouched, but otherwise looks good to me.

integration_tests/src/main/python/window_function_test.py
revans2: The test was originally in a different spot as I worked on it based off of the out of core sort patch.  Then when I upmerged and moved it I think I left an extra blank line.

PR #1804, opened 2021-02-24
Update UCX documentation for RX_QUEUE_LEN and Docker

abellina:
…e old README.

This incorporates feedback received, and `RX_QUEUE_LEN` recommendation in the configs.


docs/get-started/getting-started-on-prem.md
```diff
@@ -668,6 +684,9 @@ non-standard location.
   The main difference between get and put is the direction of transfer. A send operation under 
   `get_zcopy` will really be `RDMA READ` from the receiver, whereas the same send will be 
   `RDMA_WRITE` from the sender if `put_zcopy` is utilized.
+- `UCX_RC_RX_QUEUE_LEN=1024` and `UCX_UD_RX_QUEUE_LEN=1024`: Length of receive queue for `rc` and 
+  `ud` transports respectively. The length change is recommended as it has shown better performance
```
jlowe, line 688: `ud` transport is not defined, and it's not clear why they matter if the user is not requesting those transports in `UCX_TLS`.

abellina: @yosefe, as far as I understand, `ud` is still required whenever `rc` is set, is that correct?

yosefe: yes, it is used implicitly for address exchange, when the connection establishment is to worker address. it would not be used if we use dc or if we use client/server based connection establishment.
We can use a more generic setting `UCX_IB_QUEUE_LEN=1024` to affect both and cause less confusion

yosefe: sorry, had a type: it should be `UCX_IB_RX_QUEUE_LEN` 😥

abellina:
@sameerz moved things around as discussed.

docs/get-started/getting-started-on-prem.md
```diff
@@ -376,8 +376,7 @@ $SPARK_HOME/bin/spark-shell --master yarn \
   --files $SPARK_RAPIDS_DIR/getGpusResources.sh
   --jars ${SPARK_CUDF_JAR},${SPARK_RAPIDS_PLUGIN_JAR}
 ```
-  
-## Example Join Operation
+  ## Example Join Operation
```
jlowe, line 379: Indentation intentional?

docs/get-started/getting-started-on-prem.md
```diff
@@ -393,302 +392,14 @@ and `GpuColumnarExchange`.  Those correspond to operations that run on the GPU.
...
-- `spark.rapids.shuffle.ucx.bounceBuffers.device.count`
-- `spark.rapids.shuffle.ucx.bounceBuffers.host.count`
-- `spark.rapids.shuffle.ucx.bounceBuffers.size`
-
-#### Spillable Store
-This setting controls the amount of host memory (RAM) that can be utilized to spill GPU blocks when
-the GPU is out of memory, before going to disk. Please verify the [defaults](../configs.md).
-- `spark.rapids.memory.host.spillStorageSize`
+You can find out how to enable the accelerated shuffle in the 
+[RAPIDS Shuffle Manager](../additional-functionality/rapids-shuffle.md) documentation.
```
jlowe, line 402:
```suggestion
[RAPIDS Shuffle Manager documentation](../additional-functionality/rapids-shuffle.md).
```

jlowe (commented):
I've noticed we're a bit inconsistent about how we reference the shuffle plugin.  Sometimes it's RAPIDS shuffle.  Other times it's RapidsShuffleManager.  Yet other times it's RAPIDS Shuffle Manager.  We should pick one and be consistent.

jlowe (approved):

abellina:
Lets hold on this for a second. I am seeing another issue, and trying to find out if it's related.

docs/additional-functionality/rapids-shuffle.md
```diff
@@ -0,0 +1,321 @@
...
+in these scenarios:
+  - _GPU-to-GPU_: Shuffle blocks that were able to fit in GPU memory.
+  - _Host-to-GPU_ and _Disk-to-GPU_: Shuffle blocks that spilled to host (or disk) but will be manifested 
+  in the GPU in the downstream Spark task.
+  
+### System Setup
+
+In order to enable the _RAPIDS Shuffle Manager_, UCX user-space libraries and its dependencies must 
+be installed on the host and within Docker containers. A host has additional requirements, like the 
+MLNX_OFED driver and `nv_peer_mem` kernel module.
```
sameerz, line 42: Small nit, no need to italicize many of the terms: 

```suggestion
The RAPIDS Shuffle Manager is an implementation of the `ShuffleManager` interface in Apache Spark
that allows custom mechanisms to exchange shuffle data. It has two components: a spillable cache, 
and a transport that can utilize Remote Direct Memory Access (RDMA) and high-bandwidth transfers 
within a node that has multiple GPUs. This is possible because the plugin utilizes 
[Unified Communication X (UCX)](https://www.openucx.org/) as its transport.

- **Spillable cache**: This store keeps GPU data close by where it was produced in device memory,
but can spill in the following cases:
  - GPU out of memory: If an allocation in the GPU failed to acquire memory, spill will get triggered
    moving GPU buffers to host to allow for the original allocation to succeed.
  - Host spill store filled: If the host memory store has reached a maximum threshold 
    (`spark.rapids.memory.host.spillStorageSize`), host buffers will be spilled to disk until
    the host spill store shrinks back below said configurable threshold.
    
  Tasks local to the producing executor will short-circuit read from the cache.

- **Transport**: Handles block transfers between executors using various means like NVLink,
PCIe, Infiniband (IB), RDMA over Converged Ethernet (RoCE) or TCP, and as configured in UCX,
in these scenarios:
  - GPU-to-GPU: Shuffle blocks that were able to fit in GPU memory.
  - Host-to-GPU and Disk-to-GPU: Shuffle blocks that spilled to host (or disk) but will be manifested 
  in the GPU in the downstream Spark task.
  
### System Setup

In order to enable the RAPIDS Shuffle Manager, UCX user-space libraries and its dependencies must 
be installed on the host and within Docker containers. A host has additional requirements, like the 
MLNX_OFED driver and `nv_peer_mem` kernel module.
```

abellina:
> Lets hold on this for a second. I am seeing another issue, and trying to find out if it's related.

Issues were not related to the config recommendations proposed here.

abellina:
build

jlowe (approved):

sameerz (approved):

PR #1812, opened 2021-02-25
allow spilled buffers to be unspilled

rongou:


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBuffer.scala
```diff
@@ -94,6 +112,11 @@ trait RapidsBuffer extends AutoCloseable {
 
   val spillCallback: RapidsBuffer.SpillCallback
 
+  val unspillCallback: RapidsBuffer.UnspillCallback
```
revans2, line 115: Can we just combine this with spillCallback. We can detect from the API if the data is going to the `DEVICE` or not, and it would make adding in metrics to show unspill behavior a lot cleaner.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferCatalog.scala
```diff
@@ -235,9 +235,28 @@ object RapidsBufferCatalog extends Logging with Arm {
    * Lookup the buffer that corresponds to the specified buffer ID and acquire it.
    * NOTE: It is the responsibility of the caller to close the buffer.
    * @param id buffer identifier
+   * @param unspill if true, unspill the buffer to device storage
    * @return buffer that has been acquired
    */
-  def acquireBuffer(id: RapidsBufferId): RapidsBuffer = singleton.acquireBuffer(id)
+  def acquireBuffer(id: RapidsBufferId, unspill: Boolean = false): RapidsBuffer = {
```
revans2, line 241: I think I would prefer to have a completely separate API for the unspill case instead of a boolean flag. It would also let us change the type signature so the return type is not just a `RapidsBuffer`, but one that includes `getColumnarBatch`.  That way we can remove `getColumnarBatch` from `RapidsBuffer` and let the type checker verify that we called the correct API to unspill when we want to convert it to a batch.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDiskStore.scala
```diff
@@ -36,31 +36,37 @@ class RapidsDiskStore(
...
-        sharedBufferFiles.computeIfAbsent(id, _ => id.getDiskPath(diskBlockManager))
-      } else {
-        id.getDiskPath(diskBlockManager)
-      }
-      val fileOffset = if (id.canShareDiskPaths) {
-        // only one writer at a time for now when using shared files
-        path.synchronized {
-          copyBufferToPath(hostBuffer, path, append = true)
+    if (incoming.alreadySpilledToDisk) {
+      logDebug(s"Respilling $incoming ${incoming.id} of size ${incoming.size} to disk")
```
revans2, line 40: nit: Could we update this because we didn't actually move any bytes to spill, and it might be good that it was more obvious in the log.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -37,30 +37,36 @@ class RapidsGdsStore(
...
-      } else {
-        id.getDiskPath(diskBlockManager)
-      }
-      // When sharing files, append to the file; otherwise, write from the beginning.
-      val fileOffset = if (id.canShareDiskPaths) {
-        // only one writer at a time for now when using shared files
-        path.synchronized {
-          CuFile.appendDeviceBufferToFile(path, deviceBuffer)
+    if (other.alreadySpilledToDisk) {
+      logDebug(s"Respilling $other ${other.id} of size ${other.size} to disk")
```
revans2, line 41: Same comment here about the log message

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShuffleBufferCatalog.scala
```diff
@@ -209,7 +205,7 @@ class ShuffleBufferCatalog(
    */
   def acquireBuffer(tableId: Int): RapidsBuffer = {
     val shuffleBufferId = getShuffleBufferId(tableId)
-    acquireBuffer(shuffleBufferId)
+    acquireBuffer(shuffleBufferId, unspill = true)
```
revans2, line 208: I would expect this to be false by default.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBuffer.scala
```diff
@@ -94,6 +112,11 @@ trait RapidsBuffer extends AutoCloseable {
...
+
+  /** The file offset of this buffer in bytes if it has already been written to disk. */
+  val fileOffset: Option[Long]
```
jlowe, line 118: The point of `RapidsBuffer` is to abstract details of different buffer types, and this is specific to a single buffer type (i.e.: one stored to disk).

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferCatalog.scala
jlowe: Agree, I think performing an unspill on `acquireBuffer` is inappropriate.  There may be cases in the future where we want to examine the location of a buffer without triggering a device copy.  However I'm not sure I agree with removing `getColumnarBatch` from `RapidsBuffer`.

IMO the unspill behavior should occur in `getColumnarBatch` since that is the only buffer method that _requires_ the data to be in device memory to satisfy the request.  We could add a separate unspill API if there's a use-case for it, but I would expect most users to only call the existing `getColumnarBatch` to trigger that behavior.  They aren't going to care if the buffer had to unspill to satisfy the request nor want to deal with the logic of doing the unspill themselves before calling `getColumnarBatch`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDiskStore.scala
```diff
@@ -36,31 +36,37 @@ class RapidsDiskStore(
...
-      val path = if (id.canShareDiskPaths) {
-        sharedBufferFiles.computeIfAbsent(id, _ => id.getDiskPath(diskBlockManager))
-      } else {
-        id.getDiskPath(diskBlockManager)
-      }
-      val fileOffset = if (id.canShareDiskPaths) {
-        // only one writer at a time for now when using shared files
-        path.synchronized {
-          copyBufferToPath(hostBuffer, path, append = true)
+    if (incoming.alreadySpilledToDisk) {
```
jlowe, line 39: Checking for a buffer already being spilled in the disk store seems backwards to me.   We'd have to put this type of check in every type of buffer store except the device store, since it's not guaranteed that the spill went from device to disk.  It could be in host memory or some other store we invent later.

IMO a cleaner design is to check for the already-been-spilled logic in the one place that realy needs to do it: the device store.  The device store can simply free the buffer if it's already been spilled to _any store_, rather than checking for this in every other store except the device store.

jlowe (changes requested):
I think it's worth meeting to hash out a design for this.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDiskStore.scala
revans2: I would be fine with that too.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShuffleBufferCatalog.scala
```diff
@@ -190,15 +190,11 @@ class ShuffleBufferCatalog(
...
    */
-  def acquireBuffer(id: ShuffleBufferId): RapidsBuffer = {
-    val buffer = catalog.acquireBuffer(id)
-    // Shuffle buffers that have been read are less likely to be read again,
-    // so update the spill priority based on this access
-    val spillPriority = SpillPriorities.getShuffleOutputBufferReadPriority
-    buffer.setSpillPriority(spillPriority)
-    buffer
+  def acquireBuffer(id: ShuffleBufferId, unspill: Boolean = false): RapidsBuffer = {
+    RapidsBufferCatalog.acquireBuffer(id, unspill)
```
abellina, line 197: have you found that the "less likely" piece here is true for the common case? I see we are not updating the priority anymore.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/RapidsCachingReader.scala
```diff
@@ -142,7 +142,7 @@ class RapidsCachingReader[K, C](
       try {
         val cachedIt = cachedBufferIds.iterator.map(bufferId => {
           GpuSemaphore.acquireIfNecessary(context)
-          val cb = withResource(catalog.acquireBuffer(bufferId)) { buffer =>
+          val cb = withResource(catalog.acquireBuffer(bufferId, unspill = true)) { buffer =>
```
abellina, line 145: should this iterator return a set of batches that are "to be" unspilled later. Giving the coalesce batches that is pulling from it a chance to batch the unspills in a coherent way?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBuffer.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDiskStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShuffleBufferCatalog.scala
rongou: No longer needed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBuffer.scala
rongou: Removed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDiskStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShuffleBufferCatalog.scala
rongou: Left the priority update in for now.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/RapidsCachingReader.scala
rongou: Yes, batched spilling can be a further optimization.

rongou:
@revans2 @jlowe @abellina this is more or less working now. I still need to update the tests, and also add some tests for the new behavior. Please take another look.

jlowe (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferCatalog.scala
```diff
@@ -33,77 +33,136 @@ import org.apache.spark.sql.rapids.RapidsDiskBlockManager
...
-      if (buffer.addReference()) {
-        return buffer
+      val buffers = bufferMap.get(id)
+      if (isEmpty(buffers)) {
+        throw new NoSuchElementException(s"Cannot locate buffers associated with ID: $id")
       }
+      buffers.foreach(_.foreach(buffer =>
+        if (buffer.addReference()) {
+          return buffer
+        }
```
jlowe, line 54: If we fail to acquire the most appropriate buffer then the buffer was in the process of being freed, meaning the mapping is in the process of being updated.  We need to fully loop back around and fetch a new sequence of buffers and try to acquire the first one as it's likely the sequence has changed.  Only the first one is important for this method.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferCatalog.scala
```diff
@@ -33,77 +33,136 @@ import org.apache.spark.sql.rapids.RapidsDiskBlockManager
...
+
+  /**
+   * Check if the buffer that corresponds to the specified buffer ID is stored in a slower storage
+   * tier.
+   *
+   * @param id   buffer identifier
+   * @param tier storage tier to check
+   * @return true if the buffer is stored in multiple tiers
+   */
+  def isBufferBackedUp(id: RapidsBufferId, tier: StorageTier): Boolean = {
```
jlowe, line 87:
```suggestion
  def isBufferSpilled(id: RapidsBufferId, tier: StorageTier): Boolean = {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -304,11 +305,48 @@ abstract class RapidsBufferStore(
...
+    override def getMemoryBuffer: MemoryBuffer = materializeMemoryBuffer
+
+    override def getDeviceMemoryBuffer: DeviceMemoryBuffer = {
+      deviceRapidsBuffer match {
+        case Some(rapidsBuffer) => rapidsBuffer.getDeviceMemoryBuffer
+        case None =>
+          catalog.acquireBuffer(id, DEVICE) match {
+            case Some(acquiredBuffer) =>
+              deviceRapidsBuffer = Some(acquiredBuffer)
+              deviceRapidsBuffer.get.getDeviceMemoryBuffer
```
jlowe, line 317: This needs to close `acquiredBuffer` after calling `getDeviceMemoryBuffer` on it, otherwise we're unnecessarily pinning the device memory even after the caller releases the resulting `DeviceMemoryBuffer`.  It will never be closed until this buffer is closed, meaning unspill becomes a one-way ticket, never to be spilled again.  That's not what we want and will lead to OOM situations that are avoidable.

The `DeviceMemoryBuffer` being wielded by the caller and eventually closed by the caller will prevent the memory from being released as long as it hasn't been closed.  That's why we don't need to track the RapidsDeviceBuffer instance in this buffer instance.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -304,11 +305,48 @@ abstract class RapidsBufferStore(
...
+              }
+              try {
+                catalog.registerNewBuffer(newBuffer)
+              } catch {
+                case e: IllegalStateException =>
+                  // A device buffer is already registered.
+                  logDebug(s"Error registering new buffer $newBuffer", e)
+                  newBuffer.free()
+                  deviceRapidsBuffer = catalog.acquireBuffer(id, DEVICE)
+                  return deviceRapidsBuffer.get.getDeviceMemoryBuffer
```
jlowe, line 331: This needs to loop back around and try to acquire from the beginning to cover the case where it still returns `None`.  Yes, should be incredibly rare in practice, but theoretically possible if this thread gets indefinitely postponed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -304,11 +305,48 @@ abstract class RapidsBufferStore(
...
+              }
+              if (newBuffer.addReference()) {
+                deviceRapidsBuffer = Some(newBuffer)
+                deviceRapidsBuffer.get.getDeviceMemoryBuffer
+              } else {
```
jlowe, line 336: Same problem of locking the device buffer in place almost indefinitely which can trigger OOM in otherwise avoidable situations.

jlowe (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
rongou: Fixed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -304,6 +304,38 @@ abstract class RapidsBufferStore(
       }
     }
 
+    override def getMemoryBuffer: MemoryBuffer = internalGetMemoryBuffer
+
+    override def getDeviceMemoryBuffer: DeviceMemoryBuffer = {
+      while (true) {
```
rongou, line 310: Not sure if we need to limit the number of retries as done in the buffer catalog?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBuffer.scala
```diff
@@ -107,14 +107,35 @@ trait RapidsBuffer extends AutoCloseable {
...
+   * Materialize the memory buffer from the underlying storage.
+   *
+   * If the buffer resides in device or host memory, only reference count is incremented.
+   * If the buffer resides in secondary storage, a new host or device memory buffer is created, with
+   * the data copied to the new buffer.
+   * The caller must have successfully acquired the buffer beforehand.
+   * @see [[addReference]]
+   * @note It is the responsibility of the caller to close the buffer.
+   */
+  def materializeMemoryBuffer: MemoryBuffer
```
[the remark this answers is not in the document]
jlowe: Can you elaborate on why we can't have `copyBuffer` take both the `RapidsBuffer` being copied and the corresponding `MemoryBuffer` containing the data to copy?  Each specific buffer store knows how to manifest the data for its buffer objects without calling a public method on the buffer object.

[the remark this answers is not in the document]
rongou: As far as I can tell, only a rapids buffer knows how to materialize the underlying memory buffer. A buffer store only knows how to create the corresponding rapids buffer. Once it's created, the store has the same access level to its methods as anyone else.

In java we can mark the method package private to limit its access to within the package. Looks like in scala you can also do `private[rapids]`, but the `rapids` package is fairly wide, I'm not sure it helps that much.

jlowe (changes requested):
Looking better.  I think there are still some issues around new buffer registration and corresponding races along with new tests that need to be written, both for the catalog and for the new "unspill" behavior.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -304,6 +304,38 @@ abstract class RapidsBufferStore(
...
+        catalog.acquireBuffer(id, DEVICE) match {
+          case Some(buffer) =>
+            withResource(buffer) { buffer =>
+              return buffer.getDeviceMemoryBuffer
+            }
+          case _ =>
+            val newBuffer = {
+              logDebug(s"Unspilling $this $id to $DEVICE")
+              RapidsBufferCatalog.getDeviceStorage.copyBuffer(this, Cuda.DEFAULT_STREAM)
```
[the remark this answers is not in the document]
rongou: Yeah we were calling register twice, so I removed it from within `copyBuffer`. Moved it back there and also try to handle the exception there.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -304,6 +304,38 @@ abstract class RapidsBufferStore(
...
+            }
+            try {
+              catalog.registerNewBuffer(newBuffer)
```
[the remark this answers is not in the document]
rongou: Fixed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDeviceMemoryStore.scala
```diff
@@ -29,10 +29,27 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
...
       stream: Cuda.Stream): RapidsBufferBase = {
-    throw new IllegalStateException("should not be spilling to device memory")
+    val deviceBuffer = {
+      other.internalGetMemoryBuffer match {
+        case d: DeviceMemoryBuffer => d
+        case h: HostMemoryBuffer =>
+          try {
+            val deviceBuffer = DeviceMemoryBuffer.allocate(other.size)
+            logDebug(s"copying from host $h to device $deviceBuffer")
+            deviceBuffer.copyFromHostBuffer(h)
```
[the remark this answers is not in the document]
rongou: Looks we do have a version with stream. Fixed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -64,6 +61,10 @@ class RapidsGdsStore(
     }
   }
 
+  override protected def getMemoryBuffer(buffer: RapidsBufferBase): MemoryBuffer = {
+    buffer.asInstanceOf[RapidsGdsBuffer].materializeMemoryBuffer
```
jlowe, line 65: Looking at `RapidsGdsBuffer.materializeMemoryBuffer`, the code uses more state of the parent store than the buffer itself.  That's why I think it makes sense to not have a `materializeBuffer` interface on the buffer but rather handle it in the store directly.  Here's an example that should work after `RapidsGdsBuffer` exposes `fileOffset` as a `val` or otherwise has an accessor to it:
```suggestion
    buffer match {
      case buf: RapidsGdsBuffer =>
        val path = if (buf.id.canShareDiskPaths) {
          sharedBufferFiles.get(buf.id)
        } else {
          buf.id.getDiskPath(diskBlockManager)
        }
        closeOnExcept(DeviceMemoryBuffer.allocate(buf.size)) { devBuffer =>
          CuFile.readFileToDeviceBuffer(devBuffer, path, buf.fileOffset)
          logDebug(s"Created device buffer for $path ${buf.fileOffset}:${buf.size} via GDS")
          devBuffer
        }
      case _ => throw new IllegalArgumentException(s"not a GDS buffer: $buffer")
    }
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsHostMemoryStore.scala
```diff
@@ -71,12 +71,14 @@ class RapidsHostMemoryStore(
     (buffer, true)
   }
 
-  override protected def createBuffer(
-      other: RapidsBuffer,
+  override protected def getMemoryBuffer(buffer: RapidsBufferBase): MemoryBuffer = {
+    buffer.asInstanceOf[RapidsHostMemoryBuffer].materializeMemoryBuffer
```
jlowe, line 75:
```suggestion
    buffer.getMemoryBuffer
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsHostMemoryStore.scala
```diff
@@ -123,7 +125,7 @@ class RapidsHostMemoryStore(
       extends RapidsBufferBase(id, size, meta, spillPriority, spillCallback) {
     override val storageTier: StorageTier = StorageTier.HOST
 
-    override def getMemoryBuffer: MemoryBuffer = {
+    override def materializeMemoryBuffer: MemoryBuffer = {
```
jlowe, line 128:
```suggestion
    override def getMemoryBuffer: MemoryBuffer = {
```

jlowe (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -304,6 +328,36 @@ abstract class RapidsBufferStore(
       }
     }
 
+    override def getMemoryBuffer: MemoryBuffer = materializeMemoryBuffer
+
```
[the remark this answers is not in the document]
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -304,6 +328,36 @@ abstract class RapidsBufferStore(
...
+          case Some(buffer) =>
+            withResource(buffer) { buffer =>
+              return buffer.getDeviceMemoryBuffer
+            }
+          case _ =>
+            try {
+              val newBuffer = {
+                logDebug(s"Unspilling $this $id to $DEVICE")
+                RapidsBufferCatalog.getDeviceStorage.copyBuffer(
+                  this, materializeMemoryBuffer, Cuda.DEFAULT_STREAM)
```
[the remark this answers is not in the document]
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDeviceMemoryStore.scala
```diff
@@ -28,11 +28,31 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
...
+          }
+        case b => throw new IllegalStateException(s"Unrecognized buffer: $b")
+      }
+    }
+    new RapidsDeviceMemoryBuffer(other.id, other.size, other.meta, None,
+      deviceBuffer, other.getSpillPriority, other.spillCallback)
+  }
+
+  override protected def getMemoryBuffer(buffer: RapidsBufferBase): MemoryBuffer = {
+    buffer.asInstanceOf[RapidsDeviceMemoryBuffer].materializeMemoryBuffer
```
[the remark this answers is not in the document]
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDeviceMemoryStore.scala
```diff
@@ -163,11 +183,13 @@ class RapidsDeviceMemoryStore(catalog: RapidsBufferCatalog = RapidsBufferCatalog
...
     }
 
-    override def getMemoryBuffer: MemoryBuffer = {
+    override def getDeviceMemoryBuffer: DeviceMemoryBuffer = {
       contigBuffer.incRefCount()
       contigBuffer
     }
 
+    override def materializeMemoryBuffer: MemoryBuffer = getDeviceMemoryBuffer
+
```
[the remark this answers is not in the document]
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDiskStore.scala
```diff
@@ -64,6 +62,10 @@ class RapidsDiskStore(
     }
   }
 
+  override protected def getMemoryBuffer(buffer: RapidsBufferBase): MemoryBuffer = {
+    buffer.asInstanceOf[RapidsDiskBuffer].materializeMemoryBuffer
```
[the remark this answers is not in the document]
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDiskStore.scala
```diff
@@ -98,7 +100,7 @@ class RapidsDiskStore(
 
     override val storageTier: StorageTier = StorageTier.DISK
 
-    override def getMemoryBuffer: MemoryBuffer = synchronized {
+    override def materializeMemoryBuffer: MemoryBuffer = synchronized {
```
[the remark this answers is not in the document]
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsHostMemoryStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsHostMemoryStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -304,6 +320,34 @@ abstract class RapidsBufferStore(
...
+          case Some(buffer) =>
+            withResource(buffer) { buffer =>
+              return buffer.getDeviceMemoryBuffer
+            }
+          case _ =>
+            try {
+              val newBuffer = {
+                logDebug(s"Unspilling $this $id to $DEVICE")
+                RapidsBufferCatalog.getDeviceStorage.copyBuffer(
+                  this, getMemoryBuffer, Cuda.DEFAULT_STREAM)
```
rongou, line 335: Actually this doesn't work for the GDS case as it causes an infinite regress.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDeviceMemoryStore.scala
```diff
@@ -28,11 +28,31 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
...
+        case d: DeviceMemoryBuffer => d
+        case h: HostMemoryBuffer =>
+          try {
+            val deviceBuffer = DeviceMemoryBuffer.allocate(other.size)
+            logDebug(s"copying from host $h to device $deviceBuffer")
+            deviceBuffer.copyFromHostBuffer(h, stream)
+            deviceBuffer
+          } finally {
+            h.close()
+          }
```
jlowe, line 46:
```suggestion
          withResource(h) { _ =>
            closeOnExcept(DeviceMemoryBuffer.allocate(other.size)) { deviceBuffer =>
              logDebug(s"copying from host $h to device $deviceBuffer")
              deviceBuffer.copyFromHostBuffer(h, stream)
              deviceBuffer
            }
          }
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -201,7 +209,15 @@ abstract class RapidsBufferStore(
    * @param stream CUDA stream to use or null
```
jlowe, line 209: The new `memoryBuffer` argument needs to be documented and mention that this method will close it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
jlowe: It doesn't work because this calls the wrong `getMemoryBuffer`.  This is calling `getMemoryBuffer` on the `RapidsBuffer` rather than on the buffer store method passing `this` as the argument.  In retrospect, `getMemoryBuffer` is a poor name choice on the store method because it's easy to confuse with the buffer method.  We can use the original name of `materializeMemoryBuffer` for the buffer store method so it's harder to get confused.

Also if you think it makes more sense to move it back in the RapidsBuffer that's OK with me as long as it remains a protected method.  Externally it doesn't matter much either way whether this method lives in the store or in the buffer as long as code outside of the store cannot call it.  I don't want to allow outside code to see it since it will be confusing which one they should call, and we always want the unspill behavior for external code trying to manifest a device buffer.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -128,10 +128,18 @@ abstract class RapidsBufferStore(
...
-    newBuffer
+  def copyBuffer(buffer: RapidsBuffer, memoryBuffer: MemoryBuffer, stream: Cuda.Stream)
+  : RapidsBufferBase = {
+    val newBuffer = createBuffer(buffer, memoryBuffer, stream)
+    try {
+      buffers.add(newBuffer)
+      catalog.registerNewBuffer(newBuffer)
+      newBuffer
+    } catch {
+      case e: IllegalStateException =>
```
jlowe, line 139: A bit of future-proofing here since we don't actually care what type of exception it is:
```suggestion
      case e: Exception =>
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -304,6 +320,34 @@ abstract class RapidsBufferStore(
...
+                  this, getMemoryBuffer, Cuda.DEFAULT_STREAM)
+              }
+              if (newBuffer.addReference()) {
+                withResource(newBuffer) { newBuffer =>
+                  return newBuffer.getDeviceMemoryBuffer
+                }
+              }
+            } catch {
+              case e: IllegalStateException =>
+                logDebug("Error unspilling to device", e)
```
jlowe, line 344: I don't think we want to log the full exception stack here and call it an error when it is expected and will be retried.  If it does occur it will add a lot of spam to the debug log.  Maybe something along the line of `s"Lost device buffer registration race for buffer $id, retrying..."`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
rongou: Ok, added `materializeMemoryBuffer` as a protected method in `RapidsBufferBase`. Only `RapidsGdsBuffer` needs to override it.

rongou (commented):
Let me know if the API changes look ok. I'll start working on the tests.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
jlowe: I think the wrong exception was updated.  Instead of this being updated the exception we're suppressing later in this file was updated, see the other comment about that below.  This should still be updated since in this case we're not suppressing and no matter what the error is we want to free the buffer.

jlowe (commented):
You can start on the tests.  We should still clean up the redundant materialization methods, but that shouldn't affect the tests since it's not an exposed method.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
rongou: Should be fixed now.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -198,10 +208,22 @@ abstract class RapidsBufferStore(
...
-  protected def createBuffer(buffer: RapidsBuffer, stream: Cuda.Stream): RapidsBufferBase
+  protected def createBuffer(buffer: RapidsBuffer, memoryBuffer: MemoryBuffer, stream: Cuda.Stream)
+  : RapidsBufferBase
+
+  /**
+   * Materialize the underlying memory buffer from the specified Rapids buffer.
+   * @param buffer a buffer managed by this store
+   * @return underlying memory buffer
+   */
+  protected def materializeMemoryBuffer(buffer: RapidsBufferBase): MemoryBuffer = {
```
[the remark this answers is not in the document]
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -304,6 +339,32 @@ abstract class RapidsBufferStore(
...
+              logDebug(s"Unspilling $this $id to $DEVICE")
+              val newBuffer = RapidsBufferCatalog.getDeviceStorage.copyBuffer(
+                this, materializeMemoryBuffer, Cuda.DEFAULT_STREAM)
+              if (newBuffer.addReference()) {
+                withResource(newBuffer) { newBuffer =>
+                  return newBuffer.getDeviceMemoryBuffer
+                }
+              }
+            } catch {
+              case _: Exception =>
```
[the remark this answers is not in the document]
rongou: Fixed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsHostMemoryStore.scala
```diff
@@ -71,12 +71,10 @@ class RapidsHostMemoryStore(
     (buffer, true)
   }
 
-  override protected def createBuffer(
-      other: RapidsBuffer,
+  override protected def createBuffer(other: RapidsBuffer, otherBuffer: MemoryBuffer,
       stream: Cuda.Stream): RapidsBufferBase = {
     val (hostBuffer, isPinned) = allocateHostBuffer(other.size)
```
jlowe, line 76: We leak `otherBuffer` if this throws.  Should start the method with `withResource(otherBuffer) { _ =>` to ensure it is always closed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsDiskStore.scala
```diff
@@ -33,10 +33,8 @@ class RapidsDiskStore(
     extends RapidsBufferStore(StorageTier.DISK, catalog) {
   private[this] val sharedBufferFiles = new ConcurrentHashMap[RapidsBufferId, File]
 
-  override def createBuffer(
-      incoming: RapidsBuffer,
+  override protected def createBuffer(incoming: RapidsBuffer, incomingBuffer: MemoryBuffer,
       stream: Cuda.Stream): RapidsBufferBase = {
-    val incomingBuffer = incoming.getMemoryBuffer
     try {
```
jlowe, line 38: Can use `withResource(incomingBuffer) { _ =>` here

jlowe (commented):
Thanks for updating @rongou!  The main code looks good to me.  From my perspective all that remains are the missing tests.

tests/src/test/scala/com/nvidia/spark/rapids/RapidsBufferCatalogSuite.scala
```diff
@@ -79,20 +79,6 @@ class RapidsBufferCatalogSuite extends FunSuite with MockitoSugar {
     assertResult(expectedMeta)(meta)
   }
 
-  test("update buffer map only updates for faster tier") {
```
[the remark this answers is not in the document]
rongou: Done.

rongou (commented):
Ok, fixed the existing tests and added new tests for the buffer catalog. Had to plumb through device storage to make tests happy. Please take another look.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -269,6 +270,7 @@ abstract class RapidsBufferStore(
       override val meta: TableMeta,
       initialSpillPriority: Long,
       override val spillCallback: RapidsBuffer.SpillCallback,
+      deviceStorage: RapidsDeviceMemoryStore = RapidsBufferCatalog.getDeviceStorage,
```
jlowe, line 273: New default args should be added at the end, and this helps make it consistent with the way it was added to the store.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -40,7 +40,8 @@ object RapidsBufferStore {
  */
 abstract class RapidsBufferStore(
     val tier: StorageTier,
-    catalog: RapidsBufferCatalog = RapidsBufferCatalog.singleton)
+    catalog: RapidsBufferCatalog = RapidsBufferCatalog.singleton,
+    deviceStorage: RapidsDeviceMemoryStore = RapidsBufferCatalog.getDeviceStorage)
```
jlowe, line 44: Rather than tacking on the device store everywhere, would it be simpler to promote `getDeviceStorage` to be an overridable method on the catalog rather than a static method?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
rongou: Tried this, but there are a couple of places in shuffle manager that also calls `RapidsBufferCatalog.getDeviceStorage` directly, so we'd have to plumb the method through their catalogs. Not sure if it's an improvement.

jlowe: Cool, thanks for looking into it.  I agree that amount of change is probably not appropriate for this PR, and it's something we can address later.

jlowe (approved):
Thanks for all the work, @rongou!

rongou:
@abellina yes please take another look.

jlowe (dismissed):

abellina (approved):
We use `getMemoryBuffer` in the shuffle, so we shouldn't cause an unspill there (unless it is GDS), which is consistent with what we did before. The unspill flow makes sense to me with `getColumnarBatch`.

Thanks @rongou @jlowe

abellina:
build

rongou:
Added config for unspill with default to false. Please take another look.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -352,6 +352,14 @@ object RapidsConf {
     .bytesConf(ByteUnit.BYTE)
     .createWithDefault(ByteUnit.GiB.toBytes(1))
 
+  val UNSPILL = conf("spark.rapids.memory.gpu.unspill.enabled")
+    .doc("When a spilled GPU buffer is needed again, should it be unspilled, or only copied " +
+        "back into GPU memory temporarily. Unspilling may be useful for GPU buffers that are " +
+        "needed frequently, for example, broadcast variables; however, it may also increase GPU " +
+        "memory usage")
+      .booleanConf
```
jlowe, line 360: Do we want this to be an internal config for now so it doesn't show up in the docs?  Seems like we're still working out the kinks with unspill, so wondering if we should avoid advertising it until it's farther along.  I'm fine if others would rather keep it public.

jlowe (approved):
Looks OK to me other than we may want to keep the new config internal while we do more testing/refinement to unspill.

abellina (approved):
Thanks @rongou. LGTM, agree with making the config internal like @jlowe suggested.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
rongou: Done.

abellina (approved):

abellina:
build

PR #1813, opened 2021-02-25
Fix LEAD/LAG failures in Spark 3.1.1

### LINKED ISSUE #999: [BUG] test_multi_types_window_aggs_for_rows_lead_lag fails against Spark 3.1.0
**Describe the bug**
Nightly build of integration tests against Spark 3.1.0 are failing in the test_multi_types_window_aggs_for_lead_lag tests.

Run integration tests against latest Spark 3.1.0-SNAPSHOT

**Expected behavior**
Tests do not fail.

mythrocks:
Fixes #999.

Fixes class-loader related errors of LEAD/LAG operators, which caused test failures like:
```
: java.lang.VerifyError: Bad type on operand stack
Exception Details:
Location:
com/nvidia/spark/rapids/OffsetWindowFunctionMeta.<init>(Lorg/apache/spark/sql/catalyst/expressions/OffsetWindowFunction;Lcom/nvidia/spark/rapids/RapidsConf;Lscala/Option;Lcom/nvidia/spark/rapids/ConfKeysAndIncompat;)V @11: invokespecial
Reason:
Type 'org/apache/spark/sql/catalyst/expressions/OffsetWindowFunction' (current frame, stack[1]) is not assignable to 'org/apache/spark/sql/catalyst/expressions/Expression'
```

Also corrects for the reversal in offset semantics for `LAG()` expressions in Spark 3.1.1, causing `Lag.offset` to be negative.

Signed-off-by: Mithun RK <mythrocks@gmail.com>

### DIFF
--- a/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/OffsetWindowFunctionMeta.scala
+++ b/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/OffsetWindowFunctionMeta.scala
@@ -37,9 +37,8 @@ abstract class OffsetWindowFunctionMeta[INPUT <: OffsetWindowFunction] (
     parent: Option[RapidsMeta[_, _, _]],
     rule: DataFromReplacementRule)
   extends ExprMeta[INPUT](expr, conf, parent, rule) {
-  val input: BaseExprMeta[_] = GpuOverrides.wrapExpr(expr.input, conf, Some(this))
-  // For LAG() expressions, negate the offset value.
-  val adjustedOffset: Expression = {
+  lazy val input: BaseExprMeta[_] = GpuOverrides.wrapExpr(expr.input, conf, Some(this))
+  lazy val adjustedOffset: Expression = {
     expr match {
       case lag: Lag =>
         GpuOverrides.extractLit(lag.offset) match {
@@ -61,8 +60,23 @@ abstract class OffsetWindowFunctionMeta[INPUT <: OffsetWindowFunction] (
         throw new IllegalStateException(s"$other is not a supported window function")
     }
   }
-  val offset: BaseExprMeta[_] =
+  lazy val offset: BaseExprMeta[_] =
     GpuOverrides.wrapExpr(adjustedOffset, conf, Some(this))
-  val default: BaseExprMeta[_] = GpuOverrides.wrapExpr(expr.default, conf, Some(this))
-  override val childExprs: Seq[BaseExprMeta[_]] = Seq(input, offset, default)
+  lazy val default: BaseExprMeta[_] = GpuOverrides.wrapExpr(expr.default, conf, Some(this))
+
+  override val childExprs: Seq[BaseExprMeta[_]] = Seq.empty
+
+  override def tagExprForGpu(): Unit = {
+    expr match {
+      case Lead(_,_,_) => // Supported.
+      case Lag(_,_,_) =>  // Supported.
+      case other =>
+        willNotWorkOnGpu( s"Only LEAD/LAG offset window functions are supported. Found: $other")
+    }
+
+    if (GpuOverrides.extractLit(expr.offset).isEmpty) { // Not a literal offset.
+      willNotWorkOnGpu(
+        s"Only integer literal offsets are supported for LEAD/LAG. Found: ${expr.offset}")
+    }
+  }
 }
--- a/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/OffsetWindowFunctionMeta.scala
+++ b/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/OffsetWindowFunctionMeta.scala
@@ -43,7 +43,7 @@ abstract class OffsetWindowFunctionMeta[INPUT <: OffsetWindowFunction] (
     expr match {
       case lag: Lag =>
         GpuOverrides.extractLit(lag.offset) match {
-         case Some(Literal(offset:Int, IntegerType)) =>
+         case Some(Literal(offset: Int, IntegerType)) =>
             Literal(-offset, IntegerType)
          case _ =>
            throw new IllegalStateException(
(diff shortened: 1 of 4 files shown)


shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/OffsetWindowFunctionMeta.scala
```diff
@@ -0,0 +1,57 @@
...
+    conf: RapidsConf,
+    parent: Option[RapidsMeta[_, _, _]],
+    rule: DataFromReplacementRule)
+  extends ExprMeta[INPUT](expr, conf, parent, rule) {
+  val input: BaseExprMeta[_] = GpuOverrides.wrapExpr(expr.input, conf, Some(this))
+  // For LAG() expressions, negate the offset value.
+  val adjustedOffset: Expression = {
+    expr match {
+      case lag: Lag =>
+        lag.offset match {
```
revans2, line 45: Is ti possible to ever have an alias here? I think it would be cleaner to just use GpuOverrides.extractLit

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/Spark311Shims.scala
```diff
@@ -45,6 +45,21 @@ import org.apache.spark.sql.rapids.shims.spark311._
 import org.apache.spark.sql.types._
 import org.apache.spark.storage.{BlockId, BlockManagerId}
 
+abstract class OffsetWindowFunctionMeta2[INPUT <: OffsetWindowFunction] (
```
revans2, line 48: This is never used. Can it be deleted?

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/OffsetWindowFunctionMeta.scala
```diff
@@ -0,0 +1,57 @@
...
+          case Literal(v: Int, IntegerType) =>
+            Literal(-v)
+          case _ => expr.offset
```
revans2, line 48: Wouldn't this cause an error if this path was ever taken?  Can we throw an exception instead?

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/Spark311Shims.scala
mythrocks: Yes, of course. Apologies, this is sloppy.

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/OffsetWindowFunctionMeta.scala
mythrocks: Agreed. Fixed.

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/OffsetWindowFunctionMeta.scala
mythrocks: I haven't been able to make this turn into an alias in my testing. But it's best to defend against this possibility.

revans2 (approved):

mythrocks:
At the risk of blowing this up a little, we might consider a couple of changes more:
1. Add similarly robust offset checks in `com.nvidia.spark.rapids.OffsetWindowFunctionMeta`.
2. Add a `tagSelfForGpu()`, and fail GPU pushdown in case the offsets in the window are not suitable for acceleration.

revans2:
> At the risk of blowing this up a little, we might consider a couple of changes more:
> 
>     1. Add similarly robust offset checks in `com.nvidia.spark.rapids.OffsetWindowFunctionMeta`.
> 
>     2. Add a `tagSelfForGpu()`, and fail GPU pushdown in case the offsets in the window are not suitable for acceleration.

Do it

mythrocks:
Hmm... I wonder if adding `tagExprForGpu` might help. 

@revans2, is there a guarantee that `tagExprForGpu()` will be called before `OffsetWindowFunctionMeta::childExprs` is evaluated? If not, the exceptions we throw when calculating the `adjustedOffset` would've been thrown anyway.

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/OffsetWindowFunctionMeta.scala
```diff
@@ -0,0 +1,68 @@
...
+    parent: Option[RapidsMeta[_, _, _]],
+    rule: DataFromReplacementRule)
+  extends ExprMeta[INPUT](expr, conf, parent, rule) {
+  val input: BaseExprMeta[_] = GpuOverrides.wrapExpr(expr.input, conf, Some(this))
+  // For LAG() expressions, negate the offset value.
+  val adjustedOffset: Expression = {
+    expr match {
+      case lag: Lag =>
+        GpuOverrides.extractLit(lag.offset) match {
+         case Some(Literal(offset:Int, IntegerType)) =>
```
gerashegalov, line 46: nit: space
```suggestion
         case Some(Literal(offset: Int, IntegerType)) =>
```

gerashegalov (commented):
LGTM

gerashegalov (approved):

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/OffsetWindowFunctionMeta.scala
mythrocks: Thanks, @gerashegalov. Fixed.

mythrocks:
> If not, the exceptions we throw when calculating the adjustedOffset would've been thrown anyway.

Apologies for the delay. I have now added `tagExprForGpu()`, and set it up so that the expressions aren't converted for GPU on construction.

Could I please bother @gerashegalov  and @revans2 to see if this is agreeable?

revans2 (approved):

revans2:
build

mythrocks:
Thanks for the reviews, @revans2, @gerashegalov.

PR #1814, opened 2021-02-25
Struct to string casting functionality

### LINKED ISSUE #1698: [FEA] Support casting structs to string
Sub feature of #1604, but does not require cudf implementation. Can be accomplished with recursive casts and a string concatenate of the resulting columns.

rwlee:
Resolves #1698
Requires https://github.com/rapidsai/cudf/pull/7406 -- draft until that PR is merged

Inherits incompatibility of base casting -- e.g. floats -- because casting is used recursively


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -484,6 +488,59 @@ case class GpuCast(
...
+      }
+      withResource(GpuScalar.from(" ", StringType)) { separatorScalar =>
+        spaceColumn = ColumnVector.fromScalar(separatorScalar, input.getRowCount().toInt)
+      }
+      withResource(GpuScalar.from("[", StringType)) { bracketScalar =>
+        columns += ColumnVector.fromScalar(bracketScalar, input.getRowCount().toInt)
+      }
+      withResource(input.getBase().getChildColumnView(0)) { childView =>
+        columns += childView.asStrings()
+        coreColumns += childView.asStrings()
```
revans2, line 509: Casting to strings twice is expensive. We should just increment the reference count instead.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -484,6 +488,59 @@ case class GpuCast(
...
+        separatorColumn = ColumnVector.fromScalar(separatorScalar, input.getRowCount().toInt)
+      }
+      withResource(GpuScalar.from(" ", StringType)) { separatorScalar =>
+        spaceColumn = ColumnVector.fromScalar(separatorScalar, input.getRowCount().toInt)
+      }
+      withResource(GpuScalar.from("[", StringType)) { bracketScalar =>
+        columns += ColumnVector.fromScalar(bracketScalar, input.getRowCount().toInt)
+      }
+      withResource(input.getBase().getChildColumnView(0)) { childView =>
+        columns += childView.asStrings()
```
revans2, line 508: Why are we not recursively having the sub-columns cast to strings? There is a lot of formatting that this is not going to properly cover.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -484,6 +488,59 @@ case class GpuCast(
...
+        coreColumns += childView.asStrings()
+      }
+
+      for(childIndex <- 1 until input.getBase().getNumChildren()) {
+        withResource(input.getBase().getChildColumnView(childIndex)) { childView =>
+          columns += separatorColumn
+          columns += spaceColumn.mergeAndSetValidity(BinaryOp.BITWISE_AND, childView)
```
revans2, line 515: null handling of child columns is configurable in the latest versions of spark just like the brackets are, this has changed since 3.0.0. If it is not the legacy behavior, then we append "null" for null values.

 https://github.com/apache/spark/blob/4a3200b08ac3e7733b5a3dc7271d35e6872c5967/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L387

https://github.com/apache/spark/blob/4a3200b08ac3e7733b5a3dc7271d35e6872c5967/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L395

Otherwise it appears to be an empty string.

It would also be nice to have a comment explaining what is happening, because the BITWISE_AND is not obvious.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -766,7 +766,7 @@ class CastChecks extends ExprChecks {
   val mapChecks: TypeSig = none
   val sparkMapSig: TypeSig = STRING + MAP.nested(all)
 
-  val structChecks: TypeSig = none
+  val structChecks: TypeSig = STRING
```
revans2, line 769: We probably need some updates to how this checks nested types, because this currently says that we can cast a struct of anything to a string, and that is not technically true.  We are going to need to be recursive in how we check because we still want things like floats being cast to strings only working if we have enabled it.

https://github.com/NVIDIA/spark-rapids/blob/de00b8a238201334bd59bedcebb76e8d5022e2c2/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala#L530

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
rwlee: I was also leaking the second column, oops

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
rwlee: This was pretty nasty to implement because getting the struct column's child columns only returns `ColumnView`. Recursively calling `doColumnar` required copying each column to create a `ColumnVector`. These copies are unnecessary as far as I can tell, may be worth a followup PR.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
rwlee: Added some comments and there's a whole bunch of changes to support the new `3.1.0+` behavior with legacy support

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
rwlee: I've added recursive checks to the struct type checking when tagging the expression for GPU. `TypeChecks.scala` definition and `supported_ops.md` documentation will continue to believe that struct to string casting has blanket support -- should there be an extra note in the `supported_ops.md` documentation?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -485,6 +509,105 @@ case class GpuCast(
...
+    val columns: ArrayBuffer[ColumnVector] = new ArrayBuffer[ColumnVector]()
+    // coreColumns tracks the casted child columns
+    val coreColumns: ArrayBuffer[ColumnVector] = new ArrayBuffer[ColumnVector]()
+
+    try {
+      withResource(GpuScalar.from(leftBracket, StringType)) { bracketScalar =>
+        columns += ColumnVector.fromScalar(bracketScalar, input.getRowCount().toInt)
+      }
+      withResource(input.getBase().getChildColumnView(0)) { childView =>
+        withResource(childView.copyToColumnVector()) { childVector =>
```
revans2, line 527: Why do we need to copy this to a column vector? I understand because of the APIs we have right now it is needed, but it would be nice to avoid the copy if possible.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -485,6 +509,105 @@ case class GpuCast(
...
+          columns += doColumnar(GpuColumnVector.from(childVector, inputSchema(0).dataType))
+        }
+      }
+
+      if(legacyCastToString) {
```
revans2, line 532: nit there is usually a space after the if

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -485,6 +509,105 @@ case class GpuCast(
...
+            columns.toArray[ColumnView])) { fullResult =>
+            // Merge the validity of all child columns, fully null rows are null in the result
+            withResource(fullResult.mergeAndSetValidity(BinaryOp.BITWISE_OR,
+              coreColumns: _*)) { nulledResult =>
+              // Reflect the struct column's validity vector in the result
+              nulledResult.mergeAndSetValidity(BinaryOp.BITWISE_AND, input.getBase(), nulledResult)
+            }
+          }
+        }
+      } else {
```
revans2, line 569: The code in the if and else are a little long. it might be nice to either move it into separate methods, or find a way to commonize the code a bit more.

integration_tests/src/main/python/struct_test.py
```diff
@@ -57,3 +57,20 @@ def test_orderby_struct_2(data_gen):
         lambda spark : append_unique_int_col_to_df(spark, unary_op_df(spark, data_gen)),
         'struct_table',
         'select struct_table.a, struct_table.uniq_int from struct_table order by uniq_int')
+
+# conf with legacy cast to string on
+legacy_complex_types_to_string = {'spark.sql.legacy.castComplexTypesToString.enabled': 'true'}
+# @pytest.mark.parametrize('data_gen', [StructGen([["first", boolean_gen], ["second", byte_gen], ["third", short_gen], ["fourth", int_gen], ["fifth", long_gen], ["sixth", string_gen], ["seventh", date_gen], ["eighth", float_gen], ], ["ninth", double_gen], ["tenth", timestamp_gen]])], ids=idfn)
```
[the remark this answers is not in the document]
rwlee: This set of data gens include `incompat` types (float, double, timestamp), the line below it leaves them out. Because we use a recursive cast to string, we inherit the `incompat` issues and can't test them using a strict cpu and gpu comparison. This seemed like a reasonable way to leave the complete and ideal test set in, so what's actually running isn't accepted as exhaustive.

docs/supported_ops.md
```diff
@@ -18395,7 +18395,7 @@ and the accelerator produces the same result.
 <td> </td>
 <td> </td>
 <td> </td>
-<td><b>NS</b></td>
+<td>S</td>
```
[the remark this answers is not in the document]
rwlee: There is already a note addressing that in the header of the cast section. Should I include it inside the table as well?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
rwlee: Ideally we would not copy this column vector, which is why I had been using `toString` in earlier iterations. How might I avoid this copy?

docs/supported_ops.md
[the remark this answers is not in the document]
revans2: Yes, I personally think it would be good to have it documented here so it is clear to the end user.  We can do that in the TypeSig with a `.psNote`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
revans2: The signature of `doColumnar` takes a `GpuColumnVector` It uses the `GpuColumnVector` to extract its Spark data type so it can do the cast. A `ColumnVector` is a `ColumnView`. As such I would propose that we insert another API in between that we could use for recursive calls.

```
override def doColumnar(input: GpuColumnVector): ColumnVector =
    doColumnar(input.getBase, input.dataType())

private def doColumnar(input: ColumnView, sparkType: DataType): COlumnVector = {
   (sparkType, dataType) match {
      ...
   }
}
```

Then you can call the new doColumnar without making  copy of the data.

```
withResource(input.getBase().getChildColumnView(0)) { childView =>
  columns += doColumnar(childView, inputSchema(0).dataType)
}
```

integration_tests/src/main/python/struct_test.py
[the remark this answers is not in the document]
revans2: because `data_gen` is parameterized, I personally would prefer a version of it with the failing types marks with xfail and reason on the xfail pointing to the issue(s) that are causing it to fail. I am fine with the commented out code so long as it also has a comment explaining why it is commented out, and ideally a link to the issue(s) that need to be fixed for it.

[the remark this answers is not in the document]
rwlee: Changed to xfail for the 3 failing types.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
rwlee: There a few things around decimal casting that didn't translate nicely. In some cases they want to do a zero-copy, so I've filtered them out in the `GpuColumnVector` wrapper API. In those cases, if those cases are found when doing a `ColumnView` cast, it's copied to a ColumnVector. Reworking the decimal support to be `ColumnView` friendly version can be done in a follow up PR if necessary, but it seemed a bit out of scope and was non-trivial.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
rwlee: They approaches are different enough that I opted to split the code into separate methods. There's a lot of similarities, but how nulls are handled is too different to cleanly parameterize.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -90,14 +97,27 @@ class CastExprMeta[INPUT <: CastBase](
         "The first step may lead to precision loss. To enable this operation on the GPU, set " +
         s" ${RapidsConf.ENABLE_CAST_STRING_TO_FLOAT} to true.")
     }
+    if (fromDataType.isInstanceOf[StructType]) {
+      val checks = rule.getChecks.asInstanceOf[CastChecks]
```
revans2, line 101: When the tests run I get an error here
```
E                   py4j.protocol.Py4JJavaError: An error occurred while calling o83756.collectToPython.
E                   : java.lang.ClassCastException: scala.Some cannot be cast to com.nvidia.spark.rapids.CastChecks
E                   	at com.nvidia.spark.rapids.CastExprMeta.recursiveTagExprForGpuCheck(GpuCast.scala:100)
E                   	at com.nvidia.spark.rapids.CastExprMeta.tagExprForGpu(GpuCast.scala:44)
E                   	at com.nvidia.spark.rapids.BaseExprMeta.tagSelfForGpu(RapidsMeta.scala:761)
E                   	at com.nvidia.spark.rapids.RapidsMeta.tagForGpu(RapidsMeta.scala:267)
```

Please update it to 

```
val checks = rule.getChecks.get.asInstanceOf[CastChecks]
```

revans2 (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
revans2: Could we have a follow on issue to clean this up? The issue appears to be because of incRefCount, and there should be ways for us to work around that without too much difficulty.

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
rwlee: #2043

revans2 (approved):

Issue #1817, opened 2021-02-25
[BUG] Developer build with local modifications fails during verify phase


jlowe:
**Describe the bug**
When performing a `mvn verify` build as a developer, the build will fail if there are any checkouts.  This check was added by #1809 but should only happen on precommit builds, not developer builds.  This behavior should be controlled via a Maven profile that is not enabled by default.

**Steps/Code to reproduce bug**
Make some changes and try to build with `mvn verify`

**Expected behavior**
Developer builds succeed despite checkedout files which are common in developer environments.

pxLi:
sounds reasonable, let me add a profile for it

sameerz:
Closed by #1820

PR #1820, opened 2021-02-26
[BUG] make modified check pre-merge only

pxLi:

fix https://github.com/NVIDIA/spark-rapids/issues/1817 by adding a pre-merge profile

only when `pre-merge` profile is explicitly enabled, the modified_files check would be executed. verified locally


pxLi:
build

GaryShen2008 (approved):

PR #1839, opened 2021-03-01
Refactor join code to reduce duplicated code

### LINKED ISSUE #412: [FEA] Shim followup - Investigate commonizing join code in Shims
Since spark 3.1.0 changed the base HashJoin class, the shims had to copy a lot of code. Investigate if we can have a base class that shares as much code as possible.

jlowe:
Fixes #412.

This reduces the duplication of code between the shims by factoring out the core join code into a common base class.  I ran the join integration tests against the spark300, spark311, and spark301db shims.


jlowe:
build

tgravescs:
Note we can reference https://github.com/NVIDIA/spark-rapids/issues/412 that I never got to

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuShuffledHashJoinExec.scala
```diff
@@ -85,79 +80,10 @@ case class GpuShuffledHashJoinExec(
...
-        context.addTaskCompletionListener[Unit](_ => builtTable.close())
-
-        doJoin(builtTable, streamIter, boundCondition,
-          numOutputRows, joinOutputRows, numOutputBatches,
-          streamTime, joinTime, filterTime, totalTime)
-      }
-    }
+    isSkewJoin = isSkewJoin) {
+  override def nodeName: String = {
+    if (isSkewJoin) super.nodeName + "(skew=true)" else super.nodeName
```
gerashegalov, line 87: we can move this to `GpuShuffledHashJoinBase`

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/GpuShuffledHashJoinExec.scala
```diff
@@ -86,79 +81,10 @@ case class GpuShuffledHashJoinExec(
...
-        context.addTaskCompletionListener[Unit](_ => builtTable.close())
-
-        doJoin(builtTable, streamIter, boundCondition,
-          numOutputRows, joinOutputRows, numOutputBatches,
-          streamTime, joinTime, filterTime, totalTime)
-      }
-    }
+    isSkewJoin = isSkewJoin) {
+  override def nodeName: String = {
+    if (isSkewJoin) super.nodeName + "(skew=true)" else super.nodeName
```
gerashegalov, line 88: we can move this to `GpuShuffledHashJoinBase`

gerashegalov (changes requested):

revans2 (dismissed):
Looks good to me. I agree with @gerashegalov on his cleanup comments too.

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuShuffledHashJoinExec.scala
jlowe: I don't think we can move this because the Databricks version does not have this override.

revans2: If they don;'t set isSkewJoin, then it will always be false and the output will be identical to what it was before. It is a NOOP.

jlowe:
build

gerashegalov (approved):
LGTM

revans2 (approved):

PR #1843, opened 2021-03-02
Refactor GpuGenerateExec and Explode

sperlingxx:
This pull request is to refactor `GpuGenerateExec` in order to support GPU generators with a more general approach, as `GpuGenerateExec` we had before is specialized for `Explode(CreateArray(_))`.

In this PR,  we rework  `GpuGenerateExec` with a new interface `GpuGenerator`. The core abstract method of `GpuGenerator` is `generate`, which performing the entire generating process.  The new `GpuGenerateExec` works like a delegation of  the GpuGenerators. For new `GpuGenerateExec`, it orchestrates input columns and calls `GpuGenerator.generate`  to apply generation.

In addition, in this PR, `LiteralExprMeta` is modified to provide a tricky support of `GpuLiteral` on ArrayData: transforming `Literal(ArrayType(BaseType))` into `CreateArray(Literal(BaseType):_*)`.


sperlingxx:
build

sperlingxx:
build

sperlingxx:
build

integration_tests/src/main/python/generate_expr_test.py
```diff
@@ -63,3 +68,33 @@ def test_explode_litarray(data_gen):
             lambda spark : four_op_df(spark, data_gen).select(f.col('a'), f.col('b'), f.col('c'), 
                 f.explode(array_lit)))
 
+#sort locally because of https://github.com/NVIDIA/spark-rapids/issues/84
```
revans2, line 71: This only applies to sorting floating point values.  There are no floating point values in this test so we should remove the comment and sort on the GPU.

integration_tests/src/main/python/generate_expr_test.py
```diff
@@ -63,3 +68,33 @@ def test_explode_litarray(data_gen):
...
+def test_explode_split_string():
+    str_gen = StringGen('([ABC]{0,3}_?){0,7}').with_special_case('').with_special_pattern('.{0,10}')
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : two_col_df(spark, int_gen, str_gen).selectExpr('a', 'explode(split(b, "_"))'))
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : two_col_df(spark, int_gen, str_gen).selectExpr(
+                'a', 'explode(split(b, "_")) as c').selectExpr(
+                'a', 'explode(split(c, "A"))'))
+
+#sort locally because of https://github.com/NVIDIA/spark-rapids/issues/84
```
revans2, line 83: Again no float/double values in the tests so this does not apply.

integration_tests/src/main/python/generate_expr_test.py
```diff
@@ -63,3 +68,33 @@ def test_explode_litarray(data_gen):
...
+# After 3.1.0 is the min spark version we can drop this
+@ignore_order(local=True)
+def test_explode_from_parquet(spark_tmp_path):
```
revans2, line 86: What indication do we have that reading from parquet will cause different results compared to just doing the explode directly.  I personally would much rather see us cover all of the types that explode supports instead of covering a few things that might come before it in a plan.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
```diff
@@ -233,3 +234,32 @@ case class GpuLiteral (value: Any, dataType: DataType) extends GpuLeafExpression
...
+
+  override def convertToGpu(): GpuExpression = {
+    lit.dataType match {
+      // NOTICE: There is a temporary transformation from Literal(ArrayType(BaseType)) into
+      // CreateArray(Literal(BaseType):_*). The transformation is a walkaround support for Literal
+      // of ArrayData under GPU runtime, because cuDF scalar doesn't support nested types.
+      case ArrayType(baseType, _) =>
+        val litArray = lit.value.asInstanceOf[ArrayData]
+          .array.map(GpuLiteral(_, baseType))
+        GpuCreateArray(litArray, useStringTypeWhenEmpty = false)
```
revans2, line 252: I am a bit nervous about this change.  Because cudf does not support literal arrays I would want to see a lot more testing of literal arrays in other places in queries, not just explode.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2463,10 +2462,8 @@ object GpuOverrides {
     exec[GenerateExec] (
       "The backend for operations that generate more output rows than input rows like explode",
       ExecChecks(
-        TypeSig.commonCudfTypes + TypeSig.DECIMAL
-            .withPsNote(TypeEnum.ARRAY,
-              "Only literal arrays and the output of the array function are supported")
-            .nested(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL),
+        TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.ARRAY.nested(
+          TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.ARRAY),
```
revans2, line 2466: So this supports arrays of arrays?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2299,6 +2291,13 @@ object GpuOverrides {
         override def convertToGpu(child: Expression): GpuExpression =
           GpuMakeDecimal(child, a.precision, a.scale, a.nullOnOverflow)
       }),
+    expr[Explode](
+      "Given an input array produces a sequence of rows for each value in the array",
+      ExprChecks.unaryProject(TypeSig.all, TypeSig.all,
+        TypeSig.ARRAY.nested(TypeSig.all), TypeSig.ARRAY.nested(TypeSig.all)),
```
revans2, line 2297: Spark supports maps, but we do not.  That should be captured here.
Spark supports user defined types as part of the nesting, we do not.  That must be removed from our list of supported input types.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -701,19 +701,11 @@ object GpuOverrides {
     expr[Literal](
       "Holds a static value from the query",
       ExprChecks.projectNotLambda(
-        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.CALENDAR,
+        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.CALENDAR
+          + TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.NULL
+          + TypeSig.DECIMAL + TypeSig.CALENDAR),
```
revans2, line 706: I don't think we can support an array of calendars as a literal either.  That needs to be removed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2299,6 +2291,13 @@ object GpuOverrides {
         override def convertToGpu(child: Expression): GpuExpression =
           GpuMakeDecimal(child, a.precision, a.scale, a.nullOnOverflow)
       }),
+    expr[Explode](
+      "Given an input array produces a sequence of rows for each value in the array",
```
revans2, line 2295: Here would be a good place to add a note about outer not being supported. It might even be nice to have something for the ExprMeta that we produce that could some how tell GenerateExec if it supports outer/inner/etc, but I am not sure if it is worth the engineering effort to do it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -122,75 +201,25 @@ case class GpuGenerateExec(
...
-            }
+    val generatorProjectList: Seq[GpuExpression] =
+      GpuBindReferences.bindGpuReferences(generator.children, child.output)
+    val othersProjectList: Seq[GpuExpression] =
+      GpuBindReferences.bindGpuReferences(requiredChildOutput, child.output)
+
+    child.executeColumnar().map { inputFromChild =>
+      withResource(inputFromChild) { input =>
+        withResource(new NvtxWithMetrics("GpuGenerateExec", NvtxColor.PURPLE, totalTime)) { _ =>
+          // Project input columns, setting other columns ahead of generator's input columns.
```
revans2, line 212: We lost the optimizations we had in place to reduce total memory used when exploding a literal array.  We need to have a way to put that back in.  Similarly this has no memory optimizations.  It is very easy to have the amount of memory used in an output batch grow to a massive size because of the number of elements in the array/map passed in. At a minimum when we don't know if it is a literal/scalar value passed in for the explode, then we need to have the generator do some estimation to split up the input batch into smaller manageable chunks so we have some confidence that we will be close to the desired output batch size.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -36,75 +35,155 @@ class GpuGenerateExecSparkPlanMeta(
...
+
+case class GpuExplode(child: Expression) extends GpuExplodeBase {
+  override val position: Boolean = false
+
+  override def generate(inputBatch: ColumnarBatch,
+    generatorOffset: Int,
+    outer: Boolean): ColumnarBatch = {
+
+    require(inputBatch.numCols() - 1 == generatorOffset,
+      "GpuExplode should has and only has one input attribute")
```
revans2, line 163: Internal Error GpuExplode supports one and only one input attribute.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -36,75 +35,155 @@ class GpuGenerateExecSparkPlanMeta(
...
+   *              given `generator` is empty.
+   *
+   * @return result ColumnarBatch
+   */
+  def generate(inputBatch: ColumnarBatch,
+    generatorOffset: Int,
+    outer: Boolean): ColumnarBatch
+}
+
+trait GpuCollectionGenerator extends GpuGenerator {
```
revans2, line 105: I'm not sure that we need to match the Spark class hierarchy here.  Spark is doing outer/etc in a very different way and the CollectionGenerator in Spark is really there for code generation, as per the comments on that class. I think we can clean things up for now and add them back in, if we reach a situation where we need it.

integration_tests/src/main/python/generate_expr_test.py
```diff
@@ -63,3 +68,33 @@ def test_explode_litarray(data_gen):
             lambda spark : four_op_df(spark, data_gen).select(f.col('a'), f.col('b'), f.col('c'), 
                 f.explode(array_lit)))
 
+#sort locally because of https://github.com/NVIDIA/spark-rapids/issues/84
+# After 3.1.0 is the min spark version we can drop this
+@ignore_order(local=True)
+def test_explode_split_string():
+    str_gen = StringGen('([ABC]{0,3}_?){0,7}').with_special_case('').with_special_pattern('.{0,10}')
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : two_col_df(spark, int_gen, str_gen).selectExpr('a', 'explode(split(b, "_"))'))
```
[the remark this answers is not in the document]
sperlingxx: At here, split is to generate array data (from string). The intention of this test is to prove `GpuExplode` is able to run with child expressions other than `GpuMakeArray` and `GpuLiteral`.

integration_tests/src/main/python/generate_expr_test.py
```diff
@@ -63,3 +68,33 @@ def test_explode_litarray(data_gen):
...
+    assert_gpu_and_cpu_are_equal_collect(
```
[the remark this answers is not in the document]
sperlingxx: Done.

integration_tests/src/main/python/generate_expr_test.py
sperlingxx: Done.

integration_tests/src/main/python/generate_expr_test.py
sperlingxx: Done.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
sperlingxx: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
sperlingxx: I put back memory optimization via splitting up the input batch into smaller manageable chunks.
If it is a  literal/scalar value passed in for the explode, which means number of row expansion are fixed, we can figure out number of splits from meta data.
If not, we estimate the number of row expansion with calculating average array size of sampling data.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: I created `GeneratorExprMeta` to wrap `outerSupport` attribute for Generators.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
sperlingxx: I removed this.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: Removed.

integration_tests/src/main/python/generate_expr_test.py
```diff
@@ -27,6 +28,9 @@ def four_op_df(spark, gen, length=2048, seed=0):
         ('c', gen),
         ('d', gen)], nullable=False), length=length, seed=seed)
 
+"""
+below tests are temporarily disabled because explode_position has not been supported in cuDF-nightly
```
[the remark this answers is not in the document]
sperlingxx: I added `GpuPosExplode`. So, these tests are enabled now.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: No, we don't support nested array in current. Here is a mistake, I've fixed it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
sperlingxx: I've added some tests for Literal(ArrayData) in `ProjectExprSuite`.

integration_tests/src/main/python/generate_expr_test.py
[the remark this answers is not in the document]
revans2: But Parquet is not an expression, it is a separate plan stage. I am fine with the string split, if we document that is what we are trying to test. The parquet test I am still not convinced that we need.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
revans2: OK Can we have a follow on issue to remove this once cudf supports scalar arrays? Which should be fairly soon.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -36,75 +36,219 @@ class GpuGenerateExecSparkPlanMeta(
...
+      }
+  }
+
+  // In terms of GpuExplodeBase, split number should be array size of each row. It is because
+  // we would like to keep size of output batches close to input ones in case of OOM caused by
+  // expanded output.
+  override def inputSplitNum(genIn: ColumnarBatch, outer: Boolean): Int = {
+    // only split up input batch if it is large enough
+    if (genIn.numRows() < SPLIT_BATCH_THRESHOLD) return 1
+
```
revans2, line 158: I would prefer to see us do something more like the following in determining the splits.
This is just pseudo code.
```
val baseVectors = GpuColumnVector.extractBases(genIn)

// Get the vectors that are going to explode in size because of the array column
val columnsThatWillBeRepeated = baseVectors.slice(0, generatorOffset)
val averageRowSizeToBeRepeated = columnsThatWillBeRepeated.map(_.getDeviceMemorySize()).sum().toDouble / genIn.numRows

// Get the  average row size of the column that we are going to explode
// along with an estimate of how much the output will explode by
val (averageExplodeColumnSize, averageExplodeColumnLength) = withResource(baseVectors(generatorOffset).getChildColumnView(0)) { listDataColumn =>
    val totalCount = listDataColumn.getRowCount
    val totalSize = listDataColumn.getDeviceMemorySize
    (totalSize.toDouble/totalCount, totalCount.toDouble/genIn.numRows)
}
// The estimated size in bytes for each output row
val estimatedOutputRowSize = averageRowSizeToBeRepeated + averageExplodeColumnSize
// The maximum number of rows per split to hit the targetSize in bytes
val maxTargetOutputRowCount =  Math.min(Int.MaxValue, targetSizeBytes/estimatedOutputRowSize).toInt
// How many output rows do we think will be produced by the explode in total
val estimatedOutputRows = genIn.numRows * averageExplodeColumnLength
// how may splits will we need to keep the output size under the target size
val numSplits = ceil(estimatedOutputRows/maxTargetOutputRowCount)
```

revans2:
I also really want to see some performance and memory usage numbers for the case with a literal array column.  We have a use case where we are pushing the limit in that respect, and if the numbers are too high we are going to need to figure out a better way to lazily evaluate the columns instead of materializing all of the data up front.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
sperlingxx: According to above pseudo code, I reworked the method to determine the splits. I made some simplification on the basis of pseudo code, removing the computation of average values.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -36,75 +36,219 @@ class GpuGenerateExecSparkPlanMeta(
...
+   * generators may produce multiple records for each input record, which make output batch size
+   * much larger than input size.
+   *
+   * @param generatorInput input batch, only containing columns for generation
+   * @param outer when true, each input row will be output at least once, even if the output of the
+   *              given `generator` is empty.
+   *
+   * @return recommended number of splits for given input batch
+   */
+  def inputSplitNum(generatorInput: ColumnarBatch, outer: Boolean): Int = 1
```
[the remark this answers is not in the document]
sperlingxx: I added the helper method  `def generateSplitIndices(rows: Long, numSplits: Int): Array[Int]` into `GpuBatchUtils.scala`, which helps finding even split points given number of splits.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -36,75 +36,219 @@ class GpuGenerateExecSparkPlanMeta(
...
+          .add("value", vt, valueContainsNull)
+      }
+  }
+
+  // In terms of GpuExplodeBase, split number should be array size of each row. It is because
+  // we would like to keep size of output batches close to input ones in case of OOM caused by
+  // expanded output.
+  override def inputSplitNum(genIn: ColumnarBatch, outer: Boolean): Int = {
+    // only split up input batch if it is large enough
+    if (genIn.numRows() < SPLIT_BATCH_THRESHOLD) return 1
```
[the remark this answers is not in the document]
sperlingxx: Removed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
sperlingxx: I filed an issue for this (#1902).

sperlingxx:
build

sperlingxx:
> I also really want to see some performance and memory usage numbers for the case with a literal array column. We have a use case where we are pushing the limit in that respect, and if the numbers are too high we are going to need to figure out a better way to lazily evaluate the columns instead of materializing all of the data up front.

Hi @revans2, I conducted a  benchmark to compare the performance of two kinds of implementation on literal array column. And sadly found the implementation of current PR has much poorer performance compared with the implementation of main branch.
I tried to tune the performance of current PR with configuration `spark.rapids.sql.batchSizeBytes`. The tuning will bring up some performance improvement, but still much poorer.

I performed the benchmark with below script:
```python3
@pytest.mark.parametrize('data_gen', [LongGen()], ids=idfn)
def test_explode_litarray(data_gen):
    array_lit = gen_scalar(ArrayGen(data_gen, min_length=1000, max_length=1000, nullable=False))
    spark_fn = lambda spark : four_op_df(spark, data_gen, length=1000000)\
        .select(f.col('a'), f.col('b'), f.col('c'), f.explode(array_lit))\
        .count()
    for i in range(5):
        with_gpu_session(spark_fn)
```

Here is the result in detail: https://docs.google.com/spreadsheets/d/1S84wy9Yc68yFkJmGLd6uKfEsVS-sehiZb7q_CJNn3c4/edit?usp=sharing

integration_tests/src/main/python/join_test.py
```diff
@@ -54,7 +54,7 @@ def create_df(spark, data_gen, left_length, right_length):
     right = binary_op_df(spark, data_gen, length=right_length).withColumnRenamed("a", "r_a")\
             .withColumnRenamed("b", "r_b")
     return left, right
-
+"""
```
revans2, line 57: Why are these tests commented out?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -36,75 +38,345 @@ class GpuGenerateExecSparkPlanMeta(
...
+
+  /**
+   * Extract lazy expressions from generator if exists.
+   *
+   * This is a specialized method for GPU runtime, which is called by GpuGenerateExec to determine
+   * whether current generation plan can be executed with optimized lazy array generation or not.
+   *
+   * @return fixed length lazy expressions for generation
+   */
+  def fixedLenLazyExpressions: Seq[Expression] = Nil
```
revans2, line 136: nit: we should document that Nil means there are no expressions to extract.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -36,75 +38,345 @@ class GpuGenerateExecSparkPlanMeta(
...
+      numOutputBatches: GpuMetric,
+      totalTime: GpuMetric): Iterator[ColumnarBatch] = {
+
+    val numArrayColumns = boundLazyProjectList.length
+    val numOtherColumns = boundOthersProjectList.length
+    val numExplodeColumns = if (position) 2 else 1
+
+    new Iterator[ColumnarBatch] {
+
+      var currentBatch: ColumnarBatch = _
```
revans2, line 272: nit: can we have a follow on issue filed to make this spillable? I don't want it to happen here, but we should at least be tracking it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -36,75 +38,345 @@ class GpuGenerateExecSparkPlanMeta(
...
+        }
+        currentBatch != null
+      }
+
+      override def next(): ColumnarBatch = {
+        if (currentBatch == null || indexIntoData >= numArrayColumns) {
+          fetchNextBatch()
+        }
+        withResource(new NvtxWithMetrics("GpuGenerateExec", NvtxColor.PURPLE, totalTime)) { _ =>
+          val result = new Array[ColumnVector](numExplodeColumns + numOtherColumns)
```
revans2, line 302: nit: you can do a withResource on an Array too.

revans2 (commented):
Looks really good, just a few nits.

revans2:
Do you have updated benchmark results?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
sperlingxx: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
sperlingxx: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
sperlingxx: Done.

integration_tests/src/main/python/join_test.py
sperlingxx: My bad. It been fixed.

sperlingxx:
build

sperlingxx:
> Do you have updated benchmark results?

I just appended results for latest change.

revans2 (approved):
Great work

PR #1852, opened 2021-03-03
Fix Part Suite Tests

revans2:
Something changed in cudf where all now returns false for an empty column.  We don't use all except for tests so it is not a big deal.  This works around that change, but while looking into it I noticed that there was a separate bug where we were comparing expected to itself instead of actual.  This fixes that too.

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/GpuPartitioningSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/GpuPartitioningSuite.scala
@@ -19,7 +19,7 @@ package com.nvidia.spark.rapids
 import java.io.File
 import java.math.RoundingMode
 
-import ai.rapids.cudf.{ColumnVector, DType, Table}
+import ai.rapids.cudf.{ColumnVector, Cuda, DType, Table}
 import org.scalatest.FunSuite
 
 import org.apache.spark.SparkConf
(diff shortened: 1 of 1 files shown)


revans2:
build

tests/src/test/scala/com/nvidia/spark/rapids/GpuPartitioningSuite.scala
```diff
@@ -53,6 +53,24 @@ class GpuPartitioningSuite extends FunSuite with Arm {
...
+        Cuda.DEFAULT_STREAM)) { decompressor =>
+        compressedColumn.getTableBuffer.incRefCount()
+        decompressor.addBufferToDecompress(compressedColumn.getTableBuffer,
+          compressedColumn.getTableMeta.bufferMeta)
+        withResource(decompressor.finishAsync()) { outputBuffers =>
+          val outputBuffer = outputBuffers.head
+          // There should be only one
+          withResource(
+            MetaUtils.getTableFromMeta(outputBuffer, compressedColumn.getTableMeta)) { table =>
+            (0 until table.getNumberOfColumns).map(i => table.getColumn(i).incRefCount()).toArray
```
jlowe, line 70: I think this is going to leak.  `extractBases` made the simplifying assumption that any columns extracted are just aliased onto some other column batch memory and therefore don't need to be closed.  However for compressed batches that is an invalid assumption.

The contract for `extractBases` should change so that callers need to always close the resulting array to cover the compression case.

jlowe (changes requested):

revans2:
build

tests/src/test/scala/com/nvidia/spark/rapids/GpuPartitioningSuite.scala
revans2: Thanks for catching that. I renamed extractBases so it would not be confused with extractBases in GpuColumnVector where the contract is different.

jlowe (approved):

PR #1858, opened 2021-03-03
fix shuffle manager doc on ucx library path

rongou:

The UCX 1.9.0 debian package puts `libjucx.so` in `/usr/lib`, which conflicts with the one in our jar file.

### DIFF
--- a/docs/additional-functionality/rapids-shuffle.md
+++ b/docs/additional-functionality/rapids-shuffle.md
@@ -272,7 +272,7 @@ In this section, we are using a docker container built using the sample dockerfi
 --conf spark.executorEnv.UCX_MAX_RNDV_RAILS=1 \
 --conf spark.executorEnv.UCX_MEMTYPE_CACHE=n \
 --conf spark.executorEnv.UCX_IB_RX_QUEUE_LEN=1024 \
---conf spark.executorEnv.LD_LIBRARY_PATH=/usr/lib:/usr/lib/ucx \
+--conf spark.executorEnv.LD_LIBRARY_PATH=/usr/lib/ucx \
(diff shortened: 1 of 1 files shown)


abellina:
I can't reproduce this myself, with a system that has `libjucx.so` in `/usr/lib` with `LD_LIBRARY_PATH` set to `/usr/lib:/usr/lib/ucx` (which the part you removed should be there by default). I see us opening the libjucx.so that is bundled via the jucx jar. I am not sure why it is picking up the jucx lib from `/usr/lib` in this case, instead of the one we bundle in our dist jar. 

If I force it to load, I do get a segfault (using `LD_PRELOAD=/usr/lib/libjucx.so`). It seems the libjucx.so for UCX 1.9 in the packaging is different than the one we bundle in our jar. Can you verify using `LD_DEBUG=libs` that the library in `/usr/lib` is the one that is getting loaded for you?

`libjucx.so` shouldn't be included in UCX packaging. I chatted with @petro-rudenko, and he opened a PR to remove libjucx.so from the packaging (https://github.com/openucx/ucx/pull/6440). UCX 1.10 PR is pending. So starting ucx 1.10, libjucx.so is not going to be in /usr/lib.

In the doc we do say: set `LD_LIBRARY_PATH` if the UCX library is in a non-default location. In terms of this PR I think we should remove the `LD_LIBRARY_PATH` config provided, as that is just confusing. We could leave the bit after it where we make the user aware this needs native libraries and `LD_LIBRARY_PATH` may need to be set in some circumstances.

rongou:
Removed the line.

abellina (approved):

abellina:
build

abellina:
Note that @petro-rudenko's PR went into 1.10: https://github.com/openucx/ucx/pull/6441.

Issue #1864, opened 2021-03-04
[BUG]HostColumnarToGPU inefficient when only doing count()


tgravescs:
**Describe the bug**
If you are doing a count() operation, there are not any columns or data passed with the ColumnarBatch. The HostColumnarToGPU defaults to a row limit of 512 in this case, which is very inefficient.  We should check for this case and make the limit much higher.

In one test scenario reading from a columnar source with I believe around 15 million rows, a simple count() took 2.1 minutes when it should have taken around 3 seconds.

PR #1868, opened 2021-03-04
Increase row limit when doing count() for HostColumnarToGpu 

tgravescs:
If doing a count() with HostcolumnToGPU which ends up doing a coalesce, the row limit is currently set at 512. This can be very inefficient for larger data.  So when there aren't any columns specified just set the row limit to max integer.

This improved one query from 2 minutes down to 6 seconds.

I also added a new testing function assert_gpu_and_cpu_are_equal_count to be able to easily test count() calls.

fixes https://github.com/NVIDIA/spark-rapids/issues/1864

### DIFF
--- a/integration_tests/src/main/python/asserts.py
+++ b/integration_tests/src/main/python/asserts.py
@@ -271,9 +271,9 @@ def assert_gpu_fallback_write(write_func,
         gpu_end - gpu_start, cpu_end - cpu_start))
 
     (cpu_bring_back, cpu_collect_type) = _prep_func_for_compare(
-            lambda spark: read_func(spark, cpu_path), True)
+            lambda spark: read_func(spark, cpu_path), 'COLLECT')
     (gpu_bring_back, gpu_collect_type) = _prep_func_for_compare(
-            lambda spark: read_func(spark, gpu_path), True)
+            lambda spark: read_func(spark, gpu_path), 'COLLECT')
 
     from_cpu = with_cpu_session(cpu_bring_back, conf=conf)
     from_gpu = with_cpu_session(gpu_bring_back, conf=conf)
@@ -286,7 +286,7 @@ def assert_gpu_fallback_write(write_func,
 def assert_gpu_fallback_collect(func,
         cpu_fallback_class_name,
         conf={}):
-    (bring_back, collect_type) = _prep_func_for_compare(func, True)
+    (bring_back, collect_type) = _prep_func_for_compare(func, 'COLLECT')
     conf = _prep_incompat_conf(conf)
 
     print('### CPU RUN ###')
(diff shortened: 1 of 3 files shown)


tgravescs:
build

integration_tests/src/main/python/asserts.py
```diff
@@ -335,15 +338,24 @@ def assert_gpu_and_cpu_are_equal_collect(func, conf={}):
     In this case the data is collected back to the driver and compared here, so be
     careful about the amount of data returned.
     """
-    _assert_gpu_and_cpu_are_equal(func, True, conf=conf)
+    _assert_gpu_and_cpu_are_equal(func, 'COLLECT', conf=conf)
+
+def assert_gpu_and_cpu_are_equal_count(func, conf={}):
```
revans2, line 343: I am not a huge fan of this API. It makes it really easy to accidentally only compare the number of rows for a query instead of the actual data of the query. I understand why you are trying to do this, but the same thing can be achieved with `assert_gpu_and_cpu_are_equal_collect` and inserting a `groupBy().count()` in there.

This is not a show stopper. If you want to keep it, that is fine, I would just like a warning in the doc string about this.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
```diff
@@ -279,6 +279,11 @@ class HostToGpuCoalesceIterator(iter: Iterator[ColumnarBatch],
     // schema and desired batch size
     batchRowLimit = GpuBatchUtils.estimateRowCount(goal.targetSizeBytes,
       GpuBatchUtils.estimateGpuMemory(schema, 512), 512)
+    // when there aren't any columns, it generally means user is doing a count() and we don't
```
gerashegalov, line 282: nice find.

Let us not call `estimateRowCount/estimateGpuMemory` when there no columns:
```scala
batchRowLimit = if (batch.numCols() > 0 ) {
  GpuBatchUtils.estimateRowCount(goal.targetSizeBytes,
      GpuBatchUtils.estimateGpuMemory(schema, 512), 512)
} else {
  Integer.MAX_VALUE
}      
```

integration_tests/src/main/python/asserts.py
tgravescs: ok, I thought the _count would imply that you are only comparing counts.  We in general have no tests that test count() that I see and we have hit lots of issues with it so I thought it would be a convenient api to easily add tests without having to change the operation or a bunch of code.

I'm definitely open to changing if we think it will cause problems but I don't think groupBy().count() is as intuitive or convenient. 

Perhaps just changing the name to assert_gpu_and_cpu_row_counts_equal and add more documentation?  Or if you can think of another convenience api.

revans2: That is fine with me. I mostly want to be sure that no one uses it on accident.  Using it on purpose is fine.

tgravescs:
build

gerashegalov (approved):
LGTM, :rocket:

PR #1874, opened 2021-03-04
Update 0.4.0 with contiguous split workaround

jlowe:
Merging in the contiguous split workaround for the 0.4.0 release.

**NOTE: This needs to be merged with a merge commit, not squashed!**

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
@@ -608,13 +608,15 @@ case class GpuCoalesceBatches(child: SparkPlan, goal: CoalesceGoal)
     // cache in local vars to avoid serializing the plan
     val outputSchema = schema
     val decompressMemoryTarget = maxDecompressBatchMemory
-    val cannotSpill = child.schema.fields.exists { f =>
-      f.dataType match {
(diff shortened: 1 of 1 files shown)


jlowe:
build

jlowe:
Fixing this in #1875

abellina (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
```diff
@@ -608,13 +608,15 @@ case class GpuCoalesceBatches(child: SparkPlan, goal: CoalesceGoal)
...
-    val cannotSpill = child.schema.fields.exists { f =>
-      f.dataType match {
-        case MapType(_, _, _) | ArrayType(_, _) | StructType(_) => true
-        case _ => false
-      }
-    }
-
+    // disabling spillable batches due to
+    // https://github.com/rapidsai/cudf/issues/7514
+    val cannotSpill = true
```
gerashegalov, line 613: we can also consider leaving the code in but adding a new conf like "disable.spillable.batches" . We would flip the default to false once the bug is fixed.

jlowe: IMO configs are overkill for this.  Bugfix is already posted in cudf-0.19, will definitely be fixed in 0.5.

gerashegalov (approved):

PR #1876, opened 2021-03-05
[WIP] avoid reading spilled buffers multiple times during shuffle

rongou:

Digging through spark logs, it seems that shuffle buffers are being read multiple times. This is one place it might be happening, but this particular fix doesn't seem to help much. `BufferSendState#next()` seems be called multiple times with the same list of buffers.

This is orthogonal to unspilling, might be a useful thing to do.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
@@ -16,6 +16,8 @@
 
 package com.nvidia.spark.rapids.shuffle
 
+import scala.collection.mutable
+
 import ai.rapids.cudf.{Cuda, CudaUtil, DeviceMemoryBuffer, HostMemoryBuffer, MemoryBuffer}
 import com.nvidia.spark.rapids.{Arm, RapidsBuffer, RapidsBufferId, ShuffleMetadata, StorageTier}
 import com.nvidia.spark.rapids.RapidsPluginImplicits._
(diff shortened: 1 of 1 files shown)


abellina:
>This is one place it might be happening, but this particular fix doesn't seem to help much.

abellina:
@rongou I am wondering if in this case we could remove the extra copy to bounce buffer and the materialization of a potentially large GDS buffer to GPU memory.

This case is the case where a GDS buffer is bigger than a bounce buffer. Could we copy directly from GDS in this case into a device-backed bounce buffer?

rongou:
I moved the memory buffers up into member variables. This does have a dramatic effect on GDS spilling performance. Query 95 is down from 400 seconds to 60 seconds. But there are still some duplicate reads. Probably should still do unspilling.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
```diff
@@ -119,6 +128,7 @@ class BufferSendState(
       throw new IllegalStateException("ALREADY CLOSED!")
     }
     isClosed = true
+    cleanMemoryBuffers()
```
abellina, line 131: we should move the `cleanMemoryBuffers` call within `releaseAcquiredToCatalog`. Otherwise, we will hold onto the materialized buffer (and in GDS'es case, the copy to device memory) until `close`, instead of until the UCX callback says its ok to let go.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
```diff
@@ -171,31 +181,36 @@ class BufferSendState(
 
         acquiredBuffs.foreach { case RangeBuffer(blockRange, rapidsBuffer) =>
```
abellina, line 182: Currently the caller (when UCX callback is triggered) calls `releaseAcquiredToCatalog`, and we require this state is cleaned in https://github.com/NVIDIA/spark-rapids/pull/1876/files#diff-039d2429369bd2d3796876c1f91257f62836765e46509a6e7abc9534df22abecR146. It means the behavior of this is to allow the catalog to spill the buffer while we are working a window and waiting for possibly long amount of times. You are changing that behavior here not just for GDS, but for all tiers.

If you have a bunch of transfers in flight and each is holding onto a large shuffle block that a bounce buffer window is sliding through, I think we are going to run into trouble if we hold each of those large shuffle blocks for the entire duration of that transfer.

Were you able to check if copying from GDS to device memory at an offset is possible? And is it performant for bounce-buffer length copies? To me that seems ideal, the buffer isn't materialized wholesale, and it arrives at memory that is destined to be used for the transfer.

abellina (changes requested):

rongou:
@abellina the unspill approach seems to work well and obviate the need for this PR. Do you think we still want to do this?

abellina:
@rongou I'll review your other PR if that's the preferred approach. For me, I was just interested in the performance of straight copies from GDS for bounce-buffer size requests. In the case in this PR, you were holding onto a reference of GPU/Host (and GDS) memory for longer periods of times than we were before, so that's the part I wasn't sure about.

rongou:
Closing this. Caching should be handled by unspilling. Once that's done, will work on streaming GDS buffers to UCX.

PR #1877, opened 2021-03-05
remove docker build in pre-merge [skip ci]

pxLi:

Move image build work to an independent image manage service, save some time for each build.

Tested in my forked repo. I will submit a verify PR against our repo after this one get merged.

### DIFF
--- a/jenkins/Jenkinsfile-blossom.premerge
+++ b/jenkins/Jenkinsfile-blossom.premerge
@@ -28,8 +28,6 @@ import ipp.blossom.*
 def pluginPremerge
 
 def githubHelper // blossom github helper
-def IMAGE_PREMERGE // temp image for premerge test
-def PREMERGE_TAG
 def skipped = false
 
 pipeline {
@@ -89,7 +87,7 @@ pipeline {
 
             agent {
                 kubernetes {
-                    label "premerge-docker-${BUILD_TAG}"
+                    label "premerge-prepare-${BUILD_TAG}"
                     cloud 'sc-ipp-blossom-prod'
                     yaml pod.getDockerBuildYAML()
                     workspaceVolume persistentVolumeClaimWorkspaceVolume(claimName: "${PVC}", readOnly: false)
@@ -104,13 +102,13 @@ pipeline {
                             changelog: false,
                             poll: true,
                             scm: [
-                                    $class: 'GitSCM', branches: [[name: githubHelper.getMergedSHA()]],
+                                    $class                           : 'GitSCM', branches: [[name: githubHelper.getMergedSHA()]],
                                     doGenerateSubmoduleConfigurations: false,
-                                    submoduleCfg: [],
-                                    userRemoteConfigs: [[
-                                        credentialsId: 'github-token',
-                                        url: githubHelper.getCloneUrl(),
-                                        refspec: '+refs/pull/*/merge:refs/remotes/origin/pr/*']]
+                                    submoduleCfg                     : [],
+                                    userRemoteConfigs                : [[
+                                                                                credentialsId: 'github-token',
+                                                                                url          : githubHelper.getCloneUrl(),
+                                                                                refspec      : '+refs/pull/*/merge:refs/remotes/origin/pr/*']]
                             ]
                     )
 
@@ -118,18 +116,8 @@ pipeline {
                         def CUDA_NAME = sh(returnStdout: true,
                                 script: '. jenkins/version-def.sh>&2 && echo -n $CUDA_CLASSIFIER | sed "s/-/./g"')
 
-                        IMAGE_TAG = "dev-ubuntu16-${CUDA_NAME}"
-                        CACHE_IMAGE_NAME = "${ARTIFACTORY_NAME}/sw-spark-docker/plugin:${IMAGE_TAG}"
-
-                        sh "docker pull $CACHE_IMAGE_NAME || true"
-
-                        PREMERGE_TAG = "${IMAGE_TAG}-${BUILD_TAG}"
-                        IMAGE_PREMERGE = "${ARTIFACTORY_NAME}/sw-spark-docker-local/plugin:${PREMERGE_TAG}"
-                        def CUDA_VER = "$CUDA_NAME" - "cuda"
-                        docker.build(IMAGE_PREMERGE, "-f jenkins/Dockerfile-blossom.ubuntu16 --build-arg CUDA_VER=$CUDA_VER --cache-from $CACHE_IMAGE_NAME -t $IMAGE_PREMERGE .")
-                        uploadDocker(IMAGE_PREMERGE)
-
-                        pluginPremerge = pod.getGPUYAML("${IMAGE_PREMERGE}", "${GPU_TYPE}", '8', '32Gi') // cpu: 8, memory: 32Gi
+                        def IMAGE_PREMERGE = "$ARTIFACTORY_NAME/sw-spark-docker-local/plugin:dev-ubuntu16-$CUDA_NAME-blossom-dev"
+                        pluginPremerge = pod.getGPUYAML("${IMAGE_PREMERGE}", "${GPU_TYPE}", '8', '32Gi')
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


pxLi:
build

NvTimLiu (approved):
Looks good to me

pxLi:
@tgravescs @revans2 @jlowe 

this could save around 10 - 15 mins for each run, but it will not do docker build even we have a change to the ubuntu Dockerfile (nightly test will help detect if any issue)

what we can do  to improve for this PR is that if the PR include Dockerfile change, then we enable the docker build. 
This could make the pre-merge pipeline complicated,
1. if ubuntu CI docker file change, enable docker build w/ a temp tag and push to Artifactory (to not conflict w/ other builds)
2. run test w/ the temp image
3. post action to clean the temp image in Artifactory

But changes to dockerfile is not that frequent, so to add extra steps to check and cleanup maybe not worthy for the 10 - 15 mins time saving, we may just keep what we are doing right now (docker build for every run).

Please kindly share your thoughts about this, thanks!

jlowe:
IMO 10-15 minutes is a nice win on a build, and this win is for almost every premerge CI run.  In terms of CI resource minutes saved per typical development day where there are many premerge CI runs, that's a sizeable savings overall.

I'm not a Jenkins/Blossom expert, but it seems like we could solve the need-to-build-Docker-sometimes problem by parameterizing the core premerge CI job with the Docker tag to use and then running the premerge flow like this:
1. premerge starts by launching a parent job that checks the branch to see if it modifies the Dockerfile
2. If Dockerfile modified then it launches the core premerge job with the standard Docker tag as a parameter
3. If Dockerfile not modified then:
  1. it launches a sub-job to build the Dockerfile with a unique temporary tag
  2. runs core premerge job with the temp Docker tag as a parameter
  3. removes the temp Docker image after core premerge complete

There's minimal additional complexity for the core premerge job where there's more churn than the Docker image contents.
The parent job that handles the Dockerfile-modified checks and removes the temp Docker image if necessary seems relatively straightforward, and we already have most of the code needed for the sub-job to build the Docker image.

Thoughts?  I'm totally fine if we want to hold off merging this change until the Dockerfile-modified logic is implemented so the premerge features remain intact and we don't temporarily lose the ability to check PRs that modify the Dockerfile.

pxLi:
> IMO 10-15 minutes is a nice win on a build, and this win is for almost every premerge CI run. In terms of CI resource minutes saved per typical development day where there are many premerge CI runs, that's a sizeable savings overall.
> 
> I'm not a Jenkins/Blossom expert, but it seems like we could solve the need-to-build-Docker-sometimes problem by parameterizing the core premerge CI job with the Docker tag to use and then running the premerge flow like this:
> 
> 1. premerge starts by launching a parent job that checks the branch to see if it modifies the Dockerfile
> 2. If Dockerfile modified then it launches the core premerge job with the standard Docker tag as a parameter
> 3. If Dockerfile not modified then:
> 4. it launches a sub-job to build the Dockerfile with a unique temporary tag
> 5. runs core premerge job with the temp Docker tag as a parameter
> 6. removes the temp Docker image after core premerge complete
> 
> There's minimal additional complexity for the core premerge job where there's more churn than the Docker image contents.
> The parent job that handles the Dockerfile-modified checks and removes the temp Docker image if necessary seems relatively straightforward, and we already have most of the code needed for the sub-job to build the Docker image.
> 
> Thoughts? I'm totally fine if we want to hold off merging this change until the Dockerfile-modified logic is implemented so the premerge features remain intact and we don't temporarily lose the ability to check PRs that modify the Dockerfile.

sounds OK. The reason we have complicated step for this is because its running on kubernetes, so we have to build the image in docker-in-docker container, and push the image to registry, then pull the image and start a new pod here for test running (we cannot just run the newly built image within the parent docker), at last we have to remove the temp image from registry to save space, but this is totally achievable.

I will try to add the Dockerfile-modified logic for this one later. Thanks!

pxLi:
build

jenkins/Jenkinsfile-blossom.premerge
```diff
@@ -118,16 +119,27 @@ pipeline {
                         def CUDA_NAME = sh(returnStdout: true,
                                 script: '. jenkins/version-def.sh>&2 && echo -n $CUDA_CLASSIFIER | sed "s/-/./g"')
 
-                        IMAGE_TAG = "dev-ubuntu16-${CUDA_NAME}"
-                        CACHE_IMAGE_NAME = "${ARTIFACTORY_NAME}/sw-spark-docker/plugin:${IMAGE_TAG}"
+                        // check if pre-merge dockerfile modified
+                        def dockerfileModified = sh(returnStdout: true,
+                                script: 'BASE=$(git --no-pager log --oneline -1 | awk \'{ print $NF }\'); ' +
```
jlowe, line 124: I don't think this does what we want.  Printing `$NF` is going to print the last word on the line which will be the last word in the comment one-liner, but that's not going to be useful for obtaining a commit hash.

I think we just want to use `git rev-parse HEAD` here, no `awk` processing necessary.

jenkins/Jenkinsfile-blossom.premerge
```diff
@@ -118,16 +119,27 @@ pipeline {
...
+                                        'git --no-pager diff --name-only HEAD $(git merge-base HEAD $BASE) ' +
+                                        '-- jenkins/Dockerfile-blossom.ubuntu16 || true')
```
jlowe, line 126: It would be nice to store the name of the CI Dockerfile in a variable at the top of this script and use the variable throughout so if we change the file path later it later it's easy to update.

jlowe (changes requested):

jenkins/Jenkinsfile-blossom.premerge
pxLi: @jlowe `git --no-pager log --oneline -1 | awk '{ print $NF }'` is trying to get the base commit for diff comparing, and the HEAD ref is used directly in `git merge-base HEAD $BASE`.

In our pre-merge cases, we fetch the PRs from remote `+refs/pull/*/merge:refs/remotes/origin/pr/*`, and we used the merged code (merge HEAD to BASE branch) to run the test to avoid people using out-dated code (but no merge conflict) to pass the CI.

So the HEAD commit message is always like `Merge HEAD into BASE`, e.g. `Merge 586f28f20dbe1621c743722627766e4439cf0014 into dc66f03a4664f41a2f16ddef4a587f81a3fc41c0`. I don't know if we have a better way to get the BASE commit w/o fetching the remote BASE branch, and I don't think `git rev-parse HEAD or origin/branch-0.5` would work here. Please let me know if any better option, thanks!

jenkins/Jenkinsfile-blossom.premerge
pxLi: sure, let me update it

pxLi:
build

jenkins/Jenkinsfile-blossom.premerge
```diff
@@ -118,16 +120,27 @@ pipeline {
...
+                        if (!dockerfileModified?.trim()) {
+                            TEMP_IMAGE_BUILD = false
+                        }
+
+                        if (TEMP_IMAGE_BUILD) {
+                            IMAGE_TAG = "dev-ubuntu16-${CUDA_NAME}"
+                            PREMERGE_TAG = "${IMAGE_TAG}-${BUILD_TAG}"
+                            IMAGE_PREMERGE = "${ARTIFACTORY_NAME}/sw-spark-docker-local/plugin:${PREMERGE_TAG}"
+                            def CUDA_VER = "$CUDA_NAME" - "cuda"
+                            docker.build(IMAGE_PREMERGE, "-f jenkins/Dockerfile-blossom.ubuntu16 --build-arg CUDA_VER=$CUDA_VER -t $IMAGE_PREMERGE .")
```
jlowe, line 137:
```suggestion
                            docker.build(IMAGE_PREMERGE, "-f ${PREMERGE_DOCKERFILE} --build-arg CUDA_VER=$CUDA_VER -t $IMAGE_PREMERGE .")
```

jlowe (changes requested):

pxLi:
build

jenkins/Jenkinsfile-blossom.premerge
pxLi: nice catch, thanks!

jlowe (approved):

pxLi:
Please let me merge this one and do a following verify PR to double check later, thanks!

PR #1878, opened 2021-03-05
spillable cache for GpuCartesianRDD

sperlingxx:
Use SpillableColumnarBatch to cache stream-side data in case of re-computation caused by nested-loop. Current PR is the substitution of #1784, which based on `branch-0.4`.


sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
```diff
@@ -161,7 +181,10 @@ class GpuCartesianRDD(
         filterTime,
         totalTime)
 
-      CompletionIterator[ColumnarBatch, Iterator[ColumnarBatch]](ret, table.close())
+      CompletionIterator[ColumnarBatch, Iterator[ColumnarBatch]](ret, {
```
jlowe, line 184: Do we need to worry about cases where the iterator does not fully iterate before the task completes?  Thinking of limit situations, etc.

jlowe (changes requested):

sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
```diff
@@ -141,15 +143,33 @@ class GpuCartesianRDD(
...
+
+    rdd1.iterator(currSplit.s1, context).zipWithIndex.flatMap { case (lhs, index) =>
       val table = withResource(lhs) { lhs =>
         GpuColumnVector.from(lhs.getBatch)
       }
-      // Ideally instead of looping through and recomputing rdd2 for
-      // each batch in rdd1 we would instead cache rdd2 in a way that
-      // it could spill to disk so we can avoid re-computation
+
+      val streamIterator = if (index == 0) {
```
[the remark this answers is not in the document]
sperlingxx: I think `spillBatchBuffer.isEmpty` will not work if `rdd2` return an empty iterator.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
```diff
@@ -87,11 +89,11 @@ class GpuSerializableBatch(batch: ColumnarBatch)
 }
 
 class GpuCartesianPartition(
-    idx: Int,
-    @transient private val rdd1: RDD[_],
-    @transient private val rdd2: RDD[_],
-    s1Index: Int,
-    s2Index: Int
+  idx: Int,
```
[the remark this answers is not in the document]
sperlingxx: Thank you, @jlowe! I've configured my IDE settings with rapids style.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
sperlingxx: Yes, it is a potential problem which may leads to memory leak. But I am not sure how to fix this problem, since the problem of `CompletionIterator` also exists in spark runtime.

revans2: The way to fix it is with a task completion listener.  You have to keep track if the table has been closed or not, and have both try to close it.  That way if the CompetionIterator closes it first that is great, but if it never closes it, then when the task finishes it will close it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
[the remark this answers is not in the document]
jlowe: Then use a local boolean to track.  Point is still valid: we don't really need indices here.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
```diff
@@ -190,10 +218,10 @@ object GpuNoColumnCrossJoin extends Arm {
     // Hash aggregate explodes the rows out, so if we go too large
     // it can blow up. The size of a Long is 8 bytes so we just go with
     // that as our estimate, no nulls.
-    val maxRowCount = targetSizeBytes/8
+    val maxRowCount = targetSizeBytes / 8
```
jlowe, line 221: In the future, please don't make a lot of these whitespace changes throughout the file.

sperlingxx: Sorry, I will be careful not to do this in next time.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
sperlingxx: I've added a `TaskCompletionListener` to ensure the GPU memory will be cleaned up.

sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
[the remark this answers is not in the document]
sperlingxx: > Then use a local boolean to track. Point is still valid: we don't really need indices here.

Yes, the `zipWithIndex` has been removed.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
```diff
@@ -141,15 +143,49 @@ class GpuCartesianRDD(
...
+    var streamSideCached = false
+    // a pointer to track buildTableOnFlight
+    var buildTableOnFlight: Option[Table] = None
+
+    // Add a taskCompletionListener to ensure the release of GPU memory. This listener will work
+    // if the CompletionIterator does not fully iterate before the task completes, which may
+    // happen if there exists specific plans like `LimitExec`.
+    context.addTaskCompletionListener[Unit]((_: TaskContext) => {
+      spillBatchBuffer.safeClose()
+      if (buildTableOnFlight.isDefined) buildTableOnFlight.get.close()
```
jlowe, line 159:
```suggestion
     buildTableOnFlight.foreach(_.close())
```

sperlingxx: Change applied, Thx!

sperlingxx:
build

sperlingxx:
build

jlowe (approved):

sperlingxx:
build

revans2 (approved):

Issue #1605, opened 2021-01-27
[FEA] Support sorting on nested type keys


kuhushukla:
**Is your feature request related to a problem? Please describe.**
Once https://github.com/rapidsai/cudf/issues/7226 is implemented we would want to test and add functionality to the plugin to be able to sort on nested type keys.

sameerz:
The PR that closed this issue supports single level sorting on these data types: BOOLEAN + BYTE + SHORT + INT + LONG + FLOAT + DOUBLE + DATE + TIMESTAMP + STRING + NULL + DECIMAL

PR #1883, opened 2021-03-05
Enable sort for single-level nesting struct columns on GPU

gerashegalov:
This PR contributes to #1605. It depends on https://github.com/rapidsai/cudf/pull/7422
- adds single-level struct type 


gerashegalov:
Status:
```shell
TEST_PARALLEL=1 SPARK_HOME=~/spark-3.1.1-bin-hadoop3.2 integration_tests/run_pyspark_from_build.sh \
  -k "test_single_orderby and struct"
```
we observe the failure:
```
ai.rapids.cudf.CudfException: cuDF failure at: /home/gshegalov/gits/cudf/cpp/src/sort/sort_column.cu:112: Column type must be relationally comparable
	at ai.rapids.cudf.Table.orderBy(Native Method)
	at ai.rapids.cudf.Table.orderBy(Table.java:1441)
	at com.nvidia.spark.rapids.GpuSorter.$anonfun$sort$2(SortUtils.scala:183)
```

jlowe (changes requested):
What's here looks good, but there are some things missing.

At some point this needs to be built with `mvn verify` to generate the changes to `supported_ops.md` and those changes should be in this PR.

It would be nice to have some negative testing for this, i.e.: testing that we don't try to place sorts of deeply nested struct key columns or non-default struct sort ordering on the GPU.

gerashegalov:
build

gerashegalov:
```bash
$ TEST_PARALLEL=1 COVERAGE_SUBMIT_FLAGS=$JDBSTR SPARK_HOME=~/spark-3.1.1-bin-hadoop3.2 integration_tests/run_pyspark_from_build.sh -k "test_single_struct and orderby"

XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a ASC NULLS FIRST'>0-Struct(['child0', Struct(['child0', Byte],['child1', Short],['child2', Integer],['child3', Long],['child4', Float],['child5', Double],['child6', String],['child7', Boolean],['child8', Date],['child9', Timestamp],['child10', Null])])]
  second-level structs are not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a ASC NULLS FIRST'>0-Array(String)]
  arrays are not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a ASC NULLS FIRST'>0-Map(String(not_null),Map(String(not_null),String))]
  maps are not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a ASC NULLS FIRST'>1-Struct(['child0', Struct(['child0', Byte],['child1', Short],['child2', Integer],['child3', Long],['child4', Float],['child5', Double],['child6', String],['child7', Boolean],['child8', Date],['child9', Timestamp],['child10', Null])])]
  second-level structs are not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a ASC NULLS FIRST'>1-Array(String)]
  arrays are not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a ASC NULLS FIRST'>1-Map(String(not_null),Map(String(not_null),String))]
  maps are not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a ASC NULLS LAST'>-Struct(['child0', Byte],['child1', Short],['child2', Integer],['child3', Long],['child4', Float],['child5', Double],['child6', String],['child7', Boolean],['child8', Date],['child9', Timestamp],['child10', Null])]
  opposite null order not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a ASC NULLS LAST'>-Struct(['child0', Struct(['child0', Byte],['child1', Short],['child2', Integer],['child3', Long],['child4', Float],['child5', Double],['child6', String],['child7', Boolean],['child8', Date],['child9', Timestamp],['child10', Null])])]
  opposite null order not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a ASC NULLS LAST'>-Array(String)]
  opposite null order not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a ASC NULLS LAST'>-Map(String(not_null),Map(String(not_null),String))]
  opposite null order not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a DESC NULLS LAST'>0-Struct(['child0', Struct(['child0', Byte],['child1', Short],['child2', Integer],['child3', Long],['child4', Float],['child5', Double],['child6', String],['child7', Boolean],['child8', Date],['child9', Timestamp],['child10', Null])])]
  second-level structs are not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a DESC NULLS LAST'>0-Array(String)]
  arrays are not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a DESC NULLS LAST'>0-Map(String(not_null),Map(String(not_null),String))]
  maps are not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a DESC NULLS FIRST'>-Struct(['child0', Byte],['child1', Short],['child2', Integer],['child3', Long],['child4', Float],['child5', Double],['child6', String],['child7', Boolean],['child8', Date],['child9', Timestamp],['child10', Null])]
  opposite null order not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a DESC NULLS FIRST'>-Struct(['child0', Struct(['child0', Byte],['child1', Short],['child2', Integer],['child3', Long],['child4', Float],['child5', Double],['child6', String],['child7', Boolean],['child8', Date],['child9', Timestamp],['child10', Null])])]
  opposite null order not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a DESC NULLS FIRST'>-Array(String)]
  opposite null order not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a DESC NULLS FIRST'>-Map(String(not_null),Map(String(not_null),String))]
  opposite null order not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a DESC NULLS LAST'>1-Struct(['child0', Struct(['child0', Byte],['child1', Short],['child2', Integer],['child3', Long],['child4', Float],['child5', Double],['child6', String],['child7', Boolean],['child8', Date],['child9', Timestamp],['child10', Null])])]
  second-level structs are not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a DESC NULLS LAST'>1-Array(String)]
  arrays are not supported
XFAIL ../../src/main/python/sort_test.py::test_single_nested_order_orderby[Column<'a DESC NULLS LAST'>1-Map(String(not_null),Map(String(not_null),String))]
  maps are not supported
=============== 4 passed, 4848 deselected, 20 xfailed in 43.43s ================
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -135,6 +135,8 @@ abstract class GpuTimeMath(
 
   override lazy val resolved: Boolean = childrenResolved && checkInputDataTypes().isSuccess
 
+  protected val microSecondsInOneDay: Long = TimeUnit.DAYS.toMicros(1)
```
jlowe, line 138: Curious, what do the changes in this file have to do with sorting of struct columns?

gerashegalov: nothing in particular, it's minor refactor that I noticed in the prior reviews and thought may not  be worth a dedicated PR.

jlowe: I personally prefer such things to be separate when they're not touching files that wouldn't normally be touched by the changes.  Otherwise where does it end?  You could end up with a PR that would normally only touch 3 files but ends up touching 20 because of other 'minor refactors'.  It blurs the focus of the patch and increases its risk footprint.  It also means the commit headline belies the changes being made, since a number of "oh, by the way, here's some unrelated stuff as well" changes come along silently for the ride.

gerashegalov: I am usually mindful of these concerns, however considered this minor enough and quite localized. If you insist, I'll remove it.

jlowe: I insist. :smile:  Touching unrelated files is where I draw the line during reviews.

gerashegalov:
Breaking tests can be traced to the fact that `null` in nested integer columns seems to be interpreted as `zero` in https://github.com/rapidsai/cudf/pull/7422

The minimum repro:
```
scala> val schema = new StructType().add("nestedInt", new StructType().add("value", IntegerType)  )
schema: org.apache.spark.sql.types.StructType = StructType(StructField(nestedInt,StructType(StructField(value,IntegerType,true)),true))

scala> val df = spark.createDataFrame(sc.parallelize(Seq( Row(Row(1)), Row(Row(null)), Row(Row(-1))  )), schema)
df: org.apache.spark.sql.DataFrame = [nestedInt: struct<value: int>]

scala> df.sort("nestedInt").select("nestedInt.value").show
21/03/15 04:49:38 WARN GpuOverrides: 
*Exec <TakeOrderedAndProjectExec> will run on GPU
  *Expression <SortOrder> nestedInt#18 ASC NULLS FIRST will run on GPU
  *Expression <Alias> cast(nestedInt#18.value as string) AS value#23 will run on GPU
    *Expression <Cast> cast(nestedInt#18.value as string) will run on GPU
      *Expression <GetStructField> nestedInt#18.value will run on GPU
  !NOT_FOUND <RDDScanExec> cannot run on GPU because no GPU enabled version of operator class org.apache.spark.sql.execution.RDDScanExec could be found
    @Expression <AttributeReference> nestedInt#18 could run on GPU

+-----+
|value|
+-----+
|   -1|
| null|
|    1|
+-----+
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2610,7 +2629,7 @@ object GpuOverrides {
         }),
     exec[CollectLimitExec](
       "Reduce to single partition and apply limit",
-      ExecChecks(TypeSig.commonCudfTypes + TypeSig.DECIMAL, TypeSig.all),
+      ExecChecks(sortOrderTypeSigs, TypeSig.all),
```
revans2, line 2632: Here too we should allow more through than we do.

revans2 (commented):
Looking good.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -423,6 +423,17 @@ object GpuOverrides {
     "\\S", "\\v", "\\V", "\\w", "\\w", "\\p", "$", "\\b", "\\B", "\\A", "\\G", "\\Z", "\\z", "\\R",
     "?", "|", "(", ")", "{", "}", "\\k", "\\Q", "\\E", ":", "!", "<=", ">")
 
+  // TODO need a better name since used in multiple contexts
```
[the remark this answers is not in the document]
revans2: The TODO comment is still here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1792,16 +1808,25 @@ object GpuOverrides {
...
+          pluginSupportedOrderableSig,
           TypeSig.orderable))),
-      (a, conf, p, r) => new BaseExprMeta[SortOrder](a, conf, p, r) {
+      (sortOrder, conf, p, r) => new BaseExprMeta[SortOrder](sortOrder, conf, p, r) {
+        override def tagExprForGpu(): Unit = {
+          val directionDefaultNullOrdering = sortOrder.direction.defaultNullOrdering
+          if (isNestedType(sortOrder.dataType) &&
+              sortOrder.nullOrdering != directionDefaultNullOrdering) {
+            willNotWorkOnGpu(s"Only default null ordering $directionDefaultNullOrdering " +
+              s"supported. Found: ${sortOrder.nullOrdering}")
```
revans2, line 1823: nit: for nested types

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2610,7 +2635,7 @@ object GpuOverrides {
         }),
     exec[CollectLimitExec](
       "Reduce to single partition and apply limit",
-      ExecChecks(TypeSig.commonCudfTypes + TypeSig.DECIMAL, TypeSig.all),
+      ExecChecks(TypeSig.all - TypeSig.UDT, TypeSig.all),
```
revans2, line 2638: I think this would allow an array of UDT through, we need to make sure that UDT is removed from nested as well.

revans2: Oh and SortExec should be updated to match these too.

revans2 (approved):

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2610,7 +2634,7 @@ object GpuOverrides {
         }),
     exec[CollectLimitExec](
       "Reduce to single partition and apply limit",
-      ExecChecks(TypeSig.commonCudfTypes + TypeSig.DECIMAL, TypeSig.all),
+      ExecChecks(TypeSig.all - TypeSig.UDT.nested(), TypeSig.all),
```
revans2, line 2637: Actually thinking about it more, we might have an issue with CalendarInterval, binary, and a few other types. I'm not sure that we support them on the GPU so calling them out as supported here might end up causing us to try to move them to the GPU and get an exception.  Can we either remove them from here as listed as supported, or make sure that we have a test somewhere to verify that they work as expected?

integration_tests/src/main/python/sort_test.py
```diff
@@ -34,6 +34,40 @@ def test_single_orderby(data_gen, order):
             lambda spark : unary_op_df(spark, data_gen).orderBy(order),
             conf = allow_negative_scale_of_decimal_conf)
 
+@pytest.mark.parametrize('shuffle_parts', [
+    pytest.param(1),
+    # This xfail would normally work but device-side asserts currently lead to memmory
+    # leaks, and they cause persitent bad state and good tests start failing
+    # pytest.param(200, marks=pytest.mark.xfail(reason="https://github.com/NVIDIA/spark-rapids/issues/1607"))
```
revans2, line 41: This scares me enough that I don't want to merge it in until it is fixed. I'd rather fall back to the CPU then risk these problems.

gerashegalov: Disabled GPU for numPartitions > 1

revans2 (changes requested):
GPU Sort uses upperbound, but not in all cases.

https://github.com/NVIDIA/spark-rapids/blob/ec2dcd7e6a34197fa11cc7fcc071523bee065e1b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala#L435

without https://github.com/rapidsai/cudf/issues/7690 that means we have a ticking time bomb in the code.

Also the single partition work around makes no since to me at all. Until that is also fixed I see no way for us to merge this in.

gerashegalov:
> Also the single partition work around makes no since to me at all. Until that is also fixed I see no way for us to merge this in.

I understand the concern about GPU Sort. 

As for the single partition workaround, It's just incremental delivery which mimics the progress we have on the cuDF side. It would make sense to me  as long as it works as documented.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2491,6 +2518,14 @@ object GpuOverrides {
         override val childExprs: Seq[BaseExprMeta[_]] =
           rp.ordering.map(GpuOverrides.wrapExpr(_, conf, Some(this)))
 
+        override def tagPartForGpu() {
+          val numPartitions = rp.numPartitions
+          if (numPartitions > 1) {
```
revans2, line 2523: This needs to be combined with a check for struct/nested type in the `rp.ordering`

```
if (numPartitions > 1 && re.ordering.exists(so => isNestedType(so.dataType))) {
...
```

gerashegalov: you are right

revans2 (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2737,9 +2772,15 @@ object GpuOverrides {
       "The backend for the sort operator",
       // The SortOrder TypeSig will govern what types can actually be used as sorting key data type.
       // The types below are allowed as inputs and outputs.
-      ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.ARRAY +
```
jlowe, line 2740: The original code allowed ArrayType and nested structs to participate in the sort but not as a sort key.  This seems to remove that support?  Also shown in a loss of coverage in the change to `supported_ops.md`.  Did we not have tests verifying that before?

revans2: Great catch I missed that.  We did not have tests verifying this before.  We should have.

gerashegalov:
build

jlowe (approved):
This looks better, but I didn't see a test added for the ability to sort a table containing arrays but not using arrays as the key.  Is that going to be done as a followup then?

gerashegalov:
@jlowe these [existing tests cover](https://github.com/NVIDIA/spark-rapids/blob/ea01e33f39afeeef5e0c7e46c1bf736d22bb7869/integration_tests/src/main/python/array_test.py#L65) the scenarios you are calling out. they are passing now locally

gerashegalov:
build

jlowe:
> these existing tests cover the scenarios you are calling out. they are passing now locally

Great, thanks @gerashegalov!

jlowe (approved):

revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
gerashegalov: thanks for catching this @jlowe

PR #1889, opened 2021-03-07
Decimal DIV changes 

razajafri:
This PR deals with the changes to the Decimal `DIV` in cudf.

There is a bug in rapidsai/cudf#7435 where it's returning a Decimal col when a BOOL8 col is expected. I have commented on the PR https://github.com/rapidsai/cudf/pull/7435#issuecomment-792217382

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
@@ -194,7 +194,7 @@ trait CudfBinaryExpression extends GpuBinaryExpression {
   def outputType(l: BinaryOperable, r: BinaryOperable) : DType = {
     val over = outputTypeOverride
     if (over == null) {
-      BinaryOperable.implicitConversion(l, r)
+      BinaryOperable.implicitConversion(binaryOp, l, r)
     } else {
       over
     }
@@ -203,12 +203,17 @@ trait CudfBinaryExpression extends GpuBinaryExpression {
   override def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector = {
     val lBase = lhs.getBase
     val rBase = rhs.getBase
-    val outType = outputType(lBase, rBase)
+    val outType = if (castOutputAtEnd) {
+      BinaryOperable.implicitConversion(binaryOp, lBase, rBase)
+    } else {
+      outputType(lBase, rBase)
+    }
     val tmp = lBase.binaryOp(binaryOp, rBase, outType)
     // In some cases the output type is ignored
-    if (!outType.equals(tmp.getType) && castOutputAtEnd) {
+    val castType = outputType(lBase, rBase)
+    if (!castType.equals(tmp.getType) && castOutputAtEnd) {
       withResource(tmp) { tmp =>
-        tmp.castTo(outType)
+        tmp.castTo(castType)
       }
     } else {
       tmp
@@ -217,12 +222,17 @@ trait CudfBinaryExpression extends GpuBinaryExpression {
 
   override def doColumnar(lhs: Scalar, rhs: GpuColumnVector): ColumnVector = {
     val rBase = rhs.getBase
-    val outType = outputType(lhs, rBase)
+    val outType = if (castOutputAtEnd) {
+      BinaryOperable.implicitConversion(binaryOp, lhs, rBase)
+    } else {
+      outputType(lhs, rBase)
+    }
     val tmp = lhs.binaryOp(binaryOp, rBase, outType)
     // In some cases the output type is ignored
-    if (!outType.equals(tmp.getType) && castOutputAtEnd) {
+    val castType = outputType(lhs, rBase)
+    if (!castType.equals(tmp.getType) && castOutputAtEnd) {
       withResource(tmp) { tmp =>
-        tmp.castTo(outType)
+        tmp.castTo(castType)
       }
     } else {
       tmp
@@ -231,12 +241,17 @@ trait CudfBinaryExpression extends GpuBinaryExpression {
 
   override def doColumnar(lhs: GpuColumnVector, rhs: Scalar): ColumnVector = {
     val lBase = lhs.getBase
-    val outType = outputType(lBase, rhs)
+    val outType = if (castOutputAtEnd) {
+      BinaryOperable.implicitConversion(binaryOp, lBase, rhs)
+    } else {
+      outputType(lBase, rhs)
+    }
     val tmp = lBase.binaryOp(binaryOp, rhs, outType)
     // In some cases the output type is ignored
-    if (!outType.equals(tmp.getType) && castOutputAtEnd) {
+    val castType = outputType(lBase, rhs)
+    if (!castType.equals(tmp.getType) && castOutputAtEnd) {
       withResource(tmp) { tmp =>
-        tmp.castTo(outType)
+        tmp.castTo(castType)
       }
     } else {
       tmp
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
@@ -200,17 +200,15 @@ trait CudfBinaryExpression extends GpuBinaryExpression {
     }
   }
 
-  override def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector = {
-    val lBase = lhs.getBase
-    val rBase = rhs.getBase
+  def doColumnar(lhs: BinaryOperable, rhs: BinaryOperable): ColumnVector = {
     val outType = if (castOutputAtEnd) {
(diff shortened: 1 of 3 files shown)


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -239,22 +239,99 @@ object GpuDivideUtil {
...
+    case _ => super.dataType
+  }
+
+  def getNewDecimalType(outputScale: Int, l: DecimalType, r: DecimalType): DecimalType = {
+    val newScale = if (outputScale + r.scale > l.scale) {
+      outputScale + r.scale
+    } else {
+      l.scale - outputScale
+    }
+    val newType = DecimalType(Math.min(18, l.precision + newScale), Math.min(18, newScale))
```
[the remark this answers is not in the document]
razajafri: Good catch, there already is one that I can use Decimal.MAX_LONG_DIGITS()

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
```diff
@@ -231,12 +241,17 @@ trait CudfBinaryExpression extends GpuBinaryExpression {
 
   override def doColumnar(lhs: GpuColumnVector, rhs: Scalar): ColumnVector = {
     val lBase = lhs.getBase
-    val outType = outputType(lBase, rhs)
+    val outType = if (castOutputAtEnd) {
+      BinaryOperable.implicitConversion(binaryOp, lBase, rhs)
+    } else {
+      outputType(lBase, rhs)
+    }
```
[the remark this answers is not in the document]
razajafri: Thanks, i have updated the code

razajafri:
@gerashegalov have I addressed all your concerns? PTAL

razajafri:
@revans2 can you take a look at this PR.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -239,22 +239,98 @@ object GpuDivideUtil {
   }
 }
 
+abstract class GpuDecimalDivide(left: Expression, right: Expression) extends GpuDivModLike with
```
revans2, line 242: Why create an abstract class when only one thing implements it?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -239,22 +239,98 @@ object GpuDivideUtil {
...
+  Serializable {
+  // Override the output type as a special case for decimal
+  override def dataType: DataType = (left.dataType, right.dataType) match {
+    case (l: DecimalType, r: DecimalType) => GpuDivideUtil.decimalDataType(l, r)
+    case _ => super.dataType
+  }
+
+  def getNewDecimalType(outputScale: Int, l: DecimalType, r: DecimalType): DecimalType = {
```
revans2, line 250: The name of this method does not explain anything about what it is doing, and there are no comments to explain it either.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -239,22 +239,98 @@ object GpuDivideUtil {
...
+    val newScale = if (outputScale + r.scale > l.scale) {
```
revans2, line 251: FYI: Spark has already make the lhs and rhs precision and scale the same.  This is part of the problem with multiply and divide is that Spark makes them both longer than they need to be so that they can be the same, which makes the spark binary op code happen, but results Spark doing things really inefficiently.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -239,22 +239,98 @@ object GpuDivideUtil {
...
+      outputScale + r.scale
+    } else {
+      l.scale - outputScale
+    }
+    val newType = DecimalType(Math.min(Decimal.MAX_LONG_DIGITS, l.precision + newScale),
```
revans2, line 256: I thought that the requirement was `r.scale + l.scale == outputScale`. This feels rather convoluted to get there.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -239,22 +239,98 @@ object GpuDivideUtil {
...
+      Math.min(Decimal.MAX_LONG_DIGITS, newScale))
+    val raise = newScale - l.scale
+    require(raise <= Decimal.MAX_LONG_DIGITS, s"$newType  is not supported for GPU processing yet.")
```
revans2, line 259: Is there ever a case where this will throw? If so then we really need the checks in tagging before we get to the point that this would throw an exception.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -239,22 +239,98 @@ object GpuDivideUtil {
...
+    newType
+  }
+
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector = {
+    (left.dataType, right.dataType) match {
+      case (l: DecimalType, r: DecimalType) => {
+        val outputScale = dataType.asInstanceOf[DecimalType].scale
+        val newType = getNewDecimalType(outputScale, l, r)
+        if (outputScale + r.scale > l.scale) {
```
revans2, line 268: Again r.scale == l.scale in all cases for Spark right now.

razajafri:
@revans2 I have addressed your concerns. Knowing what I know now, the math is pretty much just adding the r.scale to the outputType.scale so after div is called we end up with outputType.scale.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -245,16 +245,58 @@ case class GpuDivide(left: Expression, right: Expression) extends GpuDivModLike
...
-    case (l: DecimalType, r: DecimalType) =>  GpuDivideUtil.decimalDataType(l, r)
+    case (l: DecimalType, r: DecimalType) => GpuDivideUtil.decimalDataType(l, r)
     case _ => super.dataType
   }
+
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector = {
+    (left.dataType, right.dataType) match {
+      case (_: DecimalType, r: DecimalType) => {
+        val outType = dataType.asInstanceOf[DecimalType]
+        val newType = DecimalType(outType.precision, outType.scale + r.scale)
```
revans2, line 266: Can you please add a comment to explain the rules here about what is needed for the divide to work? It looks like from the CUDF code that for a divide the output scale must be `lhs.scale - rhs.scale = outputScale`.  I did a much of math and I think casting the lhs up is going to be fine until we hit 128-bit support for decimal and need to potentially check for overflow, so it would be good to have a comment about that here too.

razajafri: OK, I will do that. I thought it was unnecessary as we are already checking for it in `tagExprForGpu`

razajafri: nevermind, we are only checking for the intermediary result. 

I think we should add a check for `GpuDivideUtil.decimalDataType + r.scale` instead of what we are doing now? What do you think?

razajafri:
@revans2 can you check my math on the `tagExprForGpu` method?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1733,7 +1733,25 @@ object GpuOverrides {
           // decimal place so we can round it in GpuCheckOverflow
           (childExprs.head.dataType, childExprs(1).dataType) match {
             case (l: DecimalType, r: DecimalType) =>
-              val intermediateResult = GpuDivideUtil.decimalDataType(l, r)
+              val outputType = GpuDivideUtil.decimalDataType(l, r)
+              // We will never hit a case where outputType.precision < outputType.scale + r.scale.
```
revans2, line 1737: I still want a comment about the 128 bit use case

razajafri: Thanks for the clarification, I have added the

pxLi:
FYI the new cudf has just been built and deployed to snapshot

revans2 (approved):

jlowe:
build

jlowe:
build

jlowe (approved):
Premerge is failing for 3.1.2 tests because a recent 3.2.0 fix was cherry-picked.  All other tests passed.  Committing this since it is blocking all premerge CI for other PRs, including the one that will fix the 3.1.2 tests.

Issue #1558, opened 2021-01-20
[FEA] Initialize UCX early


abellina:
We have several issues that could be improved if UCX started early:

1. IPC mem handles are slow to open, especially on systems with many GPUs
2. nv_peer_mem registration takes time
3. UCX issues happen during the query, rather than upfront. 
4. We have seen pauses when handshaking UCX worker addresses. With all else going on in the application, the late start can complicate debugging these cases.

For issues (3, 4), if there is a configuration/network issue, we currently only detect it much later after the query has partially executed, which should be enough motivation for this issue.

UCX initialization also means that a ping message should be sent between peers. This would allow for those IPC and GPUDirectRDMA mappings to be done early.

PR #1891, opened 2021-03-08
Connect UCX endpoints early during app startup

abellina:

The following PR adds a heartbeat (default on, configurable) when the `RapidsShuffleManager` is enabled, allowing UCX to start connections to peers early, so it can exchange _rkeys_. Rkeys are opaque objects that UCX uses to encapsulate serialized handles to remote memory for RDMA and also for CUDA. Exchanging rkeys early helps because opening cuda IPC mem handles, and the various RoCE/Infiniband registrations, can take considerable time.

When all peers connect early, we have seen savings of up to 20%. This number is all over the place since we are lazily doing the registration as well. Especially for queries that are sub 1 minute in run length this can help because registration can be up to a few seconds, depending on system conditions.

Prior to this patch UCX connection establishment occurred at shuffle read time. `spark.rapids.shuffle.transport.earlyStart` can be set to `false` to go back to the late binding approach.

It is based on a patch from @jlowe. And early rkey unpacking was suggested by @Akshay-Venkatesh.

Closes https://github.com/NVIDIA/spark-rapids/issues/1558


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuShuffleEnv.scala
```diff
@@ -65,23 +65,28 @@ class GpuShuffleEnv(rapidsConf: RapidsConf) extends Logging {
 object GpuShuffleEnv extends Logging {
```
jlowe, line 65: 2021 copyrights

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleServer.scala
```diff
@@ -26,6 +26,7 @@ import com.nvidia.spark.rapids.RapidsPluginImplicits._
 import com.nvidia.spark.rapids.format.TableMeta
```
jlowe, line 26: 2021 copyrights

tests/src/test/scala/com/nvidia/spark/rapids/TestUtils.scala
```diff
@@ -19,15 +19,15 @@ package com.nvidia.spark.rapids
 import java.io.File
 
 import ai.rapids.cudf.{ColumnVector, DType, HostColumnVectorCore, Table}
+import org.mockito.Mockito._
```
jlowe, line 22:
```suggestion
```

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleHeartbeatManagerTest.scala
```diff
@@ -0,0 +1,87 @@
...
+
+    val exec1 = TrampolineUtil.newBlockManagerId("1", "peer", 123, Some("rapids=123"))
+    val exec2 = TrampolineUtil.newBlockManagerId("2", "peer2", 456, Some("rapids=456"))
+
+    // first executor
+    hbMgr.registerExecutor(exec1)
+    // second executor
+    hbMgr.registerExecutor(exec2)
+
+    // first executor hearbeats (finding out about executor 2)
```
jlowe, line 43:
```suggestion
    // first executor heartbeats (finding out about executor 2)
```

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleHeartbeatManagerTest.scala
```diff
@@ -0,0 +1,87 @@
...
```
jlowe, line 65:
```suggestion
    // first executor heartbeats (finding out about executor 2)
```

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleHeartbeatManagerTest.scala
```diff
@@ -0,0 +1,87 @@
...
+    // second heartbeat from exec1, should be empty
+    val secondUpdate = hbMgr.executorHeartbeat(exec1)
+    assertResult(0)(secondUpdate.ids.size)
+
+    // first heartbeat from exec2 returns empty, it already knew about exec1 on registration
+    val firstUpdateExec2 = hbMgr.executorHeartbeat(exec2)
+    assertResult(0)(firstUpdateExec2.ids.size)
+  }
+
+  test("an executor tells hearbeat handler about new peers") {
```
jlowe, line 79:
```suggestion
  test("an executor tells heartbeat handler about new peers") {
```

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -458,31 +475,68 @@ class UCX(executorId: Int, usingWakeupFeature: Boolean = true) extends AutoClose
...
+        try {
+          buffers.foreach { buffer =>
+            val mmapParam = new UcpMemMapParams()
+                .setAddress(buffer.getAddress)
+                .setLength(buffer.getLength)
+
+            //note that this can throw, lets call back and let caller figure out how to handle
+            try {
+              val registered = context.memoryMap(mmapParam)
+              registeredMemory.synchronized {
```
jlowe, line 509: Does it make sense to lock over and over here?  `pendingRegistration` is true, so there shouldn't be anyone waiting  on this lock, instead they'll be waiting to see `pendingRegistration` turn true.  Seems like we should grab the lock outside of `buffers.foreach` to reduce the locking overhead without any downsides of slowing another thread down trying to access.  Thoughts?

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleHeartbeatManagerTest.scala
```diff
@@ -0,0 +1,87 @@
...
+
+    val exec1 = TrampolineUtil.newBlockManagerId("1", "peer", 123, Some("rapids=123"))
+    val exec2 = TrampolineUtil.newBlockManagerId("2", "peer2", 456, Some("rapids=456"))
+
+    // first executor
+    hbMgr.registerExecutor(exec1)
+    // second executor
+    hbMgr.registerExecutor(exec2)
+
+    // first executor hearbeat (finding out about executor 2)
```
jlowe, line 43:
```suggestion
    // first executor heartbeat (finding out about executor 2)
```

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleHeartbeatManagerTest.scala
```diff
@@ -0,0 +1,87 @@
...
```
jlowe, line 65:
```suggestion
    // first executor heartbeat (finding out about executor 2)
```

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
abellina: Yeap that's better for sure, fixing.

abellina:
build

jlowe (approved):

abellina:
Thanks @jlowe

PR #1893, opened 2021-03-08
Plug-in support for get_json_object

razajafri:
This PR adds `get_json_object` support to the rapids plugin. This is a DRAFT PR that shouldn't be committed until the related cudf changes are in /rapidsai/cudf#7286


pom.xml
```diff
@@ -147,7 +147,7 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <spark.version>${spark300.version}</spark.version>
         <cuda.version>cuda10-1</cuda.version>
-        <cudf.version>0.18-SNAPSHOT</cudf.version>
+        <cudf.version>0.19-SNAPSHOT</cudf.version>
```
jlowe, line 150: Need to upmerge to branch-0.5.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGetJsonObject.scala
```diff
@@ -0,0 +1,49 @@
...
+case class GpuGetJsonObject(left: Expression, right: Expression) extends GpuBinaryExpression with
+  ExpectsInputTypes {
+
+  override def dataType: DataType = StringType
+  override def inputTypes: Seq[DataType] = Seq(StringType, StringType)
+  override def nullable: Boolean = true
+  override def prettyName: String = "get_json_object"
+
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector = {
+    throw new UnsupportedOperationException("We don't support json path variable as a vector")
```
jlowe, line 33: Avoid "We" in an exception message the user sees.  IMO better to word this something like, `JSON path must be a scalar value`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGetJsonObject.scala
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  override def doColumnar(lhs: Scalar, rhs: GpuColumnVector): ColumnVector = {
+    throw new UnsupportedOperationException("We don't support json path variable as a vector")
```
jlowe, line 37: Similar exception comment cleanup here

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGetJsonObject.scala
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package com.nvidia.spark.rapids
+
+import ai.rapids.cudf.{ColumnVector, Scalar}
+
+import org.apache.spark.sql.catalyst.expressions.{ExpectsInputTypes, Expression}
+import org.apache.spark.sql.types.{DataType, StringType}
+
+case class GpuGetJsonObject(left: Expression, right: Expression) extends GpuBinaryExpression with
```
jlowe, line 24: I think it's clearer to follow the Spark code here and name these parameters `json` and `path` with overrides to `left` and `right` so it's very clear which is which up front.

razajafri:
build

razajafri:
@jlowe have I addressed all your concerns?

integration_tests/src/main/python/json_test.py
```diff
@@ -0,0 +1,34 @@
...
+from asserts import assert_gpu_and_cpu_are_equal_collect
+from data_gen import *
+from pyspark.sql.types import *
+
+def mk_json_str_gen(pattern):
+    return StringGen(pattern).with_special_case('').with_special_pattern('.{0,10}')
+
+@pytest.mark.parametrize('json_str_pattern', ["\\{\"store\": \{\"fruit\": \[\\{" "\"weight\":\d,\"type\":\"[a-z]{1,9}\"\\}\], " \
+                   "\"bicycle\":\\{\"price\":\d\d\.\d\d,\"color\":\"[a-z]{0,4}\"\\}\\},\"email\":\"[" \
+                   "a-z]{1,5}\\@[a-z]{3,10}\.com\",\"owner\":\"[a-z]{3,8}\"\\}", "\\{\"a\": \"[a-z]{1,3}\"\\}"], ids=idfn)
```
jlowe, line 26: This would be a _lot_ clearer to read without all the escaped characters.  Recommend using raw strings here, something like this (untested):
```suggestion
@pytest.mark.parametrize('json_str_pattern', [r'\{"store": \{"fruit": \[\{"weight":\d,"type":"[a-z]{1,9}"\}\], ' \
                   r'"bicycle":\{"price":\d\d\.\d\d,"color":"[a-z]{0,4}"\}\},' \
                   r'"email":"[a-z]{1,5}\@[a-z]{3,10}\.com","owner":"[a-z]{3,8}"\}',
                   r'\{"a": "[a-z]{1,3}"\}'], ids=idfn)
```

integration_tests/src/main/python/json_test.py
```diff
@@ -0,0 +1,34 @@
...
+def test_get_json_object(json_str_pattern):
+    gen = mk_json_str_gen(json_str_pattern)
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: unary_op_df(spark, gen, length=10).selectExpr(
+            'get_json_object(a,\"$.a\")',
+            'get_json_object(a, \"$.owner\")',
+            'get_json_object(a, \"$.store.fruit[0]\")'),
```
jlowe, line 33: Unnecessary escaping here.
```suggestion
            'get_json_object(a,"$.a")',
            'get_json_object(a, "$.owner")',
            'get_json_object(a, "$.store.fruit[0]")'),
```

razajafri:
build

jlowe (approved):

PR #1903, opened 2021-03-09
Fix issues with cached partitions not matching

### LINKED ISSUE #1885: [BUG] natural join on string key results in a data frame with spurious NULLs
**Describe the bug**

I have a data frame SRC that is backed by a Parquet file, and two data frames AGG1 and AGG2 that have the same set of values for a key column but different columns for the remaining columns (these column values come from performing different aggregates on SRC).  When I perform a natural join on AGG1 and AGG2, generating a new data frame JOINED, the result appears as I would expect it to:  one record for each key value, with the union of the columns of AGG1 and AGG2.  Furthermore, filtering JOINED with `isnull` or `isnan` on each column reveals no null or NaN values in any column.  However, when running on GPU (but not CPU):

1. `JOINED.dropna().count()` is approximately 0.05% of `JOINED.count()`
2. using `JOINED` on the left side of another join results in NULLs for the columns in `JOINED` in the output frame

Relevant query plans for the input relations are [here](https://gist.github.com/willb/713a10abe6df1accbec53503f46e462f) and an example REPL session exhibiting this behavior is [here](https://gist.github.com/willb/91d5bc8dfc112200885a61bbf87631b5).

1. run `spark-submit --driver-memory 16G generate.py` from [this repo](https://github.com/willb/telco-churn-augmentation)
2. submit [this reproducer script](https://gist.github.com/willb/f63bed40752cd75d27da203d8ba8fdcc) to Spark with and without the GPU plugin loaded (or run the linked notebook).

**Expected behavior**

The output should show that `customer_charges` had no records with NULL values whether running on CPU or GPU.

This requires no special Spark configuration.  I've reproduced it with a local master running under Docker with at least the following configurations:

- rapids-4-spark 0.2.0, cudf 0.15, cuda 10.1
- rapids-4-spark 0.3.0, cudf 0.17, cuda 11
- rapids-4-spark 0.4.0, cudf 0.18, cuda 11

**Additional context**

revans2:
This fixes #1885 

This is a rather large and comprehensive change for this late in the 0.4 release cycle. Please help test this as much as possible I am going to do the same on my end.

### DIFF
--- a/integration_tests/src/main/python/data_gen.py
+++ b/integration_tests/src/main/python/data_gen.py
@@ -276,6 +276,21 @@ class RepeatSeqGen(DataGen):
         self._start(rand, self._loop_values)
         self._vals = [self._child.gen() for _ in range(0, self._length)]
 
+class SetValuesGen(DataGen):
+    """A set of values that are randomly selected"""
+    def __init__(self, data_type, data):
+        super().__init__(data_type, nullable=False)
+        self.nullable = any(x is None for x in data)
+        self._vals = data
+
+    def __repr__(self):
+        return super().__repr__() + '(' + str(self._child) + ')'
+
+    def start(self, rand):
+        data = self._vals
+        length = len(data)
+        self._start(rand, lambda : data[rand.randrange(0, length)])
+
 FLOAT_MIN = -3.4028235E38
 FLOAT_MAX = 3.4028235E38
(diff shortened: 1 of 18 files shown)


revans2:
build

revans2:
I ran tests on 3.0.2, but I am seeing some tests fail on 3.1.1. I am going to start digging into those.

revans2:
This is not going to work. AQE makes it so it is impossible to do what we want to try with this.

jlowe (commented):
Filing these in case this gets reopened.

PR #1841, opened 2021-03-02
Add in support for DateAddInterval

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/1612
Signed-off-by: Niranjan Artal <nartal@nvidia.com>

PR #1906, opened 2021-03-09
Remove get call on timeZoneId

nartal1:
This a follow on PR based on review comment https://github.com/NVIDIA/spark-rapids/pull/1841#discussion_r585206250 .
Updated the code where get is called on timeZoneId(Option).

### DIFF
--- a/shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
+++ b/shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
@@ -17,7 +17,6 @@
 package com.nvidia.spark.rapids.shims.spark300
 
 import java.nio.ByteBuffer
-import java.time.ZoneId
 
 import scala.collection.mutable.ListBuffer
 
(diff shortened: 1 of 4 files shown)


gerashegalov (approved):
LGTM, thanks @nartal1

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
```diff
@@ -255,8 +255,9 @@ class Spark300Shims extends SparkShims {
                 }
               case _ =>
             }
-            if (ZoneId.of(a.timeZoneId.get).normalized() != GpuOverrides.UTC_TIMEZONE_ID) {
-              willNotWorkOnGpu("Only UTC zone id is supported")
+            a.timeZoneId.foreach {
+              case zoneId if ZoneId.of(zoneId).normalized() != GpuOverrides.UTC_TIMEZONE_ID =>
+                willNotWorkOnGpu(s"Only UTC zone id is supported. Actual zone id: $zoneId")
```
jlowe, line 260: Isn't this going to trigger a `MatchError` if this one case doesn't match?  e.g.:
```
scala> Some(2).foreach{ case x if x != 2 => println(x) }
scala.MatchError: 2 (of class java.lang.Integer)
  at $anonfun$res1$1(<console>:26)
  at scala.runtime.java8.JFunction1$mcVI$sp.apply(JFunction1$mcVI$sp.java:23)
  at scala.collection.immutable.List.foreach(List.scala:392)
  ... 47 elided
```

jlowe (changes requested):

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/Spark300Shims.scala
gerashegalov: good catch, it's either should be a `collect` or we need to have a default `case _`

jlowe: Or let's not out-fancy ourselves and just do it in a very straightforward way:
```scala
a.timeZoneId.foreach { zoneId =>
  if (ZoneId.of(zoneId).normalized() != GpuOverrides.UTC_TIMEZONE_ID) {
    willNotWorkOnGpu(s"Only UTC zone id is supported. Actual zone id: $zoneId")
  }
}
```

gerashegalov: this will do too

gerashegalov: And we should factor it out since this fragment is used many times

nartal1: Thanks for the suggestions. Will update the PR.

nartal1:
@jlowe @gerashegalov  Updated as per suggestion.
Please let me know if the additional changes(renaming, formatting) are okay to be in this PR.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -408,6 +408,19 @@ trait GpuOverridesListener {
       costOptimizations: Seq[Optimization])
 }
 
+/**
+ * trait to check if the zone id is in UTC
+ */
+trait TimeZoneCheck {
+  def checkTimeZoneId(timeZoneId: Option[String], meta: RapidsMeta[_, _, _]): Unit = {
```
jlowe, line 415: This seems more appropriate as a static function than a trait.  We aren't going to use the trait class as a discriminator, and the trait doesn't leverage anything the class provides.

gerashegalov (dismissed):
LGTM

nartal1:
build

jlowe (approved):

gerashegalov (approved):
LGTM

Issue #1907, opened 2021-03-10
[BUG] Remove the Scala version of Mortgage ETL tests from nightly test


NvTimLiu:
**Describe the bug**
The Scala API of Mortgage ETL tests is unneeded, it just runs to check if it crashes or not, and it does not check the test results either.
https://github.com/NVIDIA/spark-rapids/blob/branch-0.5/jenkins/spark-tests.sh#L81

The Mortgage ETL test has already been included in the python integration tests.
https://github.com/NVIDIA/spark-rapids/blob/branch-0.5/integration_tests/src/main/python/mortgage_test.py#L25

NvTimLiu:
close as PR #1908 merged

PR #1908, opened 2021-03-10
Remove the Scala version of Mortgage ETL tests from nightly test

NvTimLiu:
Issue: https://github.com/NVIDIA/spark-rapids/issues/1907

The Scala API of Mortgage ETL tests is unneeded, it just runs to check if it crashes or not, and it does not check the test results either.

The Mortgage ETL test has already been included in the python integration tests.


NvTimLiu:
build

PR #1909, opened 2021-03-10
Avoid pinned memory for shuffle host buffers

jlowe:
When using legacy Spark shuffle the buffers are first collected into host buffers that are later concatenated into one monolithic host buffer by `GpuShuffleCoalesceExec` that is sent to the GPU.  The intermediate buffers do not need to be allocated with pinned memory because they are not sent directly to the device. This reduces the load on the pinned memory pool.  Because the pinned memory pool allocator currently does not perform well with many allocations, see https://github.com/rapidsai/cudf/issues/7553, this change shows noticeable speedup on regular queries with the default 200 partitions.


jlowe:
build

tgravescs (approved):

PR #1910, opened 2021-03-10
Make hash partitioning match CPU

### LINKED ISSUE #1885: [BUG] natural join on string key results in a data frame with spurious NULLs

revans2:
this fixes #1885

There is one test I commented out that fails, although it looks like it is an existing issue with cudf.


integration_tests/src/main/python/join_test.py
```diff
@@ -201,3 +201,90 @@ def do_join(spark):
...
+        # so don't let the shuffle be on the GPU when we repartition and cache
+        # the data
+        spark.conf.set('spark.rapids.sql.exec.ShuffleExchangeExec', 'false')
+        left = gen_df(spark, left_gen, length=500)
+        right = gen_df(spark, right_gen, length=500)
+        cached = right.repartition('r_a').cache()
+        cached.count() # populate the cache
+        # Now turn it back on so the other half of the shuffle will be on the GPU
+        spark.conf.set('spark.rapids.sql.exec.ShuffleExchangeExec', 'true')
+        return left.join(cached, left.a == cached.r_a, join_type)
```
jlowe, line 241: Nit: There's a lot of common code between this and test_half_cache_join_cpu_left, would be good to refactor the common parts out and parameterize.  Not must-fix.

integration_tests/src/main/python/repart_test.py
```diff
@@ -17,6 +17,7 @@
 from asserts import assert_gpu_and_cpu_are_equal_collect
```
jlowe, line 17: 2021 copyrights

integration_tests/src/test/scala/com/nvidia/spark/rapids/UnaryOperatorsSuite.scala
```diff
@@ -37,6 +37,6 @@ class UnaryOperatorsSuite extends SparkQueryCompareTestSuite {
   }
```
jlowe, line 37: 2021 copyrights

integration_tests/src/test/scala/com/nvidia/spark/rapids/UnaryOperatorsSuite.scala
```diff
@@ -37,6 +37,6 @@ class UnaryOperatorsSuite extends SparkQueryCompareTestSuite {
...
 
   testSparkResultsAreEqual("Test murmur3", mixedDfWithNulls) {
-    frame => frame.selectExpr("hash(longs, doubles, 1, null, 'stock string', ints, strings)")
```
jlowe, line 40: Do we want an xfail test with a corresponding issue for double hashing?  This looks like we're punting on the issue in a different way.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuHashPartitioning.scala
```diff
@@ -18,11 +18,11 @@ package com.nvidia.spark.rapids
```
jlowe, line 18: 2021 copyrights

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuHashPartitioning.scala
```diff
@@ -47,96 +47,67 @@ case class GpuHashPartitioning(expressions: Seq[Expression], numPartitions: Int)
...
+        val parts = withResource(new NvtxRange("Calculate part", NvtxColor.CYAN)) { _ =>
+          withResource(GpuMurMur3Hash.compute(batch, expressions)) { hash =>
+            withResource(GpuScalar.from(numPartitions, IntegerType)) { partsLit =>
+              hash.pmod(partsLit, DType.INT32)
+            }
+          }
+        }
+        withResource(new NvtxRange("sort by part", NvtxColor.DARK_GREEN)) { _ =>
+          withResource(parts) { parts =>
+            val allColumns = ArrayBuffer.empty[ColumnVector]
```
jlowe, line 64:
```suggestion
            val allColumns = new ArrayBuffer[ColumnVector](batch.numCols + 1)
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2250,14 +2250,15 @@ object GpuOverrides {
       "Murmur3 hash operator",
       ExprChecks.projectNotLambda(TypeSig.INT, TypeSig.INT,
         repeatingParamCheck = Some(RepeatingParamCheck("input",
+          // Floating poitn values don't work because of -0.0 is not hashed properly
```
jlowe, line 2253:
```suggestion
          // Floating point values don't work because of -0.0 is not hashed properly
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2378,6 +2379,16 @@ object GpuOverrides {
         override val childExprs: Seq[BaseExprMeta[_]] =
           hp.expressions.map(GpuOverrides.wrapExpr(_, conf, Some(this)))
 
+        override def tagPartForGpu(): Unit = {
+          // This needs to match what murmur3 supports. in 0.5 we should make the checks
```
jlowe, line 2383:
```suggestion
          // This needs to match what murmur3 supports. In 0.5 we should make the checks
```

jlowe: Worth linking to a followup issue here?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/HashFunctions.scala
```diff
@@ -40,31 +36,37 @@ case class GpuMd5(child: Expression)
   }
 }
 
-case class GpuMurmur3Hash(child: Seq[Expression]) extends GpuExpression {
+object GpuMurMur3Hash extends Arm {
```
jlowe, line 39:
```suggestion
object GpuMurmur3Hash extends Arm {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/HashFunctions.scala
```diff
@@ -40,31 +36,37 @@ case class GpuMd5(child: Expression)
...
+  def compute(batch: ColumnarBatch, boundExpr: Seq[Expression], seed: Int = 42): ColumnVector = {
+    val newExprs = boundExpr.map { expr =>
+      expr.dataType match {
+        case ByteType | ShortType =>
+          GpuCast(expr, IntegerType)
+        case DoubleType =>
+          // We have to normalize the NaNs, but not zeros
+          // however the current cudf code does the wrong thing for -0.0
```
jlowe, line 47: Should we link to followup issue for double/float problems?

integration_tests/src/test/scala/com/nvidia/spark/rapids/UnaryOperatorsSuite.scala
revans2: I personally would be fine completely deleting the test.  It did not cover much of anything and what we have now is much better. I will mark the double/float tests as xfail for now and point them to an issue to fix them.

revans2:

revans2:
build

jlowe:
Looks like there are two test failures:
```
08:53:58  AdaptiveQueryExecSuite:
08:54:04   skewed inner join optimization *** FAILED ***
08:54:04    ArrayBuffer(99) had length 1 instead of expected length 2 (AdaptiveQueryExecSuite.scala:463)
08:54:09   skewed left outer join optimization *** FAILED ***
08:54:09    ArrayBuffer(99) had length 1 instead of expected length 2 (AdaptiveQueryExecSuite.scala:463)
```

jlowe (approved):

revans2:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
```diff
@@ -69,6 +69,14 @@ public static synchronized void debug(String name, ai.rapids.cudf.ColumnVector c
     }
   }
 
+  private static String hexString(byte[] bytes) {
+    StringBuilder str = new StringBuilder();
```
andygrove, line 73: nit: could create with capacity here to avoid reallocations

revans2:
Please don't merge this yet. A test is commented out and we are working on a fix for cudf to be in 0.18.1 so we can turn the test back on again.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
gerashegalov: nit: we can use [Hex.encodeHexString(bytes)|](https://commons.apache.org/proper/commons-codec/archives/1.10/apidocs/index.html)

jlowe:
> Please don't merge this yet.

Made this PR a draft to enforce this.

revans2:
build

revans2:
Removed the draft because we are still working out how to deal with the cudf issue. I thought we were going to merge this in as is, and then update the CUDF version.

revans2:
OK the tests are still failing waiting on a fix from cudf for 0.18.1

NvTimLiu:
Depend on https://github.com/NVIDIA/spark-rapids/pull/1927

jlowe (approved):

pxLi:
build

PR #1911, opened 2021-03-10
Fix bug where cost-based optimizer was forcing GPU CustomShuffleReaderExec onto CPU

andygrove:
The experimental cost-based optimizer determines CPU vs GPU costs per operator and forces sections of the plan back onto the CPU if there is no benefit in having that section on the GPU. Unfortunately, it was not taking into account the fact that some operators must run on GPU. For example, with AQE enabled we must have a GPU `CustomShuffleReaderExec` if the shuffle already ran on the GPU, and forcing it back onto the CPU results in an invalid plan.

This PR introduces yet another type of tag in `RapidsMeta` for tagging operators that must run on GPU and the optimizer is updated to respect this. I also added a unit test to reproduce the bug.

I'm not sure if adding another tag in `RapidsMeta` is the best solution so am looking for feedback on that. 

I tested the change locally with our benchmarks and it fixes the 13 queries that were previously failing at SF=100.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
@@ -77,7 +77,7 @@ class CostBasedOptimizer(conf: RapidsConf) extends Logging {
         val transitionCost = plan.childPlans.filter(!_.canThisBeReplaced)
             .map(costModel.transitionToGpuCost).sum
         val gpuCost = operatorGpuCost + transitionCost
-        if (gpuCost > operatorCpuCost) {
+        if (gpuCost > operatorCpuCost && !plan.mustThisBeReplaced) {
           optimizations.append(AvoidTransition(plan))
           plan.costPreventsRunningOnGpu()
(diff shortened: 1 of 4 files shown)


jlowe:
We're in the process of switching the shuffle partitioning to be CPU-compatible in #1910, so it would be interesting to pull that change in (or CBO into 0.4).  IIUC this shouldn't be necessary with that other change in place, and it would be good to test (both to verify and get additional testing on #1910).

andygrove:
> We're in the process of switching the shuffle partitioning to be CPU-compatible in #1910, so it would be interesting to pull that change in (or CBO into 0.4). IIUC this shouldn't be necessary with that other change in place, and it would be good to test (both to verify and get additional testing on #1910).

@jlowe @revans2 I branched from #1910 and added in CBO and confirmed that all the problematic queries ran without the fix that I have in this PR :tada: so I would be happy to close this if we merge the hash partitioning fix.

Making hash partitioning match certainly makes the CBO logic simpler and also allows us more flexibility in optimizing query stages when AQE is enabled because we can mix and match CPU and GPU per query stage as appropriate.

jlowe:
At this point we're set on merging #1910, so closing this.

PR #1919, opened 2021-03-11
Add support for Structs for UnionExec

razajafri:
This PR whitelabels UnionExec for structs and adds tests for 3.1.1

### DIFF
--- a/integration_tests/src/main/python/repart_test.py
+++ b/integration_tests/src/main/python/repart_test.py
@@ -21,8 +21,10 @@ from marks import ignore_order
 import pyspark.sql.functions as f
 
 nested_scalar=pytest.mark.xfail(reason="https://github.com/NVIDIA/spark-rapids/issues/1459")
-@pytest.mark.parametrize('data_gen', [pytest.param((StructGen([['child0', DecimalGen(7, 2)]]), StructGen([['child1', IntegerGen()]])), marks=nested_scalar),
-                                      (StructGen([['child0', DecimalGen(7, 2)]], nullable=False), StructGen([['child1', IntegerGen()]], nullable=False))], ids=idfn)
+@pytest.mark.parametrize('data_gen', [pytest.param((StructGen([['child0', DecimalGen(7, 2)]]),
+                                                    StructGen([['child1', IntegerGen()]])), marks=nested_scalar),
+                                      (StructGen([['child0', DecimalGen(7, 2)]], nullable=False),
+                                       StructGen([['child1', IntegerGen()]], nullable=False))], ids=idfn)
 def test_union_struct(data_gen):
     left_gen, right_gen = data_gen
     assert_gpu_and_cpu_are_equal_collect(
--- a/integration_tests/src/main/python/repart_test.py
+++ b/integration_tests/src/main/python/repart_test.py
@@ -37,7 +37,7 @@ def test_union(data_gen):
             lambda spark : binary_op_df(spark, data_gen).union(binary_op_df(spark, data_gen)))
 
 @pytest.mark.parametrize('data_gen', all_gen + [pytest.param(all_basic_struct_gen, marks=nested_scalar_mark),
-                                      StructGen([[ 'child0', DecimalGen(7, 2)]], nullable=False)])
+                                                pytest.param(StructGen([[ 'child0', DecimalGen(7, 2)]], nullable=False), marks=nested_scalar_mark)])
 @pytest.mark.skipif(is_before_spark_311(), reason="This is supported only in Spark 3.1.1+")
 def test_union_by_missing_col_name(data_gen):
     assert_gpu_and_cpu_are_equal_collect(
--- a/integration_tests/src/main/python/repart_test.py
+++ b/integration_tests/src/main/python/repart_test.py
@@ -25,13 +25,17 @@ nested_scalar_mark=pytest.mark.xfail(reason="https://github.com/NVIDIA/spark-rap
                                                     StructGen([['child1', IntegerGen()]])), marks=nested_scalar_mark),
                                       (StructGen([['child0', DecimalGen(7, 2)]], nullable=False),
                                        StructGen([['child1', IntegerGen()]], nullable=False))], ids=idfn)
-def test_union_struct(data_gen):
+@pytest.mark.skipif(is_before_spark_311(), reason="This is supported only in Spark 3.1.1+")
+# This tests the union of DF of structs with different types of cols as long as the struct itself
+# isn't null. This is a limitation in cudf because we don't support nested types as literals
+def test_union_struct_missing_children(data_gen):
     left_gen, right_gen = data_gen
     assert_gpu_and_cpu_are_equal_collect(
         lambda spark : binary_op_df(spark, left_gen).unionByName(binary_op_df(
             spark, right_gen), True))
 
 @pytest.mark.parametrize('data_gen', all_gen + [all_basic_struct_gen, StructGen([['child0', DecimalGen(7, 2)]])], ids=idfn)
+# This tests union of two DFs of two cols each. The types of the left col and right col is the same
 def test_union(data_gen):
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : binary_op_df(spark, data_gen).union(binary_op_df(spark, data_gen)))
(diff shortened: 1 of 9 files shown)


integration_tests/src/main/python/repart_test.py
```diff
@@ -15,15 +15,26 @@
...
+@pytest.mark.parametrize('data_gen', all_gen + [all_basic_struct_gen, StructGen([['child0', DecimalGen(7, 2)]])], ids=idfn)
 def test_union(data_gen):
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : binary_op_df(spark, data_gen).union(binary_op_df(spark, data_gen)))
 
-@pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
+@pytest.mark.parametrize('data_gen', all_gen + [all_basic_struct_gen, StructGen([['child0', DecimalGen(7, 2)]])], ids=idfn)
+@pytest.mark.skipif(is_before_spark_311(), reason="This is supported only in Spark 3.1.1+")
+def test_union_by_missing_col_name(data_gen):
+    if (isinstance(data_gen, StructGen)):
```
gerashegalov, line 30: xfail would stick around for all further test case iterations once activated in my experience. Some examples of the [documented](https://docs.pytest.org/en/stable/parametrize.html) way of registering [parameter-specific xfail are e.g here](https://github.com/NVIDIA/spark-rapids/blob/63a2e3df1255f09b550d56103525621cbf6bfe56/integration_tests/src/main/python/sort_test.py#L53)

gerashegalov (changes requested):

integration_tests/src/main/python/repart_test.py
razajafri: I have updated but the xfail doesn't persist or stick, can you please correct me if I am wrong. I tested it by placing the struct data_gen before the all_gen and it was not `xfail`ing every test after struct. 

Nevertheless, I have made the change to follow the convention.

razajafri:
@sameerz we won't be able to support the union of structs in cases where the struct can have a null value. I have added tests for all the cases and xfailed the ones that we don't support ATM. 

This is still dependent on the cudf support for Scalar Structs

razajafri:
@gerashegalov PTAL

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2630,7 +2630,10 @@ object GpuOverrides {
       (shuffle, conf, p, r) => new GpuShuffleMeta(shuffle, conf, p, r)),
     exec[UnionExec](
       "The backend for the union operator",
-      ExecChecks(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all),
+      ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.STRUCT
+        .withPsNote(TypeEnum.STRUCT, "unionByName will not optionally fill missing columns with " +
+          "nulls when the col has structs"))
+        .nested(), TypeSig.all),
```
gerashegalov, line 2636: do we want to allow this for multi- level nesting right away? In the sort PR we are doing [single-level first](https://github.com/NVIDIA/spark-rapids/pull/1883/files#diff-8926565c5f6bc26203bed459ac37ea0fba15f9f5488aa2412aeac26b31f80ea6R430)

razajafri: Correct me if I am wrong but won't they both result in the same code? 

https://github.com/NVIDIA/spark-rapids/blob/236caa07d353f5455edf97ca158e87920f988123/sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala#L180

https://github.com/NVIDIA/spark-rapids/blob/236caa07d353f5455edf97ca158e87920f988123/sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala#L188

because in your case you are basically passing the `nesting.initialTypes = initialTypes` so how is it different?

integration_tests/src/main/python/repart_test.py
```diff
@@ -15,15 +15,35 @@
 import pytest
 
 from asserts import assert_gpu_and_cpu_are_equal_collect
+from spark_session import is_before_spark_311
 from data_gen import *
 from marks import ignore_order
 
-@pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
+nested_scalar_mark=pytest.mark.xfail(reason="https://github.com/NVIDIA/spark-rapids/issues/1459")
```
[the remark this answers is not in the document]
razajafri: It wasn't to save the line length. I did it primarily to avoid repeating the pytest.mark every time when it was exactly the same. If you want I can change this but I think its better this way.

razajafri:
CI failed due to lack of resources

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
gerashegalov: The difference is whether Struct spills into `nested` types , you can observe the difference in the REPL:
```scala
scala> val pluginSupportedOrderableSig = (
     |     TypeSig.commonCudfTypes +
     |     TypeSig.NULL +
     |     TypeSig.DECIMAL +
     |     TypeSig.STRUCT.nested(
     |       TypeSig.commonCudfTypes +
     |       TypeSig.NULL +
     |       TypeSig.DECIMAL
     |     ))
pluginSupportedOrderableSig: com.nvidia.spark.rapids.TypeSig = com.nvidia.spark.rapids.TypeSig@468f547e

scala> val typeSigBlanketNested = (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.STRUCT).nested()

scala> val singleNested = new StructType().add("value", IntegerType)
singleNested: org.apache.spark.sql.types.StructType = StructType(StructField(value,IntegerType,true))

scala> val doubleNested = new StructType().add("nestedInt", new StructType().add("value", IntegerType))
doubleNested: org.apache.spark.sql.types.StructType = StructType(StructField(nestedInt,StructType(StructField(value,IntegerType,true)),true))

scala> pluginSupportedOrderableSig.isSupportedByPlugin(singleNested, true)
res8: Boolean = true

scala> pluginSupportedOrderableSig.isSupportedByPlugin(doubleNested, true)
res7: Boolean = false
```

But with typeSigBlanketNested e.g double nesting would be allowed
```scala
scala> typeSigBlanketNested.isSupportedByPlugin(doubleNested, true)
res11: Boolean = true
````

Comment (razajafri) [reply sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala]: Oh of course! I missed a very important thing that the struct will be a part of the initialTypes. Thanks for the explanation

I have pushed an update. PTAL

**Comment by razajafri** [comment]
build

**Comment by razajafri** [comment]
@gerashegalov can you please see if I have addressed all your concerns?

**Review by gerashegalov — APPROVED** [review]
LGTM

tried some wording on the doc but not sure it's accurate

**Comment by razajafri** [comment]
@gerashegalov your approval was clobbered due to the updated docs. Please approve again

**Review by gerashegalov — APPROVED** [review]

**Comment by sameerz** [comment]
build

**Comment by razajafri** [comment]
CI build fails because a line exceeds 100 chars but the line shows up as within 100 chars in the PR and my local build 

https://github.com/NVIDIA/spark-rapids/blob/4684a32872f09df0a024ca0b0f60204e7f0db8cd/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala#L2693

**Comment by razajafri** [comment]
Figured it out, @gerashegalov made a suggestion that I accepted that made the line exceed 100 chars. I understand that he made the suggestion in the browser and its not his fault. 

This is why I tend to not accept suggestions directly from the browser and usually make the change manually. @gerashegalov can you please bless this PR one more time after the build is successful?

**Comment by razajafri** [comment]
build

**Comment by razajafri** [comment]
build

**Comment by razajafri** [comment]
I had to force push to get rid of the error. I hope this will pass tests

**Comment by razajafri** [comment]
build

**Comment by razajafri** [comment]
build

File: integration_tests/src/main/python/sort_test.py [hunk 4]
```diff
@@ -19,7 +19,7 @@
 from marks import *
 from pyspark.sql.types import *
 import pyspark.sql.functions as f
-from spark_session import is_before_spark_310
+from spark_session import is_before_spark_311
```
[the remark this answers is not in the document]
razajafri: Do you want me to revert these changes? I thought to change it because 310 wasn't released

razajafri:
build

razajafri:
@revans2 can you please take another look?

integration_tests/src/main/python/sort_test.py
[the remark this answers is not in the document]
revans2: No need to revert it. Just would have been nice to have it called out in the PR so I know that it is in there instead of scratching my head as to why there are changes in here that don't look related.

revans2 (approved):

PR #1926, opened 2021-03-12
Consolidate RAT settings in parent pom

gerashegalov:
- Create a single config for RAT in parent.
- Update the RAT exclusion list to filter out hidden directories


gerashegalov:
build

PR #1930, opened 2021-03-12
Relax cudf version check for patch-level versions

### LINKED ISSUE #1921: [FEA] cudf version check should be lenient on later patch version
Occasionally there are hotfix releases for cudf, and currently the cudf version check must be exact.  This means if a cudf patch version is released that contains a critical fix the plugin will refuse to work with the new cudf version.

The cudf version check should allow patch versions of the same cudf minor version that are later than the version it expects.

jlowe:
Fixes #1921.

This updates the cudf version check to be lenient on patch-level differences.  The patch level, and any sub-version afterwards, can be different as long as it is more recent (i.e.: greater) than the requested version.


jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -242,6 +242,35 @@ class RapidsExecutorPlugin extends ExecutorPlugin with Logging {
...
+   * patch version then the actual patch version must be greater than or equal.
+   * For example, version 7.1 is not satisfied by version 7.2, but version 7.1 is satisfied by
+   * version 7.1.1.
+   */
+  def cudfVersionSatisfied(expected: String, actual: String): Boolean = {
+    val expSplits = expected.split('.')
+    val actSplits = actual.split('.')
+    val zipSplits = expSplits.zip(actSplits)
+    // major and minor versions must match exactly
+    if (zipSplits.take(2).exists{case (e, a) => e != a}) {
```
tgravescs, line 258: nit space before and after the {}
{ case ... a }

jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -242,6 +242,35 @@ class RapidsExecutorPlugin extends ExecutorPlugin with Logging {
...
+object RapidsExecutorPlugin {
+  /**
+   * Return true if the expected cudf version is satisfied by the actual version found.
+   * The version is satisfied if the major and minor versions match exactly. If there is a requested
+   * patch version then the actual patch version must be greater than or equal.
+   * For example, version 7.1 is not satisfied by version 7.2, but version 7.1 is satisfied by
+   * version 7.1.1.
+   */
+  def cudfVersionSatisfied(expected: String, actual: String): Boolean = {
+    val expSplits = expected.split('.')
```
gerashegalov, line 254: nit:
```
val (expMajorMinor, expPatch) = expected.split('.').splitAt(2)
val (actMajorMinor, actPatch) = actual.split('.').splitAt(2)
if (expMajorMinor != actMajorMinor) {
   false
} else {
   ...
}
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -242,6 +242,35 @@ class RapidsExecutorPlugin extends ExecutorPlugin with Logging {
...
+    val actSplits = actual.split('.')
+    val zipSplits = expSplits.zip(actSplits)
+    // major and minor versions must match exactly
+    if (zipSplits.take(2).exists{ case (e, a) => e != a }) {
+      return false
+    }
+
+    // patch and sub-patch versions can match as long as actual is more recent
+    zipSplits.drop(2).forall { case (expStr, actStr) =>
+      val expVal = Integer.parseInt(expStr)
```
gerashegalov, line 264: nit: toInt

gerashegalov (approved):
LGTM

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
jlowe: This doesn't work as-is, but I get the idea.  I can do something similar in spirit.

tgravescs (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
gerashegalov: this seems to work
```scala
    val (expMajorMinor, expPatch) = expected.split('.').splitAt(2)
    val (actMajorMinor, actPatch) = actual.split('.').splitAt(2)
    actMajorMinor.startsWith(expMajorMinor) && {
      // if NaN make expected high, actual low so
      // that act is never greater than exp
      val expPatchInts = expPatch.map(v => Try(v.toInt).getOrElse(Int.MaxValue))
      val actPatchInts = actPatch.map(v => Try(v.toInt).getOrElse(Int.MinValue))
      val zipped = expPatchInts.zipAll(actPatchInts, 0, 0)
      zipped.forall { case (e, a) => e <= a }
    }
```

also takes care of the cases where expected patch is longer than the actual one:

```scala
    assert(!RapidsExecutorPlugin.cudfVersionSatisfied("7.0.1", "7.0"))
    assert(!RapidsExecutorPlugin.cudfVersionSatisfied("7.0.1.1", "7.0.1"))
    assert(!RapidsExecutorPlugin.cudfVersionSatisfied("7.0.1-special.1", "7.0.2"))
    assert(!RapidsExecutorPlugin.cudfVersionSatisfied("7.0.2.2.2", "7.0.2.2"))
```

jlowe: Thanks!  That looks cleaner. I updated to use something very similar, except I left the expected as a raw `toInt` since I'd rather know up-front that we have an unexpected string in the expected version rather than a more cryptic cudf versions don't match when they actually do.

jlowe:
build

gerashegalov (approved):
LGTM

jlowe:
@pxLi can you look into the signoff check failure that also succeeded?  :think:  The successful run should have cleared the older error but somehow did not.

jlowe:
build

jlowe:
build

jlowe:
build

pxLi:
> @pxLi can you look into the signoff check failure that also succeeded? 🤔 The successful run should have cleared the older error but somehow did not.

I see it passed now, there could be some delay for the github action to updated the status

pxLi:
build

pxLi:
CI failed. should be waiting for https://github.com/NVIDIA/spark-rapids/pull/1889

jlowe:
> I see it passed now, there could be some delay for the github action to updated the status

It passed because I pushed a new merge commit which caused it to run again.

pxLi:
> > I see it passed now, there could be some delay for the github action to updated the status
> 
> It passed because I pushed a new merge commit which caused it to run again.

yep, I checked the github status, there were 2 terrible incidents for github action recently, 
https://www.githubstatus.com/incidents/s654n76c1bwr
https://www.githubstatus.com/incidents/wdhkj70kynwk

jlowe:
build

PR #1932, opened 2021-03-14
Create a getting started on K8s page

viadea:


docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+   
+   So please look into the Dockerfile.cuda to make sure the file names are correct, and modify if needed.
+   
+   Currently the directory in the local machine should look as below:
+   ```shell 
+   $ ls
+   Dockerfile.cuda   cudf-<version>.jar   getGpusResources.sh   rapids-4-spark_<version>.jar   spark
+   ```
+
+4. Build the Docker image with proper repository name and tag and push to repository
```
jlowe, line 57:
```suggestion
4. Build the Docker image with a proper repository name and tag and push it to the repository
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+   ```shell 
+   export IMAGE_NAME=xxx/yyy:tag
+   docker build . -f Dockerfile.cuda -t $IMAGE_NAME
+   docker push $IMAGE_NAME
+   ```
+
+## Run Spark Applications in Kubernetes Cluster
```
jlowe, line 64:
```suggestion
## Running Spark Applications in the Kubernetes Cluster
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+
+### spark-submit example pi job in `cluster` mode.
+
+This is a good way to test if RAPIDS plugin can be found.
+Because `ClassNotFoundException` is a common error if the Driver can not 
+find the jar for `com.nvidia.spark.SQLPlugin`:
```
jlowe, line 70:
```suggestion
### Submitting a Simple Test Job

This simple job will test if the RAPIDS plugin can be found.
`ClassNotFoundException` is a common error if the Spark driver can not 
find the RAPIDS Accelerator jar, resulting in an exception like this:
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+export K8SMASTER=k8s://https://<k8s-apiserver-host>:<k8s-apiserver-port>
+export SPARK_NAMESPACE=default
+export SPARK_DRIVER_NAME=exampledriver
+
+$SPARK_HOME/bin/spark-submit \
+     --master $K8SMASTER \
+     --deploy-mode cluster  \
+     --name examplejob \
+     --class org.apache.spark.examples.SparkPi \
+     --conf spark.rapids.sql.concurrentGpuTasks=1 \
```
jlowe, line 89: This is the default.
```suggestion
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+     --name mysparkshell \
+     --deploy-mode client  \
+     --conf spark.rapids.sql.concurrentGpuTasks=1 \
+     --conf spark.executor.instances=1 \
+     --conf spark.executor.resource.gpu.amount=1 \
+     --conf spark.executor.memory=4G \
+     --conf spark.executor.cores=1 \
+     --conf spark.task.cpus=1 \
+     --conf spark.task.resource.gpu.amount=1 \
+     --conf spark.rapids.memory.pinnedPool.size=2G \
```
jlowe, line 157: If we're going to ask for 2GB of pinned memory outside of the executor heap then I think we need to set `spark.kubernetes.memoryOverheadFactor` accordingly so Kubernetes expects the extra memory usage.

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+     --conf spark.executor.resource.gpu.vendor=nvidia.com \
+     --conf spark.kubernetes.container.image=$IMAGE_NAME \
+     --conf spark.executor.extraClassPath=/opt/sparkRapidsPlugin/rapids-4-spark_<version>.jar:/opt/sparkRapidsPlugin/cudf-<version>.jar   \
+     --conf spark.driver.extraClassPath=/opt/sparkRapidsPlugin/rapids-4-spark_<version>.jar:/opt/sparkRapidsPlugin/cudf-<version>.jar   \
+     --driver-memory 2G \
+     local:///opt/spark/examples/jars/spark-examples_2.12-3.0.2.jar
+```
+
+   Note: `local://` means the jar file location is inside the Docker image.
+   Since this is `cluster` mode, the Driver is running inside a POD in Kubernetes.
```
jlowe, line 113:
```suggestion
   Since this is `cluster` mode, the Spark driver is running inside a pod in Kubernetes.
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+   When the job is running, we could see both Driver POD and Executor POD(s):
```
jlowe, line 114:
```suggestion
   The driver and executor pods can be seen when the job is running:
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+```shell
+$ kubectl get pods
+NAME                               READY   STATUS    RESTARTS   AGE
+spark-pi-d11075782f399fd7-exec-1   1/1     Running   0          9s
+exampledriver                      1/1     Running   0          15s
+```
+
+   To view the Driver log, use below command:
```
jlowe, line 122:
```suggestion
   To view the Spark driver log, use below command:
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+```shell
+kubectl logs $SPARK_DRIVER_NAME
+```
+
+   To view the Spark Driver UI when the job is running:
```
jlowe, line 127:
```suggestion
   To view the Spark driver UI when the job is running first expose the driver UI port:
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+```shell
+kubectl port-forward $SPARK_DRIVER_NAME 4040:4040
+```
+   Then open web browser:
```
jlowe, line 131:
```suggestion
   Then open a web browser to the Spark driver UI page on the exposed port:
```

tgravescs:
I assume this is to replace https://nvidia.github.io/spark-rapids/docs/get-started/getting-started-on-prem.html#running-on-kubernetes ?  so we probably need to remove that section.

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+```shell
+http://localhost:4040
+```
+
+   To kill the Spark job and delete Driver POD:
```
jlowe, line 136:
```suggestion
   To kill the Spark job and delete driver pod:
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+```shell
+$SPARK_HOME/bin/spark-submit --kill spark:$SPARK_DRIVER_NAME
+kubectl delete pod $SPARK_DRIVER_NAME
+```
+
+### spark-shell in `client` mode with an interactive shell
```
jlowe, line 142:
```suggestion
### Running an Interactive Spark Shell
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+     --conf spark.locality.wait=0s \
+     --conf spark.sql.files.maxPartitionBytes=512m \
+     --conf spark.sql.shuffle.partitions=10 \
+     --conf spark.plugins=com.nvidia.spark.SQLPlugin \
+     --conf spark.kubernetes.namespace=$SPARK_NAMESPACE  \
+     --conf spark.kubernetes.driver.pod.name=$SPARK_DRIVER_NAME  \
+     --conf spark.executor.resource.gpu.discoveryScript=/opt/sparkRapidsPlugin/getGpusResources.sh \
+     --conf spark.executor.resource.gpu.vendor=nvidia.com \
+     --conf spark.kubernetes.container.image=$IMAGE_NAME \
+     --conf spark.executor.extraClassPath=/opt/sparkRapidsPlugin/rapids-4-spark_<version>.jar:/opt/sparkRapidsPlugin/cudf-<version>.jar   \
```
tgravescs, line 106: extraClassPath shouldn't be needed with kubernetes

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+
+   To kill the Spark job and delete Driver POD:
+```shell
+$SPARK_HOME/bin/spark-submit --kill spark:$SPARK_DRIVER_NAME
+kubectl delete pod $SPARK_DRIVER_NAME
+```
+
+### spark-shell in `client` mode with an interactive shell
+
+If you need an interactive spark-shell with Executor POD(s) running inside Kubernetes:
```
jlowe, line 144:
```suggestion
If you need an interactive Spark shell with executor pods running inside the Kubernetes cluster:
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+```shell
+$SPARK_HOME/bin/spark-shell \
+     --master $K8SMASTER \
+     --name mysparkshell \
+     --deploy-mode client  \
+     --conf spark.rapids.sql.concurrentGpuTasks=1 \
```
jlowe, line 150:
```suggestion
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+     --name examplejob \
+     --class org.apache.spark.examples.SparkPi \
+     --conf spark.rapids.sql.concurrentGpuTasks=1 \
+     --conf spark.executor.instances=1 \
+     --conf spark.executor.resource.gpu.amount=1 \
+     --conf spark.executor.memory=4G \
+     --conf spark.executor.cores=1 \
+     --conf spark.task.cpus=1 \
+     --conf spark.task.resource.gpu.amount=1 \
+     --conf spark.rapids.memory.pinnedPool.size=2G \
```
jlowe, line 96: If we're going to ask for 2GB of pinned memory outside of the executor heap then I think we need to set `spark.kubernetes.memoryOverheadFactor` accordingly so Kubernetes expects the extra memory usage.

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+   To view the Driver log, use below command:
+```shell
+kubectl logs $SPARK_DRIVER_NAME
+```
+
+   To view the Spark Driver UI when the job is running:
+```shell
+kubectl port-forward $SPARK_DRIVER_NAME 4040:4040
```
tgravescs, line 129: would this be cluster specific?

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+     --conf spark.kubernetes.namespace=$SPARK_NAMESPACE  \
+     --conf spark.executor.resource.gpu.discoveryScript=/opt/sparkRapidsPlugin/getGpusResources.sh \
+     --conf spark.executor.resource.gpu.vendor=nvidia.com \
+     --conf spark.kubernetes.container.image=$IMAGE_NAME \
+     --conf spark.executor.extraClassPath=/opt/sparkRapidsPlugin/rapids-4-spark_<version>.jar:/opt/sparkRapidsPlugin/cudf-<version>.jar   \
+     --driver-class-path=./cudf-<version>.jar:./rapids-4-spark_<version>.jar \
+     --driver-memory 2G 
+```
+
+Only `client` mode is applicable. If you specify this as `cluster` mode, you would see below error:
```
jlowe, line 171:
```suggestion
Only the `client` deploy mode should be used. If you specify the `cluster` deploy mode, you would see the following error:
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+```
+   Then open web browser:
+```shell
+http://localhost:4040
+```
+
+   To kill the Spark job and delete Driver POD:
+```shell
+$SPARK_HOME/bin/spark-submit --kill spark:$SPARK_DRIVER_NAME
+kubectl delete pod $SPARK_DRIVER_NAME
```
tgravescs, line 139: why do we need to delete it?

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+     --conf spark.rapids.memory.pinnedPool.size=2G \
+     --conf spark.locality.wait=0s \
+     --conf spark.sql.files.maxPartitionBytes=512m \
+     --conf spark.sql.shuffle.partitions=10 \
+     --conf spark.plugins=com.nvidia.spark.SQLPlugin \
+     --conf spark.kubernetes.namespace=$SPARK_NAMESPACE  \
+     --conf spark.executor.resource.gpu.discoveryScript=/opt/sparkRapidsPlugin/getGpusResources.sh \
+     --conf spark.executor.resource.gpu.vendor=nvidia.com \
+     --conf spark.kubernetes.container.image=$IMAGE_NAME \
+     --conf spark.executor.extraClassPath=/opt/sparkRapidsPlugin/rapids-4-spark_<version>.jar:/opt/sparkRapidsPlugin/cudf-<version>.jar   \
```
tgravescs, line 166: spark.executor.extraClassPath shouldn't be needed

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+     --driver-class-path=./cudf-<version>.jar:./rapids-4-spark_<version>.jar \
+     --driver-memory 2G 
+```
+
+Only `client` mode is applicable. If you specify this as `cluster` mode, you would see below error:
+```shell
+Cluster deploy mode is not applicable to Spark shells.
+```
+And also you notice that here we removed `--conf spark.driver.extraClassPath` but added `--driver-class-path`.
+This is because now the Driver is running on the client machine so the jar paths should be local machine path.
```
jlowe, line 176:
```suggestion
Also notice that `--conf spark.driver.extraClassPath` was removed but `--driver-class-path` was added.
This is because now the driver is running on the client machine, so the jar paths should be local filesystem paths.
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+
+Only the Executor POD(s) are running inside Kubernetes:
```
jlowe, line 178:
```suggestion
When running the shell you can see only the executor pods are running inside Kubernetes:
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+```
+$ kubectl get pods
+NAME                                     READY   STATUS    RESTARTS   AGE
+mysparkshell-bfe52e782f44841c-exec-1     1/1     Running   0          11s
+```
+
+We can run below scala code in `spark-shell` to test if GPU is enabled for our queries:
```
jlowe, line 185:
```suggestion
The following Scala code can be run in the Spark shell to test if the RAPIDS Accelerator is enabled.
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+```shell
+val df = spark.sparkContext.parallelize(Seq(1)).toDF()
+df.createOrReplaceTempView("df")
+spark.sql("SELECT value FROM df WHERE value <>1").show
+spark.sql("SELECT value FROM df WHERE value <>1").explain
+:quit
+```
+The expected explain plan should contain the GPU related operators:
```
jlowe, line 193:
```suggestion
The expected `explain` plan should contain the GPU related operators:
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+scala> spark.sql("SELECT value FROM df WHERE value <>1").explain
+== Physical Plan ==
+GpuColumnarToRow false
++- GpuFilter NOT (value#2 = 1)
+   +- GpuRowToColumnar TargetSize(2147483647)
+      +- *(1) SerializeFromObject [input[0, int, false] AS value#2]
+         +- Scan[obj#1]
+```
+
+### spark-submit python code in `client` mode.
```
jlowe, line 204:
```suggestion
### Running PySpark in Client Mode
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+
+Of course, you can `COPY` the python code in the Docker image when building it.
+Then you can spark-submit it using `cluster` mode as previous example pi job.
```
jlowe, line 207:
```suggestion
Of course, you can `COPY` the Python code in the Docker image when building it
and submit it using the `cluster` deploy mode as showin in the previous example pi job.
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+
+However if you do not want to re-build the Docker image each time and just want to submit the python code
```
jlowe, line 209:
```suggestion
However if you do not want to re-build the Docker image each time and just want to submit the Python code
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+from the client machine, you can use `client` mode.
```
jlowe, line 210:
```suggestion
from the client machine, you can use the `client` deploy mode.
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+
+```shell
+$SPARK_HOME/bin/spark-submit \
+     --master $K8SMASTER \
+     --deploy-mode client  \
+     --name mypythonjob \
+     --conf spark.rapids.sql.concurrentGpuTasks=1 \
```
jlowe, line 217:
```suggestion
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+     --conf spark.executor.instances=1 \
+     --conf spark.executor.resource.gpu.amount=1 \
+     --conf spark.executor.memory=4G \
+     --conf spark.executor.cores=1 \
+     --conf spark.task.cpus=1 \
+     --conf spark.task.resource.gpu.amount=1 \
+     --conf spark.rapids.memory.pinnedPool.size=2G \
```
jlowe, line 224: Need to set `spark.kubernetes.memoryOverheadFactor` as well.

docs/get-started/getting-started-kubernetes.md
viadea: @tgravescs This depends on: When building the container image, do we put the 2 jars into "jars" subdirectory of spark or not in the Docker container.  If yes, then no need, because those 2 jars are already there. However if we just copy it into /opt/sparkRapidsPlugin in Docker container, then we need to specify it here. This is to give a better clear picture for customers imo.

docs/get-started/getting-started-kubernetes.md
viadea: This is not. "4040" is the default port for driver and is also used by Spark doc: https://spark.apache.org/docs/latest/configuration.html

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
...
+   $ ls
+   Dockerfile.cuda   cudf-<version>.jar   getGpusResources.sh   rapids-4-spark_<version>.jar   spark
+   ```
+
+4. Build the Docker image with proper repository name and tag and push to repository
+   ```shell 
+   export IMAGE_NAME=xxx/yyy:tag
+   docker build . -f Dockerfile.cuda -t $IMAGE_NAME
+   docker push $IMAGE_NAME
+   ```
```
hyperbolic2346, line 62: Do we really require the user build a docker image to run this in Kubernetes? I would expect an automated build in docker hub and this doc to distill down to `kubectl apply -f https://github.com/NVIDIA/rapids/examples/k8s/deply.yaml` or, is complex, `helm install ...` and an example workload.

docs/get-started/getting-started-kubernetes.md
viadea: Good question. Several reasons: 1. The driver pod may not be cleaned in time after the job completes,  so if we may need to manually clean it to run the test again. 2. Due to some reasons the driver may fail(such as noclassfound exception), then we may need to clean it manually.

docs/get-started/getting-started-kubernetes.md
viadea: As mentioned in previous comment we need to set the 2 parameters if the 2 jars are not in the jars directory of spark in Docker.

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,256 @@
+---
+layout: page
+title: Kubernetes
+nav_order: 1
```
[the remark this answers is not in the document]
viadea: Fixed

docs/get-started/getting-started-kubernetes.md
tgravescs: what is multiple drivers on the same host?  Spark increments the port number for you.   Do some clusters just have aces open so I can just access it. etc... this is what I mean by cluster specific. I'm fine with this as long as we document.

docs/get-started/getting-started-kubernetes.md
viadea: @hyperbolic2346  this could one goal. Currently this guide is to give users a way to customize the docker image if they want.

docs/get-started/getting-started-kubernetes.md
viadea: Oh i see, i think configurations like spark.driver.port should not need to be documented by us and users can refer to spark guide for all configs. this may just be a quick start guide. How do you think?

docs/get-started/getting-started-kubernetes.md
viadea: @jlowe Do you mean the default spark.kubernetes.memoryOverheadFactor=0.1 is not enough? What is your suggestion on the value of spark.kubernetes.memoryOverheadFactor is we pin 2G memory in this case?

viadea:
> I assume this is to replace https://nvidia.github.io/spark-rapids/docs/get-started/getting-started-on-prem.html#running-on-kubernetes ? so we probably need to remove that section.

Yes I think once this guide is released, we can either remove that section or create a link to this doc from that doc.

docs/get-started/getting-started-kubernetes.md
```diff
@@ -0,0 +1,254 @@
...
+1. [Download Apache Spark](https://spark.apache.org/downloads.html).
+   Supported versions of Spark are listed on the [RAPIDS Accelerator download page](../download.md).  Please note that only
+   Scala version 2.12 is currently supported by the accelerator. 
+
+   Note that you can download these into a local directory and untar the Spark `.tar.gz` as a directory named `spark`.
+
+2. Download the [RAPIDS Accelerator for Spark jars](getting-started-on-prem.md#download-the-rapids-jars) and the
+  [GPU discovery script](getting-started-on-prem.md#install-the-gpu-discovery-script).
+  
+   Put the 2 jars -- `rapids-4-spark_<version>.jar`, `cudf-<version>.jar`  and `getGpusResources.sh` in the same directory as `spark`.
```
tgravescs, line 41: we should clarify this and say in the jars directory within the spark download

docs/get-started/getting-started-kubernetes.md
tgravescs: I yes there are multiple ways to do this, this doc is putting them in /jars do not needed. I'm fine with adding something else saying if you don't put them in /jars then you may need to include them somehow, or we document all the known ways of doing this.  For instance you could also use --jars with spark -submit if you have setup shared storage space. I honestly don't know if you do this if they need to be included in extra class path.

docs/get-started/getting-started-kubernetes.md
jlowe: No, 0.1 is not going to be enough even in this trivial example because 4G heap * 0.1 overhead is only ~400M of overhead, and that's not enough to cover the 2G of pinned memory (along with regular JVM overhead).  In this case it would need to be set to something like 0.6 with 0.5 to cover the 2G of pinned memory and the remaining 0.1 to cover the JVM itself as for normal Spark-on-k8s JVM jobs.

docs/get-started/getting-started-kubernetes.md
tgravescs: this will remove the logs for that job as well though won't it?

docs/get-started/getting-started-kubernetes.md
viadea: If using --jars, then no need since those jars after --jars will be sent to each executors.

docs/get-started/getting-started-kubernetes.md
viadea: Correct. maybe i can say "To delete the Driver POD:" section instead of putting it under "kill job" section

viadea: Just added  "To delete the Driver POD" section. How do you think right now?

docs/get-started/getting-started-kubernetes.md
tgravescs: so it will depends on what spark does with them and who downloads them. If the jars aren't downloaded before starting the executor and put in class path then we still may need the extraClassPath options

tgravescs: I haven't ever tried this or looked at the k8s code for this case.

docs/get-started/getting-started-kubernetes.md
tgravescs: yeah that's fine

docs/get-started/getting-started-kubernetes.md
tgravescs: yeah quick start is fine perhaps at the beginning of the doc we can add a sentence ssaying this is quick start thing can be slightly different for different cluster setups, see your spark configs.

viadea: "This is a quick start guide which uses default settings which may be different from your cluster." added.

docs/get-started/getting-started-kubernetes.md
viadea: May need to use spark.kubernetes.file.upload.path to point to a shared storage.

docs/get-started/getting-started-kubernetes.md
viadea: @tgravescs Actually this means the spark directory + 2 jars + getGpusResources.sh are in the same directory.

tgravescs: oh sorry I see.   the docker file in next step will move them.

docs/get-started/getting-started-kubernetes.md
viadea: @jlowe i just added spark.kubernetes.memoryOverheadFactor=0.6 can you take a look?thx 
I tested this in example job and it works.

docs/get-started/getting-started-kubernetes.md
tgravescs: ming this is going by these instructions, the jars are in the spark/jars directory so we should remove them from this sample command and add in a separately section if you want to states other ways to do it.

docs/get-started/getting-started-kubernetes.md
tgravescs: personally I don't like recommending changing the overhead factor as it changes with the amount of memory you specify, I would prefer to see us use 

spark.executor.memoryOverhead

which is not k8s specific so it applies to multiple resource managers and you set it at a value. If we know pinned memory is going to be 2G, set it to 3G for instance. This lets you change the executor memory size without having to worry about changing this one as well.

jlowe: Yes, +1.  I should have pointed to spark.executor.memoryOverhead.

viadea: Yep just Changed to spark.executor.memoryOverhead=3G

docs/get-started/getting-started-kubernetes.md
viadea: I just added some words to explain that:
`Note: If here you decide to put above 2 jars in the spark/jars directory which will be copied into /opt/spark/jars directory in Docker image, then in the future you do not need to specify spark.driver.extraClassPath or spark.executor.extraClassPath using cluster mode. This example just shows you a way to put customized jars or 3rd party jars.`

See if it looks good to you. Feel free to reword if needed.

docs/get-started/getting-started-kubernetes.md
viadea: Added spark.executor.memoryOverhead=3G in each spark-submit.

docs/get-started/getting-started-kubernetes.md
viadea: resolved manually

jlowe (approved):

tgravescs:
ok I think that doc is fine but how is this fitting in with existing docs:
https://nvidia.github.io/spark-rapids/docs/get-started/getting-started-on-prem.html#running-on-kubernetes

why don't we just remove it here?  or were you going to do that in a followup - if followup needs to be done right away so we don't forget about it.

tgravescs:
build

btong04:
I was able to go through the documentation and create a k8s cluster, but it looks like the GPU operator plugin isn't working properly. I tried both the deepops deployment several times (on a desktop) and the alternative option of using kubeadmin+helm (on top of a new Ubuntu 18.04 install on AWS EC2). I had trouble getting ansible to authenticate on EC2 so I went ahead with kubeadmin+helm.

Here are similar bug reports to what I encountered with the GPU operator plugin: https://github.com/NVIDIA/gpu-operator/issues/83, https://github.com/NVIDIA/gpu-operator/issues/166. Spark-shell can be started, but it complains about no resources being available when running GPU enabled tasks. Spark worker web UI on port 4040 confirms that no resources were assigned. Are there alternatives to using GPU-Operator?

I was able to confirm that my dockerfile works with CPU only spark run on k8s prior to adding the GPU operator plugin.

viadea:
> I was able to go through the documentation and create a k8s cluster, but it looks like the GPU operator plugin isn't working properly. I tried both the deepops deployment several times (on a desktop) and the alternative option of using kubeadmin+helm (on top of a new Ubuntu 18.04 install on AWS EC2). I had trouble getting ansible to authenticate on EC2 so I went ahead with kubeadmin+helm.
> 
> Here are similar bug reports to what I encountered with the GPU operator plugin: [NVIDIA/gpu-operator#83](https://github.com/NVIDIA/gpu-operator/issues/83), [NVIDIA/gpu-operator#166](https://github.com/NVIDIA/gpu-operator/issues/166). Spark-shell can be started, but it complains about no resources being available when running GPU enabled tasks. Spark worker web UI on port 4040 confirms that no resources were assigned. Are there alternatives to using GPU-Operator?
> 
> I was able to confirm that my dockerfile works with CPU only spark run on k8s prior to adding the GPU operator plugin.

As discussed, how to install a working K8s cluster with nvidia GPU support is not in this scope of this article. 
You should follow https://docs.nvidia.com/datacenter/cloud-native/kubernetes/install-k8s.html#option-2-installing-kubernetes-using-kubeadm to install a working k8s cluster.

Regarding the error message such as `User "system:serviceaccount:default:default" cannot get resource "pods" in API group "" in the namespace "default".` , you need to make sure you submit the spark job using a "service account" with proper ClusterRole binded.  For example, in K8s on Spark Doc https://spark.apache.org/docs/latest/running-on-kubernetes.html#submitting-applications-to-kubernetes , it shows how to create a service account named "spark" with proper ClusterRole binded.

Of course, you can refer to k8s doc for more k8s cluster management tasks such as how to create service account, how to create ClusterRoleBinding, how to assign Resource Quotas etc.

viadea:
@btong04 

Today I also tried to use [DeepOps](https://github.com/NVIDIA/deepops/tree/master/docs/k8s-cluster) to deploy a K8s cluster on a single EC2 machine, and it works fine for me as well.
Since [DeepOps](https://github.com/NVIDIA/deepops/tree/master/docs/k8s-cluster) used [kubespray](https://github.com/kubernetes-sigs/kubespray) as the K8s template, it may brings more complexity in terms of K8s troubleshooting.

For example, when I was testing installing K8s using DeepOPS, assume we use all default settings,
I also met several K8s related issues, here are several common K8s related issues and resolution:
1. There are 2 CoreDNS PODs with 1 POD pending.
This is by default, deployment dns-autoscaler is created in K8s cluster which requires at least 2 CoreDNS.
You can reduce it to 1 from 2 with below command:
`kubectl edit configmap dns-autoscaler --namespace=kube-system`
in above command,change `"min":2` to `"min":1`, and then delete the pending POD if it is still there or it started crashing.

2. CoreDNS pod crashed with the reason as "OOMKilled"
This is because by default, CoreDNS POD has 170MB memory limit which may be too small for big cluster. 
The fix is straightforward, just increase the deployment CoreDNS' resource limit:
`kubectl set resources deployment.v1.apps/coredns --limits=cpu=1000m,memory=1024Mi`

3. Spark on Kubernetes Job in client mode keeps failing
This may be related to the Spark job but it is nothing to do with the rapids accelerator here.
Any Spark job in client may fail if you are installing a K8s cluster on a ec2 instance using deepOps method.
The symptom is:
The Spark Driver may keep printing below messages:
`Initial job has not accepted any resources; check your cluster UI to ensure that workers are registered and have sufficient resources.`
The Spark Executor may keeps crashing and restarting, but if we use "kubectl logs" to check the Executor POD, we will get the root cause:
`Caused by: java.net.UnknownHostException: ip-xxx-xxx-xxx-xxx.cluster.local`

Here the `hostname -f` shows as ".cluster.loca" while the `hostname` shows as `.ec2.internal`.
This is very strange because the correct EC2 hostname should be `.ec2.internal`.
Since this K8s cluster created by Kubespray by default enables nodelocal dns cache with default IP as 169.254.25.10.
So the DNS server inside POD is always pointing to 169.254.25.10.

Eventually the root cause is below entry was added by Ansible in /etc/hosts:
```
# Ansible inventory hosts BEGIN
xxx.xxx.xxx.xxx ip-xxx-xxx-xxx-xxx.cluster.local ip-xxx-xxx-xxx-xxx ip-xxx-xxx-xxx-xxx.ec2.internal.cluster.local ip-xxx-xxx-xxx-xxx.ec2.internal
```
After removing above entries in /etc/hosts,  `hostname -f` become correct. 
Basically we just let DNS server to resolve the hostname.
And then Spark in client job works fine.

I am pretty sure there could be more K8s related issues but here is just my quick summary for using DeepOPS to install a K8s cluster with a single EC2 instance.
Using DeepOps way to install a enterprise ready K8s cluster installs more K8s deployments comparing to a kubeadm method, so if you just want to test the basic functions of Spark+Rapids in K8s without much K8s troubleshooting experience, I would suggest you use kubeadm method.

PR #1933, opened 2021-03-15
Split DB scripts to make them common for the build and IT pipeline

NvTimLiu:
Split DB scripts to make them common for the build and IT pipeline.

Issue: https://github.com/NVIDIA/spark-rapids/issues/1568

Related PR https://github.com/NVIDIA/spark-rapids/pull/1829

### DIFF
--- a/jenkins/databricks/build.sh
+++ b/jenkins/databricks/build.sh
@@ -103,36 +103,5 @@ mvn -B install:install-file \
 
 mvn -B -P${BUILD_PROFILES} clean package -DskipTests
 
-# Copy so we pick up new built jar and latest CuDF jar. Note that the jar names have to be
-# exactly what is in the statically setup Databricks cluster we use.
-echo "Copying rapids jars: dist/target/$RAPIDS_BUILT_JAR udf-examples/target/$RAPIDS_UDF_JAR $DB_JAR_LOC"
-sudo cp dist/target/$RAPIDS_BUILT_JAR udf-examples/target/$RAPIDS_UDF_JAR $DB_JAR_LOC
-echo "Copying cudf jars: $CUDF_JAR $DB_JAR_LOC"
-sudo cp $CUDF_JAR $DB_JAR_LOC
-
-# tests
-export PATH=/databricks/conda/envs/databricks-ml-gpu/bin:/databricks/conda/condabin:$PATH
-sudo /databricks/conda/envs/databricks-ml-gpu/bin/pip install pytest sre_yield requests pandas \
-	pyarrow findspark pytest-xdist
-cd /home/ubuntu/spark-rapids/integration_tests
-export SPARK_HOME=/databricks/spark
-# change to not point at databricks confs so we don't conflict with their settings
-export SPARK_CONF_DIR=$PWD
-export PYTHONPATH=$SPARK_HOME/python:$SPARK_HOME/python/pyspark/:$SPARK_HOME/python/lib/py4j-0.10.9-src.zip
-sudo ln -s /databricks/jars/ $SPARK_HOME/jars || true
-sudo chmod 777 /databricks/data/logs/
-sudo chmod 777 /databricks/data/logs/*
-echo { \"port\":\"15002\" } > ~/.databricks-connect
-if [ `ls $DB_JAR_LOC/rapids* | wc -l` -gt 2 ]; then
-    echo "ERROR: Too many rapids jars in $DB_JAR_LOC"
-    ls $DB_JAR_LOC/rapids*
-    exit 1
-fi
-if [ `ls $DB_JAR_LOC/cudf* | wc -l` -gt 1 ]; then
-    echo "ERROR: Too many cudf jars in $DB_JAR_LOC"
-    ls $DB_JAR_LOC/cudf*
-    exit 1
-fi
-bash run_pyspark_from_build.sh --runtime_env="databricks"
 cd /home/ubuntu
 tar -zcvf spark-rapids-built.tgz spark-rapids
--- a/jenkins/databricks/clusterutils.py
+++ b/jenkins/databricks/clusterutils.py
@@ -23,7 +23,7 @@ class ClusterUtils(object):
 
     @staticmethod
     def generate_create_templ(sshKey, cluster_name, runtime, idle_timeout,
-            num_workers, driver_node_type, worker_node_type, cluster_type,
+            num_workers, driver_node_type, worker_node_type, cloud_provider,
             printLoc=sys.stdout):
         timeStr = str(int(time.time()))
         uniq_name = cluster_name + "-" + timeStr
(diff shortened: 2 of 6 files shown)


NvTimLiu:
build

jenkins/databricks/build.sh
```diff
@@ -103,36 +103,5 @@ mvn -B install:install-file \
 
 mvn -B -P${BUILD_PROFILES} clean package -DskipTests
 
-# Copy so we pick up new built jar and latest CuDF jar. Note that the jar names have to be
```
NvTimLiu, line 106: Move integration tests scripts into 'test.sh', to make it common for DB build & IT pipelines

jenkins/databricks/params.py
```diff
@@ -0,0 +1,71 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
NvTimLiu, line 1: Make the parameter parsing scripts common for both 'run-build.py' and 'run-test.py'

jenkins/databricks/run-tests.py
```diff
@@ -19,84 +19,25 @@
 import os
 import subprocess
 from clusterutils import ClusterUtils
+import params
 
 
 def main():
-  workspace = 'https://dbc-9ff9942e-a9c4.cloud.databricks.com'
```
NvTimLiu, line 25: Make the parameter parsing scripts common for both 'run-build.py' and 'run-test.py'

jenkins/databricks/test.sh
```diff
@@ -0,0 +1,42 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -e
+
+LOCAL_JAR_PATH=$1
+
+# tests
+export PATH=/databricks/conda/envs/databricks-ml-gpu/bin:/databricks/conda/condabin:$PATH
```
NvTimLiu, line 23: Move integration tests scripts into 'test.sh', to make it common for DB build & IT pipelines

jenkins/databricks/test.sh
```diff
@@ -0,0 +1,42 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -e
+
+LOCAL_JAR_PATH=$1
```
NvTimLiu, line 20: Build pipeline builds the jars from source code, so the LOCAL_JAR_PATH is null.

IT pipeline sets the 'LOCAL_JAR_PATH' to tell where the jars are downloaded into the local directory.

jenkins/databricks/clusterutils.py
```diff
@@ -23,21 +23,23 @@ class ClusterUtils(object):
 
     @staticmethod
     def generate_create_templ(sshKey, cluster_name, runtime, idle_timeout,
-            num_workers, driver_node_type, worker_node_type,
+            num_workers, driver_node_type, worker_node_type, cluster_type,
```
NvTimLiu, line 26: Add 'cluster_type' parameter to make create.py common for aws and azure Databricks

jenkins/databricks/create.py
```diff
@@ -33,20 +33,21 @@ def main():
   num_workers = 1
   worker_type = 'g4dn.xlarge'
   driver_type = 'g4dn.xlarge'
+  cluster_type = 'aws'
```
NvTimLiu, line 36: Add 'cluster_type' parameter to make create.py common for aws and azure Databricks

NvTimLiu:
build

jenkins/databricks/clusterutils.py
tgravescs: don't we need whatever equivalent for azure or those just aren't specified?

tgravescs: also perhaps a better name for this would be like cloud_provider.

jenkins/databricks/run-build.py
```diff
@@ -0,0 +1,51 @@
...
+  rsync_command = "rsync -I -Pave \"ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -p 2200 -i %s\" %s ubuntu@%s:%s" % (params.private_key_file, params.source_tgz, master_addr, params.tgz_dest)
+  print("rsync command: %s" % rsync_command)
+  subprocess.check_call(rsync_command, shell = True)
+
+  ssh_command = "ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null ubuntu@%s -p 2200 -i %s %s %s %s %s 2>&1 | tee buildout; if [ `echo ${PIPESTATUS[0]}` -ne 0 ]; then false; else true; fi" % (master_addr, params.private_key_file, params.script_dest, params.tgz_dest, params.base_spark_pom_version, params.build_profiles)
+  print("ssh command: %s" % ssh_command)
+  subprocess.check_call(ssh_command, shell = True)
+
+  print("Copying built tarball back")
+  rsync_command = "rsync  -I -Pave \"ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -p 2200 -i %s\" ubuntu@%s:/home/ubuntu/spark-rapids-built.tgz ./" % (params.private_key_file, master_addr)
```
tgravescs, line 46: do we are now downloading the jar after the build finished before the tests run.  This is fine as long as we make sure if the tests fail this gets thrown away and we don't push it.  Can you please make sure that is true.

tgravescs:
changes overall look fine, a few minor nits. One thing I want to mention is these scripts are also used by dev to build and test on Databricks manually.  So any changes I would like to keep that easy to do.  It looks like with these changes that is still fine, just splits it into build and test separately.

jenkins/databricks/run-build.py
NvTimLiu: Yes, we make sure if the tests fail we don't push jars, and fails the pipeline.

The Databricks Jenkins steps runs orderly: Create cluster --> build jars --> run tests --> deploy jars

The pipeline stops and the follow-up will not run, only if previews step fails

jenkins/databricks/clusterutils.py
NvTimLiu: > don't we need whatever equivalent for azure or those just aren't specified?

I thinks we need not specify `whatever equivalent for azure`, as the default(unspecified) value of `cluster_type` is aws. 

Also `cluster_type` is for aws configs of creating cluster as below, any other values means we do not need the below configs. So we only need to specify the default(unspecified) value `cluster_type` as 'aws' in the create.py.

        if (cluster_type == 'aws'):
            templ['aws_attributes'] = {
                        "zone_id": "us-west-2a",
                        "first_on_demand": 1,
                        "availability": "SPOT_WITH_FALLBACK",
                        "spot_bid_price_percent": 100,
                        "ebs_volume_count": 0
            }

NvTimLiu: > also perhaps a better name for this would be like cloud_provider.

Sounds good, let me change it.

NvTimLiu:
build

tgravescs (approved):

PR #1935, opened 2021-03-15
Update Alias SQL quoting and float-to-timestamp casting to match Spark 3.2

jlowe:
Alias SQL quoting behavior changed in [SPARK-34626](https://issues.apache.org/jira/browse/SPARK-34626) and float-to-timestamp casting behavior changed in [SPARK-34727](https://issues.apache.org/jira/browse/SPARK-34727) to upcast to double first.

This adds shim-specific flags to know when to apply these changed behaviors so the plugin remains compatible with the behavior with older versions of Spark while also matching the new Spark 3.2 behavior.


jlowe:
build

revans2:
I can see why we might want the cast to be configured by the shim, but honestly these are bug fixes, and I'm not really sure why we need to break older versions when we could fix it for them, especially the SQL alias code.

revans2 (approved):
After talking it over, it turns out we have some tests that check both of these, so this is fine.

Issue #1912, opened 2021-03-11
[FEA] Let `Scalar Pandas UDF ` support array of struct type.


firestarman:
**Is your feature request related to a problem? Please describe.**
I wish the RAPIDS Accelerator for Apache Spark would support running Scalar Pandas UDF with array type as input, then it can work with `collect_list`, such as the app code as below.

```Python
@pandas_udf(returnType=IntegerType())
def pandas_udf_func(windows: pd.Series) -> pd.Series:
    return pd.Series([1]).repeat(windows.size)

spark.sql("SELECT pandas_udf_func( collect_list( struct(a, b)) ) AS ret_list FROM a_table")
```

firestarman:
Now the cuDF complains the error as below when transfering data to Python.

```
Caused by: ai.rapids.cudf.CudfException: cuDF failure at: /home/liangcail/work/projects/on_github/cudf/cpp/src/interop/to_arrow.cpp:207: Number of field names and number of children doesn't match

	at ai.rapids.cudf.Table.convertCudfToArrowTable(Native Method)
	at ai.rapids.cudf.Table.access$1500(Table.java:46)
	at ai.rapids.cudf.Table$ArrowIPCTableWriter.write(Table.java:1010)
	at org.apache.spark.sql.rapids.execution.python.GpuArrowPythonRunner$$anon$2.$anonfun$writeIteratorToStream$5(GpuArrowEvalPythonExec.scala:455)
	at org.apache.spark.sql.rapids.execution.python.GpuArrowPythonRunner$$anon$2.$anonfun$writeIteratorToStream$5$adapted(GpuArrowEvalPythonExec.scala:453)
	at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
	at com.nvidia.spark.rapids.Arm.withResource$(Arm.scala:26)
```

firestarman:
Filed an issue https://github.com/rapidsai/cudf/issues/7570 for the exception above.

PR #1944, opened 2021-03-16
Support running scalar pandas UDF with array type.

firestarman:
This PR is to support running scalar pandas UDF with array type.

1. Add array type signature for related expressions and plans.
2. Flatten the names of nested struct columns from schema, which is also required by the cudf Arrow IPC writer.

This PR depends on https://github.com/rapidsai/cudf/pull/7598 

closes https://github.com/NVIDIA/spark-rapids/issues/1912


firestarman:
rapidsai/cudf#7598 was merged. Mark as 'ready for review'.

firestarman:
build

firestarman:
build

firestarman:
build

integration_tests/src/main/python/data_gen.py
```diff
@@ -826,6 +826,19 @@ def gen_scalars_for_sql(data_gen, count, seed=0, force_no_nulls=False):
            decimal_gen_default, decimal_gen_scale_precision, decimal_gen_same_scale_precision,
            decimal_gen_64bit]
 
+common_cudf_types_gen = all_basic_gens[:-1]
```
jlowe, line 829: This is very fragile, especially since it's so far removed from where `all_basic_gens` is defined.  I'd much rather see `all_basic_gens` be based on `common_cudf_types_gen` rather than assuming we know how to slice up `all_basic_gens` accordingly.  The next time someone touches `all_basic_gens` this is almost guaranteed to break.

integration_tests/src/main/python/data_gen.py
```diff
@@ -826,6 +826,19 @@ def gen_scalars_for_sql(data_gen, count, seed=0, force_no_nulls=False):
...
+
+# Pyarrow will complain the error as below if the timestamp is out of range for both CPU and GPU,
+# so narrow down the time range to avoid exceptions causing test failures.
+#
+#     "pyarrow.lib.ArrowInvalid: Casting from timestamp[us, tz=UTC] to timestamp[ns]
+#      would result in out of bounds timestamp: 51496791452587000"
+#
+# This issue has been fixed in pyarrow by the PR https://github.com/apache/arrow/pull/7169
+# However it still requires PySpark to specify the new argument "timestamp_as_object".
+arrow_common_gen = common_cudf_types_gen[:-1] + [TimestampGen(
```
jlowe, line 839: Same slicing problem here.  We should build up from common types rather than assume we can slice out what we need.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1979,7 +1979,7 @@ object GpuOverrides {
         TypeSig.all,
         repeatingParamCheck = Some(RepeatingParamCheck(
           "param",
-          TypeSig.commonCudfTypes,
+          TypeSig.commonCudfTypes + TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.STRUCT),
```
jlowe, line 1982: We're adding array support, but explicitly disallowing array-of-array.  Intentional?  Also this says we support arrays of structs but not top-level structs?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2630,7 +2630,9 @@ object GpuOverrides {
       "The backend of the Scalar Pandas UDFs. Accelerates the data transfer between the" +
         " Java process and the Python process. It also supports scheduling GPU resources" +
         " for the Python process when enabled",
-      ExecChecks(TypeSig.commonCudfTypes, TypeSig.all),
+      ExecChecks(
+        TypeSig.commonCudfTypes + TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.STRUCT),
```
jlowe, line 2634: Same questions about the types supported here.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuArrowEvalPythonExec.scala
```diff
@@ -430,18 +430,46 @@ class GpuArrowPythonRunner(
 
       protected override def writeIteratorToStream(dataOut: DataOutputStream): Unit = {
         val writer = {
+
+          lazy val flattenColumnNames: (StructField, Boolean) => Seq[(String, Boolean)] =
+            (f, appendMe)=> {
```
jlowe, line 435:
```suggestion
            (f, appendMe) => {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuArrowEvalPythonExec.scala
```diff
@@ -430,18 +430,46 @@ class GpuArrowPythonRunner(
...
+              if (appendMe) {
+                colNames.append((f.name, f.nullable))
+              }
+              f.dataType match {
+                case structType: StructType =>
+                  colNames.appendAll(structType.flatMap(field => flattenColumnNames(field, true)))
+                case arrayType: ArrayType =>
+                  // Dig into the child but skip itself by specifying `appendMe` to false
+                  colNames.appendAll(flattenColumnNames(
+                    StructField("", arrayType.elementType, f.nullable), false))
```
jlowe, line 446: Why create an entry at all if we're not going to append it?  Seems like we can skip the `appendMe` stuff and just have a recursive method that only reports what names it has.  I think this can be significantly simplified with the observation that only structs trigger names.  For example:
```
private def flattenNames(d: DataType): Seq[String] = d match {
  case s: StructType => s.flatMap(sf => Seq(sf.name) ++ flattenNames(sf.dataType))
  case a: ArrayType => names(a.elementType)
  case m: MapType => names(m.keyType) ++ names(m.valueType)
  case _ => Nil
}
[...]
val columnNames = flattenNames(pythonInSchema)
```

jlowe (changes requested):

integration_tests/src/main/python/data_gen.py
firestarman: Thanks for reivew. Good catch. Updated.

firestarman:
build

integration_tests/src/main/python/data_gen.py
firestarman: Updated accordingly.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
firestarman: Yes, the original goal is to support only array of struct for a customer request. 

I have updated to support array of struct, array of array, top-level struct, struct of struct, struct of array, along with the integration tests, since cudf has already supported them. 

It would be nice to have them here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
firestarman: Yes, the original goal is to support only array of struct for a customer request.

I have updated to support array of struct, array of array, top-level struct, struct of struct, struct of array, along with the integration tests, since cudf has already supported them.

It would be nice to have them here.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuArrowEvalPythonExec.scala
firestarman: on longer needed

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuArrowEvalPythonExec.scala
firestarman: Good suggestion, i did a little change to support `nullable` flag, which is being used here when specifying the column names.

jlowe (approved):

revans2 (approved):
The changes look good. I would like to see some follow on work in CUDF to rethink how we deal with nested types for writes now that parquet also supports this.

PR #1954, opened 2021-03-17
Fix CBO bug where incompatible plans were produced with AQE on

andygrove:

This fixes a bug where the optimizer would try and move certain GPU operators onto CPU resulting in an incompatible plan when AQE is on.

It also changes some of the the default optimizer configuration settings to avoid trying to force all joins back onto CPU.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -126,14 +129,30 @@ class CostBasedOptimizer(conf: RapidsConf) extends Logging {
...
-    if (!plan.canThisBeReplaced) {
+    if (!plan.canThisBeReplaced || mustThisBeReplaced(plan)) {
       // reset the costs because this section of the plan was not moved to GPU
       totalGpuCost = totalCpuCost
     }
 
     (totalCpuCost, totalGpuCost)
   }
 
+  private def mustThisBeReplaced(plan: SparkPlanMeta[_]): Boolean = {
```
jlowe, line 140: Does this really mean something more like, "planning for this node cannot be changed"?  Just because we find a shuffle or other node involving data transfer in the original plan, that doesn't mean we always are replacing those nodes with GPU versions, correct?

andygrove: Good point. This method is really just determining whether the operator is consuming output from a query stage, so I have renamed it `consumesQueryStage` now.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -115,25 +118,55 @@ class CostBasedOptimizer(conf: RapidsConf) extends Logging {
...
+      plan.costPreventsRunningOnGpu()
+      plan.childPlans.foreach(recursiveCostPreventsRunningOnGpu)
+    }
+  }
+
+  /**
+   * Determines whether the specified plan will read from a query stage.
+   */
+  private def consumesQueryStage(plan: SparkPlanMeta[_]): Boolean = {
+    if (SQLConf.get.adaptiveExecutionEnabled) {
```
gerashegalov, line 155: nit: if / else looks superficial:

```suggestion
    SQLConf.get.adaptiveExecutionEnabled && plan.wrapped match {
      case _: CustomShuffleReaderExec
             | _: ShuffledHashJoinExec
             | _: BroadcastHashJoinExec
             | _: BroadcastNestedLoopJoinExec => true
         case _ => false
    }
```

andygrove: Thanks. Updated.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -115,25 +118,50 @@ class CostBasedOptimizer(conf: RapidsConf) extends Logging {
...
     }
 
     (totalCpuCost, totalGpuCost)
   }
 
+  /**
+   * Recursively force a section of the plan back onto CPU, stopping once a plan
+   * is reached that is already on CPU or cannot be moved back onto CPU.
+   */
+  final def recursiveCostPreventsRunningOnGpu(plan: SparkPlanMeta[_]): Unit = {
```
andygrove, line 144: This was supposed to be `private def` not `final def`. I will push a fix later today.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -119,23 +119,9 @@ abstract class RapidsMeta[INPUT <: BASE, BASE, OUTPUT <: BASE](
...
-      childDataWriteCmds.foreach(_.recursiveCostPreventsRunningOnGpu())
-    }
-  }
-
   final def costPreventsRunningOnGpu(): Unit = {
     cannotRunOnGpuBecauseOfCost = true
     willNotWorkOnGpu("Removed by cost-based optimizer")
-    childExprs.foreach(_.recursiveCostPreventsRunningOnGpu())
-    childParts.foreach(_.recursiveCostPreventsRunningOnGpu())
-    childScans.foreach(_.recursiveCostPreventsRunningOnGpu())
```
revans2, line 138: Why were these removed? are we not reporting anymore that they cannot run on the GPU because of cost?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -115,25 +118,50 @@ class CostBasedOptimizer(conf: RapidsConf) extends Logging {
...
   }
 
+  /**
+   * Recursively force a section of the plan back onto CPU, stopping once a plan
+   * is reached that is already on CPU or cannot be moved back onto CPU.
+   */
+  private def recursiveCostPreventsRunningOnGpu(plan: SparkPlanMeta[_]): Unit = {
+    if (plan.canThisBeReplaced && !consumesQueryStage(plan)) {
+      plan.costPreventsRunningOnGpu()
+      plan.childPlans.foreach(recursiveCostPreventsRunningOnGpu)
```
revans2, line 147: I am starting to get confused with the algorithm and what it really intends to do so I have a hard time judging what is and isn't a bug.  Because now that we stop when we hit a consumsQueryStage doesn't that mess up the cost calculations that we did to get here?

Could we have some high level comments  about what the algorithm is trying to do?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
andygrove: This code was originally based on the logic used in `recursiveSparkPlanPreventsRunningOnGpu` and `recursiveSparkPlanRemoved` but in other cases (like fixing up join consistency) we were just calling `willNotWorkOnGpu` without recursing into the expr, parts, and scans because we are just tagging that the operator should not run on GPU. 

I wasn't sure why we would want to do this recursion part other than recursing through the operators, and removing this code didn't affect the unit tests or the results when running benchmarks so I removed it. Is this something that should be added back in? If so, do you have any guidance on how to test this part in the context of the optimizer?

revans2: The question really is about what is printed out when the GPU explain is enabled.  

Before this change all of the expressions, parts, and scans that are children would report that they were "Removed by cost-based optimizer" and include the `$` at the beginning indicating that it was removed because of cost.

After this change the expressions, parts, and scans will not say anything about the cost based optimizer, but will print out a `@` at the beginning and say it could run on the GPU but will not because of the spark plan it is a part of.

Which do we want to happen?  I don't know. Both appear reasonable to me. I just assumed that the code was put in on purpose but I guess not.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
andygrove: I've added some more documentation but the intent is that if we see an operator that cannot run on GPU (because we don't support it) or an operator that is consuming a query stage (so has to stay wherever it is) then we make the GPU cost be the same as the CPU cost. We only have cost differences where this actually potential to change the plan. At least, that is the intent.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
andygrove: Thanks. I think I should put this back in then. It does require adding another type of tag to RapidsMeta to track the concept of `mustThisBeReplaced` though.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -932,14 +932,14 @@ object RapidsConf {
       .internal()
       .doc("Default cost of transitioning from GPU to CPU")
       .doubleConf
-      .createWithDefault(0.15)
+      .createWithDefault(0.1)
```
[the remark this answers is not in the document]
andygrove: The cost model is very arbitrary but the default values for GPU are that operators are 0.8 and transitions are now 0.1 (there are not time units yet but relative costs compared to CPU). So for an operator with two children that both have transitions from CPU, the GPU cost now would be 0.8 + 0.1 + 0.1 = 1.0 so same as CPU whereas before it would have been 0.8 + 0.15 + 0.15 = 1.1 which is > CPU cost of 1.0 and we were trying to push all joins back onto CPU by default.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -170,6 +171,10 @@ abstract class RapidsMeta[INPUT <: BASE, BASE, OUTPUT <: BASE](
     }
   }
 
+  final def mustBeReplaced(because: String): Unit = {
+    mustBeReplacedReasons.get.add(because)
```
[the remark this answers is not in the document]
andygrove: The various mutable sets like `cannotBeReplacedReasons` and `mustBeReplaced` are initialized in `initReasons`:

```scala
def initReasons(): Unit = {
  cannotBeReplacedReasons = Some(mutable.Set[String]())
  ...
```
and I was basically copying the same pattern that we use in methods like `willNotWorkOnGpu`:

```scala
final def willNotWorkOnGpu(because: String): Unit = {
  cannotBeReplacedReasons.get.add(because)
  ...
```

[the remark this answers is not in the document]
andygrove: It might be cleaner to change these to `cannotBeReplacedReasons.foreach(_.add(because))` but it could also hide errors if we fail to initialize any of these sets.

andygrove:
Hi @revans2 @gerashegalov @razajafri do you have any more feedback on this PR? Thanks. I think I have addressed the feedback so far.

revans2 (approved):

andygrove:
build

razajafri (approved):

andygrove:
build

andygrove:
build

PR #1955, opened 2021-03-17
Fix typo in Arrow optimization config

andygrove:

This typo was fixed in the past but seems to have reverted in a subsequent PR merge issue.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
```diff
@@ -194,7 +194,7 @@ trait CudfBinaryExpression extends GpuBinaryExpression {
   def outputType(l: BinaryOperable, r: BinaryOperable) : DType = {
     val over = outputTypeOverride
     if (over == null) {
-      BinaryOperable.implicitConversion(l, r)
+      BinaryOperable.implicitConversion(binaryOp, l, r)
```
jlowe, line 197: This seems out of place in a PR to fix a typo.  Probably best to wait for https://github.com/NVIDIA/spark-rapids/pull/1889 to be merged and upmerge afterwards to pass premerge CI.

andygrove:
build

revans2 (commented):
I am fine with fixing the typo, it is on by default and internal so I don't suspect that anyone will actually use this config in the 0.4 release, otherwise we will have to add in support for deprecating/remapping a config.

revans2 (approved):

tgravescs:
this is weird, I fixed this already, did it not go into 0.5 - https://github.com/NVIDIA/spark-rapids/pull/1652/files

tgravescs:
oh sorry, just saw comment about it being reverted

PR #1958, opened 2021-03-17
Support legacy behavior of parameterless count 

### LINKED ISSUE #1737: Spark 3.1 now supports the legacy behavior of parameterless count
Spark 3.1 supports parameterless count. 

We should test if we are supporting that, if not then we need to add this functionality

**Additional context**
Please look at this [change](https://github.com/apache/spark/commit/63eb9101f2)

razajafri:
This PR returns a Long col with a single row with a value of 0

closes #1737 


razajafri:
@abellina can you take a look since you originally worked on aggregates?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -871,6 +871,17 @@ case class GpuHashAggregateExec(
             }
           }
         }
+        // If cvs is empty, we add a single row with zero value.
+        // This is to fix a bug in the plugin where a paramater-less count wasn't returning the
+        // desired result compared to Spark-CPU.
+        // For more details go to https://github.com/NVIDIA/spark-rapids/issues/1737
+        if (cvs.isEmpty) {
```
abellina, line 878: 1. Should we assert that the aggregate function is indeed `count`?
2. Does this only happen with reductions?
3. Need tests for this in the integration tests at least. I'd also expect them in the scala tests... though I defer to @revans2 where he'd like to see new tests go in.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -871,6 +871,17 @@ case class GpuHashAggregateExec(
...
+          withResource(Scalar.fromInt(0)) { ZERO =>
```
abellina, line 879: `fromLong` removes the cast I believe.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
revans2: So in a count with no parameter what is `agg.ref` above on line 866? It is pointing to something? How is it that we are not getting an NPE above?

razajafri: `aggModeCudfAggregates` is empty so that entire loop is never executed

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
razajafri: You are right thanks

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
razajafri: > 

    1. Should we assert that the aggregate function is indeed count?
    2. Does this only happen with reductions?
    3. Need tests for this in the integration tests at least. I'd also expect them in the scala tests... though I defer to @revans2 where he'd like to see new tests go in.

1. We can, I am not sure its needed as no other aggregate is parameterless. @revans2 is this needed? 
2. I am not sure
3. I have added tests

razajafri:
I have added tests. I assumed the count was tested but it wasn't the parameterless count obviously, should've known better.

@abellina have I answered your questions? 
@revans2 PTAL

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
revans2: I am sorry but the code around this is too complicated for me to be able to really grok it and see if what you have done is correct.  I personally think we need to refactor all of hash aggregate so it is actually understandable without spending a few days to dig into it.  I'll try and find some time to dig into the code, but for now I don't feel comfortable approving this.  I'll let @abellina deal with it.  Thanks for answering my questions.

razajafri:
@abellina are you OK with this PR?

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -274,6 +274,7 @@ def test_hash_query_multiple_distincts_with_non_distinct(data_gen, conf):
         'sum(distinct a),' +
         'count(distinct b),' +
         'count(a),' +
+        'count(),' +
```
abellina, line 277: As far as I understand, `count()` needs to be enabled for it to be considered by catalyst. Am I missing something?

Also, for group by, I believe `count()` _shouldn't_ work given the change you made is only for reductions. I tried this manually in my box, and `count()` seemed to be returning 0 for groups of 1, which seemed wrong to me.

If you run out of time today don't worry, I should be able to put up a fix for this.

razajafri: I should've added better tests. The config is only required for Spark3.1+. I have updated the tests now 

>Also, for group by, I believe count() shouldn't work given the change you made is only for reductions. I tried this manually in my box, and count() seemed to be returning 0 for groups of 1, which seemed wrong to me.

Is the CPU result not matching? The unit tests compare the CPU to GPU

abellina: I re-ran this both for CPU and GPU, and they both return 0 for `count()` even if the groups each would have a positive count.  So I think the `group by` count is ok.

```
scala> spark.sql("select count(),count(*) from foo group by value%2").collect
res8: Array[org.apache.spark.sql.Row] = Array([0,5], [0,5])

scala> spark.conf.set("spark.rapids.sql.enabled", "true")

scala> spark.sql("select count(),count(*) from foo group by value%2").collect
res10: Array[org.apache.spark.sql.Row] = Array([0,5], [0,5])
```

abellina: Thanks for adding the config.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -19,7 +19,7 @@ package com.nvidia.spark.rapids
 import scala.collection.mutable.ArrayBuffer
 
 import ai.rapids.cudf
-import ai.rapids.cudf.NvtxColor
+import ai.rapids.cudf.{DType, NvtxColor, Scalar}
```
abellina, line 22: `DType` doesn't seem to be used.

abellina:
@razajafri just thought of an edge case. Two `count()` "aggs" back to back. Does your code work with this?

```
scala> spark.sql("select count(),count() from foo").explain(true)
== Parsed Logical Plan ==
'Project [unresolvedalias('count(), None), unresolvedalias('count(), None)]
+- 'UnresolvedRelation [foo], [], false

== Analyzed Logical Plan ==
count(): bigint, count(): bigint
Aggregate [count() AS count()#255L, count() AS count()#256L]
+- SubqueryAlias foo
   +- SerializeFromObject [input[0, int, false] AS value#2]
      +- ExternalRDD [obj#1]

== Optimized Logical Plan ==
Aggregate [0 AS count()#255L, 0 AS count()#256L]
+- SerializeFromObject
   +- ExternalRDD [obj#1]

== Physical Plan ==
AdaptiveSparkPlan isFinalPlan=false
+- HashAggregate(keys=[], functions=[], output=[count()#255L, count()#256L])
   +- Exchange SinglePartition, ENSURE_REQUIREMENTS, [id=#983]
      +- HashAggregate(keys=[], functions=[], output=[])
         +- SerializeFromObject
            +- Scan[obj#1]

scala> spark.sql("select count(),count() from foo").collect
res32: Array[org.apache.spark.sql.Row] = Array([0,0])
```

integration_tests/src/main/python/hash_aggregate_test.py
razajafri: yes, so false will xfail only for Spark3.1+

abellina:
build

razajafri:
@abellina PTAL

abellina:
@razajafri thanks for adding tests. I am still not clear on how this works.

My theory is that it works because there is a projection at the end that just sets 0s (i.e. the scalar you are generating is not used). A quick test would be to use something other than 0 for your scalar, to see if that makes a difference.

If the above is true, I don't think it changes the impl. It would just be good to fully understand how this is propagating to the result.

razajafri:
> My theory is that it works because there is a projection at the end that just sets 0s (i.e. the scalar you are generating is not used). A quick test would be to use something other than 0 for your scalar, to see if that makes a difference.

You are right that's how its working. I did a quick test by using a scalar value of 2 and that had no effect on the result

abellina (approved):
@razajafri thanks for the changes and for the testing. I think this makes sense and I can't think of a simpler way at this point.

razajafri:
build

abellina (approved):

PR #1969, opened 2021-03-18
Enable StructType for Parquet and InMemoryTableScanExec

razajafri:
This is a draft PR for supporting structs for `InMemoryTableScanExec`

It depends on rapidsai/cudf#7461. So until that is checked in, this will remain a draft so we don't accidentally merge it.


shims/spark311/src/main/scala/org/apache/spark/sql/rapids/shims/spark311/GpuInMemoryTableScanExec.scala
```diff
@@ -16,7 +16,8 @@
```
razajafri, line 16: This file is mostly reformatted.

shims/spark311/src/main/scala/org/apache/spark/sql/rapids/shims/spark311/GpuInMemoryTableScanExec.scala
```diff
@@ -26,90 +27,95 @@ import org.apache.spark.sql.catalyst.plans.QueryPlan
...
+        buffers,
+        relation.output,
+        attributes,
+        conf).map { cb =>
+      if (numOutputRows != null) {
         numOutputRows += cb.numRows()
-        cb
       }
+      cb
     }
```
razajafri, line 73: This is the only relevant change in this file. The rest is indentation change

shims/spark311/src/main/scala/org/apache/spark/sql/rapids/shims/spark311/GpuInMemoryTableScanExec.scala
```diff
@@ -26,90 +27,95 @@ import org.apache.spark.sql.catalyst.plans.QueryPlan
...
+  override def vectorTypes: Option[Seq[String]] =
+    relation.cacheBuilder.serializer.vectorTypes(attributes, conf)
+
+  private lazy val columnarInputRDD: RDD[ColumnarBatch] = {
+    val rapidsConf = new RapidsConf(conf)
+    val numOutputRows = if (MetricsLevel(rapidsConf.metricsLevel) == DEBUG_LEVEL) {
+      longMetric("numOutputRows")
+    } else {
+      null
+    }
```
jlowe, line 61: Is this code necessary?  `GpuExec` is already conditionally creating the output rows metric at debug level, see https://github.com/NVIDIA/spark-rapids/blob/branch-0.5/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala#L217-L221.  I think this just needs to use the already created `GpuMetric` directly and not check for null later:
```suggestion
    val numOutputRows =gpuLongMetric(GpuMetrics.NUM_OUTPUT_ROWS)
```

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/Spark311Shims.scala
```diff
@@ -275,7 +275,7 @@ class Spark311Shims extends Spark301Shims {
         }),
       GpuOverrides.exec[InMemoryTableScanExec](
         "Implementation of InMemoryTableScanExec to use GPU accelerated Caching",
-        ExecChecks(TypeSig.commonCudfTypes + TypeSig.DECIMAL, TypeSig.all),
+        ExecChecks(TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.STRUCT, TypeSig.all),
```
jlowe, line 278: Shouldn't this be something like:
```suggestion
        ExecChecks((TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.STRUCT).nested(), TypeSig.all),
```

jlowe (changes requested):

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/Spark311Shims.scala
razajafri: It is not very clear to me. Do we have documentation on this?

jlowe: The Scaladoc for TypeChecks, etc. is the documentation.  It should be clear whether you need it or not based on the generated docs which are missing in this PR.  I suspect without using `nested()` the docs will say STRUCT is supported but not any of the types within that struct.

razajafri: I used to be able to gen docs by invoking `mvn build` thats not happening now.

jlowe: `build` is not part of the Maven default lifecycle goals.  You should be running `mvn verify` to perform the doc generation per [the pom file](https://github.com/NVIDIA/spark-rapids/blob/branch-0.5/dist/pom.xml#L162-L180) and [contributing guide](https://github.com/NVIDIA/spark-rapids/blob/branch-0.5/CONTRIBUTING.md#building-from-source).

razajafri: sorry, I meant to write `mvn install`

I think something broke in my workflow because I used to see the docs generated by simply running my script. Will update and get back

razajafri: The reason why the docs are not generated is because they aren't auto-generated for the Shims. If I manually generate it, where do the docs go for the Shims?
That could be a bigger effort, should we may be consider handling it as a separate PR?

jlowe: Yes, we should handle doc generation for stuff that requires a minimum Spark version beyond the base supported Spark version as a separate issue.  I filed #1977 to track.

integration_tests/src/main/python/cache_test.py
```diff
@@ -152,7 +154,7 @@ def op_df(spark, length=2048, seed=0):
 
     assert_gpu_and_cpu_are_equal_collect(op_df, conf = enableVectorizedConf)
 
-@pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
+@pytest.mark.parametrize('data_gen', all_gen + [all_basic_struct_gen, StructGen([['child0', byte_gen]])], ids=idfn)
```
jlowe, line 157: What about testing structs containing more than one type or deeper nesting such as struct-of-structs?

razajafri: @jlowe I have updated the tests for testing nested struct. PTAL

sameerz:
Please retarget to 0.6 once that branch is open, as the dependent cudf issue is in 0.20.

razajafri:
@jlowe PTAL, the Java bindings have merged just an hour ago so the CI will fail as I understand, it builds against the nightly cudf? 

If you are OK with this, I can merge it tomorrow after the CI will see the changes in cudf  that were merged just today

docs/additional-functionality/cache-serializer.md
```diff
@@ -29,12 +29,40 @@ nav_order: 2
   `spark.sql.inMemoryColumnarStorage.enableVectorizedReader` will not be honored as the GPU
   data is always read in as columnar. If `spark.rapids.sql.enabled` is set to false
   the cached objects will still be compressed on the CPU as a part of the caching process.
-  Also note that Parquet doesn't support CalendarIntervalType or NullType out of the box, but
-  ParquetCachedBatchSerializer does by decomposing intervals to struct containing the
-  months, days and microseconds and NullType to Int column containing nulls.
-
+  Please note that Parquet doesn't support columns of type CalendarIntervalType or NullType out of
+  the box, but ParquetCachedBatchSerializer does by decomposing intervals to struct containing the
+  months, days and microseconds and NullType to Int column containing nulls. Also note that
```
jlowe, line 34: This seems like too much implementation detail because the user won't be able to access the underlying representation.  Personally I'd not mention it at all or at best say CalendarIntervalType and NullType are supported and leave it at that.  But if the detail stays:
```suggestion
  the box, but ParquetCachedBatchSerializer does by decomposing an interval to a struct containing the
  months, days and microseconds and NullType to an Int which is always null. Also note that
```

docs/additional-functionality/cache-serializer.md
```diff
@@ -29,12 +29,40 @@ nav_order: 2
...
+  ParquetCachedBatchSerializer doesn't support negative decimal scale, so if 
+  `spark.sql.legacy.allowNegativeScaleOfDecimal` is set to true ParquetCachedBatchSerializer
+  should not be used, if used, it will result in an error resulting from trying to write a
+  negative scale to parquet.
```
jlowe, line 38:
```suggestion
  should not be used. Using the serializer with negative decimal scales will generate
  an error at runtime.
```

docs/additional-functionality/cache-serializer.md
```diff
@@ -29,12 +29,40 @@ nav_order: 2
...
+  
   To use this serializer please run Spark
```
jlowe, line 40:
```suggestion
  To use this serializer, set the `spark.sql.cache.serializer` config to the serializer class as in this example:
```

docs/additional-functionality/cache-serializer.md
```diff
@@ -29,12 +29,40 @@ nav_order: 2
...
   ```
   spark-shell --conf spark.sql.cache.serializer=com.nvidia.spark.rapids.shims.spark311.ParquetCachedBatchSerializer"
   ```
   To use the default serializer don't set the `spark.sql.cache.serializer` conf
```
jlowe, line 44:
```suggestion
 ```

File: shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/Spark311Shims.scala [hunk 10]
```diff
@@ -275,12 +275,19 @@ class Spark311Shims extends Spark301Shims {
         }),
       GpuOverrides.exec[InMemoryTableScanExec](
         "Implementation of InMemoryTableScanExec to use GPU accelerated Caching",
-        ExecChecks(TypeSig.commonCudfTypes + TypeSig.DECIMAL, TypeSig.all),
+        ExecChecks((TypeSig.commonCudfTypes + TypeSig.DECIMAL.withPsNote(TypeEnum.DECIMAL,
+          "Negative scales aren't supported at the moment even with " +
+            "spark.sql.legacy.allowNegativeScaleOfDecimal set to true. This is because Parquet " +
+            "doesn't support negative scale for decimal values") + TypeSig.STRUCT).nested(),
```
jlowe, line 281: I think it would be easier to see the supported types if the PS note wasn't smack in the middle of the list, e.g.:
```suggestion
        ExecChecks((TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.STRUCT).nested()
            .withPsNote(TypeEnum.DECIMAL,
                "Negative scales aren't supported at the moment even with " +
                "spark.sql.legacy.allowNegativeScaleOfDecimal set to true. Parquet " +
                "doesn't support negative scale for decimal values"),
```

shims/spark311/src/main/scala/org/apache/spark/sql/rapids/shims/spark311/GpuInMemoryTableScanExec.scala
```diff
@@ -28,87 +28,86 @@ import org.apache.spark.sql.execution.{LeafExecNode, SparkPlan}
...
-    override val nodeName: String = {
-      relation.cacheBuilder.tableName match {
-        case Some(_) =>
-          "Scan " + relation.cacheBuilder.cachedName
-        case _ =>
-          super.nodeName
-      }
+case class GpuInMemoryTableScanExec(attributes: Seq[Attribute],
+   predicates: Seq[Expression],
+   @transient relation: InMemoryRelation) extends LeafExecNode with GpuExec {
```
jlowe, line 33: 4-space indent for continuations
```suggestion
    predicates: Seq[Expression],
    @transient relation: InMemoryRelation) extends LeafExecNode with GpuExec {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -97,18 +97,54 @@ object GpuParquetFileFormat {
...
-  def getPrecisionList(schema: StructType): Seq[Int] =  {
-    def precisionsList(t: DataType): Seq[Int] = {
-      t match {
-        case d: DecimalType => List(d.precision)
-        case _: StructType =>
-          throw new IllegalStateException("structs are not supported right now")
-        case _: ArrayType =>
-          throw new IllegalStateException("arrays are not supported right now")
-        case _ => List(0)
+  def parquetWriterOptionsHelper(schema: StructType): ParquetWriterOptions.Builder[_] = {
```
jlowe, line 100: This could use a better name and some documentation on what it does.  "Helper" doesn't convey much about what it's doing.  How about something more descriptive, like `writerOptionsFromSchema`?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -97,18 +97,54 @@ object GpuParquetFileFormat {
...
+    def addColumnMeta(name: String, nullable: Boolean,
```
jlowe, line 101: I'm not a fan of internal functions that are quite sizeable, as they get in the way of reading the function they are in.  I suggest moving this out as a private peer method.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -97,18 +97,54 @@ object GpuParquetFileFormat {
...
+          }
+          builder.withStructColumn(structBuilder.build())
+        case a: ArrayType =>
+          val listBuilder = ParquetWriterOptions.listBuilder(entry.name, entry.nullable)
+          // lists have only one child, no name is needed
+          addColumnMeta("", entry.nullable, listBuilder, a.elementType)
+          builder.withListColumn(listBuilder.build())
+        case _ =>
+          builder.withColumn(entry.nullable, entry.name)
+      }
```
jlowe, line 145: This looks like we copied `addColumnMeta` again here?

razajafri:
closed in favor of https://github.com/NVIDIA/spark-rapids/pull/2271

PR #1972, opened 2021-03-19
support GpuSize

sperlingxx:
This PR is to support `GpuSize`.
The intention is to fix test failures on  `GpuExplode` under spark 3.1.  The root cause of the failure is that in spark 3.1 newly added rule `InferFiltersFromGenerate` will insert a group by expressions include GPU unsupported expression `Size`.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
@@ -36,7 +36,7 @@ case class GpuSize(child: Expression, legacySizeOfNull: Boolean)
       return GpuColumnVector.from(GpuScalar.from(0), 0, IntegerType).getBase
     }
 
-    // Compute sizes of each cuDF.ListType to get sizes of each ArrayData and MapData, considering
+    // Compute sizes of cuDF.ListType to get sizes of each ArrayData or MapData, considering
     // MapData is represented as List of Struct in terms of cuDF.
     // We compute list size via subtracting the offset of next element(row) to the current offset.
     val collectionSize = {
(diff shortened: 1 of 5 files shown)


sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -0,0 +1,71 @@
...
+    // Compute sizes of cuDF.ListType to get sizes of each ArrayData or MapData, considering
+    // MapData is represented as List of Struct in terms of cuDF.
+    // We compute list size via subtracting the offset of next element(row) to the current offset.
+    val collectionSize = {
+      // Here is a hack: using index -1 to fetch the offset column of list.
+      // In terms of cuDF native, the offset is the first (index 0) child of list_column_view.
+      // In JNI layer, we add 1 to the child index when fetching child column of ListType to keep
+      // alignment.
+      // So, in JVM layer, we have to use -1 as index to fetch the real first child of list_column.
+      withResource(inputBase.getChildColumnView(-1)) { offset =>
```
revans2, line 48: Lets just add in JNI for this instead.

https://github.com/rapidsai/cudf/blob/6fe8b1b7be71c0f3587d3ef43c535aec376d1e45/cpp/include/cudf/lists/count_elements.hpp#L49

It does everything we want, except for the legacy null behavior, and we can fix that up without too much work.

```
var ret = intput.getBase.countElements()
if (legacySizeOfNull) {
    withResource(input.getBase.isNull()) { isNull =>
        withResource(GpuScalar.from(-1)) { negOne =>
            withResource(ret) { count =>
                ret = isNull.ifElse(negOne, count)
            }
        }
    }
}
```

revans2 (commented):
The code itself looks fine, but I just would prefer for us to do the right thing instead of putting in a hack like this.

revans2 (approved):
I missed the part about this is blocking issues with the nightly tests.  Lets merge this in as is and have a follow on to make ti work correctly.  I'll get a PR up for the JNI code ASAP so we can get a real fix in soon too.

revans2:
I filed https://github.com/NVIDIA/spark-rapids/issues/1974 as a follow on issue for this.

Issue #1974, opened 2021-03-19
Follow on issues for GpuSize


revans2:
The GpuSize work #1972 will fix the failing nightly tests, so we are going to merge it in fast, but there are a few things that should be fixed in follow on work.

1. cudf has an API to do exactly what we want. We should use that API instead of hacking together something ourselves.
2. The signature for the input parameters are not strictly correct.  They are not technically wrong either.  The problem is that they say that we can support anything as the child of a map or array, but we have not plumbed through a lot of things calendar interval, or UDT. For the types that we support we should either extend the tests to cover some examples of these, or we should remove them from the list of supported types.  Just so we are cautious and also so our documentation is as accurate as possible.

On a side note I should have a PR up shortly for `countElements` so we can get this done ASAP.

revans2:
I just put up the PR https://github.com/rapidsai/cudf/pull/7651 in CUDF for the changes that are needed there to do this the right way.

sperlingxx:
Thank you, @revans2. I've put up a PR to fix this.

sameerz:
@sperlingxx is this resolved?  If so can you mark the issue closed?

PR #1981, opened 2021-03-22
Fix GpuSize

sperlingxx:
This pull request is to address #1974.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
@@ -32,18 +32,17 @@ case class GpuSize(child: Expression, legacySizeOfNull: Boolean)
 
   override protected def doColumnar(input: GpuColumnVector): ColumnVector = {
 
-    val nullScalar = if (legacySizeOfNull) {
-      GpuScalar.from(-1)
-    } else {
-      GpuScalar.from(null, IntegerType)
-    }
     // Compute sizes of cuDF.ListType to get sizes of each ArrayData or MapData, considering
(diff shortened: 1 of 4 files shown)


sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -31,38 +31,17 @@ case class GpuSize(child: Expression, legacySizeOfNull: Boolean)
...
-      withResource(inputBase.getChildColumnView(-1)) { offset =>
-        withResource(offset.subVector(1)) { upBound =>
-          withResource(offset.subVector(0, offset.getRowCount.toInt - 1)) { lowBound =>
-            upBound.sub(lowBound)
-          }
-        }
-      }
-    }
 
     val nullScalar = if (legacySizeOfNull) {
```
revans2, line 35: I think `countElements` already does the right thing for `legacySizeOfNull == false`. That should be the common case so why do the extra computation if it is not needed.

```
withResource(input.getBase.countElements()) { collectionSize =>
    if (legacySizeOfNull) {
        withResource(GpuScalar.from(-1)) { nullScalar =>
            withResource(input.getBase.isNull) { inputIsNull =>
              inputIsNull.ifElse(nullScalar, collectionSize)
            }
        }
    }  else {
      collectionSize.incRefCount()
    }
}
```

sperlingxx: Fixed. Thx!

sperlingxx:
build

revans2 (approved):

PR #1982, opened 2021-03-22
Update ORC pushdown filter building to latest Spark logic

jlowe:
`OrcFilters` can crash with `NoSuchMethodError: org.apache.spark.sql.connector.catalog.CatalogV2Implicits$.quoteIfNeeded(Ljava/lang/String;)Ljava/lang/String;` when running on Spark 3.2.0.

Looking at our `OrcFilters`, the Spark version has changed a bit and there's some good fixes and features we're missing. Updating to the latest code picks up these fixes and also resolves the quoting method error.


jlowe:
build

gerashegalov (approved):
LGTM

PR #1991, opened 2021-03-22
Fixes creation of invalid DecimalType in GpuDivide.tagExprForGpu

### LINKED ISSUE #1984: [BUG] NDS q58 Decimal scale (59) cannot be greater than precision (38).
**Describe the bug**

NDS q58 fails in CI when running against Spark 3.1.1 against the 1TB parquet dataset with decimals.

```
org.apache.spark.sql.AnalysisException: Decimal scale (59) cannot be greater than precision (38).
  at org.apache.spark.sql.types.DecimalType.<init>(DecimalType.scala:48)
  at com.nvidia.spark.rapids.GpuOverrides$$anon$97.tagExprForGpu(GpuOverrides.scala:1736
```

razajafri:
While supporting DIV changes, I completely overlooked the fact that the precision and scale can be truncated. 
This addresses that case. 

fixes #1984

### DIFF
--- a/integration_tests/src/main/python/arithmetic_ops_test.py
+++ b/integration_tests/src/main/python/arithmetic_ops_test.py
@@ -69,7 +69,10 @@ def test_multiplication_mixed(lhs, rhs):
                 f.col('a') * f.col('b')),
             conf=allow_negative_scale_of_decimal_conf)
 
-@pytest.mark.parametrize('data_gen', [double_gen, decimal_gen_neg_scale, DecimalGen(6, 3), DecimalGen(5, 5), DecimalGen(6, 0)], ids=idfn)
+@pytest.mark.parametrize('data_gen', [double_gen, decimal_gen_neg_scale, DecimalGen(6, 3),
+ DecimalGen(5, 5), DecimalGen(6, 0),
+pytest.param(DecimalGen(38, 21), marks=pytest.mark.xfail(reason="The precision is too large to be supported on the GPU")),
+pytest.param(DecimalGen(21, 17), marks=pytest.mark.xfail(reason="The precision is too large to be supported on the GPU"))], ids=idfn)
 def test_division(data_gen):
     data_type = data_gen.data_type
     assert_gpu_and_cpu_are_equal_collect(
--- a/integration_tests/src/main/python/arithmetic_ops_test.py
+++ b/integration_tests/src/main/python/arithmetic_ops_test.py
@@ -20,6 +20,7 @@ from marks import incompat, approximate_float
 from pyspark.sql.types import *
 from spark_session import with_cpu_session, with_gpu_session, with_spark_session, is_before_spark_310
 import pyspark.sql.functions as f
+from pyspark.sql.utils import IllegalArgumentException
 
 decimal_gens_not_max_prec = [decimal_gen_neg_scale, decimal_gen_scale_precision,
         decimal_gen_same_scale_precision, decimal_gen_64bit]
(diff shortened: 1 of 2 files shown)


razajafri:
@andygrove thank you for finding this. Please take a look at this when you get a chance. This will basically behave the same as before, which is preclude the Exp from being on the GPU.

razajafri:
build

andygrove:
@razajafri I tested locally and can confirm that it fixed the issue when running with NDS query 58.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1714,29 +1714,39 @@ object GpuOverrides {
...
-              // To find out if outputType.precision < outputType.scale simplifies to p1 < s1,
-              // which is never possible
-              //
-              // case 2:
-              // outputType.precision = p1 - s1 + s2 + 6 + 1
-              // outputType.scale = 6 + 1
-              // To find out if outputType.precision < outputType.scale simplifies to p1 < 0
-              // which is never possible
+              // Case 1: OutputType.precision doesn't get truncated
+              //   We will never hit a case where outputType.precision < outputType.scale + r.scale.
```
andygrove, line 1718: I think it would be good (and fairly easy) to add a unit test for this. I had a go and found some values that do hit this case but it is possible that I am using values that couldn't happen in real life. I am not sure.

Here is one example:

```
l = DecimalType(3,0)
r = DecimalType(21,17)
outputType = DecimalType(38,23)
38 < 23 + 17
```

andygrove: Perhaps I am hitting case 2 here where precision is getting truncated?

andygrove: It does look like I am hitting case 2 here so I think the logic in the comments is sound, but a unit test would make me more confident and would protect against future regressions if any of this code gets updated.

revans2: +1 for the unit test to verify that this is working.  The logic looks good to me. I am a little concerned about the coupling between this code and the code in `GpuDivideUtil`, but that is minor for now.

razajafri: Yes, I wrote a unit test locally to verify this. I don't know why I didn't check it in. Let me do that

razajafri: > l = DecimalType(3,0)
r = DecimalType(21,17)
outputType = DecimalType(38,23)
38 < 23 + 17
So, like you said this won't happen in reality as by the time we get the call both `l` and `r` Spark will be the same precision and scale, I think it will upcast the Decimal(3,0) to Decimal(21,17).

razajafri:
@andygrove @revans2 I have added a unit test PTAL

andygrove (dismissed):
LGTM. Thanks @razajafri

razajafri:
build

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -69,7 +69,10 @@ def test_multiplication_mixed(lhs, rhs):
                 f.col('a') * f.col('b')),
             conf=allow_negative_scale_of_decimal_conf)
 
-@pytest.mark.parametrize('data_gen', [double_gen, decimal_gen_neg_scale, DecimalGen(6, 3), DecimalGen(5, 5), DecimalGen(6, 0)], ids=idfn)
+@pytest.mark.parametrize('data_gen', [double_gen, decimal_gen_neg_scale, DecimalGen(6, 3),
+ DecimalGen(5, 5), DecimalGen(6, 0),
+pytest.param(DecimalGen(38, 21), marks=pytest.mark.xfail(reason="The precision is too large to be supported on the GPU")),
+pytest.param(DecimalGen(21, 17), marks=pytest.mark.xfail(reason="The precision is too large to be supported on the GPU"))], ids=idfn)
```
revans2, line 75: If I remove you patch and keep the test there will be no real difference in the results, unless someone goes through the logs manually and looks that in one case we failed by falling back to the CPU, and in another case we failed by throwing an exception about going over the limit. In both cases the xfail ignored the exception that was thrown.

revans2 (changes requested):

integration_tests/src/main/python/arithmetic_ops_test.py
razajafri: I have updated the test to check if the test raises IllegalArgumentException. Let me know if you want me to tighten it down even more. Ideally we should be checking the message to make sure its failing because its not columnar but I am not sure how to accomplish that using the pytest xfail

razajafri:
@revans2 I tested the new test on top of branch-0.5 and it fails without my fix. PTAL

razajafri:
build

razajafri:
@revans2 do you have any more concerns?

revans2 (approved):

razajafri:
Insufficient CPU

Issue #1625, opened 2021-01-29
[FEA] Support Decimal Casts to String and String to Decimal


tgravescs:
**Is your feature request related to a problem? Please describe.**
 I was testing reading some Decimal values from a datasource v2 and just doing a show() on it and it couldn't run on GPU due to Decimal cast to String not supported.

We should add support for these, not sure on priority, but if users like using show() it might come up quite a bit.

sameerz:
Related to https://github.com/rapidsai/cudf/pull/7623

PR #1999, opened 2021-03-23
support casting string to decimal

sperlingxx:

This pull request is to support casting string to decimal, which is required by #1625. To apply HALF_UP rounding, we cast string to float64 at first instead of casting to decimal directly. 
In addition,  the support of NaN value on float to decimal conversion is appended in this PR.


sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -387,6 +387,14 @@ case class GpuCast(
               }
           }
         }
+      case (StringType, dt: DecimalType) =>
+        // To apply HALF_UP rounding strategy during casting to decimal, we firstly cast
+        // string to fp64. Then, cast fp64 to target decimal type to enforce HALF_UP rounding.
+        withResource(input.getBase.strip()) { trimmed =>
+          withResource(castStringToFloats(trimmed, ansiMode, DType.FLOAT64)) { fp =>
```
revans2, line 394: This is not compatible with what Spark does. I see you fixed some bugs with `NaN`s? But what about `-Inf` or `+Inf`.   What about if we go over the max precision that a double can support? It will end up truncating the data and we will lose data. For example if I tried to convert "9999999999999999" this way you would be off.

If we want to be able to round HALF_UP then I would suggest we parse the value as a decimal with precision + 1 and scale + 1. Then do the rounding. This means that we will not be able to cast a string to a decimal if the precision + 1 is more than we can support. But it is better than producing an incorrect answer.

sperlingxx: Hi @revans2, I found that the test data for `Double.NaN` values are also include `Double.PositiveInfinity` and `Double.NegativeInfinity` values. So, it works for `-Inf` and `+Inf`.

And I tested with values like `9999999999999999`, the GPU results may slightly differ from CPU results (like: 99999999999999999 vs 99999999999999987, 99999999999999.00 vs 99999999999999.10), but the tolerance is not beyond the threshold.

Actually, in `castFloatsToDecimal`, we enforce HALF_UP rounding strategy through "parse the value as a decimal with precision + 1 and scale + 1. Then do the rounding.".

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
revans2: > And I tested with values like 9999999999999999, the GPU results may slightly differ from CPU results (like: 99999999999999999 vs 99999999999999987, 99999999999999.00 vs 99999999999999.10), but the tolerance is not beyond the threshold.

We allow float values to not be exact only because of order of operation in computation. In all other cases if we cannot produce the exact result it is off by default.  Decimal is not float.  We do not allow decimal values to be off, because there is no order of operation ambiguity in decimal computation. 

What is more casting a float to a decimal is off by default because we cannot produce the same result that Spark does. 

https://github.com/NVIDIA/spark-rapids/blob/ec2dcd7e6a34197fa11cc7fcc071523bee065e1b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala#L535-L539

Casting a string to a float is also off by default because we cannot produce the exact same result that Spark does in all cases.

https://github.com/NVIDIA/spark-rapids/blob/ec2dcd7e6a34197fa11cc7fcc071523bee065e1b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala#L555-L564

We cannot implement string to decimal and have it on by default unless it produces the correct result in ALL cases. If you want to add another config and document how it falls over, then we can merge it in, but we will need a follow on issue to fix the bugs.

sperlingxx:
build

sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
sperlingxx: I added RapidsConf `castStringToDecimal`, and filed issue #2019 for this problem.

revans2 (approved):

PR #2006, opened 2021-03-24
Enable windowed `collect_list` by default

firestarman:
`collect_list ` already supports excluding null values, the same with Spark.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -2402,7 +2402,7 @@ object GpuOverrides {
         override def convertToGpu(): GpuExpression = GpuPosExplode(childExprs(0).convertToGpu())
       }),
     expr[CollectList](
-      "Collect a list of elements.",
+      "Collect a list of elements, now only supported by windowing.",
       // It should be 'fullAgg' eventually but now only support windowing,
       // so 'aggNotGroupByOrReduction'
       ExprChecks.aggNotGroupByOrReduction(
@@ -2415,7 +2415,7 @@ object GpuOverrides {
       (c, conf, p, r) => new ExprMeta[CollectList](c, conf, p, r) {
         override def convertToGpu(): GpuExpression = GpuCollectList(
           childExprs.head.convertToGpu(), c.mutableAggBufferOffset, c.inputAggBufferOffset)
-      }).disabledByDefault("for now it is only supported by windowing."),
+      }),
     expr[ScalarSubquery](
       "Subquery that will return only one row and one column",
       ExprChecks.projectOnly(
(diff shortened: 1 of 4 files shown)


firestarman:
build

firestarman:
Not sure we could enable it by default or keep unchanged.

firestarman:
build

jlowe (commented):
If we're going to mark this as being supported outside of windowing, don't we need corresponding tests that prove it?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2415,8 +2415,7 @@ object GpuOverrides {
       (c, conf, p, r) => new ExprMeta[CollectList](c, conf, p, r) {
         override def convertToGpu(): GpuExpression = GpuCollectList(
           childExprs.head.convertToGpu(), c.mutableAggBufferOffset, c.inputAggBufferOffset)
-      }).disabledByDefault("for now the GPU collects null values to a list, but Spark does not." +
-      " This will be fixed in future releases."),
+      }).disabledByDefault("for now it is only supported by windowing."),
```
revans2, line 2418: Do we have a test that shows that we try to run this in the case of a groupby or aggregation? `ExprChecks.aggNotGroupByOrReduction` should take care of only letting it happen as a window function.  If that is not happening, then we need to fix that code instead of turning off the operator

revans2 (changes requested):

firestarman:
> If we're going to mark this as being supported outside of windowing, don't we need corresponding tests that prove it?

Sorry for confusion.
Actually it is still only supported by windowing. This change here focuses on the doc update about the fix of excluding null values.

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
firestarman: Sorry for confusion.
Actually it is still only supported by windowing. This change here focuses on the doc update about the fix of excluding null values.
I changed to enable it by default. As you said, we have `aggNotGroupByOrReduction ` to make sure the context. In windowing, `collect_list ` works as what Spark version does now.

jlowe (approved):
Looks ok to me.  Might be nice to have negative tests to verify we are properly _not_ replacing a groupby or reduction `collect_list`.

revans2 (approved):

firestarman:
> Looks ok to me. Might be nice to have negative tests to verify we are properly _not_ replacing a groupby or reduction `collect_list`.

Good suggestion, do we have such knid of tests for reference ?

jlowe:
> Good suggestion, do we have such knid of tests for reference ?

See [`assert_gpu_fallback_collect`](https://github.com/NVIDIA/spark-rapids/blob/branch-0.5/integration_tests/src/main/python/asserts.py#L286) and tests that use it.

PR #2005, opened 2021-03-24
Update to new is_before_spark_311 function name

jlowe:
#1958 was committed with references to `is_before_spark_310` but this function was renamed to `is_before_spark_311`.  This fixes the incorrect name to allow the integration tests to be collected properly by pytest.

PR #2007, opened 2021-03-24
Fix unknown column name error when filtering ORC file with no names

jlowe:
https://github.com/rapidsai/cudf/pull/7624 recently exposed an issue in the plugin when handling an ORC file with no column names.  The plugin was filtering the stripes based on the columns being referenced but then asking for all columns in the original file schema.  After the recent cudf change, that caused an "Unknown column" exception to be thrown.

This updates the code to filter the requested columns to the referenced columns when dealing with a file that has no column names.  The existing no-column-name ORC test now passes with this change, so no new test was added.


jlowe:
This will need to be upmerged with #2005 before the CI build will pass.

jlowe:
build

tgravescs (approved):

Issue #1980, opened 2021-03-22
[FEA] `GpuWindowInPandas` supports sending multiple groups to Python.


firestarman:
The same with Spark logic, currently `GpuWindowInPandas` sends only one group one time to Python, the performance is quite bad when there are too many small groups for GPU.
Thanks to the Python implementation, we can send multiple groups (a whole batch) to Python one time for bounded windows, without change in Python. It will be good to have this.

jlowe:
@firestarman is this issue still relevant?

firestarman:
I think we can close this now, since it is low priority and no plan for it. 
Closing

PR #2008, opened 2021-03-24
Let `WindowInPandas` support sending multiple groups

firestarman:
Let `WindowInPandas ` support sending all the groups in a batch one time to Python for better performance.

Thanks to the Python implementation, we can do this without Python change for bounded windows, by having the window bounds been relative to the head of the whole batch, instead of the head of its group.

Disable it by default due to some limitations.

- Add a config to enable/disable this feature.
- Move bounds building into iterator instead of the methods of the plan to avoid serializing the whole plan.

closes https://github.com/NVIDIA/spark-rapids/issues/1980


firestarman:
build

firestarman:
build

firestarman:
@revans2 Could you help review this ?
 Yes, it is a little big, but i tried to make it as small as possible.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExecBase.scala
```diff
@@ -78,96 +77,265 @@ abstract class GpuWindowInPandasExecMetaBase(
...
+          val (lowerCV, upperCV, _) = genColumnsForWindowBounds(frame, gpInfo, 0)
+          Seq(lowerCV, upperCV)
+        case f => throw new UnsupportedOperationException(s"Unsupported window frame $f")
+      }
+    }.map(GpuColumnVector.from(_, IntegerType))
+    new ColumnarBatch(boundsCVs.toArray, batch.numRows())
+  }
+
+  /*
+   * (Currently do it be leveraging the existing APIs in cuDF Java.
```
revans2, line 181: I would prefer to have an actual explanation of what this API is supposed to do.  This code is rather complex now and the explanation I think would be more helpful.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExecBase.scala
```diff
@@ -78,96 +77,265 @@ abstract class GpuWindowInPandasExecMetaBase(
...
+        table.orderBy(partitionIndices.map(id => cudf.Table.asc(id, true)): _*)
+      }
+      val (countHostCol, numRows) = withResource(orderedTable) { table =>
+        // Yes copying the data to host, it would be OK since just copying the aggregated
+        // count column.
+        (table.getColumn(table.getNumberOfColumns - 1).copyToHost(), table.getRowCount)
+      }
+      withResource(countHostCol) { cntCol =>
+        // Verified the type of Count column is integer, so use getInt
+        (0L until numRows).map(id => cntCol.getInt(id))
```
revans2, line 209: Is this lazy? I get nervous around `map` because in some cases it will not do the calculation now, it will just capture the map function and do it later on.  If that is happening here, then the host column may have been released by the time this code executes.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExecBase.scala
```diff
@@ -78,96 +77,265 @@ abstract class GpuWindowInPandasExecMetaBase(
...
+        (batch, groupLens)
+      }
+    }
+    (nextBatch, buildInfoBatchForPython(nextBatch, gpLens))
+  }
+
+  private[this] def splitBatchPerGroups(
+      batch: ColumnarBatch,
+      gpLens: Seq[Int]): Seq[ColumnarBatch] = {
+    val splitIndices = gpLens.dropRight(1).scan(0)(_+_).tail
```
revans2, line 142: A comment here would be really helpful. It is not at all obvious why we need to remove the last element from the input sequence, do a running sum on the values, and then drop the first element from the list of indexes.

I think I understand why, but defining exactly what `gpLens` is meant to hold would make it much clearer what is happening.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExecBase.scala
```diff
@@ -78,96 +77,265 @@ abstract class GpuWindowInPandasExecMetaBase(
...
+        groupBatchesAndSizes.head
+      } else {
+        (batch, groupLens)
+      }
+    }
+    (nextBatch, buildInfoBatchForPython(nextBatch, gpLens))
+  }
+
+  private[this] def splitBatchPerGroups(
+      batch: ColumnarBatch,
```
revans2, line 140: This is closed by this function. It would be nice to have that documented.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExecBase.scala
```diff
@@ -78,96 +77,265 @@ abstract class GpuWindowInPandasExecMetaBase(
...
-            table
-              .groupBy(partitionIndices:_*)
-              .aggregate(Aggregation.count(true).onColumn(0))
-          }
+      val groupLens = genGroupInfoFromBatch(batch)
+      if (isSplitGroups && groupLens.size > 1) {
+        // Split the groups when being required and there are more than one group.
+        val groupBatchesAndSizes = splitBatchPerGroups(batch, groupLens)
+          .zip(groupLens.map(Seq(_)))
+        val tails = groupBatchesAndSizes.tail.map {
```
revans2, line 124: Again a comment here would be helpful. Having the tail on the groupBatchesAndSizes fells wrong because it looks like we are dropping some data, and it is not clear that we have to go down the head to be returned.  It also looks like there might be some leaks on failure, but that is minor compared to having the code be readable.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExecBase.scala
```diff
@@ -78,96 +77,265 @@ abstract class GpuWindowInPandasExecMetaBase(
...
         }
-        val (countHostCol, numRows) = withResource(orderedTable) { table =>
-          // Yes copying the data to host, it would be OK since just copying the aggregated
-          // count column.
-          (table.getColumn(table.getNumberOfColumns - 1).copyToHost(), table.getRowCount)
+      }
+    }
+  }
+
+  private[this] def buildInfoBatchForPython(
```
revans2, line 154: Again some comments would be nice here.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExecBase.scala
```diff
@@ -78,96 +77,265 @@ abstract class GpuWindowInPandasExecMetaBase(
...
+      batch: ColumnarBatch,
+      gpLens: Seq[_]): ColumnarBatch = {
```
revans2, line 156: I am not happy with the type of `gpLens`, at a minimum we need to document exactly what is supported because it is not clear.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuWindowInPandasExecBase.scala
```diff
@@ -78,96 +77,265 @@ abstract class GpuWindowInPandasExecMetaBase(
...
+
+    // The rolling size may need tuning.
+    val rollingSize = 20000
```
revans2, line 159: Where exactly did the 20,000 come from? Also why are we using the number of groups? Why is this not related to the amount of data being sent or number of rows being sent instead of groups. It feels like it would be close to impossible to predict how much memory the python side would need to do the processing.

firestarman:
Thanks for review, this is now low priority. The response will be delayed .

GaryShen2008:
Since we're working on the other pandas udf type support, let's move it to 0.6 target.

jlowe:
This PR has not seen activity in a while -- is it still relevant?  As mentioned before, it must be retargeted to branch-0.6 if it is needed.  branch-0.5 is closed for everything except critical fixes.

firestarman:
Closing as it is not required now

Issue #2014, opened 2021-03-24
[BUG] earlyStart option added in 0.5 can cause errors when starting UCX


abellina:
@rongou reported this one, I'll put up a fix soon. Introduced with: https://github.com/NVIDIA/spark-rapids/pull/1891

```
21/03/24 12:21:44 ERROR SparkUncaughtExceptionHandler: Uncaught exception in thread Thread[ucx-connection-server-2,5,main]
java.nio.BufferUnderflowException
	at java.nio.DirectByteBuffer.get(DirectByteBuffer.java:270)
	at java.nio.HeapByteBuffer.put(HeapByteBuffer.java:223)
	at com.nvidia.spark.rapids.shuffle.ucx.UCXConnection$.$anonfun$writeHandshakeHeader$2(UCXConnection.scala:398)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
	at com.nvidia.spark.rapids.shuffle.ucx.UCXConnection$.writeHandshakeHeader(UCXConnection.scala:396)
	at com.nvidia.spark.rapids.shuffle.ucx.UCX.handleSocket(UCX.scala:460)
	at com.nvidia.spark.rapids.shuffle.ucx.UCX.$anonfun$startManagementPort$6(UCX.scala:227)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

PR #2015, opened 2021-03-24
Fix bug where rkey buffer is getting advanced after the first handshake

abellina:
Fixes https://github.com/NVIDIA/spark-rapids/issues/2014

The bug here is that the lazy val would get mutated per the `put` interface in `ByteBuffer`. I am not sure how I missed it but yeah, that's a big bug (introduced in 0.5 by me, see issue linked).

@rongou fyi, and thank you.


abellina:
build

rongou (approved):

Issue #305, opened 2020-06-29
[FEA] Support Effecient data transfers for python UDFs


revans2:
pyspark supports sending data using arrow in some cases.  Our data is already in arrow format.  We should look at what it would take to support taking our columnar data and sending it across the wire to a python process with as little transformations as possible. In some cases we might even be able to not pull all of the data back from the GPU when we do this.

The class we need to look at is `ArrowEvalPythonExec`.  We should also pay attention to how we do this because we could reuse this for R too if we wanted to.

revans2:
I filed https://github.com/rapidsai/cudf/issues/5657 to help this in cudf.

revans2:
This was updated to be https://github.com/rapidsai/cudf/issues/1352 after some discussion with the cudf team.  On the cudf java side we are going to have to get the c++ arrow classes and then use `arrow::ipc` to write them out in a form that we want.  we will probably have to provide our own implementations of `arrow::InputStream` and `arrow::OutputStream` to tie it all together. The main issue is that those APIs assume blocking I/O so we are going to have to do callbacks into java in some cases to make this work.

revans2:
It looks like the C++ code is getting closer to going in.  It is close enough to the 0.15 release of cudf that I am not 100% sure that I will be able to get the changes to JNI in on time.  The issue is that this code is going to be like one of the file formats for both reads and writes, but it is a bit more complicated because we need to be able to do reads and writes on a stream.  We already support this for writes, but we don't support it for reads yet, so this is likely to require some new API work.

What is more these APIs need to call back into java so we need to make sure that we get this all done very efficiently.
We are also going to interact with the C++ arrow APIs for the first time here, which is going to require some work.  We might need to wrap `arrow::Buffer` in a `HostMemoryBuffer` so we can pull data back and do what we need to with it.

In general the code looks fairly simple to implement.

We create an `ArrowIPCTableWriter`.  It would follow the same API that we have for parquet, and Orc. We might need to split `WriteOptions` up into another base class, because we only need the names of the columns.

The actual core part of the implementation would include something like...
Initialization:
```
std::shared_ptr<arrow::io::OutputStream> output_stream = ...;
```

write table:
```
std::shared_ptr<arrow::Table> arrow_tab = cudf::to_arrow(cudf_table_view);
if (!initialized) {
  arrow::Result<std::shared_ptr<RecordBatchWriter>> writer = arrow::io::ipc::NewWriter(arrow_tab->schema());
[...]

PR #2035, opened 2021-03-29
Accelerate data transfer for map Pandas UDF plan

firestarman:
This PR is to accelerate the data transfer between JVM and Python for the plan `GpuMapInPandas`, by implementing the columnar way.

Also add the related integration tests.

For issue #305


firestarman:
Will run some benchmark locally before marking ready for review.

firestarman:
build

firestarman:
> Will run some benchmark locally before marking ready for review.

Got some positive numbers and tests passed on databricks. Mark as ready for review.

integration_tests/src/main/python/udf_test.py
```diff
@@ -214,6 +213,22 @@ def pandas_filter(iterator):
             conf=arrow_udf_conf)
 
 
+@pytest.mark.parametrize('data_gen', data_gens_nested_for_udf, ids=idfn)
+def test_pandas_map_udf_nested_type(data_gen):
+    # Spark supports limited types as the return type of pandas UDF.
+    # For details please go to
+    #  https://github.com/apache/spark/blob/master/python/pyspark/sql/pandas/types.py#L28
+    # So we return the data frames with one column of integral type for all types of the input.
```
jlowe, line 221: The return types don't look _that_ limited.  It supports strings, arrays (as long as they don't contain structs or timestamps), structs that don't contain other structs, NullType, DecimalType, etc.  Looks like we're missing a lot of test cases here.

jlowe (commented):
I'll defer to @revans2 for reviewing the details of the iterator changes since he's much more familiar with that code, but it does seem like we could do a better job of testing the various type combinations.

integration_tests/src/main/python/udf_test.py
firestarman: Yes, @revans2  Should we have tests to cover all the output types supported by PySpark ?

firestarman:
> > Got some positive numbers
> 
> Can you elaborate here what those numbers were? What was the use-case and how much improvement did you see?

Not easy to show the numbers here, so here is [the link](https://docs.google.com/spreadsheets/d/1xO_-YalLsl1KIgEdksmBR4UEdmpk5SVNhkFNAuhmLDI/edit#gid=74620754). 

I just ran some benchmark for data transfer locally to verify the columnar way will get better perf,  and this PR is for the issue #305 .

integration_tests/src/main/python/udf_test.py
firestarman: I updated the test for the supported output types check.

firestarman:
Forced push only for the commit signature issue. Since the latest merge commit has no signature by default. I added it by a forced push.

firestarman:
build

integration_tests/src/main/python/udf_test.py
```diff
@@ -215,17 +215,38 @@ def pandas_filter(iterator):
...
 def test_pandas_map_udf_nested_type(data_gen):
-    # Spark supports limited types as the return type of pandas UDF.
-    # For details please go to
-    #  https://github.com/apache/spark/blob/master/python/pyspark/sql/pandas/types.py#L28
-    # So we return the data frames with one column of integral type for all types of the input.
-    def size_udf(pdf_itr):
+    # Supported UDF output types by plugin: (commonCudfTypes + ARRAY).nested() + STRUCT
+    # STRUCT represents the whole dataframe in Map Pandas UDF, so no struct column in UDF output.
+    # More details is here
+    #   https://github.com/apache/spark/blob/master/python/pyspark/sql/udf.py#L119
```
jlowe, line 221: That code seems to only indicate that the top-level type is expected to be a struct, not that it doesn't support struct-in-struct?  Poking around at Spark Pandas UDF examples, it looks like struct columns are supported as a result column.

firestarman: I verified it locally, and it failed when returning a struct column (even as a top-level column) from Map Pandas UDF.
I uploaded the test code to my repo, you can find it here: https://github.com/firestarman/spark-rapids/commit/02b5b4d4e4a2547fcc35cece04f79d7f34cc70b2 
```bash
        elif self.evalType == PythonEvalType.SQL_MAP_PANDAS_ITER_UDF:
            if isinstance(self._returnType_placeholder, StructType):
                try:
                    to_arrow_type(self._returnType_placeholder)
                except TypeError:
                    raise NotImplementedError(
                        "Invalid return type in mapInPandas: "
>                       "%s is not supported" % str(self._returnType_placeholder))
E                   NotImplementedError: Invalid return type in mapInPandas: StructType(List(StructField(a,StructType(List(StructField(child0,ArrayType(ShortType,true),true),StructField(child1,StructType(List(StructField(child0,ByteType,true),StructField(child1,ShortType,true),StructField(child2,IntegerType,true),StructField(child3,LongType,true),StructField(child4,FloatType,true),StructField(child5,DoubleType,true),StructField(child6,StringType,true),StructField(child7,BooleanType,true),StructField(child8,DateType,true),StructField(child9,TimestampType,true))),true))),true))) is not supported
```

Whether it supports non-nested struct column depends on the type of Pandas UDF, I think. 
e.g. the Scalar Pandas UDF should support top-level struct column, but Map Pandas UDF does not. Since the former one does not require the top-level type being a struct, while the later one does. 
Actually Map Pandas UDF requires returning a DataFrame from UDF, not a column. And the type of a DataFrame is `struct`.

jlowe (approved):

firestarman:
@revans2 Hi, Could you take a look at this PR ? Thanks in advance.

firestarman:
I am going to merge this, if any concern, I will make new PRs.

PR #2037, opened 2021-03-29
Fix warnings from use of deprecated cudf methods

jlowe:
Updated the code to use the preferred cudf methods which cleans up many of the deprecated warnings during the build.


jlowe:
build

revans2 (approved):

PR #2046, opened 2021-03-30
support casting Decimal to String

sperlingxx:
Current PR is to support casting Decimal to String, which is requested in issue #1625.


sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -419,6 +419,9 @@ case class GpuCast(
       case (from: DecimalType, to: DecimalType) =>
         castDecimalToDecimal(input.getBase, from, to)
 
+      case (_: DecimalType, StringType) =>
```
revans2, line 422: I think we have to update the `CastChecks` to allow casting from decimal to string for this to actually impact user queries.

sperlingxx: Sorry. I forgot it.

sperlingxx: Fixed.

sperlingxx:
build

sperlingxx:
build

sperlingxx:
build

tests/src/test/scala/com/nvidia/spark/rapids/GpuExpressionTestSuite.scala
```diff
@@ -172,6 +172,11 @@ abstract class GpuExpressionTestSuite extends SparkQueryCompareTestSuite {
     }
   }
 
+  def compareStringifiedDecimals(expected: String, actual: String): Boolean = {
+    (expected == null && actual == null) ||
+        (expected != null && actual != null && Decimal(expected) == Decimal(actual))
```
revans2, line 177: I thought we wanted to produce the exact same result.  If we don't produce the exact same result then at a minimum we need to document the incompatibility and if it is a big enough difference, then we need to have a config to disable it by default.

sperlingxx:
build

sperlingxx:
build

sperlingxx:
build

sperlingxx:
build

sperlingxx:
I think this PR is ready to merge.

jlowe (approved):

PR #2050, opened 2021-03-31
stream shuffle buffers from GDS to UCX

rongou:


sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
```diff
@@ -171,32 +171,8 @@ class BufferSendState(
...
-                  case md: DeviceMemoryBuffer =>
-                    // device original => device bounce
-                    d.copyFromDeviceBufferAsync(buffOffset, md, blockRange.rangeStart,
-                      blockRange.rangeSize(), serverStream)
-                  case _ => throw new IllegalStateException("What buffer is this")
-                }
-              case _ => throw new IllegalStateException("What buffer is this")
-            }
-          }
+          rapidsBuffer.copyToMemoryBuffer(
```
abellina, line 174: In the GDS scenario, we never spill to host memory, correct?  

Mostly just checking, because the bounce buffer picked here can be host in the non-GDS case, and that would trigger the exception for the GDS RapidsBuffer.

abellina (changes requested):
To me this seems to be a good change. `BufferReceiveState` now has copy code that `BufferSendState` doesn't, so they are a bit dissimilar, but I don't think it's the end of the world (and I'd much rather have the GDS details within the GDS `RapidsBuffer`)

The other comment is that the `CudaUtil` class should go away. I have a long standing task to take care of it in cuDF, and just noting that I'll need to follow this up by to remove `CudaUtil`, I don't think this PR changes that fact, just moves the code that needs some help to the `RapidsBuffer`.

sql-plugin/src/main/scala/ai/rapids/cudf/CudaUtil.scala
```diff
@@ -36,6 +36,25 @@ object CudaUtil {
...
+   * `length` bytes, in the default stream.
+   * @param src source buffer
+   * @param srcOffset source offset
+   * @param dst destination buffer
+   * @param dstOffset destination offset
+   * @param length amount to copy
+   */
+  def copyAsync(src: MemoryBuffer, srcOffset: Long, dst: MemoryBuffer,
+      dstOffset: Long, length: Long): Unit = {
+    Cuda.asyncMemcpy(
```
jlowe, line 51: Ideally this should do some bounds-checking to avoid difficult-to-debug issues if we go beyond the buffer extents. I'd actually prefer this functionality was moved to `ai.rapids.cudf.MemoryBuffer` in cudf, but it's fine for here in the interim.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -88,6 +87,22 @@ class RapidsGdsStore(
...
+        srcOffset: Long, dst: MemoryBuffer, dstOffset: Long, length: Long): Unit = {
+      val path = if (id.canShareDiskPaths) {
+        sharedBufferFiles.get(id)
+      } else {
+        id.getDiskPath(diskBlockManager)
+      }
+      dst match {
+        case _: DeviceMemoryBuffer =>
+          val dm = dst.asInstanceOf[DeviceMemoryBuffer].slice(dstOffset, length)
+          CuFile.readFileToDeviceBuffer(dm, path, fileOffset + srcOffset)
```
jlowe, line 100: Curious, what stream is used here?  It looks like it's default, but it's unclear whether that's per-thread default or system default.

abellina: nice catch, we should match the stream we are synchronizing on (the `BufferSendState` takes a stream)

abellina: @rongou this PR is still held by this right?

rongou: I asked the GDS team they said right now this is just synchronous. I added a TODO to switch to the async api when it becomes available.

rongou (commented):
Removed `CudaUtil` and fixed the tests. Please take another look.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBuffer.scala
```diff
@@ -115,6 +115,17 @@ trait RapidsBuffer extends AutoCloseable {
    */
   def getMemoryBuffer: MemoryBuffer
 
+  /**
+   * Copy the content of this buffer into the specified memory buffer, starting from the given
+   * offset.
+   * @param srcOffset offset to start copying from.
+   * @param dst the memory buffer to copy into.
+   * @param dstOffset offset to copy into.
+   * @param length number of bytes to copy.
```
abellina, line 124: missing stream in scaladoc

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -334,6 +334,19 @@ abstract class RapidsBufferStore(
       }
     }
 
+    override def copyToMemoryBuffer(srcOffset: Long, dst: MemoryBuffer, dstOffset: Long,
+        length: Long, stream: Cuda.Stream): Unit = {
+      withResource(getMemoryBuffer) { memBuff =>
+        dst match {
+          case _: HostMemoryBuffer =>
+            dst.copyFromMemoryBuffer(dstOffset, memBuff, srcOffset, length, stream)
```
abellina, line 342: Should this just use the `Async` version? Especially if `dst` is a pinned buffer, I believe that pinned copy can be done asynchronously.

tests/src/test/scala/org/apache/spark/sql/rapids/SpillableColumnarBatchSuite.scala
```diff
@@ -18,7 +18,7 @@ package org.apache.spark.sql.rapids
```
abellina, line 18: this file needs 2021 copyrights

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleServerSuite.scala
```diff
@@ -16,10 +16,11 @@
```
abellina, line 16: 2021 copyright

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
```diff
@@ -16,7 +16,7 @@
```
abellina, line 16: 2021 copyright.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferReceiveState.scala
```diff
@@ -18,7 +18,7 @@ package com.nvidia.spark.rapids.shuffle
```
abellina, line 18: 2021 copyright

abellina (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBuffer.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
rongou: In the original code `BufferSendState` was doing a `CudaUtil.copy` instead of `copyAsync`, so I wanted to preserve the existing behavior. We can certainly change it here if you think it is appropriate.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferReceiveState.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
rongou: Done.

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleServerSuite.scala
rongou: Done.

tests/src/test/scala/org/apache/spark/sql/rapids/SpillableColumnarBatchSuite.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
abellina: OK, maybe it can be a follow up to keep changes to a minimum. Agree with it not being part of this PR then.

abellina: Can you add a TODO here please?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
abellina: This seems fine to me for now.

abellina (approved):

abellina:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -86,6 +86,23 @@ class RapidsGdsStore(
...
+      } else {
+        id.getDiskPath(diskBlockManager)
+      }
+      dst match {
+        case _: DeviceMemoryBuffer =>
+          val dm = dst.asInstanceOf[DeviceMemoryBuffer].slice(dstOffset, length)
+          // TODO: switch to async API when it's released, using the passed in CUDA stream.
+          CuFile.readFileToDeviceBuffer(dm, path, fileOffset + srcOffset)
+          logDebug(s"Created device buffer for $path $fileOffset:$size via GDS")
+        case _ => throw new IllegalStateException("GDS can only copy to device buffer, not $dst")
```
gerashegalov, line 102: interpolation `s` omitted. dst.getClass might be safer since toString might have some side effect.

gerashegalov (changes requested):

gerashegalov:
with the current PR description it sounds like this should be a draft PR [WIP]

rongou:
@gerashegalov removed the initial description. This should be considered ready now.

rongou:
build

gerashegalov:
```bash
$ TEST_PARALLEL=1 SPARK_HOME=~/spark-3.1.1-bin-hadoop3.2 \
  integration_tests/run_pyspark_from_build.sh -k test_single_nested_orderby

========= 20 passed, 5383 deselected, 100 xfailed in 188.11s (0:03:08) =========
```

abellina (approved):

gerashegalov (approved):
LGTM

PR #2055, opened 2021-03-31
Support PivotFirst

nartal1:
This PR is to support PivotFirst. 
Have added integration tests for all the types supported by Pivot.


nartal1:
build

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -181,6 +181,7 @@ def get_params(init_list, marked_params=[]):
 _init_list_no_nans_with_decimal = _init_list_no_nans + [
     _grpkey_small_decimals]
 
+
```
abellina, line 184: extra line, nit

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -223,6 +224,20 @@ def test_hash_avg_nulls_partial_only(data_gen):
...
+@pytest.mark.parametrize('conf', get_params(_confs, params_markers_for_confs), ids=idfn)
+def test_hash_grpby_pivot(data_gen, conf):
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: gen_df(spark, data_gen, length=100)
+            .groupby('a')
+            .pivot('b')
+            .agg(f.sum('c')),
+            conf=conf)
+
+
```
abellina, line 240: extra line nit

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1847,6 +1847,26 @@ object GpuOverrides {
         override def convertToGpu(): Expression =
           sortOrder.withNewChildren(childExprs.map(_.convertToGpu()))
       }),
+    expr[PivotFirst](
+      "Pivot first operator",
```
abellina, line 1851:
```suggestion
      "PivotFirst operator",
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -810,6 +810,19 @@ abstract class AggExprMeta[INPUT <: AggregateFunction](
...
 
+abstract class ImperativeAggExprMeta[INPUT <: ImperativeAggregate](
+    expr: INPUT,
+    conf: RapidsConf,
+    parent: Option[RapidsMeta[_, _, _]],
+    rule: DataFromReplacementRule)
+    extends ExprMeta[INPUT](expr, conf, parent, rule) {
+
+  override final def convertToGpu(): GpuExpression =
+    convertToGpu(childExprs(0).convertToGpu(),childExprs(1).convertToGpu())
```
abellina, line 821: @nartal1 can you double check that all `ImperativeAggregate`s have two children?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -810,6 +810,19 @@ abstract class AggExprMeta[INPUT <: AggregateFunction](
   def convertToGpu(child: Expression): GpuExpression
 }
 
+abstract class ImperativeAggExprMeta[INPUT <: ImperativeAggregate](
```
abellina, line 813: need a class header.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -375,6 +375,55 @@ case class GpuSum(child: Expression, resultType: DataType)
     Aggregation.sum().onColumn(inputs.head._2)
 }
 
+case class GpuPivotFirst(
```
abellina, line 378: class header.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -375,6 +375,55 @@ case class GpuSum(child: Expression, resultType: DataType)
...
+  pivotColumn: Expression,
+  valueColumn: Expression,
+  pivotColumnValues: Seq[Any],
+  mutableAggBufferOffset: Int = 0,
+  inputAggBufferOffset: Int = 0) extends GpuDeclarativeAggregate {
```
abellina, line 383: so this extends `GpuDeclarativeAggregate`, but the meta expression is for imperative ones. Which seems odd. If both imperative and declarative Spark CPU aggregates can be implemented by our same interface (so far anyway), perhaps we should rename ours `GpuAggregate` or something like that and use it for both kinds.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -375,6 +375,55 @@ case class GpuSum(child: Expression, resultType: DataType)
...
+
+  val valueDataType = valueColumn.dataType
+
+  override val dataType: DataType = valueDataType
+  override val nullable: Boolean = false
+
+  val pivotColAttr = pivotColumnValues.map(a => {
+    // If `a` is null, then create an AttributeReference for null column.
+    if (a == null) AttributeReference(GpuLiteral(null, valueDataType).toString, valueDataType)()
```
abellina, line 392: lets expand this block out and add curly brackets: 
```
if () {
} else {
}
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -375,6 +375,55 @@ case class GpuSum(child: Expression, resultType: DataType)
...
+    else AttributeReference(a.toString, valueDataType)()
+  })
+
+  override lazy val inputProjection: Seq[Expression] = {
+    val expr = pivotColumnValues.map(a => {
+      if (a == null) {
```
abellina, line 398: `a` -> something like `pivotValue` or `pivotColumnValue`.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -375,6 +375,55 @@ case class GpuSum(child: Expression, resultType: DataType)
...
+      } else {
+        GpuIf(GpuEqualTo(pivotColumn, GpuLiteral(a, pivotColumn.dataType)),
+          valueColumn, GpuLiteral(null, valueDataType))
+      }
+    })
+    expr
+  }
+
+  override lazy val updateExpressions: Seq[GpuExpression] = {
+    val updateArray = pivotColAttr.map(a => new CudfLastExcludeNulls(a))
```
abellina, line 409: Could you define in this aggregate's class header what the pivot looks like at each stage? I don't think we need to rehash everything that Spark has, but something quick showing why `CudfLastExcludeNulls` works here and in `mergeExpressions`.

abellina (changes requested):
Very quick first pass. I'll do another one today or Monday for sure.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
nartal1: Thanks @abellina  for catching this. All ImperativeAggregates don't have two children. I have updated the code.

nartal1:
build

abellina (changes requested):
@nartal1 I checked out your copy and started running some pivot queries. I looked at (https://github.com/apache/spark/blob/0494dc90af48ce7da0625485a4dc6917a244d580/sql/core/src/test/resources/sql-tests/inputs/pivot.sql). I think a lot of these are working, but I do see some that are causing plugin (and/or cudf) specific errors. I think it's a coding error in the plugin, and it's worth looking at.

Here are a couple of examples of queries that are failing. I think this points at there needs to be more testing for pivot first, manually and in the integration tests/scala tests as much as possible. 

```
SELECT * FROM (
  SELECT course, earnings FROM courseSales
)
PIVOT (
  sum(earnings)
  FOR course IN ('dotNET', 'Java')
);
```

Note in this case I have 2012 in there twice:

```
SELECT * FROM courseSales
PIVOT (
  sum(earnings)
  FOR year IN (2012,2013,2012)
);
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -32,7 +32,7 @@ import org.apache.spark.sql.catalyst.plans.physical.{AllTuples, ClusteredDistrib
 import org.apache.spark.sql.catalyst.util.truncatedString
 import org.apache.spark.sql.execution.{ExplainUtils, SortExec, SparkPlan, UnaryExecNode}
 import org.apache.spark.sql.execution.aggregate.{HashAggregateExec, SortAggregateExec}
-import org.apache.spark.sql.rapids.{CudfAggregate, GpuAggregateExpression, GpuDeclarativeAggregate}
+import org.apache.spark.sql.rapids.{CudfAggregate, GpuAggregate, GpuAggregateExpression}
```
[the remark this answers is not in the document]
abellina: BTW we already have `GpuAggregateFunction`... so combining/refactoring if it makes sense. We can discuss offline if you like.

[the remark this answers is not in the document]
nartal1: Sure Alessandro. We can discuss offline about refactoring. I am currently looking into the test failures which is mentioned above. Thanks for catching those.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1847,6 +1847,33 @@ object GpuOverrides {
         override def convertToGpu(): Expression =
           sortOrder.withNewChildren(childExprs.map(_.convertToGpu()))
       }),
+    expr[PivotFirst](
+      "PivotFirst operator",
+      ExprChecks.fullAgg(
```
revans2, line 1852: `fullAgg` indicates that it can work as a window function, and as a reduction too. I am 99% sure that Spark does not support pivot for window operations, and 100% sure that we cannot. I am less so about reductions.  Could you please update this so that only the operations that we support are included? I don't think that there is a helper function of this. They all assume that window operations are supported in Spark for the given aggregation.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -286,7 +286,7 @@ abstract class GpuDeclarativeAggregate extends GpuAggregateFunction with GpuUnev
 
   // mostly likely a pass through (count => sum we merged above).
   // average has a more interesting expression to compute the division of sum/count
-  val evaluateExpression: Expression
+  val evaluateExpression: Seq[Expression]
```
revans2, line 289: I think we need probably should rename this to make it plural. `evaluateExpressions`

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -375,15 +375,109 @@ case class GpuSum(child: Expression, resultType: DataType)
...
+      AttributeReference(GpuLiteral(null, valueDataType).toString, valueDataType)()
+    } else {
+      AttributeReference(pivotColumnValue.toString, valueDataType)()
+    }
+  })
+
+  override lazy val inputProjection: Seq[Expression] = {
+    val expr = pivotColumnValues.map(pivotColumnValue => {
+      if (pivotColumnValue == null) {
+        GpuIf(GpuIsNull(pivotColumn), valueColumn, GpuLiteral(null, valueDataType))
```
revans2, line 443: GpuLiteral does not support arrays, but in the function signature it looks like we support single level arrays. How are we protecting ourselves from valueDataType being an array?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1847,6 +1847,33 @@ object GpuOverrides {
...
+        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL +
+          TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL),
+        TypeSig.all,
+        Seq(ParamCheck(
+          "aggFunc",
```
revans2, line 1857: I think this is "pivotColumn" not "aggFunc"

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1847,6 +1847,33 @@ object GpuOverrides {
...
+          TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL +
+            TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL),
+          TypeSig.all)),
+        Some(RepeatingParamCheck("filter",
```
revans2, line 1861: I don't think we have a repeating parameter called filter. I think there is just a second expression called "valueColumn"

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -375,15 +375,109 @@ case class GpuSum(child: Expression, resultType: DataType)
...
+          valueColumn, GpuLiteral(null, valueDataType))
+      }
+    })
+    expr
+  }
+
+  override lazy val updateExpressions: Seq[GpuExpression] = {
+    val updateArray = pivotColAttr.map(
+      pivotColumnValue => new CudfLastExcludeNulls(pivotColumnValue))
+    GpuCreateArray(updateArray, false) :: Nil
```
revans2, line 455: Why is this here? it looks like it is dead code.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
nartal1: I did an explain on a simple query which includes an aggregate(sum) and pivot on a column. At present we dont support arrays for aggregations(sum, max etc). 
The explain of this query produced aggregations on CPU and did a RowToColumnar for pivot. 
For `valueDataType` being array, it throws the exception:
 `class org.apache.spark.sql.catalyst.util.GenericArrayData '[1,1]' is not supported as a scalar yet` .
 I don't think we can support this currently. So changed the function signature. 
 Please let me know if it's okay.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
nartal1: Sorry it's remanent from debugging. Removed it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
nartal1: Thanks Bobby. I have updated this for only groupBy aggregates. For reduction, it fails because in the second stage of aggregation, we groupBy grouping columns and rearrange the pivotColumnValues into separate columns.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1847,6 +1847,36 @@ object GpuOverrides {
         override def convertToGpu(): Expression =
           sortOrder.withNewChildren(childExprs.map(_.convertToGpu()))
       }),
+    expr[PivotFirst](
+      "PivotFirst operator",
+      ExprChecks.groupByAgg(
```
revans2, line 1852: So I was able to make PivotFirst work as a reduction (no keys in the HashAggregate). Spark does not make it simple/obvious how to do it but.

```
scala> df.groupBy().pivot("a").max("b").show
+---+---+---+---+
|  0|  1| 12|123|
+---+---+---+---+
|  7|  6|  8|  5|
+---+---+---+---+


scala> df.groupBy().pivot("a").max("b").explain
== Physical Plan ==
HashAggregate(keys=[], functions=[pivotfirst(a#2519, max(`b`)#6016, 0, 1, 12, 123, 0, 0)])
+- Exchange SinglePartition, true, [id=#309]
   +- HashAggregate(keys=[], functions=[partial_pivotfirst(a#2519, max(`b`)#6016, 0, 1, 12, 123, 0, 0)])
      +- *(2) HashAggregate(keys=[a#2519], functions=[max(b#2520)])
         +- Exchange hashpartitioning(a#2519, 200), true, [id=#304]
            +- *(1) HashAggregate(keys=[a#2519], functions=[partial_max(b#2520)])
               +- *(1) LocalTableScan [a#2519, b#2520]
```

I think it might be time to have a bit more complex API, that might end up being a bit cleaner in the long run probably using a builder to be able to set 

```
def aggBuilder(
      outputCheck: TypeSig,
      sparkOutputSig: TypeSig,
      paramCheck: Seq[ParamCheck] = Seq.empty,
      repeatingParamCheck: Option[RepeatingParamCheck] = None): ExprChecksBuilder = {
 ...
}

class ExprChecksBuilder(...) {
    def rapidsAndSparkSupportsGroupBy ...
    def rapidsAndSparkSupportsReduction ...
    def sparkSupportsGroupByAndReduction ...
    def rapidsAndSparkSupportsWindow ...
    def sparkSupportsWindow ...
    def build(): ExprChecks
}
      
```

Otherwise if we keep doing what we have been doing we will have 2^5 different aggregation APIs for type checking.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1847,6 +1847,36 @@ object GpuOverrides {
...
+        override def tagExprForGpu(): Unit = {
+          if (conf.hasNans &&
+            (pivot.pivotColumn.dataType.equals(FloatType) ||
+              pivot.pivotColumn.dataType.equals(DoubleType))) {
+            willNotWorkOnGpu("Pivot expressions over floating point columns " +
+              "that may contain NaN is disabled. You can bypass this by setting " +
+              s"${RapidsConf.HAS_NANS}=false")
+          }
+          // If pivotColumnValues doesn't have distinct values, follow back to CPU
+          if (pivot.pivotColumnValues.distinct.lengthCompare(pivot.pivotColumnValues.length) != 0) {
```
revans2, line 1873: Did this show up in testing? I don't see how this would ever happen as it would cause issues in the CPU code too.

nartal1: Yes, it showed up in testing. The result is null for all non-distinct entries in CPU.

```
spark.sql("""SELECT * FROM courseSales
PIVOT (
  sum(earnings)
  FOR year IN (2012, 2013,2012)
)""").show

+------+----+-----+----+                                                        
|course|2012| 2013|2012|
+------+----+-----+----+
|dotNET|null|48000|null|
|  Java|null|30000|null|
+------+----+-----+----+
```
In plugin, it throws cudfException:
```
  Caused by: ai.rapids.cudf.CudfException: vector::_M_range_check: __n (which is 3) >= this->size() (which is 3)
  at ai.rapids.cudf.Table.groupByAggregate(Native Method)
  at ai.rapids.cudf.Table.access$2900(Table.java:45)
  at ai.rapids.cudf.Table$AggregateOperation.aggregate(Table.java:2377)
  at com.nvidia.spark.rapids.GpuHashAggregateExec.computeAggregate(aggregate.scala:821)
  at com.nvidia.spark.rapids.GpuHashAggregateExec.$anonfun$doExecuteColumnar$1(aggregate.scala:431)
```

If we remove additional `2012` from the query, we get correct result for both CPU and GPU:
```
+------+-----+-----+
|course| 2012| 2013|
+------+-----+-----+
|dotNET|15000|48000|
|  Java|20000|30000|
+------+-----+-----+
```

revans2: hmm, that is really interesting.  We could definitely match that same functionality, but why would anyone want that??? Falling back to the CPU feels like the right thing to do.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
revans2: You removed the array from the types supported by the Type checks so now this looks fine.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
nartal1: Thanks, @revans2. Updated the PR to support queries with reduction only on Pivot. Added integration tests as well. 
I have created a follow on issue to refactor ExprChecks https://github.com/NVIDIA/spark-rapids/issues/2152.
PTAL.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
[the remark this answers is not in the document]
nartal1: @abellina , I have combined GpuAggregate with GpuAggregateFunction. PTAL.

nartal1:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1848,6 +1848,36 @@ object GpuOverrides {
...
+          if (conf.hasNans &&
+            (pivot.pivotColumn.dataType.equals(FloatType) ||
+              pivot.pivotColumn.dataType.equals(DoubleType))) {
+            willNotWorkOnGpu("Pivot expressions over floating point columns " +
+              "that may contain NaN is disabled. You can bypass this by setting " +
+              s"${RapidsConf.HAS_NANS}=false")
+          }
+          // If pivotColumnValues doesn't have distinct values, fall back to CPU
+          if (pivot.pivotColumnValues.distinct.lengthCompare(pivot.pivotColumnValues.length) != 0) {
+            willNotWorkOnGpu("PivotFirst doesnt work on non distinct")
```
revans2, line 1875:
```
willNotWorkOnGpu("PivotFirst does not work on non-distinct values")
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -1005,6 +1005,22 @@ object ExprChecks {
       Seq(ParamCheck(param1._1, param1._2, param1._3),
         ParamCheck(param2._1, param2._2, param2._3)))
 
+  /**
+   * Aggregate operation where only group by agg and reduction is supported in the plugin and in
+   * Spark.
+   */
+
```
revans2, line 1012: nit: extra line

revans2 (dismissed):
Really just down to some nits. It is looking really good.

abellina (changes requested):

nartal1:
build

nartal1:
@abellina @revans2 I think I have addressed remaining nits/review comments. PTAL.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -375,15 +366,106 @@ case class GpuSum(child: Expression, resultType: DataType)
...
+
+  override lazy val updateExpressions: Seq[GpuExpression] = {
+    pivotColAttr.map(pivotColumnValue => new CudfLastExcludeNulls(pivotColumnValue))
+  }
+
+  override lazy val mergeExpressions: Seq[GpuExpression] = {
+    pivotColAttr.map(pivotColumnValue => new CudfLastExcludeNulls(pivotColumnValue))
+  }
+
+  override lazy val evaluateExpressions: Seq[Expression] = {
```
abellina, line 449: So does this need to be a `Seq` still?

abellina (changes requested):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
abellina: I think it did at some point, but doesn't seem to anymore, and could remove some diffs.

nartal1: You are right. It's not required anymore.  I have removed it now.

abellina (approved):

abellina:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1883,6 +1883,36 @@ object GpuOverrides {
...
+          if (conf.hasNans &&
+            (pivot.pivotColumn.dataType.equals(FloatType) ||
+              pivot.pivotColumn.dataType.equals(DoubleType))) {
+            willNotWorkOnGpu("Pivot expressions over floating point columns " +
+              "that may contain NaN is disabled. You can bypass this by setting " +
+              s"${RapidsConf.HAS_NANS}=false")
+          }
+          // If pivotColumnValues doesn't have distinct values, fall back to CPU
+          if (pivot.pivotColumnValues.distinct.lengthCompare(pivot.pivotColumnValues.length) != 0) {
+            willNotWorkOnGpu("PivotFirst does not work on non-distinct values")
```
revans2, line 1910: nit: I think it would be better if we clarified a but, just because distinct is a key work in SQL that has a different meaning. Perhaps something more like.

```
willNotWorkOnGpu("PivotFirst does not work on the GPU when there are duplicate pivot values provided")
```

Sorry I didn't put this in the first time when I looked at it.

revans2 (approved):
Just one minor nit

nartal1:
build

revans2 (approved):

abellina (approved):

nartal1:
Thanks @revans2 and @abellina

Issue #1341, opened 2020-12-09
Optimize unnecessary columnar->row->columnar transitions with AQE


jlowe:
**Is your feature request related to a problem? Please describe.**
Sometimes AQE processing can lead to a query plan with what appears to be a degenerate columnar->row->columnar transition like this:
![image](https://user-images.githubusercontent.com/1360766/101564968-b4207800-3991-11eb-9de8-4d1dc544d778.png)

It would be great if this wasteful transition between formats could be avoided, keeping the data on the GPU throughout.

**Describe the solution you'd like**
Plugin transition overrides recognizes this situation and removes unnecessary transitions when possible.

viadea:
Here is one minimum reproduce which can trigger similar issue when AQE is on:

```
val df = Seq(1).toDF()
df.write.format("parquet").mode("overwrite").save("/tmp/testaqe")
spark.read.parquet("/tmp/testaqe").createOrReplaceTempView("df")
# No GpuColumnarToRow or GpuRowToColumnar
val out=spark.sql("select value from df group by value")
out.write.format("parquet").mode("overwrite").save("/tmp/testaqe_out")
 
# When ACQ is on, one pair of GpuColumnarToRow or GpuRowToColumnar is introduced.
spark.conf.set("spark.sql.adaptive.enabled", true)
out.write.format("parquet").mode("overwrite").save("/tmp/testaqe_out")
```

Basically to trigger this I need to make sure all of below 3 requirements are met:

- AQE is on
- There is aggregation before writing to parquet files
- Write to parquet files eventually

andygrove:
I've been investigating this and I think it is pretty likely that we will want to modify Spark by adding the following methods to AdaptiveSparkPlanExec:

```scala
override def supportsColumnar: Boolean = initialPlan.supportsColumnar

override protected def doExecuteColumnar(): RDD[ColumnarBatch] = {
  val rdd = getFinalPhysicalPlan().executeColumnar()
  finalPlanUpdate
  rdd
}
```

I am working on getting one example working end to end based on this.

andygrove:
The `supportsColumnar` is tricky because we need it to be based on the final physical plan rather than the initial plan but we don't want a call to this method to trigger creating the final plan. I'll know more once I get the initial test working.

andygrove:
I was able to get the test case to work by adding this method to Spark's AdaptiveSparkPlanExec:

```scala
[...]

PR #2064, opened 2021-04-01
Optimize unnecessary columnar->row->columnar transitions with AQE

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/1341


andygrove:
build

tgravescs:
note we should have an issue to see if we can change Spark itself to make this better.

jlowe (commented):
I was very surprised to see only changes in GpuFileFormatWriter for this, and it seems like this isn't a general solution to the problem.  Yes, it may resolve the specific case described in the issue, but what if the plan was a filter or project then write to Parquet instead of just writing to Parquet?  Would we have to update FilterExec and/or ProjectExec as well to cover  those cases?

tgravescs:
so that is one thing I'm still trying to fully understand about AQE to make sure I'm not missing some cases, but at the surface it appears that won't happen, the filter would get rolled into the AdaptiveSparkPlan, the only thing spark doesn't wrap are writes.

andygrove:
> I was very surprised to see only changes in GpuFileFormatWriter for this, and it seems like this isn't a general solution to the problem. Yes, it may resolve the specific case described in the issue, but what if the plan was a filter or project then write to Parquet instead of just writing to Parquet? Would we have to update FilterExec and/or ProjectExec as well to cover those cases?

My understanding is that the issue only happens when we have an adaptive query where we are writing the results out to disk (rather than calling collect). If there are FilterExec or ProjectExec involved then those would be part of the adaptive query being executed, unless there is some filter or projection push-down that happens outside of that?

jlowe:
I'm still surprised that this workaround is in a specific operator like this rather than the plan being updated explicitly in something like GpuTransitionOverrides.  Seems like a generalized solution to me if the transition overrides, which already deal with AQE transitions, looks for the case where there's an AdaptiveSparkPlan that is really columnar followed by a columnar operation (after the necessary row->column conversion).  If it finds it, it removes the row conversion and puts in one of these adaptive fixup projects.  Then we don't care that it's specifically for a write, it should work for any case where Spark wants to do this -- or am I missing something?

andygrove:
The problem is that Spark treats this as two separate queries. There is the adaptive query that gets executed (and we don't know during planning or execution if this is being called from `collect` or `write`), and then after that, we have the write operation.

andygrove:
It is possible that we can insert this new AvoidTransitions operator during the planning of the write though ... I will look at that.

andygrove:
Ok, I understand the concerns better now. `GpuFileFormatWriter` receives a plan which is `GpuRowToColumnar(AdaptiveSparkPlan(..))` and has logic for wrapping this with `GpuProjectExec` and `GpuSort` in some cases and now we have additional logic that inserts the `AvoidTransition` operator as well, and this wasn't going to work for the projection case so I'll push a change to fix that once I figure out how to add a unit test.

However, do we also want to try and pull this runtime query plan manipulation out into rules as well as part of this PR?

andygrove:
Disregard my last comment .. I get it now, and yes this can be handled in GpuOverrides for `DataWritingCommandExec`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2636,9 +2636,14 @@ object GpuOverrides {
...
-        override def convertToGpu(): GpuExec =
-          GpuDataWritingCommandExec(childDataWriteCmds.head.convertToGpu(),
-            childPlans.head.convertIfNeeded())
+        override def convertToGpu(): GpuExec = {
+          // When the input is an adaptive plan we do not get to see the GPU version until
+          // the plan is executed and sometimes the plan will have a GpuColumnarToRowExec as the
+          // final operator and we can bypass this to keep the data columnar by inserting
+          // the [[AvoidAdaptiveTransitionToRow]] operator here
+          val child = AvoidAdaptiveTransitionToRow(childPlans.head.convertIfNeeded())
+          GpuDataWritingCommandExec(childDataWriteCmds.head.convertToGpu(), child)
```
jlowe, line 2645: I still don't get why this logic is tied to this specific exec.  The transition overrides should run after we've replaced the normal execs we're going to replace, like `DataWritingCommandExec`.  The transition override should not be looking for specific GPU exec nodes, but rather the sequence of a `GpuRowToColumnarExec` with a child `AdaptiveSparkPlan` having a final physical plan with a trailing `GpuColumnarToRowExec` in it.  If it finds it, it should replace the `GpuColumnarToRowExec` with the new `AvoidAdaptiveTransitionToRow` exec.  This should work whether the node after the `GpuColumnarToRowExec` is a GPU write or any other GPU exec node and thus is more robust if Spark changes where it wants to place adaptive plans.

andygrove: Thanks, that makes sense. We can't check for the final physical plan of the `AdaptiveSparkPlanExec` at this stage, because it hasn't been created yet, but we can look for `GpuRowToColumnarExec` wrapping `AdaptiveSparkPlanExec` and insert `AvoidAdaptiveTransitionToRow`. I'll update this.

tests/src/test/scala/com/nvidia/spark/rapids/AdaptiveQueryExecSuite.scala
```diff
@@ -294,6 +297,81 @@ class AdaptiveQueryExecSuite
     }, conf)
   }
 
+  test("Avoid transitions to row when writing to Parquet") {
```
jlowe, line 300: It would be good to add a negative test where an adaptive plan appears but it is _not_ a columnar plan and therefore we should not remove the row-to-columnar transition because it is necessary.

jlowe (commented):
Thanks, @andygrove!  This is along the lines of what I expected.  Looks OK to me, although I'd like to see a test for the case when we shouldn't be removing the columnar transition.

andygrove:
The changes in this PR cause regressions when running the test suite against Spark 3.1.1 and I am working on debugging that now.

andygrove:
build

andygrove:
Build failed with the same issue I have been seeing locally when testing with Spark 3.1.1

```
12:19:45  [32mSortExecSuite:[0m
12:19:47  [31m- IGNORE ORDER, WITH DECIMALS: join longs *** FAILED ***[0m
12:19:47  [31m  org.apache.spark.SparkException: Job aborted due to stage failure: Task 77 in stage 475.0 failed 1 times, most recent failure: Lost task 77.0 in stage 475.0 (TID 685) (premerge-test-jenkins-rapids-premerge-github-1290-dgblq-0qrfv executor driver): java.lang.NullPointerException[0m
12:19:47  [31m	at com.nvidia.spark.rapids.RapidsBufferCatalog$.addTable(RapidsBufferCatalog.scala:200)[0m
```

andygrove:
build

andygrove:
@jlowe @tgravescs Tests are now passing and this is ready for re-review.

The test failures were somehow caused by adding a new SparkListener to the Spark context. This caused the Spark session to leak when running against Spark 3.1.1. Switching over to the `ExecutionPlanCaptureCallback` provided by our test suite resolved the issue.

jlowe (approved):

tgravescs (approved):

PR #2072, opened 2021-04-01
Disable CSV parsing by default and update tests to better show what is left

revans2:
This mitigates some issues with CSV parsing by disabling it by default. To really fix things we are likely going to need to fix some things with CSV parsing of strings and then work on modified versions of the code that we use to cast from string to various types.


revans2:
build

docs/compatibility.md
```diff
@@ -103,16 +103,41 @@ will produce a different result compared to the plugin.
...
+ * [spark.rapids.sql.csv.read.date.enabled](configs.md#sql.csv.read.date.enabled)
+ * [spark.rapids.sql.csv.read.double.enabled](configs.md#sql.csv.read.double.enabled)
+ * [spark.rapids.sql.csv.read.float.enabled](configs.md#sql.csv.read.float.enabled)
+ * [spark.rapids.sql.csv.read.integer.enabled](configs.md#sql.csv.read.integer.enabled)
+ * [spark.rapids.sql.csv.read.long.enabled](configs.md#sql.csv.read.long.enabled)
+ * [spark.rapids.sql.csv.read.short.enabled](configs.md#sql.csv.read.short.enabled)
+ * [spark.rapids.sql.csvTimestamps.enabled](configs.md#sql.csvTimestamps.enabled)
+
+If you know that your particular data type will be parsed correctly enough, you may enable each
+type you expect to use. Often the performance improvement is so good that it is worth dealing
```
jlowe, line 120:
```suggestion
type you expect to use. Often the performance improvement is so good that it is worth
```

docs/compatibility.md
```diff
@@ -103,16 +103,41 @@ will produce a different result compared to the plugin.
...
+checking if it is parsed correctly.
+
+Spark is generally very strict when reading CSV and if the data does not conform with the 
+expected format exactly it will result in a `null` value. The underlying parser that the SQL plugin
```
jlowe, line 124:
```suggestion
expected format exactly it will result in a `null` value. The underlying parser that the RAPIDS Accelerator
```

docs/compatibility.md
```diff
@@ -142,6 +167,11 @@ disable itself if you set a format that it does not support.
 
 As a work around you can parse the column as a timestamp and then cast it to a date.
 
+Invalid dates, ones that have the correct format, but the numbers produce invalid dates, in Spark
+can result in an exception by default and how they are parsed can be controlled through a config.
```
jlowe, line 171:
```suggestion
Invalid dates in Spark, values that have the correct format but the numbers produce invalid dates,
can result in an exception by default, and how they are parsed can be controlled through a config.
```

docs/compatibility.md
```diff
@@ -142,6 +167,11 @@ disable itself if you set a format that it does not support.
...
+The plugin does not support any of this and will produce an incorrect date.  Typically, one that
```
jlowe, line 172:
```suggestion
The RAPIDS Accelerator does not support any of this and will produce an incorrect date.  Typically, one that
```

docs/compatibility.md
```diff
@@ -163,9 +193,14 @@ portion followed by one of the following formats:
 Just like with dates all timestamp formats are actually supported at the same time.  The plugin will
 disable itself if it sees a format it cannot support.
 
+Invalid timestamps, ones that have the correct format, but the numbers produce invalid dates or
```
jlowe, line 196: Same rewording suggestions as above

docs/compatibility.md
```diff
@@ -174,6 +209,10 @@ Some floating-point values also appear to overflow but do not for the CPU as des
 
 Any number that overflows will not be turned into a null value.
 
+Also parsing of some values will not produce bit for bit identical results to what the CPU does.
+they are within round-off errors except when they are close enough to overflow to Inf or -Inf which
```
jlowe, line 213:
```suggestion
They are within round-off errors except when they are close enough to overflow to Inf or -Inf which
```

docs/compatibility.md
```diff
@@ -174,6 +209,10 @@ Some floating-point values also appear to overflow but do not for the CPU as des
...
+then results in a number being returned when the CPU would have returns null.
```
jlowe, line 214:
```suggestion
then results in a number being returned when the CPU would have returned null.
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -700,6 +693,60 @@ object RapidsConf {
...
+  // TODO should we change this config?
+  val ENABLE_CSV_TIMESTAMPS = conf("spark.rapids.sql.csvTimestamps.enabled")
+      .doc("When set to true, enables the CSV parser to read timestamps. The default output " +
+          "format for Spark includes a timezone at the end. Anything except the UTC timezone is " +
+          "not supported. Timestamps after 2038 and before 1902 are also not supported.")
+      .booleanConf
+      .createWithDefault(false)
+
+  val ENABLE_READ_CSV_DATES = conf("spark.rapids.sql.csv.read.date.enabled")
+      .doc("Parsing invalid CSV dates produces different results from spark")
```
jlowe, line 705:
```suggestion
      .doc("Parsing invalid CSV dates produces different results from Spark")
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -700,6 +693,60 @@ object RapidsConf {
...
+      .booleanConf
+      .createWithDefault(false)
+
+  val ENABLE_READ_CSV_BOOLS = conf("spark.rapids.sql.csv.read.bool.enabled")
+      .doc("Parsing invalid CSV boolean values produce true instead of null")
```
jlowe, line 710:
```suggestion
      .doc("Parsing an invalid CSV boolean value produces true instead of null")
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -700,6 +693,60 @@ object RapidsConf {
...
+
+  val ENABLE_READ_CSV_LONGS = conf("spark.rapids.sql.csv.read.long.enabled")
+      .doc("Parsing CSV longs is much more lenient and will return 0 for some " +
+          "malformed values instead of null")
+      .booleanConf
+      .createWithDefault(false)
+
+  val ENABLE_READ_CSV_FLOATS = conf("spark.rapids.sql.csv.read.float.enabled")
+      .doc("Parsing CSV floats has some issues at the min and max values for floating" +
+          "point numbers also it can be more lenient on parsing inf and -inf values")
```
jlowe, line 740:
```suggestion
      .doc("Parsing CSV floats has some issues at the min and max values for floating " +
          "point numbers and can be more lenient on parsing inf and -inf values")
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -700,6 +693,60 @@ object RapidsConf {
...
+      .booleanConf
+      .createWithDefault(false)
+
+  val ENABLE_READ_CSV_DOUBLES = conf("spark.rapids.sql.csv.read.double.enabled")
+      .doc("Parsing CSV double has some issues at the min and max values for floating" +
+          "point numbers also it can be more lenient on parsing inf and -inf values")
```
jlowe, line 746:
```suggestion
      .doc("Parsing CSV double has some issues at the min and max values for floating " +
          "point numbers and can be more lenient on parsing inf and -inf values")
```

integration_tests/src/main/python/csv_test.py
```diff
@@ -93,37 +93,150 @@
 _good_str_schema = StructType([
```
jlowe, line 93: 2021 copyrights

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBatchScanExec.scala
```diff
@@ -173,46 +173,94 @@ object GpuCSVScan {
...
     if (StandardCharsets.UTF_8.name() != parsedOptions.charset &&
         StandardCharsets.US_ASCII.name() != parsedOptions.charset) {
       meta.willNotWorkOnGpu("GpuCSVScan only supports UTF8 encoded data")
     }
 
-    if (parsedOptions.ignoreLeadingWhiteSpaceInRead) {
-      // TODO need to fix this (or at least verify that it is doing the right thing)
-      meta.willNotWorkOnGpu("GpuCSVScan does not support ignoring leading white space")
+    // TODO parsedOptions.ignoreLeadingWhiteSpaceInRead cudf always do this, but not for strings
+    // TODO parsedOptions.ignoreTrailingWhiteSpaceInRead cudf always do this, but not for strings
```
jlowe, line 186:
```suggestion
    // TODO parsedOptions.ignoreLeadingWhiteSpaceInRead cudf always does this, but not for strings
    // TODO parsedOptions.ignoreTrailingWhiteSpaceInRead cudf always does this, but not for strings
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBatchScanExec.scala
```diff
@@ -173,46 +173,94 @@ object GpuCSVScan {
...
+            s" is not supported'")
+    }
+    // parsedOptions.maxCharsPerColumn does not impact the final output it is a performance
+    // improvement if you know the maximum size
 
-    if (parsedOptions.lineSeparator.getOrElse("\n") != "\n") {
-      meta.willNotWorkOnGpu("GpuCSVScan only supports \"\\n\" as a line separator")
+    // parsedOptions.maxColumns was originally a performance optimization but is not used any more
+
+    if (readSchema.map(_.dataType).contains(DateType)) {
```
jlowe, line 217: Nit: We could make these very-similar-but-not-quite-the-same checks more data-driven and parameterized, e.g.: a sequence of Spark types and conf keys or maybe a trivial case-class to make it easier to write/read the parameters.

tests/src/test/scala/com/nvidia/spark/rapids/HashAggregatesSuite.scala
```diff
@@ -29,8 +29,9 @@ import org.apache.spark.sql.functions._
 import org.apache.spark.sql.types.{DataType, DataTypes}
```
jlowe, line 29: 2021 copyrights

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBatchScanExec.scala
revans2: I'm less convinced about this, especially for DecimalType where contains on the map is not going to work, because a DecimalType is an actual class.  We don't support it yet, but are likely to eventually.

revans2:
build

revans2:
@jlowe could you please take another look

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
jlowe: This comment was missed in the update

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
revans2: that pesky `s` should have been added now.

jlowe (approved):

PR #2073, opened 2021-04-01
Avoid use of Spark leaf, unary and binary expressions and exec nodes

### LINKED ISSUE #2052: [BUG] Spark 3.2.0 test fails due to SPARK-34906 Refactor TreeNode's children handling methods into specialized traits
**Describe the bug**
nightly spark 3.2.0 test fail with:
```
[2021-03-31T07:24:52.771Z] [32mHashAggregatesSuite:[0m
[2021-03-31T07:25:02.731Z] [31m*** RUN ABORTED ***[0m
[2021-03-31T07:25:02.731Z] [31m  java.lang.NoSuchMethodError: org.apache.spark.sql.execution.UnaryExecNode.children$(Lorg/apache/spark/sql/execution/UnaryExecNode;)Lscala/collection/Seq;[0m
[2021-03-31T07:25:02.731Z] [31m  at com.nvidia.spark.rapids.GpuHashAggregateExec.children(aggregate.scala:302)[0m
[2021-03-31T07:25:02.731Z] [31m  at org.apache.spark.sql.catalyst.trees.TreeNode.containsChild$lzycompute(TreeNode.scala:120)[0m
[2021-03-31T07:25:02.731Z] [31m  at org.apache.spark.sql.catalyst.trees.TreeNode.containsChild(TreeNode.scala:120)[0m
[2021-03-31T07:25:02.731Z] [31m  at org.apache.spark.sql.catalyst.trees.TreeNode.mapChildren(TreeNode.scala:356)[0m
[2021-03-31T07:25:02.731Z] [31m  at org.apache.spark.sql.catalyst.trees.TreeNode.transformUp(TreeNode.scala:337)[0m
[2021-03-31T07:25:02.731Z] [31m  at com.nvidia.spark.rapids.GpuOverrides.addSortsIfNeeded(GpuOverrides.scala:2992)[0m
[2021-03-31T07:25:02.731Z] [31m  at com.nvidia.spark.rapids.GpuOverrides.applyOverrides(GpuOverrides.scala:2950)[0m
[2021-03-31T07:25:02.731Z] [31m  at com.nvidia.spark.rapids.GpuOverrides.apply(GpuOverrides.scala:2911)[0m
[2021-03-31T07:25:02.731Z] [31m  at com.nvidia.spark.rapids.GpuOverrides.apply(GpuOverrides.scala:2897)[0m
[2021-03-31T07:25:02.731Z] [31m  at org.apache.spark.sql.execution.ApplyColumnarRulesAndInsertTransitions.$anonfun$apply$1(Columnar.scala:532)[0m
```

This looks like this is from:

https://issues.apache.org/jira/browse/SPARK-34906

https://github.com/apache/spark/commit/bd0990e3e813d17065c593fc74f383b494fe8146

jlowe:
Fixes #2052.

Spark 3.2 changed the class hierarchy for leaf, unary and binary expressions and exec nodes, and the plugin was deriving many GPU classes from these abstract Spark classes.  This causes issues on Spark 3.2 because the location of the `child`, `children`, `left`, `right`, etc. methods has changed, and code compiled against the Spark 3.0 code is no longer binary compatible with these classes in Spark 3.2+.


jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2183,13 +2183,18 @@ object GpuOverrides {
         ("key", TypeSig.commonCudfTypes, TypeSig.all)),
       (in, conf, p, r) => new BinaryExprMeta[ArrayContains](in, conf, p, r) {
         override def tagExprForGpu(): Unit = {
+          // avoid using BinaryExpression accessors since hierarchy changed in Spark 3.2
```
tgravescs, line 2186: wonder if we want an explicit comment saying  left and right must match order in children

sql-plugin/src/main/scala/com/nvidia/spark/GpuDataSourceV2ScanExecBase.scala
```diff
@@ -0,0 +1,88 @@
...
+  def partitions: Seq[InputPartition]
+
+  def readerFactory: PartitionReaderFactory
+
+  override def simpleString(maxFields: Int): String = {
+    val result =
+      s"$nodeName${truncatedString(output, "[", ", ", "]", maxFields)} ${scan.description()}"
+    TrampolineUtil.redact(sqlContext, result)
+  }
+
```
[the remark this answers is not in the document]
jlowe: Good catch!  Unfortunately that isn't an API in Spark 3.0.0.  Might be more straightforward to ditch this new class and just shim GpuBatchScanExec directly.

jlowe:
@tgravescs I think I've addressed your review comments.  I refactored `GpuBatchScanExec` to be a shim due to the use of `LeafExecNode` which required updating another shim interface.

@abellina during testing I saw canonicalization failures which I tracked down to mishandling canonicalization in aggregation expressions.  I've fixed those tests, but I'm not 100% positive it's the proper fix.  Please take a look at the change in AggregateFunctions.scala.

jlowe:
build

jlowe:
build

tgravescs (approved):

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuBatchScanExec.scala
```diff
@@ -0,0 +1,44 @@
...
+import com.nvidia.spark.rapids.{GpuBatchScanExecBase, GpuDataSourceRDD}
+
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.AttributeReference
+import org.apache.spark.sql.catalyst.plans.QueryPlan
+import org.apache.spark.sql.connector.read.{InputPartition, PartitionReaderFactory, Scan}
+import org.apache.spark.sql.execution.datasources.v2.DataSourceV2ScanExecBase
+
+// Added to shims since LeafExecNode changed hierarchy in Spark 3.2
```
razajafri, line 28: @jlowe why are we adding a shim in 300 when when the change was only in 3.2?

jlowe:
The premerge CI was failing for Spark 3.2 due to a crash.  I'm not quite sure what is causing the crash, but as far as I can tell it is directly associated with CostBasedOptimizerSuite.  If I remove that from test on Spark 3.2 it always passes for me, but fails almost every time if it is present.  If this stubbing is OK, I'll file a followup issue to investigate the instability on Spark 3.2 so that we can re-enable that test across all Spark versions.

I also added a few commits to fix more cases of unary classes in the class hierarchy.

jlowe:
build

tgravescs (approved):
I'm good with merging and following up on the test issues

abellina:
I was trying to test this but another PR went into Spark that is breaking 3.2 (https://github.com/apache/spark/pull/32030). It seems related, but not sure if you want to address here. 

Why the hash agg change is needed for filter is odd to me too. I'll look more at that today. I did notice the filter expression is not included here: https://github.com/NVIDIA/spark-rapids/blob/branch-0.5/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala#L150... it is in the CPU case. So was going to go down that route.

If we want to merge this and follow up with a 3.2.0 follow up that would be fine too.

jlowe:
So a bit of bad news.  [SPARK-34989](https://issues.apache.org/jira/browse/SPARK-34989) recently updated the 3.2 code so that every node needs to override `withNewChildrenInternal`.  I've updated the code so it at least compiles, but it won't run as-is since the hierarchy isn't binary compatible, e.g.:
```
HashAggregatesSuite:
*** RUN ABORTED ***
  java.lang.AbstractMethodError:
  at org.apache.spark.sql.catalyst.trees.TreeNode.$anonfun$withNewChildren$2(TreeNode.scala:279)
  at org.apache.spark.sql.catalyst.trees.CurrentOrigin$.withOrigin(TreeNode.scala:73)
  at org.apache.spark.sql.catalyst.trees.TreeNode.withNewChildren(TreeNode.scala:278)
  at org.apache.spark.sql.execution.ApplyColumnarRulesAndInsertTransitions.insertRowToColumnar(Columnar.scala:514)
  at org.apache.spark.sql.execution.ApplyColumnarRulesAndInsertTransitions.$anonfun$insertRowToColumnar$1(Columnar.scala:514)
  at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:237)
  at scala.collection.immutable.List.foreach(List.scala:392)
  at scala.collection.TraversableLike.map(TraversableLike.scala:237)
  at scala.collection.TraversableLike.map$(TraversableLike.scala:230)
  at scala.collection.immutable.List.map(List.scala:298)
  ...
```

jlowe:
Adapting to the changes in [SPARK-34989](https://issues.apache.org/jira/browse/SPARK-34989) probably needs a different approach than just throwing what doesn't work in the shim manually.

PR #2075, opened 2021-04-02
Remove debug println from CBO test

andygrove:


andygrove:
build

tests/src/test/scala/com/nvidia/spark/rapids/CostBasedOptimizerSuite.scala
```diff
@@ -386,8 +386,6 @@ class CostBasedOptimizerSuite extends SparkQueryCompareTestSuite with BeforeAndA
         col("l.more_strings_1").equalTo(col("r.more_strings_2")))
       df.collect()
 
-      println(df.queryExecution.executedPlan)
-
       df
```
jlowe, line 389: Nit: I don't think this is necessary.

jlowe (approved):
Looks good, just a small nit that's not necessary to fix.

PR #2086, opened 2021-04-07
Updating documentation for data format support

### LINKED ISSUE #2003: [DOC] Misleading documentation for spark.rapids.sql.incompatibleDateFormats.enabled
## Report incorrect documentation

**Location of incorrect documentation**

The `RapidsConf` documentation for `spark.rapids.sql.incompatibleDateFormats.enabled` is slightly misleading.

**Describe the problems or issues found in the documentation**

The documentation states that:

```
When parsing strings as dates and timestamps in functions like unix_timestamp, 
setting this to true will force all parsing onto GPU even for formats that can 
result in incorrect results when parsing invalid inputs.
```

What isn't clear here is that some formats are not supported at all on GPU and will still fall back to CPU. For example, we don't support formats that include `MMM` on GPU.

**Steps taken to verify documentation is incorrect**

```
scala> spark.conf.set("spark.sql.session.timeZone", "UTC")
scala> spark.conf.set("spark.rapids.sql.incompatibleDateFormats.enabled", "true")
scala> val df_notsupported = Seq(("2021-Dec-25 11:11:11")).toDF("ts")
scala> df_notsupported.write.format("parquet").mode("overwrite").save("/tmp/testts_notsupported.parquet")
scala> spark.read.parquet("/tmp/testts_notsupported.parquet").createOrReplaceTempView("df_notsupported")
scala> val df = spark.sql("select to_timestamp(ts, 'yyyy-MMM-dd HH:mm:ss') from df_notsupported")
df: org.apache.spark.sql.DataFrame = [to_timestamp(ts, yyyy-MMM-dd HH:mm:ss): timestamp]

scala> df.collect
21/03/23 12:46:29 WARN GpuOverrides: 
!Exec <ProjectExec> cannot run on GPU because unsupported data types in output: TimestampType; not all expressions can be replaced
  !Expression <Alias> gettimestamp(ts#7, yyyy-MMM-dd HH:mm:ss, Some(UTC), true) AS to_timestamp(ts, yyyy-MMM-dd HH:mm:ss)#9 cannot run on GPU because expression GetTimestamp gettimestamp(ts#7, yyyy-MMM-dd HH:mm:ss, Some(UTC), true) produces an unsupported type TimestampType; expression Alias gettimestamp(ts#7, yyyy-MMM-dd HH:mm:ss, Some(UTC), true) AS to_timestamp(ts, yyyy-MMM-dd HH:mm:ss)#9 produces an unsupported type TimestampType
    !Expression <GetTimestamp> gettimestamp(ts#7, yyyy-MMM-dd HH:mm:ss, Some(UTC), true) cannot run on GPU because expression GetTimestamp gettimestamp(ts#7, yyyy-MMM-dd HH:mm:ss, Some(UTC), true) produces an unsupported type TimestampType; Failed to convert Unsupported word: MMM null
      @Expression <AttributeReference> ts#7 could run on GPU
      @Expression <Literal> yyyy-MMM-dd HH:mm:ss could run on GPU
  *Exec <FileSourceScanExec> will run on GPU
```

**Suggested fix for documentation**

````
When parsing strings as dates and timestamps in functions like unix_timestamp, some formats are
fully supported on GPU, some are supported but can produce incorrect results for invalid inputs, 
and others are not supported at all. Setting this to true will force all parsing onto GPU for supported formats, 
including formats that can result in incorrect results.
```

sameerz:
This is a first pass at updating the documentation to warn users about unsupported data formats.  

@andygrove if we have any more specific information about which formats are supported but can produce incorrect results for invalid inputs, it would be good for us to clarify that in the docs as well.  

Fixes #2003


andygrove:
@sameerz I'm no longer sure which formats can produce incorrect results because there were a number of improvements since this documentation was originally written. However, @viadea did run into one example (I think) that led to this issue being filed.

andygrove:
> @sameerz I'm no longer sure which formats can produce incorrect results because there were a number of improvements since this documentation was originally written. However, @viadea did run into one example (I think) that led to this issue being filed.

Actually, that isn't correct. @viadea was using an unsupported format, not a format that can produce incorrect results.

andygrove (approved):
LGTM

sameerz:
>> @sameerz I'm no longer sure which formats can produce incorrect results because there were a number of improvements since this documentation was originally written. However, @viadea did run into one example (I think) that led to this issue being filed.

> Actually, that isn't correct. @viadea was using an unsupported format, not a format that can produce incorrect results.

@andygrove is it still accurate to say that "some are supported but can produce incorrect results for invalid inputs" or should I update the document?

sameerz:
@andygrove is going to investigate `sql.incompatibleDateFormats` and either remove it or provide more specific information about incompatible formats for the documentation.  Putting this into a draft state for the time being.

sameerz:
@andygrove please re-review

sameerz:
build

pxLi:
build

andygrove (approved):
Thanks, Sameer. I think this is much clearer now.

Issue #2090, opened 2021-04-07
[FEA] Make row count estimates available to the cost-based optimizer


andygrove:
**Is your feature request related to a problem? Please describe.**
In order to implement a real cost model in the cost-based optimizer, it is important to have an estimate of the input row count for an operator. For example, it might be worth moving to GPU for a compute-intensive operation on millions of rows, but not for just a few rows.

**Describe the solution you'd like**
Spark already contains logic for estimating data sizes (but not row counts) for a logical plan (`SizeInBytesOnlyStatsPlanVisitor`) and I propose that we based the solution on that logic.

We should also leverage actual row counts when they are available from executed query stages when AQE is on.

PR #2093, opened 2021-04-07
Initial implementation of row count estimates in cost-based optimizer

andygrove:

First pass at making estimated row counts available to the cost-based optimizer.

Closes https://github.com/NVIDIA/spark-rapids/issues/2090

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
@@ -21,8 +21,9 @@ import scala.collection.mutable.ListBuffer
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.expressions.{Alias, AttributeReference, Expression}
 import org.apache.spark.sql.catalyst.plans.{JoinType, LeftAnti, LeftSemi}
-import org.apache.spark.sql.execution.{GlobalLimitExec, LocalLimitExec, ProjectExec, SparkPlan, UnionExec}
+import org.apache.spark.sql.execution.{GlobalLimitExec, LocalLimitExec, ProjectExec, SparkPlan, TakeOrderedAndProjectExec, UnionExec}
 import org.apache.spark.sql.execution.adaptive.{CustomShuffleReaderExec, QueryStageExec}
+import org.apache.spark.sql.execution.aggregate.HashAggregateExec
 import org.apache.spark.sql.execution.exchange.ShuffleExchangeExec
 import org.apache.spark.sql.execution.joins.{BroadcastHashJoinExec, BroadcastNestedLoopJoinExec, ShuffledHashJoinExec, SortMergeJoinExec}
 import org.apache.spark.sql.internal.SQLConf
@@ -275,7 +276,7 @@ class DefaultCostModel(conf: RapidsConf) extends CostModel {
 
 /**
  * Estimate the number of rows that an operator will output. Note that these row counts are
- * the aggregate across all partitions.
+ * the aggregate across all output partitions.
  *
  * Logic is based on Spark's SizeInBytesOnlyStatsPlanVisitor. which operates on logical plans
  * and only computes data sizes, not row counts.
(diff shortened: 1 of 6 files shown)


andygrove:
build

andygrove:
@revans2 Could you review this when you get a chance

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -270,6 +273,62 @@ class DefaultCostModel(conf: RapidsConf) extends CostModel {
 
 }
 
+/**
+ * Estimate the number of rows that an operator will output. Note that these row counts are
+ * the aggregate across all partitions.
```
revans2, line 278: This is not necessarily true with how the code is written.

LocalLimit and GlobalLimit are the same code, but GlobalLimit requires a single partition, while LocalLimit does not. That means that GlobalLimit is across all partitions, because there is only a single partition, but LocalLimit is per-partition.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -270,6 +273,62 @@ class DefaultCostModel(conf: RapidsConf) extends CostModel {
...
+object RowCountPlanVisitor {
+
+  def visit(plan: SparkPlanMeta[_]): Option[BigInt] = plan.wrapped match {
+    case p: QueryStageExec =>
+      ShimLoader.getSparkShims.getQueryStageRuntimeStatistics(p).rowCount
+    case GlobalLimitExec(limit, _) =>
+      visit(plan.childPlans.head).map(_.min(limit)).orElse(Some(limit))
+    case LocalLimitExec(limit, _) =>
+      visit(plan.childPlans.head).map(_.min(limit)).orElse(Some(limit))
+    case p: SortMergeJoinExec =>
```
revans2, line 292: I would add in a TakeOrderedAndProject in here that behaves a lot like the limits, but it also sorts the data.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -270,6 +273,62 @@ class DefaultCostModel(conf: RapidsConf) extends CostModel {
...
+      estimateJoin(plan, p.joinType)
+    case p: ShuffledHashJoinExec =>
+      estimateJoin(plan, p.joinType)
+    case p: BroadcastHashJoinExec =>
+      estimateJoin(plan, p.joinType)
+    case _: UnionExec =>
+      Some(plan.childPlans.flatMap(visit).sum)
+    case _ =>
```
revans2, line 300: I would also want to add in something for aggregates.  Aggregates will change the number of output rows based off of the cardinality of the input, so it should be very similar to how we estimate a Join, with the key difference being that if there are no grouping keys (a reduction), then each partition will output a single row.

andygrove: I added in the empty grouping expression case and for other cases, it will fall through the default handling, which will just return the product of the child plan row counts. I don't know how we can get statistics on the cardinality of the grouping expressions to be able to determine anything more accurate. It would be nice if Spark could collect those kind of statistics in AQE.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
andygrove: Added

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
andygrove: I updated the logic for LocalLimit to multiple the limit by the number of output partitions for consistency.

revans2 (approved):

andygrove:
build

Issue #1445, opened 2021-01-05
[FEA] GDS Integration


rongou:
## Motivation
For many jobs running on spark-rapids, the ratio of computation to data movement is low, making them I/O bound. GPUDirect Storage ([GDS](https://docs.nvidia.com/gpudirect-storage/index.html)) enables Direct Memory Access (DMA) between storage and GPUs, with increased bandwidth and lower latency, thus could potentially reduce file I/O overhead and increase overall job performance.

## Goals
* Integrate GDS and the cuFile API into spark-rapids.
* Support reading input data and writing output data through GDS.
* Support storing Spark temporary data using GDS.
* Benchmark GDS performance with spark-rapids and provide recommendations on when to use GDS.

## Non-Goals
* GDS/cuFile is currently distributed as a user library and a kernel module, and requires installing MOFED and a few third-party libraries. In the future GDS may become part of the CUDA driver and the CUDA toolkit. The exact distribution mechanism is beyond the scope of this document.
* For distributed filesystems, GDS currently supports Lustre, NFS, and WekaFS, which are not commonly used for storing data for Spark jobs. The initial integration will not focus on distributed filesystem support.
* In a multi-node environment, GDS can write to and read from a remote storage server through NVMe-oF. This may be an alternative to the current UCX-based shuffle plugin, but probably not a full replacement.

## Assumptions
In a deployment environment, we assume the user has correctly set up NVMe/NVMe-oF support for GDS, and installed the cuFile kernel module and the cuFile user library.

## Risks
* Until the final distribution mechanism is sorted out, spark-rapids with GDS may add extra complexity to the deployment process.
* Depending on the hardware and software stack, and the particular Spark job, GDS may not provide a significant performance boost.

## Design
This [blog post](https://developer.nvidia.com/blog/gpudirect-storage/) provides a good overview of GDS. There is some public [documentation](https://docs.nvidia.com/gpudirect-storage/index.html).

For spark-rapids, integrating with GDS can be roughly considered along two dimensions: the type of data, and the type of storage.

A Spark job may deal with these types of data involving disk storage:
* Input data. This is the data a Spark job reads from disk for processing, typically in CSV or Parquet format, and usually of large size.
* Output data. This is the data a Spark job writes out. For an analytical query, the output size may be very small, but for data transformations (e.g. converting from CSV to Parquet) the output volume can be considerable.
[...]

PR #2095, opened 2021-04-08
fix refcount when materializing device buffer from GDS

rongou:

Ideally this should be covered by a test, but right now we can't run GDS tests with the plugin. I added an action item in #1445 to add tests once the CI/CD pipeline is set up to run GDS tests.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
@@ -82,7 +82,6 @@ class RapidsGdsStore(
       closeOnExcept(DeviceMemoryBuffer.allocate(size)) { buffer =>
         CuFile.readFileToDeviceBuffer(buffer, path, fileOffset)
         logDebug(s"Created device buffer for $path $fileOffset:$size via GDS")
-        buffer.incRefCount()
         buffer
       }
     }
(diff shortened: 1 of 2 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -82,6 +82,7 @@ class RapidsGdsStore(
       closeOnExcept(DeviceMemoryBuffer.allocate(size)) { buffer =>
         CuFile.readFileToDeviceBuffer(buffer, path, fileOffset)
         logDebug(s"Created device buffer for $path $fileOffset:$size via GDS")
+        buffer.incRefCount()
```
jlowe, line 85: `DeviceMemoryBuffer.allocate` returns a buffer that has a refcount of 1, and we don't close it before it's returned.  Incrementing it here seems like we're asking it to leak, since I would only expect the caller of this method to close it once.

rongou: Ah ok, I think the problem is we are closing the device buffer in `getDeviceMemoryBuffer`. Should be fixed now.

rongou:
build

jlowe (approved):

abellina (approved):

PR #2105, opened 2021-04-08
Cleanup prior Spark sessions in tests consistently

jlowe:
The tests are using multiple, different ways to shut down any prior Spark session before trying to start a new session.  This updates the code to consistently use `cleanupAnyExistingSession` which is the same cleanup code used in Spark's tests.


jlowe:
build

jlowe:
> does the GpuDeviceManager.shutdown method still get called under the hood?

Yes, via the plugin shutdown that is called when the Spark session shuts down.

abellina (approved):

tgravescs (approved):

PR #2106, opened 2021-04-08
Fix some buffer leaks

jlowe:
This fixes a few resource leaks that I saw when running the unit tests.  As part of cleaning up the buffer leak in the out-of-core sorter, I also changed it so it adds the contiguous table directly to the spill framework rather than manifesting a `Table` that is then immediately discarded.


jlowe:
build

abellina (approved):

PR #2114, opened 2021-04-10
Modified the Download page for 0.4.1 and updated doc to point to K8s guide

viadea:


tgravescs:
this seems fine but one thing I wonder if it would be more consistent to split out yarn and standalone into own sections as well that way when I go to the docs and twisty the getting started I see all of them rather then having to go into on Prem non for just yarn and standalone. That should be different PR but if folks agree we should file a followup.  @sameerz @jlowe thoughts?

tgravescs:
build

tgravescs (approved):

jlowe (approved):
> I wonder if it would be more consistent to split out yarn and standalone into own sections as well that way when I go to the docs and twisty the getting started I see all of them rather then having to go into on Prem non for just yarn and standalone. That should be different PR but if folks agree we should file a followup.

Agree, I'd like to see each setup treated consistently.

sameerz:
> > I wonder if it would be more consistent to split out yarn and standalone into own sections as well that way when I go to the docs and twisty the getting started I see all of them rather then having to go into on Prem non for just yarn and standalone. That should be different PR but if folks agree we should file a followup.
> 
> Agree, I'd like to see each setup treated consistently.

Agree, filed an issue #2121

sameerz (approved):

PR #2120, opened 2021-04-13
Materialize Stream before serialization

### LINKED ISSUE #2036: [BUG] Stackoverflow when writing wide parquet files.
**Describe the bug**
When using basic Spark SQL functions to add a new column to a wide(784 columns) parquet file, see Stackoverflow error.

inference.py :
```
from pyspark.sql import SparkSession
from pyspark.sql.functions import col
spark  = SparkSession.builder.appName("TF SPARK DEMO").getOrCreate()

df = spark.read.parquet('test_2')
new_df = df.withColumn('out', col('1')*100)
new_df.write.parquet("result")
```

Spark parameters:
```
CUDF_JAR=$PWD/cudf-0.18.1-cuda11.jar
PLUGIN_JAR=$PWD/rapids-4-spark_2.12-0.4.1.jar
SPARK_HOME=/opt/spark-3.0.2-bin-hadoop3.2

rm -rf result

$SPARK_HOME/bin/spark-submit --master spark://127.0.0.1:7077            \
--jars ${CUDF_JAR},${PLUGIN_JAR}                       \
--conf spark.plugins=com.nvidia.spark.SQLPlugin        \
--driver-memory 10G                                    \
--executor-memory 30G                                  \
--conf spark.rapids.memory.gpu.allocFraction=0.3       \
--conf spark.executor.extraClassPath=${CUDF_JAR}:${PLUGIN_JAR} \
--conf spark.driver.extraClassPath=${CUDF_JAR}:${PLUGIN_JAR} \
--conf spark.driver.extraJavaOptions="-Duser.timezone=UTC" \
--conf spark.rapids.sql.decimalType.enabled=true \
--conf spark.rapids.memory.pinnedPool.size=3g \
--conf spark.rapids.sql.hashOptimizeSort.enabled=true \
--conf spark.rapids.memory.gpu.pooling.enabled=true \
--conf spark.executor.extraJavaOptions="-Duser.timezone=UTC -Dai.rapids.cudf.prefer-pinned=true" \
--conf spark.driver.extraJavaOptions="-Duser.timezone=UTC" \
--conf spark.rapids.sql.expression.UnixTimestamp=true \
--conf spark.sql.parquet.outputTimestampType=TIMESTAMP_MICROS \
--conf spark.executor.resource.gpu.amount=1 \
--conf spark.rapids.sql.explain=ALL \
--conf spark.task.resource.gpu.amount=0.01 \
--conf spark.rapids.sql.concurrentGpuTasks=2 \
--conf spark.executor.resource.gpu.discoveryScript=./getGpusResources.sh \
--files $SPARK_HOME/examples/src/main/scripts/getGpusResources.sh \
inference.py

```

error:

```
21/03/29 09:56:24 WARN TaskSetManager: Lost task 0.2 in stage 1.0 (TID 3, 172.17.0.2, executor 1): java.lang.StackOverflowError
        at java.io.ObjectInputStream$BlockDataInputStream.readByte(ObjectInputStream.java:3244)
        at java.io.ObjectInputStream.readHandle(ObjectInputStream.java:1778)
        at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1632)
        at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2405)
        at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2329)
        at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2187)
        at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1667)
        at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2405)
        at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2329)
        at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2187)
        at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1667)
        at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2405)
        at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2329)
        at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2187)
        at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1667)
        at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2405)
        at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2329)
        at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2187)
        at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1667)
        at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2405)

```

**Expected behavior**
no error and parquet file correctly saved.

 - Environment location: Standalone
 - Spark configuration settings related to the issue: GPU:  single Titan V

**Additional context**
parquet file
[test_2.zip](https://github.com/NVIDIA/spark-rapids/files/6220927/test_2.zip)

Also, tried without plugin several times, and didn't see such error.

gerashegalov:
This is a workaround for the stackoverflow during the recursive Stream
serialization. Fixes #2036 


gerashegalov:
Test falls before the PR, succeeds after:
```
TEST_PARALLEL=1 SPARK_HOME=~/spark-3.1.1-bin-hadoop3.2 \
  integration_tests/run_pyspark_from_build.sh -k test_many_column_project
```

gerashegalov:
build

wjxiz1992:
Hi @gerashegalov, could you kindly explain a little bit why using Scala List instead of Seq is a "Materialize"? 
Per https://github.com/apache/spark/pull/24865, my understanding is  Seq[Expression] will trigger recursive serialization calls while List[Expression] will serialize 1 by 1. But I still don't have a clear map for them.

razajafri:
@gerashegalov The code change is simple enough. Can you please explain why and how this solves the problem for us laymen? :)

integration_tests/src/main/python/parquet_test.py
```diff
@@ -465,3 +464,12 @@ def test_spark_32639(std_input_path):
         lambda spark: spark.read.schema(schema_str).parquet(data_path),
         conf=original_parquet_file_reader_conf)
 
+def test_many_column_project(std_input_path):
+    data_path = f"{std_input_path}/issue-2036"
+    with_gpu_session(
```
revans2, line 469: nit: I would prefer if we validate that the output is the same on both the CPU and the GPU.  Just checking that we don't crash is good, but it is better to validate that we also got the correct answer.

revans2 (commented):
I'm not super happy with adding a 3MB binary file. Could we have fewer rows in it? Could we even generate it on the fly on the CPU instead of checking it in?

revans2:
> @gerashegalov The code change is simple enough. Can you please explain why and how this solves the problem for us laymen? :)

`Seq` can be lazy. That means something like
```
val info = foo.map(a => a + 100)
```
might not actually execute anything. Instead it stored the function `a => a + 100` along with `foo` inside a new lazy sequence class and will only run when something happens to info. This lets you do some optimizations like if foo is 10,000 elements long
```
info.head
```
only needs to process one of those elements.

This becomes a problem when we do serialization though. Apparently `Seq` in some cases stores data in something similar to a `LinkedList` that is hidden internally in `Seq` and is not optimized for java serialization. Because of this java serialization will then recursively call into itself as it serializes this data, one stack frame per item in the list. If the number of entries is long enough we can end up hitting stack limits in the JVM.

If you are passing around a `List` instead of a `Seq` internally the data is stored more like an array, or it is at least optimized for java serialization so there far fewer recursive calls during serialization, and we don't have stack limit issues.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBoundAttribute.scala
```diff
@@ -68,7 +68,7 @@ object GpuBindReferences extends Logging {
   def bindGpuReferences[A <: Expression](
```
jlowe, line 68: 2021 copyrights

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBoundAttribute.scala
```diff
@@ -68,7 +68,7 @@ object GpuBindReferences extends Logging {
...
       expressions: Seq[A],
       input: AttributeSeq): Seq[GpuExpression] =
-    expressions.map(GpuBindReferences.bindGpuReference(_, input))
+    expressions.map(GpuBindReferences.bindGpuReference(_, input)).toList
```
jlowe, line 71: Curious why `toList` was used instead of `toArray` as I've seen done elsewhere to force a manifest?  Either way, probably worth adding a comment so someone doesn't "optimize" it later. :smile:

razajafri:
> > @gerashegalov The code change is simple enough. Can you please explain why and how this solves the problem for us laymen? :)
> 
> `Seq` can be lazy. That means something like
> 
> ```
> val info = foo.map(a => a + 100)
> ```
> 
> might not actually execute anything. Instead it stored the function `a => a + 100` along with `foo` inside a new lazy sequence class and will only run when something happens to info. This lets you do some optimizations like if foo is 10,000 elements long
> 
> ```
> info.head
> ```
> 
> only needs to process one of those elements.
> 
> This becomes a problem when we do serialization though. Apparently `Seq` in some cases stores data in something similar to a `LinkedList` that is hidden internally in `Seq` and is not optimized for java serialization. Because of this java serialization will then recursively call into itself as it serializes this data, one stack frame per item in the list. If the number of entries is long enough we can end up hitting stack limits in the JVM.
> 
> If you are passing around a `List` instead of a `Seq` internally the data is stored more like an array, or it is at least optimized for java serialization so there far fewer recursive calls during serialization, and we don't have stack limit issues.

Thank you very much Bobby! that explains it

gerashegalov:
@wjxiz1992 @razajafri here is my understanding of the issue

`Seq` is just a trait with many implementations. `List` is an implementation of `Seq`, and so is `Stream` that unlike conventional  computes elements on demand. 
```scala
scala> List(1).isInstanceOf[Seq[Int]]
res4: Boolean = true

(1 to 100).isInstanceOf[Seq[Int]]
res6: Boolean = true
```
stream.map is still lazy ondemand because it just appends another function to apply prior to producing an element as @revans2 explained.

The evaluated prefix of the Stream is represented as a [linked list](https://github.com/szeiger/scala/blob/5d56f69ffc07ae96347b5a0a6c1dd5c17cfaa8df/src/library/scala/collection/immutable/LazyList.scala#L618) 
 
Prior to [this fix in Scala 2,13](https://github.com/scala/scala/pull/6676/files#diff-386ea95e2f79d6d9a0456a069023483132643a4c26aad8802d40f0250b9cfba7R477) the default Java serialization is recursive and once the Stream is long enough exhausts the thread stack size.

integration_tests/src/main/python/parquet_test.py
revans2: Could we have a comment then explaining it?

gerashegalov: it's probably in line with the comment https://github.com/NVIDIA/spark-rapids/issues/2036#issuecomment-810336351, and parts of the serialization are cached. I could not find a way to uncache it.  Instead added a flag to be able to reorder GPU and CPU runs and then we can hit the issue reliably.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBoundAttribute.scala
gerashegalov: added a comment https://github.com/NVIDIA/spark-rapids/pull/2120/files#diff-7718c93b5e463c82611c693aa760c934343e3d8047dd800f1d0c00e9ab8afa52R63

gerashegalov:
build

integration_tests/src/main/python/asserts.py
```diff
@@ -312,52 +312,70 @@ def assert_gpu_fallback_collect(func,
 
 def _assert_gpu_and_cpu_are_equal(func,
         mode,
+        is_cpu_first=True,
```
jlowe, line 315: New arguments should not appear before other default args, or we risk existing code passing a conf as the third argument and it now being interpreted as the `is_cpu_first` argument.  Code can still call it with `is_cpu_first=False` if they don't want to pass a conf explicitly.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBoundAttribute.scala
jlowe: The comment isn't anywhere near this code.  Would be nice to have a simple one-liner like `// Force list to avoid recursive Java serialization of Seq` so maintainers don't think the `toList` is superfluous.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBoundAttribute.scala
```diff
@@ -83,7 +83,7 @@ object GpuBindReferences extends Logging {
   def bindReferences[A <: Expression](
       expressions: Seq[A],
       input: AttributeSeq): Seq[A] =
-    expressions.map(GpuBindReferences.bindReference(_, input))
+    expressions.map(GpuBindReferences.bindReference(_, input)).toList
```
jlowe, line 86: Similar one-liner comment here

jlowe (dismissed):

revans2 (approved):

gerashegalov:
build

jlowe (approved):

gerashegalov:
build

Issue #1755, opened 2021-02-18
[DOC] Improve supports ops table readability


tgravescs:
The supports ops tables are kind of hard to read since some are rather large and the column names are way at the top. Its hard to scroll and figure out which columns you are looking at while figuring out the row names as well.

perhaps  we could repeat the header rows periodically throughout the table would help readability.

himanshu007-creator:
Hi i would like to work on this issue!

tgravescs:
That would be great if you would like to contribute.  Please let us know what your ideas are for improving before starting coding so we can discuss.

revans2:
Ya it would be great. I started to work on adding the header periodically through the table, just because I am in the process of adding in new checks for partitioning, and probably do the code generation for scans as well.  But if there are better ideas it is simple to remove what I have been doing.

PR #2128, opened 2021-04-14
Add in Partition type check support

### LINKED ISSUE #1915: [FEA] follow on work for hash partitioning changes
#1910 made hash partitioning match exactly what Spark does.  There were a few issues found and both #1913 and #1914 were filed to cover those.  This is intended to cover some of the other things that need to be done.

- [x] Identity hash for hash_partition and performance checks for it too.
The hash_partition JNI bindings in cudf did not include any way to pass down the hash function to use.  It really would be nice to be able to use hash_partition along with the identity hash function instead of sorting the data and doing upper bounds on it ourselves.  But we need to make sure that the performance does not get worse. I don't see how it could, but we do need to check it.  Be aware that hash_partition is using a regular modulo instead of a pmod so we will still have to calculate the exact partition we want and pass that in instead of just the murmur3 hash function. 

- [x] BitCast updates for Murmur3.
The Murmur3 implementation needed a little bit of work before we could get the individual parts passed in to match what spark was doing.  There are a very few number of places where we need to pull out the raw bits from a value (float/double #1914) and decimal when it is a decimal64 value.  In those cases it really would be nice to write the code in such a way that we can do it without having to materialize a new temporary column.  The float/double is less of a big deal because we have to normalize the NaN values ahead of time, so we will create a temp value anyways.  And for decimal we have to be extra careful because after #1717 it will only work for decimal64.  Decimal32 would still need to have the value copied out.

- [x] performance testing/tuning
Because this was put in as a really fast bug fix at the end of the 0.4 release there wasn't a lot of time to really look at how to performance tune this, or if it is needed.  We should at least do some profiling and see how it performs compared to the previous code.  If we use the identity hash function then really the only extra cost would be computing the murmur3 hash, and the savings would be cudf not computing their murmur3 hash.

- [ ] Type Checking for Partitioning
The current GpuOverrides does not do any type checking for partitioning.  The HashPartitioning does not rewrite the expressions to insert in the murmur3 with the pmod at that time, so we cannot rely on it to do the type checks for us.  The code that went into 0.4 just did the check, but no documentation was generated for it or anything like that.  We should update the type checking and documentation generation to do the right thing.

revans2:
This fixes #1915

I could add in support for input/output checks for Parquet, Orc, and CSV if people want me to, but I thought I would keep it simple to start out with.

This also addresses part of #1755 by outputting headers periodically in the tables. It is better than it was before, but I don't think it is a final solution.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-
 package com.nvidia.spark.rapids
 
 import java.io.{File, FileOutputStream}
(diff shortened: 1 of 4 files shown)


revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -601,6 +600,57 @@ object ExecChecks {
...
+  override def tag(meta: RapidsMeta[_, _, _]): Unit = {
+    val part = meta.wrapped
+    val children = meta.childExprs.map(_.wrapped.asInstanceOf[Expression]).toArray
+
+    val fixedChecks = paramCheck.toArray
+    assert (fixedChecks.length <= children.length,
+      s"${part.getClass.getSimpleName} expected at least ${fixedChecks.length} but " +
+          s"found ${children.length}")
+    fixedChecks.indices.foreach { i =>
+      val check = fixedChecks(i)
```
jlowe, line 622: Nit: can simplify with something like this:
```suggestion
    fixedChecks.zipWithIndex.foreach { case (check, i) =>
```

jlowe (dismissed):
Only a minor simplification suggestion, lgtm.

jlowe (approved):

revans2:
build

revans2:
build

Issue #2078, opened 2021-04-04
[BUG] java.lang.ArithmeticException: divide by zero when spark.sql.ansi.enabled=true


viadea:
**Describe the bug**

When `spark.sql.ansi.enabled=true` , NDS query Q7 will fail with below stacktrace:
```
Caused by: java.lang.ArithmeticException: divide by zero
  at org.apache.spark.sql.rapids.GpuDivModLike.divByZeroError(arithmetic.scala:189)
  at org.apache.spark.sql.rapids.GpuDivModLike.$anonfun$doColumnar$3(arithmetic.scala:196)
  at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
  at com.nvidia.spark.rapids.Arm.withResource$(Arm.scala:26)
  at com.nvidia.spark.rapids.CudfBinaryOperator.withResource(GpuExpressions.scala:261)
  at org.apache.spark.sql.rapids.GpuDivModLike.doColumnar(arithmetic.scala:194)
  at org.apache.spark.sql.rapids.GpuDivModLike.doColumnar$(arithmetic.scala:192)
  at org.apache.spark.sql.rapids.GpuDivide.doColumnar(arithmetic.scala:243)
  at com.nvidia.spark.rapids.GpuBinaryExpression.columnarEval(GpuExpressions.scala:167)
  at com.nvidia.spark.rapids.GpuBinaryExpression.columnarEval$(GpuExpressions.scala:158)
  at com.nvidia.spark.rapids.CudfBinaryOperator.columnarEval(GpuExpressions.scala:261)
  at com.nvidia.spark.rapids.GpuHashAggregateExec.$anonfun$doExecuteColumnar$5(aggregate.scala:502)
  at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
  at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
  at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
  at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
  at scala.collection.TraversableLike.map(TraversableLike.scala:238)
  at scala.collection.TraversableLike.map$(TraversableLike.scala:231)
  at scala.collection.AbstractTraversable.map(Traversable.scala:108)
[...]

PR #2130, opened 2021-04-14
Fix divide-by-zero in GpuAverage with ansi mode

abellina:
Fixes: https://github.com/NVIDIA/spark-rapids/issues/2078

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
@@ -269,7 +269,10 @@ object GpuDivModLike {
 }
 
 trait GpuDivModLike extends CudfBinaryArithmetic {
-  lazy val failOnError: Boolean = ShimLoader.getSparkShims.shouldFailDivByZero()
+  val failOnErrorOverride: Option[Boolean] = None
+
+  lazy val failOnError: Boolean =
+    failOnErrorOverride.getOrElse(ShimLoader.getSparkShims.shouldFailDivByZero())
 
   override def nullable: Boolean = true
 
@@ -330,7 +333,8 @@ object GpuDivideUtil {
 }
 
 // This is for doubles and floats...
-case class GpuDivide(left: Expression, right: Expression) extends GpuDivModLike {
+case class GpuDivide(left: Expression, right: Expression,
+    override val failOnErrorOverride: Option[Boolean] = None) extends GpuDivModLike {
   override def inputType: AbstractDataType = TypeCollection(DoubleType, DecimalType)
 
   override def symbol: String = "/"
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
@@ -366,7 +366,7 @@ case class GpuDivide(left: Expression, right: Expression,
     (left.dataType, right.dataType) match {
       case (_: DecimalType, r: DecimalType) => {
         val (upcastedLhs, upcastedRhs) = if (!DecimalType.is32BitDecimalType(dataType) &&
-            DecimalType.is32BitDecimalType(r)) {
+          DecimalType.is32BitDecimalType(r)) {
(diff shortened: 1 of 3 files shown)


andygrove (approved):
LGTM

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -330,7 +333,8 @@ object GpuDivideUtil {
 }
 
 // This is for doubles and floats...
-case class GpuDivide(left: Expression, right: Expression) extends GpuDivModLike {
+case class GpuDivide(left: Expression, right: Expression,
+    override val failOnErrorOverride: Option[Boolean] = None) extends GpuDivModLike {
```
andygrove, line 337: The change looks good but I was curious why we are using a different pattern to Spark which just has a plain boolean argument with a default value, rather than using an option. Was this necessary because of the way we're using the shim layer?

abellina: Yes, I tried other ways but couldn't think of something cleaner.

jlowe (approved):

revans2 (approved):

abellina:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -269,7 +269,10 @@ object GpuDivModLike {
 }
 
 trait GpuDivModLike extends CudfBinaryArithmetic {
-  lazy val failOnError: Boolean = ShimLoader.getSparkShims.shouldFailDivByZero()
+  val failOnErrorOverride: Option[Boolean] = None
+
```
gerashegalov, line 273: let us try without `failOnErrorOverride`, just make failOnError non-lazy, so we can override it

gerashegalov (changes requested):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
revans2: You can override a lazy val

```
override lazy val failOnError: Boolean = failOnErrorOverride.getOrElse(GpuDivModeLike.failOnError)
```

But I am fine with keeping this as is.

abellina:
@gerashegalov we don't have ANSI semantics in several of the operations. In the `DivModLike` ops alone, the `failOnError` flag is actually given (in Spark) to: `Divide`, `IntegralDivide` and `Remainder`. Each of these has as default `SQLConf.get.ansiEnabled`. In our case, `Pmod` falls under the `DivModLike` category also.

Doing a quick search in Spark, I am finding that:
- `Divide`: `failOnError` is passed only in the `Average` as per this PR.
- `IntegralDivide`: `failOnError` in case class signature but not overwritten. 
- `Reminder`: `failOnError` in case class signature but not overwritten. 
- `Pmod`: `failOnError` in case class signature but not overwritten. 

So the only class that really needs it is `GpuDivide`, and it seems better (and less error prone) to me to have the call to the shim in a single place `GpuDivModLike`, with the override possibility, as only `GpuDivide` requires it.

That said, if we want to match Spark more, I can look into adding the `failOnError` to all the `GpuDivModLike`s. Interested to know if we need an issue to track ANSI work for the plugin in general, as it seems that many places have forks for this.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
gerashegalov: yes, I just meant you can't use lazy as a parameter.

gerashegalov:
@abellina this is the change I am suggesting in the nut shell. https://github.com/abellina/spark-rapids/compare/agg/fix_ansi_avg...gerashegalov:agg/fix_ansi_avg?expand=1

@revans2 if we made failOnError non-lazy then we don't need an override inside the GpuDivide case class body but could do it just as a param

abellina:
@gerashegalov updated PR to incorporate your suggestion.

gerashegalov (approved):
LGTM

abellina:
build

revans2 (approved):

PR #2132, opened 2021-04-14
Log plugin and cudf versions on startup

### LINKED ISSUE #2018: [FEA] A way for user to find out the plugin version and cudf version in REPL
Currently, there is no way for a user to determine what version of plugin and cudf is the spark-shell started with. A user can list the jars if he is running Spark with jars loaded, but I think there should be a user-friendly way for the user to find the version like Spark allows by typing `sc.version`. 

We can also print the version when Spark-shell is first loaded if its an easy thing to do

jlowe:
Fixes #2018 

This logs the RAPIDS Accelerator plugin version and cudf versions found on the classpath during startup of the driver and executor.  On the driver there is a message that looks like this:
```
21/04/14 18:07:57 WARN SQLExecPlugin: RAPIDS Accelerator 0.5.0-SNAPSHOT using cudf 0.19-SNAPSHOT. To disable GPU support set `spark.rapids.sql.enabled` to false
```
and on the executors will be a log message that looks like this:
```
21/04/14 13:07:59 INFO RapidsExecutorPlugin: RAPIDS Accelerator 0.5.0-SNAPSHOT using cudf 0.19-SNAPSHOT
```

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
@@ -38,6 +38,10 @@ import org.apache.spark.sql.internal.StaticSQLConf
 import org.apache.spark.sql.rapids.GpuShuffleEnv
 import org.apache.spark.sql.util.QueryExecutionListener
 
+class PluginException(msg: String) extends RuntimeException(msg)
+
+case class CudfVersionMismatchException(errorMsg: String) extends PluginException(errorMsg)
+
 case class ColumnarOverrideRules() extends ColumnarRule with Logging {
   val overrides: Rule[SparkPlan] = GpuOverrides()
   val overrideTransitions: Rule[SparkPlan] = new GpuTransitionOverrides()
@@ -52,7 +56,9 @@ case class ColumnarOverrideRules() extends ColumnarRule with Logging {
  */
 class SQLExecPlugin extends (SparkSessionExtensions => Unit) with Logging {
   override def apply(extensions: SparkSessionExtensions): Unit = {
-    logWarning("Installing extensions to enable rapids GPU SQL support." +
+    val pluginVersion = RapidsPluginUtils.getPluginVersion
+    val cudfVersion = RapidsPluginUtils.getClasspathCudfVersion
+    logWarning(s"RAPIDS Accelerator $pluginVersion using cudf $cudfVersion." +
       s" To disable GPU support set `${RapidsConf.SQL_ENABLED}` to false")
     extensions.injectColumnar(_ => ColumnarOverrideRules())
     ShimLoader.getSparkShims.injectQueryStagePrepRule(extensions, _ => GpuQueryStagePrepOverrides())
@@ -66,8 +72,10 @@ object RapidsPluginUtils extends Logging {
   private val SERIALIZER_CONF_KEY = "spark.serializer"
   private val JAVA_SERIALIZER_NAME = classOf[JavaSerializer].getName
   private val KRYO_SERIALIZER_NAME = classOf[KryoSerializer].getName
-  private val KRYO_REGISRATOR_KEY = "spark.kryo.registrator"
-  private val KRYO_REGISRATOR_NAME = classOf[GpuKryoRegistrator].getName
+  private val KRYO_REGISTRATOR_KEY = "spark.kryo.registrator"
+  private val KRYO_REGISTRATOR_NAME = classOf[GpuKryoRegistrator].getName
+  private val CUDF_PROPS_FILENAME = "cudf-java-version-info.properties"
+  private val PLUGIN_PROPS_FILENAME = "rapids4spark-version-info.properties"
 
   def fixupConfigs(conf: SparkConf): Unit = {
     // First add in the SQL executor plugin because that is what we need at a minimum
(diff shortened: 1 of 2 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -86,22 +94,54 @@ object RapidsPluginUtils extends Logging {
...
+      throw new PluginException(s"Could not find properties file $resourceName in the classpath")
+    }
+    val props = new Properties
+    props.load(resource)
+    props
+  }
+
+  private def getVersionFromPropsFile(resourceName: String): String = {
+    val props = loadProps(resourceName)
+    Option(props.get("version")).map(_.toString).getOrElse {
```
abellina, line 143: Is this going to get called from the driver too? If so, I wonder if we should have an "undetected" state, in case cudf isn't in the class path of the driver.

jlowe: cudf has to be in the classpath of the driver.  The driver calls some cudf Java code that doesn't load libcudf.

abellina: Cool that makes sense

revans2 (dismissed):

jlowe:
build

abellina (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -52,7 +56,9 @@ case class ColumnarOverrideRules() extends ColumnarRule with Logging {
  */
 class SQLExecPlugin extends (SparkSessionExtensions => Unit) with Logging {
   override def apply(extensions: SparkSessionExtensions): Unit = {
-    logWarning("Installing extensions to enable rapids GPU SQL support." +
+    val pluginVersion = RapidsPluginUtils.getPluginVersion
+    val cudfVersion = RapidsPluginUtils.getClasspathCudfVersion
+    logWarning(s"RAPIDS Accelerator $pluginVersion using cudf $cudfVersion." +
```
gerashegalov, line 61: I would suggest to log the entire build info, i.e all properties:
```scala
scala> com.nvidia.spark.rapids.RapidsPluginUtils.getPluginProps
res2: java.util.Properties = {version=0.5.0-SNAPSHOT, user=gshegalov, url=git@github.com:NVIDIA/spark-rapids.git, date=2021-04-14T18:55:56Z, revision=122097be00609a525cc756a9fba707685c789710, cudf_version=0.19-SNAPSHOT, branch=log-version}
```
Something to the tune:
```scala
logWarning(s"RAPIDS Accelerator [buildInfo ${RapidsPluginUtils.getPluginProps}] ..." 
```
Also let us update https://github.com/NVIDIA/spark-rapids/blob/branch-0.5/docs/configs.md#rapids-accelerator-for-apache-spark-configuration on how to verify the plugin was loaded using logs and REPL

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -188,53 +228,26 @@ class RapidsExecutorPlugin extends ExecutorPlugin with Logging {
...
       if (!RapidsExecutorPlugin.cudfVersionSatisfied(expectedCudfVersion, cudfVersion)) {
-        throw CudfVersionMismatchException(s"Cudf version in the classpath is different. " +
-          s"Found $cudfVersion, RAPIDS Accelerator expects $expectedCudfVersion")
+        throw CudfVersionMismatchException(s"Found cudf version $cudfVersion, RAPIDS Accelerator " +
+            s"expects $expectedCudfVersion")
       }
+      val pluginVersion = Option(pluginProps.get("version")).map(_.toString).getOrElse {
+        throw new PluginException("No `version` property in plugin properties file")
+      }
+      logInfo(s"RAPIDS Accelerator $pluginVersion using cudf $cudfVersion")
```
gerashegalov, line 244: let us log all version props ,  see comment above

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -188,53 +228,26 @@ class RapidsExecutorPlugin extends ExecutorPlugin with Logging {
...
     } catch {
-      case x: CudfVersionMismatchException if conf.cudfVersionOverride =>
-        logWarning(s"${x.errorMsg}")
+      case x: PluginException if conf.cudfVersionOverride =>
+        logWarning(s"${x.getMessage}")
```
gerashegalov, line 247: can we add  some language here about exception being ignored due to cudfVersionOverride being on.

gerashegalov (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
jlowe: IMO this is _far_ too much detail and clogs up the output with user-confusing details.  Instead of a brief message showing the version, now I have to essentially parse a long line to figure out what versions are really used here.  For example, is the cudf version what `cudf_version` is reporting or the version later on the line after `using cudf`?  It's the latter, but how will the user know that?

I think it might be useful to log the git commit hash, possibly in parentheses, after the version number for both the plugin and cudf, but I'm not a fan of logging all the other stuff.  If we really need that, we can fish the properties file from the jar.

gerashegalov: I'd rather overlog. We don't need redundant `using cudf ` obviously if we agree to log all the properties.  Git commit is a reasonable compromise.  But given that it's OSS and there can be non-deterministic build related issues,  I'd rather see more than less diagnostics. Deferring customers to fishing in the jars might not be feasible in many environments.

jlowe: If we want to point to OSS, Spark has this properties file but does not dump it on startup.  I don't think we should either.

gerashegalov: > This appears on the driver startup every time. IMO it's unprofessional and excessive to log raw properties like this.

we can style/relabel them for clarity but I don't find it unprofessional and/or excessive  even as is.

gerashegalov: > If we want to point to OSS, Spark has this properties file but does not dump it on startup. I don't think we should either.

I meant that everyone can build OSS, so git commit and version is not sufficient to identify jars.  I was not talking about Spark as a reference.

jlowe: I kept the short version logging at the WARN level but now also log the full build props at INFO level on both the driver and executor.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -52,22 +56,31 @@ case class ColumnarOverrideRules() extends ColumnarRule with Logging {
  */
 class SQLExecPlugin extends (SparkSessionExtensions => Unit) with Logging {
   override def apply(extensions: SparkSessionExtensions): Unit = {
-    logWarning("Installing extensions to enable rapids GPU SQL support." +
+    val pluginProps = RapidsPluginUtils.loadProps(RapidsPluginUtils.PLUGIN_PROPS_FILENAME)
+    logInfo(s"RAPIDS Accelerator build: $pluginProps")
+    val cudfProps = RapidsPluginUtils.loadProps(RapidsPluginUtils.CUDF_PROPS_FILENAME)
+    logInfo(s"cudf build: $pluginProps")
```
abellina, line 62:
```suggestion
    logInfo(s"cudf build: $cudfProps")
```

jlowe: Doh!  Nice catch.  I checked the executor log but missed this in the driver log.

abellina (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
gerashegalov: Maybe for a followup PR: Instead of logWarning in interactive shells for Scala, Python, R we can `println` this string 
based on 
- main class in sun.java.command system property
- `spark.app.name` https://github.com/apache/spark/blob/60e324aa9f071260f130780fde7438911b0aecee/bin/pyspark#L68
- `spark.repl.class.outputDir` https://github.com/apache/spark/blob/3be552ccc8d26089881229edc034d2ebf2e75511/repl/src/main/scala-2.12/org/apache/spark/repl/Main.scala#L92
...

gerashegalov (approved):
LGTM

jlowe:
build

PR #2135, opened 2021-04-14
Disable Spark 3.2 shim by default

jlowe:
The fix for #2052 is going to take a long time, so in the meantime this disables the Spark 3.2 shim build by default.  It can still be built with the `spark320` profile.


jlowe:
build

tgravescs (approved):

gerashegalov (approved):

PR #2139, opened 2021-04-15
support truncate cuda version for image build [skip ci]

pxLi:
cuda image has add the update version to image tag since 11.2+, e.g.
11.2.<update_version>-runtime-ubuntu18.04 

Since cudf + cudatoolkit do not care about the update's version, we need to truncate it.

Tested locally.


pxLi:
build

NvTimLiu (approved):

PR #2124, opened 2021-04-14
init branch-0.6 [skip ci]

pxLi:
Signed-off-by: Peixin Li <pxli@nyu.edu>

initiate branch-0.6 w/ cudf  0.20-SNAPSHOT,
1. bump up plugin version to 0.6-SNAPSHOT
2. bump up cudf version to 0.20-SNAPSHOT 
3. updated Doc to use new versions plugin and cudf
4. removed cudf classifier from pom

and there are still some CUDA_CLASSIFIER configs left under jenkins folder, 
this will be cleaned up later after we finish adjusting our pipelines' setup

Please help review, thanks!
~~NOTE: currently some test cases are still failing~~ 
~~I may just skip CI and merge this one if it looks OK to unblock the following pipelines setup work~~
actually passed tests w/ latest cudfjni

PR #2142, opened 2021-04-15
Restore CUDA classifier handling to Jenkins scripts

jlowe:
Following up from #2124, this restores the CUDA classifier environment variable handling in the Jenkins scripts.  This shouldn't be needed as long as we're only supporting CUDA 11.x releases, but it will probably be needed when we start supporting future CUDA major versions.


tgravescs (approved):

jlowe:
build

Issue #2103, opened 2021-04-08
[FEA] Join support for passing ArrayType and StructType columns along when not join keys


tgravescs:
**Is your feature request related to a problem? Please describe.**
Add support for passing ArrayType and StructType columns along with Joins when its not one of the main join keys.

a customer ran into this issue.

tgravescs:
*Exec <ProjectExec> will run on GPU
      !Exec <SortMergeJoinExec> cannot run on GPU because unsupported data types in input: M#906, X#1033; unsupported data types in output: M#906, X#1033
        @Expression <AttributeReference> num#5 could run on GPU
        @Expression <AttributeReference> num#1037 could run on GPU

tgravescs:
note the error message here isn't very clear either so we may want to see if we can improve that for other operators if it applies.

tgravescs:
this is only handling sort merge join, shuffled hash join and broadcast hash join.

Issue #2134, opened 2021-04-14
[FEA] Support Array[Decimal] and struct of Decimal being passed along with join


tgravescs:
**Is your feature request related to a problem? Please describe.**
in https://github.com/NVIDIA/spark-rapids/issues/2103 we are adding support for arrays to be passed along with join (not the join key), but Array of Decimal was failing the tests so we need to investigate that further and figure out why it was failing.  I also left the struct of Decimals out of the check as it was easier, the struct of decimal tests were working so we just need to re-enable it and add the tests.

sperlingxx:
This issue is out of date. Just close it.

Issue #2133, opened 2021-04-14
[FEA] Join support for passing MapType columns along when not join keys


tgravescs:
**Is your feature request related to a problem? Please describe.**
Join support for passing MapType columns along when not join keys.

Currently joins don't allow this but it should work.  One issue here might be with Sort supporting if its a sort merge join.

jlowe:
Fixed by #3011

PR #2144, opened 2021-04-15
Allow array and struct types to pass thru when doing join

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/2103

This adds supports to SortMergeJoin, HashShuffleJoin, and broadcasthashjoin to allow array and struct types to pass thru when doing a Join. ie they can't be join keys.  Note this required adding support to BroadcastExchange as well.  I set the allow to be the same as other things like HashAggregate.  If other types can be supported we can do in followups.  I wanted to keep to what was easy for 0.5.

This does not cover array[Decimal] which seems to have a bug because it was failing tests: https://github.com/NVIDIA/spark-rapids/issues/2134

issue for adding map types as well:
https://github.com/NVIDIA/spark-rapids/issues/2133

I added tests for both positive case and make sure it fallsback to cpu when used as join keys.    I ran the tests on all the spark versions modified here including Databricks.

I didn't see any other testing of BroadcastExchange so assume it was just through the Join, if someone knows of tests I may have missed let me know.

### DIFF
--- a/docs/download.md
+++ b/docs/download.md
@@ -15,18 +15,14 @@ that is either preinstalled in the Spark classpath on all nodes or submitted wit
 RAPIDS Accelerator For Apache Spark. See the [getting-started guide](https://nvidia.github.io/spark-rapids/Getting-Started/) for more details.
 
 ## Release v0.4.1
+### Download v0.4.1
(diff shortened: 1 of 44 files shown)


tgravescs:
build

revans2 (approved):

PR #2151, opened 2021-04-15
add corresponding close call for HostMemoryOutputStream

wbo4958:


jlowe (approved):
Note that the close method for `HostMemoryOutputStream` doesn't do anything, so this PR should have net-zero effect.  Still it's useful to close at the appropriate place just in case semantics are added to the close method later. :smile:

gerashegalov:
given Jason's comment it's irrelevant but technically you want `close` inside a finally block, maybe using withResource

sameerz:
build

pxLi:
build

pxLi:
build

PR #2166, opened 2021-04-16
Script for auto prioritizing audit

razajafri:
This is a first step in trying to auto prioritize the audit commits. Currently it looks at all the changes to a file and determines if spark-rapids depends on it if it does then it flags the commit as `P1` otherwise it flags it as `P?` 

Future improvements: 
* This version isn't smart enough to differentiate between code changes and comment changes. Due to this there are many false positives i.e. many commits are flagged as P1 even though they aren't, but I have not seen any false negatives which is good
* This can be aided by a visual tool where we can push the changes directly. The devs can just go to the site and audit the commits using a GUI


razajafri:
@nartal1 PTAL

scripts/audit-plugin.sh
```diff
@@ -0,0 +1,101 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# This script analyzes the hashCommitsWithMessages.log generated by scripts/audit-spark*.sh
+# and determines which commits are directly modifying classes that we use in our plugin. Commits
```
tgravescs, line 19: what does "classes that we use in our plugin". mean?  There are different types of things we do.
1) directly copy classes and functions (ie functions aren't classes and we may have moved to other classes).
2) We have classes very similar ie cpu SortExec => GpuSortExec and we may have functionality very similar in there.

it would be nice to clarify exactly what this script does.

razajafri: this script will see if we have a dependency on that class that was modified in the spark commit. I will clarify it in the comment

tgravescs: what do you mean by dependency?  we call it or import it?  This would all be good information to add to the description as well.

scripts/audit-plugin.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
+#
+# Copyright (c) 2019-2021, NVIDIA CORPORATION. All rights reserved.
```
nartal1, line 3: Copyright year 2021

scripts/audit-plugin.sh
```diff
@@ -0,0 +1,102 @@
...
```
tgravescs, line 3: shouldn't have 2019

scripts/audit-plugin.sh
```diff
@@ -0,0 +1,101 @@
...
+    if [ ${FOUND} -ne 0 ]; then 
+      printf "%s\t%s\t%s\t%s\n" "P1" "${COMMIT_ID}" "${COMMIT_MSG}" "${CLASS}" >> ${PRIORITIZED_COMMITS}
+      break
+    fi
+  fi
+  done
+  if [ ${FOUND} -eq 0 ]; then
+    printf "%s\t%s\t%s\n" "P?" "${COMMIT_ID}" "${COMMIT_MSG}" >> ${PRIORITIZED_COMMITS}
+  fi
+done < <(cat ${WORKSPACE}/hashCommitsWithMessage.log)
```
nartal1, line 101: hashCommitsWithMessage.log is used to get the initial set of commits and not for incremental set of changes. I think we  should use $COMMIT_DIFF_LOG instead.

razajafri: is that an env variable expected by `audit-spark-3.2.sh` because I don't see it being set in that script

razajafri: Also, what about the first time we run the script? We need to get both initial and subsequent runs.

nartal1: Yes, it's environment variable setup in jenkins job. 
Thanks for pointing it out. I think we could write to $COMMIT_DIFF_LOG for both the cases and read that file in this script.

razajafri:
@nartal1 Can you PTAL I have updated the script. I have run it on my local box and it seems to be doing the job but since you are more familiar with jenkins env it would be nice if you could verify the script runs without any errors there.

nartal1:
> @nartal1 Can you PTAL I have updated the script. I have run it on my local box and it seems to be doing the job but since you are more familiar with jenkins env it would be nice if you could verify the script runs without any errors there.

Sure @razajafri. We might have to make some changes in the jenkins script. Will verify and put up a PR for that.

razajafri:
@nartal1 have you been able to verify this?

nartal1:
> @nartal1 have you been able to verify this?

Yes, I checked out your branch and verifying. Was planning to add review comments all at once  when I get it to pass. 
It's failing on `nvidia-smi` not found in CI. I think that's not required for this script as we are not testing anything specifically on GPU. Making changes to Jenkins file.

scripts/audit-plugin.sh
```diff
@@ -0,0 +1,110 @@
...
+#Set env variables
+. jenkins/version-def.sh
+
+ABSOLUTE_PATH=$(cd $(dirname $0) && pwd)
+AUDIT_PLUGIN_LOG=${ABSOLUTE_PATH}/audit-plugin.log
+if [ -e ${AUDIT_PLUGIN_LOG}]; then 
+  rm ${AUDIT_PLUGIN_LOG}
+fi
+
+PRmJECT_VER="0.5.0-SNAPSHOT"
```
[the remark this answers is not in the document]
razajafri: we don't need this variable. I added this only for testing

nartal1:
build

nartal1 (approved):

PR #2169, opened 2021-04-16
Work around bug in Spark for integration test

revans2:
This is a work around for https://issues.apache.org/jira/browse/SPARK-35108

It does this by making it so that we never have structs that can have the same data, but different keys.


revans2:
build

gerashegalov (approved):

jlowe (approved):

PR #2179, opened 2021-04-19
cuda11.0 toolkit for the Databricks integrationt test

NvTimLiu:
cuda11.0 toolkit has been installed via CONDA in the Databricks init script file: 'init_cudf_udf.sh'

We only need to export the cuda11.0 toolkit path using 'LD_LIBRARY_PATH'

### DIFF
--- a/jenkins/databricks/test.sh
+++ b/jenkins/databricks/test.sh
@@ -22,6 +22,9 @@ SPARK_CONF=$2
 
 # tests
 export PATH=/databricks/conda/envs/databricks-ml-gpu/bin:/databricks/conda/condabin:$PATH
+## cuda toolkit has been installed into conda envs via 'init_cudf_udf.sh'
+export LD_LIBRARY_PATH=/databricks/conda/envs/databricks-ml-gpu/lib
+
 sudo /databricks/conda/envs/databricks-ml-gpu/bin/pip install pytest sre_yield requests pandas \
 	pyarrow findspark pytest-xdist
 


jenkins/databricks/test.sh
```diff
@@ -22,6 +22,9 @@ SPARK_CONF=$2
 
 # tests
 export PATH=/databricks/conda/envs/databricks-ml-gpu/bin:/databricks/conda/condabin:$PATH
+## cuda toolkit has been installed into conda envs via 'init_cudf_udf.sh'
+export LD_LIBRARY_PATH=/databricks/conda/envs/databricks-ml-gpu/lib
```
NvTimLiu, line 26: conda install CUDA11 here:

https://github.com/NVIDIA/spark-rapids/blob/branch-0.6/jenkins/databricks/init_cudf_udf.sh#L29

NvTimLiu:
build

jenkins/databricks/test.sh
tgravescs: why do we need this, the dynamic linker isn't finding it automatically?

NvTimLiu: This should be conda install an issue

conda installation CLI https://github.com/NVIDIA/spark-rapids/blob/branch-0.6/jenkins/databricks/init_cudf_udf.sh#L29 
only installs cuda11 runtime jars into the its evns dir: `/databricks/conda/envs/databricks-ml-gpu/lib`

No cuda11 dynamic linker /usr/local/cuda11.0, no ldconfig files  /etc/ld.so.conf.d/cuda-11-0.conf

So we need manually add LD_LIBRARY_PATH here.

tgravescs: ok makes sense

tgravescs: but this is used for all integration tests right?   I assume this isn't how the customer would installed cuda 11, so why aren't we just installed it like that?
if the init script has more then just cudf_udf tests just in there we should modify the name.
Also is this script manually put there (dbfs:/databricks/init_scripts/init_cudf_udf.sh) I didn't see anything that was copying it but maybe its in the other repo

tgravescs: if we just install it with commands I sent you, does that cause issues for the cudf UDF tests or something?

NvTimLiu: > but this is used for all integration tests right? I assume this isn't how the customer would installed cuda 11, so why aren't we just installed it like that?
> if the init script has more then just cudf_udf tests just in there we should modify the name.
> Also is this script manually put there (dbfs:/databricks/init_scripts/init_cudf_udf.sh) I didn't see anything that was copying it but maybe its in the other repo

@tgravescs We copy init scripts to  dbfs in Jenkinsfile via CLI, do you know some python APIs to do it?
https://gitlab-master.nvidia.com/nvspark/blossom-jenkins/-/blob/main/spark-rapids/Jenkinsfile.databricks301nightly#L80

NvTimLiu: > if we just install it with commands I sent you, does that cause issues for the cudf UDF tests or something?

Yes, we can install cuda11 in another init script, it should work, too. Let me check again.

NvTimLiu: > > if we just install it with commands I sent you, does that cause issues for the cudf UDF tests or something?
> 
> Yes, we can install cuda11 in another init script, this way should be more like customer situation to install cuda11. It should not cause any issue for the cudf UDF, either. Let me check again.

tgravescs: ok, missed the copy when I grepped the code.  It can be same script but lets rename it so it doesn't seem like its specific to the cudf_udf test.

Issue #2042, opened 2021-03-29
[BUG] NDS q14a fails with "GpuColumnarToRow does not implement doExecuteBroadcast"


andygrove:
**Describe the bug**

NDS q14a fails with:

```
Caused by: java.lang.UnsupportedOperationException: GpuColumnarToRow does not implement doExecuteBroadcast
	at org.apache.spark.sql.execution.SparkPlan.doExecuteBroadcast(SparkPlan.scala:298)
	at org.apache.spark.sql.execution.InputAdapter.doExecuteBroadcast(WholeStageCodegenExec.scala:515)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeBroadcast$1(SparkPlan.scala:193)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeQuery$1(SparkPlan.scala:218)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
	at org.apache.spark.sql.execution.SparkPlan.executeQuery(SparkPlan.scala:215)
	at org.apache.spark.sql.execution.SparkPlan.executeBroadcast(SparkPlan.scala:189)
	at org.apache.spark.sql.execution.joins.BroadcastHashJoinExec.prepareBroadcast(BroadcastHashJoinExec.scala:203)
	at org.apache.spark.sql.execution.joins.BroadcastHashJoinExec.prepareRelation(BroadcastHashJoinExec.scala:217)
```

**Steps/Code to reproduce bug**

- Spark 3.0.2 or 3.1.1
- SF=100 or SF=1000
- AQE on
- Decimal support disabled
- NDS data with Decimals in it

**Expected behavior**

Query should not fail.

jlowe:
@andygrove can you run with `spark.rapids.sql.explain=ALL`?  I'm curious what the plans look like that could cause GpuColumnarToRow to be invoked in this context.  I'm assuming this query runs without problems with AQE=off, is that the case?

andygrove:
Correct, the query works fine with AQE off. It also works fine with AQE on and decimal support enabled. The full log is quite large (6MB) but this portion may be helpful:

```
Caused by: java.util.concurrent.ExecutionException: org.apache.spark.sql.catalyst.errors.package$TreeNodeException: execute, tree:
ShuffleQueryStage 3
+- Exchange SinglePartition, ENSURE_REQUIREMENTS, [id=#18032]
[...]

PR #2188, opened 2021-04-19
Support GPU broadcast exchange reuse to feed CPU BHJ when AQE is enabled

andygrove:

This PR makes it possible for a GPU exchange to be re-used by ReuseExchange even when it is feeding a CPU join. A specialized columnar-to-row transition is inserted that converts the data to the serialized format that the CPU expects.

~WIP because I am concerned that this introduces a resource leak and I am still looking into that.~

I also need to figure out what we can do for testing. I manually tested this by running NDS q14a.

Closes https://github.com/NVIDIA/spark-rapids/issues/2042

### DIFF
--- a/shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuBroadcastExchangeExec.scala
+++ b/shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuBroadcastExchangeExec.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2021, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuBroadcastExchangeExec.scala
+++ b/shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuBroadcastExchangeExec.scala
@@ -20,7 +20,7 @@ import org.apache.spark.sql.execution.SparkPlan
 import org.apache.spark.sql.rapids.execution.GpuBroadcastExchangeExecBase
 
 case class GpuBroadcastExchangeExec(
-    mode: BroadcastMode,
+    override val mode: BroadcastMode,
     child: SparkPlan) extends GpuBroadcastExchangeExecBase(mode, child) {
 
   override def doCanonicalize(): SparkPlan = {
--- a/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/GpuBroadcastExchangeExec.scala
+++ b/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/GpuBroadcastExchangeExec.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2021, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/GpuBroadcastExchangeExec.scala
+++ b/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/GpuBroadcastExchangeExec.scala
@@ -24,7 +24,7 @@ import org.apache.spark.sql.execution.exchange.BroadcastExchangeLike
 import org.apache.spark.sql.rapids.execution.GpuBroadcastExchangeExecBase
 
 case class GpuBroadcastExchangeExec(
-    mode: BroadcastMode,
+    override val mode: BroadcastMode,
     child: SparkPlan) extends GpuBroadcastExchangeExecBase(mode, child) with BroadcastExchangeLike {
 
   override def runId: UUID = _runId
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVector.java
@@ -63,7 +63,7 @@ public final class RapidsHostColumnVector extends RapidsHostColumnVectorCore {
   /**
    * Sets up the data type of this column vector.
    */
-  RapidsHostColumnVector(DataType type, ai.rapids.cudf.HostColumnVector cudfCv) {
+  public RapidsHostColumnVector(DataType type, ai.rapids.cudf.HostColumnVector cudfCv) {
     super(type, cudfCv);
     // TODO need some checks to be sure everything matches
     this.cudfCv = cudfCv;
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVector.java
@@ -1,6 +1,6 @@
 
 /*
- * Copyright (c) 2020, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2021, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
@@ -166,7 +166,7 @@ class ColumnarToRowIterator(batches: Iterator[ColumnarBatch],
   @transient var cb: ColumnarBatch = null
   var it: java.util.Iterator[InternalRow] = null
 
-  Option(TaskContext.get()).foreach(_.addTaskCompletionListener[Unit](_ => closeCurrentBatch()))
+  TaskContext.get().addTaskCompletionListener[Unit](_ => closeCurrentBatch())
 
   private def closeCurrentBatch(): Unit = {
     if (cb != null) {
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
@@ -166,7 +166,7 @@ class ColumnarToRowIterator(batches: Iterator[ColumnarBatch],
   @transient var cb: ColumnarBatch = null
   var it: java.util.Iterator[InternalRow] = null
 
-  TaskContext.get().addTaskCompletionListener[Unit](_ => closeCurrentBatch())
+  Option(TaskContext.get()).foreach(_.addTaskCompletionListener[Unit](_ => closeCurrentBatch()))
 
   private def closeCurrentBatch(): Unit = {
     if (cb != null) {
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
@@ -31,7 +31,7 @@ import org.apache.spark.sql.execution.datasources.v2.DataSourceV2ScanExecBase
 import org.apache.spark.sql.execution.exchange.{Exchange, ReusedExchangeExec, ShuffleExchangeExec}
 import org.apache.spark.sql.execution.joins.{BroadcastHashJoinExec, BroadcastNestedLoopJoinExec}
 import org.apache.spark.sql.rapids.{GpuDataSourceScanExec, GpuFileSourceScanExec, GpuInputFileBlockLength, GpuInputFileBlockStart, GpuInputFileName, GpuShuffleEnv}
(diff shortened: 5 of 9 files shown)


andygrove:
@revans2 @tgravescs I would appreciate a review to make sure I understood this correctly.

tgravescs:
without me having looked at the details is there a reason we use a separate GpuBroadcastColumnarToRow vs just implementing doExecuteBroadcast in GpuColumnarToRow?  Would be good to state that in description.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastColumnarToRowExec.scala
```diff
@@ -0,0 +1,204 @@
+package org.apache.spark.sql.rapids.execution
```
revans2, line 1: Need copyright and license header.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastColumnarToRowExec.scala
```diff
@@ -0,0 +1,204 @@
...
+    val broadcastTime = gpuLongMetric(BUILD_TIME)
+    val totalTime = gpuLongMetric(TOTAL_TIME)
+
+    val task = new Callable[Broadcast[Any]]() {
+      override def call(): Broadcast[Any] = {
+        // This will run in another thread. Set the execution id so that we can connect these jobs
+        // with the correct execution.
+        SQLExecution.withExecutionId(sqlContext.sparkSession, executionId) {
+          val totalRange = new MetricRange(totalTime)
+          try {
```
revans2, line 86: Lets use withResources if possible

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastColumnarToRowExec.scala
```diff
@@ -0,0 +1,204 @@
...
+            // Setup a job group here so later it may get cancelled by groupId if necessary.
+            sparkContext.setJobGroup(_runId.toString, s"broadcast exchange (runId ${_runId})",
+              interruptOnCancel = true)
+
+            val f = GpuColumnarToRowExecParent
+                .makeIteratorFunc(broadcast.child.output, numOutputRows, NoopMetric, totalTime)
+
+            val collectRange = new NvtxWithMetrics("broadcast collect", NvtxColor.GREEN,
+              collectTime)
+            val serializedBatch: SerializeConcatHostBuffersDeserializeBatch = try {
```
revans2, line 96: I don't think this is going to work. `SerializeConcatHostBuffersDeserializeBatch` will deserialize the data to the GPU.  This code runs on the driver, which is not guaranteed to have a GPU.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastColumnarToRowExec.scala
```diff
@@ -0,0 +1,204 @@
...
+        // This will run in another thread. Set the execution id so that we can connect these jobs
+        // with the correct execution.
+        SQLExecution.withExecutionId(sqlContext.sparkSession, executionId) {
+          val totalRange = new MetricRange(totalTime)
+          try {
+            // Setup a job group here so later it may get cancelled by groupId if necessary.
+            sparkContext.setJobGroup(_runId.toString, s"broadcast exchange (runId ${_runId})",
+              interruptOnCancel = true)
+
+            val f = GpuColumnarToRowExecParent
```
revans2, line 91: This runs code on the GPU, but this code is running on the driver that is not guaranteed to have a GPU. If we can deserialize the concatenated batch to an array of HostColumnVector instances, then we can probably make this work.

andygrove: @revans2 I've updated the code so that the driver collects the serialized batches and then creates host column vectors. Could you review that part to see if I am on the right track?

That part of the code appears to work correctly but now I hit a failure when trying to create the hashed relation because the code is expecting to see `UnsafeRow` rather than `InternalRow`. I don't know if that means this whole approach is not going to work or not.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
```diff
@@ -166,7 +166,7 @@ class ColumnarToRowIterator(batches: Iterator[ColumnarBatch],
   @transient var cb: ColumnarBatch = null
   var it: java.util.Iterator[InternalRow] = null
 
-  TaskContext.get().addTaskCompletionListener[Unit](_ => closeCurrentBatch())
+  Option(TaskContext.get()).foreach(_.addTaskCompletionListener[Unit](_ => closeCurrentBatch()))
```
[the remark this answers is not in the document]
tgravescs: what were your finding here?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -135,7 +135,11 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
     // in future query stages. Note that because these query stages have already executed, we
     // don't need to recurse down and optimize them again
     case ColumnarToRowExec(e: BroadcastQueryStageExec) =>
-      getColumnarToRowExec(e)
+      e.plan match {
+        case ReusedExchangeExec(_, b: GpuBroadcastExchangeExecBase) =>
```
tgravescs, line 139: comment here would be nice

andygrove: done

andygrove:
@tgravescs @revans2 This is ready for re-review. This is what the plan looks like now:

gpu2cpu

sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVector.java
```diff
@@ -63,7 +63,7 @@
   /**
    * Sets up the data type of this column vector.
    */
-  RapidsHostColumnVector(DataType type, ai.rapids.cudf.HostColumnVector cudfCv) {
+  public RapidsHostColumnVector(DataType type, ai.rapids.cudf.HostColumnVector cudfCv) {
```
revans2, line 66: needs copyright update

shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/GpuBroadcastExchangeExec.scala
```diff
@@ -24,7 +24,7 @@ import org.apache.spark.sql.execution.exchange.BroadcastExchangeLike
 import org.apache.spark.sql.rapids.execution.GpuBroadcastExchangeExecBase
 
 case class GpuBroadcastExchangeExec(
-    mode: BroadcastMode,
+    override val mode: BroadcastMode,
```
revans2, line 27: needs copyright update

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuBroadcastExchangeExec.scala
```diff
@@ -20,7 +20,7 @@ import org.apache.spark.sql.execution.SparkPlan
 import org.apache.spark.sql.rapids.execution.GpuBroadcastExchangeExecBase
 
 case class GpuBroadcastExchangeExec(
-    mode: BroadcastMode,
+    override val mode: BroadcastMode,
```
revans2, line 23: needs copyright update

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastToCpuExec.scala
```diff
@@ -0,0 +1,238 @@
...
+import org.apache.spark.util.KnownSizeEstimation
+
+/**
+ * This is a specialized version of GpuColumnarToRow that wraps a GpuBroadcastExchange and
+ * converts the columnar results containing cuDF tables into Spark rows so that the results
+ * can feed a CPU BroadcastHashJoin. This is required for exchange reuse in AQE.
+ *
+ * @param child Input to broadcast
+ */
+case class GpuBroadcastToCpuExec(mode: BroadcastMode, child: SparkPlan)
```
revans2, line 51: Can we have a follow on issue to look at how much of the code we can combine with `GpuBroadcastExchangeExec`. The reason I don't ask to do it here is because this is a P1 issue, and I don't want to get in the way of shipping.

Inside `GpuBroadcastExchangeExec` there are two APIs `doExecuteBroadcast` and `executeColumnarBroadcast`. Our code always calls the second one. The CPU code calls the first one but it is implemented to do exactly the same thing. At a minimum the first one should throw an exception but we probably could have it do the same thing that is happening here.  We might even be able to figure out a way so we broadcast the data twice, but we don't execute the broadcast twice.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastToCpuExec.scala
```diff
@@ -0,0 +1,238 @@
...
+  val _runId: UUID = UUID.randomUUID()
+
+  @transient
+  lazy val relationFuture: Future[Broadcast[Any]] = {
+    // relationFuture is used in "doExecute". Therefore we can get the execution id correctly here.
+    val executionId = sparkContext.getLocalProperty(SQLExecution.EXECUTION_ID_KEY)
+    val numOutputRows = gpuLongMetric(NUM_OUTPUT_ROWS)
+    val collectTime = gpuLongMetric(COLLECT_TIME)
+    val buildTime = gpuLongMetric(BUILD_TIME)
+    val broadcastTime = gpuLongMetric(BUILD_TIME)
```
revans2, line 96: Not BUILD_TIME

revans2 (commented):
Really just some nits and a follow on issue, unless you want to make the changes now instead.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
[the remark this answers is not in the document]
andygrove: I was able to remove this after the refactor to avoid using GPU on the driver.

andygrove:
Thanks for the review @revans2. I pushed some commits to fix the simple feedback items and filed https://github.com/NVIDIA/spark-rapids/issues/2219 for a refactor to reduce code duplication.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastToCpuExec.scala
```diff
@@ -0,0 +1,238 @@
...
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.vectorized.ColumnarBatch
+import org.apache.spark.util.KnownSizeEstimation
+
+/**
+ * This is a specialized version of GpuColumnarToRow that wraps a GpuBroadcastExchange and
+ * converts the columnar results containing cuDF tables into Spark rows so that the results
+ * can feed a CPU BroadcastHashJoin. This is required for exchange reuse in AQE.
+ *
+ * @param child Input to broadcast
```
tgravescs, line 49: add mode param here

tgravescs:
just to note, I tried out a version of this with Q14a and that query took about 2.2 minutes on my setup vs 1.8 on the CPU. With the Spark fix we have proposed it took 1.2 minute.  I'm assuming the difference in time there was in a CPU broadcasthashjoin vs a GPU broadcasthashjoin.
the explain text output in the UI also seems not show the GPUBroadcastToCpu but the graph at the top does.  That might be a spark issue though.

andygrove:
> the explain text output in the UI also seems not show the GPUBroadcastToCpu but the graph at the top does. That might be a spark issue though.

This issue seems specific to Spark 3.1.1 and works fine with 3.0.2 so it does seem to be a regression in Spark somehow, although I have not finished investigating this.

revans2:
Things look good I didn't see a follow on issue for code cleanup, and I would like to see us dig into the performance issues that Tom saw
https://github.com/NVIDIA/spark-rapids/pull/2188#issuecomment-824115676

tgravescs:
Yeah I'm wondering if there is a way we can keep things GPU based on what I saw on the plans  in the runs with spark fix, but we can do as followup.

andygrove:
Here's the issue for code cleanup: https://github.com/NVIDIA/spark-rapids/issues/2219

tgravescs (approved):

andygrove:
build

andygrove:
I looked into the explain issue some more and setting `spark.sql.ui.explainMode` to `extended` causes Spark to just call toString on the plan which was the original behavior and this does correctly show the new operator. Setting to `formatted` (the default) does not.

Note that even though this is a UI setting, it actually has to be set before running the query because it affects the data that actually gets recorded in the event log.

andygrove:
Test failed due to `Temporary failure in name resolution: Unknown host` with an internal host. Will retry.

andygrove:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastToCpuExec.scala
andygrove: Here is the follow up issue to remove code duplication: https://github.com/NVIDIA/spark-rapids/issues/2219

PR #2190, opened 2021-04-19
JNI fixes for StringWordCount native UDF example

jlowe:
This fixes two problems in the StringWordCount native UDF example:
- Java class moved to the com.nvidia.spark.rapids.udf.hive package but corresponding JNI function names were not updated
- The JNI function could fail to return a value if an error occurred


jlowe:
build

revans2 (approved):

PR #2191, opened 2021-04-19
Use CPM to fetch libcudf dependency for native UDF example build

jlowe:
This simplifies the native UDF build by using CPM to pull in libcudf rather than manually building it in a Docker image.  This work was heavily influenced by the RAPIDS example project using libcudf via CPM at https://github.com/rapidsai/rapids-examples/tree/main/strings_udf.

### DIFF
--- a/udf-examples/Dockerfile
+++ b/udf-examples/Dockerfile
@@ -16,62 +16,55 @@
 
 # A container that can be used to build UDF native code against libcudf
 ARG CUDA_VERSION=11.0
-ARG CUDA_SHORT_VERSION=${CUDA_VERSION}
 ARG LINUX_VERSION=ubuntu18.04
-ARG CUDF_REPO=https://github.com/rapidsai/cudf
-ARG CUDF_TAG=branch-0.20
-ARG CC=7
-ARG CXX=7
 
 FROM nvidia/cuda:${CUDA_VERSION}-devel-${LINUX_VERSION}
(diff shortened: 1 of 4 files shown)


jlowe:
build

udf-examples/Dockerfile
```diff
@@ -16,62 +16,55 @@
...
+ENV PARALLEL_LEVEL=10
 
-# Install conda
-ADD https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh /miniconda.sh
-RUN sh /miniconda.sh -b -p /conda && /conda/bin/conda update -n base conda
-ENV PATH=${PATH}:/conda/bin
-# Enables "source activate conda"
-SHELL ["/bin/bash", "-c"]
+RUN GCC_VERSION=$(bash -c '\
+CUDA_VERSION=$(nvcc --version | head -n4 | tail -n1 | cut -d" " -f5 | cut -d"," -f1); \
```
revans2, line 28: nit: having CUDA_VERSION in the shell script can be a little confusing with it is also an ARG to the Dockerfile.

revans2 (approved):
Looks fine just one nit, that is not a big deal

udf-examples/Dockerfile
jlowe: If it's just a minor nit, I'd like to keep this in-step with the original RAPIDS example since this came verbatim from there.

Issue #2203, opened 2021-04-20
[BUG] Spark nightly cache tests fail with -- master flag


razajafri:
**Describe the bug**
When submitting cache tests in the integration tests [here](https://blossom.nvidia.com/sw-gpu-spark-jenkins/job/rapids_integration-dev-github/) with command 

```
spark-submit --master spark://raza-linux-2:7077 --executor-memory 12G --total-executor-cores 6 --conf spark.sql.shuffle.partitions=12 --conf spark.driver.extraClassPath=/home/rjafri/audit-plugin/jars/cudf-0.20-SNAPSHOT-cuda11.jar:/home/rjafri/audit-plugin/jars/rapids-4-spark_2.12-0.6.0-SNAPSHOT.jar:/home/rjafri/audit-plugin/jars/rapids-4-spark-udf-examples_2.12-0.6.0-SNAPSHOT.jar --conf spark.executor.extraClassPath=/home/rjafri/audit-plugin/jars/cudf-0.20-SNAPSHOT-cuda11.jar:/home/rjafri/audit-plugin/jars/rapids-4-spark_2.12-0.6.0-SNAPSHOT.jar:/home/rjafri/audit-plugin/jars/rapids-4-spark-udf-examples_2.12-0.6.0-SNAPSHOT.jar --conf spark.driver.extraJavaOptions=-Duser.timezone=UTC --conf spark.executor.extraJavaOptions=-Duser.timezone=UTC --conf spark.sql.session.timeZone=UTC --conf spark.sql.cache.serializer=com.nvidia.spark.rapids.shims.spark311.ParquetCachedBatchSerializer --jars /home/rjafri/audit-plugin/jars/rapids-4-spark-integration-tests_2.12-0.6.0-SNAPSHOT.jar ./runtests.py -v -rfExXs --std_input_path=/home/rjafri/audit-plugin/integration_tests/src/test/resources/ -k cache_test.py
```

They fail due to an exception. 

``` E                   py4j.protocol.Py4JJavaError: An error occurred while calling o282.collectToPython.
E                   : org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 4.0 failed 4 times, most recent failure: Lost task 0.3 in stage 4.0 (TID 33) (192.168.1.151 executor 0): java.lang.NullPointerException
E                   	at org.apache.spark.sql.internal.SQLConf.setConfWithCheck(SQLConf.scala:3856)
[...]

PR #2204, opened 2021-04-20
Run nightly tests for ParquetCachedBatchSerializer

razajafri:
~~Cache tests fail when the jenkins server run them with `--master` flag. This will fix the error by removing the flag. Further investigation is needed to know why it fails with the flag it's being tracked in #2203.~~

This PR is being retrofitted to push the changes needed for enabling the nightly cache tests. 


tgravescs:
no, I don't want to check by removing master here... figure out what is wrong or just revert the change.

razajafri:
> no, I don't want to check by removing master here... figure out what is wrong or just revert the change.
> it would be confusing to run this test differently then the rest of the tests.

reverting until I narrow down this problem https://github.com/NVIDIA/spark-rapids/pull/2208

tgravescs:
note changes were reverted in https://github.com/NVIDIA/spark-rapids/pull/2208/files

razajafri:
This will be ready after #2227 is merged

razajafri:
build

jenkins/spark-tests.sh
```diff
@@ -93,6 +93,12 @@ pushd $RAPIDS_INT_TESTS_HOME
 TEST_TYPE="nightly"
 spark-submit $BASE_SPARK_SUBMIT_ARGS --jars $RAPIDS_TEST_JAR ./runtests.py -v -rfExXs --std_input_path="$WORKSPACE/integration_tests/src/test/resources/" --test_type=$TEST_TYPE
 spark-submit $BASE_SPARK_SUBMIT_ARGS $CUDF_UDF_TEST_ARGS --jars $RAPIDS_TEST_JAR ./runtests.py -m "cudf_udf" -v -rfExXs --cudf_udf --test_type=$TEST_TYPE
+#only run cache tests with our serializer in nightly test for Spark version >= 3.1.1
+if [ "$(printf '%s\n' "3.1.1" "$SPARK_VER" | sort -V | head -n1)" = "3.1.1" ]; then
```
revans2, line 97: 1. Why exactly is it that we should not run the cache tests unless the version is 3.1.1 or later?
2. Why is it OK to run them in the pre-commit tests but not in the nightly tests?
3. Why not change the test itself to skip things if the versions are wrong/it is configured incorrectly.  We already check versions in test https://github.com/NVIDIA/spark-rapids/blob/6274fadf2393ebf7d47ead7e322e672776cbefc1/integration_tests/src/main/python/cache_test.py#L118-L119

razajafri: 1. Because we are already running the `cache_test.py` as part of the integration tests, whats the point of running it again.
2. Its completely OK and we are running `cache_test.py` as part of the precommit. May be we need another PR to address the precommit test so we run the Serializer test as well. 
3. We actually don't want to skip the tests as we want to run them with the default serializer

I hope I have answered the questions if I have misunderstood please let me know and I will try to understand better

revans2: Then how exactly does this "Fixed nightly build failure in cache tests"?

We are running a test again, this time with a slightly different configuration and that some how makes the existing test stop failing?

razajafri: Sorry, I forgot to change the title when I changed the description. 

The tests weren't enabled for the serializer, this PR will run them now

revans2 (approved):

PR #2209, opened 2021-04-20
Fixed a few issues with out of core sort

revans2:
This fixes an off by one error when and entire batch is already sorted. It fixes an issue when inserting batches into the pending queue, and it fixes an issue when sorting only rows, no columns. Any of what could lead to data corruption.

### DIFF
--- a/integration_tests/src/main/python/sort_test.py
+++ b/integration_tests/src/main/python/sort_test.py
@@ -163,7 +163,7 @@ def test_single_orderby_with_skew(data_gen):
             conf = allow_negative_scale_of_decimal_conf)
 
 # This is primarily to test the out of core sort with multiple batches. For this we set the data size to
-# be realatively large (1 MiB accross all tasks) and the target size to be small 16 KiB. This means we
+# be relatively large (1 MiB across all tasks) and the target size to be small (16 KiB). This means we
 # should see around 64 batches of data. So this is the most valid if there are less than 64 tasks
 # in the cluster, but it should still work even then.
 def test_large_orderby():
--- a/integration_tests/src/main/python/sort_test.py
+++ b/integration_tests/src/main/python/sort_test.py
@@ -161,3 +161,13 @@ def test_single_orderby_with_skew(data_gen):
                     .orderBy(f.col('a'))\
                     .selectExpr('a'),
             conf = allow_negative_scale_of_decimal_conf)
+
+# This is primarily to test the out of core sort with multiple batches. For this we set the data size to
+# be realatively large (1 MiB accross all tasks) and the target size to be small 16 KiB. This means we
+# should see around 64 batches of data. So this is the most valid if there are less than 64 tasks
+# in the cluster, but it should still work even then.
+def test_large_orderby():
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : unary_op_df(spark, long_gen, length=1024*128)\
+                    .orderBy(f.col('a')),
+            conf = {'spark.rapids.sql.batchSizeBytes': '16384'})
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
@@ -295,13 +295,14 @@ case class GpuOutOfCoreSortIterator(
         sorted.add(sp)
       }
     } else {
-      val splitIndexes = if (sortedOffset > 0) {
+      val hasFullySortedData = sortedOffset > 0
+      val splitIndexes = if (hasFullySortedData) {
         sortedOffset until rows by targetRowCount
       } else {
         targetRowCount until rows by targetRowCount
       }
       // Get back the first row so we can sort the batches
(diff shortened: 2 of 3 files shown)


revans2:
build

integration_tests/src/main/python/sort_test.py
```diff
@@ -161,3 +161,13 @@ def test_single_orderby_with_skew(data_gen):
                     .orderBy(f.col('a'))\
                     .selectExpr('a'),
             conf = allow_negative_scale_of_decimal_conf)
+
+# This is primarily to test the out of core sort with multiple batches. For this we set the data size to
+# be realatively large (1 MiB accross all tasks) and the target size to be small 16 KiB. This means we
```
jlowe, line 166:
```suggestion
# be relatively large (1 MiB across all tasks) and the target size to be small (16 KiB). This means we
```

abellina:
```suggestion
# be relatively large (1 MiB across all tasks) and the target size to be small 16 KiB. This means we
```

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -296,13 +295,13 @@ case class GpuOutOfCoreSortIterator(
         sorted.add(sp)
       }
     } else {
-      val splitIndexes = if (sortedOffset >= 0) {
+      val splitIndexes = if (sortedOffset > 0) {
```
abellina, line 298: nit, since we use `sortedOffset > 0` in three different places, maybe it would be good to create a `val` above this line and use that? 

```
val offsetIsNonZero = sortedOffset > 0
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -418,7 +417,7 @@ case class GpuOutOfCoreSortIterator(
         // First we want figure out what is fully sorted from what is not
         val sortSplitOffset = if (pending.isEmpty) {
```
abellina, line 418: do we need `sortSplitOffset`? Seems like instead it cold be replaced with `pending.isEmpty` down in the next if statement, and could remove this if statement.

abellina (changes requested):
small comments

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
revans2: That would not cover the case where pending is not empty, but the next batch in pending does not overlap with what we just sorted.

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -284,7 +283,7 @@ case class GpuOutOfCoreSortIterator(
     // Protect ourselves from large rows when there are small targetSizes
     val targetRowCount = Math.max((targetBatchSize/averageRowSize).toInt, 1024)
 
-    if (sortedOffset == rows - 1) {
+    if (sortedOffset == rows) {
```
gerashegalov, line 286: I think the code would read easier if we renamed sortedOffset to sortedRows or numSortedRows

revans2: I personally think of it in terms of offsets instead of number of rows. The fact that they end up being equal is just because the sorted values are at the first part of the batch.

pxLi:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
abellina: Got it, I missed the use right below of `sortSplitOffset`.

abellina (approved):

jlowe (approved):

revans2:
@gerashegalov is it OK if I merge this? or do you want me to make changes because the memory model is not OK

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
gerashegalov: Not a blocking issue. 

Just to explain my thingking: I think of an offset as the start position of an array, or a range the way it's used on [L300]( https://github.com/NVIDIA/spark-rapids/pull/2209/files#diff-da27912783f99ffd588b32b729fec7de13fb138238df012e5e2afb87e4b483daR300)  `[sortedOffset, rows)`. Since it describes the unsorted range,  if we wanted to use the term offset, I'd call it unsortedOffset.  On the other hand, we can view it as the definition of the sorted area `[0, sortedOffset)` in which case `sortedRows` works better for me.

gerashegalov (approved):
it's more of my mental model. no problem either way.

PR #2226, opened 2021-04-21
Add cosine similarity native UDF example

jlowe:
Adding a Spark Java UDF that computes the cosine similarity between two float vectors.  The implementation is custom CUDA code and shows how a RAPIDS accelerator UDF can handle nested type inputs.

This also adds a native performance benchmark for cosine similarity along with the cmake framework for building benchmarks.  Benchmarks are not built by default but can be requested via `-DBUILD_UDF_BENCHMARKS=ON`.

As with other native UDFs, this does not build unless the `udf-native-examples` build profile is specified.

### DIFF
--- a/integration_tests/src/main/python/data_gen.py
+++ b/integration_tests/src/main/python/data_gen.py
@@ -313,7 +313,10 @@ class FloatGen(DataGen):
 
     def _fixup_nans(self, v):
         if self._no_nans and (math.isnan(v) or v == math.inf or v == -math.inf):
-            v = None
+            if self.nullable:
+                v = None
+            else:
+                v = 0.0
         return v
 
     def start(self, rand):
@@ -372,7 +375,10 @@ class DoubleGen(DataGen):
 
     def _fixup_nans(self, v):
         if self._no_nans and (math.isnan(v) or v == math.inf or v == -math.inf):
-            v = None
+            if self.nullable:
+                v = None
+            else:
+                v = 0.0
         return v
 
     def start(self, rand):
--- a/integration_tests/src/main/python/data_gen.py
+++ b/integration_tests/src/main/python/data_gen.py
@@ -313,10 +313,7 @@ class FloatGen(DataGen):
 
     def _fixup_nans(self, v):
         if self._no_nans and (math.isnan(v) or v == math.inf or v == -math.inf):
-            if self.nullable:
-                v = None
-            else:
-                v = 0.0
+            v = None if self.nullable else 0.0
         return v
 
     def start(self, rand):
@@ -375,10 +372,7 @@ class DoubleGen(DataGen):
 
     def _fixup_nans(self, v):
         if self._no_nans and (math.isnan(v) or v == math.inf or v == -math.inf):
-            if self.nullable:
-                v = None
-            else:
-                v = 0.0
+            v = None if self.nullable else 0.0
         return v
 
     def start(self, rand):
--- a/integration_tests/src/main/python/rapids_udf_test.py
+++ b/integration_tests/src/main/python/rapids_udf_test.py
@@ -62,6 +62,7 @@ def test_hive_generic_udf():
 @rapids_udf_example_native
 def test_hive_simple_udf_native(enable_rapids_udf_example_native):
     with_spark_session(skip_if_no_hive)
+    data_gens = [["s", StringGen('.{0,30}')]]
     def evalfn(spark):
         load_hive_udf_or_skip_test(spark, "wordcount", "com.nvidia.spark.rapids.udf.hive.StringWordCount")
         return gen_df(spark, data_gens)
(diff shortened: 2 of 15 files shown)


jlowe:
build

udf-examples/src/main/java/com/nvidia/spark/rapids/udf/java/NativeUDFExamplesLoader.java
```diff
@@ -0,0 +1,42 @@
...
+
+package com.nvidia.spark.rapids.udf.java;
+
+import ai.rapids.cudf.NativeDepsLoader;
+
+import java.io.IOException;
+
+/** Loads the native dependencies for UDF examples with a native implementation */
+public class NativeUDFExamplesLoader {
+  private static volatile boolean isLoaded = false;
```
gerashegalov, line 25: nit: false is already Java's initialization default for booleans

udf-examples/src/main/java/com/nvidia/spark/rapids/udf/java/NativeUDFExamplesLoader.java
```diff
@@ -0,0 +1,42 @@
...
+
+  /** Loads native UDF code if necessary */
+  public static void ensureLoaded() {
+    if (!isLoaded) {
```
gerashegalov, line 29: nothing prevents 2 concurrent calls `ensureLoaded` to enter `load()`. Let us not mix volatile and synchronized, and just make `ensureLoaded` synchronized too.

jlowe: Nice catch!  I waffled between two designs on this, and ended in a bad mix of the two. 😞  The intent was to do double-checked locking to avoid the synchronization overhead, as there's potentially many threads trying to do this simultaneously.  But it was missing the second check, and single-checked locking doesn't work. :smile:

However I agree, let's keep it simple for now, we can always optimize if profiles show this is a problem.

integration_tests/src/main/python/data_gen.py
```diff
@@ -313,7 +313,10 @@ def __init__(self, nullable=True,
 
     def _fixup_nans(self, v):
         if self._no_nans and (math.isnan(v) or v == math.inf or v == -math.inf):
-            v = None
+            if self.nullable:
```
abellina, line 316: suggest: `v = None if self.nullable else 0.0`

integration_tests/src/main/python/rapids_udf_test.py
```diff
@@ -88,3 +89,24 @@ def evalfn(spark):
...
+    def evalfn(spark):
+        class RangeFloatGen(FloatGen):
+            def start(self, rand):
+                def gen_float(): return rand.uniform(-1000.0, 1000.0)
+                self._start(rand, gen_float)
+        load_java_udf_or_skip_test(spark, "cosine_similarity", "com.nvidia.spark.rapids.udf.java.CosineSimilarity")
+        arraygen = ArrayGen(RangeFloatGen(nullable=False), min_length=8, max_length=8)
+        df = binary_op_df(spark, arraygen, length=16)
+        return df.selectExpr("cosine_similarity(a, b)")
+    assert_gpu_and_cpu_are_equal_collect(evalfn)
```
abellina, line 104: Ok so for the CPU test, even though we've registered the `cosine_similarity` udf, the CPU implementation of `cosine_similarity` is picked up in the plan? or is it being skipped, just trying to understand this part.

abellina (commented):
small nit

integration_tests/src/main/python/rapids_udf_test.py
abellina: Thanks! Missed that.

udf-examples/src/main/cpp/benchmarks/synchronization/synchronization.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include <rmm/cuda_stream_view.hpp>
+#include <rmm/device_buffer.hpp>
+
+cuda_event_timer::cuda_event_timer(benchmark::State& state,
+                                   bool flush_l2_cache,
+                                   rmm::cuda_stream_view stream)
+  : stream(stream), p_state(&state)
+{
+  // flush all of L2$
+  if (flush_l2_cache) {
```
abellina, line 30: Is this done to help the timer code be more consistent? My guess yes, but it wasn't entirely clear. Maybe expanding the comment?

abellina:
@jlowe looks really good. Mostly curious on the L2 thing.

udf-examples/src/main/cpp/benchmarks/synchronization/synchronization.cpp
jlowe: It's to make the benchmark more consistent by reducing the impact of caching from run-to-run.  Standard stuff.  This file and the benchmark fixture file were lifted from cudf essentially verbatim because I couldn't easily access them during the build otherwise.

abellina (approved):

gerashegalov (approved):

revans2 (approved):

PR #2235, opened 2021-04-22
Fix distributed cache to read requested schema 

razajafri:
This PR fixes a bug which caused a failure when reading the cache if the column order was different from the stored cache. There is a point where we read the schema directly from the stored cache and use it to initialize the `VectorizedColumnReader`s but the `WritableColumnVectors` are still initialized using the originalRequestSchema which caused a failure because of a mismatch between the type of `WriteableColumnVector` and `VectorizedColumnReader`. 

A defensive piece of code is also added that protects against a case where the requested schema has a column that doesn't exist in the cache. 

Another change that this PR makes is that the Hadoop Conf will not be broadcasted, instead it will be created on the executor side to save the driver from broadcasting all the properties inside the object. 

Unit test was added to test

### DIFF
--- a/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
+++ b/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
@@ -831,7 +831,8 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
           this.totalRowCount += block.getRowCount
         }
 
-        // initialize missingColumns
+        // initialize missingColumns to cover the case where requested column isn't present in the
+        // cache, which should never happen but just in case it does
         val columns = reqParquetSchema.getColumns
         val paths = reqParquetSchema.getPaths
         val fileSchema = parquetFileReader.getFooter.getFileMetaData.getSchema
--- a/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
+++ b/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
@@ -55,7 +55,7 @@ import org.apache.spark.sql.execution.datasources.parquet.rapids.shims.spark311.
 import org.apache.spark.sql.execution.vectorized.{OffHeapColumnVector, WritableColumnVector}
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.internal.SQLConf.LegacyBehaviorPolicy
-import org.apache.spark.sql.types.{StructType, _}
+import org.apache.spark.sql.types._
 import org.apache.spark.sql.vectorized.ColumnarBatch
 import org.apache.spark.storage.StorageLevel
 import org.apache.spark.unsafe.types.CalendarInterval
@@ -849,13 +849,12 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
               throw new UnsupportedOperationException("Schema evolution not supported.")
             }
             missingColumns(i) = false
-          }
-          else {
+          } else {
             if (columns.get(i).getMaxDefinitionLevel == 0) {
               // Column is missing in data but the required data is non-nullable.
               // This file is invalid.
-              throw new IOException("Required column is missing in data file. Col: "
-                + colPath.toList)
+              throw new IOException(s"Required column is missing in data file. Col: " +
+                s"${colPath.toList}")
             }
             missingColumns(i) = true
           }
--- a/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
+++ b/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
@@ -853,8 +853,7 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
             if (columns.get(i).getMaxDefinitionLevel == 0) {
               // Column is missing in data but the required data is non-nullable.
               // This file is invalid.
-              throw new IOException(s"Required column is missing in data file. Col: " +
-                s"${colPath.toList}")
+              throw new IOException(s"Required column is missing in data file: ${colPath.toList}")
             }
             missingColumns(i) = true
           }
--- a/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
+++ b/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
@@ -43,7 +43,7 @@ import org.apache.parquet.io.{DelegatingPositionOutputStream, DelegatingSeekable
 import org.apache.spark.TaskContext
 import org.apache.spark.broadcast.Broadcast
 import org.apache.spark.rdd.RDD
-import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.{vectorized, SparkSession}
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, Expression, GenericInternalRow, SpecializedGetters, UnsafeRow}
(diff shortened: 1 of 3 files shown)


shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -834,6 +831,36 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
+          }
+          val colPath = paths.get(i)
+          if (fileSchema.containsPath(colPath)) {
+            val fd = fileSchema.getColumnDescription(colPath)
+            if (!(fd == columns.get(i))) {
+              throw new UnsupportedOperationException("Schema evolution not supported.")
+            }
+            missingColumns(i) = false
+          }
+          else {
```
jlowe, line 853:
```suggestion
          } else {
```

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -834,6 +831,36 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
+            if (columns.get(i).getMaxDefinitionLevel == 0) {
+              // Column is missing in data but the required data is non-nullable.
+              // This file is invalid.
+              throw new IOException("Required column is missing in data file. Col: "
+                + colPath.toList)
```
jlowe, line 858:
```suggestion
              throw new IOException(s"Required column is missing in data file: ${colPath.toList}")
```

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -485,12 +483,11 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
       })
       cbRdd.mapPartitions(iter => CloseableColumnBatchIterator(iter))
     } else {
-      val broadcastedHadoopConf = getBroadcastedHadoopConf(conf, selectedAttributes)
```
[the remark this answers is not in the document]
razajafri: I will update the description. I wanted to minimize the broadcasted variables and since we are creating a new SQLConf on the executor side, it made sense to create the hadoopConf on the executor as well.

razajafri:
@jlowe I think I have addressed your concerns PTAL

razajafri:
build

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -849,13 +849,12 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
-          }
-          else {
+          } else {
             if (columns.get(i).getMaxDefinitionLevel == 0) {
               // Column is missing in data but the required data is non-nullable.
               // This file is invalid.
-              throw new IOException("Required column is missing in data file. Col: "
-                + colPath.toList)
+              throw new IOException(s"Required column is missing in data file. Col: " +
+                s"${colPath.toList}")
```
jlowe, line 857: There's no point in formatting a string that doesn't have format directives, and the `Col: ` is redundant since the message already mentions a column is missing.
```suggestion
              throw new IOException(s"Required column is missing in data file: ${colPath.toList}")
```

jlowe:
> Another change that this PR makes is that the Hadoop Conf will not be broadcasted, instead it will be created on the executor side to save the driver from broadcasting all the properties inside the object.

I don't think we necessarily want this behavior, right?  Spark explicitly broadcasts the Hadoop conf elsewhere, and it solves problems where an executor runs on a node that has a Hadoop conf setup that doesn't match what the driver has.  By broadcasting the driver's view of the conf, it's a consistent conf across the Spark cluster.

razajafri:
> > Another change that this PR makes is that the Hadoop Conf will not be broadcasted, instead it will be created on the executor side to save the driver from broadcasting all the properties inside the object.
> 
> I don't think we necessarily want this behavior, right? Spark explicitly broadcasts the Hadoop conf elsewhere, and it solves problems where an executor runs on a node that has a Hadoop conf setup that doesn't match what the driver has. By broadcasting the driver's view of the conf, it's a consistent conf across the Spark cluster.

I thought it was more efficient to create the HadoopConf on the driver and broadcast it once, but @gerashegalov is of the opinion that it's more efficient to create it on the executor side. @gerashegalov can you please chime in on why you thought that was the case?

jlowe:
> @gerashegalov is of the opinion that it's more efficient to create it on the executor side.

Ah, I remember now.  It's different in this case because we just need _a_ Hadoop conf, not _the_ Hadoop conf.  The code is explicitly creating a conf with no defaults and setting up the config for Parquet.  I think that should be OK as long as we've covered all the settings that affect how we use Parquet.

jlowe:
Looks like the latest code isn't building? 
```
17:05:11  [INFO] Compiling 2 Scala sources to /home/jenkins/agent/workspace/jenkins-rapids_premerge-github-1436/tests-spark310+/target/test-classes ...
17:05:16  [ERROR] [Error] /home/jenkins/agent/workspace/jenkins-rapids_premerge-github-1436/tests-spark310+/src/test/scala/com/nvidia/spark/rapids/Spark310ParquetWriterSuite.scala:176: too many arguments (4) for constructor CachedBatchIteratorProducer: (iter: Iterator[org.apache.spark.sql.vectorized.ColumnarBatch], cachedAttributes: Seq[org.apache.spark.sql.catalyst.expressions.Attribute], sharedConf: org.apache.spark.broadcast.Broadcast[Map[String,String]])ser.CachedBatchIteratorProducer[org.apache.spark.sql.vectorized.ColumnarBatch]
17:05:16  [ERROR] one error found
```

razajafri:
> > @gerashegalov is of the opinion that it's more efficient to create it on the executor side.
> 
> Ah, I remember now. It's different in this case because we just need _a_ Hadoop conf, not _the_ Hadoop conf. The code is explicitly creating a conf with no defaults and setting up the config for Parquet. I think that should be OK as long as we've covered all the settings that affect how we use Parquet.

I would still think your original comment about Spark broadcast being more efficient makes more sense and we should do it that way. The reason why we shouldn't do it is because, as I understand, the broadcast variable is read-only, while over here we are modifying the hadoopConf by resetting the Parquet schema calculated from the batch. 



> Looks like the latest code isn't building?
> 
> ```
> 17:05:11  [INFO] Compiling 2 Scala sources to /home/jenkins/agent/workspace/jenkins-rapids_premerge-github-1436/tests-spark310+/target/test-classes ...
> 17:05:16  [ERROR] [Error] /home/jenkins/agent/workspace/jenkins-rapids_premerge-github-1436/tests-spark310+/src/test/scala/com/nvidia/spark/rapids/Spark310ParquetWriterSuite.scala:176: too many arguments (4) for constructor CachedBatchIteratorProducer: (iter: Iterator[org.apache.spark.sql.vectorized.ColumnarBatch], cachedAttributes: Seq[org.apache.spark.sql.catalyst.expressions.Attribute], sharedConf: org.apache.spark.broadcast.Broadcast[Map[String,String]])ser.CachedBatchIteratorProducer[org.apache.spark.sql.vectorized.ColumnarBatch]
> 17:05:16  [ERROR] one error found
> ```

Pushed an update

razajafri:
build

jlowe (dismissed):
One of my comments about the exception string wasn't addressed, but it's minor.

razajafri:
@jlowe I have addressed your nit as well

razajafri:
build

gerashegalov:
> gerashegalov is of the opinion that it's more efficient to create it on the executor side.

Hadoop Conf objects are not exactly lightweight as a map. Even without defaults, each object is tracked  https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java#L830

I probably don't know all the context here but my general thinking was that instead of 
1. Constructing the conf on the driver  
2. Serializing and broadacsting the conf from the driver to executors
3. Deserializing to re-construct the conf on the Executors 

we could use a more lightweight broadcast variable to pass the info. Obviously we can't avoid patching the conf on the executors if it's not known apriori.

PR #2239, opened 2021-04-23
Do not cache the batches in Map and Group Map Panda UDF nodes.

### LINKED ISSUE #2238: [BUG] Unnecessary to cache the batches that will be sent to Python in `FlatMapGroupInPandas`.
Actually it should be an improvment more than a bug. 
The code snip is as below,
```scala
        .map { groupBatch =>
          // Cache the input batches for release after writing done.
          queue.add(groupBatch, spillCallback)
          groupBatch
        }
```
The Python runner will close the batches after writing them to Python, so no need to cache them in the queue for release.
Need to clean this.

firestarman:
The Python runner will close the batches after writing them to Python, so no need to cache them in the queue for the later release.
    
Then the batch queue is no longer needed, removed it too.

Closes #2238


firestarman:
build

firestarman:
@revans2 Could you take a look ? It is a simple PR.

jlowe (approved):

revans2 (approved):

PR #2243, opened 2021-04-23
Remove ShuffleExchangeExec from the automatic allow list in tests

### LINKED ISSUE #2146: [FEA] Improve handling of shuffles during integration tests
While working on #2145 I noticed that the existing repartition test was not actually testing anything on the GPU.  The test was supposed to be testing that a partition and shuffle matched the CPU, but the plan optimized out the GPU shuffle since the other parts of the plan were not columnar.  The test should have failed since the plan was not columnar in any way, but it went ahead and compared CPU to CPU which isn't helpful.

There should be a mechanism to avoid ignoring shuffles that are not on the GPU or force shuffles to be translated to the GPU during planning.

revans2:
This fixes #2146


revans2:
build

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -174,12 +174,14 @@ def get_params(init_list, marked_params=[]):
 
 # Pytest marker for list of operators allowed to run on the CPU,
 # esp. useful in partial and final only modes.
+# but this ends up allowing close to everything beig off the GPU so I am not sure how
```
jlowe, line 177:
```suggestion
# but this ends up allowing close to everything being off the GPU so I am not sure how
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -430,15 +430,14 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
...
       case _: RDDScanExec => () // Ignored
-      case _: ShuffleExchangeExec if conf.cpuRangePartitioningPermitted => {
-        // Ignored for now, we don't force it to the GPU if
-        // children are not on the gpu
-      }
       case other =>
         if (!plan.supportsColumnar &&
+            // There are some python execs that are not columnar because of a little
+            // used feature. This prevents those from failing tests. It also allows
+            // the columnar to row transitions to need to be explicitly allowed either.
```
jlowe, line 439: "also allows..to need to be explicitly allowed either" is difficult to parse what is intended by the comment.

revans2:
build

tests/src/test/scala/com/nvidia/spark/rapids/AdaptiveQueryExecSuite.scala
```diff
@@ -294,11 +296,12 @@ class AdaptiveQueryExecSuite
   }
 
   test("Avoid transitions to row when writing to Parquet") {
-
+    logError("Avoid transitions to row when writing to Parquet")
```
jlowe, line 299: Are these logs leftover from a debug session?  Normally we don't log errors in the tests.

revans2: Yes, but we have no good way to tell what logs came from which test. Normally we have not been doing it, but debugging without something like it is really painful.

jlowe (approved):

revans2:
build

revans2:
@jlowe I resolved the merge conflicts and this should be good to go now.

jlowe (approved):

PR #2249, opened 2021-04-23
Update UDF native example libcudf dependency to 0.20

jlowe:
#2226 originally targeted branch-0.5 which is based on cudf 0.19, but when it was retargeted to branch-0.6 the cudf dependency wasn't updated to 0.20.  This doesn't break the build given the simple libcudf interfaces the UDF native examples require, as they haven't changed between libcudf 0.19 and 0.20.  However it's best to be consistent with the libcudf dependency.


jlowe:
build

revans2 (approved):

Issue #2060, opened 2021-04-01
[FEA] Support element_at function on GPU


Dooyoung-Hwang:
I request a feature of element_at function on GPU.

The API link : https://spark.apache.org/docs/latest/api/sql/index.html#element_at

```
> SELECT element_at(array(1, 2, 3), 2);
 2
> SELECT element_at(map(1, 'a', 2, 'b'), 2);
 b
```

Driver log :
I!NOT_FOUND <ElementAt> element_at cannot run on GPU because no GPU enabled version of expression class org.apache.spark.sql.catalyst.expressions.ElementAt could be found

revans2:
We really need some more information to understand how you want to use this.  Spark is very generic in what it supports compared to CUDF.  So we can support some of this right now with just a little bit of work.  But full support is going to take a lot more, so it would be good to understand exactly what the requirements are.

Do you want to use this for arrays or maps or both? 
Is the key/index always a constant value or can it be a column of data?
What are they types that you need to support for the array (just the values) or map (both the key and the value)?

Dooyoung-Hwang:
We just want to get an element of ArrayType of atomic DataType with index. We found a workaround for this case and applied and ran the job with GPU.

[Before]
```python
    # prediction : Array of float
    features = ["c1", "c2", "c3"]
    for i in range(len(features)):
        inf_df = inf_df.withColumn(features[i], element_at("prediction", i + 1))
```

[After]
```python
    # prediction : Array of float
    features = ["c1", "c2", "c3"]
    for i in range(len(features)):
        inf_df = inf_df.withColumn(features[i], col("prediction")[i])
```

sameerz:
Tying `element_at` to what we already support for element indexing is the reason for the good first issue label.  Supporting additional features, like columnar indexing (as opposed to indexing with a literal) will be more challenging.

wjxiz1992:
> We really need some more information to understand how you want to use this. Spark is very generic in what it supports compared to CUDF. So we can support some of this right now with just a little bit of work. But full support is going to take a lot more, so it would be good to understand exactly what the requirements are.
> 
> Do you want to use this for arrays or maps or both?
> Is the key/index always a constant value or can it be a column of data?
> What are they types that you need to support for the array (just the values) or map (both the key and the value)?

Hi @revans2 , since SKT's requirement is the simple case (not nested), can I take this feature issue?

revans2:
> can I take this feature issue? To support the simple case first.

Sure

[...]

PR #2260, opened 2021-04-25
Support ElementAt

wjxiz1992:
This PR is to close https://github.com/NVIDIA/spark-rapids/issues/2060. This gives initial support for `element_at`.

Note https://github.com/apache/spark/pull/30297 introduce different behaviors for element_at when invalid index or key is given to lhs with ansi mode enabled and disabled. 
A [follow-up issue](https://github.com/NVIDIA/spark-rapids/issues/2272) is created to track the difference.

For `select element_at(Array(1,2,3), 4)`:

Spark 3.0.2 + spark.sql.ansi.enabled=true:
```
+-----------------------------+
|element_at(array(1, 2, 3), 4)|
+-----------------------------+
|                         null|
+-----------------------------+
```

Spark 3.1.1 + spark.sql.ansi.enabled=true:
```
java.lang.ArrayIndexOutOfBoundsException: Invalid index: 4, numElements: 3
```

For `select element_at(map(1, 'a', 2, 'b'), 3)`:

Spark 3.0.2 + spark.sql.ansi.enabled=true:
```
+------------------------------+
|element_at(map(1, a, 2, b), 3)|
+------------------------------+
|                          null|
+------------------------------+
```
Spark 3.1.1 + spark.sql.ansi.enabled=true:
```
java.util.NoSuchElementException: Key 3 does not exist.
```

Signed-off-by: Allen Xu <wjxiz1992@gmail.com>


integration_tests/src/main/python/array_test.py
```diff
@@ -109,3 +109,12 @@ def main_df(spark):
...
     assert_gpu_and_cpu_are_equal_collect(main_df)
+
+@pytest.mark.parametrize('data_gen', [byte_gen, short_gen, int_gen, long_gen,
+                                      FloatGen(no_nans=True), DoubleGen(no_nans=True),
+                                      string_gen, boolean_gen, date_gen, timestamp_gen], ids=idfn)
+def test_array_element_at(data_gen):
+    arr_gen = ArrayGen(data_gen)
+    assert_gpu_and_cpu_are_equal_collect(lambda spark: unary_op_df(
+        spark, arr_gen).select(element_at(col('a'), 1),
+                               element_at(col('a'), -1)), no_nans_conf)
```
sperlingxx, line 120: bottom line missed

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2244,6 +2244,15 @@ object GpuOverrides {
         ("map", TypeSig.MAP.nested(TypeSig.STRING), TypeSig.MAP.nested(TypeSig.all)),
         ("key", TypeSig.lit(TypeEnum.STRING), TypeSig.all)),
       (in, conf, p, r) => new GpuGetMapValueMeta(in, conf, p, r)),
+    expr[ElementAt](
+      "Returns element of array at given(1-based) index in value if column is array. " +
+        "Returns value for the given key in value if column is map.",
+      ExprChecks.binaryProjectNotLambda(
+        TypeSig.commonCudfTypes, TypeSig.all,
```
sperlingxx, line 2251: Shall we also support `TypeSig.NULL`, `TypeSig.DECIMAL`, Array of Struct and nested Array, just like `GpuGetArrayItem`?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -186,6 +186,99 @@ case class GpuGetMapValue(child: Expression, key: Expression)
...
+  override def doColumnar(lhs: Scalar, rhs: GpuColumnVector): ColumnVector =
+    throw new IllegalStateException("This is not supported yet")
+
+  override def doColumnar(lhs: GpuColumnVector, rhs: Scalar): ColumnVector = {
+    lhs.dataType match {
+      case _: ArrayType => {
+        if (rhs.isValid) {
+          if (rhs.getInt > 0) {
+            // SQL 1-based index
+            lhs.getBase.extractListElement(rhs.getInt - 1)
```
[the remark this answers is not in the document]
wjxiz1992: Yes, that's also my concern. According to my test for an existing "table" like:
```
+------------------------+
|col_1                   |
+------------------------+
|[Java, Scala, C++]      |
|[Java, Scala, C++, a, b]|
+------------------------+
```

when I do 
```
spark.sql("select element_at(col_1, 4) from table;").show()
```

current GPU implementation will return 
```
+--------------------+
|element_at(col_1, 4)|
+--------------------+
|                null|
|                   a|
+--------------------+

```

Which is the same to what CPU returns.
I think somehow cuDF side has already dealt with this outOfBound case. Not sure if we should still handle it in plugin side.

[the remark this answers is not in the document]
firestarman: Personally just keep the same behavior with CPU version, no matter it is done in cudf or plugin ? 
What will the CPU version produce if the index is out of bound ?

docs/supported_ops.md
```diff
@@ -5169,6 +5169,138 @@ Accelerator support is described below.
 <td> </td>
 </tr>
 <tr>
+<td rowSpan="6">ElementAt</td>
```
firestarman, line 5172: Seems too much change in this doc for just adding a new expression, could you take a double check ?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -186,6 +186,99 @@ case class GpuGetMapValue(child: Expression, key: Expression)
...
+    conf: RapidsConf,
+    parent: Option[RapidsMeta[_, _, _]],
+    rule: DataFromReplacementRule)
+  extends BinaryExprMeta[ElementAt](expr, conf, parent, rule) {
+  override def convertToGpu(lhs: Expression, rhs: Expression): GpuExpression = {
+    GpuElementAt(lhs, rhs)
+  }
+}
+
+case class GpuElementAt(left: Expression, right: Expression)
```
firestarman, line 200: Seems the CPU has the third argument `failOnError`, but is skipped here. Is it OK ?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -186,6 +186,99 @@ case class GpuGetMapValue(child: Expression, key: Expression)
...
+      case (e1, _) if (!e1.isInstanceOf[MapType] && !e1.isInstanceOf[ArrayType]) =>
+        TypeCheckResult.TypeCheckFailure(s"The first argument to function $prettyName should " +
+          s"have been ${ArrayType.simpleString} or ${MapType.simpleString} type, but its " +
+          s"${left.dataType.catalogString} type.")
+      case _ => TypeCheckResult.TypeCheckSuccess
+    }
+  }
+
+  // Eventually we need something more full featured like
+  // GetArrayItemUtil.computeNullabilityFromArray
```
firestarman, line 240: We'd better do it instead of the hard code here.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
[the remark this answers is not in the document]
wjxiz1992: > Personally I prefer to just keep the same behavior with CPU version, no matter it is done in cudf or plugin.
> What will the CPU version produce if the index is out of bound ?

it will return "null". the same as GPU returns.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -186,6 +186,99 @@ case class GpuGetMapValue(child: Expression, key: Expression)
...
+          if (rhs.getInt > 0) {
+            // SQL 1-based index
+            lhs.getBase.extractListElement(rhs.getInt - 1)
+          } else if (rhs.getInt == 0) {
+            throw new ArrayIndexOutOfBoundsException("SQL array indices start at 1")
+          } else {
+            lhs.getBase.extractListElement(rhs.getInt)
+          }
+        } else {
+          withResource(Scalar.fromNull(
```
firestarman, line 262: Is it possible for CPU version to specify an invalid value for the ordinal (aka index) ?
If not, the index scalar is supposed to be valid here. So I would suggest raising an `illegalArgumentException` instead of null.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -186,6 +186,99 @@ case class GpuGetMapValue(child: Expression, key: Expression)
...
+          }
+        }
+      }
+      case _: MapType => {
+        lhs.getBase.getMapValue(rhs)
+      }
+    }
+  }
+
+  override def doColumnar(numRows: Int, lhs: Scalar, rhs: Scalar): ColumnVector =
```
firestarman, line 274: I am not sure, but what is the user case for this one ?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -186,6 +186,99 @@ case class GpuGetMapValue(child: Expression, key: Expression)
...
+    withResource(GpuColumnVector.from(lhs, numRows, left.dataType)) { expandedLhs =>
```
firestarman, line 275:
```suggestion
    withResource(GpuColumnVector.from(lhs, numRows, lhs.dataType)) { expandedLhs =>
```
?

docs/supported_ops.md
revans2: This is correct we added in a feature to put header rows periodically in the table, and this is one of the side effects.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2244,6 +2244,15 @@ object GpuOverrides {
...
+        ("left", TypeSig.ARRAY.nested(TypeSig.commonCudfTypes) +
```
revans2, line 2252: Type checking is very simple with how it handles nested types. They all end up being combined and this would allow not just arrays of any of the common cudf types, but also maps.

If we want to support all of them we are going to need extra checks that if it is a map then it is only for the types that we support, and we might want to put some notes on them too.

revans2, line 2252: I am not sure that left/right are the right names to use for the parameters. I think it would be clearer to use something like `array/map` and `index/key` for it, which is what uses for their own documentation.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
revans2: This can happen in a few cases with optimizations. It is not common, but we need to cover it.  However this will not work until we can get scalar lists and scalar maps to work.  That is OK for now because it is very rare.

revans2 (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2244,6 +2244,15 @@ object GpuOverrides {
...
+          TypeSig.MAP.nested(TypeSig.STRING), TypeSig.all),
```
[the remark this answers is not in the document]
wjxiz1992: I tested this case, and got an error like 
```
scala> spark.sql("select element_at(2, 2)").collect
org.apache.spark.sql.AnalysisException: cannot resolve 'element_at(2, 2)' due to data type mismatch: The first argument to function element_at should have been array or map type, but its int type.; line 1 pos 7;
'Project [unresolvedalias(element_at(2, 2), None)]
+- OneRowRelation

  at org.apache.spark.sql.catalyst.analysis.package$AnalysisErrorAt.failAnalysis(package.scala:42)
  at org.apache.spark.sql.catalyst.analysis.CheckAnalysis$$anonfun$$nestedInanonfun$checkAnalysis$1$2.applyOrElse(CheckAnalysis.scala:149)
  at org.apache.spark.sql.catalyst.analysis.CheckAnalysis$$anonfun$$nestedInanonfun$checkAnalysis$1$2.applyOrElse(CheckAnalysis.scala:140)
  at org.apache.spark.sql.catalyst.trees.TreeNode.$anonfun$transformUp$2(TreeNode.scala:333)

```

Generally, I am following existing `GetArrayItem` and `GetMapValue` for this expression.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: I filed another issue https://github.com/NVIDIA/spark-rapids/issues/2272 for it's different behaviours.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: For 3.0.0 based version(before 3.1.1), I think we can put nullable = true here, because cuDF is acting the same as CPU.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: Yes, this problem is mainly related to the issue I mention above. for 3.0.0 version, it will return `null` for invalid index/key

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: it needs to know the `left` dataType, the Scalar doesn't provide that.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
wjxiz1992: Agree, it's more clear, updated.

wjxiz1992: Added extra type check in `tagExprForGpu`, also removed the class for meta, moved it under the type check part directly to make it more intuitive.

integration_tests/src/main/python/array_test.py
wjxiz1992: Added.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
wjxiz1992: Added. thanks!

integration_tests/src/main/python/array_test.py
```diff
@@ -109,3 +109,12 @@ def main_df(spark):
         chk_val = df.select(col('a')[0].alias('t')).filter(~isnan(col('t'))).collect()[0][0]
         return df.select(array_contains(col('a'), chk_val))
     assert_gpu_and_cpu_are_equal_collect(main_df)
+
+@pytest.mark.parametrize('data_gen', [byte_gen, short_gen, int_gen, long_gen,
+                                      FloatGen(no_nans=True), DoubleGen(no_nans=True),
```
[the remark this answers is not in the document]
wjxiz1992: Sorry, copied from other tests directly. I've updated the parameters here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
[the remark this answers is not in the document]
firestarman: Better to update the type sig for Spark part.
If it is 'all', there will be many unsupported types against Spark in the `supported_ops ` doc. But actually it is not, since Spark does support only map and array.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
firestarman: OK, this should be tracked by the issue 2272 too.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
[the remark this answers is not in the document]
wjxiz1992: good suggestion! Updated.

firestarman:
build

firestarman (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2244,6 +2244,40 @@ object GpuOverrides {
...
+      "Returns element of array at given(1-based) index in value if column is array. " +
+        "Returns value for the given key in value if column is map.",
+      ExprChecks.binaryProjectNotLambda(
+        (TypeSig.commonCudfTypes + TypeSig.ARRAY + TypeSig.STRUCT + TypeSig.NULL +
+          TypeSig.DECIMAL + TypeSig.MAP).nested(), TypeSig.all,
+        ("array/map", TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.ARRAY +
+          TypeSig.STRUCT + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.MAP) +
+          TypeSig.MAP.nested(TypeSig.STRING)
+            .withPsNote(TypeEnum.MAP ,"If it's map, only string is supported. " +
+              "Extra check is inside the expression metadata"),
```
revans2, line 2257: This is user facing. The "Extra check is inside the expression metadata" means nothing to an end user and should just be a regular comment.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2244,6 +2244,40 @@ object GpuOverrides {
...
+          TypeSig.ARRAY.nested(TypeSig.all) + TypeSig.MAP.nested(TypeSig.all)),
+        ("index/key", (TypeSig.lit(TypeEnum.INT) + TypeSig.lit(TypeEnum.STRING))
+          .withPsNote(TypeEnum.INT, "If it's the index for array, only INT is supported." +
```
revans2, line 2260: We should have a separate withPsNote for TypeEnum.STRING. Perhaps something more like

```
.withPsNote(TypeEnum.INT, "ints are only supported as array indexes, not as maps keys")
.withPsNote(TypeEnum.STRING, "strings are only supported as map keys, not array indexes")
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2244,6 +2244,40 @@ object GpuOverrides {
...
+            "If it's the key for map, only STRING is supported"),
+          TypeSig.all)),
+      (in, conf, p, r) => new BinaryExprMeta[ElementAt](in, conf, p, r) {
+        override def tagExprForGpu(): Unit = {
+          // To distinguish the supported nested type between Array and Map
+          in.left.dataType match {
```
revans2, line 2266: I don't think this is checking all the right things. Could we just try to match exactly what the other operators are doing to be sure?

```scala
val checks = in.left.datatType match {
  case _: MapType =>
     // This should match exactly with the checks for GetMapValue
    ExprChecks.binaryProjectNotLambda(TypeSig.STRING, TypeSig.all,
        ("map", TypeSig.MAP.nested(TypeSig.STRING), TypeSig.MAP.nested(TypeSig.all)),
        ("key", TypeSig.lit(TypeEnum.STRING), TypeSig.all))
  case _: ArrayType =>
    // This should match exactly with the checks for GetArrayItem
    ExprChecks.binaryProjectNotLambda(
        (TypeSig.commonCudfTypes + TypeSig.ARRAY + TypeSig.STRUCT + TypeSig.NULL +
            TypeSig.DECIMAL + TypeSig.MAP).nested(),
        TypeSig.all,
        ("array", TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.ARRAY +
            TypeSig.STRUCT + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.MAP),
            TypeSig.ARRAY.nested(TypeSig.all)),
        ("ordinal", TypeSig.lit(TypeEnum.INT), TypeSig.INT))
  case _ => throw new ...
}
checks.tag(this)
```

sperlingxx (commented):
I found that `ElementAt` is located at `collectionOperations.scala` instead of `complexTypeExtractors.scala` in Spark. Shall we move `GpuElementAt` into `collectionOperations.scala` to follow the structure of Spark SQL ?

revans2:
> I found that `ElementAt` is located at `collectionOperations.scala` instead of `complexTypeExtractors.scala` in Spark. Shall we move `GpuElementAt` into `collectionOperations.scala` to follow the structure of Spark SQL ?

For most expressions I think it is good to try and mirror where Spark has them so yes.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
wjxiz1992: Updated.

wjxiz1992:
Rebased to latest branch-0.6.

wjxiz1992:
build

wjxiz1992:
build

wjxiz1992:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -0,0 +1,133 @@
...
+    lhs.dataType match {
+      case _: ArrayType => {
+        if (rhs.isValid) {
+          if (rhs.getInt > 0) {
+            // SQL 1-based index
+            lhs.getBase.extractListElement(rhs.getInt - 1)
+          } else if (rhs.getInt == 0) {
+            throw new ArrayIndexOutOfBoundsException("SQL array indices start at 1")
+          } else {
+            lhs.getBase.extractListElement(rhs.getInt)
```
revans2, line 112: Does this have similar ANSI issues as https://github.com/NVIDIA/spark-rapids/pull/2350?

Should we refactor the code so that the core part of GpuGetArrayItem and GpuGetMapValue are in static methods so we can define GetElementAt in terms of them?

wjxiz1992: Yes, GetArrayItem and GetMapValue both have ANSI mode problem.
For GetMapValue, I think we can extract the common code for core part.
For GetArrayItem, there's a little difference: element_at requires 1 as the start index, while GetArrayItem uses 0. There might be a parameter like "start_zero: Boolean" for its core function.

I'll take a more close look at it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -210,4 +203,45 @@ case class GpuArrayContains(left: Expression, right: Expression)
...
+        } else if (ordinal.getInt == 0) {
+          throw new ArrayIndexOutOfBoundsException("SQL array indices start at 1")
+        } else {
+          array.getBase.extractListElement(ordinal.getInt)
+        }
+      } else {
+        withResource(Scalar.fromNull(
+          GpuColumnVector.getNonNestedRapidsType(dataType))) { nullScalar =>
+          ColumnVector.fromScalar(nullScalar, array.getRowCount.toInt)
+        }
```
wjxiz1992, line 237: keep the duplicated codes just to make the logic more intuitive.

wjxiz1992:
build

wjxiz1992:
Hi @revans2 , could you help take another look at this? Thanks!

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
wjxiz1992: Hi @revans2 , I refactored the part and move common codes into static methods, could you help take a look ? Thanks!

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -210,4 +203,45 @@ case class GpuArrayContains(left: Expression, right: Expression)
     lhs.getBase.listContainsColumn(rhs.getBase)
 
   override def prettyName: String = "array_contains"
-}
\ No newline at end of file
+}
+
+object GetArrayItemUtil {
+  def evalColumnar(array: GpuColumnVector, ordinal: Scalar, dataType: DataType,
```
firestarman, line 209: Public API needs doc

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -210,4 +203,45 @@ case class GpuArrayContains(left: Expression, right: Expression)
...
+          GpuColumnVector.getNonNestedRapidsType(dataType))) { nullScalar =>
+          ColumnVector.fromScalar(nullScalar, array.getRowCount.toInt)
+        }
+      }
+    }
+  }
+}
+
+object GetMapValueUtil {
+  def evalColumnar(map: GpuColumnVector, key: Scalar): ColumnVector = {
```
firestarman, line 244: Again public API needs doc

firestarman: I am wondering whether we really need this API since it is quite simple .

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -17,6 +17,7 @@
 package org.apache.spark.sql.rapids
 
 import ai.rapids.cudf.{ColumnVector, Scalar}
+import com.nvidia.spark.RebaseHelper.withResource
```
firestarman, line 20: I do not know this `withResource`, but we had better use this the one from `Arm.scala` by extending your class/object from `Arm`.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -0,0 +1,119 @@
+/*
```
wjxiz1992, line 1: This file is moved from rapids to sql, because:
1) to match the path for this file in Spark
2) to be able to call private [sql] class like `AbstractDataType`, `ArrayType.simpleString`

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: Done.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: Yes, currently map only supports StringType as key&value which looks simple. 
We may need more code to deal with other more types in the future.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: Nice catch.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
firestarman: NIT. A standard doc would be better, including params, returns.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
firestarman: nit: A standard doc would be better, including params, returns.

firestarman:
LGTM

firestarman:
build

wjxiz1992:
CI failing due to a window related PR, this PR has been reverted.
Need to wait until the new cuDF jar is ready.

revans2:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -210,4 +203,45 @@ case class GpuArrayContains(left: Expression, right: Expression)
     lhs.getBase.listContainsColumn(rhs.getBase)
 
   override def prettyName: String = "array_contains"
-}
\ No newline at end of file
+}
+
+object GetArrayItemUtil {
+  def evalColumnar(array: GpuColumnVector, ordinal: Scalar, dataType: DataType,
+                   zeroIndexed: Boolean): ColumnVector = {
```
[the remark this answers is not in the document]
revans2: I asked for a common implementation because I thought it could be combined cleanly.  This is not combined cleanly. At a minimum we need some documentation to explain exactly what is happening because it is not obvious from just the API. However, I am inclined to admin that my request was a bad one because the corner cases are handled differently enough that it is not simple to combine them, and perhaps we should leave these separate.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -210,4 +202,49 @@ case class GpuArrayContains(left: Expression, right: Expression)
...
+        if (ordinal.getInt > 0) {
+          // SQL 1-based index
+          array.getBase.extractListElement(ordinal.getInt - 1)
+        } else if (ordinal.getInt == 0) {
+          throw new ArrayIndexOutOfBoundsException("SQL array indices start at 1")
+        } else {
+          array.getBase.extractListElement(ordinal.getInt)
+        }
+      } else {
+        withResource(Scalar.fromNull(
```
revans2, line 235: can we future proof this a little? We already have the data type this is supposed to be, it is just the Spark version of it.

```
withResource(GpuScalar.from(null, dataType))  { nullScalar =>
```

This also goes for above, but if we are going to revert things so they are separate again you only need to do this for new code.

wjxiz1992: Sure, will do it when the GpuScalar PR gets merged.

wjxiz1992:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: Updated.

revans2 (approved):

wjxiz1992:
build

firestarman (dismissed):

wjxiz1992:
build

wjxiz1992:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -0,0 +1,131 @@
...
+            // SQL 1-based index
+            lhs.getBase.extractListElement(ordinalValue - 1)
+          } else if (ordinalValue == 0) {
+            throw new ArrayIndexOutOfBoundsException("SQL array indices start at 1")
+          } else {
+            lhs.getBase.extractListElement(ordinalValue)
+          }
+        } else {
+          GpuColumnVector.columnVectorFromNull(lhs.getRowCount.toInt, dataType)
+          }
```
firestarman, line 117:
```suggestion
        }
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -0,0 +1,131 @@
...
+        }
```
firestarman, line 118:
```suggestion
      }
```

firestarman (approved):

firestarman:
build

PR #2261, opened 2021-04-25
Make CBO row count test more robust

andygrove:

Due to the non-deterministic nature of AQE (query stages run concurrently and can result in different final plans depending on the order of completion), one of the new CBO tests was sometimes failing in CI. This PR makes the test less specific.


andygrove:
build

pxLi (approved):

Issue #1143, opened 2020-11-17
[FEA] Improve performance of Cache plugin


razajafri:
- [ ] Improve the Parquet Serializer to handle large batches that may exceed Int.MAX_SIZE
- [ ] Implementing `buildFilter` method might make the performance better 
- [ ] Look into NVComp to see if we can provide a better performance than Parquet - Might not be needed

razajafri:
This is a follow on to our work on Spark Cache that was done as a part of #444

PR #2271, opened 2021-04-27
Parquet support for Structs

razajafri:
This PR adds support for Structs in Parquet writer. It depends on https://github.com/rapidsai/cudf/pull/7998


razajafri:
build

razajafri:
CI is failing because of the cudf dependency

revans2:
Moved to draft because what it depends on is not merged in so no point in worrying about CI right now.

docs/additional-functionality/cache-serializer.md
```diff
@@ -29,22 +29,28 @@ nav_order: 2
...
   | ------ | -----|
   | 3.1.1 | com.nvidia.spark.rapids.shims.spark311.ParquetCachedBatchSerializer |
   | 3.1.2 | com.nvidia.spark.rapids.shims.spark312.ParquetCachedBatchSerializer |
-  | 3.2.0 | com.nvidia.spark.rapids.shims.spark320.ParquetCachedBatchSerializer |
+  | 3.2.0 | com.nvidia.spark.rapids.shims.spark320.ParquetCachedBatchSerializer | 
+  
+##          Supported Types                       
+ 
+ All types are supported on the CPU, on the Gpu, ArrayType, MapType and BinaryType are not
+ supported
```
revans2, line 55:
```
supported. If an unsupported type is encountered the Rapids Accelerator for Apache Spark will fall back to using the CPU for caching.
```

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -272,8 +272,16 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
 
-  def isSupportedByCudf(schema: Seq[Attribute]): Boolean = {
-    schema.forall(a => GpuColumnVector.isNonNestedSupportedType(a.dataType))
+  def isSchemaSupportedByCudf(schema: Seq[Attribute]): Boolean = {
+    schema.forall(field => isSupportedByCudf(field.dataType))
+  }
+
+  def isSupportedByCudf(dataType: DataType): Boolean = {
+    dataType match {
+      // TODO: have a case for ArrayType
```
revans2, line 281: nit: I assume this is a follow on issue? If so it might be cleaner to say

```
// TODO: when arrays are supported for cudf writes add it here.
```

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -448,7 +454,7 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
   private def getColumnNames(
      selectedAttributes: Seq[Attribute],
      cacheAttributes: Seq[Attribute]): Seq[String] = {
-    selectedAttributes.map(a => "_col" + cacheAttributes.map(_.exprId).indexOf(a.exprId))
+    selectedAttributes.map(a => a.name)
```
revans2, line 457: First off `cacheAttributes` is not used any more so it should be removed from the method signature. Second parquet has limits on the names it supports. It also does not want to have duplicate names, because reading it is ambiguous.  Spark does not have the same limitations. It is fine with duplicate names, because the expression id disambiguates them. Are we sure that this change will never run into limitations on parquet names?

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/Spark311Shims.scala
```diff
@@ -276,12 +276,20 @@ class Spark311Shims extends Spark301Shims {
         }),
       GpuOverrides.exec[InMemoryTableScanExec](
         "Implementation of InMemoryTableScanExec to use GPU accelerated Caching",
-        ExecChecks(TypeSig.commonCudfTypes + TypeSig.DECIMAL, TypeSig.all),
+        ExecChecks((TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.STRUCT).nested()
+          .withPsNote(TypeEnum.DECIMAL,
+            "Negative scales aren't supported at the moment even with " +
+              "spark.sql.legacy.allowNegativeScaleOfDecimal set to true. This is because Parquet " +
```
revans2, line 282: Is there a follow on issue for this? Probably not that important. We could hack around it, but it is probably not worth it to do so.

shims/spark311/src/main/scala/org/apache/spark/sql/rapids/shims/spark311/GpuInMemoryTableScanExec.scala
```diff
@@ -28,85 +28,85 @@ import org.apache.spark.sql.execution.{LeafExecNode, SparkPlan}
...
   @transient relation: InMemoryRelation) extends LeafExecNode with GpuExec {
 
-    override val nodeName: String = {
-      relation.cacheBuilder.tableName match {
-        case Some(_) =>
-          "Scan " + relation.cacheBuilder.cachedName
-        case _ =>
-          super.nodeName
-      }
+  override val nodeName: String = {
```
revans2, line 35: A lot of this is white space changes, which makes it really hard to review. Could you revert the changes to just indentation please?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -97,18 +98,34 @@ object GpuParquetFileFormat {
...
+  (builder: ParquetColumnWriterOptions.NestedBuilder[T, V], schema: StructType): T = {
+    // TODO once https://github.com/rapidsai/cudf/issues/7654 is fixed go back to actually
+    // setting if the output is nullable or not everywhere we have hard-coded nullable=true
+    schema.foreach(field =>
+      field.dataType match {
+        case dt: DecimalType =>
+          builder.withDecimalColumn(field.name, dt.precision, true)
+        case TimestampType =>
+          builder.withTimestampColumn(field.name,
+            // TODO once https://github.com/rapidsai/cudf/issues/8070 is fixed uncomment the line
```
revans2, line 111: We need a P1 blocker issue filed against us for this too.  At a minimum we need to fall back to the CPU for this.

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/Spark311Shims.scala
razajafri: I have added it to this issue https://github.com/NVIDIA/spark-rapids/issues/1143

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
razajafri: I created this when I saw the issue. https://github.com/rapidsai/cudf/issues/8070

razajafri: This is the plugin bug tracking this https://github.com/NVIDIA/spark-rapids/issues/2302

shims/spark311/src/main/scala/org/apache/spark/sql/rapids/shims/spark311/GpuInMemoryTableScanExec.scala
razajafri: This file only has indentation changes. I can revert and create a separate patch for it

revans2: I would prefer a separate patch.

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
razajafri: I didn't know about Spark not having a limitation on col names, thanks. The reason why I made the change was because now we are forcing people to set the column name on the column before writing parquet, but there's no guarantee there won't be duplicates. I will revert.

razajafri:
@revans2 PTAL, I have added code to fallback to CPU if either `int96AsTimestamp` or `outputTimestamp == INT96` is true. Once I get an approval on this, I will merge the cudf change

razajafri:
fixes #2302

razajafri:
I made necessary changes to the tests so we avoid the overflow scenario when writing values to INT96.

revans2 (commented):
We made the test avoid the INT96 overflow issue but we didn't fix the issue or document it in any way. If @sameerz really wants to get this merged in soon and do the INT96 work as a follow on, then I am fine with it. But I would much rather just fix it now, especially if the tests are passing now, even if we don't know 100% why.

integration_tests/src/main/python/parquet_write_test.py
```diff
@@ -29,15 +29,21 @@
...
-
+parquet_decimal_gens=[decimal_gen_default, decimal_gen_scale_precision, decimal_gen_same_scale_precision, decimal_gen_64bit]
+parquet_decimal_struct_gen= StructGen([['child'+str(ind), sub_gen] for ind, sub_gen in enumerate(parquet_decimal_gens)])
 writer_confs={'spark.sql.legacy.parquet.datetimeRebaseModeInWrite': 'CORRECTED',
               'spark.sql.legacy.parquet.int96RebaseModeInWrite': 'CORRECTED'}
 
+
+parquet_basic_gen =[byte_gen, short_gen, int_gen, long_gen, float_gen, double_gen,
+                    string_gen, boolean_gen, date_gen,
+                    # we are limiting TimestampGen to avoid overflowing the INT96 value
```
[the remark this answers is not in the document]
razajafri: Spark doesn't overflow and writes the value as a binary. 

Thanks for pointing this out. I have made the change to throw an exception in case of overflow. PTAL

razajafri:
@revans2 have I addressed your concerns? Please let me know

integration_tests/src/main/python/parquet_write_test.py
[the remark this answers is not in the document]
revans2: @razajafri we need a follow on issue to "fix" this eventually. It might not ever be a high enough priority to actually fix, but we should still have it on the backlog.

@jlowe what do you think? In a query we are never really writing more than one dataset per query so if they hit the error then can turn off parquet writes. Part of me wants a separate config, but another part says that is complexity for no real benefit.

docs/compatibility.md
```diff
@@ -244,6 +244,10 @@ values occurring before the transition between the Julian and Gregorian calendar
 When writing `spark.sql.legacy.parquet.datetimeRebaseModeInWrite` is currently ignored as described
 [here](https://github.com/NVIDIA/spark-rapids/issues/144).
 
+When `spark.sql.parquet.outputTimestampType` is set to `INT96`, the timestamps will overflow and 
+result in an `IllegalArgumentException` thrown, if any value is before 
+September 21, 1677 12:12:43 AM or it is after April 11, 2262 11:47:17 PM.
```
revans2, line 249: Can we document how to work around this? like disable the parquet writer with the config ...

razajafri:
> @razajafri we need a follow on issue to "fix" this eventually. It might not ever be a high enough priority to actually fix, but we should still have it on the backlog.

There is already an open issue tracking this https://github.com/rapidsai/cudf/issues/8070

revans2:
> There is already an open issue tracking this [rapidsai/cudf#8070](https://github.com/rapidsai/cudf/issues/8070)
> 
> Once that is fixed we can remove the test boundaries

Can we have one on the plugin side? because we would need to remove the checks/docs that this is adding in. It would also be good to have traceability in the comments for both the tests and the code as well.

razajafri:
> > There is already an open issue tracking this [rapidsai/cudf#8070](https://github.com/rapidsai/cudf/issues/8070)
> > Once that is fixed we can remove the test boundaries
> 
> Can we have one on the plugin side? because we would need to remove the checks/docs that this is adding in. It would also be good to have traceability in the comments for both the tests and the code as well.

I think you caught me while I was creating the plugin side issue. :)

docs/compatibility.md
```diff
@@ -244,6 +244,13 @@ values occurring before the transition between the Julian and Gregorian calendar
 When writing `spark.sql.legacy.parquet.datetimeRebaseModeInWrite` is currently ignored as described
 [here](https://github.com/NVIDIA/spark-rapids/issues/144).
 
+When `spark.sql.parquet.outputTimestampType` is set to `INT96`, the timestamps will overflow and 
+result in an `IllegalArgumentException` thrown, if any value is before 
+September 21, 1677 12:12:43 AM or it is after April 11, 2262 11:47:17 PM. To get around this
+issue, turn off the ParquetWriter and ParquetReader acceleration by either setting 
+`spark.rapids.sql.format.parquet.writer.int96.enabled` to false or 
+`spark.sql.parquet.outputTimestampType` to `TIMESTAMP_MICROS` or `TIMESTAMP_MILLIS` 
+
```
razajafri, line 253: Let me change the wording of that sentence, that doesn't sound right

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -62,7 +63,14 @@ object GpuParquetFileFormat {
...
 
     if (sqlConf.writeLegacyParquetFormat) {
-      meta.willNotWorkOnGpu(s"Spark legacy format is not supported")
+      meta.willNotWorkOnGpu("Spark legacy format is not supported")
+    }
+
+    if (!meta.conf.isParquetInt96WriteEnabled && sqlConf.parquetOutputTimestampType ==
+      ParquetOutputTimestampType.INT96) {
+      meta.willNotWorkOnGpu("Writing INT96 is disabled, if you want to enable it turn it on by " +
+        "setting the `spark.rapids.sql.format.parquet.writer.int96.enabled` to true. NOTE: check " +
```
revans2, line 72: nit: we often use something like `s"setting the ${RapidsConf.ENABLE_PARQUET_INT96_WRITE.key} to true."` just to avoid any possible typos.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -273,6 +293,31 @@ class GpuParquetWriter(
               new GpuColumnVector(DataTypes.TimestampType, withResource(cv.getBase()) { v =>
                 v.castTo(DType.TIMESTAMP_MILLISECONDS)
               })
+            case DataTypes.TimestampType
+              if outputTimestampType == ParquetOutputTimestampType.INT96.toString =>
+                withResource(Scalar.fromLong(Long.MaxValue / 1000)) { upper =>
+                  withResource(Scalar.fromLong(Long.MinValue / 1000)) { lower =>
+                    withResource(cv.getBase().bitCastTo(DType.INT64).greaterOrEqualTo(upper)) {
```
revans2, line 300: bitCastTo returns a ColumnView that still has to be closed. You cannot chain them like this.

razajafri: Sorry, I keep forgetting the ColumnView is also a CPP side object and important to be closed

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
razajafri: Fixed

revans2 (dismissed):

razajafri:
build

razajafri:
build

jlowe (approved):

revans2 (approved):

PR #2274, opened 2021-04-27
Enable struct columns for GpuHashAggregateExec

### LINKED ISSUE #2273: [FEA] Enable struct type columns for GpuHashAggregateExec
RAPIDS for Spark currently does not support hash aggregation for nested types
```python
schema = StructType([
    StructField('name', 
        StructType([
            StructField('first', StringType()),
            StructField('last', StringType()),
        ])),
    StructField('intCol', IntegerType())
])
df = spark.createDataFrame(data, schema)

df.select(countDistinct('name')).explain('formatted')

!Exec <HashAggregateExec> cannot run on GPU because Nested types in grouping expressions are not supported;
```

Queries aggregating by struct-column values should be supported

project struct fields to flatten the struct

gerashegalov:
Fixes #2273
    

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2858,8 +2858,9 @@ object GpuOverrides {
     exec[HashAggregateExec](
       "The backend for hash based aggregations",
       ExecChecks(
-        (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.MAP + TypeSig.ARRAY)
-          .nested(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL),
+        (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL +
+          TypeSig.MAP + TypeSig.ARRAY + TypeSig.STRUCT)
+            .nested(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL),
```
jlowe, line 2863: Double-checking that this intends only to support single-level structs and no nesting whatsoever beyond the first level.  I ask because MapType is supported and is doubly-nested (array of structs), so wondering if we're supposed to support nested structs here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -97,12 +97,13 @@ class GpuHashAggregateMeta(
...
-      dtype.isInstanceOf[ArrayType] || dtype.isInstanceOf[StructType]
-        || dtype.isInstanceOf[MapType])) {
-      willNotWorkOnGpu("Nested types in grouping expressions are not supported")
-    }
+    agg.groupingExpressions
+      .find(_.dataType match {
+        case _@(ArrayType(_, _) | MapType(_, _, _)) => true
+        case _ => false
+      })
+      .foreach(_ => willNotWorkOnGpu("Nested types in grouping expressions are not supported"))
```
abellina, line 105:
```suggestion
      .foreach(_ => willNotWorkOnGpu("Array and Map in grouping expressions are not supported"))
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
gerashegalov: Yes, my intention is to first to add support for single-level structs

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -631,18 +631,20 @@ class ExecChecks private(
     val plan = meta.wrapped.asInstanceOf[SparkPlan]
     val allowDecimal = meta.conf.decimalTypeEnabled
 
-    val unsupportedOutputTypes = plan.output
-      .filterNot(attr => check.isSupportedByPlugin(attr.dataType, allowDecimal))
-      .toSet
+    val unsupportedOutputTypes = plan.output.collect {
+      case attr if !check.isSupportedByPlugin(attr.dataType, allowDecimal) => attr.dataType
```
gerashegalov, line 635: This was an unfortunate regression introduced by https://github.com/NVIDIA/spark-rapids/commit/64f9d9b4414103fa424283712371e2b9bddcd8fe#diff-c5cb89e178628729e961e2e0e9be2689893254638c5ce467250a263f7867112dR558

gerashegalov:
build

gerashegalov:
looking into the test failures

gerashegalov:
The discrepancy between my local test passing and the CI is that I tested with Spark 3.1.1. Looking into what needs to be fixed for Spark 3.0.x

gerashegalov:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
jlowe, line 2863: Should there be a ps note for the ARRAY and MAP type checks to note that they are not allowed in the grouping expressions?

jlowe (approved):

jlowe:
Has a merge conflict that needs to be resolved but otherwise lgtm.

gerashegalov:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1935,7 +1935,7 @@ object GpuOverrides {
       ExprChecks.fullAgg(
         TypeSig.LONG, TypeSig.LONG,
         repeatingParamCheck = Some(RepeatingParamCheck(
-          "input", TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all))),
+          "input", pluginSupportedOrderableSig, TypeSig.all))),
```
revans2, line 1938: This confuses me. Even if the signatures are the same, there is no reason why count needs input to be orderable. Also why limit count to single level structs, why not all structs or lists/etc? Am I missing something?

gerashegalov: > This confuses me. Even if the signatures are the same, there is no reason why count needs input to be orderable. 

If it's about the naming, I anticipated this question but forgot to rename the variable. 

> Also why limit count to single level structs, why not all structs or lists/etc? 

I think it's the same line of thinking as why only the single-level  structs were enabled for sort. This would be the first step to establish single-level as a baseline and then have a follow to allow deeper or arbitrary nesting. I believe it was suggested to me previously by @sameerz . Maybe it's no longer relevant.

abellina:
@gerashegalov and I discussed on the side one issue which I am suggesting a change for. As of this comment, the PR enables the struct type full stop, but we run into trouble if the struct isn't cast to string:

For example, I wrote a query like this, where `location` is a struct:
```
spark.sql("select location, count(*) from foo group by location").collect
```

But got this:
```
21/05/04 13:32:01 WARN TaskSetManager: Lost task 7.0 in stage 18.0 (TID 97) (192.168.50.80 executor 1): java.lang.IllegalArgumentException: StructType(StructField(lat,DoubleType,true), StructField(long,DoubleType,true)) is not supported for GPU processing yet.
	at com.nvidia.spark.rapids.GpuColumnVector.getNonNestedRapidsType(GpuColumnVector.java:479)
```

Now if `location` is cast to string beforehand, no issue is seen.

So I believe the approach here should be to allow it in the agg expression as @gerashegalov has done, but disable structs if the grouping expression or the aggregate function sees `StructType`. Anything but the `IllegalArgumentException` basically ;)

gerashegalov:
Thanks @abellina! The query corresponds  to the `NO_CAST `parameter in the tests in this PR. It curently crashes in the execution phase. I'll rework so it's intercepted in the planning phase for fallback.

gerashegalov:
build

jlowe (approved):

abellina (approved):

revans2 (approved):

PR #2285, opened 2021-04-28
Require single batch for full outer join streaming

### LINKED ISSUE #2278: [BUG] FullOuter join can produce too many results
This impacts all versions of the plugin. If you do a FullOuter join and happen to have multiple batches stream through the join the plugin will potentially output duplicate results. This is because if the build batch has keys that are not in the current stream batch the join will output them for that batch.

In the short term we can fall back to the CPU for FullOuter joins or we can have the stream side also require a single batch.  In the long term we are going to need to do a sort merge like join in the worst case.

revans2:
This fixes #2278 

But it is probably not the final solution. I will file a follow on issue to look into a sort merge join implementation that can do an out of core join for full outer, but could also be used for the other joins if the build side table is too large.

This also adds in configs for the various types of joins that we support.  This way if someone is hitting problems with a particular join that is running out or memory they will now have a bit more control over what goes back to the CPU instead of all SortMergeJoins.


revans2:
build

jlowe (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -611,6 +611,41 @@ object RapidsConf {
       .booleanConf
       .createWithDefault(false)
 
+  val ENABLE_INNER_JOIN = conf("spark.rapids.sql.join.inner.enabled")
```
gerashegalov, line 614: consider creating a joinType-based config generation using enumeration  https://stackoverflow.com/a/34550542/15132525 and a joinType-parametrized conf getter to avoid repetitive code throughout this PR.

gerashegalov (approved):
LGTM overall

revans2:
build

jlowe (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
revans2: `JoinType` is a sealed abstract class that has a type hierarchy to it that does not fit with the enum pattern you listed here.

`CrossJoin` is an `InnerLikeJoin` which is a `JoinType`.

There are also JoinType sub classes that we don't deal with yet (ExistanceJoin), and ones that we will never deal with because they are just used for planning (NaturalJoin and UsingJoin).

I like the idea but I don't know how to really implement it in practice.

gerashegalov: Then we could list explicitly the  join types we care about:
```scala
  val joinTypeEnabledMap: Map[String, ConfEntry[Boolean]] =
    Seq(
      Inner,
      Cross,
      LeftOuter,
      ...
    ).map { jt =>
    val key = s"spark.rapids.sql.join.${jt}.enabled"
    key -> conf(key)
      .doc(s"When set to true ${jt.sql} joins are enabled on the GPU")
      .booleanConf
      .createWithDefault(true)
  }.toMap
```

then in https://github.com/NVIDIA/spark-rapids/blob/575f645690a17065ee2c5175a54fdecc2e1631a2/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala#L1383

```scala
  def isJoinTypeEnabled(joinType: JoinType): Boolean = joinTypeEnabledMap
    .get(s"spark.rapids.sql.join.${joinType}.enabled")
    .map(confEntry => get(confEntry))
    .getOrElse(true)
```


and in https://github.com/NVIDIA/spark-rapids/pull/2285/files#diff-753a0d906f1f2d01d8bb08ef13e5205b1710cf915bb0f5fcdd55c3b665d2900dR28

```scala
object JoinTypeChecks {
  def tagForGpu(joinType: JoinType, meta: RapidsMeta[_, _, _]): Unit = {
    val conf = meta.conf
   if !conf.isJoinTypeEnabled(joinType) {
     meta.willNotWorkOnGpu("${joinType.sql} joins have been disabled. To enable set " +
            s"spark.rapids.sql.join.${joinType}.enabled" to true")
   }
}
```

PR #2289, opened 2021-04-28
Update docs to warn against 450.80.02 driver with 10.x toolkit

sameerz:
Update download.md to warn against 450.80.02 driver with 10.x toolkit

### DIFF
--- a/docs/download.md
+++ b/docs/download.md
@@ -100,6 +100,10 @@ The list of all supported operations is provided [here](supported_ops.md).
 For a detailed list of changes, please refer to the
 [CHANGELOG](https://github.com/NVIDIA/spark-rapids/blob/main/CHANGELOG.md). 
 
+**_Note:_** Using the CUDA 450.80.02 driver in combination with the CUDA 10.1 or 10.2 toolkit may
+result in long read times when reading a file that is snappy compressed.  If using the CUDA
+450.80.02 driver we recommend running with the CUDA 11.0 toolkit.  This issue is resolved in the 0.5
+and higher release.  
 
 ## Release v0.4.0
 ### Download v0.4.0
@@ -150,6 +154,10 @@ The list of all supported operations is provided [here](supported_ops.md).
 For a detailed list of changes, please refer to the
 [CHANGELOG](https://github.com/NVIDIA/spark-rapids/blob/main/CHANGELOG.md). 
 
+**_Note:_** Using the CUDA 450.80.02 driver in combination with the CUDA 10.1 or 10.2 toolkit may
+result in long read times when reading a file that is snappy compressed.  If using the CUDA
+450.80.02 driver we recommend running with the CUDA 11.0 toolkit.  This issue is resolved in the 0.5
+and higher release. 
 
 ## Release v0.3.0
 ### Download v0.3.0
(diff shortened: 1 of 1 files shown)


jlowe (dismissed):

jlowe:
Shouldn't this be targeting branch-0.5?

jlowe (approved):

jlowe:
build

docs/download.md
```diff
@@ -100,6 +100,10 @@ The list of all supported operations is provided [here](supported_ops.md).
 For a detailed list of changes, please refer to the
 [CHANGELOG](https://github.com/NVIDIA/spark-rapids/blob/main/CHANGELOG.md). 
 
+**_Note:_** Using the CUDA 450.80.02 driver in combination with the CUDA 10.1 or 10.2 toolkit may
```
viadea, line 103: If this issue only happens on specific Driver version, could we ask users to use a different(or newer) version of driver other than 450.80.02?

jlowe: I don't think a driver with the fix is available yet.

tgravescs, line 103: we should probabaly say 450.80.02 or newer driver. I don't think they put out a fixed version yet, once they do we would want to update and say 450.80.02 thru XXXX version

docs/download.md
```diff
@@ -100,6 +100,10 @@ The list of all supported operations is provided [here](supported_ops.md).
...
+result in long read times when reading a file that is snappy compressed.  If using the CUDA
+450.80.02 driver we recommend running with the CUDA 11.0 toolkit.  This issue is resolved in the 0.5
```
tgravescs, line 105: do we want to change 0.5 to v0.5.0 to match syntax above?

docs/download.md
viadea, line 103:
```suggestion
**_Note:_** Using the NVIDIA driver 450.80.02 in combination with the CUDA 10.1 or 10.2 toolkit may
```

viadea: If we use "NVIDIA driver" term which can match the download link's term. This can avoid confusing between "CUDA toolkit" version vs "CUDA driver" version.
And also this is what mentioned in https://docs.nvidia.com/deploy/cuda-compatibility/index.html#cuda-intro

docs/download.md
```diff
@@ -150,6 +154,10 @@ The list of all supported operations is provided [here](supported_ops.md).
 For a detailed list of changes, please refer to the
 [CHANGELOG](https://github.com/NVIDIA/spark-rapids/blob/main/CHANGELOG.md). 
 
+**_Note:_** Using the CUDA 450.80.02 driver in combination with the CUDA 10.1 or 10.2 toolkit may
```
viadea, line 157:
```suggestion
**_Note:_** Using the NVIDIA driver 450.80.02 in combination with the CUDA 10.1 or 10.2 toolkit may
```

docs/download.md
```diff
@@ -197,6 +205,10 @@ The list of all supported operations is provided [here](supported_ops.md).
...
```
viadea, line 208:
```suggestion
**_Note:_** Using the NVIDIA driver 450.80.02 in combination with the CUDA 10.1 or 10.2 toolkit may
```

docs/download.md
```diff
@@ -247,6 +259,10 @@ The list of all supported operations is provided
...
```
viadea, line 262:
```suggestion
**_Note:_** Using the NVIDIA driver 450.80.02 in combination with the CUDA 10.1 or 10.2 toolkit may
```

viadea (approved):

docs/download.md
sameerz: Updated the text to indicate a user can use a newer version of the driver.

docs/download.md
sameerz: Yes, updated to 0.5.0

docs/download.md
sameerz: Updated to refer to the Nvidia driver.

sameerz: Found that the subsequent 450.102.04 did not have the fix, but versions after that did.  Updated the docs to reflect that.

sameerz:
build

viadea (approved):

docs/download.md
tgravescs: what about 460.X drivers?  I assume same bug applies to them until certain version, if they put out fix?

PR #2295, opened 2021-04-29
batch small buffers when spilling via GDS

rongou:

When spilling via GDS, if the user specifies a large number of shuffle partitions, each shuffle buffer can be fairly small and very inefficient to write to disk individually. This PR adds a batch write buffer that collects these smaller buffers and writes them out in a single `cuFileWrite` call. Preliminary results show this greatly increases the write throughput and reduces the overall query time.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
@@ -19,18 +19,26 @@ package com.nvidia.spark.rapids
 import java.io.File
 import java.util.concurrent.ConcurrentHashMap
 
+import scala.collection.mutable.ArrayBuffer
+
 import ai.rapids.cudf._
 import com.nvidia.spark.rapids.StorageTier.StorageTier
 import com.nvidia.spark.rapids.format.TableMeta
 
-import org.apache.spark.sql.rapids.RapidsDiskBlockManager
+import org.apache.spark.sql.rapids.{RapidsDiskBlockManager, TempSpillBufferId}
 
 /** A buffer store using GPUDirect Storage (GDS). */
 class RapidsGdsStore(
     diskBlockManager: RapidsDiskBlockManager,
     catalog: RapidsBufferCatalog = RapidsBufferCatalog.singleton)
     extends RapidsBufferStore(StorageTier.GDS, catalog) with Arm {
+  private val BlockSize = 4096;
+  private val BatchWriteBufferSize = 1 << 27; // 128 MiB
   private[this] val sharedBufferFiles = new ConcurrentHashMap[RapidsBufferId, File]
+  private[this] val batchWriteBuffer = CuFileBuffer.allocate(BatchWriteBufferSize, true)
+  private[this] var batchWriteBufferId = TempSpillBufferId()
+  private[this] var batchWriteBufferOffset = 0L
+  private[this] val pendingBuffers = ArrayBuffer.empty[RapidsGdsBuffer]
 
   override protected def createBuffer(other: RapidsBuffer, otherBuffer: MemoryBuffer,
       stream: Cuda.Stream): RapidsBufferBase = {
@@ -39,30 +47,75 @@ class RapidsGdsStore(
         case d: DeviceMemoryBuffer => d
         case _ => throw new IllegalStateException("copying from buffer without device memory")
       }
-      val id = other.id
-      val path = if (id.canShareDiskPaths) {
-        sharedBufferFiles.computeIfAbsent(id, _ => id.getDiskPath(diskBlockManager))
+      if (isBatched(deviceBuffer.getLength)) {
+        batchSpill(other, deviceBuffer)
       } else {
-        id.getDiskPath(diskBlockManager)
+        singleShotSpill(other, deviceBuffer)
       }
-      // When sharing files, append to the file; otherwise, write from the beginning.
-      val fileOffset = if (id.canShareDiskPaths) {
-        // only one writer at a time for now when using shared files
-        path.synchronized {
-          CuFile.appendDeviceBufferToFile(path, deviceBuffer)
-        }
-      } else {
-        CuFile.writeDeviceBufferToFile(path, 0, deviceBuffer)
-        0
-      }
-      logDebug(s"Spilled to $path $fileOffset:${other.size} via GDS")
-      new RapidsGdsBuffer(id, fileOffset, other.size, other.meta, other.getSpillPriority,
-        other.spillCallback)
     }
+  }
+
+  private def isBatched(length: Long): Boolean = {
+    length < BatchWriteBufferSize
+  }
+
+  private def singleShotSpill(
+      other: RapidsBuffer, deviceBuffer: DeviceMemoryBuffer): RapidsBufferBase = {
+    val id = other.id
+    val path = if (id.canShareDiskPaths) {
+      sharedBufferFiles.computeIfAbsent(id, _ => id.getDiskPath(diskBlockManager))
+    } else {
+      id.getDiskPath(diskBlockManager)
+    }
+    // When sharing files, append to the file; otherwise, write from the beginning.
+    val fileOffset = if (id.canShareDiskPaths) {
+      // only one writer at a time for now when using shared files
+      path.synchronized {
+        CuFile.appendDeviceBufferToFile(path, deviceBuffer)
+      }
+    } else {
+      CuFile.writeDeviceBufferToFile(path, 0, deviceBuffer)
+      0
+    }
+    logDebug(s"Spilled to $path $fileOffset:${other.size} via GDS")
+    new RapidsGdsBuffer(id, None, fileOffset, other.size, other.meta, other.getSpillPriority,
+      other.spillCallback)
+  }
+
+  private def batchSpill(
+      other: RapidsBuffer, deviceBuffer: DeviceMemoryBuffer): RapidsBufferBase = this.synchronized {
+      if (deviceBuffer.getLength > BatchWriteBufferSize - batchWriteBufferOffset) {
+        val path = batchWriteBufferId.getDiskPath(diskBlockManager).getAbsolutePath
+        withResource(new CuFileWriteHandle(path)) { handle =>
+          handle.write(batchWriteBuffer, batchWriteBufferOffset, 0)
+          logDebug(s"Spilled to $path 0:$batchWriteBufferOffset via GDS")
+        }
+        pendingBuffers.foreach(_.cuFileBuffer = None)
+        pendingBuffers.clear
+        batchWriteBufferId = TempSpillBufferId()
+        batchWriteBufferOffset = 0
+      }
+
+      val currentBufferOffset = batchWriteBufferOffset
+      batchWriteBuffer.copyFromMemoryBuffer(
+        batchWriteBufferOffset, deviceBuffer, 0, deviceBuffer.getLength, Cuda.DEFAULT_STREAM)
+      batchWriteBufferOffset += alignUp(deviceBuffer.getLength)
+
+      val id = other.id
+      sharedBufferFiles.computeIfAbsent(id, _ => batchWriteBufferId.getDiskPath(diskBlockManager))
+      val gdsBuffer = new RapidsGdsBuffer(id, Some(batchWriteBuffer), currentBufferOffset,
+        other.size, other.meta, other.getSpillPriority, other.spillCallback)
+      pendingBuffers += gdsBuffer
+      gdsBuffer
+    }
+
+  private def alignUp(length: Long): Long = {
+    (length + BlockSize - 1) & ~(BlockSize - 1)
   }
 
   class RapidsGdsBuffer(
       id: RapidsBufferId,
+      var cuFileBuffer: Option[CuFileBuffer],
       val fileOffset: Long,
       size: Long,
       meta: TableMeta,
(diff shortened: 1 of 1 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -39,30 +47,75 @@ class RapidsGdsStore(
...
+      if (deviceBuffer.getLength > BatchWriteBufferSize - batchWriteBufferOffset) {
+        val path = batchWriteBufferId.getDiskPath(diskBlockManager).getAbsolutePath
+        withResource(new CuFileWriteHandle(path)) { handle =>
+          handle.write(batchWriteBuffer, batchWriteBufferOffset, 0)
+          logDebug(s"Spilled to $path 0:$batchWriteBufferOffset via GDS")
         }
-      } else {
-        CuFile.writeDeviceBufferToFile(path, 0, deviceBuffer)
-        0
+        pendingBuffers.foreach(_.cuFileBuffer = None)
```
jlowe, line 93: I'm confused by this line.  `CuFileBuffer` resources need to be explicitly closed, we can't just overwrite the references with `None`.  To me this seems like a clear leak or if this is being closed earlier then the `None` overwrite should occur there rather than here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -39,30 +47,75 @@ class RapidsGdsStore(
...
+      0
+    }
+    logDebug(s"Spilled to $path $fileOffset:${other.size} via GDS")
+    new RapidsGdsBuffer(id, None, fileOffset, other.size, other.meta, other.getSpillPriority,
+      other.spillCallback)
+  }
+
+  private def batchSpill(
+      other: RapidsBuffer, deviceBuffer: DeviceMemoryBuffer): RapidsBufferBase = this.synchronized {
+      if (deviceBuffer.getLength > BatchWriteBufferSize - batchWriteBufferOffset) {
```
jlowe, line 87: Indentation is off starting here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -19,18 +19,26 @@ package com.nvidia.spark.rapids
...
 
-import org.apache.spark.sql.rapids.RapidsDiskBlockManager
+import org.apache.spark.sql.rapids.{RapidsDiskBlockManager, TempSpillBufferId}
 
 /** A buffer store using GPUDirect Storage (GDS). */
 class RapidsGdsStore(
     diskBlockManager: RapidsDiskBlockManager,
     catalog: RapidsBufferCatalog = RapidsBufferCatalog.singleton)
     extends RapidsBufferStore(StorageTier.GDS, catalog) with Arm {
+  private val BlockSize = 4096;
```
jlowe, line 35:
```suggestion
  private val BlockSize = 4096
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -39,30 +47,75 @@ class RapidsGdsStore(
...
+    }
+    logDebug(s"Spilled to $path $fileOffset:${other.size} via GDS")
+    new RapidsGdsBuffer(id, None, fileOffset, other.size, other.meta, other.getSpillPriority,
+      other.spillCallback)
+  }
+
+  private def batchSpill(
+      other: RapidsBuffer, deviceBuffer: DeviceMemoryBuffer): RapidsBufferBase = this.synchronized {
+      if (deviceBuffer.getLength > BatchWriteBufferSize - batchWriteBufferOffset) {
+        val path = batchWriteBufferId.getDiskPath(diskBlockManager).getAbsolutePath
```
jlowe, line 88: Rather than cache the `TempSpillBufferId` which is only used to obtain the disk path, we should be caching the disk path.  Every time we call `getDiskPath` it is calling through to the disk block manager which needs to do existence checks on the local filesystem.  Those redundant syscalls can be avoided if we cache the path rather than the ID, and that makes the code simpler to follow since we only care about the path anyway.  The temp ID is just one of many ways to get a temp path.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -39,30 +47,75 @@ class RapidsGdsStore(
...
+        withResource(new CuFileWriteHandle(path)) { handle =>
+          handle.write(batchWriteBuffer, batchWriteBufferOffset, 0)
+          logDebug(s"Spilled to $path 0:$batchWriteBufferOffset via GDS")
         }
-      } else {
-        CuFile.writeDeviceBufferToFile(path, 0, deviceBuffer)
-        0
+        pendingBuffers.foreach(_.cuFileBuffer = None)
+        pendingBuffers.clear
```
jlowe, line 94: I see the code writing `pendingBuffers` but otherwise never reading it.  I expected to see `materializeMemoryBuffer` and `copyToMemoryBuffer` to check if this buffer was one of the pending buffers and do the copy from device memory rather than the fetch from GDS first, but that logic seems to be missing.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -74,39 +127,48 @@ class RapidsGdsStore(
...
+            dm.copyFromMemoryBuffer(
+              0, cuFileBuffer.get, fileOffset + srcOffset, size, Cuda.DEFAULT_STREAM)
+          }
         case _ => throw new IllegalStateException(
           s"GDS can only copy to device buffer, not ${dst.getClass}")
       }
     }
 
     override protected def releaseResources(): Unit = {
       // Buffers that share paths must be cleaned up elsewhere
```
jlowe, line 170: This comment isn't addressed in this PR.  Currently when buffers share files there needs to be something external to the store that cleans up the files that are shared.  Before this, the only buffers sharing files were shuffle buffers, and the shuffle manager cleans up all the files when the shuffle session is garbage collected, so we don't accumulate files forever.  I don't see anything like that happening here, so it seems for a long-running GDS session there's nothing that will ever reap the files corresponding to batched buffers even if all the buffers in the batch have been destroyed.

jlowe (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
jlowe: Wondering if it would be clearer and more consistent if this did use reference-counting, otherwise we've got some code that needs to close a cufilebuffer and other code that _must not_ close the cufile buffer, and we may confuse which is which.  If we always close when we're done then it could be simpler.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -18,19 +18,28 @@ package com.nvidia.spark.rapids
...
+import org.apache.spark.sql.rapids.{RapidsDiskBlockManager, TempSpillBufferId}
 
 /** A buffer store using GPUDirect Storage (GDS). */
 class RapidsGdsStore(
     diskBlockManager: RapidsDiskBlockManager,
     catalog: RapidsBufferCatalog = RapidsBufferCatalog.singleton)
     extends RapidsBufferStore(StorageTier.GDS, catalog) with Arm {
-  private[this] val sharedBufferFiles = new ConcurrentHashMap[RapidsBufferId, File]
+  private val blockSize = 4096
+  private val batchWriteBufferSize = 1 << 27 // 128 MiB
```
jlowe, line 37: Curious, how did we arrive at this number?  Experimentally or just guessing?  Would be nice to make it as low as possible as it's a permanent allocation.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -39,80 +48,166 @@ class RapidsGdsStore(
...
   class RapidsGdsBuffer(
       id: RapidsBufferId,
+      path: File,
       val fileOffset: Long,
       size: Long,
       meta: TableMeta,
       spillPriority: Long,
-      spillCallback: RapidsBuffer.SpillCallback)
+      spillCallback: RapidsBuffer.SpillCallback,
+      var isPending: Boolean)
```
jlowe, line 153: After thinking about the general design some more, I don't think this, or the previous `Option` approach, is going to work in light of multithreaded processing and spill. Consider the following scenario:
1. Buffer is asked to materialize/copy the buffer
1. It checks the pending flag and sees that it is pending
1. Current thread that will materialize/copy the pending buffer gets preempted
1. Another thread triggers a spill and causes the batched buffer to flush
1. Data from another buffer starts being written to the big batch buffer
1. Original thread now resumes and proceeds to copy from the big batch buffer but risks copying unrelated data, corrupting the result

There needs to be synchronized access to the pending set to determine safely whether a buffer is pending and its data is guaranteed to be in the batched buffer for the duration of a transfer.  I don't think poking a flag/option asynchronously in each buffer is going to be safe.

jlowe (commented):
It would be good to have at least some testing of this new functionality, for example a test that verifies that buffers below the threshold are collected into a single file, large buffers are not, and that the batched file is removed only after the last buffer that was in the batch is freed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
rongou: Yes, that's why I added the synchronized blocks to the copying/flushing methods. The store can only flush a buffer (i.e. clear the pending status) after it's done copying.

abellina (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
jlowe: But what about the `stream` argument?  We can leave `copyToMemoryBuffer` and unlock before the copy has actually completed, and if another copy occurs on a different stream (i.e.: DEFAULT_STREAM vs. the stream passed to `copyToMemoryBuffer`) I think we still could have issues.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
abellina: UCX is using BAR1 aggressively too. We normally max out the BAR1 usage by splitting it between send and receive buffers. Make sure that it isn't UCX doing the mapping (we configure it using the bounce buffer pool size).

rongou: Ok, made this configurable and default to 8 MiB.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
rongou: Added stream synchronize calls.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -18,19 +18,28 @@ package com.nvidia.spark.rapids
...
 
 /** A buffer store using GPUDirect Storage (GDS). */
 class RapidsGdsStore(
     diskBlockManager: RapidsDiskBlockManager,
     catalog: RapidsBufferCatalog = RapidsBufferCatalog.singleton)
     extends RapidsBufferStore(StorageTier.GDS, catalog) with Arm {
-  private[this] val sharedBufferFiles = new ConcurrentHashMap[RapidsBufferId, File]
+  private val blockSize = 4096
+  private val batchWriteBufferSize = 1 << 27 // 128 MiB
+  private[this] val batchedBuffers = new ConcurrentHashMap[File, Set[RapidsBufferId]]
```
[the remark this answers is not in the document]
rongou: Broke apart the two cases into separate classes.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -39,12 +44,31 @@ class RapidsGdsStore(
...
+      override val meta: TableMeta,
+      spillPriority: Long,
+      override val spillCallback: RapidsBuffer.SpillCallback)
+      extends RapidsBufferBase(id, size, meta, spillPriority, spillCallback) {
+    override val storageTier: StorageTier = StorageTier.GDS
+
+    override def getMemoryBuffer: MemoryBuffer = getDeviceMemoryBuffer
+  }
+
+  class SingleShotSpiller() {
```
jlowe, line 67: I like seeing the batch logic encapsulated in a separate class to keep the main GDS store cleaner, but in this single-shot case it doesn't seem worth it. This class has no instance data, implements no common interface, and provides only one method.  This makes me wonder why not move `RapidsGdsSingleShotBuffer` out of this class and just make the `spill` method here named `singleShotSpill` in RapidsGdsStore directly.  I'm fine if we keep it, but it should be an `object` rather than a `class` since the method can be static.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -56,64 +80,177 @@ class RapidsGdsStore(
...
-        sharedBufferFiles.get(id)
-      } else {
-        id.getDiskPath(diskBlockManager)
+      override def copyToMemoryBuffer(srcOffset: Long, dst: MemoryBuffer, dstOffset: Long,
+          length: Long, stream: Cuda.Stream): Unit = {
+        dst match {
+          case dmOriginal: DeviceMemoryBuffer =>
+            val dm = dmOriginal.slice(dstOffset, length)
+            // TODO: switch to async API when it's released, using the passed in CUDA stream.
+            CuFile.readFileToDeviceBuffer(dm, path, fileOffset + srcOffset)
```
jlowe, line 106: Is CuFile using the system default stream here?  Wondering if we need to do any synchronization on `stream` before calling the `CuFile` interface since it's not taking the stream parameter.

rongou: Let me check with the GDS team.

jlowe (commented):
Thanks for the updates, @rongou!  All I think is left are the tests.

rongou:
Added tests. PTAL

rongou:
build

jlowe (approved):

Issue #2011, opened 2021-03-24
[FEA] Support creation of 2D array type


Dooyoung-Hwang:
### Creation of 2D array type with GPU plan

I request a feature to create a ArrayType[ArrayType[AtomicType]] by applying the Array function to ArrayType.

- Test code
```python
data = [(1,0.1, 0.4), (2, 0.2, 0.5), (3, -1.0, 2.2)]

schema = StructType([ \
    StructField("a",IntegerType(),True), \
    StructField("b",FloatType(),True), \
    StructField("c",FloatType(),True)
  ])

df1 = spark.createDataFrame(data=data,schema=schema).withColumn("arr1", array("b", "c")).withColumn("arrOfArray", array("arr1"))
print(df1._jdf.queryExecution())
```

- Output message from driver log.
  !Expression array(array(b#6029, c#6030)) cannot run on GPU because expression CreateArray array(b#6029, c#6030) produces an unsupported type ArrayType(FloatType,true); expression CreateArray array(array(b#6029, c#6030)) produces an unsupported type ArrayType(ArrayType(FloatType,true),false)

revans2:
This should be doable.  There really is only one thing preventing this from fully being implemented #1902. This is not a blocker, because we could just disallow cases where we cannot make it work.

sameerz:
Depends on https://github.com/rapidsai/cudf/issues/5887

sameerz:
Depends on https://github.com/rapidsai/cudf/issues/8077

wbo4958:
The current native CUDF has supported the nested List data type and the java bindings seems only support 1-D List build from ColumnVector. I can think of 2 solutions based on the above information.

The first one is to leverage the existing [makeList](https://github.com/rapidsai/cudf/blob/branch-0.20/java/src/main/native/src/ColumnVectorJni.cpp#L126) API to accept ArrayList<ArrayList<ColumnVector> > (2D ColumnVector), and then 
1. flatten the columns to 1D to leverage the cudf::interleave_columns to generate child data column
2. generate child offset according to dimensions of columns
3. create a nested child list column
4. generate the parent offset
5. create a parent list column

The second one is to ask cudf side to provide an API to create nested list columns.

the advantage of 1st solution is we don't depend on anything, the  disadvantage of 1st solution is we can support 2D, but we don't support 3D, 4D ...

The advantage of 2nd solution is we can support any dimensions array type. the disadvantage is we need to depend on cudf API.

[...]

PR #2299, opened 2021-04-29
support creating array of array

wbo4958:
This PR is for https://github.com/NVIDIA/spark-rapids/issues/2011

Currently, Cudf only supports creating 1D or 2D array, See https://github.com/rapidsai/cudf/pull/8046

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
@@ -250,8 +250,8 @@ abstract class GpuExplodeBase extends GpuUnevaluableUnaryExpression with GpuGene
 
   override def fixedLenLazyExpressions: Seq[Expression] = child match {
     // GpuLiteral of ArrayData will be converted to GpuCreateArray with GpuLiterals
-    case GpuCreateArray(expressions, _) => expressions
-    case GpuAlias(GpuCreateArray(expressions, _), _) => expressions
+    case GpuCreateArray(expressions, _, _) => expressions
+    case GpuAlias(GpuCreateArray(expressions, _, _), _) => expressions
     case _ => Nil
   }
 
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
@@ -250,7 +250,7 @@ abstract class GpuExplodeBase extends GpuUnevaluableUnaryExpression with GpuGene
 
   override def fixedLenLazyExpressions: Seq[Expression] = child match {
     // GpuLiteral of ArrayData will be converted to GpuCreateArray with GpuLiterals
-    case GpuCreateArray(expressions, __) => expressions
+    case GpuCreateArray(expressions, _) => expressions
     case GpuAlias(GpuCreateArray(expressions, _), _) => expressions
     case _ => Nil
   }
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
@@ -250,8 +250,8 @@ abstract class GpuExplodeBase extends GpuUnevaluableUnaryExpression with GpuGene
 
   override def fixedLenLazyExpressions: Seq[Expression] = child match {
     // GpuLiteral of ArrayData will be converted to GpuCreateArray with GpuLiterals
-    case GpuCreateArray(expressions, _, _) => expressions
-    case GpuAlias(GpuCreateArray(expressions, _, _), _) => expressions
+    case GpuCreateArray(expressions, __) => expressions
+    case GpuAlias(GpuCreateArray(expressions, _), _) => expressions
     case _ => Nil
   }
 
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -2369,9 +2369,8 @@ object GpuOverrides {
                 TypeSig.BOOLEAN + TypeSig.DATE + TypeSig.TIMESTAMP),
           TypeSig.all))),
       (in, conf, p, r) => new ExprMeta[CreateArray](in, conf, p, r) {
-        override def convertToGpu(): GpuExpression = {
+        override def convertToGpu(): GpuExpression =
           GpuCreateArray(childExprs.map(_.convertToGpu()), wrapped.useStringTypeWhenEmpty)
-        }
       }),
     expr[StringLocate](
       "Substring search operator",
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -2366,9 +2366,19 @@ object GpuOverrides {
           TypeSig.numeric + TypeSig.NULL + TypeSig.STRING +
               TypeSig.BOOLEAN + TypeSig.DATE + TypeSig.TIMESTAMP +
               TypeSig.ARRAY.nested(TypeSig.numeric + TypeSig.NULL + TypeSig.STRING +
-                TypeSig.BOOLEAN + TypeSig.DATE + TypeSig.TIMESTAMP),
+                TypeSig.BOOLEAN + TypeSig.DATE + TypeSig.TIMESTAMP + TypeSig.ARRAY),
           TypeSig.all))),
       (in, conf, p, r) => new ExprMeta[CreateArray](in, conf, p, r) {
+
+        override def tagExprForGpu(): Unit = {
+          wrapped.dataType match {
+            case ArrayType(ArrayType(ArrayType(_, _), _), _) =>
+              willNotWorkOnGpu("Only support to create array or array of array, Found: " +
+                s"${wrapped.dataType}")
+            case _ =>
+          }
+        }
+
         override def convertToGpu(): GpuExpression =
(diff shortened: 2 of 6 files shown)


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -77,7 +77,7 @@ case class GpuCreateArray(children: Seq[Expression], useStringTypeWhenEmpty: Boo
         }
       }
       GpuColumnVector.from(ColumnVector.makeList(numRows,
-        GpuColumnVector.getNonNestedRapidsType(dataType.elementType),
+        GpuColumnVector.getArrayNestedRapidsType(dataType.elementType),
```
razajafri, line 80:
```suggestion
      GpuColumnVector.from(ColumnVector.makeList(numRows,
        dataType.elementType match {
          case _:ArrayType => DType.LIST
          case _ => GpuColumnVector.getNonNestedRapidsType(dataType.elementType)
        },
        columns: _*), dataType)
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2284,13 +2284,16 @@ object GpuOverrides {
       }),
     expr[CreateArray](
       " Returns an array with the given elements",
+      // Cudf only supports 2D Array.
```
revans2, line 2287: This is not true cudf can support very nested arrays.  If a nested type is in nested, then it implies recursion.

```
TypeSig.Array.nested(Typeside.BOOLEAN + TypeSig.ARRAY)
```
Matches `ArrayType(BooleanType)` and `ArrayType(ArrayType(BooleanType))` but also as deeply nested as we want to go. `ArrayType(ArrayType(ArrayType(...(BooleanType))))))))))))))))`

If you only want one level deep then we re going to have to make some changes to how the type checking code works.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
revans2: This is NOT going to work. We need CUDF API changes.  makeList needs the dataType so it can know what to do if we are creating an empty column (no rows).  I'm not sure how often that will happen in practice but if it does happen this code will produce a bogus column. The type will not make any since because there will be a LIST column with no child data column.

revans2 (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
wbo4958: @revans2, Yeah, For now, seems the API makeList can only create 2D Array. So Seems the current TypeCheck is no longer cover this scenario.  Can you share some ideas about this?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
revans2: I added new comments for what I want.  Really all it ends up being is a comment in the code explaining why the situation and why it is OK to do what we are doing.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
```diff
@@ -247,6 +247,17 @@ class LiteralExprMeta(
     p: Option[RapidsMeta[_, _, _]],
     r: DataFromReplacementRule) extends ExprMeta[Literal](lit, conf, p, r) {
 
+  override def tagExprForGpu(): Unit = {
+    lit.dataType match {
+      case ArrayType(_, _) =>
+        if (lit.value == null) {
+          willNotWorkOnGpu("array(array(x), null) is not supported." +
+            "Please try array(array(x), array(null))")
```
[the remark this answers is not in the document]
wbo4958: Done.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -77,8 +76,10 @@ case class GpuCreateArray(children: Seq[Expression], useStringTypeWhenEmpty: Boo
         }
       }
       GpuColumnVector.from(ColumnVector.makeList(numRows,
```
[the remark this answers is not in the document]
wbo4958: Done.

wbo4958:
build

wbo4958:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -27,8 +27,9 @@ import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types.{ArrayType, DataType, Metadata, NullType, StringType, StructField, StructType}
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
-case class GpuCreateArray(children: Seq[Expression], useStringTypeWhenEmpty: Boolean)
-    extends GpuExpression {
+case class GpuCreateArray(children: Seq[Expression], useStringTypeWhenEmpty: Boolean,
+    /* use refer type for children whose type can't be determined */
+    referType: Option[DataType] = None) extends GpuExpression {
```
revans2, line 32: Why can it not be determined? each `Expression` has a type associated with it. In what situation is there a `Expression` that says it is the wrong type? Is this because we don't support array literals? The PRs are in progress for array literal support and I am inclined to just hold off on this and all of the work arounds in it until we actually have proper support.

wbo4958: Hi @revans2, Sorry for my unclear comment in the code. What I wanted to say is. For the case `array(array(1), array()`, with this PR, it will be converted to
``` console
GpuCreateArray
     |--children(0):  GpuCreateArray
               |-- GpuLiteral(value = 1, dataType=IntegerType).
     |--children(1):  GpuCreateArray no children
```
So the dataType of child(0) will be ArrayType(IntegerType) and the dataType of  child(1) will be `ArrayType(NullType) or ArrayType(StringType) that depends on LEGACY_CREATE_EMPTY_COLLECTION_USING_STRING_TYPE`

Later in somewhere, GpuCreateArray will be resolved and call `checkInputDataTypes` which requires all the children should have the same data type. If false. the GpuCreateArray will become un-resolved and cause some issues. So I add a refer type (from parent element of array) for the case no children and bypass `checkInputDataTypes`.

wbo4958:
build

revans2:
Can we please not rebase. It can really confuse github with outstanding commits. Just merge.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2360,13 +2360,25 @@ object GpuOverrides {
...
-              TypeSig.BOOLEAN + TypeSig.DATE + TypeSig.TIMESTAMP,
+              TypeSig.BOOLEAN + TypeSig.DATE + TypeSig.TIMESTAMP +
+              TypeSig.ARRAY.nested(TypeSig.numeric + TypeSig.NULL + TypeSig.STRING +
+                TypeSig.BOOLEAN + TypeSig.DATE + TypeSig.TIMESTAMP + TypeSig.ARRAY),
           TypeSig.all))),
       (in, conf, p, r) => new ExprMeta[CreateArray](in, conf, p, r) {
+
+        override def tagExprForGpu(): Unit = {
+          wrapped.dataType match {
+            case ArrayType(ArrayType(ArrayType(_, _), _), _) =>
```
revans2, line 2375: Where did this limitation come from? Does cudf only support so much nesting?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -250,7 +250,7 @@ abstract class GpuExplodeBase extends GpuUnevaluableUnaryExpression with GpuGene
 
   override def fixedLenLazyExpressions: Seq[Expression] = child match {
     // GpuLiteral of ArrayData will be converted to GpuCreateArray with GpuLiterals
-    case GpuCreateArray(expressions, _) => expressions
+    case GpuCreateArray(expressions, __) => expressions
```
revans2, line 253: Why do we have a `__` in here instead of a `_`?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
wbo4958: I tested like array(array(array(a))) and libcudf will throw the exception, I read the PR for making list and found there is a comment  in  `Currently, only lists columns of one depth level are supported.` https://github.com/rapidsai/cudf/pull/8046

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
wbo4958: Done

wbo4958:
build

wbo4958:
build

revans2:
Please do not rebase your patch. Just merge in the new branch. It is much less likely to confuse github.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
revans2: Could you file a follow on issue for more deeply nested? It is not something we have to work on. I just want to be sure that it is documented.

revans2 (approved):

jlowe (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
wbo4958: sure, filed the following issue https://github.com/NVIDIA/spark-rapids/issues/2453

Issue #20, opened 2020-05-28
[FEA] Support out of core joins


revans2:
**Is your feature request related to a problem? Please describe.**
Some joins can be very large and with data skew can make it difficult to do what we do now where we hold one of the join tables in memory while streaming through the other side in batches.

This is to detect when that is not going to work and switch over to a sort merge join when needed preferably using #19 for the sorting.

sameerz:
We should see what Blazing's approach for out of core joins is.

revans2:
We are working with cudf on getting something that can just return the gather maps 

https://github.com/rapidsai/cudf/issues/6480

and then we can try and chunk the output result with a little help from

https://github.com/rapidsai/cudf/issues/7408

jlowe:
@revans2 do you consider this addressed by the recent join work or are there still items left to address?

revans2:
@jlowe I am not ready to say that yet. It is really close and I would not have a problem if we wanted to close this and just track the follow on work for a sort merge join fallback with #2252

This is required to be able to support FullOuter join as out of core, and also to be able to support large join data on both sides. Right now all of the joins except cross join require that we can fit all of the data for at least one side in a single batch. In some cases you can work around this with more partitioning, but not in all cases.

jlowe:
Since it's not done I'd rather leave it open to track.  I suspected there was some extra work left, and now that you've linked what it is here, I'm good. Thanks!

revans2:
Sorry I pointed to the wrong issue. It should be #2354. I thought I had hit the button to file it but didn't and my search for "sort merge" turned up the wrong issue.

PR #2310, opened 2021-04-29
Allow batching the output of a join

revans2:
This is the first step for out of core join. This at least partially addresses #20 

This depends on https://github.com/rapidsai/cudf/pull/8118 to go in first.

For most cases that I have tested this is strictly better than what was before. If the output of the join fits in the output batch size then the join will happen just like it does today.  If the output is larger than that we now can output it in multiple batches.  The problem that I have found is that the gather map is not spillable and after a single batch is output the GPU Semaphore is released. This means that for contrived joins that explode evenly, each active task will have a potentially large gather map in memory.  I think I can make it spillable without a lot of work. If I can then I might just do it. But I also want to spend some time running benchmarks to see if this can help fix some of the exploding join issues have have seen there.


integration_tests/src/main/python/asserts.py
```diff
@@ -28,15 +28,15 @@
...
+        assert len(cpu) == len(gpu), "CPU and GPU row have different lengths at {} CPU: {} GPU: {}".format(path, len(cpu), len(gpu))
         if hasattr(cpu, "__fields__") and hasattr(gpu, "__fields__"):
             for field in cpu.__fields__:
                 _assert_equal(cpu[field], gpu[field], float_check, path + [field])
         else:
             for index in range(len(cpu)):
                 _assert_equal(cpu[index], gpu[index], float_check, path + [index])
     elif (t is list):
-        assert len(cpu) == len(gpu), "CPU and GPU list have different lengths at {}".format(path)
+        assert len(cpu) == len(gpu), "CPU and GPU row have different lengths at {} CPU: {} GPU: {}".format(path, len(cpu), len(gpu))
```
revans2, line 39: put list back in, it is not a row...

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -79,12 +81,669 @@ object GpuHashJoin {
...
+ * All instances should be spillable.
+ * The life cycle of this assumes that when it is created that the data and
+ * gather maps will be used shortly.
+ * If you are not going to use these for a while, like when returning from an iterator,
+ * then allowSpilling should be called so that the cached data is released and spilling
+ * can be allowed.  If you need/want to use the data again, just start using it, and it
+ * will be cached yet again until allowSpilling is called.
+ * When you are completely done with this object call close on it.
+ */
+trait JoinGatherer extends AutoCloseable with Arm {
```
revans2, line 150: Would it be better to move this to its own file? I would like to use it with the broadcast nested loop join/cross join code too.

revans2 (commented):
I'll try to add in some more java docs too.

revans2:
I have been adding in spilling for the gather maps which let me push things a bit further and found a bug in the gather implementation.

https://github.com/rapidsai/cudf/issues/8121

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -79,12 +81,669 @@ object GpuHashJoin {
...
+              withResource(new Table(bound)) { boundTab =>
+                sumTable.lowerBound(boundTab, OrderByArg.asc(0))
+              }
+            }
+          }
+        }
+        withResource(cutoff) { cutoff =>
+          withResource(cutoff.copyToHost()) { hostCutoff =>
+            hostCutoff.getInt(0)
+            hostCutoff.getInt(0)
```
abellina, line 219: It seems a little odd to have two `getInt(0)`'s here.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
abellina: +1

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -79,12 +81,669 @@ object GpuHashJoin {
...
+   */
+  def realCheapPerRowSizeEstimate: Double
+
+  /**
+   * Get the bit count size map for the next n rows to be gathered.
+   */
+  def getBitSizeMap(n: Int): ColumnView
+
+  /**
+   * If the data is all fixed width return the size of each row, otherwise return null.
```
abellina, line 184:
```suggestion
   * If the data is all fixed width return the size of each row, otherwise return None.
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -79,12 +81,669 @@ object GpuHashJoin {
...
+    gatheredUpTo >= totalRows
+  }
+
+  override def numRowsLeft: Long = totalRows - gatheredUpTo
+
+  override def allowSpilling(): Unit = {
+    data.allowSpilling()
+  }
+
+  override def getBitSizeMap(n: Int): ColumnView = synchronized {
```
abellina, line 384: Feels like this function could use a function header comment. As I understand it, it computes bit size per row of a cudf Table and returns a single column of longs with those values.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -79,12 +81,669 @@ object GpuHashJoin {
...
+    val inputBitCounts = withResource(GpuColumnVector.from(cb)) { table =>
+      withResource(table.rowBitCount()) { bits =>
+        bits.castTo(DType.INT64)
+      }
+    }
+    // Gather the bit counts so we know what the output table will look like
+    val gatheredBitCount = withResource(inputBitCounts) { inputBitCounts =>
+      withResource(gatherMap.toColumnView(gatheredUpTo, n)) { gatherView =>
+        // Gather only works on a table so wrap the single column
+        val gatheredTab = withResource(new Table(inputBitCounts)) { table =>
```
abellina, line 395: nit: `getheredTable`

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -79,12 +81,669 @@ object GpuHashJoin {
...
+    left.allowSpilling()
+    right.allowSpilling()
+  }
+
+  override def realCheapPerRowSizeEstimate: Double =
+    left.realCheapPerRowSizeEstimate + right.realCheapPerRowSizeEstimate
+
+  override def getBitSizeMap(n: Int): ColumnView = {
+    (left.getFixedWidthBitSize, right.getFixedWidthBitSize) match {
+      case (Some(l), Some(r)) => // This should never happen, but just in case
```
abellina, line 464: should we blow up in this case if it never should happen?

abellina: also, why shouldn't it happen?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -145,249 +804,165 @@ trait GpuHashJoin extends GpuExec {
...
+      case x =>
+        throw new IllegalArgumentException(s"GpuHashJoin should not take $x as the JoinType")
+    }
+
+    val intermediateOutput = leftData ++ rightData
+
+    val finalProject: Seq[Expression] = joinType match {
+      case _: InnerLike | LeftOuter | RightOuter | FullOuter =>
+        remappedLeftOutput ++ remappedRightOutput
+//      case j: ExistenceJoin =>
```
abellina, line 878: Forgot to remove? Or need follow on work?

abellina (changes requested):
First pass @revans2

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -79,12 +81,669 @@ object GpuHashJoin {
...
+        nextCb = nextCbFromGatherer()
+      } else {
+        mayContinue = false
+      }
+      totalTime += (System.nanoTime() - startTime)
+      initialJoin = false
+    }
+    if (nextCb.isEmpty) {
+      // Nothing is left to return so close ASAP.
+      close()
```
jlowe, line 729: Is there a risk of a double-close error on a cudf column vector if we close here and the caller separately closes after this returns false?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -145,249 +804,165 @@ trait GpuHashJoin extends GpuExec {
...
-    new ColumnarBatch(c.asInstanceOf[Array[ColumnVector]], left.numRows())
+  protected lazy val (leftData, rightData, intermediateOutput, finalProject) = {
+    require(leftKeys.map(_.dataType) == rightKeys.map(_.dataType),
+      "Join keys from two sides should have same types")
+    val (leftData, remappedLeftOutput, rightData, remappedRightOutput) = joinType match {
+      case FullOuter | RightOuter | LeftOuter =>
+        // We cannot dedupe anything here because the we can get nulls in the key columns
+        // at least one side
+        (left.output, left.output, right.output, right.output)
+      case _: InnerLike | LeftSemi | LeftAnti =>
```
jlowe, line 866: Unlike inner join, semi and anti joins don't need anything from the right table except the keys.  I see there's special handling of these joins just below, but that's only for the final project.  It looks like we'll still cart around the right table data for these joins.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -79,12 +81,669 @@ object GpuHashJoin {
...
+    case _ => None
+  }
+}
+
+/**
+ * JoinGatherer for a single map/table
+ */
+class JoinGathererImpl(
+    // TODO need a way to spill/cache the GatherMap
+    private val gatherMap: GatherMap,
```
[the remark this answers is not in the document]
revans2: Technically if you don't make it a val, except for a case class, you are accessing a closure value instead of an actual val. They might be implemented the same way, but I wanted to be explicit.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
revans2: Yes, but double frees are OK, although I will protect against it anyway to future proof it.

abellina (changes requested):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
revans2: ctrl-d in InteliJ duplicates a line. I hit it on accident some times and do not notice, just wonder why my copy didn't work.

revans2:
I think I have addressed all of the review comments.  I put together my own quick hack for https://github.com/rapidsai/cudf/issues/8121 and it is not enough to be able to run q72.  Even at a batch size of 26m and 200 partitions it took over 6 mins to finish one of the join tasks our of 200 and failed on the next one.  We are going to have to really think about what we want to try and do to support query 72. But all of the others run with reasonable configurations.

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuBroadcastHashJoinExec.scala
```diff
@@ -141,28 +143,18 @@ case class GpuBroadcastHashJoinExec(
...
-    }
-
     val rdd = streamedPlan.executeColumnar()
-    rdd.mapPartitions(it =>
-      doJoin(builtTable, it, boundCondition, numOutputRows, joinOutputRows,
-        numOutputBatches, streamTime, joinTime, filterTime, totalTime))
+    rdd.mapPartitions { it =>
+      val builtBatch = broadcastRelation.value.batch
+      GpuColumnVector.extractBases(builtBatch).foreach(_.noWarnLeakExpected())
+      doJoin(builtBatch, it, targetSize, spillCallback,
```
jlowe, line 155: Do we want to cache `targetSize` in a local `val` as is done for other parameters just above?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinBase.scala
```diff
@@ -68,37 +69,25 @@ abstract class GpuShuffledHashJoinBase(
     val joinTime = gpuLongMetric(JOIN_TIME)
     val filterTime = gpuLongMetric(FILTER_TIME)
     val joinOutputRows = gpuLongMetric(JOIN_OUTPUT_ROWS)
-
-    val boundCondition = condition.map(GpuBindReferences.bindReference(_, output))
+    val spillCallback = GpuMetric.makeSpillCallback(allMetrics)
+    val localBuildOutput: Seq[Attribute] = buildPlan.output
```
jlowe, line 73: Similar comment re: caching the `targetSize` in a local to help avoid serializing too much

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -79,12 +81,273 @@ object GpuHashJoin {
...
+    exprs.map {
+      case a: AttributeReference => Some(a.toAttribute)
+      case GpuAlias(a: AttributeReference, _) if includeAlias => Some(a.toAttribute)
+      case _ => None
+    }
+
+  /**
+   * Filter rows from the batch where all of the keys are null.
+   */
+  def filterNulls(cb: ColumnarBatch, boundKeys: Seq[Expression]): ColumnarBatch = {
```
abellina, line 96: Should we move here the comment we had before? Sorry if I've missed it somewhere else:

> // This is a work around added in response to https://github.com/NVIDIA/spark-rapids/issues/1643.
  // to deal with slowness arising from many nulls in the build-side of the join. The work around
  // should be removed when https://github.com/rapidsai/cudf/issues/7300 is addressed.

revans2: No because it is not just for that issue any longer.  The Struct join support has extended it.

revans2:
build

revans2:
I upmerged and had to update the code for the new struct join support.  A good thing too because it exposed a bug in my filtering code.  It would only have been a performance regression before the struct code, but afterwards it became an error.  This should be all ready to go now. The dependency is merged.

shims/spark300/src/main/scala/com/nvidia/spark/rapids/shims/spark300/GpuBroadcastHashJoinExec.scala
revans2: done

jlowe (approved):

abellina (approved):

PR #2311, opened 2021-04-30
Update Spark Operator getting started guide.

viadea:

### DIFF
--- a/docs/get-started/getting-started-kubernetes.md
+++ b/docs/get-started/getting-started-kubernetes.md
@@ -263,5 +263,121 @@ sqlContext.sql("SELECT * FROM df WHERE value<>1").show()
 sc.stop()
 ```
 
+## Running Spark Applications using Spark Operator
+
+Using Spark Operator is another way to submit Spark Applications into a Kubernetes Cluster.
+
+1. Copy Spark Application jars/files into the docker image when preparing docker image.
+
+For example, we copy above `test.py` in directory `/opt/sparkRapidsPlugin` inside the docker image.
+
+2. Create spark operator using `helm`.
+
+Follow [Spark Operator quick start guide](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/docs/quick-start-guide.md)
+
+3. Create a Spark Application YAML file
+
+For example, below file named `testpython-rapids.yaml` is created:
+
+```
+apiVersion: "sparkoperator.k8s.io/v1beta2"
+kind: SparkApplication
+metadata:
+  name: testpython-rapids
+  namespace: default
+spec:
+  sparkConf:
+    "spark.ui.port": "4045"
+    "spark.rapids.sql.concurrentGpuTasks": "1"
+    "spark.executor.resource.gpu.amount": "1"
+    "spark.task.resource.gpu.amount": "1"
+    "spark.executor.memory": "1g"
+    "spark.rapids.memory.pinnedPool.size": "2g"
+    "spark.executor.memoryOverhead": "3g"
+    "spark.locality.wait": "0s"
+    "spark.sql.files.maxPartitionBytes": "512m"
+    "spark.sql.shuffle.partitions": "10"
+    "spark.plugins": "com.nvidia.spark.SQLPlugin"
+    "spark.executor.resource.gpu.discoveryScript": "/opt/sparkRapidsPlugin/getGpusResources.sh"
+    "spark.executor.resource.gpu.vendor": "nvidia.com"
+    "spark.executor.extraClassPath": "/opt/sparkRapidsPlugin/rapids-4-spark.jar:/opt/sparkRapidsPlugin/cudf.jar"
+    "spark.driver.extraClassPath": "/opt/sparkRapidsPlugin/rapids-4-spark.jar:/opt/sparkRapidsPlugin/cudf.jar"
+  type: Python
+  pythonVersion: 3
+  mode: cluster
+  image: "<IMAGE_NAME>"
+  imagePullPolicy: Always
+  mainApplicationFile: "local:///opt/sparkRapidsPlugin/test.py"
+  sparkVersion: "3.1.1"
+  restartPolicy:
+    type: Never
+  volumes:
+    - name: "test-volume"
+      hostPath:
+        path: "/tmp"
+        type: Directory
+  driver:
+    cores: 1
+    coreLimit: "1200m"
+    memory: "1024m"
+    labels:
+      version: 3.1.1
+    serviceAccount: spark
+    volumeMounts:
+      - name: "test-volume"
+        mountPath: "/tmp"
+  executor:
+    cores: 1
+    instances: 1
+    memory: "5000m"
+    gpu:
+      name: "nvidia.com/gpu"
+      quantity: 1
+    labels:
+      version: 3.1.1
+    volumeMounts:
+      - name: "test-volume"
+        mountPath: "/tmp"
+```
+
+4. Submit the Spark Application
+
+```
+sparkctl create testpython-rapids.yaml
+```
+
+Note: sparkctl can be built from the Spark Operator repo after [installing golang](https://golang.org/doc/install):
+```
+cd sparkctl
+go build -o sparkctl
+```
+
+5. Check the driver log
+
+```
+sparkctl log testpython-rapids
+```
+
+6. Check the status of this Spark Application
+
+```
+sparkctl status testpython-rapids
+```
+
+7. Port forwarding when Spark driver is running
+
+```
+sparkctl forward testpython-rapids --local-port 1234 --remote-port 4045
+```
+
+Then open browser with `http://localhost:1234/` to check Spark UI.
+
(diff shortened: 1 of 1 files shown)


docs/get-started/getting-started-kubernetes.md
```diff
@@ -263,5 +263,121 @@ sqlContext.sql("SELECT * FROM df WHERE value<>1").show()
 sc.stop()
 ```
 
+## Running Spark Applications using Spark Operator
+
+Using Spark Operator is another way to submit Spark Applications into a Kubernetes Cluster.
+
+1. Copy Spark Application jars/files into the docker image when preparing docker image.
```
tgravescs, line 270: so is there any difference between this and the regular instructions for k8s?  either way we make the jars available in the docker image and set the configs appropriately.  Is the docker image different or can you use the same one?

docs/get-started/getting-started-kubernetes.md
```diff
@@ -263,5 +263,121 @@ sqlContext.sql("SELECT * FROM df WHERE value<>1").show()
...
+
+For example, we copy above `test.py` in directory `/opt/sparkRapidsPlugin` inside the docker image.
+
+2. Create spark operator using `helm`.
+
+Follow [Spark Operator quick start guide](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/docs/quick-start-guide.md)
+
+3. Create a Spark Application YAML file
+
+For example, below file named `testpython-rapids.yaml` is created:
```
jlowe, line 280:
```suggestion
For example, create a file named `testpython-rapids.yaml` with the following contents:
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -263,5 +263,121 @@ sqlContext.sql("SELECT * FROM df WHERE value<>1").show()
...
+kind: SparkApplication
+metadata:
+  name: testpython-rapids
+  namespace: default
+spec:
+  sparkConf:
+    "spark.ui.port": "4045"
+    "spark.rapids.sql.concurrentGpuTasks": "1"
+    "spark.executor.resource.gpu.amount": "1"
+    "spark.task.resource.gpu.amount": "1"
```
jlowe, line 293: I realize this is done elsewhere in this file and not just here, but do we really want all these examples to setup an executor that can only run 1 task concurrently?

docs/get-started/getting-started-kubernetes.md
```diff
@@ -263,5 +263,121 @@ sqlContext.sql("SELECT * FROM df WHERE value<>1").show()
...
+        mountPath: "/tmp"
+```
+
+4. Submit the Spark Application
+
+```
+sparkctl create testpython-rapids.yaml
+```
+
+Note: sparkctl can be built from the Spark Operator repo after [installing golang](https://golang.org/doc/install):
```
jlowe, line 349:
```suggestion
Note: `sparkctl` can be built from the Spark Operator repo after [installing golang](https://golang.org/doc/install):
```

docs/get-started/getting-started-kubernetes.md
```diff
@@ -263,5 +263,121 @@ sqlContext.sql("SELECT * FROM df WHERE value<>1").show()
...
+
+Then open browser with `http://localhost:1234/` to check Spark UI.
+
+8. Delete the Spark Application
+
+```
+sparkctl delete testpython-rapids
+```
+
+
```
jlowe, line 381:
```suggestion
```

docs/get-started/getting-started-kubernetes.md
viadea: Correct. as per the Spark Operator user guide : "A SparkApplication should set .spec.deployMode to cluster, as client is not currently implemented. The driver pod will then run spark-submit in client mode internally to run the driver program. "

docs/get-started/getting-started-kubernetes.md
viadea: I checked all previous examples and all are using 1 task/executor. Since this is for examples, I am fine. 
But if we have this concern, we can point to them to our tuning guide saying something like "For more parameter tuning related topic, refer to this tuning guide." How do you think?

docs/get-started/getting-started-kubernetes.md
viadea: @tgravescs  i just added a new sentence
```
This is because currently only cluster deployment mode is supported by Spark Operator.
```

docs/get-started/getting-started-kubernetes.md
jlowe: It doesn't have to be addressed in this PR, I'm more wondering in general if it's a good idea for these examples to have such a suboptimal setup without calling it out as such.

viadea: I recalled one reason why we put minimum resource here is sometimes people are testing a single node K8s cluster with 1 GPU inside. This can at least run through without much troubleshooting done to figure out GPU resource not enough issue.

tgravescs: 1 is the default so we could just leave it off.

jlowe: but I think that's even worse, right?  We don't want only one task running per executor, we would normally want users to set this to a fraction of a GPU.  I'm fine with this going in as-is since it's consistent with the other examples, but I think there should be a followup to address the very suboptimal config especially if the query doesn't run fully on the GPU.

tgravescs: Why is that worse?  Our defaults should be something we think are a good starting point, otherwise why is it the default? If you are saying that we should put it in at some other value for most people to start with, we should just change the default and not specify the config explicitly.

tgravescs: Oh sorry, I think I was talking bout different configs, I you meant the spark.task.resource.gpu.amount, then agree, setting this lower in example is good.  I was referring to spark.rapids.sql.concurrentGpuTasks

jlowe: I mean it's worse in the sense that it's hidden.  The default makes sense from a safety, it-will-work-if-anything-works perspective, but it's going to be very inefficient for many real-world queries.  My concern is users will set this up per the instructions, get something to run, see the performance is terrible, and stop there.

Like I said above, I'm totally cool with leaving it in (or removing it since it's the default), but we should mention in htis document that the setup is very much untuned/suboptimal and point them to the tuning guide.  The tuning guide needs to discuss setting this config, as it currently doesn't mention it at all.

docs/get-started/getting-started-kubernetes.md
tgravescs: I'm not sure that addresses my concern, I was thinking more along the lines of the docker image. Above we state exactly how to create it and here we just say copy it into the docker image. If the docker image is created the exact same way, perhaps we just reference that section.

My other thought was actually do we even want this here at all?  It seems like a standard spark operator way of doing things so if there is nothing special do we want to maintain our own instructions?

docs/get-started/getting-started-kubernetes.md
jlowe: > you meant the spark.task.resource.gpu.amount

Yes, that's what I'm worried about.  Running 1 task per executor and leaving a ton of CPU cores completely unused is not going to be stellar performance for legacy shuffle or other parts of the query that are not GPU-bound.

I'll file a followup issue for addressing this, as the tuning guide really needs to address this and that's outside the scope of this PR.

jlowe: Filed #2328 to track updating the tuning guide and mentioned updating the examples in other docs, so we don't need to deal with that particular issue in this PR.

docs/get-started/getting-started-kubernetes.md
viadea: I see. Actually if we modify the initial dockerfile to put the test.py inside, then yes, it is the same docker image as previous examples, no difference. 
Since users may not need that test.py for their real production image, so we did not COPY it in the previous example.

IMO, the extra steps for "Spark Operator"'s most important goal is to provide an YAML file example. 
Other that, I did not put detailed steps on how to install "Spark Operator" here, and just point it to the external links.

tgravescs: ok so I think we should clarify the docker image part. The docker has preparing the docker image at the top so I take that applies to all sections, but here we say again copy the jars to it, but we have already done that. 

Perhaps we need to split this into separate sub-sections. Like submitting with spark cli, submitting with spark operator and those are sections after the docker image creation, which is prerequisite.

viadea: I just added an line when preparing the docker image saying that 
```
You can modify the Dockerfile to copy your application into the docker image, i.e. test.py.
```

Then now I changed it to "locate" the application file/jar instead of COPY.
Here we assume the test.py is the user's application file which is already copied into the docker image.

tgravescs (approved):

jlowe (approved):

jlowe:
build

Issue #1902, opened 2021-03-09
[FEA] support GpuLiteral of ArrayType


sperlingxx:
**Is your feature request related to a problem? Please describe.**
Currently, `GpuLiteral` doesn't support nested types (Array, Struct, Map). Because underlying cuDF.Scalar doesn't support nested rapids types. Meanwhile, GpuLiteral of ArrayType is the essential feature for `GpuExplode` and other GpuExpressions. For now, we have a substituted implementation, which provide partial support of  GpuLiteral on ArrayData. We should replace the temporary solution once cuDF support Scalar on ListType.

revans2:
https://github.com/rapidsai/cudf/pull/7584 is adding in support for list scalars, and it looks like it will support full nesting of types because it takes a column_view as the data passed in.  We will need to make java APIs for it and also verify that it will work to create a ColumnVector from a scalar value.

PR #2313, opened 2021-04-30
Supports `GpuLiteral` of array type

firestarman:
This PR is to support `GpuLiteral ` of array type, including nestesd array and struct types.

Also update the `GpuInSet ` to accept `Seq[Any]` instead of `Seq[Literal]` to avoid the uncessary `Literal ` creation, since `GpuInSet `only needs the literal values.

closes https://github.com/NVIDIA/spark-rapids/issues/1902


firestarman:
Marked as draft, because integration is running ...

firestarman:
local it passed, verifying on DB ...

firestarman:
build

firestarman:
Depends on https://github.com/rapidsai/cudf/pull/8077

firestarman:
CI falied due to the cudf jar not being updated to include the change rapidsai/cudf#8077.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpandExec.scala
```diff
@@ -170,6 +170,10 @@ class GpuExpandIterator(
             }
             (cv, false)
           case cv: GpuColumnVector => (cv, false)
+          case scalar: Scalar =>
+            withResource(scalar) { s =>
```
jlowe, line 174: If the scalar is null, do we want to reuse the null column as if it were the `null` case above?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
```diff
@@ -50,6 +50,10 @@ object GpuProjectExec extends Arm {
         val result = expr.columnarEval(cb)
         result match {
           case cv: ColumnVector => cv
+          case scalar: Scalar =>
```
jlowe, line 53: The same code to handle the `expr.columnarEval` result keeps popping up, making me think it would be worth factoring out this boilerplate code into something in `GpuExpressionUtils` that can be easily reused.  This can be a followup issue rather than tackled here if desired.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
```diff
@@ -79,7 +91,132 @@ object GpuScalar {
...
+  }
+
+  /** Casts each element in `data` array to type T */
+  private def castTo[T](data: Array[Any]): Seq[T] = data.map(_.asInstanceOf[T])
+
+  /** Casts each element in `data` array to type T, and applies 'f' on the casted element */
+  private def castTo[T, R](data: Array[Any])(f: T => R): Seq[R] =
+    data.map(e => f(e.asInstanceOf[T]))
+
+  /** Applies 'f' on each element in `data` array, and casts the result to type T*/
```
jlowe, line 114:
```suggestion
  /** Applies 'f' on each element in `data` array, and casts the result to type T */
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
```diff
@@ -79,7 +91,132 @@ object GpuScalar {
...
+      case LongType => ColumnVector.fromBoxedLongs(castTo[JLong](array): _*)
+      case ShortType => ColumnVector.fromBoxedShorts(castTo[JShort](array): _*)
+      case FloatType => ColumnVector.fromBoxedFloats(castTo[JFloat](array): _*)
+      case DoubleType => ColumnVector.fromBoxedDoubles(castTo[JDouble](array): _*)
+      case IntegerType => ColumnVector.fromBoxedInts(castTo[Integer](array): _*)
+      case BooleanType => ColumnVector.fromBoxedBooleans(castTo[JBoolean](array): _*)
+      case TimestampType =>
+        ColumnVector.timestampMicroSecondsFromBoxedLongs(castTo[JLong](array): _*)
+      case StringType =>
+        ColumnVector.fromStrings(castTo[UTF8String, String](array)(convertUTF8StringTo(_)): _*)
```
jlowe, line 190: This copies the string _three times_: once when converting from UTF8String to String, another when getting the bytes out of the string, and yet another when copying the bytes to the device.  It would be nice to completely avoid Java's `String` type when dealing with Spark's `StringType` since `String` is so expensive to wield due to the required copying.  Using `HostColumnVector.Builder`'s `appendUTF8String` would be more efficient here.  Would be nice if `ColumnVector` had a `fromUTF8Strings` that took a vararg of `byte[]` so we could avoid the `HostColumnVector.Builder` direct usage.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
```diff
@@ -79,7 +91,132 @@ object GpuScalar {
...
+
+    withResource(listView) { list =>
+      Scalar.listFromColumnView(list);
+    }
+  }
+
+  @deprecated("This will be removed. Since no need to infer the type again because" +
+    " Spark does it already. Besides it is difficult to support nested type.", "")
   def from(v: Any): Scalar = v match {
+    case s: Scalar => s.incRefCount()
```
jlowe, line 219: If `GpuScalar.from` is going to handle `Scalar` directly, is it still required to explicitly handle `Scalar` after an `expr.columnarEval`?  Seems like we can let the existing `case other => GpuScalar.from...` code expand the scalar to a column.

firestarman:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
```diff
@@ -481,6 +481,20 @@ public static DType getNonNestedRapidsType(DataType type) {
     return result;
   }
 
+  /**
+   * Converts the Spark DataType to Rapids DType. It will cover the nested types, which are
+   * being supported more and more.
+   */
+  public static DType sparkDataTypeToRapidsType(DataType type) {
```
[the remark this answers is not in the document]
firestarman: Thanks for review, updated to `toRapidsType`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpandExec.scala
firestarman: Updated, if the scalar is null, it will go into the last case to create a `ColumnVector ` with a null scalar, the same as the before.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
firestarman: Good suggestion, the followup issue is here https://github.com/NVIDIA/spark-rapids/issues/2327

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
firestarman: Nice catch, updated

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
firestarman: Good to know, updated to use `ColumnVector.build` now.
The followup issue is here https://github.com/rapidsai/cudf/issues/8137

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
firestarman: This is to cover the case that some nodes will close a closeable object returned from `expr.columnarEval`. e.g. `GpuBinaryExpression` and `GpuUnaryExpression` .

With this, no need to add the `scalar ` case specially and the `scalar` is treated as a closeable object.

firestarman: We may want to unify the result handling of `expr.columnarEval` for these two kinds of cases.

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
firestarman: This can also be trakced by the issue #2327

revans2 (changes requested):
I started to review this, but stopped early because there are a lot of things happening here and bugs that we need to address.

1. I cannot see how the cudf JNI code for creating a null list scalar will produce the correct thing.  It produces a `list_scalar` with they type `LIST`, but nothing after that. If I create a `ColumnVector` from that scalar does it have a child data column with a type associated with it? From my understanding if it does not then it is invalid. We at least need an empty child data column, and if that is a list or a struct we need the full set of empty child data columns to fully express the type.
2. I see the desire to move away from returning a Scala literal value to return an actual `Scalar`. I am fine with this, but I don't want it to be half done.  I don't want to support both Scalar and Scala Literal values being returned. If we are going to go with Scalar, because it is potentially a large change, can we please do one PR first to refactor the code completely. Then we can add LIST support on top of that.
3. Similarly to be able to properly support nested literal values I don't think we can continue to use some of the existing cudf APIs. I really would like to see a separate PR that has us stop using any of the Scalar factory APIs outside of GpuScalar itself.

Also now that I have had time to look at the JNI work for list scalar there are a number of changes I would like to see made to it. I am really sorry that I didn't get around to reviewing it early on.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpandExec.scala
```diff
@@ -170,6 +170,10 @@ class GpuExpandIterator(
             }
```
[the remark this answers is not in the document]
jlowe: Filed #2332

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
```diff
@@ -79,8 +91,145 @@ object GpuScalar {
...
+        ColumnVector.timestampMicroSecondsFromBoxedLongs(castTo[JLong](array): _*)
+      case StringType =>
+        // To be updated. https://github.com/rapidsai/cudf/issues/8137
+        val rows = castTo[UTF8String, Array[Byte]](array)(convertUTF8StringTo(_))
+        ColumnVector.build(DType.STRING, rows.length, b => {
+          rows.foreach {
+            case null => b.appendNull()
+            case s => b.appendUTF8String(s.asInstanceOf[Array[Byte]])
+          }
+        })
```
jlowe, line 202: Nit: If we're going to go through array elements one-by-one, to call `appendUTF8String`, why not skip the double-pass and extra array obj creation in `rows` and instead just walk `array `directly calling `appendUTF8String(e.getBytes)` for each element that is non-null.

firestarman: Good suggestion, will update later

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpandExec.scala
firestarman: The case I met is at line 90 in file "nullExpressions.scala ", where it is trying to extract the value from a cudf Scalar and return. But pulling out list data is difficult, so I change to return a cudf Scalar directly as below.
```diff   
-       GpuScalar.extract(runningScalar)
+      // Returns the Scalar directly instead of pulling data out of GPU.
+      runningScalar.incRefCount()
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
```diff
@@ -98,12 +247,18 @@ object GpuScalar {
...
+      // is needed since Spark has already inferred the DataType for the literal value.
+      throw new IllegalStateException("Please calls 'from(Any, DataType)' instead to" +
+        " create a Scalar of array type.")
     case _ =>
       throw new IllegalStateException(s"${v.getClass} '$v' is not supported as a scalar yet")
   }
 
   def from(v: Any, t: DataType): Scalar = v match {
-    case _ if v == null => Scalar.fromNull(GpuColumnVector.getNonNestedRapidsType(t))
+    case _ if v == null => Scalar.fromNull(GpuColumnVector.toRapidsType(t))
```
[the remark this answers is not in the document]
firestarman: So I got a question.
How to define a null LIST Scalar ?  Besides setting the validity flag to `false`, we still need to create an empty column with the correct children for it, instead of the default column ?

firestarman:
> I started to review this, but stopped early because there are a lot of things happening here and bugs that we need to address.
> 
> 1. I cannot see how the cudf JNI code for creating a null list scalar will produce the correct thing.  It produces a `list_scalar` with they type `LIST`, but nothing after that. ....e.
> 2. I see the desire to move away from returning a Scala literal value to return an actual `Scalar`. ...
> 3. Similarly ...
> 
> ...

Thanks a lot for this review.
Before I continue the work on this PR, we may want to align with some coming changes.

1.  I was wrong to deal with the null `LIST  Scalar`. So we still need to create an empty column with the correct children for a null `LIST  Scalar` , instead of the default column created by cpp, right ? If yes, a special API is probably needed for creating a null `LIST Scalar`. Such as `Scalar listFromNull(HostColumnVector.DataType dt)`.

2.  The change is becoming more complicated than I thought since it needs some refactor work going first. To support the `LIST Scalar`, I plan to
    - Let `GpuLiteral ` return a cudf `Scalar ` instead of the current Scala value `Any` from its `columnarEval`, by calling the "GpuScalar.from" factory APIs to convert a Scala value to a `Scalar`.  This is different from the CPU `Literal `, who is returning the Scala value directly.

3. Is there any limitation or concern or design about the result from the `columnarEval` for all the GPU expressions ? According to the change above, we should have only two kinds of objects, `GpuColumnVector ` and cudf `Scalar `. Is it reasonable ?

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
```diff
@@ -481,6 +481,20 @@ public static DType getNonNestedRapidsType(DataType type) {
     return result;
   }
 
+  /**
+   * Converts the Spark DataType to Rapids DType. It will cover the nested types, which are
+   * being supported more and more.
+   */
+  public static DType toRapidsType(DataType type) {
```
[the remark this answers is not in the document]
firestarman: Will remove this.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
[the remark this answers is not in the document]
revans2: Yes, this is why the c++ code does not let you make an empty nested column with just the data_type.

https://github.com/rapidsai/cudf/blob/2ead87c7d0245e5fef08c6fde44c8f7d978874ac/cpp/src/column/column_factories.cpp#L67-L68

But instead wrote `cudf::empty_like` so you could make an empty column that looks like an existing one.

https://github.com/rapidsai/cudf/blob/2ead87c7d0245e5fef08c6fde44c8f7d978874ac/cpp/include/cudf/copying.hpp#L177-L183

And also why we check that when we get a column that all of the children columns exist and are the proper types.

https://github.com/NVIDIA/spark-rapids/blob/c6931355fc32736ba1e7383dbcd0dfee4ca760c1/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java#L616-L622

Otherwise we risk producing an invalid memory layout at some point. The fact that you could make a list scalar with no column view feels really bad to me.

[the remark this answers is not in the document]
firestarman: Filed a PR https://github.com/rapidsai/cudf/pull/8173 to support this.

firestarman:
Depends on https://github.com/rapidsai/cudf/pull/8173, back to draft

firestarman:
The issue #2349 was filed to track the refactor work.

firestarman:
Continued work on this PR since the refactor work has been done.

firestarman:
build

firestarman:
build

tests/src/test/scala/com/nvidia/spark/rapids/ProjectExprSuite.scala
```diff
@@ -54,6 +54,11 @@ class ProjectExprSuite extends SparkQueryCompareTestSuite {
         lit(BigDecimal(-2.12314e-8)),
         lit(Array(1, 2, 3, 4, 5)), lit(Array(1.2, 3.4, 5.6)),
         lit(Array("a", "b", null, "")),
+        lit(Array(Array(1, 2), null, Array(3, 4))),
```
wjxiz1992, line 57: The decription of this pr is only "Array" , I guess you mean array of array, and array of structs.

wjxiz1992 (approved):
LGTM, only description mismatch

tests/src/test/scala/com/nvidia/spark/rapids/ProjectExprSuite.scala
firestarman: I thought array type also covered array of array, and array of structs, is it ?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
```diff
@@ -60,14 +56,132 @@ object GpuScalar extends Arm with Logging {
...
+    // StructData does not support utf8 string yet, so parses it here instead of calling the
+    //  `convertUTF8StringTo`. Tracked by https://github.com/rapidsai/cudf/issues/8137
+    case StringType => element.asInstanceOf[UTF8String].toString
+    case dt: DecimalType => convertDecimalTo(element.asInstanceOf[Decimal], dt) match {
+      case Left(i) => i
+      case Right(l) => l
+    }
+    case ArrayType(eType, _) =>
+      val data = element.asInstanceOf[ArrayData]
+      data.array.map(convertElementTo(_, eType)).toList.asJava
```
wbo4958, line 118: is there a possibility that `data.arrray` equals null?

tests/src/test/scala/com/nvidia/spark/rapids/ProjectExprSuite.scala
wjxiz1992: I mean when I first read the description, my feeling is "before this PR, array literal is not supported and this PR starts to support it."

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
```diff
@@ -329,6 +490,7 @@ object GpuLiteral {
     val cpuLiteral = Literal.default(dataType)
     GpuLiteral(cpuLiteral.value, cpuLiteral.dataType)
   }
+
```
wbo4958, line 493: empty line?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
firestarman: It can not be null according to the Spark code of [ArrayData creation](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/CatalystTypeConverters.scala#L165) .

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
firestarman: yes, easy to distinguish the ends of the function and the `GpuLiteral ` object.

tests/src/test/scala/com/nvidia/spark/rapids/ProjectExprSuite.scala
revans2: I am happy with doing more than is promised, especially if it is simple to add in. Please do not remove array of structs even if it is beyond the requirements.

revans2 (approved):
This looks good to me now. It is a clean small patch that does not break any existing tests. I just had one nit about the TODO that I am happy to see done in a follow on issue if we want to.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuInSet.scala
```diff
@@ -16,16 +16,14 @@
```
jlowe, line 16: 2021 copyrights

integration_tests/src/main/python/conditionals_test.py
```diff
@@ -138,9 +138,7 @@ def test_ifnull(data_gen):
                 'ifnull({}, b)'.format(null_lit),
```
jlowe, line 138: 2021 copyrights

gerashegalov: @firestarman would you mind trying the automated copyright update https://github.com/NVIDIA/spark-rapids/blob/branch-0.6/CONTRIBUTING.md#pre-commit-hooks

tests/src/test/scala/com/nvidia/spark/rapids/ProjectExprSuite.scala
firestarman: Updated

integration_tests/src/main/python/conditionals_test.py
```diff
@@ -138,9 +138,7 @@ def test_ifnull(data_gen):
...
                 'ifnull(a, {})'.format(null_lit)))
 
-# GpuLiteral does not support array type yet, so make the test separate from the above one.
-# Supporting GpuLiteral of array type is tracked by
-#     https://github.com/NVIDIA/spark-rapids/issues/1902
+# TODO Merge this with the test `test_case_when` above.
```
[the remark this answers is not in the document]
firestarman: I would like to do it in another PR, here is the issue https://github.com/NVIDIA/spark-rapids/issues/2445 .

integration_tests/src/main/python/conditionals_test.py
firestarman: Thanks a lot @gerashegalov. Great tool, will try it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuInSet.scala
firestarman: nice catch

integration_tests/src/main/python/conditionals_test.py
firestarman: nice catch. Updated

firestarman:
build

wbo4958 (approved):

firestarman:
Since I have addressed the 3 comments of small issues (two for copy rights and one for a following issue), I am going to merge this to unblock @wbo4958 's work.

PR #2316, opened 2021-04-30
Update doc to note that single quoted json strings are not ok

sameerz:
Update doc to note that `get_json_object` on the GPU only supports double quoted strings in JSON per the http://json.org/ spec.  Spark appears to support single quoted strings in JSON per https://github.com/apache/spark/blob/4e8701a77dff729c4e8e0ad39c16e2717c2c32fe/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/jsonExpressions.scala#L108 .  Single quote support will be addressed in a future release.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -454,5 +454,5 @@ on constant folding and parts of the query will not be accelerated if
 ## JSON string handling
 The 0.5 release introduces the `get_json_object` operator.  The JSON specification only allows
 double quotes around strings, whereas Spark allows single quotes.  The `get_json_object` operation
-on the GPU will return `None` in Pyspark or `Null` in Scala when trying to match a string surrounded
+on the GPU will return `None` in PySpark or `Null` in Scala when trying to match a string surrounded
 by single quotes. 
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -455,4 +455,4 @@ on constant folding and parts of the query will not be accelerated if
 The 0.5 release introduces the `get_json_object` operation.  The JSON specification only allows
 double quotes around strings, whereas Spark allows single quotes.  The `get_json_object` operation
 on the GPU will return `None` in PySpark or `Null` in Scala when trying to match a string surrounded
-by single quotes.
+by single quotes.  The behavior will be updated in a future release to more closely match Spark. 
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -452,7 +452,7 @@ on constant folding and parts of the query will not be accelerated if
 `org.apache.spark.sql.catalyst.optimizer.ConstantFolding` is excluded as a rule.
 
 ## JSON string handling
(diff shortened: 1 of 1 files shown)


docs/compatibility.md
```diff
@@ -450,3 +450,9 @@ ConstantFolding is an operator optimization rule in Catalyst that replaces expre
 be statically evaluated with their equivalent literal values. The RAPIDS Accelerator relies
 on constant folding and parts of the query will not be accelerated if 
 `org.apache.spark.sql.catalyst.optimizer.ConstantFolding` is excluded as a rule.
+
+## JSON string handling
+The 0.5 release introduces the `get_json_object` operator.  The JSON specification only allows
+double quotes around strings, whereas Spark allows single quotes.  The `get_json_object` operation
+on the GPU will return `None` in Pyspark or `Null` in Scala when trying to match a string surrounded
```
jlowe, line 457:
```suggestion
on the GPU will return `None` in PySpark or `Null` in Scala when trying to match a string surrounded
```

jlowe (commented):
Do we need to mention unescaped control characters as well?  I assume the cudf json parser does this differently.  cc: @nvdbaranec

docs/compatibility.md
```diff
@@ -450,3 +450,9 @@ ConstantFolding is an operator optimization rule in Catalyst that replaces expre
 be statically evaluated with their equivalent literal values. The RAPIDS Accelerator relies
 on constant folding and parts of the query will not be accelerated if 
 `org.apache.spark.sql.catalyst.optimizer.ConstantFolding` is excluded as a rule.
+
+## JSON string handling
+The 0.5 release introduces the `get_json_object` operator.  The JSON specification only allows
```
abellina, line 455:
```suggestion
The 0.5 release introduces the `get_json_object` operation.  The JSON specification only allows
```

nvdbaranec:
> 
> 
> Do we need to mention unescaped control characters as well? I assume the cudf json parser does this differently. cc: @nvdbaranec

We don't do any explicit filtering for them at the moment, so they "should" just come through.

nvdbaranec (commented):
Worth mentioning that this behavior will be fixed in the future?

sameerz:
> Worth mentioning that this behavior will be fixed in the future?

Updated to mention this will be addressed in the future.

sameerz:
build

nvdbaranec:
Actually, to add to the confusion a bit.   The JSONPath string (the query itself) allows single quotes.  In fact the spec for it requires them.  It's the input JSON data itself that requires the double quotes.  Might be worth a clarification.

sameerz:
>  The JSONPath string (the query itself) allows single quotes. In fact the spec for it requires them. It's the input JSON data itself that requires the double quotes. Might be worth a clarification.

Updated to note that the double quote requirement is for strings in JSON data.

nvdbaranec (approved):

sameerz:
build

PR #2318, opened 2021-05-01
Fix ColumnarToRowIterator handling of empty batches

### LINKED ISSUE #2317: [BUG] GpuColumnarToRowIterator can stop after receiving an empty batch
**Describe the bug**
When GpuColumnarToRowIterator receives an empty device columnar batch it can return `false` for `hasNext` despite more input batches awaiting to be iterated.

Performing a join where there are input tables that have null join keys, the target batch size is very small (e.g.: 10 bytes), and the shuffle is forced off the GPU.  What should occur is that the query will filter for nulls before the join on the GPU with `GpuFilter` then follow that up with `ColumnarToRowExec` before the CPU shuffle.  With a very small target batch size, batches can consist of a single row, and some of those batches will be filtered to nothing before being passed to `ColumnarToRowExec`.  That exec will terminate the iteration early when it sees the empty batch.

jlowe:
Fixes #2317

This updates `loadNextBatch` to pull from the input batch iterator in a loop as long as there are empty batches being received.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
@@ -161,12 +161,12 @@ class AcceleratedColumnarToRowIterator(
 class ColumnarToRowIterator(batches: Iterator[ColumnarBatch],
     numInputBatches: GpuMetric,
     numOutputRows: GpuMetric,
-    totalTime: GpuMetric) extends Iterator[InternalRow] {
+    totalTime: GpuMetric) extends Iterator[InternalRow] with Arm {
   // GPU batches read in must be closed by the receiver (us)
   @transient var cb: ColumnarBatch = null
   var it: java.util.Iterator[InternalRow] = null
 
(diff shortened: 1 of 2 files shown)


jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
```diff
@@ -177,31 +177,43 @@ class ColumnarToRowIterator(batches: Iterator[ColumnarBatch],
...
+          cb = new ColumnarBatch(GpuColumnVector.extractColumns(devCb).map(_.copyToHost()),
+            devCb.numRows())
+          it = cb.rowIterator()
+          // In order to match the numOutputRows metric in the generated code we update
+          // numOutputRows for each batch. This is less accurate than doing it at output
+          // because it will over count the number of rows output in the case of a limit,
+          // but it is more efficient.
+          numOutputRows += cb.numRows()
+      }
+      return true
```
revans2, line 212: nit: I think it is more scala like to have it be functional.

```
if (devCb.numRows > 0) {
  ...
  true
} else {
  false
}
```

revans2 (approved):

PR #2323, opened 2021-05-01
Update CBO to use fixed cost model and to respect estimated row counts

andygrove:
Changes in this PR:

- Switch from relative cost per operator to absolute cost that takes estimated row count into account
- CPU and GPU have separate cost models (both very simplistic still)
- Added cost for `GetStructField`

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
@@ -243,26 +243,26 @@ class CpuCostModel(conf: RapidsConf) extends CostModel {
   def getCost(plan: SparkPlanMeta[_]): Double = {
     val rowCount = RowCountPlanVisitor.visit(plan).map(_.toDouble)
       .getOrElse(conf.defaultRowCount.toDouble)
-    val costPerRow = plan.conf.getOperatorCost(plan.wrapped.getClass.getSimpleName).getOrElse {
+    plan.conf.getOperatorCost(plan.wrapped.getClass.getSimpleName).getOrElse {
       plan.wrapped match {
         case _: ProjectExec =>
           // the cost of a projection is the sum of its expressions
           plan.childExprs
-            .map(expr => exprCost(expr.asInstanceOf[BaseExprMeta[Expression]]))
+            .map(expr => exprCost(expr.asInstanceOf[BaseExprMeta[Expression]], rowCount))
             .sum
-        case _ => 1.0
+        case _ => 1.0 * rowCount
       }
     }
-    rowCount * costPerRow
   }
 
-  private def exprCost[INPUT <: Expression](expr: BaseExprMeta[INPUT]): Double = {
+  private def exprCost[INPUT <: Expression](expr: BaseExprMeta[INPUT], rowCount: Double): Double = {
 
     val childExprCost = expr.childExprs
-      .map(e => exprCost(e.asInstanceOf[BaseExprMeta[Expression]])).sum
+      .map(e => exprCost(e.asInstanceOf[BaseExprMeta[Expression]], rowCount)).sum
 
     // always check for user overrides first
-    childExprCost + expr.conf.getExpressionCost(expr.getClass.getSimpleName).getOrElse {
+    val totalExprCost = childExprCost + expr.conf
+        .getExpressionCost(expr.getClass.getSimpleName).getOrElse {
       expr match {
         case _ =>
           // many of our BaseExprMeta implementations are anonymous classes so we look directly at
@@ -276,6 +276,7 @@ class CpuCostModel(conf: RapidsConf) extends CostModel {
           }
       }
     }
+    rowCount * totalExprCost
   }
 }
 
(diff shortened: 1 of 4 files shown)


andygrove:
build

andygrove:
@revans2 This PR does not attempt to implement the memory access cost model we discussed but it does now provide separate cost models for CPU and GPU that are based on absolute rather than relative costs and takes row count estimates into account.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -211,49 +234,59 @@ trait CostModel {
...
-    conf.defaultTransitionToCpuCost
+  private def exprCost[INPUT <: Expression](expr: BaseExprMeta[INPUT]): Double = {
+    // always check for user overrides first
+    expr.conf.getExpressionCost(expr.getClass.getSimpleName).getOrElse {
+      expr match {
+        case _ =>
+          // many of our BaseExprMeta implementations are anonymous classes so we look directly at
+          // the wrapped expressions in some cases
+          expr.wrapped match {
+            case _: AttributeReference => 0.01
```
revans2, line 267: nit: instead of a magic number of 0.01 it would be nice to have a common value somewhere. for these

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -279,21 +308,23 @@ class DefaultCostModel(conf: RapidsConf) extends CostModel {
...
           expr.conf.getExpressionCost(s"Cast${cast.fromType}To${cast.toType}")
-              .getOrElse(conf.defaultExpressionCost)
+            .getOrElse(conf.defaultExpressionCost)
         case _ =>
           // many of our BaseExprMeta implementations are anonymous classes so we look directly at
           // the wrapped expressions in some cases
           expr.wrapped match {
-            case _: AttributeReference => 1.0 // no benefit on GPU
-            case Alias(_: AttributeReference, _) => 1.0 // no benefit on GPU
+            case _: AttributeReference => 0.01 // no benefit on GPU
```
revans2, line 316: Actually the cost of an AttributeReference here is probably close to free, because we are just incrementing a reference count, but the cost of a GetStructField needs to copy the data out, so they should be different.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -279,21 +308,23 @@ class DefaultCostModel(conf: RapidsConf) extends CostModel {
           // different CAST operations have different costs, so we allow these to be configured
           // based on the data types involved
```
revans2, line 309: I think it would be a lot more accurate to recursively walk the expression tree and sum up the values instead of trying to match things at the top level.  That way `5 + foo` can have the correct cost compared to `5 + foo.bar`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -264,11 +297,7 @@ class DefaultCostModel(conf: RapidsConf) extends CostModel {
         case _ => conf.defaultOperatorCost
       }
     }
-
-    plan.cpuCost = cpuCost
-    plan.gpuCost = gpuCost
-
-    (cpuCost, gpuCost)
+    rowCount * perRowCost
```
revans2, line 300: This is a good first step but longer term I think we might want to turn this model on its head a bit and pass the row count to the cost estimation for the expression.  On the GPU there is a fixed cost to launching a kernel, so `5 + foo` has some cost that is fixed no matter the number of rows, and a second potion that is relative to the cost of doing the add. If we don't include this in some form then we cannot fall back to the CPU for really small row counts.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -211,89 +234,109 @@ trait CostModel {
...
 
-class DefaultCostModel(conf: RapidsConf) extends CostModel {
+class CpuCostModel(conf: RapidsConf) extends CostModel {
 
-  def transitionToGpuCost(plan: SparkPlanMeta[_]) = {
-    // this is a placeholder for now - we would want to try and calculate the transition cost
-    // based on the data types and size (if known)
-    conf.defaultTransitionToGpuCost
+  private val attrRefCost = conf.getCpuExpressionCost("AttributeReference").getOrElse(0d)
+  private val structFieldCost = conf.getCpuExpressionCost("GetStructField").getOrElse(0.05d)
```
andygrove, line 244: I'm not crazy about these different levels of configs but not sure what the right answer is. I don't think we want to add a specific config for every single operator and expression so having a generic mechanism seems to make sense for now but then we need defaults as well.

andygrove:
@revans2 This is ready for another review now

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -138,29 +145,29 @@ class CostBasedOptimizer extends Optimizer with Logging {
...
             val (childCpuCost, childGpuCost) = childCosts
-            val transitionCost = costModel.transitionToCpuCost(child)
+            val transitionCost = transitionToCpuCost(conf, child)
             val childGpuTotal = childGpuCost + transitionCost
             if (child.canThisBeReplaced && !consumesQueryStage(child)
                 && childGpuTotal > childCpuCost) {
               // force this child plan back onto CPU
               optimizations.append(ReplaceSection(
-                child.asInstanceOf[SparkPlanMeta[SparkPlan]], totalCpuCost, totalGpuCost))
+                child, totalCpuCost, totalGpuCost /*childCpuCost, childGpuTotal*/))
```
revans2, line 154: Do we plan to uncomment these in the future?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
revans2: Having configs for each of these things is really only going to help with testing. No real user will be able to set these in any meaningful way, not should they.

revans2 (approved):
This all looks fine, but I really think we could simplify things by looking at memory access patterns instead, and having a default cost estimation for expressions. We could also allow for overrides for specific expressions that we know we need to tune.

For example a normal expression will read all of its inputs and produce a single output.

```
def singleExprCost(expr:...): Long = {
  expr.conf.getGpuExpressionCost(...).getOrElse {
    val readCost = expr.childExprs.map(child => getGpuOutputSizeEstimate(child, rowCount)).sum * GPU_READ_SPEED
    val writeCost = getGpuOutputSizeEstimate(expr, rowCount) * GPU_WRITE_SPEED
    val kernelLaunchCost = getKernelLaunchEstimate(expr) * KERNEL_LAUNCH_OVERHEAD
    // TODO possibly add in some extra per row computation cost...
    readCost + writeCost + kernelLaunchCost
  }
}
```
The costs here would all be measured in time. 

Inside things like `getKernelLaunchEstimate` we would have Tables that could override the default values for different expressions. For `getGpuOutputSizeEstimate` we would look at things like the data type and for variable width types guess what it would look like.

I would think the CPU cost would look very similar, but with different override tables and different constants.

andygrove:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
andygrove: No. I pushed a change to remove those.

jlowe:
build

andygrove:
@revans2 could you re-approve please - I dismissed your approval when removing the commented out parameters

jlowe (approved):

PR #2337, opened 2021-05-04
Update the databricks shim for struct joins.

revans2:
This updates the databricks shim layer for struct joins. It really was just missing support for nested decimal which was added by the struct join patch.

I also cleaned up an unused import in several shims while looking for it.


revans2:
build

tgravescs (approved):

jlowe (approved):

PR #2338, opened 2021-05-04
Commonize tagging for hash and sort aggregates

jlowe:
Currently hash and sort aggregates have completely separate code for tagging the plan despite both being translated to use `GpuHashAggregateExec` as the backend GPU implementation.

This updates the tagging code to be common and also cleans up the code (using `withResource`, proper indentation in places, etc.).  Happy to separate out some of these changes if desired.

Notable changes include the removal of some tagging restrictions for sort aggregate since they weren't there in hash aggregate.  String literals are no longer restricted and `hasNans` doesn't have to be set to false for aggregations on floating point types.


jlowe:
build

abellina (approved):

PR #2339, opened 2021-05-04
Remove support for Spark 3.0.0

jlowe:
This removes support for Spark 3.0.0, updating the minimum Spark version to Spark 3.0.1.  The spark300 and spark300emr shim projects are removed along with `SparkShims` methods that were only needed to implement compile differences between Spark 3.0.0 and Spark 3.0.1.  Documentation references to Spark 3.0.0 were also updated to use Spark 3.0.1 instead.


jlowe:
@revan2 I noticed the supported_ops doc updated when I copied the `First` and `Last` behavior from spark301 shim into the base shim.  I kept the behavior unchanged, but I'm not sure it's what is intended, so it would be good to get some input there.

@nartal1 I updated the audit code and scripts to reflect the removal of the spark300 shim, if you could give some feedback on those changes that would be appreciated.

tests/src/test/scala/com/nvidia/spark/rapids/AdaptiveQueryExecSuite.scala
```diff
@@ -85,7 +85,7 @@ class AdaptiveQueryExecSuite
 
   private def findTopLevelGpuBroadcastHashJoin(plan: SparkPlan): Seq[GpuExec] = {
     collect(plan) {
-      case j: GpuExec if ShimLoader.getSparkShims.isBroadcastExchangeLike(j) => j
+      case j: GpuExec if j.isInstanceOf[BroadcastExchangeLike] => j
```
revans2, line 88: nit: why not just
```
case j: BroadcastExchangeLike => j
```

revans2 (approved):
Just one small nit that I can ignore.

tests/src/test/scala/com/nvidia/spark/rapids/AdaptiveQueryExecSuite.scala
jlowe: That doesn't compile since it needs to return a sequence of `GpuExec` but `BroadcastExchangeLike` does not implement that trait. I do think I can simplify it a bit by using `with BroadcastExchangeLike` instead of the `isInstanceOf` check.

jlowe:
build

revans2 (approved):

tgravescs (approved):

jlowe:
Spark 3.1.1 tests failed because it was trying to lookup `TimeSub` which was removed in that version.  I refactored the shim expressions handling to explicitly call this out.  In the process I also realized that the `First`and `Last` expressions no longer need to be in the shims after 3.0.0 is removed, so I moved those aggregation expressions back into the common code.

jlowe:
build

revans2 (approved):

tgravescs (approved):

nartal1:
> @nartal1 I updated the audit code and scripts to reflect the removal of the spark300 shim, if you could give some feedback on those changes that would be appreciated.
 The changes for audit scripts LGTM. Thanks @jlowe .

nartal1 (approved):

PR #2340, opened 2021-05-05
Allow partial materialization of broadcast nested loop join and cartesian exec

revans2:
This is the second out of core join PR. This adds in support for doing out of core joins on cross joins (CarteaseanExec) and inner/cross joins for  broadcast nested loop joins.  Because these joins always explode in size we didn't have them enabled by default but with this work we can have them on by default now.  This has one big advantage over the other join implementations, in that I don't have to materialize the gather maps so even very very large joins we can output batches without having very large gather maps.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
@@ -268,13 +268,12 @@ case class GpuCartesianProductExec(
         numOutputRows,
         numOutputBatches)
     } else {
-      val targetSize = RapidsConf.GPU_BATCH_SIZE_BYTES.get(conf)
       val spillCallback = GpuMetric.makeSpillCallback(allMetrics)
 
       new GpuCartesianRDD(sparkContext,
         boundCondition,
         spillCallback,
-        targetSize,
+        targetSizeBytes,
         joinTime,
         joinOutputRows,
         numOutputRows,
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
@@ -20,8 +20,9 @@ import java.io.{IOException, ObjectInputStream, ObjectOutputStream}
 
 import scala.collection.mutable
 
-import ai.rapids.cudf.{JCudfSerialization, NvtxColor, NvtxRange, Table}
-import com.nvidia.spark.rapids.{Arm, GpuBindReferences, GpuBuildLeft, GpuColumnVector, GpuExec, GpuExpression, GpuMetric, GpuSemaphore, MetricsLevel, RapidsBuffer, SpillableColumnarBatch, SpillPriorities}
+import ai.rapids.cudf.{JCudfSerialization, NvtxColor, NvtxRange}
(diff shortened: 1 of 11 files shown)


revans2:
build

revans2:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExec.scala
```diff
@@ -264,51 +395,27 @@ abstract class GpuBroadcastNestedLoopJoinExecBase(
...
         numOutputBatches)
-    } else if (broadcast.output.isEmpty) {
-      assert(boundCondition.isEmpty)
-
-      lazy val buildCount: Int = computeBuildRowCount(broadcastRelation, buildTime, buildDataSize)
-
+    } else {
+      lazy val builtBatch: ColumnarBatch =
+        makeBuiltBatch(broadcastRelation, buildTime, buildDataSize)
+      val targetSize = RapidsConf.GPU_BATCH_SIZE_BYTES.get(conf)
```
jlowe, line 406: Why not use `targetSizeBytes` here rather than re-parsing the conf setting?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
```diff
@@ -327,15 +262,19 @@ case class GpuCartesianProductExec(
...
       //  that happens with the built in cartesian, but writing another custom RDD
       //  just for this use case is not worth it without an explicit use case.
-      GpuNoColumnCrossJoin.divideIntoBatches(
+      GpuBroadcastNestedLoopJoinExecBase.divideIntoBatches(
         l.cartesian(r).map(p => p._1 * p._2),
         targetSizeBytes,
         numOutputRows,
         numOutputBatches)
     } else {
+      val targetSize = RapidsConf.GPU_BATCH_SIZE_BYTES.get(conf)
```
jlowe, line 271: Why not use `targetSizeBytes` here?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
```diff
@@ -20,8 +20,9 @@ import java.io.{IOException, ObjectInputStream, ObjectOutputStream}
 
 import scala.collection.mutable
 
-import ai.rapids.cudf.{JCudfSerialization, NvtxColor, NvtxRange, Table}
-import com.nvidia.spark.rapids.{Arm, GpuBindReferences, GpuBuildLeft, GpuColumnVector, GpuExec, GpuExpression, GpuMetric, GpuSemaphore, MetricsLevel, RapidsBuffer, SpillableColumnarBatch, SpillPriorities}
+import ai.rapids.cudf.{JCudfSerialization, NvtxColor, NvtxRange}
+import com.nvidia.spark.rapids.{Arm, GpuBindReferences, GpuBuildLeft, GpuColumnVector, GpuExec, GpuMetric, GpuSemaphore, LazySpillableColumnarBatch, MetricsLevel, RapidsConf}
```
jlowe, line 24:
```suggestion
import com.nvidia.spark.rapids.{Arm, GpuBindReferences, GpuBuildLeft, GpuColumnVector, GpuExec, GpuMetric, GpuSemaphore, LazySpillableColumnarBatch, MetricsLevel}
```

jlowe (approved):

jlowe:
build

PR #2343, opened 2021-05-05
Add more description to Decimal support.

viadea:


docs/supported_ops.md
```diff
@@ -21,6 +21,29 @@ decimals are disabled by default in the plugin, because it is supported by a sma
 number of operations presently, which can result in a lot of data movement to and
 from the GPU, slowing down processing in some cases.
 
+Result `Decimal` precision and scale follow the same rule as CPU mode in Apache Spark:
```
revans2, line 24: This file is auto-generated so you should not edit it directly.

docs/supported_ops.md
```diff
@@ -21,6 +21,29 @@ decimals are disabled by default in the plugin, because it is supported by a sma
...
+ *   e1 % e2      min(p1-s1, p2-s2) + max(s1, s2)         max(s1, s2)
+ *   e1 union e2  max(s1, s2) + max(p1-s1, p2-s2)         max(s1, s2)
+```
+
+However Spark inserts "PromotePrecision" to CAST both sides to the same type.
+GPU mode may fall back to CPU even if the result Decimal precision is within 18 digits.
+For example, `Decimal(8,2)` x `Decimal(6,3)` resulting in `Decimal (15,5)` runs on CPU,
+because due to PromotePrecision, GPU mode assumes the result is `Decimal(19,6)`.
+
+Currently `mod` is not supported on `Decimal` and will fallback on CPU.
```
revans2, line 45: I'm not sure we need to list that here. `mod` or `Remainder` is in the table below and it clearly shows that it does not support decimal. We also don't support it for a large number of other operators. Not sure why mod needs to be called out specially.  If it is just because of binary ops then we should probably call out pmod too.

docs/supported_ops.md
viadea: Sure I just rollback the changes, and modified `TypeChecks.scala` instead.

docs/supported_ops.md
viadea: It makes sense. I just removed `mod` related statement in this section.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -1265,6 +1265,26 @@ object SupportedOpsDocs {
     println("decimals are disabled by default in the plugin, because it is supported by a small")
     println("number of operations presently, which can result in a lot of data movement to and")
     println("from the GPU, slowing down processing in some cases.")
+    println("Result `Decimal` precision and scale follow the same rule as CPU mode in Apache Spark:")
```
revans2, line 1268: This looks fine. The only issue now is that you need to do a build so that the generated file is updated and check that file in too.  The precommit build checks to see if you didn't do this so we can make sure that the docs are always up to date.

viadea:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
viadea: Sure I just build and them pushed the changes on doc as well.

revans2 (approved):

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -1265,6 +1265,26 @@ object SupportedOpsDocs {
...
+    println()
+    println("```")
+    println(" * In particular, if we have expressions e1 and e2 with precision/scale p1/s2 and p2/s2")
```
sameerz, line 1271:
```suggestion
    println(" * In particular, if we have expressions e1 and e2 with precision/scale p1/s1 and p2/s2")
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -1265,6 +1265,26 @@ object SupportedOpsDocs {
...
+    println(" *   e1 * e2      p1 + p2 + 1                             s1 + s2")
+    println(" *   e1 / e2      p1 - s1 + s2 + max(6, s1 + p2 + 1)      max(6, s1 + p2 + 1)")
+    println(" *   e1 % e2      min(p1-s1, p2-s2) + max(s1, s2)         max(s1, s2)")
+    println(" *   e1 union e2  max(s1, s2) + max(p1-s1, p2-s2)         max(s1, s2)")
+    println("```")
+    println()
+    println("However Spark inserts `PromotePrecision` to CAST both sides to the same type.")
+    println("GPU mode may fall back to CPU even if the result Decimal precision is within 18 digits.")
+    println("For example, `Decimal(8,2)` x `Decimal(6,3)` resulting in `Decimal (15,5)` runs on CPU,")
+    println("because due to `PromotePrecision`, GPU mode assumes the result is `Decimal(19,6)`.")
```
sameerz, line 1287: In this example, why does `PromotePrecision` promote the precision by three and scale by one more than the operations dictate?

sameerz (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
revans2: For multiply the regular output formula is  `precision = p1 + p2 + 1` and `scale = s1 + s2` so for the inputs in the example we get 

`precision = 8 + 6 + 1 => 15` and 
`scale = 3 + 2 => 5` or 
`Decimal(15, 5)`. 

Promote precision widens both inputs prior to any of this so that they are the same. It does this with `precision = max(s1, s2) + max(p1-s1, p2-s2)` and `scale = max(s1, s2)`. So when we do this we get

 `precision = max(2, 3) + max(8 -2, 6 - 3) => 3 + max(6, 3) => 3 + 6 => 9` and
 `scale = max(2,3) => 3' or
`Decimal(9, 3)` for both inputs.

Now if we do the original multiply calculation again we get.

`precision = 9 + 9 + 1 => 19` and
`scale = 3 + 3 => 6` or
`Decimal(19, 6)` which is the actual output type of multiply.  Then Spark inserts a `CheckOverflow` afterwards to look at the `Decimal(19, 6)` value that is returned and convert it back into a `Decimal(15, 5)` value.

So the final plan ends up looking like 
```
CheckOverflow(Multiply(Cast(LHS as Decimal(9, 3)), Cast(RHS as Decimal(9, 3))), Decimal(15, 5)
```

Why does spark do this? All I can tell it is because binary op decided that it needs the type of both inputs to be the same, and a `Decimal(6, 3)` is not the same type as a `Decimal(8, 2)` so they decided to "fix" it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
viadea: @revans2  @sameerz 
I accidentally polluted my branch. Can I cancel this PR and open a new PR?

viadea:
Let me open a new PR since my original branch got polluted accidentally.

Issue #2272, opened 2021-04-27
[FEA] elementAt, getArrayItem and getMapValue have different behaviors since Spark-3.1.1 when honoring AnsiMode


wjxiz1992:
**Is your feature request related to a problem? Please describe.**
https://github.com/apache/spark/pull/30297 introduce different behaviors for element_at, getArrayItem and getMapValue when invalid index or key is given to lhs with ansi mode enabled and disabled.

For `select element_at(Array(1,2,3), 4)`:

Spark 3.0.2 + spark.sql.ansi.enabled=true:
```
+-----------------------------+
|element_at(array(1, 2, 3), 4)|
+-----------------------------+
|                         null|
+-----------------------------+
```

Spark 3.1.1 + spark.sql.ansi.enabled=true:
```
java.lang.ArrayIndexOutOfBoundsException: Invalid index: 4, numElements: 3
```

For `select element_at(map(1, 'a', 2, 'b'), 3)`:

Spark 3.0.2 + spark.sql.ansi.enabled=true:
```
+------------------------------+
|element_at(map(1, a, 2, b), 3)|
+------------------------------+
|                          null|
+------------------------------+
```
Spark 3.1.1 + spark.sql.ansi.enabled=true:
```
java.util.NoSuchElementException: Key 3 does not exist.
```
Initial support for elementAt, getArrayItem will always return null for invalid index or key.

**Describe the solution you'd like**
Put corresponding code into 3.1.1 shim layers.

- [ ] GetArrayItem
- [ ] GetMapValue
- [ ] ElementAt

**Additional context**
Initial PR: https://github.com/NVIDIA/spark-rapids/pull/2260

Issue #2276, opened 2021-04-27
[BUG] SPARK-33386 - ansi-mode changed ElementAt/Elt/GetArray behavior in Spark 3.1.1 - fallback to cpu


tgravescs:
**Describe the bug**
Spark made a change to ElementAt/Elt/GetArray with ansi-mode that throws runtime ArrayIndexOutOfBoundsException in Spark 3.1.1 instead of returning NULL. We need to make the same change in our plugin which will be done under https://github.com/NVIDIA/spark-rapids/issues/2272.  Until that is done we should fall back to the CPU for those when ansi-mode is turned on.

See other details of that change:
[SPARK-33386][SQL] Accessing array elements in ElementAt/Elt/GetArray Item should failed if index is out of bound
https://issues.apache.org/jira/browse/SPARK-33386
https://github.com/apache/spark/commit/6d31daeb6a

PR #2350, opened 2021-05-06
AnsiMode support for GetArrayItem GetMapValue and ElementAt for Spark 3.1.1

wjxiz1992:
Fix https://github.com/NVIDIA/spark-rapids/issues/2272 and https://github.com/NVIDIA/spark-rapids/issues/2276
This PR adds shim support for GetArrayItem, GetMapValue and ElementAt to match the CPU behavior on Spark 3.1.1.

This relies on https://github.com/rapidsai/cudf/pull/8209 and https://github.com/NVIDIA/spark-rapids/pull/2260.

More:
This adds an parameter `all_null` for ArrayGen in the data_gen part in integration_test. 
This parameter is used to create null array instead of empty array. 
Null array is used to create a corner case for GetArrayItem:

For a dataframe like:

``` 
+------------------------+
|col_1                   |
+------------------------+
|null                    |
|null                    |
+------------------------+
```
`df.select(col("col_1")[2]).show()` will return without exception:

```
+--------+
|col_1[2]|
+--------+
|    null|
|    null|
+--------+
```


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -119,8 +119,15 @@ case class GpuGetArrayItem(child: Expression, ordinal: Expression)
 
   override def doColumnar(lhs: GpuColumnVector, ordinal: Scalar): ColumnVector = {
     // Need to handle negative indexes...
-    if (ordinal.isValid && ordinal.getInt >= 0) {
-      lhs.getBase.extractListElement(ordinal.getInt)
+    if (ordinal.isValid) {
+      val minNumElements = lhs.getBase.countElements.min.getInt
+      if ( (ordinal.getInt < 0 | minNumElements < ordinal.getInt + 1) & failOnError) {
```
razajafri, line 124: why are we using bitwise or/and instead of logical ones?

wjxiz1992: I must be drunk...Fix.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -87,15 +87,15 @@ class GpuGetArrayItemMeta(
   override def convertToGpu(
       arr: Expression,
       ordinal: Expression): GpuExpression =
-    GpuGetArrayItem(arr, ordinal)
+    GpuGetArrayItem(arr, ordinal, SQLConf.get.ansiEnabled)
```
revans2, line 90: This is wrong. In version prior to 3.1.1 the default value should be false, not based off of the `ansiEnabled` config. Otherwise we will fail when spark does not.

integration_tests/src/main/python/array_test.py
```diff
@@ -109,3 +109,11 @@ def main_df(spark):
...
         return df.select(array_contains(col('a'), chk_val))
     assert_gpu_and_cpu_are_equal_collect(main_df)
+
+@pytest.mark.parametrize('data_gen', array_gens_sample, ids=idfn)
+def test_get_array_item_ansi_fail(data_gen):
+    assert_gpu_and_cpu_error(lambda spark: unary_op_df(
+        spark, data_gen).select(col('a')[100]).collect(),
+                               conf={'spark.sql.ansi.enabled':True,
+                                     'spark.sql.legacy.allowNegativeScaleOfDecimal': True},
+                               error_message='java.lang.ArrayIndexOutOfBoundsException')
```
revans2, line 119: Can we add a separate test for ansi enabled prior to 3.1.1 so we can verify that we are doing the right thing for those versions of Spark?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -119,8 +119,15 @@ case class GpuGetArrayItem(child: Expression, ordinal: Expression)
 
   override def doColumnar(lhs: GpuColumnVector, ordinal: Scalar): ColumnVector = {
     // Need to handle negative indexes...
-    if (ordinal.isValid && ordinal.getInt >= 0) {
-      lhs.getBase.extractListElement(ordinal.getInt)
+    if (ordinal.isValid) {
+      val minNumElements = lhs.getBase.countElements.min.getInt
```
revans2, line 123: This leaks a ColumnVector and a Scalar.  The result of `countElements` must be closed and so much the result of `min`

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -119,8 +119,15 @@ case class GpuGetArrayItem(child: Expression, ordinal: Expression)
...
+      if ( (ordinal.getInt < 0 || minNumElements < ordinal.getInt + 1) && failOnError) {
```
revans2, line 124: What is supposed to happen with a `null` array?  `countElements` will return a `null` for a `null` array and min skips over nulls unless all of them are null.  So is `null[1]` an error in ansi mode or not? If it is then this code will completely miss it. If it is not an error, then we will get an exception, or possibly data corruption when we try to get the int value from the result of `min` if the batch is all null arrays.

revans2 (changes requested):
A few more bugs that I saw when I took a closer look

integration_tests/src/main/python/array_test.py
```diff
@@ -109,3 +110,20 @@ def main_df(spark):
...
+@pytest.mark.skipif(is_before_spark_311(), reason="This will throw exception only in Spark 3.1.1+")
+@pytest.mark.parametrize('data_gen', array_gens_sample, ids=idfn)
+def test_get_array_item_ansi_fail(data_gen):
+    assert_gpu_and_cpu_error(lambda spark: unary_op_df(
+        spark, data_gen).select(col('a')[100]).collect(),
+                               conf={'spark.sql.ansi.enabled':True,
+                                     'spark.sql.legacy.allowNegativeScaleOfDecimal': True},
+                               error_message='java.lang.ArrayIndexOutOfBoundsException')
+
+@pytest.mark.skipif(not is_before_spark_311(), reason="This will throw exception only in Spark 3.1.1+")
```
revans2, line 123: The reason looks like a copy and paste. Under this test is in not clear what it means. It might be nice to update both to say something like `"In Spark 3.1.1+ ANSI mode array index throws on out of range indexes"`

revans2 (commented):
Looking better. My main concern right now is that I don't see how we are setting `failOnError` properly for Spark3.1.1+ I think we need to either check the version number in the meta, which is brittle, or preferably put the rule into the Shim layer so Spark 3.1.1 can override the behavior.

integration_tests/src/main/python/array_test.py
revans2: Sorry I missed that. Looks good then all I have is this nit.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: I did a simple test about this problem:

```
// row data is like:

+------------------------+
|col_1                   |
+------------------------+
|null                    |
|[Java, Scala, C++, a, b]|
+------------------------+

df.select(col("col_1")[2]).show()

+--------+
|col_1[2]|
+--------+
|    null|
|     C++|
+--------+
```
CPU and GPU will return the same result for this case.
But like you said, error occurs when the column contains all null arrays.  
the `countElements` works well, but the `min` will return 0 for this case and will throw the exception here.
(CPU will still return null for them)

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/Spark311Shims.scala
```diff
@@ -17,13 +17,11 @@
 package com.nvidia.spark.rapids.shims.spark311
 
 import java.nio.ByteBuffer
-
```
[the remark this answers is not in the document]
wjxiz1992: Updated.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -87,15 +87,15 @@ class GpuGetArrayItemMeta(
...
+    GpuGetArrayItem(arr, ordinal, SQLConf.get.ansiEnabled)
 }
 
 /**
  * Returns the field at `ordinal` in the Array `child`.
  *
  * We need to do type checking here as `ordinal` expression maybe unresolved.
  */
-case class GpuGetArrayItem(child: Expression, ordinal: Expression)
+case class GpuGetArrayItem(child: Expression, ordinal: Expression, failOnError: Boolean = false)
```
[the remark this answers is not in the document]
wjxiz1992: Done.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: Set to false and add comment.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -165,8 +181,22 @@ case class GpuGetMapValue(child: Expression, key: Expression)
 
   override def prettyName: String = "getMapValue"
 
-  override def doColumnar(lhs: GpuColumnVector, rhs: GpuScalar): ColumnVector =
-    lhs.getBase.getMapValue(rhs.getBase)
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuScalar): ColumnVector = {
+    if (failOnError){
```
firestarman, line 185: This can be simplified as 
``` Scala
    if (failOnError){
      withResource(lhs.getBase.getMapKeyExistence(rhs.getBase)) { keyExistenceColumn =>
        withResource(keyExistenceColumn.all) { exist =>
          if (!exist.getBoolean) {
            throw new NoSuchElementException(s"Key: ${rhs.getValue.asInstanceOf[String]} does " +
              s"not exist in one of the rows in the map column")
          }
        }
      }
    }
    lhs.getBase.getMapValue(rhs.getBase)
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -165,8 +181,22 @@ case class GpuGetMapValue(child: Expression, key: Expression)
...
+      withResource(lhs.getBase.getMapKeyExistence(rhs.getBase)) { keyExistenceColumn =>
+        withResource(keyExistenceColumn.all) { exist =>
+          if (exist.getBoolean) {
+            return lhs.getBase.getMapValue(rhs.getBase)
```
firestarman, line 189: `return` is not necessary

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -165,8 +181,22 @@ case class GpuGetMapValue(child: Expression, key: Expression)
...
+          } else {
+            throw new NoSuchElementException(s"Key: ${rhs.getValue.asInstanceOf[String]} does " +
```
firestarman, line 191: getValue is supposed to return a UTF8String.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -165,8 +181,22 @@ case class GpuGetMapValue(child: Expression, key: Expression)
...
+              s"not exist in one of the rows in the map column")
```
firestarman, line 192:
```suggestion
              s"not exist in at least one of the rows in the map column")
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: Updated.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -119,13 +120,26 @@ case class GpuGetArrayItem(child: Expression, ordinal: Expression)
...
+                  s"$minNumElements")
+            } else {
+              lhs.getBase.extractListElement(ordinal)
+            }
+        }
+      }
     } else {
       GpuColumnVector.columnVectorFromNull(lhs.getRowCount.toInt, dataType)
-    }
+      }
```
firestarman, line 142:
```suggestion
    }
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -119,13 +120,26 @@ case class GpuGetArrayItem(child: Expression, ordinal: Expression)
...
+              throw new ArrayIndexOutOfBoundsException(
+                s"Invalid index: ${ordinal}, minimum numElements in this ColumnVector: " +
+                  s"$minNumElements")
+            } else {
+              lhs.getBase.extractListElement(ordinal)
+            }
+        }
+      }
     } else {
       GpuColumnVector.columnVectorFromNull(lhs.getRowCount.toInt, dataType)
```
firestarman, line 141: How does Spark handle the null case ?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -119,13 +120,26 @@ case class GpuGetArrayItem(child: Expression, ordinal: Expression)
...
-    // Need to handle negative indexes...
     val ordinal = ordinalS.getValue.asInstanceOf[Int]
-    if (ordinalS.isValid && ordinal >= 0) {
-      lhs.getBase.extractListElement(ordinal)
+    if (ordinalS.isValid) {
+      withResource(lhs.getBase.countElements) { numElementsCV =>
+        withResource(numElementsCV.min) {
+          minScalar =>
+            val minNumElements = minScalar.getInt
+            if ( (ordinal < 0 || minNumElements < ordinal + 1) &&
```
firestarman, line 129:
```suggestion
            if ((ordinal < 0 || minNumElements < ordinal + 1) &&
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -103,21 +103,55 @@ case class GpuElementAt(left: Expression, right: Expression)
...
           GpuColumnVector.columnVectorFromNull(lhs.getRowCount.toInt, dataType)
         }
       }
       case _: MapType => {
-        lhs.getBase.getMapValue(rhs.getBase)
+        if (failOnError) {
+          withResource(lhs.getBase.getMapKeyExistence(rhs.getBase)){ keyExistenceColumn =>
+            withResource(keyExistenceColumn.all()) { exist =>
+              if (exist.getBoolean) {
+                return lhs.getBase.getMapValue(rhs.getBase)
```
firestarman, line 144: `return ` is not necessary.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: Spark will return null, for like null[5].This case is covered by integration_test `test_array_element_at_all_null_ansi_not_fail`

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
wjxiz1992: Thanks, forgot to add integration test for this case so I didn't catch it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -103,21 +104,55 @@ case class GpuElementAt(left: Expression, right: Expression)
...
+            val ordinalValue = rhs.getValue.asInstanceOf[Int]
+            withResource(lhs.getBase.countElements) { numElementsCV =>
+              withResource(numElementsCV.min) { minScalar =>
+                val minNumElements = minScalar.getInt
+                // index out of bound
+                // Note: when the column is containing all null arrays, CPU will not throw, so make
+                // GPU to behave the same.
+                if (minNumElements < math.abs(ordinalValue) &&
+                  lhs.getBase.getNullCount != lhs.getBase.getRowCount &&
+                  failOnError) {
```
firestarman, line 120: nit: The below can avoid the evaluation of the follwoing two expressions if `failOnError ` is false.
```Scala
                if (failOnError &&
                  minNumElements < math.abs(ordinalValue) &&
                  lhs.getBase.getNullCount != lhs.getBase.getRowCount) {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -119,10 +121,22 @@ case class GpuGetArrayItem(child: Expression, ordinal: Expression)
...
     val ordinal = ordinalS.getValue.asInstanceOf[Int]
-    if (ordinalS.isValid && ordinal >= 0) {
-      lhs.getBase.extractListElement(ordinal)
+    if (ordinalS.isValid) {
+      withResource(lhs.getBase.countElements) { numElementsCV =>
+        withResource(numElementsCV.min) { minScalar =>
+          val minNumElements = minScalar.getInt
+          if ((ordinal < 0 || minNumElements < ordinal + 1) &&
+            numElementsCV.getRowCount != numElementsCV.getNullCount &&
+            failOnError) {
```
firestarman, line 131: nit: The same as the above about moving `failOnError`.

firestarman (commented):
LGTM, only some nits.

wjxiz1992:
build

revans2 (dismissed):

wjxiz1992:
build

firestarman (approved):

PR #2351, opened 2021-05-06
Fix NPE when spilling with debug logging enabled

jlowe:
This ensures that a valid `BufferMeta` is created for `SpillableBuffer`'s table metadata.  This will avoid an NPE when code tries to examine the size of a buffer via its metadata or whether the data is compressed.


revans2 (approved):
Thanks for fixing my bad code.

jlowe:
build

PR #2352, opened 2021-05-06
Fixed issues with fixed width join calcs and optimized project and filter for noop cases

revans2:
Also has project and filter avoid later contig splits by doing a noop in some cases.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
```diff
@@ -155,7 +155,7 @@ object JoinGatherer extends Arm {
       val rowEstimate: Long = gatherer.getFixedWidthBitSize match {
         case Some(fixedSize) =>
```
jlowe, line 156: Nit: Would be a bit more readable if this was named `fixedBitSize`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
```diff
@@ -44,19 +45,45 @@ object GpuProjectExec extends Arm {
...
+    case _ => None
+  }
+
+  private def extractSingleBoundIndex(boundExprs: Seq[Expression]): Seq[Option[Int]] =
+    boundExprs.map(extractSingleBoundIndex)
+
+  def isNoopProject(cb: ColumnarBatch, boundExprs: Seq[Expression]): Boolean = {
+    if (boundExprs.length == cb.numCols()) {
+      extractSingleBoundIndex(boundExprs).zip(0 until cb.numCols()).forall {
+        case (Some(foundIndex), expectedIndex) if foundIndex == expectedIndex => true
```
jlowe, line 61:
```suggestion
        case (Some(foundIndex), expectedIndex) => foundIndex == expectedIndex
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
```diff
@@ -119,21 +146,33 @@ object GpuFilter extends Arm {
...
-        (0 until batch.numCols()).map(i => batch.column(i).dataType())
-      filtered = tbl.filter(filterConditionCv.getBase)
-      GpuColumnVector.from(filtered, colTypes.toArray)
-    } finally {
-      Seq(filtered, tbl, filterConditionCv, batch).safeClose()
+    withResource(batch) { batch =>
+      val checkedFilterMap = withResource(
+        GpuProjectExec.projectSingle(batch, boundCondition)) { filterMap =>
+        // If  filter is a noop then return a None
+        withResource(filterMap.getBase.all()) { allIncluded =>
```
jlowe, line 155: Has there been any performance analysis of adding this extra reduction operation and synchronization to check for the filter corner case?  Should be a nice memory win for the common case of filtering for nulls after loading from input with no nulls, but not sure how much this is costing for the other cases.

jlowe (approved):
All minor nits that aren't important and just a question, but otherwise lgtm.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
revans2: I ran q24a and q95. I didn't see anything in the nsys profile that indicated that this was causing a performance change at all good or bad. I am happy to remove it if you feel the complexity is not worth it. 

The profile put the `all` check at about the same speed as calculating the null count for a column. But that said I also didn't see any significant decrease in the use of contiguous split either. There really are just not that many cases where the input to a filter is already contiguous. They typically are happening right after a join, in our case, or after reading in input. This is because the filters are pushed to be as soon as possible in the query.

The profiles for the project also showed very little change. It is just metadata, but I did see a small drop in the number of contig splits needed for shuffled joins, but it was only on the data side not the key side. I was going to file a follow on issue to look into it more, but like I said I am happy to remove the extra code if the complexity is not worth it.

revans2:
Closing this as there is too much mixed together in it. I'll split it up and put up separate PRs.

PR #2356, opened 2021-05-06
Fix issue with fixed width size calculation for join

revans2:
I zigged, aka divided by 8, when I should have zagged, multiplied by 8, when converting from bits to bytes.


revans2:
build

jlowe (approved):

PR #2357, opened 2021-05-06
Have broadcast exchange exec produce a contiguous table

revans2:
This should help reduce the number of contiguous splits we have to do on large broadcast joins.  It is a minor performance optimization.


revans2:
build

jlowe (approved):

Issue #2359, opened 2021-05-06
[BUG] Alias is different in spark 3.1.0 but our canonicalization code doesn't handle


tgravescs:
**Describe the bug**
https://github.com/NVIDIA/spark-rapids/blame/branch-0.6/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala#L265  

That uses Alias() but alias has a different signature  in 3.1.0 - https://github.com/NVIDIA/spark-rapids/commit/cce1ec1ed401bcd7465f5bef9ab28ed38aa5a9a8  

I got an error running the tests on databricks 8.2 about Alias not being present which made me track this down.  Although I only got this error after upmerging to the latest 0.6 branch.  I'm not sure why our existing test on Spark 3.1.1 aren't catching this, need to investigate. It might be we just haven't ever exercised that code for it to fail.

 ```
py4j.protocol.Py4JJavaError: An error occurred while calling o1338.collectToPython.
E                   : java.lang.NoSuchMethodError: org.apache.spark.sql.catalyst.expressions.Alias.<init>(Lorg/apache/spark/sql/catalyst/expressions/Expression;Ljava/lang/String;Lorg/apache/spark/sql/catalyst/expressions/ExprId;Lscala/collection/Seq;Lscala/Option;)V
at com.nvidia.spark.rapids.GpuExec.$anonfun$doCanonicalize$2(GpuExec.scala:265)
E                   	at org.apache.spark.sql.catalyst.plans.QueryPlan.$anonfun$mapExpressions$1(QueryPlan.scala:117)
E                   	at org.apache.spark.sql.catalyst.trees.CurrentOrigin$.withOrigin(TreeNode.scala:81)
E                   	at org.apache.spark.sql.catalyst.plans.QueryPlan.transformExpression$1(QueryPlan.scala:117)
E                   	at org.apache.spark.sql.catalyst.plans.QueryPlan.recursiveTransform$1(QueryPlan.scala:128)
E                   	at org.apache.spark.sql.catalyst.plans.QueryPlan.$anonfun$mapExpressions$3(QueryPlan.scala:133)
E                   	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
E                   	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
E                   	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
[...]

PR #2361, opened 2021-05-06
Fix Alias to properly call the shim version

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/2359

Alias needs to be shimmed before spark 3.1.0 made changes to it. https://github.com/NVIDIA/spark-rapids/commit/cce1ec1ed401bcd7465f5bef9ab28ed38aa5a9a8

It appears that got lost and our tests never caught this.   

specifically call updatedPlan.canonicalized for testing cases which would have caught this error.  It doesn't test anything other then making sure it can generate the canonicalized plan.


tgravescs:
build

jlowe (approved):

PR #2362, opened 2021-05-06
ParquetCachedBatchSerializer (PCBS) shouldn't write int96

razajafri:
I made a mistake writing int96 to cache

### DIFF
--- a/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
+++ b/shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
@@ -402,7 +402,8 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
       StructField(s"_col$index", field.dataType, field.nullable, field.metadata)
     }))
     val opts = GpuParquetFileFormat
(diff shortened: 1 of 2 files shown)


shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -402,7 +402,8 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
       StructField(s"_col$index", field.dataType, field.nullable, field.metadata)
     }))
     val opts = GpuParquetFileFormat
-      .parquetWriterOptionsFromSchema(ParquetWriterOptions.builder(), schemaWithUnambiguousNames)
+      .parquetWriterOptionsFromSchema(ParquetWriterOptions.builder(), schemaWithUnambiguousNames,
+        false)
```
andygrove, line 406:
```suggestion
        writeInt96 = false)
```

nartal1 (approved):

razajafri:
build

andygrove (approved):
LGTM. I pulled the code locally and checkecd the usage and it looks like this only changes behavior for the cache case where we no longer write INT96.

PR #2369, opened 2021-05-07
Print full column type when type conversion check fails

jlowe:
This updates the type conversion check error message to include the full cudf type when working with nested types rather than just printing `LIST` or `STRUCT` as it does today.  this can be helpful when implementing RAPIDS accelerated UDFs involving nested types to debug type conversion errors.


jlowe:
build

revans2 (approved):

jlowe:
build

Issue #2013, opened 2021-03-24
[FEA] Support concatenating ArrayType columns


Dooyoung-Hwang:
I request a feature of concatenating columns of ArrayType into one column on GPU.

- Test code
```python
data = [(1,[0.1, 0.4], [0.1, 0.4], 1.0, -1.0), (2,  [0.3, 0.5], [0.3, 0.5], 2.0, -1.0), (3, [0.3, 0.5], [2.1, 2.2], 3.0, -1.0)]
schema = StructType([ \
    StructField("a",IntegerType(),True), \
    StructField("b",ArrayType(FloatType()),True), \
    StructField("c",ArrayType(FloatType()), True), \
    StructField("d",FloatType(),True), \
    StructField("e",FloatType(),True)
  ])

df1 = spark.createDataFrame(data=data,schema=schema).withColumn("arr1", concat("b", "c"))
print(df1._jdf.queryExecution())
```

- Output log from spark-driver
```
!Expression concat(b#7221, c#7222) cannot run on GPU because expression Concat concat(b#7221, c#7222) produces an unsupported type ArrayType(FloatType,true); expression AttributeReference b#7221 produces an unsupported type ArrayType(FloatType,true); expression AttributeReference c#7222 produces an unsupported type ArrayType(FloatType,true)
```

revans2:
I filed https://github.com/rapidsai/cudf/issues/7767 for this.

PR #2379, opened 2021-05-10
support GpuConcat on ArrayType

sperlingxx:
Current PR is to support `GpuConcat` on ArrayType, which is required in issue #2013.
In addition, this PR also introduces some refinement on the implementation of string concatenation.


sperlingxx:
I found another small problem of cuDF implementation during running tests on Array of String locally. And I've put up a [PR](https://github.com/rapidsai/cudf/pull/8210) to fix it.

jlowe (dismissed):

jlowe:
build

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -16,10 +16,75 @@
...
+
+  override def columnarEval(batch: ColumnarBatch): Any = dataType match {
+    case StringType => stringConcat(batch)
+    case ArrayType(_, _) => listConcat(batch)
+  }
+
+  private def stringConcat(batch: ColumnarBatch): GpuColumnVector = {
+    val rows = batch.numRows()
+
+    withResource(ArrayBuffer[ColumnVector]()) { buffer =>
```
gerashegalov, line 49: nit: here and later: `ArrayBuffer.empty[ColumnVector]` for readability.

gerashegalov (dismissed):
LGTM

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
sperlingxx: Done.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -16,10 +16,75 @@
...
+  override def columnarEval(batch: ColumnarBatch): Any = dataType match {
+    case StringType => stringConcat(batch)
+    case ArrayType(_, _) => listConcat(batch)
+  }
+
+  private def stringConcat(batch: ColumnarBatch): GpuColumnVector = {
+    val rows = batch.numRows()
+
+    withResource(ArrayBuffer.empty[ColumnVector]) { buffer =>
+      withResource(GpuScalar.from(null, StringType)) { nullScalar =>
```
jlowe, line 50: Nit: We know the exact type we need, so no need to dispatch on the type at runtime.
```suggestion
      withResource(Scalar.fromString(null)) { nullScalar =>
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -16,10 +16,75 @@
...
+        // build input buffer
+        children.foreach { child =>
+          child.columnarEval(batch) match {
+            case cv: GpuColumnVector =>
+              buffer += cv.getBase
+            case null =>
+              buffer += GpuColumnVector.from(nullScalar, rows, StringType).getBase
+            case sv: Any =>
+              val scalar = GpuScalar.from(sv.asInstanceOf[UTF8String].toString, StringType)
+              withResource(scalar) { scalar =>
```
jlowe, line 60: There really should be a `Scalar.fromBytesAsString` so we can avoid the needless copying of the bytes to a String just so we can copy them back into bytes again for the Scalar.  :facepalm:  I filed https://github.com/rapidsai/cudf/issues/8233 to track that.
```suggestion
              withResource(Scalar.fromString(sv.asInstanceOf[UTF8String].toString)) { scalar =>
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -16,10 +16,75 @@
...
+                buffer += GpuColumnVector.from(scalar, rows, StringType).getBase
+              }
+          }
+        }
+        // run string concatenate
+        withResource(GpuScalar.from("", StringType)) { emptyScalar =>
```
jlowe, line 66:
```suggestion
        withResource(Scalar.fromString("")) { emptyScalar =>
```

sperlingxx:
build

jlowe (approved):

jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -16,10 +16,74 @@
...
+            case sv: Any =>
+              withResource(Scalar.fromString(sv.asInstanceOf[UTF8String].toString)) { scalar =>
+                buffer += GpuColumnVector.from(scalar, rows, StringType).getBase
+              }
+          }
+        }
+        // run string concatenate
+        withResource(Scalar.fromString("")) { emptyScalar =>
+          GpuColumnVector.from(ColumnVector.stringConcatenate(emptyScalar, nullScalar,
+            buffer.toArray[ColumnView]), StringType)
```
revans2, line 67: Could you please file a bug against cudf to remove the assert for at least two entries, and an issue against ourselves to add a test to verify that we can concat just a single string column?

https://github.com/rapidsai/cudf/blob/92f067dcd80ce11c0431432f3817bebfa97816a0/java/src/main/java/ai/rapids/cudf/ColumnVector.java#L512

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -16,10 +16,74 @@
...
+    withResource(ArrayBuffer[ColumnVector]()) { buffer =>
+      // build input buffer
+      children.foreach { child =>
+        child.columnarEval(batch) match {
+          case cv: GpuColumnVector => buffer += cv.getBase
+          case _ => throw new UnsupportedOperationException("Unsupported GpuScalar of List")
+        }
+      }
+      // run list concatenate
+      GpuColumnVector.from(ColumnVector.listConcatenateByRow(buffer: _*), dataType)
```
revans2, line 83: We verify in the java code that there are at least two columns.

https://github.com/rapidsai/cudf/blob/92f067dcd80ce11c0431432f3817bebfa97816a0/java/src/main/java/ai/rapids/cudf/ColumnVector.java#L551

I don't know if this is a bug in the java code or if the underlying list concat in cudf really does need at least two columns.  Either way CPU Spark supports concat with a single column.

```
scala> spark.range(2).selectExpr("concat(array(id, id))").show
+---------------------+
|concat(array(id, id))|
+---------------------+
|               [0, 0]|
|               [1, 1]|
+---------------------+
```

If the cudf implementation supports it, then we should file an issue just like with the string support and fix it before this goes in. If not then we need to put in a work around here.

revans2 (changes requested):
Mostly nits but there are some corner cases that string concat missed and that this also is missing.

I tested the string concat code and it works if we have asserts disabled. If we enable them it fails. I don't know about list concat. There is probably also an issue with no columns being passed in. I am fine with letting this slide for now because you have to go out of your way to disable expression folding for it to be an issue, but it might be good to just fall back to the CPU if there are no children. That way we don't have to ever worry about it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -16,10 +16,74 @@
...
+  override def columnarEval(batch: ColumnarBatch): Any = dataType match {
+    case StringType => stringConcat(batch)
+    case ArrayType(_, _) => listConcat(batch)
+  }
+
+  private def stringConcat(batch: ColumnarBatch): GpuColumnVector = {
+    val rows = batch.numRows()
+
+    withResource(ArrayBuffer.empty[ColumnVector]) { buffer =>
+      withResource(Scalar.fromString(null)) { nullScalar =>
```
revans2, line 50: Looks like you didn't import everything you need from cudf. Compilation is failing.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
sperlingxx: I checked native side, both `cudf::strings::concatenate` and `cudf::lists::concatenate_rows` supports single column inputs. And I filed issues against Java side assertion:
* https://github.com/rapidsai/cudf/issues/8246
* https://github.com/NVIDIA/spark-rapids/issues/2414

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -16,10 +16,74 @@
...
+              buffer += GpuColumnVector.from(nullScalar, rows, StringType).getBase
+            case sv: Any =>
+              withResource(Scalar.fromString(sv.asInstanceOf[UTF8String].toString)) { scalar =>
+                buffer += GpuColumnVector.from(scalar, rows, StringType).getBase
+              }
+          }
+        }
+        // run string concatenate
+        withResource(Scalar.fromString("")) { emptyScalar =>
+          GpuColumnVector.from(ColumnVector.stringConcatenate(emptyScalar, nullScalar,
```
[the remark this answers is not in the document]
sperlingxx: Fixed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -16,10 +16,74 @@
...
+          child.columnarEval(batch) match {
+            case cv: GpuColumnVector =>
+              buffer += cv.getBase
+            case null =>
+              buffer += GpuColumnVector.from(nullScalar, rows, StringType).getBase
+            case sv: Any =>
+              withResource(Scalar.fromString(sv.asInstanceOf[UTF8String].toString)) { scalar =>
+                buffer += GpuColumnVector.from(scalar, rows, StringType).getBase
+              }
+          }
```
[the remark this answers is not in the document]
sperlingxx: Done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -16,10 +16,74 @@
...
+    withResource(ArrayBuffer.empty[ColumnVector]) { buffer =>
+      withResource(Scalar.fromString(null)) { nullScalar =>
+        // build input buffer
+        children.foreach { child =>
+          child.columnarEval(batch) match {
+            case cv: GpuColumnVector =>
+              buffer += cv.getBase
+            case null =>
+              buffer += GpuColumnVector.from(nullScalar, rows, StringType).getBase
+            case sv: Any =>
```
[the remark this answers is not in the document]
sperlingxx: Done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/collectionOperations.scala
```diff
@@ -16,10 +16,74 @@
...
+      super.dataType
+    }
+  }
+
+  override def nullable: Boolean = children.exists(_.nullable)
+
+  override def columnarEval(batch: ColumnarBatch): Any = dataType match {
+    case StringType => stringConcat(batch)
+    case ArrayType(_, _) => listConcat(batch)
+  }
```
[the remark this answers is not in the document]
sperlingxx: Done

revans2 (approved):
Still would like to see us prevent against the empty concat when folding is disabled, but it is such a rare corner case it is not important enough to hold this up over it.

revans2:
build

sperlingxx:
build

sperlingxx:
> Still would like to see us prevent against the empty concat when folding is disabled, but it is such a rare corner case it is not important enough to hold this up over it.

I added missing support of empty concat. And I also found that cuDF concat can be simply bypassed in single column concat.

jlowe (approved):

integration_tests/src/main/python/collection_ops_test.py
```diff
@@ -17,10 +17,48 @@
...
 from data_gen import *
 from pyspark.sql.types import *
+from string_test import mk_str_gen
+import pyspark.sql.functions as f
 
 nested_gens = [ArrayGen(LongGen()),
                StructGen([("a", LongGen())]),
                MapGen(StringGen(pattern='key_[0-9]', nullable=False), StringGen())]
+# additional test for NonNull Array because of https://github.com/rapidsai/cudf/pull/8181
+non_nested_array_gens = [ArrayGen(sub_gen, nullable=nullable)
```
wjxiz1992, line 27: Not familar with the concat op, but do we need to check _null array_ along with _array[null]_ ?

sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -16,37 +16,63 @@
...
+      GpuColumnVector.from(
+        ColumnVector.stringConcatenate(buffer.toArray[ColumnView]), StringType)
+    }
+  }
+
+  private def listConcat(batch: ColumnarBatch): GpuColumnVector = {
+    withResource(ArrayBuffer[ColumnVector]()) { buffer =>
+      // build input buffer
+      children.foreach { child =>
+        child.columnarEval(batch) match {
```
revans2, line 69: This should use `columnarEvalToColumn` as well. That way if you get a GpuScalar back it will not blow up.

integration_tests/src/main/python/collection_ops_test.py
```diff
@@ -17,10 +17,48 @@
...
+@pytest.mark.parametrize('data_gen', non_nested_array_gens, ids=idfn)
+def test_concat_list(data_gen):
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: binary_op_df(spark, data_gen).selectExpr('concat(a)'))
+
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: binary_op_df(spark, data_gen).selectExpr('concat(a, b)'))
+
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: three_col_df(spark, data_gen, data_gen, data_gen
```
revans2, line 39: Sorry I forgot this before. We need to test with literal arrays as well. with concat. There was nothing in the checks that would have prevented concat from supporting literal arrays so we need to test them.

sperlingxx:
build

integration_tests/src/main/python/collection_ops_test.py
sperlingxx: For now, concat on `GpuScalar` is also supported. And corresponding tests are appended.

revans2 (approved):
Looks good

integration_tests/src/main/python/collection_ops_test.py
```diff
@@ -17,10 +17,69 @@
...
+            lambda spark: binary_op_df(spark, gen).select(
+                f.concat(),
+                f.concat(f.col('a')),
+                f.concat(s1),
+                f.concat(f.col('a'), f.col('b')),
+                f.concat(f.col('a'), f.col('b'), f.col('a')),
+                f.concat(s1, f.col('b')),
+                f.concat(f.col('a'), s2),
+                f.concat(f.lit(None).cast('string'), f.col('b')),
+                f.concat(f.col('a'), f.lit(None).cast('string')),
```
tgravescs, line 80: do we have an all nulls test case?  I guess this line might hit it if col a can generate nulls.

tgravescs: similar for arrays, do we want an all nulls test case for arrays

sperlingxx: Frankly, I am not sure whether there exists rows consisting of all null fields, with default random seed.

sperlingxx: But I think the all null case may be not that special, since the output row will be null only if one input field is null.

PR #2381, opened 2021-05-10
Add support for Databricks 8.2 runtime

tgravescs:
Add support for Databricks 8.2 runtime.  This is a bit odd because it reports that it is based on Spark version 3.1.0 when really its based on Spark 3.1.1.  I made the version be spark 310db to match what it actually reports and thinking if they change it to 3.1.1 then perhaps we will need another shim.  If objections or thoughts on this let me know.

There were additional incompatibilities here, the main one in the base python runner so much of the UDF pandas operations that used the python runner had to move to the shim.  I tried to refactor things such that you can easily diff the files copied there with the base files. I did think about trying to pull apart more, but with the spark 3.2.0 shim layer stuff coming this seems to fit that model better anyway. I removed some classes from the base arrow exec to move to their own files.

There are also a few changes in classes themselves. I tried to tag all of these with "DB SPECIFIC".  Databricks has a serializer specific for the grouped map udf in pandas as well.

The GpuWindowInPandas needed doExecuteColumnar to move to the shim to properly handle that now in the 310db shim. since it was already split apart I left it that way.  I also had to change the package for it to be in the org.apache.spark one, so the diff shows it as deleted and added.

This does not include the files needed to enable the CI for it, that will come after.


tgravescs:
build

tgravescs:
note alot of the files are copies of other files.  These are essentially copies of spark301db:

GpuWindowInPandasExec.scala
GpuSortMergeJoinExec.scala
GpuBroadcastExchangeExec.scala
GpuShuffledHashJoinExec.scala
GpuBroadcastHashJoinExec.scala
GpuBroadcastNestedLoopJoinExec.scala
GpuShuffleExchangeExec.scala
GpuFileScanRDD.scala

If you diff there should be package name change and potentially comment tagging databricks specific

These are also copies but not in spark301db, they are copies of the base files due to base python runner changes and need to be compiled against that base version in databricks. So if you diff those they should be very small.

GpuFlatMapGroupsInPandasExec.scala - contains serializer change for group map udf in pandas
BatchGroupUtils.scala
GpuAggregateInPandasExec.scala
GpuArrowEvalPythonExec.scala

new files for databricks specific:
spark310db/src/main/scala/org/apache/spark/sql/rapids/shims/spark310db/GpuGroupUDFArrowPythonRunner.scala

shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/GpuWindowInPandasExec.scala
```diff
@@ -1,42 +0,0 @@
-/*
```
jlowe, line 1: I was a bit surprised to see this class moved and have so much difference from before when it's the 301 shim and not a Databricks thing.  It looks like all Databricks needs to do is override `doColumnar` from the existing base class, but i understand it's not as consistent as what we've done with the other classes where nothing is overridden in the base class by the shim version, as the base class is truly reusable across shim versions.

I could see it either way, but this PR would probably be a lot smaller if we went the "keep 301 where it is" route.  Curious what others think.

shims/spark310db/src/main/scala/com/nvidia/spark/rapids/shims/spark310db/GpuShuffleExchangeExec.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  override def getShuffleRDD(
+      partitionSpecs: Array[ShufflePartitionSpec],
+      partitionSizes: Option[Array[Long]]): RDD[_] = {
+    throw new UnsupportedOperationException
+  }
+
+  // DB SPECIFIC - throw if called since we don't know how its used
+  override def withNewOutputPartitioning(outputPartitioning: Partitioning) = {
+    throw new UnsupportedOperationException
```
jlowe, line 43: Do we need a followup issue to track this?

shims/spark310db/src/main/scala/org/apache/spark/sql/rapids/shims/spark310db/GpuGroupUDFArrowPythonRunner.scala
```diff
@@ -0,0 +1,144 @@
...
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.rapids.execution.python.GpuPythonHelper
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.util.ArrowUtils
+import org.apache.spark.sql.vectorized.ColumnarBatch
+import org.apache.spark.util.Utils
+import org.apache.spark.sql.rapids.execution.python.GpuPythonUDF
+
+/**
+ * Group Map UDF specific serializer for databricks.
```
jlowe, line 52:
```suggestion
 * Group Map UDF specific serializer for Databricks.
```

shims/spark310db/src/main/scala/org/apache/spark/sql/rapids/shims/spark310db/GpuGroupUDFArrowPythonRunner.scala
```diff
@@ -0,0 +1,144 @@
...
+ */
+class GpuGroupUDFArrowPythonRunner(
```
jlowe, line 54: Might be helpful to expand the Scaladoc for this class to give an overview of what it does differently than the Apache Spark serializer.

shims/spark310db/src/main/scala/org/apache/spark/sql/rapids/shims/spark310db/GpuGroupUDFArrowPythonRunner.scala
```diff
@@ -0,0 +1,144 @@
...
+          PythonRDD.writeUTF(v, dataOut)
+        }
+
+        PythonUDFRunner.writeUDFs(dataOut, funcs, argOffsets)
+      }
+
+      protected override def writeIteratorToStream(dataOut: DataOutputStream): Unit = {
+        // write out number of columns
+        Utils.tryWithSafeFinally {
+          // databricks uses special GroupUDFSerializer so we have to match the logic it expects,
```
jlowe, line 90:
```suggestion
          // Databricks uses special GroupUDFSerializer so we have to match the logic it expects,
```

shims/spark310db/src/main/scala/org/apache/spark/sql/rapids/shims/spark310db/GpuGroupUDFArrowPythonRunner.scala
```diff
@@ -0,0 +1,144 @@
...
+          // which is a arrow stream for each grouped data
```
jlowe, line 91:
```suggestion
          // which is an Arrow stream for each grouped data
```

shims/spark310db/src/main/scala/org/apache/spark/sql/rapids/shims/spark310db/GpuGroupUDFArrowPythonRunner.scala
```diff
@@ -0,0 +1,144 @@
...
+          val builder = ArrowIPCWriterOptions.builder()
+          builder.withMaxChunkSize(batchSize)
+          builder.withCallback((table: Table) => {
+            table.close()
+            GpuSemaphore.releaseIfNecessary(TaskContext.get())
+          })
+          // Flatten the names of nested struct columns, required by cudf arrow IPC writer.
```
jlowe, line 98:
```suggestion
          // Flatten the names of nested struct columns, required by cudf Arrow IPC writer.
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/python/GpuPythonUDF.scala
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright (c) 2020-2021, NVIDIA CORPORATION.
```
jlowe, line 2: 2020 copyrights intentional on this new file?

pom.xml
```diff
@@ -131,13 +131,25 @@
 	        <rat.consoleOutput>true</rat.consoleOutput>
             </properties>
         </profile>
+        <profile>
+            <id>databricks310</id>
+            <properties>
+	        <rat.consoleOutput>true</rat.consoleOutput>
```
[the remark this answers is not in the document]
tgravescs: I'd be fine with that, I guess only downside might be if it scrolls to far on screen for developer, but I think it usually quits pretty quickly afterwards.  I'll change it and if we decide we don't like it we can always change back.

shims/aggregator/pom.xml
```diff
@@ -53,9 +64,37 @@
...
+                    <groupId>com.nvidia</groupId>
+                    <artifactId>rapids-4-spark-shims-spark310-databricks_${scala.binary.version}</artifactId>
+                    <version>${project.version}</version>
+                    <scope>compile</scope>
+                </dependency>
             </dependencies>
         </profile>
-       <profile>
+        <profile>
+            <id>include-databricks-301</id>
```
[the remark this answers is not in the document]
tgravescs: ok so the -Pdatabricks301 profile for instance is also in the shims/pom.xml which triggers the build, then in this pom file -Pdatabricks301 results in it being included in the jar.  The -Pinclude-databricks-301 doesn't trigger the build of it but includes it in the jar by pulling from maven repo.  We could change to do -Pdatabricks301 -Pinclude-databricks-301 but to me that isn't as user friendly for building.

jenkins/databricks/build.sh
```diff
@@ -21,8 +21,16 @@ SPARKSRCTGZ=$1
...
-BASE_SPARK_VERSION=${BASE_SPARK_VERSION:-'3.0.1'}
+BASE_SPARK_VERSION_TO_INSTALL_DATABRICKS_JARS=3.1.0
+BUILD_PROFILES=${BUILD_PROFILES:-'databricks310,!snapshot-shims'}
+BASE_SPARK_VERSION=${BASE_SPARK_VERSION:-'3.1.1'}
+# the version of spark used when we install the databricks jars in .m2
+# 3.1.0-databricks is add because its actually based on Spark 3.1.1
+BASE_SPARK_VERSION_TO_INSTALL_DATABRICKS_JARS=${BASE_SPARK_VERSION_TO_INSTALL_DATABRICKS_JARS:-$BASE_SPARK_VERSION}
+SPARK_VERSION_TO_INSTALL_DATABRICKS_JARS=$BASE_SPARK_VERSION_TO_INSTALL_DATABRICKS_JARS-databricks
+# something like spark_3_1 or spark_3_0
+SPARK_MAJOR_VERSION_NUM_STRING=$(echo ${BASE_SPARK_VERSION} | sed 's/\./\_/g' | cut -d _ -f 1,2)
```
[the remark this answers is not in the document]
jlowe: Ah, I missed that this is not some arbitrary name on our part but we're matching an external "standard."

shims/aggregator/pom.xml
```diff
@@ -43,6 +43,17 @@
...
+                    <groupId>com.nvidia</groupId>
+                    <artifactId>rapids-4-spark-shims-spark310-databricks_${scala.binary.version}</artifactId>
+                    <version>${project.version}</version>
+                    <scope>compile</scope>
+                </dependency>
+            </dependencies>
+        </profile>
         <profile>
             <!-- use a separate profile to just pull databricks from maven repository without building it -->
             <id>include-databricks</id>
```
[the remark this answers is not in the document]
tgravescs: yeah if its alright I'll file a followup for this because I need to do the CI/CD for databricks 8.2 and at that point was thinking of removing some individual ones again.  I added it to https://github.com/NVIDIA/spark-rapids/issues/2390

pom.xml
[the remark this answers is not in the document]
tgravescs: so it looks like it still writes out the rat.txt file, this just causes it to also print a summary and unapproved files to the screen.  That seems pretty safe to me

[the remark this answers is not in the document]
tgravescs: I left this change off for now

jenkins/databricks/build.sh
```diff
@@ -21,8 +21,16 @@ SPARKSRCTGZ=$1
 # version of Apache Spark we are building against
 BASE_SPARK_VERSION=$2
 BUILD_PROFILES=$3
-BUILD_PROFILES=${BUILD_PROFILES:-'databricks301,!snapshot-shims'}
-BASE_SPARK_VERSION=${BASE_SPARK_VERSION:-'3.0.1'}
+BASE_SPARK_VERSION_TO_INSTALL_DATABRICKS_JARS=3.1.0
+BUILD_PROFILES=${BUILD_PROFILES:-'databricks310,!snapshot-shims'}
+BASE_SPARK_VERSION=${BASE_SPARK_VERSION:-'3.1.1'}
+# the version of spark used when we install the databricks jars in .m2
```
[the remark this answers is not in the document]
tgravescs: done

shims/spark310db/src/main/scala/org/apache/spark/sql/rapids/shims/spark310db/GpuGroupUDFArrowPythonRunner.scala
tgravescs: done

shims/spark310db/src/main/scala/org/apache/spark/sql/rapids/shims/spark310db/GpuGroupUDFArrowPythonRunner.scala
tgravescs: done

tgravescs:
build

shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/GpuWindowInPandasExec.scala
tgravescs: looking into reverting since it will make review easier and ideally we want to remove 3.0.1 shim anyway.

integration_tests/src/main/python/spark_session.py
```diff
@@ -94,5 +94,6 @@ def with_gpu_session(func, conf={}):
     copy['spark.rapids.sql.decimalType.enabled'] = 'true'
     return with_spark_session(func, conf=copy)
 
+# databricks runtime 8.x and lower show 3.1.0 even though really 3.1.1
 def is_before_spark_311():
-    return spark_version() < "3.1.1"
+    return (is_databricks_runtime() and spark_version() < "3.1.0") or (not is_databricks_runtime() and spark_version() < "3.1.1")
```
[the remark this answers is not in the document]
tgravescs: I committed this without thinking about it, this won't work because databricks is < 3.1.1 so it matches the first one which we don't want because we want the 310db to show as not before 311, I'm going revert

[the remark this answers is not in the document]
tgravescs: Note tests fail on databricks 310db which is why I caught this

PR #2383, opened 2021-05-10
Remove unneeded join key "optimization"

revans2:
When putting in the initial out of core join support I added in a fancy optimization so that in some cases we could avoid gathering keys on one side of the join or the other, because the keys end up being duplicates of each other. It is still a good optimization in theory, and we may try to re-implement it in the future, but for now this removes that optimization.  This is because 

1) It added in a lot of complexity that was hard to reason about, and I am about to try and dynamically fall back to the CPU in some cases and the extra complexity gets in the way.
2) It ended up splitting the keys from the data for the built side. This resulted in contiguous split calls for the built side keys, which in all but some very specific cases were duplicates of data in the `builtData`, so we ended up spilling the keys twice. If we do implement it again we would need to be very specific about only doing this on the stream side.
3) I am fairly sure that it introduced a bug in the filter code. I'm not sure anyone would ever hit it because you would have to do an inner equality join with a conditional, where the condition included one of the equality keys on the left hand column and something from the right hand side too.

So instead of trying to fix all of that, I thought I would just remove it for now and we can come back to look at it in the future if we need a bit more optimization.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
@@ -177,11 +177,10 @@ object GpuHashJoin extends Arm {
  * An iterator that does a hash join against a stream of batches.
  */
 class HashJoinIterator(
-    inputBuiltKeys: ColumnarBatch,
-    inputBuiltData: ColumnarBatch,
+    builtInput: ColumnarBatch,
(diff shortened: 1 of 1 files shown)


revans2:
build

revans2:
I ran some performance tests on query 14a and 24a. The difference in performance appears to be within the run to run noise.

jlowe:
> The difference in performance appears to be within the run to run noise.

This is not only about the time spent gathering the redundant keys but also the memory spent holding the redundant keys, correct?

PR #2392, opened 2021-05-11
Closed unclosed views to avoid memory leak

### LINKED ISSUE #2355: [BUG] convertDecimal64ToDecimal32Wrapper leaks ColumnView instances
`com.nvidia.spark.rapids.ParquetPartitionReaderBase#convertDecimal64ToDecimal32Wrapper` instantiates `ColumnView` instances without closing them.  For example, `child` and `newChild` are both `ColumnView` instances and neither are closed.
```scala
      } else if (dt == DType.LIST) {
        val child = cv.getChildColumnView(0)
        val newChild = convertDecimal64ToDecimal32(child, precision, newCols)
        if (child == newChild) {
          cv
        } else {
          cv.replaceListChild(newChild)
        }
      } else if (dt == DType.STRUCT) {
```
Each `ColumnView` corresponds to a native structure that needs to be closed to release the native memory.

razajafri:
fixes #2355 

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
@@ -737,7 +737,7 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
     def convertDecimal64ToDecimal32(
        cv: ColumnView,
        precision: Int,
-       toClose: ArrayBuilder[ColumnView]): ColumnView = {
+       toClose: ArrayBuffer[ColumnView]): ColumnView = {
       val dt = cv.getType
       if (!dt.isNestedType) {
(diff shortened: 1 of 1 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -725,10 +725,19 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
...
+     * @param toClose     The ColumnViews to be closed after this method is done. This will also
+     *                    contain the returned view
+     */
     def convertDecimal64ToDecimal32(
-        cv: ColumnView,
-        precision: Int,
-        newCols: ArrayBuilder[ColumnView]): ColumnView = {
+       cv: ColumnView,
+       precision: Int,
+       toClose: ArrayBuilder[ColumnView]): ColumnView = {
```
jlowe, line 740: Curious why `ArrayBuilder` instead of `ArrayBuffer` here?  Typically we've used `ArrayBuffer`.

jlowe (changes requested):

Issue #2315, opened 2021-04-30
[BUG] legacy struct cast to string crashes on a two field struct


gerashegalov:
**Describe the bug**
In Spark 3.0.x or in Spark 3.1.+ with `spark.sql.legacy.castComplexTypesToString.enabled=true` queries from an RDD source may crash with

```
java.lang.AssertionError:  value at 15 is null
        at ai.rapids.cudf.HostColumnVectorCore.assertsForGet(HostColumnVectorCore.java:228)
        at ai.rapids.cudf.HostColumnVectorCore.getUTF8(HostColumnVectorCore.java:355)
        at com.nvidia.spark.rapids.RapidsHostColumnVectorCore.getUTF8String(RapidsHostColumnVectorCore.java:177)
        at org.apache.spark.sql.vectorized.ColumnarBatchRow.getUTF8String(ColumnarBatch.java:220)
        at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.apply(Unknown Source)
        at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.apply(Unknown Source)
        at scala.collection.Iterator$$anon$10.next(Iterator.scala:459)
        at org.apache.spark.sql.execution.SparkPlan.$anonfun$getByteArrayRdd$1(SparkPlan.scala:346)
        at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2(RDD.scala:898)
        at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2$adapted(RDD.scala:898)
        at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
        at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:373)
        at org.apache.spark.rdd.RDD.iterator(RDD.scala:337)
```

**Steps/Code to reproduce bug**
Minimum repro:
```python
import sys
sys.path.append(r'/path/to/localgit/spark-rapids/integration_tests/src/main/python')

import pyspark.sql.functions as F
import pytest
from data_gen import *

key_data_gen = StructGen([
        ('a', IntegerGen(min_val=0, max_val=4)),
        ('b', IntegerGen(min_val=5, max_val=9)),
    ], nullable=False)
val_data_gen = IntegerGen()
df = two_col_df(spark, key_data_gen, val_data_gen)

[...]

PR #2395, opened 2021-05-11
Unify legacy and 3.1.x struct cast implementations

### LINKED ISSUE #2309: [BUG] legacy cast of a struct column to string with a single nested null column yields null instead of '[]'
**Describe the bug**
In Spark 3.0.x and Spark 3.1+ with the legacy mode, casting a struct with a single nested null-valued field produces a null string instead of `[]`

```python
import pyspark.sql.functions as F
import pyspark.sql.types as T

# for Spark 3.1+
spark.conf.set('spark.sql.legacy.castComplexTypesToString.enabled', True)

data = [ (('gera',),), ((None,),), (None,) ]
df = spark.createDataFrame(data)
df.select(df._1.cast(T.StringType())).collect()
```
Broken output:
```
[Row(_1='[gera]'), Row(_1=None), Row(_1=None)]
```

**Expected behavior**
Correct output:
```
[Row(_1='[gera]'), Row(_1='[]'), Row(_1=None)]
```
Reproduces in a local REPL 

**Additional context**
CI test failures in #2274

gerashegalov:
Refactors struct cast to string such that there no need for a dedicated method handling the legacy mode cast. Fixes #2309 and #2315
    
### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
@@ -529,8 +529,7 @@ case class GpuCast(
       spaceColumn: ColumnVector,
       leftColumn: ColumnVector,
       rightColumn: ColumnVector) = {
-      val columns = ArrayBuffer.empty[ColumnVector]
-      try {
+      withResource(ArrayBuffer.empty[ColumnVector]) { columns =>
         // legacy: [firstCol
         //   3.1+: {firstCol
         columns += leftColumn.incRefCount()
@@ -555,8 +554,6 @@ case class GpuCast(
         withResource(ColumnVector.stringConcatenate(emptyScalar, nullScalar, columns.toArray))(
           _.mergeAndSetValidity(BinaryOp.BITWISE_AND, input) // original whole row is null
         )
-      } finally {
-        columns.safeClose()
       }
     }
 
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
@@ -271,7 +271,7 @@ case class GpuCast(
       case (TimestampType, StringType) =>
         castTimestampToString(input)
       case (StructType(fields), StringType) =>
-        castStructToString(input, legacyCastToString, fields)
+        castStructToString(input, fields)
 
       // ansi cast from larger-than-integer integral types, to integer
       case (LongType, IntegerType) if ansiMode =>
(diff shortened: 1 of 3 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -510,126 +510,71 @@ case class GpuCast(
...
-  private def castStructToString(input: ColumnView,
-    legacyCastToString: Boolean, inputSchema: Array[StructField]): ColumnVector = {
-    if (legacyCastToString) {
-      legacyStructToString(input, inputSchema)
-    } else {
-      modernStructToString(input,inputSchema)
+    withResource(Array(
+      Scalar.fromString(emptyStr),
+      Scalar.fromString(nullStr),
+      Scalar.fromString(separatorStr),
```
jlowe, line 568: This will leak GPU resources if any of these `fromString` methods throw after the first.  Something like this would cover that possibility:
```scala
  withResource(Seq(emptyStr, nullStr, separatorStr, spaceStr, leftStr, rightStr).safeMap(Scalar.fromString).toArray) { ...
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -510,126 +510,71 @@ case class GpuCast(
...
+      Scalar.fromString(spaceStr),
+      Scalar.fromString(leftStr),
+      Scalar.fromString(rightStr)
+    )) { case Array(emptyScalar, nullScalar, columnScalars@_*) =>
+      withResource(columnScalars.map(s => ColumnVector.fromScalar(s, numRows)))(columns => {
```
jlowe, line 573: If the `map` throws while trying to build the sequence then we can leak
```suggestion
      withResource(columnScalars.safeMap(s => ColumnVector.fromScalar(s, numRows)))(columns => {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
jlowe: This isn't trying to close the strings but the Scalar instances constructed from those strings.  Note the `safeMap` after the `Seq`, so we're not trying to close a `Seq[String]` but a `Seq[Scalar]`.

gerashegalov:
build

jlowe (approved):
Small nit but otherwise lgtm

jlowe (changes requested):
Actually now that #2274 has been merged, we should pull out the xfail markers in those tests to work around the bugs being fixed here.

jlowe (approved):

jlowe:
build

PR #2397, opened 2021-05-11
Include memory access costs in cost models (cost-based optimizer)

andygrove:

This PR starts to introduce the mechansim for including memory access costs in the cost model. The default settings are 320 GB/s for GPU and 30 GB/s for CPU.

This also contains a fix for a bug that was highlighted during testing where CBO could move a BroadcastExchange back to CPU but leave the BHJ on GPU, resulting in an invalid plan.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
@@ -21,10 +21,9 @@ import scala.collection.mutable.ListBuffer
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.expressions.{Alias, AttributeReference, Expression, GetStructField}
 import org.apache.spark.sql.catalyst.plans.{JoinType, LeftAnti, LeftSemi}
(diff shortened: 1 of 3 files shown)


revans2 (dismissed):

revans2:
> @revans2 This is what I have so far for incorporating memory access costs. It's very early and I have plenty to fix already but wanted to get your feedback early on to make sure I am heading in the right direction. I will handle kernel costs in a separate PR.

Jason had a good point when we talked that it is not really kernel costs as much as it is a fixed cost per operation. The cost of launching a kernel is likely small compared to other parts of the computation.

andygrove:
@revans2 This is ready for review now.

andygrove:
I filed a follow-on issue for improving our handling of BHJ: https://github.com/NVIDIA/spark-rapids/issues/2417

revans2 (dismissed):
This looks fine to me, but I would like to hear from others

andygrove:
I filed https://github.com/NVIDIA/spark-rapids/issues/2430 for adding kernel launch costs

andygrove:
build

andygrove:
Rebased to pull in fix for test failure

andygrove:
build

andygrove:
@revans2 Any objection to me merging this now so we can see the impact in the automated benchmarks?

revans2 (approved):

Issue #2390, opened 2021-05-11
[FEA] CI/CD for databricks 8.2 runtime


tgravescs:
**Is your feature request related to a problem? Please describe.**
Adding support for databricks 8.2 here: https://github.com/NVIDIA/spark-rapids/pull/2381

We need to setup ci/cd and make sure the databricks jar gets pushed first before we run the nightly that would pull it.

Also comment in review: https://github.com/NVIDIA/spark-rapids/pull/2381#discussion_r629690099
to potentially clean up the pom files.

PR #2405, opened 2021-05-13
Activate databricks 8.2 runtime jar being included in nightly and use property instead of profile

tgravescs:
All the builds are setup and jars for the new databricks 8.2 runtime are pushed, so now activate including both databricks runtimes in the nightly jar. Also change include-databricks to be a property so we don't duplicate code.

I also had a type in the include-databricks-301.. the final dash shouldn't have been there.

This should be last PR for enabling all the CI for the databricks 8.2 runtime.

fixes https://github.com/NVIDIA/spark-rapids/issues/2390


tgravescs:
build

tgravescs:
@jlowe I think this is what your comment meant by using properties, if not let me know

jlowe (approved):

Issue #1588, opened 2021-01-26
Rapids shuffle - UCX active messages


abellina:
JUCX is adding active messages to its API (https://github.com/openucx/ucx/pull/6155). Active messages are a big change but they should remove the need for tag management, and bring also stability improvements (with some error handling already done for active messages).

This task is for the transport to change to use this, instead of tags, as a proof-of-concept at first, but document here findings and a likely PR once JUCX 1.11 is published.

Issue #980, opened 2020-10-19
[FEA] flatbuffer shuffle metadata limit


abellina:
We currently have hard limits on shuffle metadata limits in the Rapids Shuffle Manager, which are required to be negotiated ahead of time since we currently use UCX for metadata exchange.

If limits are reached, or exceeded, we may need to chunk up a request into multiple ones. This is to handle such cases:

1. Metadata request (block info): normally when metadata response is large (many tables/columns/etc), but could happen in a case where we have many partitions on the request side as well.
2. Transfer request (block transfer initiation): if we have lots of partitions, the transfer request message could get large as well. This is an easier fix => send multiple transfer requests. But in some cases that seems sub-optimal (what if could handle all those tiny transfers with 1 bounce buffer).

Either case can be worked around by setting a higher ceiling for metadata limit, but that's not a good user experience.

Issue #2275, opened 2021-04-27
[FEA] UCP error callback deal with cleanup


abellina:
With https://github.com/NVIDIA/spark-rapids/pull/2131, we now have a way to get a UCP error handler callback. This callback is currently setup to close and remove the endpoint from the list of endpoints.

We need to clean up the state regarding those endpoints, as we'll have requests pending for them, and we'll have clients who depend on them. 

This task is to figure out what needs to happen... and how to test.

abellina:
Note comment from @jlowe should be handled with this issue:

> Is the explicit predicate function really necessary or is this equivalent to: endpoints.values().remove(ucpEndpoint)

PR #2409, opened 2021-05-13
Change shuffle metadata messages to use UCX Active Messages

abellina:
Closes https://github.com/NVIDIA/spark-rapids/issues/1588
Closes https://github.com/NVIDIA/spark-rapids/issues/980

Please set whitespace off when reviewing the bulk of this, it makes it easier to understand. Also there are changes here to the flatbuffer java files, and those are generated. I did do some minor `withResources` cleanup, can split that up into another PR.

This PR changes the metadata/transfer request messaging from UCX tags to UCX Active Messages. The advantage of Active Messages is that error handling in UCX is going this way, and when things go wrong Active Messages will clean up resources that UCX is using (like fragments from their pools for pipelined protocols). The other big one is that it removes the need to worry about the max metadata size. I haven't refactored the metadata pool yet to make it remove this setting alltogether, can do as a follow up.

It was tested with UCX 1.10.x, and 1.11. With UCX 1.10.x the RNDV mode is the only mode supported (Active Messages could also be EAGER, and they have an automatic mode to switch based on size), so I set "rndv" set as default. The config is internal, but I should say also there doesn't seem to be a big need for EAGER mode right now, not from the 3TB tests I've run.

I had to tweak client/server/iterator tests here, and they all pass. I made some changes and need to retest in a cluster setup, but from my testing at 3TB I see no performance regressions. There is some bit manipulation I'll write tests for in this PR.


abellina:
If anyone would like me to walk them throw the code, I am happy to do so.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -903,6 +903,12 @@ object RapidsConf {
       .bytesConf(ByteUnit.BYTE)
       .createWithDefault(1024 * 1024 * 1024)
 
+  val SHUFFLE_UCX_ACTIVE_MESSAGES_MODE = conf("spark.rapids.shuffle.ucx.activeMessages.mode")
+    .doc("UCX Active Messages can be RNDV, EAGER, or automatic. We set RNDV by default " +
+      "because version 1.10.x of UCX does not well with EAGER modes (or automatic).")
```
jlowe, line 908: "does not well"

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -903,6 +903,12 @@ object RapidsConf {
       .bytesConf(ByteUnit.BYTE)
       .createWithDefault(1024 * 1024 * 1024)
 
+  val SHUFFLE_UCX_ACTIVE_MESSAGES_MODE = conf("spark.rapids.shuffle.ucx.activeMessages.mode")
+    .doc("UCX Active Messages can be RNDV, EAGER, or automatic. We set RNDV by default " +
```
jlowe, line 907: Should "automatic" be capitalized in this description for consistency with the others?

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -27,20 +27,51 @@ import scala.util.Random
 
 import ai.rapids.cudf.{MemoryBuffer, NvtxColor, NvtxRange}
 import com.google.common.util.concurrent.ThreadFactoryBuilder
-import com.nvidia.spark.rapids.{GpuDeviceManager, RapidsConf}
-import com.nvidia.spark.rapids.shuffle.{AddressLengthTag, ClientConnection, MemoryRegistrationCallback, TransportUtils}
+import com.nvidia.spark.rapids.{Arm, GpuDeviceManager, RapidsConf}
+import com.nvidia.spark.rapids.shuffle.{AddressLengthTag, ClientConnection, MemoryRegistrationCallback, , TransportUtils}
```
jlowe, line 31: Something weird happened to the import list here.
```suggestion
import com.nvidia.spark.rapids.shuffle.{AddressLengthTag, ClientConnection, MemoryRegistrationCallback, TransportUtils}
```

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +374,217 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+
+  def sendActiveMessage(endpointId: Long, am: UCXActiveMessage,
+                        dataAddress: Long, dataSize: Long, cb: UcxCallback): Unit = {
+    require(am.header.isDefined,
+      "An Active Message request must include a header so the response can find the " +
+        "correct callback.")
+    onWorkerThreadAsync(() => {
+      val ep = endpoints.get(endpointId)
+      if (ep == null) {
+        throw new IllegalStateException(s"Trying to send a message to an endpoint that doesn't exist ${endpointId}")
```
jlowe, line 556: scalastyle line length exceeded

jlowe (changes requested):

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
abellina: Yeah not sure how that's building.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
abellina: I reworded this doc in latest commit.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +374,218 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+      case "rndv" =>
+        UcpConstants.UCP_AM_SEND_FLAG_RNDV
+      case "auto" =>
+        0L
+    }
+  }
+
+  def sendActiveMessage(endpointId: Long, am: UCXActiveMessage,
+                        dataAddress: Long, dataSize: Long, cb: UcxCallback): Unit = {
+    require(am.header.isDefined,
```
jlowe, line 550: Nit: Wonder if it makes sense to prevent this from occurring.  Looks like this only needs the message ID and the header from the `UCXActiveMessage` (Oddly no callbacks associated with it are used? I continue to be baffled by what UCXActiveMessage really means.). If the message ID and header are passed separately then it forces the caller to produce the header at the call site rather than passing an AM that may or may not have one.

Definitely not must-fix, just a thought.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -608,6 +834,12 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
 
   override def close(): Unit = {
     onWorkerThreadAsync(() => {
+      amRegistrations.forEach { (activeMessageId, _) =>
+        logInfo(s"Removing Active Message registration for " +
```
jlowe, line 838: Do we want this at info or debug level?

jlowe (commented):
@abellina this is the first pass through UCX.scala still have the other files to go.

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleClientSuite.scala
```diff
@@ -98,6 +98,8 @@ class RapidsShuffleClientSuite extends RapidsShuffleTestHelper {
       val tm = ptrs(t).tableMeta
       verifyTableMeta(expected, tm)
     }
+
+    assertResult(true)(response.isClosed)
```
jlowe, line 102:
```suggestion
    assert(response.isClosed)
```

Applies to other occurrences in this file as well.

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleServerSuite.scala
```diff
@@ -120,18 +121,21 @@ class RapidsShuffleServerSuite extends RapidsShuffleTestHelper {
       bounceBuffer
     }
     assertResult(true)(bb.deviceBounceBuffer.isClosed)
+    assertResult(true)(transferRequest.isClosed)
```
jlowe, line 124:
```suggestion
    assert(transferRequest.isClosed)
```
Applies elsewhere in this file.

sql-plugin/src/main/format/ShuffleTransferResponse.fbs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020, NVIDIA CORPORATION.
+// Copyright (c) 2020-2021, NVIDIA CORPORATION.
```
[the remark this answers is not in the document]
abellina: good catch, undoing.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -903,6 +903,13 @@ object RapidsConf {
       .bytesConf(ByteUnit.BYTE)
       .createWithDefault(1024 * 1024 * 1024)
 
+  val SHUFFLE_UCX_ACTIVE_MESSAGES_MODE = conf("spark.rapids.shuffle.ucx.activeMessages.mode")
+    .doc("Set to 'rndv', 'eager', or 'auto' to indicate what UCX Active Message mode to " +
+      "use. We set 'rndv' (Rendezvous) by default because UCX 1.10.x doesn't support 'eager' " +
+      "fully.  This restriction can be lifted if the user is running UCX 1.11+.")
+    .stringConf
+    .createWithDefault("rndv")
```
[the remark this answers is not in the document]
abellina: Better than than nothing. Then the matching piece will take care of it dynamically.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCXTransaction.scala
```diff
@@ -360,5 +368,85 @@ private[ucx] class UCXTransaction(conn: UCXConnection, val txId: Long)
   }
 
   var callbackCalled: Boolean = false
+
+  var activeMessageData: Option[RefCountedDirectByteBuffer] = None
```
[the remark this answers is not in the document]
abellina: done.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCXTransaction.scala
```diff
@@ -360,5 +368,85 @@ private[ucx] class UCXTransaction(conn: UCXConnection, val txId: Long)
...
+
+  override def respond(response: ByteBuffer,
```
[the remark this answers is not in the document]
abellina: I have a function header in the interface I am overriding. Should I add it here too?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleServer.scala
```diff
@@ -388,47 +343,54 @@ class RapidsShuffleServer(transport: RapidsShuffleTransport,
...
                   bufferSendState.close()
                   bssExec.notifyAll()
                 }
               }
-            } finally {
-              bufferTx.close()
             })
       }
-    }
+      }
```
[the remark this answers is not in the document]
abellina: fixed

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCXConnection.scala
```diff
@@ -40,21 +45,98 @@ private[ucx] abstract class UCXTagCallback {
...
+                       response: ByteBuffer,
+                       cb: TransactionCallback): Transaction = {
+
+    val tx = createTransaction
+    tx.start(UCXTransactionType.Request, 1, cb)
+
+    logDebug(s"Responding to ${peerExecutorId} at ${TransportUtils.toHex(header)} " +
+      s"with ${response}")
+
+    val responseAm = UCXActiveMessage(composeResponseAmId(messageType), Option(header))
```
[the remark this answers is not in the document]
abellina: I picked `Option` because @gerashegalov has beating `Some` out of me, this being the main point:
```
scala> Option(null)
res2: Option[Null] = None

scala> Some(null)
res3: Some[Null] = Some(null)
```
What do you think @jlowe if we keep `Option`?

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -124,10 +161,11 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
   // Error handler that would be invoked on endpoint failure.
   private val epErrorHandler = new UcpEndpointErrorHandler {
     override def onError(ucpEndpoint: UcpEndpoint, errorCode: Int, errorString: String): Unit = {
-      endpoints.values().removeIf(ep => ep == ucpEndpoint)
-      ucpEndpoint.close()
-      if (errorCode != UcsConstants.STATUS.UCS_ERR_CONNECTION_RESET) {
-        logError(s"Endpoint to $ucpEndpoint got error: $errorString")
+      withResource(ucpEndpoint) { _ =>
+        if (errorCode != UcsConstants.STATUS.UCS_ERR_CONNECTION_RESET) {
+          logError(s"Endpoint to $ucpEndpoint got error: $errorString")
```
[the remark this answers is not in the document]
abellina: Yeap it does. Separate PR if that's OK. I had filed this before to propagate the error: https://github.com/NVIDIA/spark-rapids/issues/2275

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -124,10 +161,11 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+        }
+        endpoints.values().removeIf(ep => ep == ucpEndpoint)
```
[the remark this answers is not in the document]
abellina: this is behind flags right now and is disabled by default. I'll tackle with the #2275, if that's OK.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCXConnection.scala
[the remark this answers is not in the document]
jlowe: In all the cases I marked, it is known not to be `null`.  It's not must-fix from my perspective, just an unnecessary use of `Option`.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCXTransaction.scala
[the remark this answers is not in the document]
jlowe: No, that's fine, we can use the interface doc.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -27,20 +27,51 @@ import scala.util.Random
...
+import org.apache.spark.sql.rapids.storage.RapidsStorageUtils
 import org.apache.spark.storage.BlockManagerId
 
 case class WorkerAddress(address: ByteBuffer)
 
 case class Rkeys(rkeys: Seq[ByteBuffer])
 
+/**
+ * A simple wrapper specifically for an Active Message Id and a small (Long) header
+ * we are using per message.
```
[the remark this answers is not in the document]
abellina: Simplified this one, and updated comment.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +374,217 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+        "Attempting to set an Active Message response handler without a callback")
+      require(singleCallback == null)
+      perHeaderCallbacks.put(am.header.get, am.cb.get)
+    }
+
+    private var singleCallback: () => UCXAmCallback = null
+    private val perHeaderCallbacks = new ConcurrentHashMap[Long, UCXAmCallback]()
+  }
+
+  private val amRegistrations = new ConcurrentHashMap[Int, ActiveMessageRegistration]()
```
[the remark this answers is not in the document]
abellina: done

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +374,217 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+    amRegistrations.computeIfAbsent(activeMessageId,
+      _ => {
+        val reg = new ActiveMessageRegistration(activeMessageId)
+        reg.setSingleActiveMessageHandler(activeMessage)
+        registerActiveMessage(activeMessageId, reg)
+        reg
+      })
+  }
+
+  private def registerActiveMessage(activeMessageId: Int, reg: ActiveMessageRegistration): Unit = {
```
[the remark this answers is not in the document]
abellina: nice catch.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
abellina: There should be 4 of these. But yeah I'll move to debug.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +374,217 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+  /**
+   * Register a request handler (the server will use this)
+   * @note This function will be called once for the server for an `activeMessageId`
+   * @param activeMessage
+   */
+  def registerRequestHandler(activeMessage: UCXActiveMessage): Unit = {
+    logDebug(s"Register Active Message $activeMessage request handler")
+    val activeMessageId = activeMessage.activeMessageId
+
+    require(!amRegistrations.containsKey(activeMessageId),
```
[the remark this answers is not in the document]
abellina: fixed

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +374,218 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+
+  private def activeMessageMode: Long = {
+    rapidsConf.shuffleUcxActiveMessagesMode match {
+      case "eager" =>
+        UcpConstants.UCP_AM_SEND_FLAG_EAGER
+      case "rndv" =>
+        UcpConstants.UCP_AM_SEND_FLAG_RNDV
+      case "auto" =>
+        0L
+    }
```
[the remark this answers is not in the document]
abellina: fixed

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
abellina: Consider it done. Hopefully `UCXActiveMessage` makes more sense now.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -27,20 +27,51 @@ import scala.util.Random
...
 
 case class Rkeys(rkeys: Seq[ByteBuffer])
 
+/**
+ * A simple wrapper specifically for an Active Message Id and a small (Long) header
+ * we are using per message.
+ */
+case class UCXActiveMessage(activeMessageId: Int,
+                            header: Option[Long] = None,
+                            cb: Option[UCXAmCallback] = None,
```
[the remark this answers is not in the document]
abellina: removed callbacks from `UCXActiveMessage`. Now it's just id and header.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +374,217 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+  }
+
+  private val amRegistrations = new ConcurrentHashMap[Int, ActiveMessageRegistration]()
+
+  /**
+   * Register a response handler (clients will use this)
+   * @note This function will be called for each client, with the same `am.activeMessageId`
+   * @param activeMessage - An Active Message to register response interest for
+   */
+  def registerResponseHandler(activeMessage: UCXActiveMessage): Unit = {
```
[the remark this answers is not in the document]
abellina: Changed these signatures.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +374,217 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+        reg
+      })
+
+    reg.addActiveMessageHandler(activeMessage)
+  }
+
+  /**
+   * Register a request handler (the server will use this)
+   * @note This function will be called once for the server for an `activeMessageId`
+   * @param activeMessage
```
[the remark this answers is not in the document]
abellina: done.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +362,235 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+   *   - On a request, a callback is instantiated using `requestCallbackGen`, which creates
+   *   a transaction each time. This is one way to handle several requests inbound to a server.
+   *
+   * For responses:
+   *   - `activeMessageId` for responses is the value of the `RequestType` enum with an extra
+   *   bit flipped (see `UCXConnection.composeResponseAmId`). These are also set once, as requests
+   *   are sent out.
+   *
+   *   - The Active Message header is used to pick the correct callback to call. In this case
+   *   there could be serveral expected responses, for a single response `activeMessageId`, so the
```
jlowe, line 389:
```suggestion
   *   there could be several expected responses, for a single response `activeMessageId`, so the
```

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +374,218 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+            // RNDV case: we get a direct buffer and UCX will fill it with data at `receive`
+            // callback
+            val resp = cb.onHostMessageReceived(amData.getLength)
+
+            val receiveAm = amData.receive(UcxUtils.getAddress(resp.getBuffer()),
+              new UcxCallback {
+                override def onError(ucsStatus: Int, errorMsg: String): Unit = {
+                  withResource(resp) { _ =>
+                    withResource(amData) { _ =>
+                      logError(s"Error receiving: $ucsStatus $errorMsg => $activeMessageId ${hdr}")
```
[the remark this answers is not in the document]
abellina: Taken care of. Will let the caller deal with the error, rather than continue down the `log-then-error` path.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +362,235 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+   *   are sent out.
+   *
+   *   - The Active Message header is used to pick the correct callback to call. In this case
+   *   there could be serveral expected responses, for a single response `activeMessageId`, so the
+   *   server echoes back our header so we can invoke the correct response callback.
+   *
+   *   - Each response received at the response activeMessageId, will be demuxed using the header:
+   *   responseActiveMessageId1 -> [callbackForHeader1, callbackForHeader2, ..., callbackForHeaderN]
+   */
+  private class ActiveMessageRegistration(val activeMessageId: Int) {
```
jlowe, line 395: This feels like a class that is trying to be two different things at once, which makes me wonder if there should be an interface like this:
```scala
trait ActiveMessageRegistration {
  val activeMessageId
  def getCallback(header: Long): UCXAmCallback
}
```
And then there's two types that implement this interface:
- a request registration that takes the callback generator as a constructor argument
- a response registration that provides the `addResponseActiveMessageHandler` method

Not must-fix, could also be a followup PR to clean it up, but all the "make sure I'm the right type of instance" checks in the methods makes me wonder if it would be cleaner to make the registration types explicit.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +362,235 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+    rapidsConf.shuffleUcxActiveMessagesMode match {
+      case "eager" =>
+        UcpConstants.UCP_AM_SEND_FLAG_EAGER
+      case "rndv" =>
+        UcpConstants.UCP_AM_SEND_FLAG_RNDV
+      case "auto" =>
+        0L
+      case _ =>
+        throw new IllegalArgumentException(
+          s"${rapidsConf.shuffleUcxActiveMessagesMode} is an invalid Active Message mode. ")
```
jlowe, line 552: This message needs to mention `RapidsConf.SHUFFLE_UCX_ACTIVE_MESSAGES_MODE.key` so the user knows this came from an invalid setting of that config.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCXConnection.scala
```diff
@@ -40,21 +45,98 @@ private[ucx] abstract class UCXTagCallback {
...
+
+      tx.start(UCXTransactionType.Request, 1, cb)
+
+      override def onSuccess(am: UCXActiveMessage, buff: RefCountedDirectByteBuffer): Unit = {
+        logDebug(s"At requestHandler for ${requestType} and am: " +
+          s"$am")
+        tx.completeWithSuccess(requestType, Option(am.header), Option(buff))
+      }
+
+      override def onHostMessageReceived(size: Long): RefCountedDirectByteBuffer = {
```
petro-rudenko, line 88: I would consider initially handle both metadata and data buffers with AM. Maybe split to some public API, that would handle both cases. Even left `onDeviceMessageReceived` unimplemented, but to see the whole picture. Trying to falling my or with AM: https://github.com/NVIDIA/spark-rapids/pull/2416 - want it to work with both GPU and host memory

abellina: Should RNDV for GPU buffers work in 1.10 for Active Messages? Just checking, I have done mostly host buffers.

I would really like to add GPU support as an different PR, need to discuss with others as well on the timing.

abellina:
@jlowe I think I have addressed your comments. Could you take another look?

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -345,6 +362,258 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
+   *   responseActiveMessageId1 -> [callbackForHeader1, callbackForHeader2, ..., callbackForHeaderN]
+   */
+  trait ActiveMessageRegistration {
+    val activeMessageId: Int
+    def getCallback(header: Long): UCXAmCallback
+  }
+
+  class RequestActiveMessageRegistration(override val activeMessageId: Int)
+      extends ActiveMessageRegistration {
+    private var requestCallbackGen: () => UCXAmCallback = null
```
jlowe, line 402: Why isn't this a constructor argument?  There's only one place that sets it, and that's immediately after constructing the instance.  Is there a valid reason to have an instance where this remains null for any significant duration after construction?

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCXConnection.scala
```diff
@@ -124,46 +230,72 @@ class UCXConnection(peerExecutorId: Int, ucx: UCX) extends Connection with Loggi
...
+    }
+    amId
+  }
+
+  protected def composeRequestHeader(executorId: Long, txId: Long): Long = {
+    composeTag(executorId << 32, txId)
+  }
+
+  def extractExecutorId(header: Long): Long = {
+    (header & upperBitsMask) >> 32
```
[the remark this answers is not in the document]
abellina: I made some changes to this to shift-then-mask (https://github.com/NVIDIA/spark-rapids/pull/2409/commits/42397f8e0101ee389c6667b658bfa35d5f77eb2a). I also added some requires around the sign of executorId, since it better not be negative.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
abellina: Making the change, that will be cleaner for sure.

abellina: done.

abellina: whoops trigger happy.. one sec

jlowe (approved):

abellina:
build

Issue #2408, opened 2021-05-13
[FEA] Split stream side join batches into smaller sizes if the join explodes


revans2:
We have started to work on joins that are too large to fit in GPU memory. The first step in this was to get back the gather maps from the join instead of materializing the results directly, and then be able to output the results in batches that are the appropriate size. For the most part this is working really well (All of the NDS queries except for 1 we are able to run at scale factor 3000). For query 72, however, it is not great. In a typical run we are seeing a join that really explodes. When joining a 600MB table by a 400MB table we see the join try to allocate over 260GB just for the gather map. (The row count on average will increase 2,000x).

We run into issues in part because this alloc request is so large that we end up spilling a lot of data to host memory, or even disk, when we could avoid doing it if we knew the join would explode, so we could split the input batches up.

Part of the problem is that this is an equi-join with an inequality join as well. So the 2,000x increase in size is before filtering. After the filter it is only a 200x explosion in size. So the AST work for Joins in CUDF will be a huge win for us. But even then the gather map for this join would be 26GB, which is larger than can fit on a typical 16GB GPU.

The idea is to take https://github.com/rapidsai/cudf/issues/8237 and check the size against a budget for the gather map. If the size exceeds that budget, then we use the result to be able to cut the input batches down into smaller pieces that should fit.

Note that for Left-semi and Left-anti joins the gather map will never be larger than the number of rows in the left table, and they will always build on the right side. So with that we already know our limit, it should be reasonable, so this feature does not apply here.

For full outer joins we need complete sets of keys for both sides.  With that we would need to do something more complicated. Probably starting with a sort merge join #2354 and then if we have a customer where the join is still exploding we would need to slice it up into smaller chunks that honor key boundaries.

revans2:
Will file a separate follow on for anything missing after #2433 goes in.

PR #2411, opened 2021-05-13
Allow inner joins to fall back to the CPU if the gather map is too large

revans2:
This is a stop gap until #2408 can be implemented.  This is primarily to help get NDS query 72 to not crash because the join explodes.  This code runs and I have manually tested it on NDS Query 72 at scale factor 200, but I have not written any automated tests for it. Hopefully #2408 comes quickly and we can rip all of this out and put in better tests, but if people really want me to add in some kind of test for this I can, but it will likely involve putting in a special config for testing that will cause inner joins to throw an OutOfMemoryError to trigger the fallback.

The metrics are not perfect when this falls back to the CPU. Again if people want me to get the metrics to work well I can do it, but it will take a little time. The join output rows metric, however, is not something that I will be able to make work the same as on the GPU code and probably will go away once we have AST joins working once cudf supports it.

The performance of this is not great.  For NDS query 72 scale factor 200 on Spark 3.0.2 I was able to run it end to end on the CPU in 14 mins (12 cores). The same query on the GPU would crash before this, but after this it now takes between 38 mins and an hour depending on how you configure batch size, temperature of the GPU,.... The slower runs are due to spilling on larger batch sizes because I had all 12 cores enabled.  Another part of the reason for this is that this patch does not use code generation for the join.  Spark supports it for this use case, but we do not because of the added complexity involved. To put that in perspective I was able to run the rest of NDS at scale factor 200 on the GPU in under 35 mins.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/execution/joins/rapids/CpuSortMergeJoinFallback.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/execution/joins/rapids/CpuSortMergeJoinFallback.scala
@@ -1,0 +1,170 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.execution.joins.rapids
+
+import com.nvidia.spark.rapids.{ColumnarToRowIterator, GpuOutOfCoreSortIterator, RowToColumnarIterator}
+import com.nvidia.spark.rapids.RapidsBuffer.SpillCallback
+
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.{Ascending, Attribute, Expression, JoinedRow, Predicate, Projection, RowOrdering, SortOrder, UnsafeProjection, UnsafeRow}
+import org.apache.spark.sql.catalyst.plans.{InnerLike, JoinType}
+import org.apache.spark.sql.execution.{ExternalAppendOnlyUnsafeRowArray, RowIterator}
+import org.apache.spark.sql.execution.joins.SortMergeJoinScanner
+import org.apache.spark.sql.vectorized.ColumnarBatch
+
+class CpuSortMergeJoinFallback(
+    val joinType: JoinType,
+    val left: Iterator[ColumnarBatch],
+    val leftSchema: Seq[Attribute],
+    val unboundLeftGpuKeys: Seq[Expression],
+    val unboundLeftCpuKeys: Seq[Expression],
+    val right: Iterator[ColumnarBatch],
+    val rightSchema: Seq[Attribute],
+    val unboundRightGpuKeys: Seq[Expression],
+    val unboundRightCpuKeys: Seq[Expression],
+    val cpuCondition: Option[Expression],
+    val outputSchema: Seq[Attribute],
+    private val targetSize: Long,
+    private val spillThreshold: Int,
+    private val inMemoryThreshold: Int,
+    private val spillCallback: SpillCallback,
+    private val cleanupResources: () => Unit) extends Iterator[ColumnarBatch] {
+
+  private val leftGpuOrder = unboundLeftGpuKeys.map(SortOrder(_, Ascending))
+  private val sortedLeft = GpuOutOfCoreSortIterator.simple(left, leftGpuOrder, leftSchema,
+    targetSize, spillCallback)
+  // This is an ugly hack to avoid data not being spillable when we fall back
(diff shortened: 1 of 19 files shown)


revans2:
build

sql-plugin/src/main/scala/org/apache/spark/sql/execution/joins/rapids/CpuSortMergeJoinFallback.scala
```diff
@@ -0,0 +1,170 @@
...
+
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.{Ascending, Attribute, Expression, JoinedRow, Predicate, Projection, RowOrdering, SortOrder, UnsafeProjection, UnsafeRow}
+import org.apache.spark.sql.catalyst.plans.{InnerLike, JoinType}
+import org.apache.spark.sql.execution.{ExternalAppendOnlyUnsafeRowArray, RowIterator}
+import org.apache.spark.sql.execution.joins.SortMergeJoinScanner
+import org.apache.spark.sql.vectorized.ColumnarBatch
+
+class CpuSortMergeJoinFallback(
+    val joinType: JoinType,
```
jlowe, line 30: Wow, this is a lot of parameters with a lot of the same types.  Keeping the call sites ordered properly will be challenging, good thing there aren't that many sites.

Should this take the `CpuJoinInfo` directly rather than as separate parameters?

sql-plugin/src/main/scala/org/apache/spark/sql/execution/joins/rapids/CpuSortMergeJoinFallback.scala
```diff
@@ -0,0 +1,170 @@
...
+    val rightSchema: Seq[Attribute],
+    val unboundRightGpuKeys: Seq[Expression],
+    val unboundRightCpuKeys: Seq[Expression],
+    val cpuCondition: Option[Expression],
+    val outputSchema: Seq[Attribute],
+    private val targetSize: Long,
+    private val spillThreshold: Int,
+    private val inMemoryThreshold: Int,
+    private val spillCallback: SpillCallback,
+    private val cleanupResources: () => Unit) extends Iterator[ColumnarBatch] {
```
jlowe, line 45: In practice `cleanupResources` seems to be a no-op always?  Maybe I misread something.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -187,9 +189,14 @@ class HashJoinIterator(
     val buildSide: GpuBuildSide,
     val compareNullsEqual: Boolean, // This is a workaround to how cudf support joins for structs
     private val spillCallback: SpillCallback,
+    private val condition: Option[Expression],
+    private val fallback: Option[
+        (Iterator[ColumnarBatch], Iterator[ColumnarBatch]) => Iterator[ColumnarBatch]],
+    private val joinOutputRows: GpuMetric,
```
jlowe, line 195: Not for this PR, but wondering if we should have a case class or something to help organize the metrics and/or join key/attr params to try to keep the number of parameters to this and related join code reasonable.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -230,8 +238,23 @@ class HashJoinIterator(
         // We are about to return something. We got everything we need from it so now let it spill
         // if there is more to be gathered later on.
         gathererStore.foreach(_.allowSpilling())
+        joinOutputRows += ret.get.numRows()
+      }
+
+      if (condition.isDefined && ret.isDefined) {
```
jlowe, line 244: Nit: I think this can be simplified since `ret.flatMap` will return `None` if ret is not defined.
```suggestion
      if (condition.isDefined) {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -346,11 +373,31 @@ class HashJoinIterator(
...
+              gathererStore = joinGatherer(builtKeys, built, cb)
+            }
+          } catch {
+            case _: OutOfMemoryError if fallback.isDefined =>
+              logWarning("Join was too large falling back to CPU")
+              fallbackStreamIter = Some(Seq(GpuColumnVector.incRefCounts(cb)).toIterator ++ stream)
+              fallbackBuildIter = Some(Seq(GpuColumnVector.incRefCounts(built.getBatch)).toIterator)
+              // As soon as we start the fallback the GpuSemaphore could be released so
+              // close everything now just to be sure
+              close()
```
jlowe, line 391: At first I thought `close()` should set `fallbackIter = None` when closing and we should be checking for closed first thing in `hasNext`, but now I see that would be very bad given how this works.  It's quite odd to be intentionally using a closed iterator, but I get the motivation.  It would be good to have comments sprinkled in or at least an overview comment on the class describing this unique behavior.

A way to avoid the odd close-myself-then-continue-as-normal behavior is to have this iterator wrap the actual iterator being used.  When we fallback we don't close ourselves but instead close the underlying GPU iterator and fire up the CPU iterator to replace it.  That would also eliminate all the "if fallback" checks sprinkled throughout.  Not must-fix, but maybe a followup if you agree.

revans2:
After running some performance tests on NDS query 72 at scale factor 3000 on a large cluster and at scale factor 200 on a smaller cluster I have decided to close this issue in favor of #2433 which produced much better results and has far fewer downsides for long term support.

Issue #2406, opened 2021-05-13
[BUG] test_struct_groupby_count failed


jlowe:
On last night's Dataproc integration test build, test_struct_groupby_count failed.  From the failure log:
```
06:40:46  cpu = 2, gpu = 3
06:40:46  float_check = <function get_float_check.<locals>.<lambda> at 0x7f3960bddd30>
06:40:46  path = [0, 'a', 'a', 'aa']
```

Full log for reference:
```
06:40:46  [31m[1m_ test_struct_groupby_count[Struct(not_null)(('a', Struct(('aa', Integer),('ab', Integer))))] _[0m
06:40:46  
06:40:46  key_data_gen = Struct(not_null)(('a', Struct(('aa', Integer),('ab', Integer))))
06:40:46  
06:40:46      [37m@allow_non_gpu[39;49;00m([96many[39;49;00m = [94mTrue[39;49;00m)
06:40:46      [37m@pytest[39;49;00m.mark.parametrize([33m'[39;49;00m[33mkey_data_gen[39;49;00m[33m'[39;49;00m, [
06:40:46          StructGen([
06:40:46              ([33m'[39;49;00m[33ma[39;49;00m[33m'[39;49;00m, StructGen([
06:40:46                  ([33m'[39;49;00m[33maa[39;49;00m[33m'[39;49;00m, IntegerGen(min_val=[94m0[39;49;00m, max_val=[94m9[39;49;00m))
06:40:46              ]))], nullable=[94mFalse[39;49;00m),
06:40:46          StructGen([
06:40:46              ([33m'[39;49;00m[33ma[39;49;00m[33m'[39;49;00m, StructGen([
06:40:46                  ([33m'[39;49;00m[33maa[39;49;00m[33m'[39;49;00m, IntegerGen(min_val=[94m0[39;49;00m, max_val=[94m4[39;49;00m)),
06:40:46                  ([33m'[39;49;00m[33mab[39;49;00m[33m'[39;49;00m, IntegerGen(min_val=[94m5[39;49;00m, max_val=[94m9[39;49;00m)),
06:40:46              ]))], nullable=[94mFalse[39;49;00m),
06:40:46      ], ids=idfn)
06:40:46      [94mdef[39;49;00m [92mtest_struct_groupby_count[39;49;00m(key_data_gen):
06:40:46          [94mdef[39;49;00m [92mgroup_by_count[39;49;00m(spark):
[...]

PR #2410, opened 2021-05-13
Simplify launching integration tests a pseudo-distributed mode

gerashegalov:
1. Pre-set executor classpath
2. Easier than full spec of PYSP_TEST_spark_master for local-cluster

Example
```bash
TEST_PARALLEL=4 NUM_LOCAL_EXECS=2 SPARK_HOME=~/dist/spark-3.0.1-bin-hadoop3.2 \
./integration_tests/run_pyspark_from_build.sh -k test_struct_groupby_count
```

Signed-off-by: Gera Shegalov <gera@apache.org>

PR #2413, opened 2021-05-14
Ignore group order in test_struct_groupby_count

gerashegalov:
Contributes to #2406.


gerashegalov:
build

gerashegalov:
It may not be the entire fix but the test_struct_groupby_count fails when run manually on dataproc and locally using #2410 , and succeeds with this PR.

jlowe (approved):
I believe this is the complete fix.  Every other test in this file that does a groupby makes sure to ignore the ordering.

PR #2429, opened 2021-05-17
Change RMM_ALLOC_FRACTION to represent percentage of available memory, rather than total memory, for initial allocation

andygrove:
Signed-off-by: Andy Grove <andygrove73@gmail.com>

When running on a desktop where I am using the GPU to drive the display, I need to reduce the value for the `spark.rapids.memory.gpu.allocFraction` settings. 

This PR changes the meaning of `RMM_ALLOC_FRACTION` to be based on % of **available** rather than **total** GPU memory. Note that this is just used for the initial allocation and doesn't not restrict total memory usage.

~~Rather than modifying the source code and remembering to undo the change before committing changes, I would prefer to be able to override this setting from the command line when running Maven.~~

~~This PR provides a generic way to override RapidsConf settings using Java system properties.~~

~~For example, when running tests with Maven, I can now run:~~

```bash
mvn clean verify -DargLine="-Dspark.rapids.memory.gpu.allocFraction=0.7"
```

~~When running tests from an IDE, I can specify `-Dspark.rapids.memory.gpu.allocFraction=0.7`.~~


CONTRIBUTING.md
```diff
@@ -45,6 +45,19 @@ After a successful build the RAPIDS Accelerator jar will be in the `dist/target/
...
+
+It is sometimes necessary to override default memory settings when running the test suite. This can be achieved 
+by using Java system properties to override any configuration specific that is specific to the RAPIDS Accelerator. 
+
+For example, if the GPU is being used to drive a display then there is less memory available, and it may be 
+necessary to reduce the `spark.rapids.memory.gpu.allocFraction` setting from the default value. The following 
+syntax can be used to pass Java system properties to the Maven Surefire plugin. 
+
+```shell script
+mvn verify -DargLine="-Dspark.rapids.memory.gpu.allocFraction=0.7"
```
revans2, line 58: I understand why you want this, but it scares me.  Spark does not support this and if someone sets a config through a system property in production I'm not sure how we can debug it. Spark will not see it so it will not be logged/displayed in the typical places. We already have something that does this just for the tests but it is not well documented.

https://github.com/NVIDIA/spark-rapids/blob/247b75847bc455aad799c30bdb792a4acea7c255/tests/src/test/scala/com/nvidia/spark/rapids/SparkQueryCompareTestSuite.scala#L88-L94

Could you use that instead? and document it a little better.

jlowe:
The side-channel hacks to adjust the Spark configs are interesting but IMO don't get to the crux of the problem: the defaults _do not work_ for this kind of setup.  That's a bad initial experience for someone just trying things out on their desktop with their display GPU.

Should we make at least the initial pool allocation fraction be based on free memory rather than total memory?  Or is there an alternative where the system defaults still perform well on our expected setup but don't crash outright on this setup?

revans2:
> The side-channel hacks to adjust the Spark configs are interesting but IMO don't get to the crux of the problem: the defaults _do not work_ for this kind of setup. That's a bad initial experience for someone just trying things out on their desktop with their display GPU.
> 
> Should we make at least the initial pool allocation fraction be based on free memory rather than total memory? Or is there an alternative where the system defaults still perform well on our expected setup but don't crash outright on this setup?

We initially had it on free memory, but it makes it impossible to share a GPU among tasks in a production setting, and can mask a GPU accidentally being shared.

I am fine with us having different defaults for the unit/integration tests that look at the free GPU memory and try to configure things accordingly, so long as we also fail if there is a bad setup.

jlowe:
> it makes it impossible to share a GPU among tasks in a production setting

Not following.  All I'm talking about is the initial RMM pool size, not the maximum pool size or anything that has to do with tasks.  The RMM pool will try to grow from the initial size as it does today if the application attempts to allocate beyond that initial size.

> can mask a GPU accidentally being shared.

Yes, that is definitely the case, as we _are trying to share_ in this setup.  That's sort of my point -- should we make it easier to share by default.  Maybe the answer is no.  If that's the case, maybe we should emit a more useful error message when the initial pool size cannot be allocated to point users to the possibility that the GPU is being used by another process concurrently?  Bonus points for emitting details about the GPU that is being used, amount of free memory on it, amount trying to be allocated, etc.

revans2:
If it is just the initial pool size then I would be fine with a change like that. We would need a lot of good documentation explaining it.

andygrove:
I've updated this and it works well for my desktop. I wasn't sure whether to use the term `available` or `free` in the documentation but I went for `available` since that is the term we use when describing the behavior of `spark.rapids.python.memory.gpu.allocFraction` and hopefully I'm using the term consistently with that.

andygrove:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala
```diff
@@ -168,7 +168,7 @@ object GpuDeviceManager extends Logging {
     // Align workaround for https://github.com/rapidsai/rmm/issues/527
     def truncateToAlignment(x: Long): Long = x & ~511L
 
-    var initialAllocation = truncateToAlignment((conf.rmmAllocFraction * info.total).toLong)
+    var initialAllocation = truncateToAlignment((conf.rmmAllocFraction * info.free).toLong)
     if (initialAllocation > info.free) {
```
revans2, line 172: This check is useless now, but we probably want to have some checks for a minimum amount of free memory and warn about that, even before the initial allocation.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala
```diff
@@ -168,7 +168,7 @@ object GpuDeviceManager extends Logging {
     // Align workaround for https://github.com/rapidsai/rmm/issues/527
     def truncateToAlignment(x: Long): Long = x & ~511L
 
-    var initialAllocation = truncateToAlignment((conf.rmmAllocFraction * info.total).toLong)
+    var initialAllocation = truncateToAlignment((conf.rmmAllocFraction * info.free).toLong)
```
tgravescs, line 171: so what if free is tiny?   do we want a minimum, sometimes I can see this having caught bad setups. Meaning you tried to start 2 exectuors on same gpu when you shouldn't have, so now this will hide that kind of failures. I guess you will probably start seeing failures but I'm wondering how hard they will be to debug.

tgravescs: oops I now see Bobby made similar comment about minimum.

andygrove: I had missed Bobby's previous comment somehow. I am working on addressing this now. I also noticed that some of the existing error messages need a little rework as well.

abellina:
Is this what we want? Perhaps I misunderstand this PR.

Say I set allocFraction to 50% and launch N executors using a 32GB GPU:

- First executor: 16 GB
- Second executor: 8 GB
- Third executor: 4 GB 
- Fourth executor: 2 GB
- and so on.

andygrove:
That's a good point @abellina ... I will rethink the approach and look at having the initial allocation for the tests be set based on free memory instead

revans2:
@abellina @andygrove the reason why we are doing this is because what you described does not happen in practice.  Yes it happens for our tests, but that is the only place. We are not recommending that people share a GPU between different executors, and we have never recommended that. In fact Spark, kuberneted, and Yarn do not support doing it. We had to hack things up to make it work for our tests.

abellina:
Maybe there could be a config that sets what we are basing the fraction from: free or total. Use free as default, especially for production setups, but use total if you know what you are doing (the test hack)?

But agree that basing it off free by default is going to make for a better experience.

revans2:
Why add a config that no one is going to set? If we hit a situation where we need a config we can look into it. But for now I would just have a hard coded minimum amount of free memory where we output a warning if the initial allocation would be too low.  That is not going to save us in all cases (because there will be races) but it should help

abellina:
> Why add a config that no one is going to set?

Never mind. I missed that since this is just the init setting, it will grow as necessary. This doesn't preclude the application I had in mind (to test shuffle locally), as I originally thought. Thanks @revans2 @andygrove.

andygrove:
This is ready for another review. I added the minimum check. I could not figure out a good way to test it though.

abellina (approved):

jlowe (dismissed):

jlowe:
build

jlowe (commented):
The integration tests failed, probably because it intentionally shares the GPU to run multiple tests in parallel.  The integration tests probably need to be updated to set the new min alloc config.

andygrove:
build

andygrove:
build

abellina (dismissed):

integration_tests/src/main/python/spark_session.py
```diff
@@ -83,6 +83,7 @@ def with_gpu_session(func, conf={}):
     """
     copy = dict(conf)
     copy['spark.rapids.sql.enabled'] = 'true'
+    copy['spark.rapids.memory.gpu.minAllocFraction'] = '0'
```
abellina: nit, could we add a comment here (and the other place where this is set for the tests)? So we remember why this is needed.

jlowe (approved):

integration_tests/src/main/python/spark_session.py
andygrove: Done

andygrove:
build

abellina (approved):
Thanks @andygrove

revans2 (approved):

andygrove:
build

PR #2432, opened 2021-05-17
Switch to single-level structs in hash aggregate tests and work around value-dedupe bug 

### LINKED ISSUE #2428: [BUG] Intermittent ValueError in test_struct_groupby_count
**Describe the bug**
Runs of test_struct_groupby_count intermittently throw exceptions like:
```
=================================== FAILURES ===================================
_ test_struct_groupby_count[Struct(not_null)(('a', Struct(('aa', Integer))))] __

self = Row(a=None), item = 'aa'

    def __getitem__(self, item):
        if isinstance(item, (int, slice)):
            return super(Row, self).__getitem__(item)
        try:
            # it will be slow when it has many fields,
            # but this will not be used in normal cases
>           idx = self.__fields__.index(item)
E           ValueError: 'aa' is not in list

```

Repeatedly running the test as  a dedicated Spark app per iteration
```bash
while (($? == 0)); do 
  ((iter++)) 
  NUM_LOCAL_EXECS=2 TEST_PARALLEL=1 LD_LIBRARY_PATH=$CONDA_PREFIX/lib \
  SPARK_HOME=~/dist/spark-3.1.1-bin-hadoop3.2 \
  ./integration_tests/run_pyspark_from_build.sh -k test_struct_groupby_count  
done
```

reproduces the error after a few iterations. 

Running multiple iterations in the same pytest app does not reproduce the error easily unless the first iteration already hits the bug

```bash 
NUM_LOCAL_EXECS=2 TEST_PARALLEL=1 LD_LIBRARY_PATH=$CONDA_PREFIX/lib \ 
SPARK_HOME=~/dist/spark-3.1.1-bin-hadoop3.2 \
./integration_tests/run_pyspark_from_build.sh -k test_struct_groupby_count --count=100 -x
```

**Expected behavior**
the test should work reliably

local-cluster mode reproduces the failure

gerashegalov:
Closes #2428. Due to a subtle difference between the single-column
gen_df and two_col_df copy-and-pasting of the StructGen between tests
introduced testing of double-nestedd structs in single-column tests
using gen_df. 

Workaround https://github.com/apache/spark/pull/31778
    
### DIFF
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -605,22 +605,28 @@ def test_subquery_in_agg(adaptive, expr):
         conf = {"spark.sql.adaptive.enabled" : adaptive})
 
 
+def assert_single_level_struct(df):
+    first_level_dt = df.schema['a'].dataType
+    second_level_dt = first_level_dt['aa'].dataType
+    assert isinstance(first_level_dt, StructType)
+    assert isinstance(second_level_dt, IntegerType)
+
+
 @allow_non_gpu(any = True)
 @pytest.mark.parametrize('key_data_gen', [
     StructGen([
-        ('a', StructGen([
-            ('aa', IntegerGen(min_val=0, max_val=9))
-        ]))], nullable=False),
+        ('aa', IntegerGen(min_val=0, max_val=9)),
+    ], nullable=False),
     StructGen([
-        ('a', StructGen([
-            ('aa', IntegerGen(min_val=0, max_val=4)),
-            ('ab', IntegerGen(min_val=5, max_val=9)),
-        ]))], nullable=False),
+        ('aa', IntegerGen(min_val=0, max_val=4)),
+        ('ab', IntegerGen(min_val=5, max_val=9)),
+    ], nullable=False),
 ], ids=idfn)
 @ignore_order(local=True)
 def test_struct_groupby_count(key_data_gen):
     def group_by_count(spark):
         df = two_col_df(spark, key_data_gen, IntegerGen())
+        assert_single_level_struct(df)
         return df.groupBy(df.a).count()
     assert_gpu_and_cpu_are_equal_collect(group_by_count)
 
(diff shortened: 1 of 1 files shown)


integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -605,40 +605,47 @@ def test_subquery_in_agg(adaptive, expr):
         conf = {"spark.sql.adaptive.enabled" : adaptive})
 
 
+def assert_single_level_struct(df):
```
revans2, line 608: The bug in Spark is about a value collisions. Spark will try to save space sending data to python and looks for `Row`s with duplicate values in them.  A `Row` is used to represent both a regular row as well as a Struct. So if I have a batch like

```
Row('a' = null)
Row('a' = Row('aa' = null))
```

There is still the possibility that this bug will be hit, even with no struct nesting, simply because a top level row in spark is treated like a struct so inherently all structs are nested under a top level row struct.  The best way to remove the possibility of hitting this bug is to have a different number of entries in each struct and also different from the rows.

gerashegalov:
with the PR the command from #2432 
```bash
while (($? == 0)); do 
  ((iter++)) 
  NUM_LOCAL_EXECS=2 TEST_PARALLEL=1 LD_LIBRARY_PATH=$CONDA_PREFIX/lib \
  SPARK_HOME=~/dist/spark-3.1.1-bin-hadoop3.2 \
  ./integration_tests/run_pyspark_from_build.sh -k test_struct_groupby_count  
done
```
has not reproduced the issue for 10 minutes.

gerashegalov:
build

integration_tests/src/main/python/hash_aggregate_test.py
gerashegalov: thanks you for explaining the Spark bug @revans2

gerashegalov:
build

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -605,40 +605,55 @@ def test_subquery_in_agg(adaptive, expr):
...
 @allow_non_gpu(any = True)
 @pytest.mark.parametrize('key_data_gen', [
     StructGen([
-        ('a', StructGen([
-            ('aa', IntegerGen(min_val=0, max_val=9))
-        ]))], nullable=False),
+        ('aa', IntegerGen(min_val=0, max_val=9)),
+    ], nullable=False),
     StructGen([
-        ('a', StructGen([
```
jlowe, line 615: Given the issue is not specific to multi-level nesting of structs, are we just giving up trying to test multi-level struct support here?

gerashegalov: These tests did not intend to test deeper nesting. Per PR description a 2-level-nested struct was introduced by copy-n-paste of parameters. The same parameter generate different nesting depending on whether passed to `two_col_df` or to `gen_df` directly. The reason why it went unnoticed  is that this is a CPU fallback test without casting the struct to string first.

jlowe: Ah, sorry, I missed that in the description.  Thanks!

jlowe (approved):

revans2 (commented):
The changes look OK to me. I am a little concerned that we are dropping tests for nested structs, especially if they are supported. And we are enforcing this in all of the tests with no explanation why.

revans2:
> The changes look OK to me. I am a little concerned that we are dropping tests for nested structs, especially if they are supported. And we are enforcing this in all of the tests with no explanation why.

Sorry I missed that the test was a fallback and it was falling back in the wrong way. +1 lets get this in.

gerashegalov:
added a TODO comment for #2438  for clarification

revans2 (approved):

gerashegalov:
build

jlowe (approved):

Issue #2012, opened 2021-03-24
[FEA] Support Window functions(lead & lag) for ArrayType


Dooyoung-Hwang:
I request a feature to run lead & lag functions with ArrayType[AtomicType], ArrayType[ArrayType[AtomicType]] on GPU.

- Test code
```python
from pyspark.sql import Window
data = [(1,0.1, 0.4, 1), (2, 0.2, 0.5,2), (3, -1.0, 2.2,3)]
schema = StructType([ \
    StructField("a",IntegerType(),True), \
    StructField("b",FloatType(),True), \
    StructField("c",FloatType(),True), \
    StructField("d",IntegerType(),True)
  ])
  
win_spec = Window().partitionBy(col("a")).orderBy(desc("d"))
df1 = spark.createDataFrame(data=data,schema=schema).withColumn("arr1", array("b", "c")) \
        .withColumn("arr2", lead(col("arr1"), 1).over(win_spec))
df2 = spark.createDataFrame(data=data,schema=schema).withColumn("arr1", array("b", "c")) \
        .withColumn("arr2", lag(col("arr1"), 1).over(win_spec)) \

print(df1._jdf.queryExecution())
print(df2._jdf.queryExecution())
```

- Output message in driver log
```
!Expression lead(arr1#6082, 1, null) cannot run on GPU because expression Lead lead(arr1#6082, 1, null) produces an unsupported type ArrayType(FloatType,true); expression AttributeReference arr1#6082 produces an unsupported type ArrayType(FloatType,true)
!Expression lag(arr1#6187, 1, null) cannot run on GPU because expression Lag lag(arr1#6187, 1, null) produces an unsupported type ArrayType(FloatType,true); expression AttributeReference arr1#6187 produces an unsupported type ArrayType(FloatType,true)
```

wbo4958:
@mythrocks Do you have any idea if cudf has supported this ?

GaryShen2008:
> 
> 
> @mythrocks Do you have any idea if cudf has supported this ?

Mithun has merged the PR https://github.com/rapidsai/cudf/pull/8062

wbo4958:
close this issue, Since the PRs has been merged.

PR #2435, opened 2021-05-18
support lead/lag on arrays

wbo4958:
This PR is for this issue https://github.com/NVIDIA/spark-rapids/issues/2012

### DIFF
--- a/integration_tests/src/main/python/window_function_test.py
+++ b/integration_tests/src/main/python/window_function_test.py
@@ -273,13 +273,15 @@ def test_multi_types_window_aggs_for_rows_lead_lag(a_gen, b_gen, c_gen):
     assert_gpu_and_cpu_are_equal_collect(do_it, conf={'spark.rapids.sql.hasNans': 'false'})
 
 
-lead_lag_array_data_gens = [ArrayGen(int_gen, max_length=10),
-                            ArrayGen(ArrayGen(int_gen, max_length=10), max_length=10),
-                            ArrayGen(ArrayGen(ArrayGen(int_gen, max_length=10), max_length=10), max_length=10)]
-# lead and lag now support array type, so redo the tests, but just for array
-# SortExec does not support array type, so sort the result locally.
+lead_lag_array_data_gens =\
+    [ArrayGen(sub_gen, max_length=10) for sub_gen in lead_lag_data_gens] + \
+    [ArrayGen(ArrayGen(sub_gen, max_length=10), max_length=10) for sub_gen in lead_lag_data_gens] + \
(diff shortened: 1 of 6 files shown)


wbo4958:
build

wbo4958:
build

integration_tests/src/main/python/window_function_test.py
```diff
@@ -272,6 +272,39 @@ def do_it(spark):
                 .withColumn('row_num', f.row_number().over(baseWindowSpec))
     assert_gpu_and_cpu_are_equal_collect(do_it, conf={'spark.rapids.sql.hasNans': 'false'})
 
+
+lead_lag_array_data_gens = [ArrayGen(int_gen, max_length=10),
+                            ArrayGen(ArrayGen(int_gen, max_length=10), max_length=10),
+                            ArrayGen(ArrayGen(ArrayGen(int_gen, max_length=10), max_length=10), max_length=10)]
+# lead and lag now support array type, so redo the tests, but just for array
```
revans2, line 279: Could we update this comment to say something more like.

```
# lead and lag are supported for arrays, but the other window operations like min and max are not right now
# once they are all supported the tests should be combined.
```

integration_tests/src/main/python/window_function_test.py
```diff
@@ -272,6 +272,39 @@ def do_it(spark):
                 .withColumn('row_num', f.row_number().over(baseWindowSpec))
     assert_gpu_and_cpu_are_equal_collect(do_it, conf={'spark.rapids.sql.hasNans': 'false'})
 
+
+lead_lag_array_data_gens = [ArrayGen(int_gen, max_length=10),
+                            ArrayGen(ArrayGen(int_gen, max_length=10), max_length=10),
```
revans2, line 277: nit: Could we have a few more types than just ints?  So that we feel that we have covered more possible types.

integration_tests/src/main/python/window_function_test.py
```diff
@@ -272,6 +272,39 @@ def do_it(spark):
...
+                            ArrayGen(ArrayGen(ArrayGen(int_gen, max_length=10), max_length=10), max_length=10)]
+# lead and lag now support array type, so redo the tests, but just for array
+# SortExec does not support array type, so sort the result locally.
+@ignore_order(local=True)
+@approximate_float
```
revans2, line 282: Can we remove this? There is nothing in this test that deals with floats right now, and if there were there is nothing in the test that would make it so the floats are not exact. I know this is a copy and paste from above and the one above also has it wrong.

wbo4958:
Done

integration_tests/src/main/python/window_function_test.py
wbo4958: Done

Issue #2427, opened 2021-05-17
[BUG] test_union_struct_missing_children[(Struct(not_null) failed in databricks310 and spark 311


pxLi:
**Describe the bug**
rapids_databricks_nightly ID 3,4
spark 3.1.1 integration tests ID 173

test_union_struct_missing_children will fail after 3.1.1 (inclusive)

**Steps/Code to reproduce bug**
``` bash
cd spark-rapids/integration_tests
export SPARK_HOME=3.1.1_VERSION
export TEST=test_union_struct_missing_children
./run_pyspark_from_build.sh
```
FAILED ../../src/main/python/repart_test.py::test_union_struct_missing_children[(Struct(not_null)(['child0', Decimal(7,2)]), Struct(not_null)(['child1', Integer]))]

```bash
[2021-05-16T06:29:57.480Z] answer = 'xro243542'
[2021-05-16T06:29:57.480Z] gateway_client = <py4j.java_gateway.GatewayClient object at 0x7f26c6b11940>
[2021-05-16T06:29:57.480Z] target_id = 'o243539', name = 'collectToPython'
[2021-05-16T06:29:57.480Z] 
[2021-05-16T06:29:57.480Z]     [94mdef[39;49;00m [92mget_return_value[39;49;00m(answer, gateway_client, target_id=[94mNone[39;49;00m, name=[94mNone[39;49;00m):
[2021-05-16T06:29:57.480Z]         [33m"""Converts an answer received from the Java gateway into a Python object.[39;49;00m
[2021-05-16T06:29:57.480Z]     [33m[39;49;00m
[2021-05-16T06:29:57.480Z]     [33m    For example, string representation of integers are converted to Python[39;49;00m
[2021-05-16T06:29:57.480Z]     [33m    integer, string representation of objects are converted to JavaObject[39;49;00m
[2021-05-16T06:29:57.480Z]     [33m    instances, etc.[39;49;00m
[2021-05-16T06:29:57.480Z]     [33m[39;49;00m
[2021-05-16T06:29:57.480Z]     [33m    :param answer: the string returned by the Java gateway[39;49;00m
[2021-05-16T06:29:57.480Z]     [33m    :param gateway_client: the gateway client used to communicate with the Java[39;49;00m
[2021-05-16T06:29:57.480Z]     [33m        Gateway. Only necessary if the answer is a reference (e.g., object,[39;49;00m
[...]

PR #2442, opened 2021-05-18
fix GpuCreateNamedStruct not serializable issue

wbo4958:
The return from columnarEval of GpuLiteral has been changed to GpuScalar which
will be as the type of names(a private field in GpuCreateNamedStruct). And the
names will be serialized before spark scheduling, but the GpuScalar is not serializable,
it will cause "org.apache.spark.SparkException: Task not serializable".

On the other hand, the calling `columnarEval(null)` in the driver side is really dangerous,
This PR also pull it apart manually

Closes https://github.com/NVIDIA/spark-rapids/issues/2427
Closes  https://github.com/NVIDIA/spark-rapids/issues/2443

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
@@ -89,6 +89,18 @@ object GpuExpressionsUtils extends Arm {
    */
   def columnarEvalToColumn(expr: Expression, batch: ColumnarBatch): GpuColumnVector =
     resolveColumnVector(expr.columnarEval(batch), batch.numRows, expr.dataType)
+
+  /**
+   * Extract the GpuLiteral
+   * @param exp the input expression to be extracted
+   * @return an optional GpuLiteral
+   */
+  @scala.annotation.tailrec
+  def extractGpuLit(exp: Expression): Option[GpuLiteral] = exp match {
+    case gl: GpuLiteral => Some(gl)
+    case ga: GpuAlias => extractGpuLit(ga.child)
+    case _ => None
+  }
 }
 
 /**
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
@@ -17,7 +17,7 @@
 package org.apache.spark.sql.rapids
 
 import ai.rapids.cudf.{ColumnVector, DType}
-import com.nvidia.spark.rapids.{GpuAlias, GpuColumnVector, GpuExpression, GpuExpressionsUtils, GpuLiteral}
+import com.nvidia.spark.rapids.{GpuColumnVector, GpuExpression, GpuExpressionsUtils}
 import com.nvidia.spark.rapids.RapidsPluginImplicits.ReallyAGpuExpression
 
 import org.apache.spark.sql.catalyst.analysis.{TypeCheckResult, TypeCoercion}
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
@@ -94,10 +94,11 @@ case class GpuCreateNamedStruct(children: Seq[Expression]) extends GpuExpression
   // And on the other hand, the calling for columnarEval(null) in the driver side is
   // dangerous for GpuExpressions, we'll have to pull it apart manually.
   private lazy val names = nameExprs.map {
-    case gl: GpuLiteral => gl.value
-    case GpuAlias(gl: GpuLiteral, _) => gl.value
     case ge: GpuExpression =>
-      throw new IllegalStateException(s"Unexpected GPU expression $ge")
+      GpuExpressionsUtils.extractGpuLit(ge) match {
+        case Some(gpuLiteral) => gpuLiteral.value
+        case None => throw new IllegalStateException(s"Unexpected GPU expression $ge")
+      }
     case e => e.eval(EmptyRow)
   }
 
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
@@ -17,7 +17,7 @@
 package org.apache.spark.sql.rapids
 
 import ai.rapids.cudf.{ColumnVector, DType}
-import com.nvidia.spark.rapids.{GpuColumnVector, GpuExpression, GpuExpressionsUtils, GpuLiteral}
+import com.nvidia.spark.rapids.{GpuAlias, GpuColumnVector, GpuExpression, GpuExpressionsUtils, GpuLiteral}
 import com.nvidia.spark.rapids.RapidsPluginImplicits.ReallyAGpuExpression
 
 import org.apache.spark.sql.catalyst.analysis.{TypeCheckResult, TypeCoercion}
@@ -87,8 +87,15 @@ case class GpuCreateNamedStruct(children: Seq[Expression]) extends GpuExpression
     case Seq(name, value) => (name, value)
   }.toList.unzip
 
+  // Names will be serialized before Spark scheduling, and the returned type GpuScalar
+  // from GpuLiteral.columnarEval(null) can't be serializable, which causes
+  // `org.apache.spark.SparkException: Task not serializable` issue.
+  //
+  // And on the other hand, the calling for columnarEval(null) in the driver side is
+  // dangerous for GpuExpressions, we'll have to pull it apart manually.
   private lazy val names = nameExprs.map {
     case gl: GpuLiteral => gl.value
+    case GpuAlias(gl: GpuLiteral, _) => gl.value
     case ge: GpuExpression =>
       throw new IllegalStateException(s"Unexpected GPU expression $ge")
     case e => e.eval(EmptyRow)
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
@@ -17,7 +17,7 @@
 package org.apache.spark.sql.rapids
 
 import ai.rapids.cudf.{ColumnVector, DType}
-import com.nvidia.spark.rapids.{GpuColumnVector, GpuExpression, GpuExpressionsUtils, GpuScalar}
+import com.nvidia.spark.rapids.{GpuColumnVector, GpuExpression, GpuExpressionsUtils, GpuLiteral}
 import com.nvidia.spark.rapids.RapidsPluginImplicits.ReallyAGpuExpression
 
 import org.apache.spark.sql.catalyst.analysis.{TypeCheckResult, TypeCoercion}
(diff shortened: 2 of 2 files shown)


wbo4958:
build

wbo4958:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -82,7 +82,13 @@ case class GpuCreateNamedStruct(children: Seq[Expression]) extends GpuExpression
   }.toList.unzip
 
   private lazy val names = nameExprs.map {
-    case g: GpuExpression => g.columnarEval(null)
+    case g: GpuExpression =>
```
firestarman, line 85: It would be better
```Scala
    case gLit: GpuLiteral => gLit.value
    case ge: GpuExpression => throw new IllegalStateException("Unexpected GPU expression")
```

wbo4958: Hi @firestarman, One question is how do we ensure it must be GpuLiteral? Is there any possibility that the nameExprs is some other GpuExpression?

firestarman: Yes, you are right, we need to do verification and take care of all possible cases. Calling `columnarEval` with an invalid batch (null) is dangerous .

wbo4958: Yeah, But I will take your suggestion

sperlingxx: Great fix! To me, `ret` here can not belong to types other than `GpuScalar`.  So, we can assert it is a `GpuScalar`.

firestarman: This `columnarEval(null)` existed before, so it is not the issue of this PR, but we can update it in the PR.

wbo4958: Done

wbo4958:
build

firestarman (dismissed):

wbo4958:
@revans2 Could you review this PR

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -82,7 +82,9 @@ case class GpuCreateNamedStruct(children: Seq[Expression]) extends GpuExpression
   }.toList.unzip
 
   private lazy val names = nameExprs.map {
-    case g: GpuExpression => g.columnarEval(null)
+    case gl: GpuLiteral => gl.value
+    case ge: GpuExpression =>
+      throw new IllegalStateException(s"Unexpected GPU expression $ge")
```
revans2, line 87: I have tried to make it so the Literal is wrapped in an Alias, which should totally be valid, but I have not been able to make it happen with the public spark APIs. If we are going to throw an exception on something we don't expect, then we should be absolutely positive that we got what we expected, and all of the default literal checks being use allow for an alias. So either still do the eval and then pull the name out of the `GpuScalar` that is returned. Or add in a second rule for an Alias of a literal, just to be on the safe side.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -82,7 +82,9 @@ case class GpuCreateNamedStruct(children: Seq[Expression]) extends GpuExpression
   }.toList.unzip
 
   private lazy val names = nameExprs.map {
-    case g: GpuExpression => g.columnarEval(null)
+    case gl: GpuLiteral => gl.value
```
revans2, line 85: The reason for this change is not clear. Could we get a comment to explain why this is happening so no one accidentally changes it back.

revans2 (changes requested):
I left a comment before, but marking as needs changes because ti was already approved.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
jlowe: > still do the eval and then pull the name out of the GpuScalar that is returned.

We can't do this because this is happening on the driver side, correct?  That's one of the drawbacks of the recent `GpuScalar` refactor -- it's practically impossible to call `columnarEval` without requiring a GPU to do it, so it's effectively off-limits to the driver.  That would be a simpler way to handle any Alias or similar shenanigans, but since we can't call `columnarEval` or `eval` to solve it, we'll have to pull it apart manually instead IIUC.

wbo4958: I aggress with Jason, and the new commit has added GpuAlias case. BTW, I see the name check has been limited to literal, Do we really add the GpuAlias case?

``` scala
object CreateNamedStructCheck extends ExprChecks {
  val nameSig: TypeSig = TypeSig.lit(TypeEnum.STRING)
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
wbo4958: Done

wbo4958:
build

wbo4958:
since there is some conflicts, I rebased the PR

wbo4958:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -87,8 +87,15 @@ case class GpuCreateNamedStruct(children: Seq[Expression]) extends GpuExpression
...
 
+  // Names will be serialized before Spark scheduling, and the returned type GpuScalar
+  // from GpuLiteral.columnarEval(null) can't be serializable, which causes
+  // `org.apache.spark.SparkException: Task not serializable` issue.
+  //
+  // And on the other hand, the calling for columnarEval(null) in the driver side is
+  // dangerous for GpuExpressions, we'll have to pull it apart manually.
   private lazy val names = nameExprs.map {
     case gl: GpuLiteral => gl.value
+    case GpuAlias(gl: GpuLiteral, _) => gl.value
```
jlowe, line 98: Technically Alias can nest arbitrarily, so I think we'd want to match on the result of calling something like what `GpuOverrides.extractLit` does but for `GpuLiteral` or what Spark's `AliasHelper.trimAliases` does, probably handling `GpuAlias` since it's already been translated at that point.

wbo4958:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
wbo4958: Yeah, Good suggestion, from the safe side, I added the extractGpulit for this case.
Thx

wbo4958: Done

firestarman:
LGTM

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -87,8 +87,18 @@ case class GpuCreateNamedStruct(children: Seq[Expression]) extends GpuExpression
...
+  // Names will be serialized before Spark scheduling, and the returned type GpuScalar
+  // from GpuLiteral.columnarEval(null) can't be serializable, which causes
+  // `org.apache.spark.SparkException: Task not serializable` issue.
+  //
+  // And on the other hand, the calling for columnarEval(null) in the driver side is
+  // dangerous for GpuExpressions, we'll have to pull it apart manually.
   private lazy val names = nameExprs.map {
-    case g: GpuExpression => g.columnarEval(null)
+    case ge: GpuExpression =>
+      GpuExpressionsUtils.extractGpuLit(ge) match {
```
gerashegalov, line 98: nit:
```scala
GpuExpressionsUtils.extractGpuLit(ge).map(_.value)
   .getOrElse(throw new IllegalStateException(s"Unexpected GPU expression $ge"))
```

wbo4958: Thx Gera

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
firestarman: > it's practically impossible to call columnarEval without requiring a GPU to do it, so it's effectively off-limits to the driver.

If a `GpuScalar ` is created from a Scala value (`GpuLiteral` will create such a `GpuScalar ` in its `columnarEval`) and only the `getValue` will be called to get the internal Scala value, then it is safe to use on the driver side without GPU. Becasue the GpuScalar for this case is totally a wrapper of a Scala value, nothing to do with GPU. 

But I am a little confused. Could we use the `columnarEval` on driver side with an invalid input for a generic GPU expression before the GpuScalar refactor?  I think it is still not always safe because the processing inside the `columnarEval` may require a GPU. It may be safe if we know the instance of the GPU expression. And if we know the instance of a GPU expression, I think we can use its direct methods or members instead.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
```diff
@@ -89,6 +89,18 @@ object GpuExpressionsUtils extends Arm {
...
   def columnarEvalToColumn(expr: Expression, batch: ColumnarBatch): GpuColumnVector =
     resolveColumnVector(expr.columnarEval(batch), batch.numRows, expr.dataType)
+
+  /**
+   * Extract the GpuLiteral
+   * @param exp the input expression to be extracted
+   * @return an optional GpuLiteral
+   */
+  @scala.annotation.tailrec
+  def extractGpuLit(exp: Expression): Option[GpuLiteral] = exp match {
```
revans2, line 99: nit: This really feels like it should be in an object instead of a class (static method instead of non-static)

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
revans2: +1, but I would take it a step further because `extractGpuLit` takes an `Expression` as input not a `GpuExpression` and we should always get `GpuExpression`s in `nameExprs` or we have other problems.

```
private lazy val names = nameExprs.map { exp =>
  GpuExpressionsUtils.extractGpuLit(exp)
    .getOrElse(throw new IllegalStateException(s"Unexpected Expression $exp"))
}
```

revans2 (approved):
The change looks fine, and I would be OK with checking it in as is, but I would like to see it cleaned up a bit.

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
wbo4958: It's in the `object GpuExpressionsUtils`.

revans2 (approved):
not exactly what I wanted, but good enough.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
revans2: I don't think you understood. I don't want `extractGpuLit` to take a `GpuExpression`.  It is good like it is. I just don't think we need to check for the non-GpuExpression case when computing `names`.  It should never happen so it is OK to throw an exception if we see it., Which is why I said we could simplify the processing to just be.

```
private lazy val names = nameExprs.map { exp =>
  GpuExpressionsUtils.extractGpuLit(exp)
    .getOrElse(throw new IllegalStateException(s"Unexpected Expression $exp"))
}
```

wbo4958: Oh, Yeah, You're right. Sorry for the misunderstanding.

Issue #2421, opened 2021-05-14
[BUG] NPE when running mapInPandas Pandas UDF in 0.5GA


viadea:
**Describe the bug**
NPE when running mapInPandas Pandas UDF in 0.5GA. 
The same code works fine in 0.4.1GA if I just swap the jars.
Note: All other pandas UDF tests work fine on 0.5GA.

**Steps/Code to reproduce bug**
Please provide a list of steps or a code sample to reproduce the issue.
Avoid posting private or sensitive data.

```
import pyspark
from pyspark.sql.functions import udf
from pyspark.sql.functions import col, count, rand, collect_list, explode, struct, count, lit
from pyspark.sql.functions import pandas_udf
from pyspark.sql.types import *
from pyspark.sql import Row,SparkSession
import pandas as pd
import numpy as np
import cudf

spark=SparkSession \
  .builder \
  .appName("Pandas-UDF-Test") \
  .config("spark.rapids.sql.python.gpu.enabled","true") \
  .config("spark.rapids.memory.gpu.allocFraction","0.5") \
  .config("spark.rapids.python.concurrentPythonWorkers","2") \
  .config("spark.rapids.python.memory.gpu.allocFraction","0.1") \
  .config("spark.rapids.python.memory.gpu.maxAllocFraction","0.1") \
  .config("spark.sql.execution.arrow.pyspark.enabled","true") \
  .config("spark.rapids.sql.exec.ArrowEvalPythonExec","true") \
  .config("spark.rapids.sql.exec.MapInPandasExec","true") \
  .config("spark.rapids.sql.exec.FlatMapGroupsInPandasExec","true") \
  .config("spark.rapids.sql.exec.AggregateInPandasExec","true") \
  .config("spark.rapids.sql.exec.FlatMapCoGroupsInPandasExec","true") \
  .config("spark.rapids.sql.exec.WindowInPandasExec","true") \
  .getOrCreate()

df = spark.createDataFrame([(1, 21), (2, 30)], ("id", "age"))

def filter_func(iterator):
    for pdf in iterator:
        yield pdf[pdf.id == 1]

out = df.mapInPandas(filter_func, schema=df.schema)
out.printSchema()
out.show()

```

Failed with below stacktrace:

```
py4j.protocol.Py4JJavaError: An error occurred while calling o116.showString.
[...]

PR #2447, opened 2021-05-19
Filter out the nulls after slicing the batches.

firestarman:
This PR is to fix the NPE in `GpuShuffleExchange` by filtering out the nulls after slicing the batches for `GpuRangePartitioner` and `GpuSinglePartitioning`. Just like what the `GpuHashPartitioning` does [here](https://github.com/NVIDIA/spark-rapids/blob/branch-21.06/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuHashPartitioning.scala#L89). 

The root cause of this NPE is the batch array returned from the method [sliceInternalGpuOrCpu ](https://github.com/NVIDIA/spark-rapids/blob/branch-21.06/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuPartitioning.scala#L98)may contain nulls that generated by the function ['sliceBatch'](https://github.com/NVIDIA/spark-rapids/blob/branch-21.06/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuPartitioning.scala#L35) for the corner case of empty batch (row number is 0). So even a single batch is expected, we still need to filter out the null.

Closes https://github.com/NVIDIA/spark-rapids/issues/2421 

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuPartitioning.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuPartitioning.scala
@@ -111,7 +111,7 @@ trait GpuPartitioning extends Partitioning with Arm {
         sliceInternalOnGpu(numRows, partitionIndexes, partitionColumns)
       } else {
         sliceInternalOnCpu(numRows, partitionIndexes, partitionColumns)
-      }
+      }.filter(_ != null)
     } finally {
       sliceRange.close()
     }
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuPartitioning.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuPartitioning.scala
@@ -111,7 +111,7 @@ trait GpuPartitioning extends Partitioning with Arm {
         sliceInternalOnGpu(numRows, partitionIndexes, partitionColumns)
       } else {
         sliceInternalOnCpu(numRows, partitionIndexes, partitionColumns)
-      }.filter(_ != null)
+      }
     } finally {
       sliceRange.close()
     }
--- a/docs/supported_ops.md
+++ b/docs/supported_ops.md
@@ -9588,7 +9588,7 @@ Accelerator support is described below.
 <td>S</td>
 <td><b>NS</b></td>
 <td>S</td>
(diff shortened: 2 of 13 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuPartitioning.scala
```diff
@@ -111,7 +111,7 @@ trait GpuPartitioning extends Partitioning with Arm {
         sliceInternalOnGpu(numRows, partitionIndexes, partitionColumns)
       } else {
         sliceInternalOnCpu(numRows, partitionIndexes, partitionColumns)
-      }
+      }.filter(_ != null)
```
wbo4958, line 114: is it safe to add this here, I checked some of them like (GpuSinglePartitioning,  GpuRangePartitioner) don't filter out the null? 

BTW, if it is safe, This PR should get rid of the other redundant filter

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuPartitioning.scala
firestarman: > is it safe to add this here, I checked some of them like (GpuSinglePartitioning, GpuRangePartitioner) don't filter out the null?
> 

That's why I add the filtering here, to make sure it takes effect for all the callers.

firestarman: Removed the redundant filter

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuPartitioning.scala
firestarman: Some tests failed, yes, it is not safe to do the filtering here. Updated

firestarman:
build

firestarman:
I marked the PR as draft, because I am not sure whether the linked issue is reproduceable on branch-21.06 (it was 0.6 before), which is the base branch of the PR.

I tried several times locally but can not get the NPE.

revans2 (approved):
nit: if it is simple, and only if it is simple, can we get a test that causes this to happen more regularly?

jlowe (approved):

firestarman:
Thanks @viadea for the verification on branch-21.06. 
I am going to merge this PR since it works.

PR #2448, opened 2021-05-19
support creating list ColumnVector for Literal(ArrayType(NullType))

wbo4958:
This PR is bringing back the feature to support creating list ColumnVector for
ArrayType(NullType). 

The previous implementation was a workaround to generate GpuCreateArray
from LiteralExprMeta for ArrayType(NullType) and then create the corresponding
list ColumnVector in GpuCreateArray. 

This PR just implement the logic based on the feature GpuLiteral for array
instead of GpuCreateArray.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
@@ -176,8 +176,8 @@ object GpuScalar extends Arm with Logging {
         val colType = resolveElementType(elementType)
         val rows = seq.map(convertElementTo(_, elementType))
         ColumnVector.fromStructs(colType, rows.asInstanceOf[Seq[HostColumnVector.StructData]]: _*)
-      case NullType => // Byte is used for NullType
-        ColumnVector.fromBoxedBytes(seq.asInstanceOf[Seq[JByte]]: _*)
+      case NullType =>
+        GpuColumnVector.columnVectorFromNull(seq.size, NullType)
       case u =>
         throw new IllegalArgumentException(s"Unsupported element type ($u) to create a" +
           s" ColumnVector.")
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
@@ -176,6 +176,8 @@ object GpuScalar extends Arm with Logging {
         val colType = resolveElementType(elementType)
         val rows = seq.map(convertElementTo(_, elementType))
         ColumnVector.fromStructs(colType, rows.asInstanceOf[Seq[HostColumnVector.StructData]]: _*)
+      case NullType => // Byte is used for NullType
+        ColumnVector.fromBoxedBytes(seq.asInstanceOf[Seq[JByte]]: _*)
       case u =>
         throw new IllegalArgumentException(s"Unsupported element type ($u) to create a" +
           s" ColumnVector.")
--- a/tests/src/test/scala/com/nvidia/spark/rapids/ProjectExprSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/ProjectExprSuite.scala
@@ -61,6 +61,7 @@ class ProjectExprSuite extends SparkQueryCompareTestSuite {
             Array(StructField("id", IntegerType), StructField("name", StringType)))))),
         new Column(Literal.create(List(BigDecimal(123L, 2), BigDecimal(-1444L, 2)),
           ArrayType(DecimalType(10, 2)))))
+        .selectExpr("array(null)", "array(array(null))", "array()")
   }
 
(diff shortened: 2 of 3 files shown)


wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
```diff
@@ -176,6 +176,8 @@ object GpuScalar extends Arm with Logging {
         val colType = resolveElementType(elementType)
         val rows = seq.map(convertElementTo(_, elementType))
         ColumnVector.fromStructs(colType, rows.asInstanceOf[Seq[HostColumnVector.StructData]]: _*)
+      case NullType => // Byte is used for NullType
+        ColumnVector.fromBoxedBytes(seq.asInstanceOf[Seq[JByte]]: _*)
```
wbo4958, line 180: Why we are using Byte for NullType, pls refer to https://github.com/NVIDIA/spark-rapids/blob/branch-21.06/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java#L456

firestarman: So here a ColumnVector filled with nulls is expected ? If yes, you can try
`GpuColumnVector.columnVectorFromNull(seq.size, NullType)`, then you do not need the comment here.

tests/src/test/scala/com/nvidia/spark/rapids/ProjectExprSuite.scala
```diff
@@ -61,6 +61,7 @@ class ProjectExprSuite extends SparkQueryCompareTestSuite {
             Array(StructField("id", IntegerType), StructField("name", StringType)))))),
         new Column(Literal.create(List(BigDecimal(123L, 2), BigDecimal(-1444L, 2)),
           ArrayType(DecimalType(10, 2)))))
```
firestarman, line 63: Could you update to use the same code style here instead of `selectExpr` ? Like
```Scala
lit(Array(null)),
lit(Array(Array(null)))
```

tests/src/test/scala/com/nvidia/spark/rapids/ProjectExprSuite.scala
```diff
@@ -61,6 +61,7 @@ class ProjectExprSuite extends SparkQueryCompareTestSuite {
...
+        .selectExpr("array(null)", "array(array(null))", "array()")
```
firestarman, line 64: I am a little confused.  Does the `ArrayType(NullType)` cover the `array(array(null)` ?

firestarman: Discussed offline.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
revans2: Yes, cudf does not support a null type so we just use a byte because it is the smallest available. and @firestarman is correct please use `GpuColumnVector.columnVectorFromNull(seq.length, NullType)`

revans2 (approved):
This looks good to me but there are some nits that would be good to clean up.

tests/src/test/scala/com/nvidia/spark/rapids/ProjectExprSuite.scala
wbo4958: It didn't work, I tried this before, it will throw exception, 

```
Unsupported component type class java.lang.Object in arrays;
org.apache.spark.sql.AnalysisException: Unsupported component type class java.lang.Object in arrays;
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
wbo4958: Done

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
wbo4958: @revans2 Seems seq.size equals seq.length

revans2: Yes, it is super minor. Scala prefers `length` over `size`. Java has `length` for arrays and `size` for everything else. In Scala both work but idiomatic Scala prefers length, so it is consistent everywhere.

tests/src/test/scala/com/nvidia/spark/rapids/ProjectExprSuite.scala
revans2: This is Scala vs Spark types.  An `Array(null)` in Scala is an `Array[Null]`. But Spark does not map `scala.runtime.Null` to the SQL null type. This might be a bug in Spark, but apparently unless you are doing crazy tests very few people do things like this.

revans2 (approved):

PR #2451, opened 2021-05-19
Update tuning docs to add batch size recommendations.

revans2:


docs/tuning-guide.md
```diff
@@ -176,15 +220,42 @@ Default value: `0`
...
-into a single batch to make the GPU processing more efficient.  This setting controls the upper
-limit on the concatenation process.  Setting this value too low can result in a large amount of
-GPU processing overhead and slower task execution.  Setting this value too high can lead to GPU
-out of memory errors.  If tasks fail due to GPU out of memory errors after the query input
-partitions have been read, try setting this to a lower value.
+in a series of columnar batches. During processing multiple batches may be concatenated
+into a single batch to make the GPU processing more efficient, or in other cases the output size
+of an operation like join or sort will target that batch size for output. This setting controls the upper
+limit on batches that are output by these tasks.  Setting this value too low can result in a
+large amount of  GPU processing overhead and slower task execution.  Setting this value too high
```
tgravescs, line 230: extra space after "of"

docs/tuning-guide.md
```diff
@@ -176,15 +220,42 @@ Default value: `0`
...
+can lead to GPU out of memory errors.  If tasks fail due to GPU out of memory errors after the
+query input partitions have been read, try setting this to a lower value. The maximum size is just
+under 2 GiB. In general, we recommend setting the batch size to
+
+```
+min(2GiB - 1 byte, (gpu_memory - 1 GiB) / gpu_concurrency / 4)
+```
+
+Where `gpu_memory` is the amount of memory on the GPU, or the maximum pool size if using pooling.
+We then subtract from that 1 GiB for overhead in holding cuda kernels/etc and divide by the
```
jlowe, line 240:
```suggestion
We then subtract from that 1 GiB for overhead in holding CUDA kernels/etc and divide by the
```

docs/tuning-guide.md
```diff
@@ -64,20 +73,47 @@ the GPU and host memory as the transfer can be performed asynchronously from the
...
+The GPU Accelerator employs different algorithms that allow it to process more data than can fit in
+the GPU's memory. We do not support this for all operations, and are constantly trying to add more.
+The way that this can work is by spilling parts of the data to host memory or to disk, and then
+reading them back in when needed. Spilling in general is more expensive than not spilling, but
+spilling to a slow disk can be especially expensive. If you see spilling to disk happening in
+your query, and you are not using GPU Direct storage for spilling, you may want to add more
+spill storage.
+
+You can configure this larger than the amount of pinned memory you have, and the Accelerator will
+end up using non-pinned memory when this happens.
```
jlowe, line 115: Technically the host memory store will use non-pinned memory when it detects there are less than the configured maximum bytes of host spill buffers being tracked in host memory, tries to allocate a new buffer from the pinned pool, and fails.  So even if the host spill store size is configured to be less than the pinned pool size, it's not guaranteed all host spills will be to pinned memory -- depends on the usage of the pinned pool by other entities.  Not sure we need to go into this much detail here, but making sure that distinction was understood.

tgravescs (approved):

jlowe (approved):

jlowe:
build

PR #2456, opened 2021-05-20
Fall back to the CPU for literal array values on case/when

### LINKED ISSUE #2455: [BUG] CaseWhen crashes on literal arrays
**Describe the bug**
#2388 added in support for arrays in the case/when statement. But never tested for literal arrays. #2445 was filed as a follow on to "fix" the tests, but really just showed that case/when is broken right now.  If you try to use a literal array in a case when statement it crashes with a cudf error. We need to fall back to the CPU if this is not supported.

revans2:
This fixes #2455

### DIFF
--- a/integration_tests/src/main/python/conditionals_test.py
+++ b/integration_tests/src/main/python/conditionals_test.py
@@ -14,9 +14,9 @@
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


revans2:
build

jlowe (approved):

PR #2458, opened 2021-05-20
Split batch by key for window operations

### LINKED ISSUE #1856: [FEA] Create a batch chunking iterator and integrate it with GpuWindowExec
This is a part of #1789

There are several possible places where an iterator that splits sorted batches up by a given set of partition keys is useful. The one we currently care about is with GpuWindowExec. For GpuWindowExec there are two sets of keys that are passed into it.  The order by key(s) and the partition by keys(s). So if we wanted to run a command like 

```
ROW_NUMBER() OVER (PARTITION BY A ORDER BY B) AS ROWNUM
```

Then the code would hash partition/shuffle on the column A and then sort within each partition on A, B, so that all of the data  for each unique A group is next to each other. like in the following table

|A|B|
|-|-| 
|1|0|
|1|5|
|6|-4|
|6|0|
|6|5|

Currently in GpuWindowExec this is all brought into a single batch and sent to cudf for processing.  But each window operation will never look outside of the current group in the partition by keys. So in the above example we would get the same result if we split the batch up into two in between A=1 and A=6 and then processed them each separately.

This could reduce the maximum memory usage a lot especially now that we have out of core sorting working too. The goal here is to then create an iterator that will take in an iterator of sorted batches and keys to split/chunk on.  It would then need to take the batch, look at the chunk on key(s) for the last entry in that batch, and split it at the lower bound position, because we don't know if all of the keys for that item have been sorted yet.

Everything below the lower bound can go on for processing in the window function.  Everything above it needs to be saved for the next batch so it can be appended to it and sent for processing.

We need to make sure that we use SpillableColumnarBatches for storing the data so we can spill if needed.  We also need to make sure that we cover corner cases like there is only one chunk on key in the entire batch so this then devolves into a concatenate batches operation.

There are a number of optimizations that might be nice to work on with this.

1. If out of core sort is disabled we should not bother with this. It is already a single batch, why complicate it.
2. GpuSort is going to have cached all of the data and it will know when it is outputting the last batch.  It might be nice to have it tag that last batch some how so we don't have to save a small part of the final batch just because we didn't know that no more data is coming.
3. GpuSort knows in some cases what the next lowest row to be output looks like. We might be able to include that in the metadata too to avoid splitting.  This is much less likely to have a real world impact so probably skip this optimization unless you are feeling ambitious.

**Additional context**
If there are no partition by keys then all of the data is part of the same window and will go to a single task for processing.  In that case we need to make sure that we are gathering all of the data into a single batch for processing.

revans2:
This fixes #1856

It makes it so that window operations do not need a `RequireSingleBatch` for anything except when there is no partitioning. This allows us to do window operations where there are a lot of keys, but more data than fits on the GPU. I have not updated the tests yet. I'll post that shortly, but I have done manual testing on very large data sets and it has essentially no performance impact on queries that do fit in GPU memory.


revans2:
Switched to draft. I found a bug in cudf while adding the tests and I want to fix it before adding the tests. I also have a new idea on how to get the framework to insert the chunking only when needed and I want to work on that too.

revans2:
I added tests and fixed some issues, including #2473

But if found a bug in cudf as a part of the tests so this is also blocked by https://github.com/rapidsai/cudf/pull/8314

revans2:
I am going to move the test fixes to another PR, and then move this to the next release, and I don't think the benefit is worth the added risk right now.

revans2:
build

revans2:
This is actually waiting for a new branch because it is a bit too risky for 21.06

revans2:
build

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuKeyBatchingIterator.scala
```diff
@@ -0,0 +1,243 @@
...
+    pendingSize = 0
+  }
+
+  override def hasNext: Boolean = pending.nonEmpty || iter.hasNext
+
+  private def getKeyCutoff(cb: ColumnarBatch): Int = {
+    val candidates = withResource(sorter.appendProjectedColumns(cb)) { appended =>
+      withResource(GpuColumnVector.from(appended)) { table =>
+        // With data skew we don't want to include too much more in a batch than is necessary
+        // so we are going to search in up to 16 evenly spaced locations in the batch, including the
```
jlowe, line 67:
```suggestion
        // so we are going to search up to 16 evenly spaced locations in the batch, including the
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuKeyBatchingIterator.scala
```diff
@@ -0,0 +1,243 @@
...
+        // last entry. From that we will see what cutoff works and is closest to the size we want.
+        // The 16 is an arbitrary number. It was picked because hopefully it is small enough that
+        // the indexes will all fit in the CPU cache at once so we can process the data in an
+        // efficient way.
+        val maxRow = cb.numRows() - 1
+        val rowsPerBatch = Math.max(1, Math.ceil(cb.numRows()/16.0).toInt)
+        val probePoints = (1 to 16).map(idx => Math.min(maxRow, idx * rowsPerBatch)).distinct
```
jlowe, line 74: Do we want `-1` as a probe point?  Wondering about the case where `cb.numRows()` == 0 which results in maxRow == -1.

revans2: I will try to guard against it and filter them out.

revans2:
build

revans2:
@jlowe I have addressed your comments please take another look

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuKeyBatchingIterator.scala
```diff
@@ -0,0 +1,249 @@
...
+  private def getKeyCutoff(cb: ColumnarBatch): Int = {
+    val candidates = withResource(sorter.appendProjectedColumns(cb)) { appended =>
+      withResource(GpuColumnVector.from(appended)) { table =>
+        // With data skew we don't want to include too much more in a batch than is necessary
+        // so we are going to search up to 16 evenly spaced locations in the batch, including the
+        // last entry. From that we will see what cutoff works and is closest to the size we want.
+        // The 16 is an arbitrary number. It was picked because hopefully it is small enough that
+        // the indexes will all fit in the CPU cache at once so we can process the data in an
+        // efficient way.
+        val maxRow = cb.numRows() - 1
```
jlowe, line 72: Nit: Would be nice to have a `require(maxRow >=0)` to protect in case someone later refactors and forgets to avoid passing empty batches to this function.

jlowe (approved):
Only a minor nit that's fine to leave, lgtm.

revans2:
build

jlowe (approved):

PR #2460, opened 2021-05-20
align GDS reads/writes to 4 KiB

rongou:
Depends on https://github.com/rapidsai/cudf/pull/8301.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -102,17 +104,18 @@ class RapidsGdsStore(
   }
 
   private def singleShotSpill(other: RapidsBuffer, deviceBuffer: DeviceMemoryBuffer)
-      : RapidsBufferBase = {
+  : RapidsBufferBase = {
     val id = other.id
     val path = id.getDiskPath(diskBlockManager)
+    val alignedSize = RapidsGdsStore.alignBufferSize(deviceBuffer)
```
jlowe, line 110: It would be nice to have comments explaining why it's desired and safe to deliberately walk off the known end of the buffer, otherwise every occurrence of `alignBufferSize` looks broken and pointless.

As an aside, I can see the desire to align the address of both the src and dest, but is the length really that important when we're not appending/sharing a file?

rongou: Ok, added some comments about the alignment approach.

jlowe (dismissed):

rongou:
build

revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -100,11 +100,17 @@ class RapidsGdsStore(
       dst match {
         case dmOriginal: DeviceMemoryBuffer =>
           val dm = dmOriginal.slice(dstOffset, length)
+          val alignedSize = if (RapidsGdsStore.isAligned(srcOffset)) {
```
jlowe, line 103: Is this fallback necessary in practice?  Curious about the scenario if it is needed, because the entire premise of assuming it's safe to read past the end of the buffer is predicated on the aligned allocator being used, and checking the start address for 4K alignment does not guarantee the aligned allocator is being used.

rongou: Turned this off for now.

jlowe (approved):

rongou:
Still doing some tweaking and test, changed to WIP for now.

rongou:
build

rongou:
build

jlowe (approved):

Issue #2461, opened 2021-05-20
[FEA] Report gpuOpTime instead of totalTime for project, filter, window, limit


andygrove:
**Is your feature request related to a problem? Please describe.**
See #2334 for the motivation for this. This issue is to update metrics for a number of operators to report their current metric as gpuOpTime instead of totalTime, since the current metric they are using is purely the cost of applying the operation on the GPU and does not include any costs of executing or fetching data from child operators.

- GpuProjectExec
- GpuFilterExec
- GpuWindowExec
- GpuLocalLimitExec / GpuGlobalLimitExec

**Describe the solution you'd like**
Report `gpuOpTime` instead of `totalTime`.

**Describe alternatives you've considered**
None

**Additional context**
None

Issue #2334, opened 2021-05-03
[FEA] Record GPU time and Fetch time separately, instead of recording Total Time


andygrove:
**Is your feature request related to a problem? Please describe.**
The `totalTime` metric can be confusing because it often (but not always) includes the execution time of the child operator(s). It would be better to separately record `fetchTime` and `gpuOpTime`, although this may be difficult or not possible in some cases.

Making this change will also provide better metrics that we can use when profiling queries or developing a better cost model for CBO.

**Describe the solution you'd like**

Review the following operators and determine whether we can improve the metrics to record the GPU time and fetch time separately. File new issues for those that we plan on doing this for.

- [x] GpuBroadcastExchangeExec - No action required
- [x] GpuBroadcastHashJoinExec - No action required
- [x] GpuBroadcastNestedLoopJoinExec - No action required
- [x] GpuBroadcastToCpuExec - No action required
- [x] GpuCartesianProductExec - No action required
- [x] GpuCoalesceBatchesExec - No action required
- [x] GpuHashAggregateExec - No action required
- [x] GpuRowToColumnarExec https://github.com/NVIDIA/spark-rapids/issues/2820
- [x] GpuShuffleCoalesceExec - No action required
- [x] GpuShuffledHashJoinExec - No action required
- [x] GpuSortExec - No action required
- [x] GpuTopN - No action required
- [x] HostColumnarToGpuExec - No action required
- [x] GpuUnionExec #2586
- [x] GpuExpandExec #2512
- [x] GpuGenerateExec #2512
- [x] GpuRangeExec #2512
- [x] GpuColumnarToRowExec #2269
- [x] GpuFilterExec #2461 
- [x] GpuGlobalLimitExec #2461
- [x] GpuLocalLimitExec #2461
- [x] GpuProjectExec #2461
- [x] GpuWindowExec #2461

**Describe alternatives you've considered**
None

**Additional context**
None

andygrove:
@nartal1 fyi - this may have an impact on the profiling tools

PR #2462, opened 2021-05-20
Report gpuOpTime instead of totalTime for project, filter, limit, and window

andygrove:

See #2334 for the motivation for this. This issue is to update metrics for  a number of operators to report their current metric as `gpuOpTime` instead of `totalTime`, since the current metric they are using is purely the cost of applying the operation on the GPU and does not include any costs of executing or fetching data from child operators.

- GpuProjectExec
- GpuFilterExec
- GpuWindowExec
- GpuLocalLimitExec / GpuGlobalLimitExec

Closes #https://github.com/NVIDIA/spark-rapids/issues/2461

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
@@ -35,8 +35,8 @@ import org.apache.spark.sql.vectorized.{ColumnarBatch, ColumnVector}
 
 object GpuProjectExec extends Arm {
   def projectAndClose[A <: Expression](cb: ColumnarBatch, boundExprs: Seq[A],
-      gpuOpTime: GpuMetric): ColumnarBatch = {
-    val nvtxRange = new NvtxWithMetrics("ProjectExec", NvtxColor.CYAN, gpuOpTime)
+      totalTime: GpuMetric): ColumnarBatch = {
+    val nvtxRange = new NvtxWithMetrics("ProjectExec", NvtxColor.CYAN, totalTime)
     try {
       project(cb, boundExprs)
     } finally {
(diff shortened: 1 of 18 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
```diff
@@ -93,6 +93,13 @@ case class GpuProjectExec(
    child: SparkPlan
  ) extends UnaryExecNode with GpuExec {
 
+  // we do not record totalTime for this operator
```
jlowe, line 96: This smells like boilerplate that's going to be replicated quite a few times as this change ripples through operators.  Should this be moved to a `GpuExec` object method for reuse?

andygrove: Yes, one option would be to remove `totalTime` from the default set of metrics and explicitly add it as an `additionalMetric` into all the operators that currently use it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
```diff
@@ -35,8 +35,8 @@ import org.apache.spark.sql.vectorized.{ColumnarBatch, ColumnVector}
 
 object GpuProjectExec extends Arm {
   def projectAndClose[A <: Expression](cb: ColumnarBatch, boundExprs: Seq[A],
-      gpuOpTime: GpuMetric): ColumnarBatch = {
-    val nvtxRange = new NvtxWithMetrics("ProjectExec", NvtxColor.CYAN, gpuOpTime)
+      totalTime: GpuMetric): ColumnarBatch = {
+    val nvtxRange = new NvtxWithMetrics("ProjectExec", NvtxColor.CYAN, totalTime)
```
jlowe: Why was this change made?  This is a utility method that will be called elsewhere, and this shouldn't be passed a total time metric if one exists.  It should be passed a metric designed to measure time spent on the GPU, as the original parameter name implied.

andygrove: Thanks. I had changed this in the first commit and then inadvertently changed it back in the second commit when merging changes from the other PR. Updated now.

jlowe (approved):

jlowe:
build

andygrove:
build

jlowe (approved):

revans2 (approved):

Issue #2465, opened 2021-05-21
[FEA] GpuFilterExec should report gpuOpTime not totalTime


andygrove:
Is your feature request related to a problem? Please describe.
See #2334 for the motivation for this. This issue is to update GpuFilterExec to report its current metric as gpuOpTime instead of totalTime, since the current metric is purely the cost of applying the filter on the GPU and does not include any costs of executing or fetching data from child operators.

Describe the solution you'd like
Report gpuOpTime instead of totalTime.

Describe alternatives you've considered
None

Additional context
None

andygrove:
Merged with https://github.com/NVIDIA/spark-rapids/issues/2461

PR #2466, opened 2021-05-21
GpuFilterExec report gpuOpTime instead of totalTime

andygrove:

See #2334 for the motivation for this. This issue is to update GpuFilterExec to report its current metric as gpuOpTime instead of totalTime, since the current metric is purely the cost of applying the filter on the GPU and does not include any costs of executing or fetching data from child operators.

Closes https://github.com/NVIDIA/spark-rapids/issues/2465

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
@@ -181,6 +181,12 @@ object GpuFilter extends Arm {
 case class GpuFilterExec(condition: Expression, child: SparkPlan)
     extends UnaryExecNode with GpuPredicateHelper with GpuExec {
 
+  // we do not record totalTime for this operator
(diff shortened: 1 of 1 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
```diff
@@ -181,6 +181,12 @@ object GpuFilter extends Arm {
 case class GpuFilterExec(condition: Expression, child: SparkPlan)
     extends UnaryExecNode with GpuPredicateHelper with GpuExec {
 
+  // we do not record totalTime for this operator
```
jlowe, line 184: Similar [comment](https://github.com/NVIDIA/spark-rapids/pull/2462#discussion_r636961988) as I made in #2462 about refactoring this for reuse.  May be simpler to just do them both in the same PR since the change is so small.

andygrove:
Closing this and applying changes to https://github.com/NVIDIA/spark-rapids/pull/2462

Issue #2470, opened 2021-05-21
[FEA] SortOrder supports arrays


wbo4958:
**Is your feature request related to a problem? Please describe.**
``` SQL
        SELECT
            a,
            b,
            c,
            d,
            LEAD(d, 1) OVER (PARTITION BY a ORDER BY b,c,d ASC) lead_array_1
        FROM table
```

if the type of column `d`  is Array, then any expressions and plans with SortOrder will not be accelerated by GPU

``` console
!Exec <WindowExec> cannot run on GPU because not all expressions can be replaced
  @Expression <Alias> lead(d#3, 1, null) windowspecdefinition(a#0L, b#1L ASC NULLS FIRST, c#2L ASC NULLS FIRST, d#3 ASC NULLS FIRST, specifiedwindowframe(RowFrame, 1, 1)) AS lead_array_1#15 could run on GPU
    @Expression <WindowExpression> lead(d#3, 1, null) windowspecdefinition(a#0L, b#1L ASC NULLS FIRST, c#2L ASC NULLS FIRST, d#3 ASC NULLS FIRST, specifiedwindowframe(RowFrame, 1, 1)) could run on GPU
      @Expression <Lead> lead(d#3, 1, null) could run on GPU
      !Expression <WindowSpecDefinition> windowspecdefinition(a#0L, b#1L ASC NULLS FIRST, c#2L ASC NULLS FIRST, d#3 ASC NULLS FIRST, specifiedwindowframe(RowFrame, 1, 1)) cannot run on GPU because expression SortOrder d#3 ASC NULLS FIRST produces an unsupported type ArrayType(IntegerType,true)
        @Expression <AttributeReference> a#0L could run on GPU
        @Expression <SortOrder> b#1L ASC NULLS FIRST could run on GPU
          @Expression <AttributeReference> b#1L could run on GPU
        @Expression <SortOrder> c#2L ASC NULLS FIRST could run on GPU
```

ttnghia:
Sorting non-nested arrays is now supported in libcudf so I believe that this can be partially fixed if needed. Support for sorting nested arrays may need to wait further.

PR #2471, opened 2021-05-21
skip test_window_aggs_for_rows_lead_lag_on_arrays

wbo4958:
If some rows of order-by columns (it's `a,b,c` in the test) are equal,
then it may fail because CPU and GPU can't guarantee the order for the
same rows, while lead/lag is typically depending on row's order.

The solution is we should add the aggregation column `d` and the default column
`d_default` columns into the order-by to guarantee the order. But for now,
sorting on array has not been supported yet, see
https://github.com/NVIDIA/spark-rapids/issues/2470.

So this PR just skip the test


wbo4958:
build

pxLi (approved):

revans2:
So fix this test instead of skipping it!  There are ways to generate data that has no ambiguity in the order.

PR #2472, opened 2021-05-21
Fix hash_aggregate_test

### LINKED ISSUE #2404: [BUG] test_hash_pivot_groupby_nan_fallback failed on Dataproc
**Describe the bug**

Build Id: rapids_it-Dataproc/178

FAILED integration_tests/src/main/python/hash_aggregate_test.py::test_hash_pivot_groupby_nan_fallback[[('a', RepeatSeq(Float)), ('b', Float), ('c', Long)]][IGNORE_ORDER({'local': True}), INCOMPAT, APPROXIMATE_FLOAT, ALLOW_NON_GPU(HashAggregateExec,PivotFirst,AggregateExpression,Alias,GetArrayItem,Literal,ShuffleExchangeExec,HashPartitioning,KnownFloatingPointNormalized,NormalizeNaNAndZero)]
FAILED integration_tests/src/main/python/hash_aggregate_test.py::test_struct_groupby_count[Struct(not_null)(('a', Struct(('aa', Integer),('ab', Integer))))][ALLOW_NON_GPU(ANY)]
FAILED integration_tests/src/main/python/repart_test.py::test_round_robin_sort_fallback[[('a', Array(String))]][IGNORE_ORDER({'local': True}), ALLOW_NON_GPU(ShuffleExchangeExec,RoundRobinPartitioning)]

 =================================== FAILURES ===================================
   test_hash_pivot_groupby_nan_fallback[[('a', RepeatSeq(Float)), ('b', Float), ('c', Long)]] 
  
  data_gen = [('a', RepeatSeq(Float)), ('b', Float), ('c', Long)]
  
      @approximate_float
      @ignore_order(local=True)
      @allow_non_gpu(HashAggregateExec', PivotFirst', AggregateExpression', Alias', GetArrayItem',
              Literal', ShuffleExchangeExec', HashPartitioning', KnownFloatingPointNormalized',
              NormalizeNaNAndZero')
      @incompat
      @pytest.mark.parametrize(data_gen', [_grpkey_floats_with_nulls_and_nans], ids=idfn)
      def test_hash_pivot_groupby_nan_fallback(data_gen):
  >       assert_gpu_fallback_collect(
              lambda spark: gen_df(spark, data_gen, length=100)
                  .groupby(a')
                  .pivot(b')
                  .agg(f.sum(c')),
              "PivotFirst",
              conf=_nans_float_conf)
  
  integration_tests/src/main/python/hash_aggregate_test.py:293:

sperlingxx:
Fix #2404.

In my opinion, the root cause of this issue seems to be incorrect plan capture. It happens occasionally, since we produce two SQL jobs in the failed test case.

### DIFF
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -20,7 +20,7 @@ from functools import reduce
 from pyspark.sql.types import *
 from marks import *
 import pyspark.sql.functions as f
-from spark_session import is_before_spark_311
+from spark_session import is_before_spark_311, with_cpu_session
 
 _no_nans_float_conf = {'spark.rapids.sql.variableFloatAgg.enabled': 'true',
                        'spark.rapids.sql.hasNans': 'false',
@@ -281,7 +281,6 @@ def test_hash_grpby_pivot(data_gen, conf):
             .agg(f.sum('c')),
         conf=conf)
 
-
 @approximate_float
 @ignore_order(local=True)
 @allow_non_gpu('HashAggregateExec', 'PivotFirst', 'AggregateExpression', 'Alias', 'GetArrayItem',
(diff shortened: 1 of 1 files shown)


sperlingxx:
build

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -290,10 +289,19 @@ def test_hash_grpby_pivot(data_gen, conf):
 @incompat
 @pytest.mark.parametrize('data_gen', [_grpkey_floats_with_nulls_and_nans], ids=idfn)
 def test_hash_pivot_groupby_nan_fallback(data_gen):
+    # collect values to pivot in previous, to avoid this preparation job being captured
+    def fetch_pivot_values(spark):
+        max_values = spark._jsparkSession.sessionState().conf().dataFramePivotMaxValues()
```
revans2, line 294: nit: I am a little confused why we need max values (default 10,000) if we are only creating 100 values at the most?

revans2 (approved):
Just one nit, and it is not a big one. It would also be nice to have a follow on issue, even if it is low priority, to understand better why the plan capture is failing. There are several other places where this could pop up if we are not careful and it would be good to understand it so we can deal with it properly.

nartal1 (approved):

sperlingxx:
> Just one nit, and it is not a big one. It would also be nice to have a follow on issue, even if it is low priority, to understand better why the plan capture is failing. There are several other places where this could pop up if we are not careful and it would be good to understand it so we can deal with it properly.

I filed the follow on issue #2485

Issue #2474, opened 2021-05-21
[BUG] when ucp listener enabled we bind 16 times always


abellina:
Noticed this when looking at more error handling using UCX's UCPListener:
```
[1621604466.240135] [test-server:9155 :0]   tcp_listener.c:134  UCX  DEBUG created a TCP listener 0x7f48d00950c0 on cm 0x7f48d00801f0 with fd: 389 listening on 192.168.50.80:36221
[1621604466.240202] [test-server:9155 :0]   tcp_listener.c:134  UCX  DEBUG created a TCP listener 0x7f48d0095890 on cm 0x7f48d00801f0 with fd: 390 listening on 192.168.50.80:36222
[1621604466.240257] [test-server:9155 :0]   tcp_listener.c:134  UCX  DEBUG created a TCP listener 0x7f48d007d2b0 on cm 0x7f48d00801f0 with fd: 391 listening on 192.168.50.80:36223
[1621604466.240289] [test-server:9155 :0]   tcp_listener.c:134  UCX  DEBUG created a TCP listener 0x7f48d007d280 on cm 0x7f48d00801f0 with fd: 392 listening on 192.168.50.80:36224
[1621604466.240320] [test-server:9155 :0]   tcp_listener.c:134  UCX  DEBUG created a TCP listener 0x7f48d0095ce0 on cm 0x7f48d00801f0 with fd: 393 listening on 192.168.50.80:36225
[1621604466.240351] [test-server:9155 :0]   tcp_listener.c:134  UCX  DEBUG created a TCP listener 0x7f48d0095e30 on cm 0x7f48d00801f0 with fd: 394 listening on 192.168.50.80:36226
[1621604466.240384] [test-server:9155 :0]   tcp_listener.c:134  UCX  DEBUG created a TCP listener 0x7f48d0096120 on cm 0x7f48d00801f0 with fd: 395 listening on 192.168.50.80:36227
[1621604466.240415] [test-server:9155 :0]   tcp_listener.c:134  UCX  DEBUG created a TCP listener 0x7f48d0096270 on cm 0x7f48d00801f0 with fd: 396 listening on 192.168.50.80:36228
[1621604466.240446] [test-server:9155 :0]   tcp_listener.c:134  UCX  DEBUG created a TCP listener 0x7f48d00963c0 on cm 0x7f48d00801f0 with fd: 397 listening on 192.168.50.80:36229
[...]

PR #2476, opened 2021-05-21
Fix for UCP Listener created spark.port.maxRetries times

abellina:

Closes https://github.com/NVIDIA/spark-rapids/issues/2474

This bug shows when the UCP listener connection method is enabled (this is an internal config, and is disabled by default). It was just connecting up to all the retry attempts, instead of exiting once the listener could bind.

### DIFF
--- a/shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
+++ b/shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
@@ -180,16 +180,19 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
           // TODO: remove this once ucx1.11 with random port selection would be released
           1024 + Random.nextInt(65535 - 1024)
         }
-        for (i <- 0 until maxRetries) {
-          val sockAddress = new InetSocketAddress(executor.host, startPort + i)
+        var attempt = 0
+        while (listener.isEmpty && attempt < maxRetries) {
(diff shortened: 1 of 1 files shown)


abellina:
@petro-rudenko fyi

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -180,16 +180,19 @@ class UCX(executor: BlockManagerId, rapidsConf: RapidsConf) extends AutoCloseabl
...
             listener = Option(worker.newListener(ucpListenerParams))
           } catch {
-            case ex: UcxException =>
-              logDebug(s"Failed to bind UcpListener on $sockAddress")
+            case _: UcxException =>
+              logDebug(s"Failed to bind UcpListener on $sockAddress. " +
+                s"Attempt ${attempt + 1} out of $maxRetries.")
               listener = None
           }
+          attempt += 1
```
jlowe, line 195: Nit: Move this to the top of the loop and we won't need to "fix it" in the log message.

jlowe (dismissed):
Minor nit but otherwise lgtm.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
abellina: Made that change.

jlowe:
build

petro-rudenko (approved):
Thanks

Issue #63, opened 2020-05-29
[FEA] support ConcatWs sql function


revans2:
**Is your feature request related to a problem? Please describe.**
it would be great to support the `concat_ws` SQL function

https://github.com/rapidsai/cudf/issues/3726 was filed to get support from cudf.

sameerz:
Follow up work on cudf, concatenating arrays of strings: https://github.com/rapidsai/cudf/issues/7727

tgravescs:
The Spark behavior of concatws:

Separator parameter Api differences:
- SQL can pass column of strings as separator
- python and scala api's, only string as separator

Null Behavior:
- if separator is NULL, columns are null
- if column value is null it leaves off separator and value
- if passing in array of all nulls, then its left out without separator
- if array contains a null, it skips it and leaves separator off

Behavior:
- 0 input columns with separate specified return column of empty strings
- 1 column just return column unless Null separator then need column of null match number of rows
- if column value is empty is puts on separator
- if passing in empty array, then its left off without separator
- if passing array with empty strings, separator included
- 2 or more columns add in separator between column values
- For SQL if column specified as separator, the separator in the corresponding row is used

tgravescs:
Note that concat and concat_ws have different behavior for nulls when all rows are null:

```
+-------+
|nullcol|
+-------+
|   null|
|notnull|
+-------+

>>> spark.sql("select concat_ws('-', nullcol, nullcol) as res from df").show(truncate=False)
+---------------+
|res            |
+---------------+
|               |
|notnull-notnull|
+---------------+

>>> spark.sql("select concat(nullcol, nullcol) as res from df").show(truncate=False)
+--------------+
|res           |
+--------------+
|null          |
|notnullnotnull|
+--------------+
```

tgravescs:
similar cudf behavior for arrays with nulls doesn't match Spark, cudf will put null if any elements in array null, spark skips them. if all nulls, then get empty string

```
>>> res = spark.sql("select concat_ws('-', array(1, 2, null, 4)) as res from df")
>>> res.show()
+-----+
|  res|
+-----+
|1-2-4|
|1-2-4|
+-----+

>>> spark.sql("select concat_ws('-', array(null, null)) as res from df").show(truncate=False)
+---+
|res|
+---+
|   |
|   |
+---+
```

tgravescs:
I discovered another weird case with the CPU where is you are concatenating an empty array and then another value, it leaves off the separator:

```
[...]

PR #2479, opened 2021-05-21
Support concat with separator on GPU

tgravescs:
This depends on CUDF PR: https://github.com/rapidsai/cudf/pull/8289

fixes https://github.com/NVIDIA/spark-rapids/issues/63

Note the issue has a bunch of Spark behavior for concat_ws as well as some examples if you want to see them. (https://github.com/NVIDIA/spark-rapids/issues/63#issuecomment-840668148).
You can pay attention to the null handling as that is what we found some incompatibilities with in CUDF. 

At a high level we add a new GpuConcatWs expression that supports taking Strings or Arrays or Strings and returns a String column.   If only a separator is specified we fall back to the CPU, otherwise as long as 1 other column with the separator we process it.  The code handles 2 types of separators, scalar and the spark sql api allows a column of separators.  We use the corresponding CUDF api based on the separator passed in.  concat_ws also allows an Array of Strings, so the strings within a single Array[String] column have to be concatenated first using a different CUDF api and then those are combined with any other columns to get the final output.

I ran a test on larger customer dataset and it helped mostly due to not going off gpu and then coming back.  80million rows, concatenating 2 sting columns, gpuconcatws took around 200ms.


tgravescs:
build

tgravescs:
oops, build will fail until cudf issue in.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -263,6 +267,127 @@ case class GpuStringTrimRight(column: Expression, trimParameters: Option[Express
...
+          s" $other. Please convert it to a GpuScalar or a GpuColumnVector before returning.")
+      }
+  }
+
+  override def columnarEval(batch: ColumnarBatch): Any = {
+    val numRows = batch.numRows()
+    var sep: Either[GpuScalar, GpuColumnVector] = null
+    try {
+      // get the separator and check for null scalar separator
+      val sepScalarNullOpt = children.head.columnarEval(batch) match {
```
jlowe, line 349: This code may be cleaner using `withResourceIfAllowed(children.head.columnarEval(batch))` to make sure we always close it.  On a related note, I think `sepScalar` leaks.

andygrove, line 349: This block returns `sepScalarNullOpt` but also sets the mutable `sep` as a side-effect. Could we just return a tuple of `(sep, sepScalarNullOpt)` instead?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -263,6 +267,127 @@ case class GpuStringTrimRight(column: Expression, trimParameters: Option[Express
...
+              columns += resolveColumnVectorAndConcatArrayCols(_, numRows, sep, batch).getBase
+            }
+            sep match {
+              case Left(sepScalar) =>
+                if (columns.size == 1) {
+                  processSingleColScalarSep(columns.head)
+                } else {
+                  stringConcatSeparatorScalar(columns, sepScalar)
+                }
+              case Right(sepVec) => stringConcatSeparatorVector(columns, sepVec)
```
revans2, line 376: FYI a change was just merged into cudf that added an assert for the scalar separator case that broke the java tests. Not sure if there is an assertion for the vector case too (it looks like we don't have unit tests for that)

tgravescs: this is handled actually above to process the scalar separator.  If I missed a case let me know

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
tgravescs: wow, certainly looks like it, weird I never saw it reported as leaked, I'll update

tgravescs: so originally I had this as separate function but that has potential to leak the sep.  as far as I know you can't assign an existing var to tuple returned, if you know a way let me know.  If we don't set sep directly then there is a chance it leaks as the finally wouldn't catch it.  its very small but was just trying to set such that finally would be able to clean up.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -263,6 +267,127 @@ case class GpuStringTrimRight(column: Expression, trimParameters: Option[Express
...
+
+  override def columnarEval(batch: ColumnarBatch): Any = {
+    val numRows = batch.numRows()
+    var sep: Either[GpuScalar, GpuColumnVector] = null
+    try {
+      // get the separator and check for null scalar separator
+      val sepScalarNullOpt = children.head.columnarEval(batch) match {
+        case sepScalar: GpuScalar =>
+          sep = Left(sepScalar)
+          if (sepScalar.getBase.isValid() == false) {
```
andygrove, line 352:
```suggestion
          if (!sepScalar.getBase.isValid()) {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -263,6 +267,127 @@ case class GpuStringTrimRight(column: Expression, trimParameters: Option[Express
...
+            None
+          }
+        case sepVec: GpuColumnVector =>
+          sep = Right(sepVec)
+          None
+      }
+      sepScalarNullOpt match {
+        case Some(col) => col
+        case None =>
+          withResource(ArrayBuffer.empty[ColumnVector]) { columns =>
```
andygrove, line 365: This could be

```scala
sepScalarNullOpt.getOrElse {
  withResource(ArrayBuffer.empty[ColumnVector]) { columns =>
    ...
  }
}
```   
```

File: sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala (hunk 1 above)
Comment (andygrove) [reply sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala]: I played with a couple of ideas but they turned out to be more trouble than they were worth

File: sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala [hunk 5]
```diff
@@ -263,6 +267,127 @@ case class GpuStringTrimRight(column: Expression, trimParameters: Option[Express
...
+            }
+          }
+      }
+    } finally {
+      if (sep != null) {
+        sep match {
+          case Right(sepVec) => sepVec.close()
+          case Left(_) =>
+        }
+      }
```
andygrove: This might be more idiomatic Scala style:

```suggestion
      Option(sep).foreach {
        case Right(sepVec) => sepVec.close()
        case _ =>
      }
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -263,6 +267,127 @@ case class GpuStringTrimRight(column: Expression, trimParameters: Option[Express
...
+          sep = Right(sepVec.incRefCount())
+          None
+      }
+      sepScalarNullOpt match {
+        case Some(col) => col
+        case None =>
+          withResource(ArrayBuffer.empty[ColumnVector]) { columns =>
+            children.tail.foreach {
+              columns += resolveColumnVectorAndConcatArrayCols(_, numRows, sep, batch).getBase
+            }
```
andygrove, line 368: nit: I think it would be better to use `map` here rather than a `foreach` with side effects.

```suggestion
          columns += children.tail.map {
            resolveColumnVectorAndConcatArrayCols(_, numRows, sep, batch).getBase
          }
```

tgravescs:
CUDF changes all merged in now.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -263,6 +267,123 @@ case class GpuStringTrimRight(column: Expression, trimParameters: Option[Express
...
+          s" $other. Please convert it to a GpuScalar or a GpuColumnVector before returning.")
+      }
+  }
+
+  override def columnarEval(batch: ColumnarBatch): Any = {
+    val numRows = batch.numRows()
+    var sep: Either[GpuScalar, GpuColumnVector] = null
+    try {
+      // get the separator and check for null scalar separator
+      val sepScalarNullOpt = withResourceIfAllowed(children.head.columnarEval(batch)) {
```
jlowe, line 349: Is there a point to using `withResourceIfAllowed` if every path within is going to increment the refcount?  My original suggestion to use it was so that we didn't need to close the scalar in the finally block, but since it's going out of it's way to do that anyway, it's wasteful to try to close something that will never actually close here.  We should either wrap the whole code block in the `withResourceIfAllowed` so the expr evaluation is always closed at the end, whether it's a scalar or vector, or remove it and the increfcounts.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2511,6 +2511,27 @@ object GpuOverrides {
...
+            // If only a separator specified and its a column, Spark returns an empty 
+            // string for all entries unless they are null, then it returns null.
+            // This seems like edge case so instead of handling on GPU just fallback.
+            willNotWorkOnGpu("Only specifying separator column not supported on GPU")
+          }
+        }
+        override final def convertToGpu(): GpuExpression = {
+          convertToGpu(childExprs.map(_.convertToGpu()))
+        }
+        def convertToGpu(child: Seq[Expression]): GpuExpression = GpuConcatWs(child)
```
revans2, line 2533: nit: Not really sure why this is separated into two functions.

```
override final def convertToGpu(): GpuExpression =
  GpuConcatWs(childExprs.map(_.convertToGpu()))
```

feels just as clear to me and is smaller.

tgravescs:
build

tgravescs:
oops, still need to remove one thing

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -263,6 +267,119 @@ case class GpuStringTrimRight(column: Expression, trimParameters: Option[Express
...
+
+  private def resolveColumnVectorAndConcatArrayCols(
+      expr: Expression,
+      numRows: Int,
+      colOrScalarSep: Any,
+      batch: ColumnarBatch): GpuColumnVector = {
+    withResourceIfAllowed(expr.columnarEval(batch)) {
+      case vector: GpuColumnVector =>
+        vector.dataType() match {
+          case ArrayType(st: StringType, nullable) => concatArrayCol(colOrScalarSep, vector.getBase)
```
revans2, line 326: nit: nullable is not used, I thought generally we wanted to use `_` for those

```
case ArrayType(st: StringType, _) => concatArrayCol(colOrScalarSep, vector.getBase)
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -263,6 +267,119 @@ case class GpuStringTrimRight(column: Expression, trimParameters: Option[Express
...
+          case _ => vector.incRefCount()
+        }
+      case s: GpuScalar =>
+        s.dataType match {
+          case ArrayType(st: StringType, nullable) =>
```
revans2, line 331: here too nullable is not used

revans2 (approved):
Just two very small nits

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -263,6 +267,119 @@ case class GpuStringTrimRight(column: Expression, trimParameters: Option[Express
...
+      case sepScalar: GpuScalar =>
+        if (!sepScalar.getBase.isValid()) {
+          // if null scalar separator just return a column of all nulls
+          Some(GpuColumnVector.from(sepScalar, numRows, dataType))
+        } else {
+          None
+        }
+      case sepVec: GpuColumnVector =>
+        None
+    }
```
andygrove, line 356: nit: this could be simplified 
```suggestion
    colOrScalarSep match {
      case sepScalar: GpuScalar if !sepScalar.getBase.isValid =>
        // if null scalar separator just return a column of all nulls
        Some(GpuColumnVector.from(sepScalar, numRows, dataType))
      case _ =>
        None
    }
```

jlowe (dismissed):

tgravescs:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -263,6 +267,115 @@ case class GpuStringTrimRight(column: Expression, trimParameters: Option[Express
...
+      }
+  }
+
+  private def checkScalarSeparatorNull(colOrScalarSep: Any,
+      numRows: Int): Option[GpuColumnVector] = {
+    colOrScalarSep match {
+      case sepScalar: GpuScalar if (!sepScalar.getBase.isValid()) =>
+        // if null scalar separator just return a column of all nulls
+        Some(GpuColumnVector.from(sepScalar, numRows, dataType))
+      case sepVec: GpuColumnVector =>
```
andygrove, line 350: This is a bit different from my suggest change and this doesn't handle the GpuScalar case where the base is valid

tgravescs:
build

andygrove (approved):

revans2 (approved):

Issue #2372, opened 2021-05-08
[BUG] FAILED ../../src/main/python/udf_cudf_test.py::test_window


pxLi:
**Describe the bug**

db_nightly build ID: 169 170

_RAPIDS Accelerator build: {version=0.6.0-SNAPSHOT, user=ubuntu, url=https://github.com/NVIDIA/spark-rapids.git, date=2021-05-08T05:22:20Z, revision=710296c32aa6033a4d25928fe0366a128061ec20, cudf_version=0.20-SNAPSHOT, branch=HEAD}
cudf build: {version=0.20-SNAPSHOT, user=, date=2021-05-08T02:05:06Z, revision=3813d9bed12358b6e29710f1f28ede65f7ff5da8, branch=HEAD}_

```bash
[2021-05-08T06:05:59.088Z] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[2021-05-08T06:05:59.088Z] 
[2021-05-08T06:05:59.088Z] answer = 'xro2229'
[2021-05-08T06:05:59.088Z] gateway_client = <py4j.java_gateway.GatewayClient object at 0x7f3f57235d10>
[2021-05-08T06:05:59.088Z] target_id = 'o2226', name = 'collectToPython'
[2021-05-08T06:05:59.088Z] 
[2021-05-08T06:05:59.088Z]     [94mdef[39;49;00m [92mget_return_value[39;49;00m(answer, gateway_client, target_id=[94mNone[39;49;00m, name=[94mNone[39;49;00m):
[2021-05-08T06:05:59.088Z]         [33m"""Converts an answer received from the Java gateway into a Python object.[39;49;00m
[2021-05-08T06:05:59.088Z]     [33m[39;49;00m
[2021-05-08T06:05:59.088Z]     [33m    For example, string representation of integers are converted to Python[39;49;00m
[2021-05-08T06:05:59.088Z]     [33m    integer, string representation of objects are converted to JavaObject[39;49;00m
[2021-05-08T06:05:59.088Z]     [33m    instances, etc.[39;49;00m
[2021-05-08T06:05:59.088Z]     [33m[39;49;00m
[2021-05-08T06:05:59.088Z]     [33m    :param answer: the string returned by the Java gateway[39;49;00m
[...]

PR #2484, opened 2021-05-24
Fix the failing test `test_window` on Databricks

firestarman:
This small PR is to fix the failing test `test_window` on Databricks by making the `outReference` be lazy.

Because on Databricks, it will get some invalid expressions (e.g.  `none#0L`, `none@1L`.) in the `projectList` if binding the references on the driver side. 

closes https://github.com/NVIDIA/spark-rapids/issues/2372

### DIFF
--- a/integration_tests/src/main/python/udf_cudf_test.py
+++ b/integration_tests/src/main/python/udf_cudf_test.py
@@ -14,7 +14,7 @@
 
 import pytest
 
-from conftest import is_at_least_precommit_run, is_databricks_runtime
+from conftest import is_at_least_precommit_run
 
 from pyspark.sql.pandas.utils import require_minimum_pyarrow_version, require_minimum_pandas_version
 try:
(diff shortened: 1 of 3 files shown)


firestarman:
Verification on DB is running ... https://blossom.nvidia.com/sw-gpu-spark-jenkins/view/Testing/job/lc-db/ .

firestarman:
build

pxLi (approved):
LGTM. 

still not sure about the root cause of abnormal expression, we can get back to this if we saw something-related again~

revans2 (commented):
I am happy that this was fixed, but it would have been nice to add a comment to the `outReferences` explaining why it needs to be lazy. Often when there a subtle bugs like this it is easy for someone to accidentally undo the fix without a warning that there is a problem to be aware of.

tgravescs:
Agree with Bobby, I tried to make sure in the new shim that we document changes made specific to Databricks, so if we can followup and add comment that would be great

firestarman:
Made a PR https://github.com/NVIDIA/spark-rapids/pull/2496 to add the comments.

Issue #2468, opened 2021-05-21
[FEA] Remove the null replacement in `GpuConditionalExpression`


firestarman:
According to [ the comment here ](https://github.com/NVIDIA/spark-rapids/blob/branch-21.06/sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala#L38),
```Scala
// TODO: This null replacement is no longer necessary when
// https://github.com/rapidsai/cudf/issues/3856 is fixed.
```
Suppose now it is safe to remove the null replacement in `GpuConditionalExpression` since the issue https://github.com/rapidsai/cudf/issues/3856 has been fixed by https://github.com/rapidsai/cudf/pull/3911.

PR #2486, opened 2021-05-24
Remove the null replacement in `computePredicate`

firestarman:
This small PR is to remove the  null replacement in [GpuConditionalExpression.computePredicate](https://github.com/NVIDIA/spark-rapids/blob/branch-21.06/sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala#L38).

Since the issue https://github.com/rapidsai/cudf/issues/3856 mentioned in its comment has been fixed already by the PR https://github.com/rapidsai/cudf/pull/3911.

closes https://github.com/NVIDIA/spark-rapids/issues/2468

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
@@ -25,21 +25,13 @@ import org.apache.spark.sql.types.{BooleanType, DataType}
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
 abstract class GpuConditionalExpression extends ComplexTypeMergingExpression with GpuExpression {
-  private def computePredicate(
-      batch: ColumnarBatch,
-      predicateExpr: Expression): GpuColumnVector = {
-    predicateExpr.columnarEval(batch) match {
-      case gcv: GpuColumnVector => gcv
-      case _ => throw new IllegalStateException("Predicate result is not a column")
-    }
-  }
 
   protected def computeIfElse(
       batch: ColumnarBatch,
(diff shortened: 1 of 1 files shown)


firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -28,23 +28,9 @@ abstract class GpuConditionalExpression extends ComplexTypeMergingExpression wit
   private def computePredicate(
```
revans2, line 28: nit: This gets called in just two places in the code, and they are in this same class.  Could we just replace the entire method with `GpuExpressionsUtils.columnarEvalToColumn`

revans2 (commented):
Overall the code looks fine, I just thing we could delete more of it and still get the same result.

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: Done.
Verified the Scala tests on my node and they passed.
Triggered pre-merge build to run the integration tests.

I still have a little concern of using the `GpuExpressionsUtils.columnarEvalToColumn` here, because it implicitly allows a `GpuScalar` as the result of the predicate expression, but the original code did not.

revans2: Yes, but if a GpuScalar shows up it will do the right thing, so it should be more robust than the old code.

revans2 (approved):

Issue #2407, opened 2021-05-13
[BUG] test_round_robin_sort_fallback failed


jlowe:
Last night's Dataproc integration test run saw a failure in test_round_robin_sort_fallback.  From the failure log:
```
06:40:46  [1m[31mE                   : java.lang.AssertionError: assertion failed: Could not find ShuffleExchangeExec in the GPU plan[0m
06:40:46  [1m[31mE                   GpuColumnarToRow false[0m
06:40:46  [1m[31mE                   +- GpuShuffleCoalesce 2147483647[0m
06:40:46  [1m[31mE                      +- GpuColumnarExchange gpuroundrobinpartitioning(2345), REPARTITION_WITH_NUM, [id=#616183][0m
06:40:46  [1m[31mE                         +- GpuProject [a#749732, 1 AS x#749734][0m
06:40:46  [1m[31mE                            +- GpuRowToColumnar TargetSize(2147483647)[0m
06:40:46  [1m[31mE                               +- *(1) Scan ExistingRDD[a#749732][0m
```

Full log for reference:
```
06:40:46  [31m[1m____________ test_round_robin_sort_fallback[[('a', Array(String))]] ____________[0m
06:40:46  
06:40:46  data_gen = [('a', Array(String))]
06:40:46  
06:40:46      [37m@allow_non_gpu[39;49;00m([33m'[39;49;00m[33mShuffleExchangeExec[39;49;00m[33m'[39;49;00m, [33m'[39;49;00m[33mRoundRobinPartitioning[39;49;00m[33m'[39;49;00m)
06:40:46      [37m@pytest[39;49;00m.mark.parametrize([33m'[39;49;00m[33mdata_gen[39;49;00m[33m'[39;49;00m, [[([33m'[39;49;00m[33ma[39;49;00m[33m'[39;49;00m, ArrayGen(string_gen))],
06:40:46          [([33m'[39;49;00m[33ma[39;49;00m[33m'[39;49;00m, StructGen([
06:40:46            ([33m'[39;49;00m[33ma_1[39;49;00m[33m'[39;49;00m, StructGen([
06:40:46              ([33m'[39;49;00m[33ma_1_1[39;49;00m[33m'[39;49;00m, int_gen),
06:40:46              ([33m'[39;49;00m[33ma_1_2[39;49;00m[33m'[39;49;00m, float_gen),
[...]

PR #2490, opened 2021-05-24
Add temporary logging for Dataproc round robin fallback issue

jlowe:
This adds temporary logging to help debug #2407.  The test is only failing in Dataproc and I am unable to reproduce it despite trying to follow the Dataproc scripts closely.  This logging should help explain whether we really are failing to fallback or if this is a case of capturing the wrong plan.

After this is merged, I'll file a followup to ensure the logging is removed before 21.06 is released.


jlowe:
build

revans2 (approved):
I assume that you know what logging you need/want.

Issue #2489, opened 2021-05-24
[BUG] Queries with window expressions fail when cost-based optimizer is enabled


andygrove:
**Describe the bug**

With the RAPIDS Accelerator cost-based optimizer enabled, queries with window expressions fail when we attempt to estimate data sizes, because `WindowFrame.dataType` throws `UnsupportedOperationException`.

```
[2021-05-22T11:55:11.887Z] java.lang.UnsupportedOperationException: dataType^M
[2021-05-22T11:55:11.887Z]      at org.apache.****.sql.catalyst.expressions.WindowFrame.dataType(windowExpressions.scala:168)^M
[2021-05-22T11:55:11.887Z]      at org.apache.****.sql.catalyst.expressions.WindowFrame.dataType$(windowExpressions.scala:168)^M
[2021-05-22T11:55:11.887Z]      at org.apache.****.sql.catalyst.expressions.SpecifiedWindowFrame.dataType(windowExpressions.scala:179)^M
[2021-05-22T11:55:11.887Z]      at com.nvidia.****.rapids.BaseExprMeta.dataType(RapidsMeta.scala:758)^M
[2021-05-22T11:55:11.887Z]      at com.nvidia.****.rapids.CpuCostModel.exprCost(CostBasedOptimizer.scala:282)^M
```

**Steps/Code to reproduce bug**
Run NDS query q89 with CBO enabled.

**Expected behavior**
Query should not fail.

**Environment details (please complete the following information)**
N/A

**Additional context**
None

revans2:
Ya there is some oddness with data type for a `WindowSpecDefinition`. For what ever reason they throw `nsupportedOperationException("dataType")`, so we probably need to have some special case code for them like we have [here](https://github.com/NVIDIA/spark-rapids/blob/a813edf7b65023fd104a653bb5eeee65c5d9ee23/sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala#L474-L480).

PR #2491, opened 2021-05-24
Fix regression in cost-based optimizer when calculating cost for Window operations

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/2489

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
@@ -262,8 +262,7 @@ class CpuCostModel(conf: RapidsConf) extends CostModel {
   }
 
   private def exprCost[INPUT <: Expression](expr: BaseExprMeta[INPUT], rowCount: Double): Double = {
-    if (MemoryCostHelper.isWindowExpr(expr)) {
-      // Window expressions are Unevaluable and accessing dataType causes an exception
+    if (MemoryCostHelper.isExcludedFromCost(expr)) {
       return 0
     }
 
@@ -312,8 +311,7 @@ class GpuCostModel(conf: RapidsConf) extends CostModel {
   }
 
   private def exprCost[INPUT <: Expression](expr: BaseExprMeta[INPUT], rowCount: Double): Double = {
-    if (MemoryCostHelper.isWindowExpr(expr)) {
-      // Window expressions are Unevaluable and accessing dataType causes an exception
+    if (MemoryCostHelper.isExcludedFromCost(expr)) {
       return 0
     }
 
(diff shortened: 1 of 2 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -351,6 +359,13 @@ object MemoryCostHelper {
   def calculateCost(dataSize: Long, memorySpeed: Double): Double = {
     (dataSize / GIGABYTE) / memorySpeed
   }
+
+  def isWindowExpr[INPUT <: Expression](expr: BaseExprMeta[INPUT]) = {
```
revans2, line 363: Can we change the name of this? A `WindowExpression` is something different that does have a type. It has children that might now, but the name to be feels broken.

andygrove: Would `isUnevaluableWindowExpr` be appropriate?

gerashegalov: Maybe this method can be a catch all for all expressions whose cost cannot be evaluated, and the name should not refer to a specific flavor of expressions?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -308,6 +312,10 @@ class GpuCostModel(conf: RapidsConf) extends CostModel {
   }
 
   private def exprCost[INPUT <: Expression](expr: BaseExprMeta[INPUT], rowCount: Double): Double = {
+    if (MemoryCostHelper.isWindowExpr(expr)) {
+      // Window expressions are Unevaluable and accessing dataType causes an exception
+      return 0
```
gerashegalov, line 317: Can we use `0d` for consistency with the uses of zero just below it.  I think`0.0` is actually more Scala-esque but consistency is most important.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -308,6 +312,10 @@ class GpuCostModel(conf: RapidsConf) extends CostModel {
   }
 
   private def exprCost[INPUT <: Expression](expr: BaseExprMeta[INPUT], rowCount: Double): Double = {
+    if (MemoryCostHelper.isWindowExpr(expr)) {
```
gerashegalov, line 315: We can also make CBO best effort to minimize effect on production systems:
```scala
def exprCost(expr) = Try(doExprCost(expr)).recover {
   case _ if (!RapidsConf.isTestEnabled) => 0.0 
 // add some diagnostics logging too, or have a counter for unsuccessful cost estimates
}.get
```

andygrove: This feature is experimental and disabled by default. The implementation is likely to change a lot still so I think it might be premature for this type of hardening.

revans2 (approved):

andygrove:
build

gerashegalov (approved):

PR #2493, opened 2021-05-24
improve window agg test for range numeric types

wbo4958:
This change makes the partition-by column to be repeatable

### DIFF
--- a/integration_tests/src/main/python/window_function_test.py
+++ b/integration_tests/src/main/python/window_function_test.py
@@ -58,19 +58,19 @@ _grpkey_decimals_with_nulls = [
     ('c', DecimalGen(precision=8, scale=3, nullable=True))]
 
 _grpkey_byte_with_nulls = [
-    ('a', IntegerGen()),
+    ('a', RepeatSeqGen(int_gen, length=20)),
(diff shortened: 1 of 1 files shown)


wbo4958:
build

integration_tests/src/main/python/window_function_test.py
```diff
@@ -58,19 +58,19 @@
     ('c', DecimalGen(precision=8, scale=3, nullable=True))]
 
 _grpkey_byte_with_nulls = [
-    ('a', IntegerGen()),
+    ('a', RepeatSeqGen(int_gen, length=20)),
```
firestarman, line 61: Could you have a comment about why making this change ?

wbo4958: The rows in partition-by columns should be repeatable, which is  a real-world case.

wbo4958: If the rows are always different, let's have an assumption, each group only has 1 row. then all the agg functions will work on only 1 row, which is not a really good test case.

wjxiz1992 (approved):
LGTM. 
nit: could put the important explanation in description body.

PR #2498, opened 2021-05-25
Add in basic support for scalar maps and allow nesting in named_struct

### LINKED ISSUE #2436: [FEA] Support nested types in CreateNamedStruct
We have basic support for CreateNamedStruct but we cannot put nested types into a struct because cudf does not support scalars of nested types. That is slowly changing and we should add in support for Structs, Arrays, and Maps to be children of names structs.

revans2:
This fixes #2436 but will need a follow on for scalar structs support in named_struct


revans2:
build

jlowe (approved):

PR #2503, opened 2021-05-25
Remove temporary logging and adjust test column names

### LINKED ISSUE #2494: Remove the round robin partitioning debug logging
#2490 added temporary logging to help debug #2407, and this issue tracks removing the logging before the 21.06 release.

jlowe:
Fixes #2494.

This removes the debug logging added in #2490.  It also adjusts the column names in `test_round_robin_sort_fallback` to disambiguate the schema from a very similar query being built in `test_repartition_df`.  As seen in the Dataproc test failures, the test that executes just before the one that fails ends up having the same schema, so it's not easy to tell if the test accidentally captured the plan from the previous query.  From the Dataproc test log before this change:
```
06:46:47  integration_tests/src/main/python/repart_test.py::test_repartition_df[4096-2345-[('a', Array(String))]][IGNORE_ORDER({'local': True})] [32mPASSED[0m[33m [ 81%][0m
06:46:47  integration_tests/src/main/python/repart_test.py::test_round_robin_sort_fallback[[('a', Array(String))]][IGNORE_ORDER({'local': True}), ALLOW_NON_GPU(ShuffleExchangeExec,RoundRobinPartitioning)] [31mFAILED[0m[31m [ 81%][0m
```

### DIFF
--- a/integration_tests/src/main/python/repart_test.py
+++ b/integration_tests/src/main/python/repart_test.py
@@ -81,7 +81,7 @@ def test_repartition_df(data_gen, num_parts, length):
                 'spark.sql.legacy.allowNegativeScaleOfDecimal': 'true'})
 
 @allow_non_gpu('ShuffleExchangeExec', 'RoundRobinPartitioning')
-@pytest.mark.parametrize('data_gen', [[('a', ArrayGen(string_gen))],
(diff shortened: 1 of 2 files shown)


jlowe:
build

revans2 (approved):

PR #2506, opened 2021-05-26
Support creating Scalars/ColumnVectors from utf8 strings directly.

firestarman:
Update to create Scalars/ColumnVectors from utf8 strings directly, since the issue https://github.com/rapidsai/cudf/issues/8137 has been solved.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
@@ -110,9 +110,7 @@ object GpuScalar extends Arm with Logging {
   /** Converts an element for nested lists */
   private def convertElementTo(element: Any, elementType: DataType): Any = elementType match {
     case _ if element == null => null
-    // StructData does not support utf8 string yet, so parses it here instead of calling the
-    //  `convertUTF8StringTo`. Tracked by https://github.com/rapidsai/cudf/issues/8137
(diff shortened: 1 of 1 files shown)


firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/literals.scala
```diff
@@ -110,9 +110,7 @@ object GpuScalar extends Arm with Logging {
   /** Converts an element for nested lists */
   private def convertElementTo(element: Any, elementType: DataType): Any = elementType match {
     case _ if element == null => null
-    // StructData does not support utf8 string yet, so parses it here instead of calling the
-    //  `convertUTF8StringTo`. Tracked by https://github.com/rapidsai/cudf/issues/8137
-    case StringType => element.asInstanceOf[UTF8String].toString
+    case StringType => element.asInstanceOf[UTF8String].getBytes
```
wjxiz1992, line 113: Only StringType are representable by Bytes? Bytes cannot be casted to other types? for example an IntegerType?

firestarman: Here is only for `StringType`, and it is array of byte inside an `UTF8String`.

wjxiz1992 (approved):
LGTM

PR #2508, opened 2021-05-26
Refactor the code for conditional expressions

firestarman:
Refactor the code for conditional expressions, and add tests for the case of `(column, column)` that was missing before.

### DIFF
--- a/integration_tests/src/main/python/conditionals_test.py
+++ b/integration_tests/src/main/python/conditionals_test.py
@@ -50,10 +50,16 @@ def test_case_when(data_gen):
         command = command.when(f.col('_b'+ str(x)), f.col('_c' + str(x)))
     command = command.otherwise(s1)
     data_type = data_gen.data_type
+    # `command` covers the case of (column, scalar), so the followings are for
+    # (column, column)
+    # (scalar, scalar)  -> the default `otherwise` is a scalar.
+    # (scalar, column)
+    # (others) in sequence.
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : gen_df(spark, gen).select(command,
+                f.when(f.col('_b0'), f.col('_c0')).otherwise(f.col('_c1')),
                 f.when(f.col('_b0'), s1),
-                f.when(f.col('_b0'), f.col('_c0')).otherwise(s1),
+                f.when(f.col('_b0'), s1).otherwise(f.col('_c0')),
                 f.when(f.col('_b0'), f.lit(None).cast(data_type)).otherwise(f.col('_c0')),
                 f.when(f.lit(False), f.col('_c0'))))
 
--- a/integration_tests/src/main/python/conditionals_test.py
+++ b/integration_tests/src/main/python/conditionals_test.py
@@ -39,7 +39,7 @@ def test_if_else(data_gen):
                 'IF(a, {}, c)'.format(null_lit)))
 
 @pytest.mark.parametrize('data_gen', all_gens, ids=idfn)
-def test_case_when_debug(data_gen):
+def test_case_when(data_gen):
     num_cmps = 20
     s1 = gen_scalar(data_gen, force_no_nulls=not isinstance(data_gen, NullGen))
     # we want lots of false
--- a/integration_tests/src/main/python/conditionals_test.py
+++ b/integration_tests/src/main/python/conditionals_test.py
@@ -28,7 +28,9 @@ def test_if_else(data_gen):
     null_lit = get_null_lit_string(data_gen.data_type)
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : three_col_df(spark, boolean_gen, data_gen, data_gen).selectExpr(
-                # A literal predicate is not supported yet
+                'IF(TRUE, b, c)',
+                'IF(TRUE, {}, {})'.format(s1, null_lit),
+                'IF(FALSE, {}, {})'.format(s1, null_lit),
                 'IF(a, b, c)',
                 'IF(a, {}, c)'.format(s1),
                 'IF(a, b, {})'.format(s2),
@@ -47,14 +49,15 @@ def test_case_when(data_gen):
     gen = StructGen(gen_cols, nullable=False)
     command = f.when(f.col('_b0'), f.col('_c0'))
     for x in range(1, num_cmps):
-        command = command.when(f.col('_b'+ str(x)), f.col('_c' + str(x)))
-    command = command.otherwise(s1)
+        command = command.when(f.col('_b'+ str(x)), f.col('_c' + str(x)))\
+                         .when(f.lit(True), f.col('_c' + str(x)))
+    command = command.when(f.lit(True), s1).otherwise(s1)
     data_type = data_gen.data_type
-    # `command` covers the case of (column, scalar), so the followings are for
-    # (column, column)
-    # (scalar, scalar)  -> the default `otherwise` is a scalar.
-    # (scalar, column)
-    # (others) in sequence.
+    # `command` covers the case of (column, scalar) for values, so the followings are for
+    #   (column, column)
+    #   (scalar, scalar)  -> the default `otherwise` is a scalar.
+    #   (scalar, column)
+    #   (others) in sequence.
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : gen_df(spark, gen).select(command,
                 f.when(f.col('_b0'), f.col('_c0')).otherwise(f.col('_c1')),
--- a/integration_tests/src/main/python/conditionals_test.py
+++ b/integration_tests/src/main/python/conditionals_test.py
@@ -39,7 +39,7 @@ def test_if_else(data_gen):
                 'IF(a, {}, c)'.format(null_lit)))
 
 @pytest.mark.parametrize('data_gen', all_gens, ids=idfn)
-def test_case_when(data_gen):
+def test_case_when_debug(data_gen):
     num_cmps = 20
     s1 = gen_scalar(data_gen, force_no_nulls=not isinstance(data_gen, NullGen))
     # we want lots of false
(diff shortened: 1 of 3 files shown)


firestarman:
build

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -16,84 +16,42 @@
...
-        }
-        GpuColumnVector.from(result, dataType)
-      } finally {
-        trueResult match {
-          case a: AutoCloseable => a.close()
-          case _ =>
+      falseValue: Any): GpuColumnVector = {
+    withResource(GpuExpressionsUtils.columnarEvalToColumn(predExpr, batch)) { pred =>
+      withResourceIfAllowed(trueExpr.columnarEval(batch)) { trueRet =>
+        withResourceIfAllowed(falseValue) { falseRet =>
```
revans2, line 35: Please move this to be the first thing that we do. Otherwise if there is an exception thrown when computing `trueExpr` or `predExpr` the false expression will leak.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -16,84 +16,42 @@
...
-          case u =>
-            throw new IllegalStateException(s"Unexpected inputs $u")
-        }
-        GpuColumnVector.from(result, dataType)
-      } finally {
-        trueResult match {
-          case a: AutoCloseable => a.close()
-          case _ =>
+      falseValue: Any): GpuColumnVector = {
+    withResource(GpuExpressionsUtils.columnarEvalToColumn(predExpr, batch)) { pred =>
```
revans2, line 33: This fixed an issue with how we deal with the predicate expression (It cannot be a literal).
Could you please remove the existing checks in GpuOverrides and add in some tests to make sure it is working?

1. https://github.com/NVIDIA/spark-rapids/blob/203769fc3669cd8e0ed76c152a29f01f451e112d/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala#L1741-L1743
2. https://github.com/NVIDIA/spark-rapids/blob/203769fc3669cd8e0ed76c152a29f01f451e112d/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala#L1700-L1703

firestarman:
please ignore the branch name.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: Nice catch. Updated

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: Done.

I also found if the `predicate` in `If` or `CaseWhen` is a single literal, seems the expression will be optimized, leaving only an `Alias`. 
Expression tree for  `'IF(a, b, c)'` .
```Bash
  *Expression <Alias> if (a#696) b#697 else c#698 AS (IF(a, b, c))#705 will run on GPU
    *Expression <If> if (a#696) b#697 else c#698 will run on GPU
```

Expression tree for `'IF(TRUE, b, c)'`.
```Bash
*Expression <Alias> b#697 AS (IF(true, b, c))#702 will run on GPU
```

firestarman:
build

integration_tests/src/main/python/conditionals_test.py
```diff
@@ -47,13 +49,20 @@ def test_case_when(data_gen):
     gen = StructGen(gen_cols, nullable=False)
     command = f.when(f.col('_b0'), f.col('_c0'))
     for x in range(1, num_cmps):
-        command = command.when(f.col('_b'+ str(x)), f.col('_c' + str(x)))
-    command = command.otherwise(s1)
+        command = command.when(f.col('_b'+ str(x)), f.col('_c' + str(x)))\
+                         .when(f.lit(True), f.col('_c' + str(x)))
```
revans2, line 53: My concern with putting a lit true for here and below on line 54 is that it ensures that the evaluation essentially stops at that point, and it does not matter what we evaluate after this.  Internally we are walking from the bottom up, but the test should be agnostic of that.

I would rather we have them be below. as separate case/when statements

revans2 (commented):
Just a nit with the tests now

integration_tests/src/main/python/conditionals_test.py
firestarman: Done.

firestarman:
build

firestarman:
I am going to merge this, if any concern, will file following PRs.

integration_tests/src/main/python/conditionals_test.py
```diff
@@ -28,7 +28,9 @@ def test_if_else(data_gen):
     null_lit = get_null_lit_string(data_gen.data_type)
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : three_col_df(spark, boolean_gen, data_gen, data_gen).selectExpr(
-                # A literal predicate is not supported yet
+                'IF(TRUE, b, c)',
+                'IF(TRUE, {}, {})'.format(s1, null_lit),
+                'IF(FALSE, {}, {})'.format(s1, null_lit),
```
wbo4958, line 33: Does it support a literal predicate? like IF (3>2, b, c)

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -16,84 +16,42 @@
...
-import ai.rapids.cudf.Scalar
 import com.nvidia.spark.rapids.RapidsPluginImplicits._
 
 import org.apache.spark.sql.catalyst.analysis.{TypeCheckResult, TypeCoercion}
 import org.apache.spark.sql.catalyst.expressions.{ComplexTypeMergingExpression, Expression}
 import org.apache.spark.sql.types.{BooleanType, DataType}
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
-abstract class GpuConditionalExpression extends ComplexTypeMergingExpression with GpuExpression {
+trait GpuConditionalExpression extends ComplexTypeMergingExpression with GpuExpression {
```
wbo4958, line 26: any reason to change to trait?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -170,24 +128,12 @@ case class GpuCaseWhen(
...
-    } finally {
-      elseExpr match {
-        case a: AutoCloseable => a.close()
-        case _ =>
-      }
+    // `elseRet` will be closed in `computeIfElse`.
+    val elseRet = elseValue
+      .map(_.columnarEval(batch))
+      .getOrElse(GpuScalar(null, branches.last._2.dataType))
+    branches.foldRight[Any](elseRet) { case ((predicateExpr, trueExpr), falseRet) =>
```
wbo4958, line 135: If you get an AutoClose there, it should be closed there after using instead of closing it in another function.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: `trait ` is always preferred if no requirements for the members. And here is the case.

integration_tests/src/main/python/conditionals_test.py
firestarman: Yes

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: `trait` is always preferred if no requirements for members, and here is the case.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: I think it is ok since this `computeIfElse` is not public. We know all the places it is used. Besides closing it in `computeIfElse` can keep the code quite simple.

sperlingxx:
LGTM

sperlingxx (approved):
LGTM

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -16,84 +16,42 @@
...
-        }
-        GpuColumnVector.from(result, dataType)
-      } finally {
-        trueResult match {
-          case a: AutoCloseable => a.close()
-          case _ =>
+      falseValue: Any): GpuColumnVector = {
+    withResourceIfAllowed(falseValue) { falseRet =>
+      withResource(GpuExpressionsUtils.columnarEvalToColumn(predExpr, batch)) { pred =>
+        withResourceIfAllowed(trueExpr.columnarEval(batch)) { trueRet =>
```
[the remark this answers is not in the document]
revans2: The problem is that `columnarEval` returns an `Any`. So technically it is possible because of the signature. In reality it should never happen.  In the future we could change the signature, but it is not super simple to do right now because of how partitioning works. I am also not 100% sure what the return type should be.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -16,84 +16,42 @@
...
-          case t: GpuScalar => predicate.getBase.ifElse(t.getBase, falseValue)
-          case u =>
-            throw new IllegalStateException(s"Unexpected inputs $u")
-        }
-        GpuColumnVector.from(result, dataType)
-      } finally {
-        trueResult match {
-          case a: AutoCloseable => a.close()
-          case _ =>
+      falseValue: Any): GpuColumnVector = {
```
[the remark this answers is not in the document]
revans2: The problem is there is no common base between `GpuColumnVector` and `GpuScalar` except for `Any`. We could update to be an `Either`.  That might make it a little cleaner to deal with.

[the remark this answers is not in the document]
firestarman: I just want to avoid `match case` inisde this `computeIfElse` since it will be called in each loop of foldRight with a evaluated false value for `GpuCaseWhen`.

PR #2513, opened 2021-05-26
Report opTime not totalTime for expand, range, and generate execs

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/2512


sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
```diff
@@ -328,7 +328,7 @@ case class GpuRangeExec(range: org.apache.spark.sql.catalyst.plans.logical.Range
                 } else false
 
               override def next(): ColumnarBatch =
-                withResource(new NvtxWithMetrics("GpuRange", NvtxColor.DARK_GREEN, totalTime)) {
+                withResource(new NvtxWithMetrics("GpuRange", NvtxColor.DARK_GREEN, gpuOpTime)) {
```
jlowe, line 331: This will time the acquisition of the GPU semaphore which is _not_ the time the GPU was actually performing operations.  IMO gpuOpTime should not be measuring that time, and total time (or some other metric measuring the entire time spent in an operator, not just how long GPU kernels take) is more appropriate.

revans2: This gets especially hard for anything that spills because we might end up grabbing the semaphore at a really odd time when we try to get a batch. i.e. we processed something and wrote out a batch. The semaphore was released and then it called hasNext or next, which caused us to get a batch out of a SpillableBatch which will grab the semaphore. If we want to exclude the semaphore from timings we need to be very careful.

revans2: Just to be clear I am good with the goal to avoid measuring getting the semaphore. It just means we are going to have to be very careful about how we do it.

jlowe: Fine with this as-is, given we are tracking the overall issue in #2527

jlowe (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala
```diff
@@ -75,7 +75,7 @@ object GpuMetric extends Logging {
   val DESCRIPTION_PARTITION_SIZE = "partition data size"
   val DESCRIPTION_NUM_PARTITIONS = "partitions"
   val DESCRIPTION_TOTAL_TIME = "total time"
-  val DESCRIPTION_GPU_OP_TIME = "GPU time"
+  val DESCRIPTION_OP_TIME = "GPU time"
```
revans2, line 78: Are we still going to call it "GPU time"? This is what the user sees, "opTime" is just a key for the map.

andygrove: oops, missed that. updated.

revans2 (approved):

andygrove:
build

PR #2514, opened 2021-05-26
Add Struct support for ParquetWriter

razajafri:


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2815,7 +2815,8 @@ object GpuOverrides {
     exec[DataWritingCommandExec](
       "Writing data",
       ExecChecks((TypeSig.commonCudfTypes +
-        TypeSig.DECIMAL.withPsNote(TypeEnum.DECIMAL, "Only supported for Parquet")).nested(),
+        TypeSig.DECIMAL.withPsNote(TypeEnum.DECIMAL, "Only supported for Parquet") +
+        TypeSig.STRUCT.withPsNote(TypeEnum.STRUCT, "Only supported for Parquet")).nested(),
```
jlowe, line 2819: Are we missing a test?  When the struct support went in, the fact that we forgot to update this should have been caught by the test.

razajafri: Thanks for pointing this out @jlowe I remembered that we added struct support for `GpuInMemoryTableScanExec` and confused it with `DataWritingCommandExec`. We have to turn this on as part of this PR. I have updated the PR name to reflect it

razajafri:
build

integration_tests/src/main/python/parquet_write_test.py
```diff
@@ -42,11 +42,11 @@
...
-parquet_write_gens_list = [
-    parquet_basic_gen,
-    pytest.param(parquet_basic_gen + [decimal_gen_default,
-                  decimal_gen_scale_precision, decimal_gen_same_scale_precision, decimal_gen_64bit],
-                 marks=pytest.mark.allow_non_gpu("CoalesceExec"))]
+parquet_write_gens_list = [parquet_basic_gen,
+                           pytest.param([decimal_gen_default, parquet_basic_struct_gen,
+                           StructGen([['child0', StructGen([['child1', byte_gen]])]]),
+                           decimal_gen_scale_precision, decimal_gen_same_scale_precision, decimal_gen_64bit],
+                                        marks=pytest.mark.allow_non_gpu("CoalesceExec"))]
```
revans2, line 49: Why is `CoalesceExec` not on the GPU? and is there an open issue that will put it on the GPU? If not we should file one and reference it here. If so we should just reference it here.

jlowe: I assume this is fixed by #2530 which added struct support for coalesce.

razajafri: I have updated the PR to not allow Coalesce with Structs to run on CPU

razajafri: no, its not refflecting it here for some reason, unless you go to the last commit

jlowe: Are you referring to a commit that has yet to be pushed to his PR? Because I still see it in https://github.com/NVIDIA/spark-rapids/pull/2514/files and at https://github.com/razajafri/spark-rapids-1/blob/parquet-doc/integration_tests/src/main/python/parquet_write_test.py#L49

razajafri: OK you are seeing the latest, I wasn't seeing it until I looked at the commit specifically.

So that is needed because CoalesceExec doesn't support Decimals yet

jlowe: > So that is needed because CoalesceExec doesn't support Decimals yet

#2531 explicitly states it adds decimal support and appears to be testing it.  So if that is somehow not working in practice an issue needs to be filed and fixed.

razajafri: Sorry for wasting your time. I didn't see `_commonTypes` was added and thought it was `TypeSig.commonCudfTypes`

jlowe (approved):

jlowe:
build

PR #2515, opened 2021-05-26
Remove fetchTime and standardize on collectTime

andygrove:

We have two metrics `collectTime` and `fetchTime` that appear to have the same meaning. This PR removes `fetchTime` and standardizes on `collectTime`.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala
```diff
@@ -78,7 +77,6 @@ object GpuMetric extends Logging {
   val DESCRIPTION_GPU_OP_TIME = "GPU time"
   val DESCRIPTION_FETCH_TIME = "fetch time"
```
jlowe, line 78: Maybe we should take this opportunity to elaborate a bit on what 'fetch' means here?  Elsewhere in Spark, fetch implies shuffle, so it would be good to clarify.

jlowe (commented):
Note that Spark already has a `collectTime` metric for some of its operators, if that influences which way we want to go.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala
andygrove: I updated this PR to remove `fetchTime` and replace it with `collectTime` (the opposite of what I was originally proposing). This is consistent with how Spark is using their `collectTime` metric.

jlowe (approved):

jlowe:
build

PR #2520, opened 2021-05-27
Add event logs for integration tests

### LINKED ISSUE #2518: [FEA] Enable inspecting integration test jobs in Spark History Server
As a spark-rapids developer I wish I wish I could review integration tests in SHS. This includes the actual configuration and the timeline, and allows an easy answer to questions like in https://github.com/NVIDIA/spark-rapids/issues/2368#issuecomment-848673450
> The other important part is to be able to match the same number of tasks. How many tasks did this run with, typically the number of CPU threads that your processor has?

- A simple switch to enable even logging in integration tests
- Description allowing easy mapping from jobs to tests

Manual and tedious

**Additional context**
N/A

gerashegalov:
Closes #2518 

### DIFF
--- a/integration_tests/src/main/python/spark_init_internal.py
+++ b/integration_tests/src/main/python/spark_init_internal.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020-2021, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+import os
 
 try:
     from pyspark.sql import SparkSession
@@ -26,7 +28,6 @@ def _spark__init():
     # DO NOT SET ANY OTHER CONFIGS HERE!!!
     # due to bugs in pyspark/pytest it looks like any configs set here
     # can be reset in the middle of a test if specific operations are done (some types of cast etc)
-    import os
     _sb = SparkSession.builder
     _sb.config('spark.plugins', 'com.nvidia.spark.SQLPlugin') \
             .config("spark.sql.adaptive.enabled", "false") \
@@ -40,12 +41,11 @@ def _spark__init():
 
     if ('PYTEST_XDIST_WORKER' in os.environ):
         wid = os.environ['PYTEST_XDIST_WORKER']
-        d = "./derby_{}".format(wid)
-        if not os.path.exists(d):
-            os.makedirs(d)
-        _sb.config('spark.driver.extraJavaOptions', driver_opts + ' -Dderby.system.home={}'.format(d))
+        _create_derby_dir(_sb, driver_opts, wid)
+        _create_event_log_dir(_sb, wid)
     else:
         _sb.config('spark.driver.extraJavaOptions', driver_opts)
+        _create_event_log_dir(_sb, 'gw0')
  
     # enableHiveSupport() is needed for parquet bucket tests
     _s = _sb.enableHiveSupport() \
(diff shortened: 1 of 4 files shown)


gerashegalov:
Screenshot from 2021-05-26 18-28-00

gerashegalov:
build

integration_tests/src/main/python/spark_init_internal.py
```diff
@@ -55,6 +55,20 @@ def _spark__init():
     _s.sparkContext.setLogLevel("WARN")
     return _s
 
+
+def _create_derby_dir(sb, driver_opts, wid):
```
revans2, line 59: `_create_*_dir` feels like all it should do is to create that dir, not setup/change configuration.  We either need a better name or split this up.

integration_tests/src/main/python/spark_init_internal.py
```diff
@@ -55,6 +55,20 @@ def _spark__init():
...
+    d = "./derby_{}".format(wid)
+    if not os.path.exists(d):
+        os.makedirs(d)
+    sb.config('spark.driver.extraJavaOptions', driver_opts + ' -Dderby.system.home={}'.format(d))
+
+def _create_event_log_dir(sb, wid):
+    d = "./eventlog_{}".format(wid)
+    if not os.path.exists(d):
+        os.makedirs(d)
+    sb.config('spark.eventLog.dir', "file://{}".format(os.path.abspath(d)))
```
revans2, line 69: I don't think this is what we want for a distributed environment like EMR. We use find spark specifically so that we can set configs like this in python before the cluster comes up. In distributed stand alone mode I think this config will be ignored for YARN or Kubernetes I think it will make it so the history logs are written to a local file in the best case (and lost because it is a remote directory) or crash with not being able to create the directory. I personally would like to see us have an option to not try to configure this at all instead and ideally check to see if the user already configured it and if they did don't bother. I'm not sure how we can do that in all cases though, especially if it is configured cluster wide.

integration_tests/src/main/python/spark_session.py
```diff
@@ -64,11 +65,20 @@ def _check_for_proper_return_values(something):
...
     return ret
 
+
+def _add_job_description(conf):
+    is_gpu_job = conf.get('spark.rapids.sql.enabled', False)
+    job_type = 'GPU' if str(is_gpu_job).lower() == str(True).lower() else 'CPU'
+    job_desc = '{}[{}]'.format(os.environ.get('PYTEST_CURRENT_TEST'), job_type)
+    _spark.sparkContext.setJobDescription(job_desc)
+
+
```
abellina, line 81: nit, extra line.

integration_tests/src/main/python/spark_session.py
```diff
@@ -64,11 +65,20 @@ def _check_for_proper_return_values(something):
...
     """Run func that takes a spark session as input with the given configs set."""
     reset_spark_session_conf()
+    _add_job_description(conf)
     _set_all_confs(conf)
     ret = func(_spark)
     _check_for_proper_return_values(ret)
     return ret
 
+
+def _add_job_description(conf):
```
abellina, line 75: This is a nice change to get the name of the test as the job's name.

integration_tests/src/main/python/spark_init_internal.py
abellina: Agree, I wonder if this would be better as an option (for those local runs). It looks like right now it's set as default which would break current setups. I.e. as it stands if you point the integration tests to yarn or a standalone cluster that has the right configs, it does the right thing.

abellina (changes requested):

integration_tests/src/main/python/spark_init_internal.py
gerashegalov: thanks for the reminders @revans2 and @abellina  . Agree that we should not override if event log collection is already configured. Non-local execution was not my primary target and albeit important does not need to be addressed right away. 

BTW what worked  well for me in the past on YARN is to use  containers local `{{LOG_DIR}}` as spark event log destination because it's typically subject of log aggregation on prem or services like EMR. Then there is no need to overload a central SHS with jobs most of which will never be looked at.

integration_tests/src/main/python/spark_session.py
gerashegalov: It's my IDE reminding me of the need for 2 lines before the function definition:
```
PEP 8: E302 expected 2 blank lines, found

integration_tests/src/main/python/spark_init_internal.py
revans2: The point is that we run/test in multiple environments and if this cannot support all of them, then it is a problem that needs to be addressed right away. I am not sure if we can force the history directory to a `{{LOG_DIR}}` on all distributed setups, like EMR. This file being changed is common to all places that we run this code with no way to opt out.

gerashegalov: This point was not lost on me. I'll address it.

gerashegalov:
build

revans2 (approved):

abellina (approved):

abellina:
Thanks @gerashegalov

Issue #1712, opened 2021-02-11
[BUG] Scala UDF compiler can decompile UDFs with RAPIDS implementation


jlowe:
Now that Scala UDFs have support for a RAPIDS implementation since #1636, there's the possibility that the Scala UDF compiler could end up compiling a user's Scala UDF into Catalyst expressions despite the fact that it already provides a RAPIDS columnar implementation.  The UDF compiler should detect and avoid trying to translate UDFs that implement the `RapidsUDF` interface.

GaryShen2008:
Remove from v21.06. Let's target in 21.08.

jlowe:
Fixed by #1712

PR #2523, opened 2021-05-27
Not compile RapidsUDF when udf compiler is enabled

wjxiz1992:
To close: https://github.com/NVIDIA/spark-rapids/issues/1712.
Originally, when UDF Compiler is enabled, it will always try to compile the bytecodes for any instance of a ScalaUDF. 
This PR detects if a ScalaUDF implements the RapidsUDF interface. If yes, don't compile its bytecode because user has already provided a columnar evaluation method for the function.

Signed-off-by: Allen Xu <wjxiz1992@gmail.com>

### DIFF
--- a/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
+++ b/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
@@ -51,13 +51,8 @@ case class LogicalPlanRules() extends Rule[LogicalPlan] with Logging {
     val conf = new RapidsConf(plan.conf)
     // iterating over NamedExpression
     exp match {
-      case f: ScalaUDF => // found a ScalaUDF
-        // If it implements the RapidsUDF interface, no need to compile it.
-        if (!getRapidsUDFInstance(f.function).isEmpty) {
-          exp
-        } else {
-          GpuScalaUDFLogical(f).compile(conf.isTestEnabled)
-        }
+      case f: ScalaUDF if getRapidsUDFInstance(f.function).isEmpty =>
+        GpuScalaUDFLogical(f).compile(conf.isTestEnabled)
       case _ =>
         if (exp == null) {
           exp
--- a/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
+++ b/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
@@ -25,6 +25,7 @@ import org.apache.spark.sql.SparkSessionExtensions
 import org.apache.spark.sql.catalyst.expressions.{Alias, Expression, NamedExpression, ScalaUDF}
 import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, Project}
 import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.rapids.GpuScalaUDF.getRapidsUDFInstance
 
 class Plugin extends Function1[SparkSessionExtensions, Unit] with Logging {
   override def apply(extensions: SparkSessionExtensions): Unit = {
@@ -52,7 +53,7 @@ case class LogicalPlanRules() extends Rule[LogicalPlan] with Logging {
     exp match {
       case f: ScalaUDF => // found a ScalaUDF
         // If it implements the RapidsUDF interface, no need to compile it.
-        if (f.function.isInstanceOf[RapidsUDF]) {
+        if (!getRapidsUDFInstance(f.function).isEmpty) {
           exp
         } else {
           GpuScalaUDFLogical(f).compile(conf.isTestEnabled)
--- a/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
+++ b/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
@@ -19,6 +19,7 @@ package com.nvidia.spark.udf
 import ai.rapids.cudf.{NvtxColor, NvtxRange}
 import com.nvidia.spark.RapidsUDF
 import com.nvidia.spark.rapids.RapidsConf
+
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.SparkSessionExtensions
 import org.apache.spark.sql.catalyst.expressions.{Alias, Expression, NamedExpression, ScalaUDF}
--- a/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
+++ b/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
@@ -17,7 +17,6 @@
 package com.nvidia.spark.udf
 
 import ai.rapids.cudf.{NvtxColor, NvtxRange}
-import com.nvidia.spark.RapidsUDF
 import com.nvidia.spark.rapids.RapidsConf
 
 import org.apache.spark.internal.Logging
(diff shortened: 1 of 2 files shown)


udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
```diff
@@ -17,8 +17,8 @@
 package com.nvidia.spark.udf
 
 import ai.rapids.cudf.{NvtxColor, NvtxRange}
+import com.nvidia.spark.RapidsUDF
 import com.nvidia.spark.rapids.RapidsConf
-
```
firestarman, line 21: This empty line is required, otherwise `mvn verify` would fail.

wjxiz1992: Done.

revans2 (approved):

revans2:
build

udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
```diff
@@ -50,7 +51,12 @@ case class LogicalPlanRules() extends Rule[LogicalPlan] with Logging {
     // iterating over NamedExpression
     exp match {
       case f: ScalaUDF => // found a ScalaUDF
-        GpuScalaUDFLogical(f).compile(conf.isTestEnabled)
+        // If it implements the RapidsUDF interface, no need to compile it.
+        if (f.function.isInstanceOf[RapidsUDF]) {
```
jlowe, line 55: Was this tested?  I don't think this is going to work as it should never match.  `f.function` is always a lambda function that wraps the actual user's class, if any.  This needs to do something like what [`getRapidsUDFInstance` does in `GpuScalaUDF`](https://github.com/NVIDIA/spark-rapids/blob/branch-21.06/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalaUDF.scala#L70-L105) to examine the lambda and see if there's a `RapidsUDF` instance within.

jlowe (changes requested):

udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
wjxiz1992: I set breakpoint at this line with 2 existing test cases: 1. [a simple UDF compiler test](https://github.com/NVIDIA/spark-rapids/blob/branch-21.06/udf-compiler/src/test/scala/com/nvidia/spark/OpcodeSuite.scala#L73) 2.[URLDecode test](https://github.com/NVIDIA/spark-rapids/blob/branch-21.06/tests/src/test/scala/com/nvidia/spark/rapids/ScalaUDFSuite.scala#L24)

For 1. it returns false. 
For 2. it returns true.

During my try, I firstly used `f.function.getClass.getInterfaces` and found it contains the `RapidsUDF`, but then I found `isInstanceOf` also works for this case.

image

jlowe: I stand correct, there are some instances where `f.function` could be the user's class directly.  However there are also cases where it can be wrapped by Spark in a lambda function, see [the corresponding code in Spark's UDFRegistration](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/UDFRegistration.scala#L793-L795).

Therefore I still stand by my comment.  This should not rely solely on direct examination of the instance class.  Also from a maintenance perspective, we should not use different logic in the UDF compiler to detect RapidsUDF instances than the code supporting RapidsUDF.  `getRapidsUDFInstance` should be refactored into a method that can be reused by both the UDF compiler and the RAPIDS UDF support code.

wjxiz1992: Hi @jlowe, I have 2 questions for this:
1) How to compose a test case to cover the lambda function case? I tried to put a lambda Function[ , ](implements RapidsUDF) into the register method but it is still able to detect the RapidsUDF and not use serializedLambda.
2) For refactoring, I found it easy to just remove _private_ keyword for `getRapidsUDFInstance` to let it be used all over, but I don't know if it's appropriate.

udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
```diff
@@ -50,7 +52,12 @@ case class LogicalPlanRules() extends Rule[LogicalPlan] with Logging {
     // iterating over NamedExpression
     exp match {
       case f: ScalaUDF => // found a ScalaUDF
-        GpuScalaUDFLogical(f).compile(conf.isTestEnabled)
+        // If it implements the RapidsUDF interface, no need to compile it.
+        if (!getRapidsUDFInstance(f.function).isEmpty) {
```
abellina, line 56:
```suggestion
        if (getRapidsUDFInstance(f.function).isDefined) {
```

udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
```diff
@@ -50,7 +52,12 @@ case class LogicalPlanRules() extends Rule[LogicalPlan] with Logging {
...
+          exp
```
jlowe, line 57: If we're not going to compile the UDF then the expression should be treated like any other expression and recurse, in case one of its input expressions (which could be yet another UDF) could be compiled, e.g.:
```scala
  case f: ScalaUDF if getRapidsUDFInstance(f.function).isDefined {
    GpuScalaUDFLogical(f).compile(conf.isTestEnabled)
  case _ =>
....
```

wjxiz1992: sorry, missed this. Updated.

udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
```diff
@@ -49,7 +51,7 @@ case class LogicalPlanRules() extends Rule[LogicalPlan] with Logging {
     val conf = new RapidsConf(plan.conf)
     // iterating over NamedExpression
     exp match {
-      case f: ScalaUDF => // found a ScalaUDF
+      case f: ScalaUDF if getRapidsUDFInstance(f.function).isEmpty =>
```
abellina, line 54: Nit, could there be a comment above this stating why the check is required?

udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
```diff
@@ -17,13 +17,15 @@
 package com.nvidia.spark.udf
 
 import ai.rapids.cudf.{NvtxColor, NvtxRange}
+import com.nvidia.spark.RapidsUDF
```
jlowe, line 20:
```suggestion
```

udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
wjxiz1992: Added.

udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
wjxiz1992: Done.

jlowe (approved):

jlowe:
build

abellina (approved):

PR #2525, opened 2021-05-27
support interval.microseconds for range window TimeStampType

wbo4958:
1. This change supports interval.microseconds for TimeStampType.

2.fix a range window on DateType bug.

  When the order-by column type is DateType, the value bounded type
  should be IntegerType instead of CalendarIntervalType.
  In other words, Interval is not working when order-by column is DateType.


wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -600,13 +601,12 @@ class GpuSpecifiedWindowFrameMeta(
...
-            if (interval.microseconds != 0 || interval.months != 0) { // DAYS == 0 is permitted.
-              willNotWorkOnGpu(s"Bounds for Range-based window frames must be specified" +
-                s" only in DAYS. Found $interval")
+          case Literal(ci: CalendarInterval, CalendarIntervalType) =>
+            if (ci.months != 0) {
+              willNotWorkOnGpu("interval months isn't supported")
             }
-            interval.days.toLong
+            // return the total microseconds
+            ci.days * TimeUnit.DAYS.toMicros(1) + ci.microseconds
```
revans2, line 609: This calculation can overflow. In the normal SQL case it should be very unlikely to happen, but we should still detect it before it happens and fall back to the CPU.

wbo4958: Yeah, the overflow checking is quite necessary, Thx for the commenting. Done

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -600,16 +601,23 @@ class GpuSpecifiedWindowFrameMeta(
...
+              willNotWorkOnGpu("interval months isn't supported")
+            }
+            // return the total microseconds
+            try {
+              Math.addExact(
+                Math.multiplyExact(ci.days.toLong, TimeUnit.DAYS.toMicros(1)),
+                ci.microseconds)
+            } catch {
+              case e: ArithmeticException =>
+                willNotWorkOnGpu("interval total microseconds has overflow")
```
revans2, line 615: I would like to see a better message. It is not clear how the user can fix this or what exactly has happened.

Perhaps something more like

```
willNotWorkOnGpu(s"windows over timestamps are converted to microseconds and $ci is too large to fit")
```

revans2 (commented):
Just a nit

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
wbo4958: Thx @revans2. Done

wbo4958:
build

revans2 (approved):

Issue #2522, opened 2021-05-27
[BUG] Integration tests FAILED on Databricks 8.2 for it's Spark Runtime updated to 3.1.1


NvTimLiu:
**Describe the bug**

Databricks 8.2 update its spark runtime from 310 to 311, causing Databricks shims FAILURE when running IT against it.
We need to update our DB shims to 311 to meet with the update of DB8.2

ubuntu@dd5:~$ /databricks/spark/bin/spark-submit --version
Welcome to
      ____              __
     / __/__  ___ _____/ /__
    _\ \/ _ \/ _ `/ __/  '_/
   /___/ .__/\_,_/_/ /_/\_\   version 3.1.1
      /_/

Using Scala version 2.12.10, OpenJDK 64-Bit Server VM, 1.8.0_282

 

[2021-05-27T07:18:53.678Z] Setting default log level to "WARN".

[2021-05-27T07:18:53.679Z] To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).

[2021-05-27T07:18:53.679Z] ImportError while loading conftest '/home/ubuntu/spark-rapids/integration_tests/src/main/python/conftest.py'.

[2021-05-27T07:18:53.679Z] ../../src/main/python/conftest.py:17: in <module>

[2021-05-27T07:18:53.679Z]     from spark_init_internal import get_spark_i_know_what_i_am_doing

[2021-05-27T07:18:53.679Z] ../../src/main/python/spark_init_internal.py:58: in <module>

[2021-05-27T07:18:53.679Z]     _spark = _spark__init()

[2021-05-27T07:18:53.679Z] ../../src/main/python/spark_init_internal.py:51: in _spark__init

[2021-05-27T07:18:53.679Z]     _s = _sb.enableHiveSupport() \

[2021-05-27T07:18:53.679Z] /databricks/spark/python/pyspark/sql/session.py:229: in getOrCreate

[2021-05-27T07:18:53.679Z]     sc = SparkContext.getOrCreate(sparkConf)

[2021-05-27T07:18:53.679Z] /databricks/spark/python/pyspark/context.py:392: in getOrCreate

[2021-05-27T07:18:53.679Z]     SparkContext(conf=conf or SparkConf())

[2021-05-27T07:18:53.679Z] /databricks/spark/python/pyspark/context.py:147: in __init__

[...]

PR #2529, opened 2021-05-27
Change Databricks 310 shim to be 311 to match reported spark.version

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/2522

Databricks changed its version for runtime 8.2 to report the spark version as 3.1.1 which is what it should have been.  Update the shim layer to match.

All tests pass on databricks runtime 8.2


tgravescs:
build

jlowe (approved):

PR #2531, opened 2021-05-27
Support structs and decimals in coalesce

### LINKED ISSUE #2530: [FEA] Add support for Struct columns in CoalesceExec
I wish the RAPIDS Accelerator for Apache Spark would support structs when coalescing dataframes

df.coalesce() should be executed on GPU even in presence of struct columns

CPU fallback works

**Additional context**
N/A

gerashegalov:
Closes #2530

### DIFF
--- a/docs/supported_ops.md
+++ b/docs/supported_ops.md
@@ -140,13 +140,13 @@ Accelerator supports are described below.
 <td>S</td>
 <td>S*</td>
 <td>S</td>
-<td><b>NS</b></td>
+<td>S*</td>
 <td>S</td>
 <td><b>NS</b></td>
 <td><b>NS</b></td>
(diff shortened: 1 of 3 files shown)


gerashegalov:
build

jlowe (commented):
This looks good to me, but shouldn't we be able to support arrays, maps and other types with it as well?  I don't think in practice `GpuCoalesceExec` actually cares about the underlying types involved.

jlowe (approved):
Approving in case we want to handle the other types that should "just work" in a followup issue.  If we don't handle those other types here it would be good to file a tracking issue.

gerashegalov:
build

revans2 (approved):

PR #2532, opened 2021-05-27
date_format should not suggest enabling incompatibleDateFormats for formats we cannot support

### LINKED ISSUE #2517: [FEA] date_format should not suggest enabling incompatibleDateFormats for formats we cannot support
When using a date_format expression with a format string that is not compatible with GPU, such as 'F' we see this EXPLAIN output:

```
cannot run on GPU because incompatible format 'F'. Set spark.rapids.sql.incompatibleDateFormats.enabled=true to force onto GPU
```

However, setting this flag will not make this run on GPU because it is unsupported.

If we know we can't support it on GPU, we should not suggest trying to make it run on GPU.

None

**Additional context**
None

andygrove:

Closes #2517.

### DIFF
--- a/integration_tests/src/main/python/date_time_test.py
+++ b/integration_tests/src/main/python/date_time_test.py
@@ -218,14 +218,14 @@ def test_date_format(data_gen, date_format):
 unsupported_date_formats = ['F']
 @pytest.mark.parametrize('date_format', unsupported_date_formats, ids=idfn)
 @pytest.mark.parametrize('data_gen', date_n_time_gens, ids=idfn)
-@allow_non_gpu('ProjectExec,date_format,Alias,DateFormatClass,Literal,Cast')
+@allow_non_gpu('ProjectExec,Alias,DateFormatClass,Literal,Cast')
 def test_date_format_f(data_gen, date_format):
     assert_gpu_fallback_collect(
         lambda spark : unary_op_df(spark, data_gen).selectExpr("date_format(a, '{}')".format(date_format)), 'ProjectExec')
 
 @pytest.mark.parametrize('date_format', unsupported_date_formats, ids=idfn)
 @pytest.mark.parametrize('data_gen', date_n_time_gens, ids=idfn)
-@allow_non_gpu('ProjectExec,date_format,Alias,DateFormatClass,Literal,Cast')
+@allow_non_gpu('ProjectExec,Alias,DateFormatClass,Literal,Cast')
 def test_date_format_f_incompat(data_gen, date_format):
     # note that we can't support it even with incompatibleDateFormats enabled
     conf = {"spark.rapids.sql.incompatibleDateFormats.enabled": "true"}
@@ -235,7 +235,7 @@ def test_date_format_f_incompat(data_gen, date_format):
 maybe_supported_date_formats = ['dd-MM-yyyy']
 @pytest.mark.parametrize('date_format', maybe_supported_date_formats, ids=idfn)
 @pytest.mark.parametrize('data_gen', date_n_time_gens, ids=idfn)
-@allow_non_gpu('ProjectExec,date_format,Alias,DateFormatClass,Literal,Cast')
+@allow_non_gpu('ProjectExec,Alias,DateFormatClass,Literal,Cast')
 def test_date_format_maybe(data_gen, date_format):
     assert_gpu_fallback_collect(
         lambda spark : unary_op_df(spark, data_gen).selectExpr("date_format(a, '{}')".format(date_format)),
--- a/integration_tests/src/main/python/date_time_test.py
+++ b/integration_tests/src/main/python/date_time_test.py
@@ -13,10 +13,10 @@
 # limitations under the License.
 
 import pytest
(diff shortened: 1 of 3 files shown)


integration_tests/src/main/python/date_time_test.py
```diff
@@ -214,3 +214,36 @@ def test_string_unix_timestamp(data_gen, date_form):
...
+
+unsupported_date_formats = ['F']
+@pytest.mark.parametrize('date_format', unsupported_date_formats, ids=idfn)
+@pytest.mark.parametrize('data_gen', date_n_time_gens, ids=idfn)
+@pytest.mark.xfail(reason='cuDF does not support these formats')
+def test_date_format(data_gen, date_format):
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark : unary_op_df(spark, data_gen).selectExpr("date_format(a, '{}')".format(date_format)))
+
+unsupported_date_formats_force = ['F']
```
wjxiz1992, line 226: This seems to be duplicated?

integration_tests/src/main/python/date_time_test.py
```diff
@@ -214,3 +214,36 @@ def test_string_unix_timestamp(data_gen, date_form):
...
+
+maybe_supported_date_formats = ['dd-MM-yyyy']
+@pytest.mark.parametrize('date_format', maybe_supported_date_formats, ids=idfn)
+@pytest.mark.parametrize('data_gen', date_n_time_gens, ids=idfn)
+@pytest.mark.xfail(reason='These formats can be parsed but are not certified as supported')
+def test_date_format(data_gen, date_format):
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark : unary_op_df(spark, data_gen).selectExpr("date_format(a, '{}')".format(date_format)))
+
+maybe_supported_date_formats_force = ['dd-MM-yyyy']
```
wjxiz1992, line 243: Same question here

integration_tests/src/main/python/date_time_test.py
andygrove: The second version of the test specifies `conf = {"spark.rapids.sql.incompatibleDateFormats.enabled": "true"}`

wjxiz1992: Oh, sorry, I mean the variable `unsupported_date_formats_force` is assigned but not used.

integration_tests/src/main/python/date_time_test.py
```diff
@@ -214,3 +214,36 @@ def test_string_unix_timestamp(data_gen, date_form):
 def test_date_format(data_gen, date_format):
     assert_gpu_and_cpu_are_equal_collect(
         lambda spark : unary_op_df(spark, data_gen).selectExpr("date_format(a, '{}')".format(date_format)))
+
+unsupported_date_formats = ['F']
+@pytest.mark.parametrize('date_format', unsupported_date_formats, ids=idfn)
+@pytest.mark.parametrize('data_gen', date_n_time_gens, ids=idfn)
+@pytest.mark.xfail(reason='cuDF does not support these formats')
```
revans2, line 221: Why xfail and not test that we fall back to the CPU?

integration_tests/src/main/python/date_time_test.py
andygrove: Thanks, this is fixed now. There were numerous other issues with the tests that are also resolved now.

integration_tests/src/main/python/date_time_test.py
andygrove: Thanks, this is fixed now.

andygrove:
build

sameerz:
@andygrove can we retarget this to 21.08?

revans2 (approved):

sameerz:
Moving to branch-21.08 after @Salonijain27 confirmed with the PR author.

integration_tests/src/main/python/date_time_test.py
```diff
@@ -214,3 +214,36 @@ def test_string_unix_timestamp(data_gen, date_form):
 def test_date_format(data_gen, date_format):
     assert_gpu_and_cpu_are_equal_collect(
         lambda spark : unary_op_df(spark, data_gen).selectExpr("date_format(a, '{}')".format(date_format)))
+
+unsupported_date_formats = ['F']
+@pytest.mark.parametrize('date_format', unsupported_date_formats, ids=idfn)
+@pytest.mark.parametrize('data_gen', date_n_time_gens, ids=idfn)
+@allow_non_gpu('ProjectExec,date_format,Alias,DateFormatClass,Literal,Cast')
```
revans2, line 221: nit: I don't think that `date_format` needs to be in the `allow_non_gpu` tag.

revans2 (approved):

integration_tests/src/main/python/date_time_test.py
andygrove: This has been updated now

revans2 (approved):

andygrove:
build

PR #2536, opened 2021-05-28
Don't do an extra shuffle in some TopN cases

revans2:
This is a very small performance improvement for TakeOrderedAndProject. It just skips doing an extra shuffle if the input partitioning is already a single partition.  This can happen when doing a window operation with no partitioning.  Not common, but it can help.


revans2:
build

PR #2548, opened 2021-06-01
support collect_set in rolling window

sperlingxx:
This PR is to support `GpuCollectSet` as a rolling window aggregation function. It is marked as WIP because it relied on a [cuDF fix](https://github.com/rapidsai/cudf/pull/8406).

### DIFF
--- a/integration_tests/src/main/python/window_function_test.py
+++ b/integration_tests/src/main/python/window_function_test.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import math
 import pytest
 
 from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_and_cpu_are_equal_sql
@@ -415,34 +415,20 @@ def test_window_aggs_for_ranges_timestamps(data_gen):
         'from window_agg_table',
         conf = {'spark.rapids.sql.castFloatToDecimal.enabled': True})
 
-<<<<<<< HEAD
-_gen_data_for_collect = [
-=======
-@pytest.mark.xfail(reason="[UNSUPPORTED] Ranges over non-timestamp columns "
-                          "(https://github.com/NVIDIA/spark-rapids/issues/216)")
-@ignore_order
-@pytest.mark.parametrize('data_gen', [_grpkey_longs_with_timestamps], ids=idfn)
-def test_window_aggs_for_ranges_of_dates(data_gen):
-    assert_gpu_and_cpu_are_equal_sql(
-        lambda spark: gen_df(spark, data_gen, length=2048),
-        "window_agg_table",
-        'select '
-        ' sum(c) over '
-        '   (partition by a order by b asc  '
-        '       range between 1 preceding and 1 following) as sum_c_asc '
-        'from window_agg_table'
-    )
-
 
 _gen_data_for_collect_list = [
->>>>>>> 8e9a7431... support collect_set in rolling window
     ('a', RepeatSeqGen(LongGen(), length=20)),
     ('b', IntegerGen()),
+    ('c_bool', BooleanGen()),
+    ('c_short', ShortGen()),
     ('c_int', IntegerGen()),
     ('c_long', LongGen()),
-    ('c_time', DateGen()),
+    ('c_date', DateGen()),
+    ('c_ts', TimestampGen()),
+    ('c_byte', ByteGen()),
     ('c_string', StringGen()),
     ('c_float', FloatGen()),
+    ('c_double', DoubleGen()),
     ('c_decimal', DecimalGen(precision=8, scale=3)),
     ('c_struct', StructGen(children=[
         ['child_int', IntegerGen()],
@@ -459,16 +445,26 @@ def test_window_aggs_for_rows_collect_list():
         "window_collect_table",
         '''
         select
+          collect_list(c_bool) over
+            (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_bool,
+          collect_list(c_short) over
+            (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_short,
           collect_list(c_int) over
             (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_int,
           collect_list(c_long) over
             (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_long,
-          collect_list(c_time) over
-            (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_time,
+          collect_list(c_date) over
+            (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_date,
+          collect_list(c_ts) over
+            (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_ts,
+          collect_list(c_byte) over
+            (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_byte,
           collect_list(c_string) over
             (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_string,
           collect_list(c_float) over
             (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_float,
+          collect_list(c_double) over
+            (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_double,
           collect_list(c_decimal) over
             (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_decimal,
           collect_list(c_struct) over
@@ -481,7 +477,7 @@ def test_window_aggs_for_rows_collect_list():
 @ignore_order(local=True)
 def test_running_window_function_exec_for_all_aggs():
     assert_gpu_and_cpu_are_equal_sql(
-        lambda spark : gen_df(spark, _gen_data_for_collect),
+        lambda spark : gen_df(spark, _gen_data_for_collect_list),
         "window_collect_table",
         '''
         select
@@ -489,8 +485,8 @@ def test_running_window_function_exec_for_all_aggs():
             (partition by a order by b,c_int rows between UNBOUNDED PRECEDING AND CURRENT ROW) as sum_int,
           min(c_long) over
             (partition by a order by b,c_int rows between UNBOUNDED PRECEDING AND CURRENT ROW) as min_long,
-          max(c_time) over
-            (partition by a order by b,c_int rows between UNBOUNDED PRECEDING AND CURRENT ROW) as max_time,
+          max(c_date) over
+            (partition by a order by b,c_int rows between UNBOUNDED PRECEDING AND CURRENT ROW) as max_date,
           count(1) over
             (partition by a order by b,c_int rows between UNBOUNDED PRECEDING AND CURRENT ROW) as count_1,
           count(*) over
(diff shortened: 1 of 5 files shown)


integration_tests/src/main/python/window_function_test.py
```diff
@@ -444,3 +445,53 @@ def test_window_aggs_for_rows_collect_list():
...
+    ('c_long', RepeatSeqGen(LongGen(), length=15)),
+    ('c_time', RepeatSeqGen(DateGen(), length=15)),
+    ('c_string', RepeatSeqGen(StringGen(), length=15)),
+    ('c_float', RepeatSeqGen(FloatGen(), length=15)),
+    ('c_decimal', RepeatSeqGen(DecimalGen(precision=8, scale=3), length=15)),
+]
+
+
+# Uses sort_array function to reorder the outputs of collect_set, to make the result deterministic.
+@allow_non_gpu('ProjectExec', 'Alias', 'Literal', 'SortArray')
```
revans2, line 465: Can you file a follow on issue to support `sort_array`? In it's most basic form we should be able to support it.  For more complicated ones with actual higher order functions it is going to be a bit more difficult.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2625,9 +2625,8 @@ object GpuOverrides {
         override def convertToGpu(): GpuExpression = GpuPosExplode(childExprs.head.convertToGpu())
       }),
     expr[CollectList](
-      "Collect a list of elements, now only supported by windowing.",
-      // It should be 'fullAgg' eventually but now only support windowing,
-      // so 'aggNotGroupByOrReduction'
+      "Collect a list of non-unique elements, only supported in rolling window in current.",
```
revans2, line 2628: I think the original explanation was clearer. If we want to follow what Spark wrote it would be "Collect the values into a list"

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2639,6 +2638,20 @@ object GpuOverrides {
         override def convertToGpu(): GpuExpression = GpuCollectList(
           childExprs.head.convertToGpu(), c.mutableAggBufferOffset, c.inputAggBufferOffset)
       }),
+    expr[CollectSet](
+      "Collect a set of unique elements, only supported in rolling window in current.",
```
revans2, line 2642: Could we make this.
```
"Collect a set of elements, only supported by windowing.",
```

revans2 (commented):
Just a few nits, but I don't think this is going to get in to 21.06 because it depends on things in cudf that are still pending, and cudf 21.06 is frozen.  If we are going to keep this targeted here we are going to need @sameerz to work with the cudf team to get the remaining dependencies in.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -658,47 +658,72 @@ case class GpuLast(child: Expression, ignoreNulls: Boolean)
...
+  override def prettyName: String = "collect_list"
+
+  override def windowAggregation(
+      inputs: Seq[(ColumnVector, Int)]): AggregationOnColumn[CollectListAggregation] = {
+    Aggregation.collectList().onColumn(inputs.head._2)
+  }
+}
+
+/**
+ * Collects and returns a set of unique elements.
```
firestarman, line 712: nit: `set` already means the elements in it are unique.
```suggestion
 * Collects and returns a set of elements.
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -658,47 +658,72 @@ case class GpuLast(child: Expression, ignoreNulls: Boolean)
...
-  override def aggBufferAttributes: Seq[AttributeReference] = cudfList :: Nil
+  override val inputProjection: Seq[Expression] = Seq(childExpression)
+
+  override def aggBufferAttributes: Seq[AttributeReference] = {
+    throw new UnsupportedOperationException
+  }
+}
+
+/**
+ * Collects and returns a list of non-unique elements.
```
firestarman, line 692: `list` does not care whether the elements are unique.
```suggestion
 * Collects and returns a list of elements.
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -658,47 +658,72 @@ case class GpuLast(child: Expression, ignoreNulls: Boolean)
...
+    childExpression: Expression,
+    mutableAggBufferOffset: Int = 0,
+    inputAggBufferOffset: Int = 0)
+    extends GpuCollectBase[CollectSetAggregation] {
+
+  override def prettyName: String = "collect_set"
+
+  override def windowAggregation(
+      inputs: Seq[(ColumnVector, Int)]): AggregationOnColumn[CollectSetAggregation] = {
+    Aggregation.collectSet().onColumn(inputs.head._2)
```
firestarman, line 727: Setting the parameters explicitly would be better than relying on the default behavior of `collectSet()`. Since it may be broken if any change in `collectSet()` unexpectedly.
```suggestion
    Aggregation.collectSet(NullPolicy.EXCLUDE, NullEquality.UNEQUAL, NaNEquality.UNEQUAL).onColumn(inputs.head._2)
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -658,47 +658,72 @@ case class GpuLast(child: Expression, ignoreNulls: Boolean)
...
+    childExpression: Expression,
+    mutableAggBufferOffset: Int = 0,
+    inputAggBufferOffset: Int = 0)
+    extends GpuCollectBase[CollectListAggregation] {
+
+  override def prettyName: String = "collect_list"
+
+  override def windowAggregation(
+      inputs: Seq[(ColumnVector, Int)]): AggregationOnColumn[CollectListAggregation] = {
+    Aggregation.collectList().onColumn(inputs.head._2)
```
firestarman, line 707: Setting the parameters explicitly would be better than relying on the default behavior of `collectList()`. Since it may be broken if any change in `collectList()` unexpectedly.
```suggestion
    Aggregation.collectList(NullPolicy.EXCLUDE).onColumn(inputs.head._2)
```
Yes I used the default one before but now I think it is not a good idea.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2639,6 +2638,20 @@ object GpuOverrides {
...
+      // GpuCollectSet is not yet supported under GroupBy and Reduction context.
+      // Compared to CollectList, StructType is NOT in GpuCollectSet because underlying
```
firestarman, line 2644: Missing `supported ` ?
```suggestion
      // Compared to CollectList, StructType is NOT supported in GpuCollectSet because underlying
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -658,47 +658,72 @@ case class GpuLast(child: Expression, ignoreNulls: Boolean)
...
- */
-case class GpuCollectList(child: Expression,
-                          mutableAggBufferOffset: Int = 0,
-                          inputAggBufferOffset: Int = 0)
-  extends GpuAggregateFunction with GpuAggregateWindowFunction[CollectListAggregation] {
+trait GpuCollectBase[T <: Aggregation with RollingAggregation[T]] extends GpuAggregateFunction
+    with GpuAggregateWindowFunction[T] {
 
-  def this(child: Expression) = this(child, 0, 0)
+  def childExpression: Expression
```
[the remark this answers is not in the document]
sperlingxx: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: Fixed.

integration_tests/src/main/python/window_function_test.py
sperlingxx: Done.

revans2: https://github.com/rapidsai/cudf/blob/48a2663f11b3c36ebed1b4300307cbbd6375c6ca/cpp/include/cudf/lists/sorting.hpp#L51

It is not perfect because Spark uses a higher order function for this, but we can support the default ordering at least, and possibly use pattern matching to figure out other pieces.

integration_tests/src/main/python/window_function_test.py
```diff
@@ -415,7 +415,23 @@ def test_window_aggs_for_ranges_timestamps(data_gen):
         'from window_agg_table',
         conf = {'spark.rapids.sql.castFloatToDecimal.enabled': True})
 
-_gen_data_for_collect = [
+@pytest.mark.xfail(reason="[UNSUPPORTED] Ranges over non-timestamp columns "
+                          "(https://github.com/NVIDIA/spark-rapids/issues/216)")
```
revans2, line 419: This was closed. Is it still an issue?

firestarman:
LGTM

integration_tests/src/main/python/window_function_test.py
sperlingxx: Sorry, I mistakenly kept this test by when resolving merge conflicts.

sperlingxx:
build

sperlingxx:
build

sperlingxx:
build

sperlingxx:
build

integration_tests/src/main/python/window_function_test.py
```diff
@@ -455,3 +456,55 @@ def test_window_aggs_for_rows_collect_list():
             (partition by a order by b,c_int rows between CURRENT ROW and UNBOUNDED FOLLOWING) as collect_struct
         from window_collect_table
         ''')
+
+
+# Generates some repeated values to test the deduplication of GpuCollectSet.
+# And GpuCollectSet does not yet support struct type.
+_gen_data_for_collect_set = [
+    ('a', RepeatSeqGen(LongGen(), length=20)),
+    ('b', IntegerGen()),
```
revans2, line 465: Could you make this a `LongRangeGen` instead, just so we know that there is no ambiguity to the sort order?

integration_tests/src/main/python/window_function_test.py
```diff
@@ -455,3 +456,55 @@ def test_window_aggs_for_rows_collect_list():
...
+    ('c_int', RepeatSeqGen(IntegerGen(), length=15)),
+    ('c_long', RepeatSeqGen(LongGen(), length=15)),
+    ('c_time', RepeatSeqGen(DateGen(), length=15)),
+    ('c_string', RepeatSeqGen(StringGen(), length=15)),
+    ('c_float', RepeatSeqGen(FloatGen(), length=15)),
+    ('c_decimal', RepeatSeqGen(DecimalGen(precision=8, scale=3), length=15)),
```
revans2, line 471: Could you add in a few more types here?  We say we support bytes, shorts, doubles, and dates. It would be really nice to add them to the tests too.

integration_tests/src/main/python/window_function_test.py
```diff
@@ -455,3 +456,55 @@ def test_window_aggs_for_rows_collect_list():
...
+# Generates some repeated values to test the deduplication of GpuCollectSet.
+# And GpuCollectSet does not yet support struct type.
+_gen_data_for_collect_set = [
+    ('a', RepeatSeqGen(LongGen(), length=20)),
+    ('b', IntegerGen()),
+    ('c_int', RepeatSeqGen(IntegerGen(), length=15)),
+    ('c_long', RepeatSeqGen(LongGen(), length=15)),
+    ('c_time', RepeatSeqGen(DateGen(), length=15)),
+    ('c_string', RepeatSeqGen(StringGen(), length=15)),
+    ('c_float', RepeatSeqGen(FloatGen(), length=15)),
```
revans2, line 470: Have you verified that there are NaNs in this?  We are doing data gen and the default weight for NaNs is only going to insert them about 5% of the time.  We configure the operation as ` NaNEquality.UNEQUAL` I really would like to know for sure that this is being tested and 15 items makes it feel a bit less likely that they will show up.

sperlingxx:
build

integration_tests/src/main/python/window_function_test.py
sperlingxx: Done.

integration_tests/src/main/python/window_function_test.py
sperlingxx: Done.

integration_tests/src/main/python/window_function_test.py
sperlingxx: I added extra data column (`FloatGen().with_special_case(math.nan, 200.0)`) into test, which produces a lot of NaN values.

revans2 (dismissed):

sperlingxx:
build

revans2 (approved):

revans2:
build

PR #2566, opened 2021-06-02
expose unspill config option

rongou:
Unspill turns out to be useful when we try to align GDS IO, since whole buffers can be read back aligned to 4 KiB, bypassing GDS's internal bounce buffer.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -357,7 +357,6 @@ object RapidsConf {
         "back into GPU memory temporarily. Unspilling may be useful for GPU buffers that are " +
         "needed frequently, for example, broadcast variables; however, it may also increase GPU " +
         "memory usage")
-      .internal()
       .booleanConf
       .createWithDefault(false)
```
jlowe, line 361: Do we still want to default this to `false`?  Do we have any cases where it is known to hurt performance?

rongou: With GDS spilling it seems like a net win, but with host memory spilling it's less clear, maybe slightly slower.

jlowe (approved):

jlowe:
build

abellina (approved):

PR #2570, opened 2021-06-03
Added nested structs to UnionExec

razajafri:
This PR adds nested structs to UnionExec

#### This was tested by
Adding a 4-level nested StructGen to existing tests

### DIFF
--- a/integration_tests/src/main/python/repart_test.py
+++ b/integration_tests/src/main/python/repart_test.py
@@ -23,6 +23,17 @@ import pyspark.sql.functions as f
 nested_scalar_mark=pytest.mark.xfail(reason="https://github.com/NVIDIA/spark-rapids/issues/1459")
 @pytest.mark.parametrize('data_gen', [pytest.param((StructGen([['child0', DecimalGen(7, 2)]]),
                                                     StructGen([['child1', IntegerGen()]])), marks=nested_scalar_mark),
+                                      # left_struct(child0 = 4 level nested, child1 = Int)
+                                      # right_struct(child0 = 4 level nested, child1 = missing)
+                                      (StructGen([['child0', StructGen([['child0', StructGen([['child0', StructGen([['child0',
+                                                             StructGen([['child0', DecimalGen(7, 2)]])]])]])]])], ['child1', IntegerGen()]], nullable=False),
+                                       StructGen([['child0', StructGen([['child0', StructGen([['child0', StructGen([['child0',
+                                                            StructGen([['child0', DecimalGen(7, 2)]])]])]])]])]], nullable=False)),
+                                      # left_struct(child0 = 4 level nested, child1=missing)
+                                      # right_struct(child0 = missing, child1 = Int)
+                                      (StructGen([['child0', StructGen([['child0', StructGen([['child0', StructGen([['child0',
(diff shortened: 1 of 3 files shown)


razajafri:
build

integration_tests/src/main/python/repart_test.py
```diff
@@ -23,6 +23,17 @@
 nested_scalar_mark=pytest.mark.xfail(reason="https://github.com/NVIDIA/spark-rapids/issues/1459")
 @pytest.mark.parametrize('data_gen', [pytest.param((StructGen([['child0', DecimalGen(7, 2)]]),
                                                     StructGen([['child1', IntegerGen()]])), marks=nested_scalar_mark),
+                                      # left_struct(child0 = 4 level nested, child1 = Int)
+                                      # right_struct(child0 = 4 level nested, child1 = missing)
+                                      (StructGen([['child0', StructGen([['child0', StructGen([['child0', StructGen([['child0',
```
jlowe, line 28: Do we have to worry about [SPARK-35108](https://issues.apache.org/jira/browse/SPARK-35108)?  We may need to vary the number of fields at each level so there's no accidental aliasing to trigger that Spark bug.

sameerz:
@razajafri please target new features to branch-21.08.

integration_tests/src/main/python/repart_test.py
razajafri: Thanks for this. I have updated the PR to have a different number of children at each level for tests testing against spark < 3.1. The fix for the bug went in 3.0.3 I believe

integration_tests/src/main/python/repart_test.py
```diff
@@ -21,8 +21,38 @@
...
+                    ['child2', BooleanGen()],
+                    ['child3', BooleanGen()]
+                ])],
+                ['child1', BooleanGen()],
+                ['child2', BooleanGen()]
+            ])],
+            ['child1', BooleanGen()]
+        ])],
+        ['child1', BooleanGen()]
+    ])]])
```
jlowe, line 42: It looks like there's still the same number of children at two of the levels?  Maybe I'm misreading the indentation, but it looks like this to me:
```
child0
|-child0
| |-child0
| | |-child0
| | | |-child0
| | | |-child1
| | | |-child2
| | | |-child3
| | |-child1
| | |-child2
| |-child1
|-child1
```
Note how the first and second levels both have 2 children.  Did you mean to drop the last child so the first level only had one child instead of two?

integration_tests/src/main/python/repart_test.py
jlowe: I see the new `nested_struct` was added, but it's not used here?  Seems like most structs generated for this test only have one child.

razajafri: so this test will skip for versions < 311. The bug SPARK-35108 was fixed for 3.0.3+. I could've added it to this test like I did for the others too but this one needs a tuple of structs for left and right and I didn't want to compromise the readability unless it is needed. 

I could still do it if you want me to

integration_tests/src/main/python/repart_test.py
razajafri: Good catch. I have removed one level

jlowe (approved):

jlowe:
build

Issue #2477, opened 2021-05-21
[BUG] test_single_sort_in_part is failed in nightly UCX and AQE (no UCX) integration 


jlowe:
The test_single_sort_in_part test failed with a comparison failure in a recent UCX EGX standalone integration test:
```
11:08:43  =================================== FAILURES ===================================
11:08:43  [31m[1m___ test_single_sort_in_part[Column<b'a DESC NULLS FIRST'>-Float(not_null)] ____[0m
11:08:43  
11:08:43  data_gen = Float(not_null), order = Column<b'a DESC NULLS FIRST'>
11:08:43  
11:08:43      @pytest.mark.parametrize('data_gen', orderable_gens + orderable_not_null_gen, ids=idfn)
11:08:43      @pytest.mark.parametrize('order', [f.col('a').asc(), f.col('a').asc_nulls_last(), f.col('a').desc(), f.col('a').desc_nulls_first()], ids=idfn)
11:08:43      def test_single_sort_in_part(data_gen, order):
11:08:43  >       assert_gpu_and_cpu_are_equal_collect(
11:08:43                  lambda spark : unary_op_df(spark, data_gen).sortWithinPartitions(order),
11:08:43                  conf = allow_negative_scale_of_decimal_conf)
11:08:43  
11:08:43  [1m[31m../../src/main/python/sort_test.py[0m:112: 
11:08:43  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
11:08:43  [1m[31m../../src/main/python/asserts.py[0m:364: in assert_gpu_and_cpu_are_equal_collect
11:08:43      _assert_gpu_and_cpu_are_equal(func, 'COLLECT', conf=conf, is_cpu_first=is_cpu_first)
11:08:43  [1m[31m../../src/main/python/asserts.py[0m:356: in _assert_gpu_and_cpu_are_equal
11:08:43      assert_equal(from_cpu, from_gpu)
11:08:43  [1m[31m../../src/main/python/asserts.py[0m:93: in assert_equal
11:08:43      _assert_equal(cpu, gpu, float_check=get_float_check(), path=[])
11:08:43  [1m[31m../../src/main/python/asserts.py[0m:41: in _assert_equal
11:08:43      _assert_equal(cpu[index], gpu[index], float_check, path + [index])
[...]

PR #2571, opened 2021-06-03
Change test_single_sort_in_part to print source data frame on failure

abellina:

Relates #2477. This is an attempt to capture the source data frame when `test_single_sort_in_part` fails. The output prints to stdout, so the pytest framework outputs it after the fact, like so:

```
----------------------------- Captured stdout call -----------------------------
### CPU RUN ###
COLLECTED
[Row(a=-341142443), Row(a=48424546), Row(a=-100515324), Row(a=1752831507), Row(a=-936999309), Row(a=1727289611), Row(a=987119867), Row(a=790204970),...
```

The output happens for both CPU and GPU sessions which is overkill and I can work to make that a single output if we are inclined.

I also made this an explicit single call, but was going down the route of making it for every test function without api changes. I figure for now it's probably best just to capture the output of this flaky test, we can discuss when/how to capture and output the dataframe for every test here or in a follow on.

### DIFF
--- a/integration_tests/src/main/python/asserts.py
+++ b/integration_tests/src/main/python/asserts.py
@@ -380,6 +380,15 @@ def assert_gpu_and_cpu_are_equal_collect(func, conf={}, is_cpu_first=True):
     """
     _assert_gpu_and_cpu_are_equal(func, 'COLLECT', conf=conf, is_cpu_first=is_cpu_first)
 
+def assert_gpu_and_cpu_are_equal_collect_debug(df_fun, result_fun, conf={}, debug=False, is_cpu_first=True):
+    def do_it_all(spark):
+        df = df_fun(spark)
+        if debug:
+            data_gen.debug_df(df)
+        return result_fun(df)
+
+    assert_gpu_and_cpu_are_equal_collect(do_it_all, conf, is_cpu_first=is_cpu_first)
+
 def assert_gpu_and_cpu_are_equal_iterator(func, conf={}, is_cpu_first=True):
     """
     Assert when running func on both the CPU and the GPU that the results are equal.
--- a/integration_tests/src/main/python/asserts.py
+++ b/integration_tests/src/main/python/asserts.py
@@ -380,15 +380,6 @@ def assert_gpu_and_cpu_are_equal_collect(func, conf={}, is_cpu_first=True):
     """
     _assert_gpu_and_cpu_are_equal(func, 'COLLECT', conf=conf, is_cpu_first=is_cpu_first)
 
-def assert_gpu_and_cpu_are_equal_collect_debug(df_fun, result_fun, conf={}, debug=False, is_cpu_first=True):
-    def do_it_all(spark):
-        df = df_fun(spark)
-        if debug:
-            data_gen.debug_df(df)
-        return result_fun(df)
-
-    assert_gpu_and_cpu_are_equal_collect(do_it_all, conf, is_cpu_first=is_cpu_first)
-
 def assert_gpu_and_cpu_are_equal_iterator(func, conf={}, is_cpu_first=True):
     """
     Assert when running func on both the CPU and the GPU that the results are equal.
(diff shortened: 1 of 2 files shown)


integration_tests/src/main/python/asserts.py
```diff
@@ -380,6 +380,15 @@ def assert_gpu_and_cpu_are_equal_collect(func, conf={}, is_cpu_first=True):
     """
     _assert_gpu_and_cpu_are_equal(func, 'COLLECT', conf=conf, is_cpu_first=is_cpu_first)
 
+def assert_gpu_and_cpu_are_equal_collect_debug(df_fun, result_fun, conf={}, debug=False, is_cpu_first=True):
```
jlowe, line 383: The name of this function is `_debug`, so why would I ever want to pass `debug=False`, and why would that be a default?

integration_tests/src/main/python/sort_test.py
```diff
@@ -105,9 +105,11 @@ def test_single_nested_orderby_with_limit(data_gen, order):
 @pytest.mark.parametrize('data_gen', orderable_gens + orderable_not_null_gen, ids=idfn)
 @pytest.mark.parametrize('order', [f.col('a').asc(), f.col('a').asc_nulls_last(), f.col('a').desc(), f.col('a').desc_nulls_first()], ids=idfn)
 def test_single_sort_in_part(data_gen, order):
-    assert_gpu_and_cpu_are_equal_collect(
-            lambda spark : unary_op_df(spark, data_gen).sortWithinPartitions(order),
-            conf = allow_negative_scale_of_decimal_conf)
+    assert_gpu_and_cpu_are_equal_collect_debug(
```
jlowe, line 108: It would be nice to make this easier to use, as I suspect we'll have other instances in the future where we'll want to debug-dump a dataframe like this.  The issue with the way it's implemented now is that I can't just change `assert_gpu_and_cpu_are_equal_collect` to `assert_gpu_and_cpu_are_equal_collect_debug` and walk away. I need to also provide a result function and hope I coded it the same way it used to work with `assert_gpu_and_cpu_are_equal_collect` did.

If we added a new `COLLECT_WITH_DEBUG` collect type and corresponding code to dump after the test's dataframe call is made then I think we could make it so tacking on `_debug` to the assert call is all that's required to debug it.

For this specific case where you're wanting to dump the dataframe after the unary_op but before the sort, we actually don't need a new assert method at all.  We could write it as something like this:
```python
def test_single_sort_in_part(data_gen, order):
    def doit(spark):
        df = unary_op_df(spark, data_gen)
        data_gen.debug_df(df)
        return df.sortWithinPartitions(order)
    assert_gpu_and_cpu_are_equal_collect(doit, conf = allow_negative_scale_of_decimal_conf)
```

abellina: > If we added a new COLLECT_WITH_DEBUG collect type and corresponding code to dump after the test's dataframe call is made then I think we could make it so tacking on _debug to the assert call is all that's required to debug it.

The issue is that what we want to capture is the source dataframe, not the result dataframe. When assert* is called, we already created the result dataframe. There already exists a `COLLECT_WITH_DATAFRAME`, but if I understand correctly, the dataframe that is captured is the one after `sortWithinPartitions` in our case. I agree it can be made easier to use, thought of some options:

1. Just have the `data_gen` instance regenerate. It should create a new Random variable from the same seed, the result should be exactly the same. Need a way to capture which `data_gen` (which is not consistent, i.e. there's a `three_col_df` for example, that can take different `data_gen` intances)

2. Could record generated dataframes in temporary views or some other variable (thought of within `data_gen`). This seems like a bad idea creating side effects.

3. It does seem that a general recipe is: we make a dataframe, then we perform an action on it. The API for this _debug assert was somewhat used in the _sql asserts, where we create a temp view and give it a name, so it can be acted on. I wonder if we really ought to write test in this way?

jlowe (approved):

jlowe:
build

revans2 (approved):
FYI: debug_df returns the original df, so you could have just written it as.

```
                lambda spark : debug_df(unary_op_df(spark, data_gen)).sortWithinPartitions(order),
```

abellina:
let me change that @revans2

abellina:
@revans2 changed with: https://github.com/NVIDIA/spark-rapids/pull/2571/commits/c17fc7ff1bfc8c15e2a15455bf7f4af3543bf257. That is better. Thanks!

revans2 (approved):

abellina:
build

abellina:
ok looks like I hit some CI timeout here. I am going to wait a bit and retrigger later.

sameerz:
build

Issue #2307, opened 2021-04-29
[FEA] Enable Parquet writing for arrays


sameerz:
**Is your feature request related to a problem? Please describe.**
The cuIO Parquet writer supports writing arrays.  We should enable that support in the plugin.  

**Describe the solution you'd like**
Support for writing list types in the plugin for Parquet.  

**Describe alternatives you've considered**
N/A

**Additional context**
Support for writing structs is is being addressed in PR #2271

PR #2582, opened 2021-06-04
support parquet write arrays

wbo4958:
This PR enables parquet write for arrays. Please refer to https://github.com/NVIDIA/spark-rapids/issues/2307

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -761,7 +761,8 @@ object GpuOverrides {
     (ParquetFormatType, FileFormatChecks(
       cudfRead = (TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.STRUCT + TypeSig.ARRAY +
           TypeSig.MAP).nested(),
-      cudfWrite = (TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.STRUCT).nested(),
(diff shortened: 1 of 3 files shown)


wbo4958:
build

wbo4958:
build

wbo4958:
@razajafri could you help to review this PR?

revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2828,7 +2829,8 @@ object GpuOverrides {
       "Writing data",
       ExecChecks((TypeSig.commonCudfTypes +
         TypeSig.DECIMAL.withPsNote(TypeEnum.DECIMAL, "Only supported for Parquet") +
-        TypeSig.STRUCT.withPsNote(TypeEnum.STRUCT, "Only supported for Parquet")).nested(),
+        TypeSig.STRUCT.withPsNote(TypeEnum.STRUCT, "Only supported for Parquet") +
```
razajafri, line 2832: nit: would be good to have an API that takes an array of `TypeEnum` so we have to duplicate the message `Only supported for Parquet`

razajafri: We can create a s follow-on issue for this as an improvement if you want

razajafri (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
wbo4958: ok, Thx @razajafri, Will create a follow-on issue for this

PR #2587, opened 2021-06-04
Handle UCX connection timeouts from heartbeats more gracefully

abellina:
This patch adds an internal config for the UCX management port connection from the client side. The config defaults to `0`, which was what we were doing before.

If there is a timeout for a connection initiated due to a heartbeat (not due to the act of needing a client to fetch blocks at shuffle time), we log and ignore the exception. There will be a more invasive change in 21.08 to remove such stale executors, but I thought not to add in 21.06 to not add too much risk.

The change also improves the error message to include the `BlockManagerId`. Instead of just:
```
 21/05/27 23:07:11 ERROR RapidsShuffleHeartbeatEndpoint: Error initializing shuffle
 java.net.ConnectException: Connection timed out (Connection timed out)
```

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
@@ -999,6 +999,13 @@ object RapidsConf {
     .stringConf
     .createWithDefault(null)
 
+  val SHUFFLE_UCX_MGMT_CONNECTION_TIMEOUT =
+    conf("spark.rapids.shuffle.ucx.managementConnectionTimeout")
+    .doc("The timeout for client connections to a remote peer")
+    .internal()
+    .integerConf
+    .createWithDefault(0)
+
   val SHUFFLE_UCX_BOUNCE_BUFFERS_SIZE = conf("spark.rapids.shuffle.ucx.bounceBuffers.size")
     .doc("The size of bounce buffer to use in bytes. Note that this size will be the same " +
       "for device and host memory")
@@ -1535,6 +1542,8 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
   lazy val shuffleUcxListenerStartPort: Int = get(SHUFFLE_UCX_LISTENER_START_PORT)
 
   lazy val shuffleUcxMgmtHost: String = get(SHUFFLE_UCX_MGMT_SERVER_HOST)
+
+  lazy val shuffleUcxMgmtConnTimeout: Int = get(SHUFFLE_UCX_MGMT_CONNECTION_TIMEOUT)
 
   lazy val shuffleUcxBounceBuffersSize: Long = get(SHUFFLE_UCX_BOUNCE_BUFFERS_SIZE)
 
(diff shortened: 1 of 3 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -999,6 +999,13 @@ object RapidsConf {
     .stringConf
     .createWithDefault(null)
 
+  val SHUFFLE_UCX_MGMT_CONNECTION_TIMEOUT =
+    conf("spark.rapids.shuffle.ucx.managementConnectionTimeout")
+    .doc("The timeout for client connections to a remote peer")
+    .internal()
+    .integerConf
+    .createWithDefault(0)
```
jlowe, line 1007: Why are we disabling a timeout by default?  Given this is an internal config, curious what the use-case is for this code.  Are we going to suggest users set this undocumented config in certain situations, use it in a test, or ...?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/RapidsShuffleInternalManagerBase.scala
```diff
@@ -203,7 +203,23 @@ abstract class RapidsShuffleInternalManagerBase(conf: SparkConf, isDriver: Boole
...
+          // We ignore the exception after logging in this instance because
+          // we may have a peer that doesn't exist anymore by the time `addPeer` is invoked
+          // due to a heartbeat response from the driver, or the peer may have a temporary network
+          // issue.
+          //
+          // This is safe because `addPeer` is only invoked due to a heartbeat that is used to
+          // opportunistically hide cost of initializing transport connections. The transport
+          // will re-try if it must fetch from this executor at a later time, in that case
+          // a connection failure causes the tasks to fail.
+          logError(s"Unable to connect to peer $peer, ignoring!", ex)
```
jlowe, line 220: Should this be a warning rather than an error?  Error implies something is not going to work whereas a warning implies something odd happened but the situation can be recovered.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
abellina: It's just for our internal testing, and I expect it to be removed in 21.08. I didn't change it from what it was default (`new Socket(host, port)` does the same). But I can choose a more sensible value.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/RapidsShuffleInternalManagerBase.scala
abellina: Fixed in: https://github.com/NVIDIA/spark-rapids/pull/2587/commits/87f14f63bf9e4c24729e7ffb6e2b1545d475382a

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
abellina: I'd like to be able to control this while we get the UcpListener to be in a state where it can be enabled by default. Then the UCX timeouts will be the ones we use, so this setting won't apply (i.e. it would be via `UCX_` env variables instead).

jlowe: If we're going to rip this out shortly in 21.08 I'm fine with leaving it at what it effectively was before.

jlowe (approved):

jlowe:
build

Issue #1791, opened 2021-02-23
[FEA] GPU support for the plan RunningWindowFunctionExec on db


firestarman:
**Is your feature request related to a problem? Please describe.**
On databricks, the plan `RunningWindowFunctionExec `, which has not been supported by GPU, will be applied instead of the `WindowExec` when matching some requirements. Then some window related IT tests will fail. 
Some related ssues: https://github.com/NVIDIA/spark-rapids/issues/1680 https://github.com/NVIDIA/spark-rapids/issues/203

**Describe the solution you'd like**
Replace the plan `RunningWindowFunctionExec` with a new or existing GPU plan correctly.

**Additional context**
`RunningWindowRunctionExec` does not exist outside of Databricks, and will NOT be used when

- window function is a SizeBasedWindowFunction, or
- window function is a OffsetWindowFunction, or
- window function is a PythonUDF, or
- window frame specification is not SpecifiedWindowFrame(RowFrame, UnboundedPreceding, CurrentRow)

Since databricks is not open source, so it may not be easy to understand what `RunningWindowRunctionExec` does and how to support it on GPU perfectly.

mythrocks:
For reference, we need the following window spec to reproduce the problem:
```sql
... OVER (PARTITION BY id ORDER BY dt ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)
```
Hence, the term `RunningWindowFunctionExec`. Reversing the boundaries (i.e. `CURRENT ROW AND UNBOUNDED FOLLOWING`) does not suffice, for some reason.

PR #2596, opened 2021-06-04
Initial running window memory optimization

revans2:
This adds in an initial optimization for some running window commands.  Specifically it allows min, max, row_number, and count to not need any re-grouping of the batches after the initial sort for running window operations.  Performance testing showed that it was faster than the version that splits batches by key. (13 seconds doing a row number for 2 billion rows with 10,000 rows per batch vs 15 seconds for the existing version)  This also has the added advantage that it allows us to do the processing for very large windows without needing to keep all of the data in GPU memory.

Right now if all of the operations in a WindowExec can be done as a rolling window this will do them that way. If even one of them cannot we fall back to doing it the original way. When we support other optimizations like fixed row window sizes we should look at splitting a single window operation up into multiple execs assuming that all of them can support working without needing a full window of the data.

This is draft right now because it falls back to the old way on databricks.  Once https://github.com/NVIDIA/spark-rapids/issues/1791 is in I will update this to deal with it and test on databricks. For now any reviews are welcome.


revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -856,7 +857,90 @@ trait GpuAggregateWindowFunction[T <: Aggregation with RollingAggregation[T]]
...
+   * Save any state needed from the previous output that will be needed to fix up the next batch.
+   * You don't need to worry about the partition columns. Those will be saved separately.
+   * @param finalOutputColumn the output of fixup.
+   */
+  def updateState(finalOutputColumn: GpuColumnVector): Unit
+
+  /**
+   * Fix up `windowedColumnOutput` with any stored state from previous batches.
+   * @param samePartitionMask a boolean mask with no nulls in it. `true` indicates the row
+   *                          is for the same partition by keys that was the last row in the
```
abellina, line 871: ok so technically you can an inbound `windowedColumnOutput` with keys `[key1, key1, key2, key2]`, where `samePartitionMask` would be `[false, true, false, true]`. From the header, I am not entirely sure if `previous batches` means previous calls to `fixUp`.

abellina: would be nice to note when `sameParttionMask` can be a vector, and when it is a scalar.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -856,7 +857,90 @@ trait GpuAggregateWindowFunction[T <: Aggregation with RollingAggregation[T]]
...
+   */
+  def fixUp(
+      samePartitionMask: Either[GpuColumnVector, Boolean],
+      windowedColumnOutput: GpuColumnVector): GpuColumnVector
+}
+
+/**
+ * For many operations a running window (unbounded preceding to current row) can
+ * process the data without dividing the data up into batches that contain all of the data
+ * for a given group by key set. Instead we store a small amount of state from a previous result
```
abellina, line 887: would be nice to expand on "we store a small amount of state". What state are we keeping?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -856,7 +857,90 @@ trait GpuAggregateWindowFunction[T <: Aggregation with RollingAggregation[T]]
...
+ * GpuAggregateWindowFunction can still work for running window queries. This just reduces the
+ * maximum amount of memory needed to process them.
+ */
+trait GpuBatchedRunningWindowFunction[T <: Aggregation with RollingAggregation[T]]
+    extends GpuAggregateWindowFunction[T] {
+
+  /**
+   * Get a new class that can be used to fix up batched RunningWindowOperations.
+   */
+  def newFixer(): BatchedRunningWindowFixer
```
abellina, line 898: Seems like a "fixer" is an aggregate function used to combine two partial values from the same key. Could there be a more descriptive name than "fixer"? Wondering if something like "combiner" captures its function, but I am not entirely sure yet.

abellina (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
revans2: Aggregate function to me means it is doing an aggregation. This is not doing an aggregation. It is fixing up the result of an already done aggregation based off of data that aggregation was missing. Spark already has an `AggregateFunction` so I really want to avoid that term or anything like it. `Combiner` shows up in Spark as a part of shuffle (Which is the same as in Hadoop it lets you do an aggregation as you shuffle the data), and as a more generic aggregation API for RDDs.  I guess this could be called a `Combiner`, but it does not really fit. A combiner typically is a function that takes two of the same type of value and outputs a new one. It is intended to be used in an iterative reduction/aggregation.  This is not that.  Which is why I called it a "Fixer". I appreciate naming help and I agree "Fixer" is not a good name, but that is why it is a `BatchedRunningWindowFixer` to try and make it less ambiguous.

abellina: Agree combine is not a good a choice either. Regardless of name, it doesn't seem entirely clear why we need to fix batches up. Could a short example be included in a comment on how the fixing process works? This is the reason why I was trying to suggest a different name, I am not entirely sure why we are fixing and what the fix is.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -78,24 +86,200 @@ class GpuWindowExecMeta(windowExec: WindowExec,
...
+  def isRunningWindow(spec: GpuWindowSpecDefinition): Boolean =
+    spec.frameSpecification match {
+      case sf: GpuSpecifiedWindowFrame =>
+        (sf.lower, sf.upper) match {
+          case (GpuSpecialFrameBoundary(UnboundedPreceding), GpuSpecialFrameBoundary(CurrentRow)) =>
+            true
+          case _ => false
+        }
+      case _ => false
+    }
```
gerashegalov, line 98: nit:

```suggestion
  def isRunningWindow(spec: GpuWindowSpecDefinition): Boolean = spec match {
    case GpuWindowSpecDefinition(_, _, GpuSpecifiedWindowFrame(_,
      GpuSpecialFrameBoundary(UnboundedPreceding), GpuSpecialFrameBoundary(CurrentRow))) => true
    case _ => false
  }
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
revans2: That is up to the operation.

revans2:
build

revans2:
I have addressed the review comments and tested this on databricks.  We should be good to go with it now.

gerashegalov:
please resolve conflicts

revans2:
build

gerashegalov (approved):
LGTM

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -856,7 +857,137 @@ trait GpuAggregateWindowFunction[T <: Aggregation with RollingAggregation[T]]
   def windowAggregation(inputs: Seq[(ColumnVector, Int)]): AggregationOnColumn[T]
 }
 
-case class GpuRowNumber() extends GpuAggregateWindowFunction[RowNumberAggregation] {
+/**
+ * Provides a way to process running window operations without needing to buffer and split the
+ * batches on partition by boundaries. When this happens part of a partition by key set may
+ * have been processed in the last batch, and the rest of it will need to be updated. For example
+ * if we are doing a running max operation. We may first get in something like
```
abellina, line 864:
```suggestion
 * if we are doing a running min operation. We may first get in something like
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -856,7 +857,137 @@ trait GpuAggregateWindowFunction[T <: Aggregation with RollingAggregation[T]]
...
+ *
+ * We generate the window result again and get
+ *
+ * <code>
+ *    MINS: 11, 5, 13, 13
+ * </code>
+ * But we cannot output this yet because there may have been overlap with the previous batch.
+ * The framework will figure that out and pass data into `fixUp` to do the fixing. It will
+ * pass in MINS, and also a column of boolean values true, true, false, false to indicate
+ * whcih rows overlapped with the previous batch.  In our min example fixUp will do a min
```
abellina, line 892:
```suggestion
 * which rows overlapped with the previous batch.  In our min example fixUp will do a min
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -869,6 +1000,9 @@ case class GpuRowNumber() extends GpuAggregateWindowFunction[RowNumberAggregatio
     assert(inputs.isEmpty, inputs)
     Aggregation.rowNumber().onColumn(0)
   }
+
+  override def newFixer(): BatchedRunningWindowFixer =
+    new BatchedRunningWindowBinaryFixer(BinaryOp.ADD, "row_number")
```
abellina, line 1005: I assume this is a kind of no-op for row_number, and that the partitioner is looking at all columns.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -856,7 +857,137 @@ trait GpuAggregateWindowFunction[T <: Aggregation with RollingAggregation[T]]
...
+ */
+class BatchedRunningWindowBinaryFixer(val binOp: BinaryOp, val name: String)
+    extends BatchedRunningWindowFixer with Arm with Logging {
+  private var previousResult: Option[Scalar] = None
+
+  override def updateState(finalOutputColumn: GpuColumnVector): Unit = {
+    logDebug(s"$name: updateState from $previousResult to...")
+    previousResult.foreach(_.close)
+    previousResult =
+      Some(finalOutputColumn.getBase.getScalarElement(finalOutputColumn.getRowCount.toInt - 1))
```
abellina, line 962: could you get a case where `getRowCount==0`?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -856,7 +857,137 @@ trait GpuAggregateWindowFunction[T <: Aggregation with RollingAggregation[T]]
...
+      case (Some(prev), scala.util.Right(mask)) =>
+        if (mask) {
+          GpuColumnVector.from(windowedColumnOutput.getBase.binaryOp(binOp, prev, prev.getType),
+            windowedColumnOutput.dataType())
+        } else {
+          // The mask is all false so do nothing
+          windowedColumnOutput.incRefCount()
+        }
+      case (Some(prev), scala.util.Left(mask)) =>
+        val base = windowedColumnOutput.getBase
```
abellina, line 980: does this need to check `if(mask)`?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -137,13 +180,160 @@ class GpuWindowExecMeta(windowExec: WindowExec,
...
+      val numRows = cb.numRows
+      numOutputBatches += 1
+      numOutputRows += numRows
+      withResource(new MetricRange(opTime)) { _ =>
+        withResource(GpuProjectExec.projectAndClose(cb, boundProjectList, NoopMetric)) { full =>
+          closeOnExcept(ArrayBuffer[ColumnVector]()) { newColumns =>
+            boundProjectList.indices.foreach { idx =>
+              val column = full.column(idx).asInstanceOf[GpuColumnVector]
+              val fixer = fixers.get(idx)
+              if (fixer.isDefined) {
```
abellina, line 222: nit, would be nice to keep `fixer.get` in a val to reuse in the couple of places utilized here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -856,7 +857,137 @@ trait GpuAggregateWindowFunction[T <: Aggregation with RollingAggregation[T]]
...
+  }
+
+  override def fixUp(samePartitionMask: Either[GpuColumnVector, Boolean],
+      windowedColumnOutput: GpuColumnVector): GpuColumnVector = {
+    logDebug(s"$name: fix up $previousResult $samePartitionMask")
+    (previousResult, samePartitionMask) match {
+      case (None, _) => windowedColumnOutput.incRefCount()
+      case (Some(prev), scala.util.Right(mask)) =>
+        if (mask) {
+          GpuColumnVector.from(windowedColumnOutput.getBase.binaryOp(binOp, prev, prev.getType),
```
abellina, line 973: so if mask == true, this is creating a new column vector from the result of `binaryOp`, should it consume `windowedColumnOutput`?

abellina (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
revans2: `row_number` only applies within a given partition.  The output is identical to doing a `count(1)`. The main difference is that `row_number` only supports a running window so there are no odd corner cases when you ask for a row number.

```
scala> val df = spark.range((20).toLong).selectExpr("id DIV 10 as a", "id as b")
df: org.apache.spark.sql.DataFrame = [a: bigint, b: bigint]
 
scala> spark.time(df.coalesce(1).selectExpr("a", "b", "row_number() over (partition by a order by b) as row_num", "count(1) over (partition by a order by b rows between UNBOUNDED PRECEDING and CURRENT ROW) as count_row_num" ).show(100)) 
+---+---+-------+-------------+
|  a|  b|row_num|count_row_num|
+---+---+-------+-------------+
|  0|  0|      1|            1|
|  0|  1|      2|            2|
|  0|  2|      3|            3|
|  0|  3|      4|            4|
|  0|  4|      5|            5|
|  0|  5|      6|            6|
|  0|  6|      7|            7|
|  0|  7|      8|            8|
|  0|  8|      9|            9|
|  0|  9|     10|           10|
|  1| 10|      1|            1|
|  1| 11|      2|            2|
|  1| 12|      3|            3|
|  1| 13|      4|            4|
|  1| 14|      5|            5|
|  1| 15|      6|            6|
|  1| 16|      7|            7|
|  1| 17|      8|            8|
|  1| 18|      9|            9|
|  1| 19|     10|           10|
+---+---+-------+-------------+

Time taken: 293 ms


```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
revans2: Left indicates that it is a `GpuColumnVector`, not a boolean so the `ifElse` call just blow this takes over for `if(mask)` in the other cases.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
revans2: No I will update the docs.

revans2:
@abellina I think I have addressed all of your comments. Please take another look

revans2:
build

abellina (approved):

PR #2600, opened 2021-06-05
Accelerate `RunningWindow` queries on GPU

### LINKED ISSUE #1791: [FEA] GPU support for the plan RunningWindowFunctionExec on db

mythrocks:
Fixes #1791.

A window-spec defined as `ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW` is termed a "running" window, because for each successive row in a row-group, the window increases in size by `1`.

"Running" windows have special handling on Databricks, via a `RunningWindowFunctionExec`. `spark-rapids` does not currently recognize or replace this exec with a GPU equivalent, thereby causing "running" window queries to fail thus:
```
java.lang.IllegalStateException: Found an aggregation function in an unexpected context Some(!NOT_FOUND <RunningWindowFunctionExec> NOT EVALUATED FOR GPU YET
...
  at com.nvidia.spark.rapids.ExpressionContext$.getAggregateFunctionContext(RapidsMeta.scala:723)
  at com.nvidia.spark.rapids.BaseExprMeta.context$lzycompute(RapidsMeta.scala:764)
  at com.nvidia.spark.rapids.BaseExprMeta.context(RapidsMeta.scala:760)
  at com.nvidia.spark.rapids.ExprChecksImpl.tag(TypeChecks.scala:732)
  at com.nvidia.spark.rapids.BaseExprMeta.$anonfun$tagSelfForGpu$3(RapidsMeta.scala:774)
  at com.nvidia.spark.rapids.BaseExprMeta.$anonfun$tagSelfForGpu$3$adapted(RapidsMeta.scala:774)
  at scala.Option.foreach(Option.scala:407)
  at com.nvidia.spark.rapids.BaseExprMeta.tagSelfForGpu(RapidsMeta.scala:774)
  at com.nvidia.spark.rapids.RapidsMeta.tagForGpu(RapidsMeta.scala:279)
...
```
This commit aims to recognize and replace `RunningWindowFunctionExec` with a GPU implementation. In this first pass, it will delegate to `GpuWindowExec`. This has the potential to be optimized further, (i.e. evaluated in a single pass), once `rapids/cudf` has the support for it.


shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
```diff
@@ -23,8 +23,8 @@ import com.nvidia.spark.rapids._
 import org.apache.arrow.memory.ReferenceManager
 import org.apache.arrow.vector.ValueVector
 import org.apache.hadoop.fs.Path
-
 import org.apache.spark.SparkEnv
```
abellina, line 26: import groups look a bit off here, did you forget run `mvn verify`?

shims/spark301db/src/main/scala/com/nvidia/spark/rapids/shims/spark301db/GpuRunningWindowExec.scala
```diff
@@ -0,0 +1,58 @@
...
+    runningWindowFunctionExec.partitionSpec.map(GpuOverrides.wrapExpr(_, conf, Some(this)))
+  val orderSpec: Seq[BaseExprMeta[SortOrder]] =
+    runningWindowFunctionExec.orderSpec.map(GpuOverrides.wrapExpr(_, conf, Some(this)))
+
+  override def tagPlanForGpu(): Unit = {
+    windowExpressions.map(meta => meta.wrapped)
+        .filter(expr => !expr.isInstanceOf[NamedExpression])
+        .foreach(_ => willNotWorkOnGpu(because = "Unexpected query plan with Windowing functions; " +
+            "cannot convert for GPU execution. " +
+            "(Detail: WindowExpression not wrapped in `NamedExpression`.)"))
```
abellina, line 45: why does the `WindowExpression` need to be wrapped in `NamedExpression`? A comment may help.

shims/spark311db/src/main/scala/com/nvidia/spark/rapids/shims/spark311db/Spark311dbShims.scala
```diff
@@ -99,6 +104,17 @@ class Spark311dbShims extends Spark311Shims {
             )
           }
         }).disabledByDefault("it only supports row based frame for now"),
+      GpuOverrides.exec[RunningWindowFunctionExec](
```
abellina, line 107: does it make sense to find a way to combine both shim impls? If not appropriate for this PR, perhaps a follow up issue.

shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
tgravescs: yeah, put the newline back above to separate spark from others and newline below should be removed

shims/spark311db/src/main/scala/com/nvidia/spark/rapids/shims/spark311db/Spark311dbShims.scala
tgravescs: much of this I explicitly left as copied when I created it.
1) its safer to build against the specific Databricks version
2) spark 3.2.0 changes will probably end up doing this anyway
3) hopefully the spark301dbshim will go away soon

it is a bit of maintenance overhead but doesn't change much and for this it seems safer just to have 2 copies for a little while until we get to the changes needed to support spark 3.2.0

shims/spark301db/src/main/scala/com/nvidia/spark/rapids/shims/spark301db/GpuRunningWindowExec.scala
```diff
@@ -0,0 +1,58 @@
...
+
+package com.nvidia.spark.rapids.shims.spark301db
+
+import com.databricks.sql.execution.window.RunningWindowFunctionExec
+import com.nvidia.spark.rapids.{BaseExprMeta, DataFromReplacementRule, GpuExec, GpuOverrides, GpuWindowExec, RapidsConf, RapidsMeta, SparkPlanMeta}
+
+import org.apache.spark.sql.catalyst.expressions.{Expression, NamedExpression, SortOrder}
+
+/**
+ * GPU-based window-exec implementation, analogous to RunningWindowFunctionExec.
```
tgravescs, line 25: so is the only difference using the RunningWindowFunctionExec to get access tot he variables?

tgravescs:
build

shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
mythrocks: @abellina is right about my not running `mvn verify`. I'll restore the newline and check with `mvn verify`.

shims/spark301db/src/main/scala/com/nvidia/spark/rapids/shims/spark301db/GpuRunningWindowExec.scala
mythrocks: Yes, sir. This class follows Databricks's conventions in `WindowExec`:
1. Only the aggregation result column is returned.
2. The aggregation inputs can be accessed via `windowExpressionList`.

shims/spark301db/src/main/scala/com/nvidia/spark/rapids/shims/spark301db/GpuRunningWindowExec.scala
mythrocks: The `WindowExpression` is unwrapped from the `NamedExpression` downstream in `GpuWindowExec`, to pull out the column/attribute.
The input is always a `NamedExpression`, as far as I've seen. The check is defensive, on the off chance that this changes.

shims/spark311db/src/main/scala/com/nvidia/spark/rapids/shims/spark311db/GpuRunningWindowExec.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import com.databricks.sql.execution.window.RunningWindowFunctionExec
+import com.nvidia.spark.rapids.{BaseExprMeta, DataFromReplacementRule, GpuExec, GpuOverrides, GpuWindowExec, RapidsConf, RapidsMeta, SparkPlanMeta}
+
+import org.apache.spark.sql.catalyst.expressions.{Expression, NamedExpression, SortOrder}
+
+/**
+ * GPU-based window-exec implementation, analogous to RunningWindowFunctionExec.
+ */
+class GpuRunningWindowExecMeta(runningWindowFunctionExec: RunningWindowFunctionExec,
```
revans2, line 27: nit: Would it be possible to add in an abstract class that is the base for both `GpuRunningWindowExecMeta` and `GpuWindowExecMeta`? The main reason I ask is because there is a bit of code that is copied and pasted here, not all of it, but for my PR there is a lot more of it, and I would like to avoid duplicating code as much as possible.

revans2 (commented):
Just a nit. It looks good.

shims/spark301db/src/main/scala/com/nvidia/spark/rapids/shims/spark301db/GpuRunningWindowExec.scala
mythrocks: P.S. Will document.

shims/spark301db/src/main/scala/com/nvidia/spark/rapids/shims/spark301db/GpuRunningWindowExec.scala
revans2: All output columns from a `SparkPlan` stage has to have a name.  `output` is a `Seq` of `Attribute`s and generally a `NamedExpression` is used to help convey that information to the `SparkPlan` like `resultExpressions` in `HashAggregate`

shims/spark311db/src/main/scala/com/nvidia/spark/rapids/shims/spark311db/GpuRunningWindowExec.scala
mythrocks: Ok, this is proving non-trivial right now. `RunningWindowFunctionExec` and `WindowExec` aren't related classes. `GpuBaseWindowExecMeta` will need to be genericized. And Scala's generics don't seem to work similarly to C++'s. I'm trying to sort this out now.

shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
mythrocks: This should be fixed now.

tgravescs (approved):

tgravescs:
build

shims/spark311db/src/main/scala/com/nvidia/spark/rapids/shims/spark311db/GpuRunningWindowExec.scala
mythrocks: I think I have figured out a way to do this. Testing now.

mythrocks: Sorry for the delay. This warranted testing on Apache Spark, Databricks 7.3, and Databricks 8.2. This should be resolved now.

mythrocks:
Note: The build against Databricks 8.2 is likely to fail because of https://github.com/NVIDIA/spark-rapids/issues/2628. We can re-kick the tests, once #2628 is resolved.

mythrocks:
@revans2, thanks for solving #2637. I'll rebase when that's been checked in.

mythrocks:
Rebased now, to include #2637.

mythrocks:
build

sameerz:
build

tgravescs (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -24,11 +24,82 @@ import org.apache.spark.sql.execution.{SparkPlan, UnaryExecNode}
 import org.apache.spark.sql.execution.window.WindowExec
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
-class GpuWindowExecMeta(windowExec: WindowExec,
+/**
```
abellina, line 27: please update copyrights in this file.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -712,7 +712,8 @@ object ExpressionContext {
     val parent = findParentPlanMeta(meta)
     assert(parent.isDefined, "It is expected that an aggregate function is a child of a SparkPlan")
     parent.get.wrapped match {
-      case _: WindowExecBase => WindowAggExprContext
+      case agg if ShimLoader.getSparkShims.isWindowFunctionExec(agg.asInstanceOf[SparkPlan]) =>
```
abellina, line 715: nit: this would not match on `agg` if it somehow is not a `SparkPlan`, and remove the explicit `asInstanceOf`

```
case agg: SparkPlan if ShimLoader.getSparkShims.isWindowFunctionExec(agg) =>
```

abellina (commented):
Just minor things.

revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
mythrocks: > if it somehow is not a `SparkPlan`.

I'm attempting to catch an `Exec`, i.e. either `WindowExec` or `RunningWindowFunctionExec`. Their most common base class seems to be `SparkPlan`. I wouldn't be able to handle a non-`SparkPlan` anyway. :]

> remove the explicit `asInstanceOf`

Thanks. Let me try that.

mythrocks: > if it somehow is not a `SparkPlan`.

I'm attempting to catch an `Exec`, i.e. either `WindowExec` or `RunningWindowFunctionExec`. Their most common base class seems to be `SparkPlan`. I wouldn't be able to handle a non-`SparkPlan` anyway. :]

> remove the explicit `asInstanceOf`

Thanks. Let me try that.

mythrocks: > Thanks. Let me try that.

Actually, no, I can't. `parent.get.wrapped` is apparently of type `Any`. And `isWindowFunctionExec()` expects a `SparkPlan`.  :/

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
mythrocks: D'oh! Fixed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, NVIDIA CORPORATION.
+ * Copyright (c) 2020-21, NVIDIA CORPORATION.
```
revans2, line 2: I think typically it is `2020-2021`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
mythrocks: @abellina: I can't seem to avoid the cast.Perhaps I've misunderstood the suggestion?

abellina: When you match `agg: SparkPlan`, `agg` essentially becomes: `parent.get.wrapped.asInstanceOf[SparkPlan]`. Did you get a compilation error?

abellina: Ok, I stand corrected. I tried a simpler example than this. I think `agg` is `SparkPlan` only in the body of the case. Never mind :)

mythrocks: Yep, I misunderstood the suggestion. Let me try this. Will update shortly.

revans2 (dismissed):

abellina (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
mythrocks: Yep, this is cleaner. Thanks, @abellina. Sorry I misread it the first time.

revans2 (approved):

revans2:
build

mythrocks:
Thanks for the reviews, all.

PR #2616, opened 2021-06-07
support GpuSortArray

### LINKED ISSUE #2557: [FEA] support sort_array on GPU
The support of `sort_array` function is missing now, which sorts the input array in ascending or descending order according to the natural ordering of the array elements. 

~It appears that an additional cuDF API like `ColumnVector.listSortByRow` is required. ~

sperlingxx:
Closes #2557.

This PR is to support sort_array function on GPU with cuDF method `ColumnVector.listSortRows`.

### DIFF
--- a/integration_tests/src/main/python/collection_ops_test.py
+++ b/integration_tests/src/main/python/collection_ops_test.py
@@ -27,6 +27,8 @@ nested_gens = [ArrayGen(LongGen()),
 non_nested_array_gens = [ArrayGen(sub_gen, nullable=nullable)
                          for nullable in [True, False] for sub_gen in all_gen + [null_gen]]
 
+non_null_nested_array_gens = [g for g in non_nested_array_gens if not g.nullable]
+
 @pytest.mark.parametrize('data_gen', non_nested_array_gens, ids=idfn)
 def test_concat_list(data_gen):
     assert_gpu_and_cpu_are_equal_collect(
@@ -96,10 +98,9 @@ def test_size_of_map(data_gen, size_of_null):
             lambda spark: unary_op_df(spark, data_gen).selectExpr('size(a)'),
             conf={'spark.sql.legacy.sizeOfNull': size_of_null})
 
-@pytest.mark.parametrize('data_gen', [g for g in non_nested_array_gens if g.nullable == False][:1], ids=idfn)
-@pytest.mark.parametrize('is_ascending', ['True', 'False'], ids=idfn)
+@pytest.mark.parametrize('data_gen', non_null_nested_array_gens, ids=idfn)
+@pytest.mark.parametrize('is_ascending', [True, False], ids=idfn)
 def test_sort_array(data_gen, is_ascending):
     assert_gpu_and_cpu_are_equal_collect(
-        lambda spark: unary_op_df(spark, data_gen).selectExpr('sort_array(a, false) as b'),
-        conf={}
-    )
+        lambda spark: unary_op_df(spark, data_gen).select(
+            f.sort_array(f.col('a'), is_ascending)))
--- a/integration_tests/src/main/python/collection_ops_test.py
+++ b/integration_tests/src/main/python/collection_ops_test.py
@@ -95,3 +95,11 @@ def test_size_of_map(data_gen, size_of_null):
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark: unary_op_df(spark, data_gen).selectExpr('size(a)'),
             conf={'spark.sql.legacy.sizeOfNull': size_of_null})
+
+@pytest.mark.parametrize('data_gen', [g for g in non_nested_array_gens if g.nullable == False][:1], ids=idfn)
+@pytest.mark.parametrize('is_ascending', ['True', 'False'], ids=idfn)
+def test_sort_array(data_gen, is_ascending):
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: unary_op_df(spark, data_gen).selectExpr('sort_array(a, false) as b'),
+        conf={}
+    )
--- a/integration_tests/src/main/python/collection_ops_test.py
+++ b/integration_tests/src/main/python/collection_ops_test.py
@@ -14,9 +14,10 @@
 
 import pytest
 
-from asserts import assert_gpu_and_cpu_are_equal_collect
+from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_and_cpu_are_equal_sql
 from data_gen import *
 from pyspark.sql.types import *
+from spark_session import with_cpu_session
 from string_test import mk_str_gen
 import pyspark.sql.functions as f
 
@@ -26,8 +27,6 @@ nested_gens = [ArrayGen(LongGen()),
 # additional test for NonNull Array because of https://github.com/rapidsai/cudf/pull/8181
 non_nested_array_gens = [ArrayGen(sub_gen, nullable=nullable)
                          for nullable in [True, False] for sub_gen in all_gen + [null_gen]]
-
-non_null_nested_array_gens = [g for g in non_nested_array_gens if not g.nullable]
 
 @pytest.mark.parametrize('data_gen', non_nested_array_gens, ids=idfn)
 def test_concat_list(data_gen):
@@ -98,9 +97,17 @@ def test_size_of_map(data_gen, size_of_null):
             lambda spark: unary_op_df(spark, data_gen).selectExpr('size(a)'),
             conf={'spark.sql.legacy.sizeOfNull': size_of_null})
 
-@pytest.mark.parametrize('data_gen', non_null_nested_array_gens, ids=idfn)
+@pytest.mark.parametrize('data_gen', non_nested_array_gens, ids=idfn)
 @pytest.mark.parametrize('is_ascending', [True, False], ids=idfn)
 def test_sort_array(data_gen, is_ascending):
     assert_gpu_and_cpu_are_equal_collect(
         lambda spark: unary_op_df(spark, data_gen).select(
             f.sort_array(f.col('a'), is_ascending)))
+
+@pytest.mark.parametrize('data_gen', non_nested_array_gens, ids=idfn)
+@pytest.mark.parametrize('is_ascending', [True, False], ids=idfn)
+def test_sort_array_lit(data_gen, is_ascending):
+    array_lit = gen_scalar(data_gen)
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: unary_op_df(spark, data_gen, length=10).select(
+            f.sort_array(f.lit(array_lit), is_ascending)))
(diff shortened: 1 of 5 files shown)


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -215,3 +215,56 @@ case class GpuSize(child: Expression, legacySizeOfNull: Boolean)
...
+    case ArrayType(dt, _) =>
+      val dtSimple = dt.catalogString
+      TypeCheckResult.TypeCheckFailure(
+        s"$prettyName does not support sorting array of type $dtSimple which is not orderable")
+    case _ =>
+      TypeCheckResult.TypeCheckFailure(s"$prettyName only supports array input.")
+  }
+
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector =
+    throw new IllegalStateException("This is not supported")
```
abellina, line 249: Suggest: 
```
    throw new IllegalArgumentException("lhs has to be a vector and rhs has to be a scalar for " +
      "the sort_array operator to work")
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -215,3 +215,56 @@ case class GpuSize(child: Expression, legacySizeOfNull: Boolean)
...
+
+  override def doColumnar(lhs: GpuScalar, rhs: GpuColumnVector): ColumnVector =
+    throw new IllegalStateException("This is not supported")
+
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuScalar): ColumnVector = {
+    val isDescending = isDescendingOrder(rhs)
+    lhs.getBase.listSortRows(isDescending, true)
+  }
+
+  override def doColumnar(numRows: Int, lhs: GpuScalar, rhs: GpuScalar): ColumnVector = {
```
abellina, line 259: so we can end up here if the expressions we are working with evaluate to scalars? Could you add a test that shows this use case? I could be missing it from the integration test.

integration_tests/src/main/python/collection_ops_test.py
```diff
@@ -27,6 +27,8 @@
 non_nested_array_gens = [ArrayGen(sub_gen, nullable=nullable)
                          for nullable in [True, False] for sub_gen in all_gen + [null_gen]]
 
+non_null_nested_array_gens = [g for g in non_nested_array_gens if not g.nullable]
```
revans2, line 30: why non-null? If we cannot support null ordering properly we either have to fall back to the CPU when we see it or we have to fix it in cudf.  Becuase we don't really have a customer asking for this right now I would rather see us fall back to the CPU.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2362,6 +2362,20 @@ object GpuOverrides {
...
           GpuArrayContains(lhs, rhs)
       }),
+    expr[SortArray](
+      "Returns a sorted array with the input array and the ascending / descending order",
+      ExprChecks.binaryProjectNotLambda(
+        TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL),
+        TypeSig.ARRAY.nested(TypeSig.all),
+        ("array", TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL),
+            TypeSig.ARRAY.nested(TypeSig.all)),
+        ("ascendingOrder", TypeSig.BOOLEAN, TypeSig.BOOLEAN)),
```
revans2, line 2372: If you only support literal values for this. then tag it as such.

```
("ascendingOrder",  TypeSig.lit(TypeEnum.BOOLEAN), TypeSig.BOOLEAN)
```

That will guarantee that we fall back to the CPU if we see a non-literal value.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2362,6 +2362,20 @@ object GpuOverrides {
...
+      (in, conf, p, r) => new BinaryExprMeta[SortArray](in, conf, p, r) {
+        override def convertToGpu(lhs: Expression, rhs: Expression): GpuExpression = {
```
revans2, line 2374: If we don't support literal arrays we have to tag it here so we fall back to the CPU.

revans2 (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2362,6 +2362,20 @@ object GpuOverrides {
         override def convertToGpu(lhs: Expression, rhs: Expression): GpuExpression =
           GpuArrayContains(lhs, rhs)
       }),
+    expr[SortArray](
+      "Returns a sorted array with the input array and the ascending / descending order",
+      ExprChecks.binaryProjectNotLambda(
+        TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL),
+        TypeSig.ARRAY.nested(TypeSig.all),
+        ("array", TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL),
```
gerashegalov, line 2370: we have a constant for this set of types 
```suggestion
        ("array", TypeSig.ARRAY.nested(_commonTypes),
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2362,6 +2362,20 @@ object GpuOverrides {
         override def convertToGpu(lhs: Expression, rhs: Expression): GpuExpression =
           GpuArrayContains(lhs, rhs)
       }),
+    expr[SortArray](
+      "Returns a sorted array with the input array and the ascending / descending order",
+      ExprChecks.binaryProjectNotLambda(
+        TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL),
```
gerashegalov, line 2368:
```suggestion
        TypeSig.ARRAY.nested(_commonTypes),
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2362,6 +2362,20 @@ object GpuOverrides {
...
+        TypeSig.ARRAY.nested(TypeSig.all),
+        ("array", TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL),
+            TypeSig.ARRAY.nested(TypeSig.all)),
+        ("ascendingOrder", TypeSig.BOOLEAN, TypeSig.BOOLEAN)),
+      (in, conf, p, r) => new BinaryExprMeta[SortArray](in, conf, p, r) {
```
gerashegalov, line 2373: maybe easier to read if we adjusted naming of components of pattern matching to the specific case like 
`(sortExpression, conf, p,  r)  => `

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -215,3 +215,56 @@ case class GpuSize(child: Expression, legacySizeOfNull: Boolean)
...
+
+  override def inputTypes: Seq[AbstractDataType] = Seq(ArrayType, BooleanType)
+
+  override def checkInputDataTypes(): TypeCheckResult = base.dataType match {
+    case ArrayType(dt, _) if RowOrdering.isOrderable(dt) =>
+      ascendingOrder match {
+        // replace Literal with GpuLiteral here
+        case GpuLiteral(_: Boolean, BooleanType) =>
+          TypeCheckResult.TypeCheckSuccess
+        case _ =>
```
gerashegalov, line 236: for easier diagnostics let us log what was provided. Same on L244

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
sperlingxx: Changed. Thanks!

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
sperlingxx: Yes, I made up the test case for scalar.

sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
sperlingxx: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala

integration_tests/src/main/python/collection_ops_test.py
sperlingxx: Sorry, it shouldn't be non-null specific at here. I removed the non-null constriction on test data.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: I think we can support literal arrays. And I made up the corresponding test.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: Yes, and if I am not wrong, sort_array also only supports `ascendingOrder` as literal value in CPU. Because its `checkInputDataTypes` method only returns `TypeCheckSuccess` if `ascendingOrder` is a Literal.
```scala
  override def checkInputDataTypes(): TypeCheckResult = base.dataType match {
    case ArrayType(dt, _) if RowOrdering.isOrderable(dt) =>
      ascendingOrder match {
        case Literal(_: Boolean, BooleanType) =>
          TypeCheckResult.TypeCheckSuccess
        case _ =>
          TypeCheckResult.TypeCheckFailure(
            "Sort order in second argument requires a boolean literal.")
      }
    case ArrayType(dt, _) =>
      val dtSimple = dt.catalogString
      TypeCheckResult.TypeCheckFailure(
        s"$prettyName does not support sorting array of type $dtSimple which is not orderable")
    case _ =>
      TypeCheckResult.TypeCheckFailure(s"$prettyName only supports array input.")
  }
```

revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
revans2: You are 100% correct. I missed that part.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
revans2: What you wrote is good enough for me. The test is likely not testing what we expect it to, but the code looks like it should work fine so I am OK with it.

PR #2612, opened 2021-06-07
Profiling tool, add in job to stage, duration, executor cpu time, fix writing to HDFS

tgravescs:
This has a few  things in it. 

- change file writer to work with writing to HDFS
- Add support for profiling tool printing job to stages to sqlid
- Add profiling tool printing the sql duration and executor cpu time, and shows with contains dataset

Output for new profiling tables look like:

```
Job Information:
+--------+-----+------------+-----+
|appIndex|jobID|stageIds    |sqlID|
+--------+-----+------------+-----+
|1       |0    |[0]         |null |
|1       |1    |[1, 2, 3, 4]|0    |
+--------+-----+------------+-----+
```

```
SQL Duration and Executor CPU Time Percent
+--------+------------------------------+-----+------------+-------------------+------------+------------------+-------------------------+
|appIndex|App ID                        |sqlID|SQL Duration|Contains Dataset Op|App Duration|Potential Problems|Executor CPU Time Percent|
+--------+------------------------------+-----+------------+-------------------+------------+------------------+-------------------------+
|1       |application_1603128018386_7759|0    |11042       |false              |119990      |null              |68.48                    |
+--------+------------------------------+-----+------------+-------------------+------------+------------------+-------------------------+

```

PR #2618, opened 2021-06-07
Correct an issue for README for tools and also correct s3 solution in Args.scala

viadea:
Correct an issue for README for tools and also correct s3 solution in Args.scala

### DIFF
--- a/tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/ProfileArgs.scala
+++ b/tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/ProfileArgs.scala
@@ -33,8 +33,7 @@ rapids-4-spark-tools_2.12-<version>.jar /path/to/eventlog1 /path/to/eventlog2
  rapids-4-spark-tools_2.12-<version>.jar /path/to/DirOfManyEventLogs
 
 # If any event log is from S3:
-# Need to download hadoop-aws-<version>.jar and aws-java-sdk-<version>.jar firstly,
-# and then put fs.s3a.access.key and fs.s3a.secret.key in $SPARK_HOME/conf/hdfs-site.xml.
+# Need to download hadoop-aws-<version>.jar and aws-java-sdk-<version>.jar firstly.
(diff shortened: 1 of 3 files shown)


tgravescs (approved):
changes look fine, will like have conflicts when https://github.com/NVIDIA/spark-rapids/pull/2612 is merged

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/ProfileArgs.scala
```diff
@@ -33,8 +33,8 @@ rapids-4-spark-tools_2.12-<version>.jar /path/to/eventlog1 /path/to/eventlog2
  rapids-4-spark-tools_2.12-<version>.jar /path/to/DirOfManyEventLogs
 
 # If any event log is from S3:
-export AWS_ACCESS_KEY_ID=xxx
-export AWS_SECRET_ACCESS_KEY=xxx
+# Need to download hadoop-aws-<version>.jar and aws-java-sdk-<version>.jar firstly,
+# and then put fs.s3a.access.key and fs.s3a.secret.key in $SPARK_HOME/conf/hdfs-site.xml.
```
gerashegalov, line 37: S3a is a separate filesystem from HDFS. It's typically stored in core-site.xml. But you can also just configure in spark via  `spark.hadoop.fs.s3a...` settings. Environment variables should work too if EnvironmentVariableCredentialsProvider is configured. 

I suggest  that we don't go into details in our repo, and simply defer/link Hadoop/Spark docs for this.

gerashegalov (changes requested):

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/ProfileArgs.scala
viadea: @gerashegalov I just added a doc link in README and removed the instructions in 2 Args.scala files. 
How is now?thx

gerashegalov (approved):

tgravescs (approved):

tgravescs:
build

nartal1 (approved):

PR #2620, opened 2021-06-07
Change aggregation of executor CPU and run time for Qualification tool to speed up query

tgravescs:
Running the qualification tool over all nds results and having it include the executor cpu time percent was taking about 45 minutes and required 30G driver memory.  To address this I changed the way we aggregate those metrics just for the qualification side so that we don't need to keep all tasks and create the individual tasks table.  Now it just aggregates them up per stage on the fly and we can simply look it up per stage and aggregate them based on that.  That improved the run time to be about 15 minutes and I used 5G memory (I might be able to use less, this is what I set it at).

since the performance is better I turned it on by default and changed config to allow removing it if you don't want it for some reason.

### DIFF
--- a/tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/EventsProcessor.scala
+++ b/tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/EventsProcessor.scala
@@ -262,8 +262,8 @@ class EventsProcessor(forQualification: Boolean = false) extends Logging {
       val taskSum = app.stageTaskQualificationEnd.getOrElseUpdate(stageAndAttempt, {
         new StageTaskQualificationSummary(event.stageId,
           event.stageAttemptId,
-          event.taskMetrics.executorRunTime,
-          NANOSECONDS.toMillis(event.taskMetrics.executorCpuTime))
+          0,
+          0)
       })
       taskSum.executorRunTime += event.taskMetrics.executorRunTime
       taskSum.executorCPUTime += NANOSECONDS.toMillis(event.taskMetrics.executorCpuTime)
--- a/tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/EventsProcessor.scala
+++ b/tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/EventsProcessor.scala
@@ -259,8 +259,8 @@ class EventsProcessor(forQualification: Boolean = false) extends Logging {
 
     if (forQualification) {
       val stageAndAttempt = s"${event.stageId}:${event.stageAttemptId}"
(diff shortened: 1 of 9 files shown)


tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/EventsProcessor.scala
```diff
@@ -26,9 +28,9 @@ import org.apache.spark.scheduler._
 import org.apache.spark.sql.execution.ui.{SparkListenerDriverAccumUpdates, SparkListenerSQLAdaptiveExecutionUpdate, SparkListenerSQLAdaptiveSQLMetricUpdates, SparkListenerSQLExecutionEnd, SparkListenerSQLExecutionStart}
 
 /**
- * This object is to process all events and do validation in the end.
+ * This class is to process all events and do validation in the end.
  */
-object EventsProcessor extends Logging {
+class EventsProcessor(forQualification: Boolean = false) extends Logging {
```
gerashegalov, line 33: maybe make an enum to make it easier to add different use cases later

tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/EventsProcessor.scala
```diff
@@ -246,75 +248,86 @@ object EventsProcessor extends Logging {
...
+      for (res <- event.taskInfo.accumulables) {
+        try {
+          val value = res.value.getOrElse("").toString.toLong
+          val thisMetric = TaskStageAccumCase(
+            event.stageId, event.stageAttemptId, Some(event.taskInfo.taskId),
+            res.id, res.name, Some(value), res.internal)
+          app.taskStageAccum += thisMetric
+        } catch {
+          case e: ClassCastException =>
+            logWarning("ClassCastException when parsing accumulables for task "
```
gerashegalov, line 279: Since we are already modify this code, maybe let us combine it into a single log statement for the same issue instead of three

tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/EventsProcessor.scala
tgravescs: we can do this later when we don't have so tight deadline, its functionally correct for now

tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/EventsProcessor.scala
tgravescs: same here, it's not required right now so we can do followup.  I would like to make pass through all code next release to clean some things up

nartal1:
build

nartal1 (approved):

gerashegalov (approved):

PR #2623, opened 2021-06-08
Fixed minor mistakes in documentation

razajafri:

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -36,21 +36,6 @@ The jar is generated in below directory :
 
 `./tools/target/rapids-4-spark-tools_2.12-<version>.jar`
 
-## How to use this tool
-
-This tool parses the spark event log(s) and creates an output report.
-Acceptable input spark event log paths are local filesystem,hdfs,S3 or mixed.
-It can be file path,directory path or mixed.
-
-Below are examples of possible inputs:
-```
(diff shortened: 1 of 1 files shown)


tools/README.md
```diff
@@ -36,9 +36,22 @@ The jar is generated in below directory :
 
 `./tools/target/rapids-4-spark-tools_2.12-<version>.jar`
 
+## How to use this tool
```
nartal1, line 39: We have `How to use this tool` section in both qualification and profiling sections below but it doesn't have examples of possible inputs.  Would it make sense to remove redundancy. Either keep this here and remove from qualification and profiling sections OR  remove here and move `possible inputs`  to both qualification and profiling.

razajafri: Yes, the `how to use this tool` headline can be renamed IMO for this section. May be `Overview of tools` or something similar?

razajafri: @tgravescs what do you think?

tgravescs: I specifically move the  How to use this tool sections below each one because they are 2 different tools.

Are those sections not clear?

razajafri: I think, I updated an older version of the doc before your changes got applied. I will remove this change.

tgravescs (approved):

razajafri:
build

PR #2651, opened 2021-06-09
Profiling tool - fix reporting contains dataset when sql time 0

tgravescs:
There was a bug on profiling side with reporting contains dataset when the sql time was 0.  Simply add another field to the table to explicitly set it.


tgravescs:
build

nartal1 (approved):

PR #2657, opened 2021-06-09
Print CPU and GPU output when _assert_equal fails to help debug given…

abellina:
… the failed path

This seems like a simple way to capture the output of a test on failure, to help debug. One could take this output and them diff the cpu and the gpu, so one can match that with the "path", as already provided by the test framework on failure.

Re: https://github.com/NVIDIA/spark-rapids/issues/2477.


integration_tests/src/main/python/asserts.py
```diff
@@ -90,7 +90,11 @@ def _assert_equal(cpu, gpu, float_check, path):
 
 def assert_equal(cpu, gpu):
     """Verify that the result from the CPU and the GPU are equal"""
-    _assert_equal(cpu, gpu, float_check=get_float_check(), path=[])
+    try:
+      _assert_equal(cpu, gpu, float_check=get_float_check(), path=[])
+    except Exception as ex:
+      print("CPU OUTPUT: %s \n\n GPU OUTPUT: %s" % (cpu, gpu))
+      raise ex
```
gerashegalov, line 97: nit: we can re-raise the last exception with less boilerplate
```python
try:
    ...
except:
    ...
    raise
```

integration_tests/src/main/python/asserts.py
```diff
@@ -90,7 +90,11 @@ def _assert_equal(cpu, gpu, float_check, path):
 
 def assert_equal(cpu, gpu):
     """Verify that the result from the CPU and the GPU are equal"""
-    _assert_equal(cpu, gpu, float_check=get_float_check(), path=[])
+    try:
+      _assert_equal(cpu, gpu, float_check=get_float_check(), path=[])
+    except Exception as ex:
+      print("CPU OUTPUT: %s \n\n GPU OUTPUT: %s" % (cpu, gpu))
```
gerashegalov, line 96: nit: multi-line f string

integration_tests/src/main/python/asserts.py
abellina: done

integration_tests/src/main/python/asserts.py
abellina: Not entirely sure if I understood this comment. But I split this into two print statements, please let me know if I misunderstood.

integration_tests/src/main/python/asserts.py
```diff
@@ -90,7 +90,12 @@ def _assert_equal(cpu, gpu, float_check, path):
 
 def assert_equal(cpu, gpu):
     """Verify that the result from the CPU and the GPU are equal"""
-    _assert_equal(cpu, gpu, float_check=get_float_check(), path=[])
+    try:
+      _assert_equal(cpu, gpu, float_check=get_float_check(), path=[])
+    except:
+      print("CPU OUTPUT: %s" % cpu)
+      print("GPU OUTPUT: %s" % gpu)
```
gerashegalov, line 97: this looks good.

I the original comment I was thinking:
```
      print(f'CPU OUTPUT: {cpu}\n'
            f'GPU OUTPUT: {gpu}')
```

gerashegalov (approved):

sameerz:
build

pxLi:
build

pxLi:
build

revans2 (approved):

integration_tests/src/main/python/asserts.py
revans2, line 97: nit: technically this is not guaranteed to be the output for the GPU or the CPU because we could have sorted it locally first, but I think everyone will understand and I think this is the right place to print out the info.

revans2:
@abellina is this supposed to be 21.06? Should this be updated to 21.08?

PR #2658, opened 2021-06-09
Qualification tool - add in estimating the App end time when the event log missing application end event

tgravescs:
Many of the databricks event logs are missing the application end event so the application duration comes out as 0.  To allow us to still score it, use the latest job end time or sql execution time instead and then add a column to indicate that we estimated the app duration in case user wants to know.

Note this does not change it for the profiling tool.


tgravescs:
build

tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/ApplicationInfo.scala
```diff
@@ -320,15 +321,24 @@ class ApplicationInfo(
     if (this.appStart.nonEmpty) {
       val appStartNew: ArrayBuffer[ApplicationCase] = ArrayBuffer[ApplicationCase]()
       for (res <- this.appStart) {
-        val durationResult = ProfileUtils.OptionLongMinusLong(this.appEndTime, res.startTime)
+
+        val estimatedResult = Option(this.appEndTime.getOrElse {
+          logWarning("Application End Time is unknown, estimating based on job and sql end times!")
+          // estimate the app end with job or sql end times
+          val sqlEndTime = this.sqlEndTime.values.toSeq.sorted(Ordering[Long].reverse).head
```
gerashegalov, line 328: what if sqlEndTimes (or  jobEndTime) is empty? Let us use something like `headOption.getOrElse(0L)

we don't need sort to determine the max. `if (sqlEndtime.isEmpty) 0L else sqlEndTime.values.max`

tgravescs: good catch

tgravescs:
I'm going to close this and rebase and open a new one

Issue #2652, opened 2021-06-09
[BUG] No Job Found. Exiting.


revans2:
I was following the instructions at 

https://github.com/NVIDIA/spark-rapids/tree/branch-21.08/tools

It took a bit to be able to find where the event logs are stored (It was not obvious and took a but of digging), but when I had found them at that point everything in the directory is called something like `app-<timestamp>-number`. I had no idea what any of these apps were so I decided to use all of them

```
spark-submit --driver-memory 5g --class com.nvidia.spark.rapids.tool.qualification.QualificationMain ./tools/target/rapids-4-spark-tools_2.12-21.08.0-SNAPSHOT.jar spark-eventlog/*
```

And the logging is far from clear what happened or why there was an error.

```
21/06/09 12:51:51 INFO ApplicationInfo: Parsing Event Log File: .../spark-eventlog/app-20201029093814-0001
21/06/09 12:51:52 INFO ApplicationInfo: Total number of events parsed: 15
21/06/09 12:51:53 INFO SharedState: Setting hive.metastore.warehouse.dir ('null') to the value of spark.sql.warehouse.dir ('spark-warehouse').
21/06/09 12:51:53 INFO SharedState: Warehouse path is 'spark-warehouse'.
21/06/09 12:51:53 INFO CodeGenerator: Code generated in 153.741923 ms
21/06/09 12:51:54 INFO ApplicationInfo: No SQL Execution Found. Skipping generating SQL Execution DataFrame.
21/06/09 12:51:54 ERROR ApplicationInfo: No Job Found. Exiting.
```

From what I can tell `app-20201029093814-0001` didn't have any SQL execution in it. OK so skip it or print that out, but go on and process everything else that I passed in.  Don't exit.

tgravescs:
I'm guessing the * at the end only expanded to the single event log, we should verify though.

Leave off the start to read all files in a directory or specify them individually.

tgravescs:
>> It took a bit to be able to find where the event logs are stored (It was not obvious and took a but of digging),

Can you clarify this?  you didn't know where spark put them on just any application run?  If so we should perhaps document the spark configs used or point to spark docs

revans2:
It would be good to point to the spark docs/etc. I had to look up in my config to again to find the directory. Not the end of the world, but it was not simple.

> I'm guessing the * at the end only expanded to the single event log, we should verify though.
>
> Leave off the star to read all files in a directory or specify them individually.

[...]

PR #2663, opened 2021-06-09
Qualification tool - add in estimating the App end time when the event log missing application end event

tgravescs:
Many of the databricks event logs are missing the application end event so the application duration comes out as 0. To allow us to still score it, use the latest job end time or sql execution time instead and then add a column to indicate that we estimated the app duration in case user wants to know.

Note this does not change it for the profiling tool.

### DIFF
--- a/tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/ApplicationInfo.scala
+++ b/tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/ApplicationInfo.scala
@@ -19,6 +19,7 @@ package org.apache.spark.sql.rapids.tool.profiling
 import scala.collection.Map
 import scala.collection.mutable.{ArrayBuffer, HashMap}
 import scala.io.{Codec, Source}
+import scala.math.Ordering
 
 import com.nvidia.spark.rapids.tool.ToolTextFileWriter
 import com.nvidia.spark.rapids.tool.profiling._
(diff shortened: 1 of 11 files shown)


tgravescs:
build

tgravescs:
@gerashegalov new pr, I addressed your one comment. I plan to add more tests related to empty jobs/sql under https://github.com/NVIDIA/spark-rapids/issues/2652

gerashegalov (approved):

tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/ApplicationInfo.scala
```diff
@@ -320,15 +321,33 @@ class ApplicationInfo(
...
       val appStartNew: ArrayBuffer[ApplicationCase] = ArrayBuffer[ApplicationCase]()
       for (res <- this.appStart) {
-        val durationResult = ProfileUtils.OptionLongMinusLong(this.appEndTime, res.startTime)
+
+        val estimatedResult =
+          this.appEndTime match {
+            case Some(t) => this.appEndTime
+            case None =>
+              if (this.sqlEndTime.isEmpty && this.jobEndTime.isEmpty) {
+                None
```
nartal1, line 330: Just for my understanding. If both are empty then it throws NoSuchElementException?

nartal1 (approved):

tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/ApplicationInfo.scala
tgravescs: it returns None which is set to estimatedResult and that gets passed into OptionLongMinusLong which will just end up returning None for the durationResult which will result in null duration in table

nartal1: Thanks !

PR #2670, opened 2021-06-10
WIP: allow using host memory when GDS spill is enabled

rongou:


rongou:
This doesn't seem to help with GDS spill performance.

PR #2671, opened 2021-06-10
Clean profiling&qualification tool README

viadea:

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -161,9 +161,9 @@ rapids-4-spark-tools_2.12-<version>.jar \
 /path/to/eventlog1 /path/to/eventlog2 /directory/with/eventlogs
 ```
 
-Note: We suggest you also save the output of the `spark-submit` to a log file for troubleshooting.
+### Options (`--help` output)
 
-### Options (How to run `--help`)
+  Note: `--help` should be before the trailing event logs.
 ```bash
 $SPARK_HOME/bin/spark-submit \
 --class com.nvidia.spark.rapids.tool.qualification.QualificationMain \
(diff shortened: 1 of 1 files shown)


tools/README.md
```diff
@@ -137,7 +161,9 @@ rapids-4-spark-tools_2.12-<version>.jar \
 /path/to/eventlog1 /path/to/eventlog2 /directory/with/eventlogs
 ```
 
-### Options
+Note: We suggest you also save the output of the `spark-submit` to a log file for troubleshooting.
+
+### Options (How to run `--help`)
```
tgravescs, line 166: This seems out of place? What is the how to run —help? Can we add a statement saying help must be before the event logs files

viadea: 1. For the `spark-submit` saving output section, initially I put after "how to" section. Now I moved to "output" section.
2. Basically the "options" section is the output of "--help".  So I just changed "How to run --help" to "--help output".
3. Also put a words saying "--help should be before the trailing event logs." in the "options" section.

How is now?

tools/README.md
```diff
@@ -12,7 +12,8 @@ GPU generated event logs.
 ## Prerequisites
 - Spark 3.0.1 or newer installed
 - Java 8 or above
-- Complete Spark event log(s) from Spark 3.0 or above version.
+- Complete Spark event log(s) from Spark 3.0 or above version
```
tgravescs, line 15: so one other thing we might mention is how to find event logs.  Perhaps reference spark config 

spark.eventLog.dir



<br class="Apple-interchange-newline">

viadea: I would suggest they just use UI , and there is a download button on the right side.
It is easier, especially for CSP users because many CSP users may not even ssh to the nodes.

viadea: Just added on how to download spark event log.

tgravescs: can we say both, because if the applications have already run, they can look in the event log directory. You can find it by look at the spark config spark.eventLog.dir

tgravescs: if you push another change ask it to build so the build goes overnight

viadea: Sure. Done.

tgravescs (approved):

tgravescs:
build

pxLi:
build

Issue #2382, opened 2021-05-10
Support writing Intervals on GPU when caching


razajafri:
If a DataFrame containing intervals is cached, today we will cache it only using CPU as cudf doesn't support intervals. But neither does Parquet-MR or VectorizedWriter support writing intervals, but  we get around this by decomposing intervals to months, years and microseconds. We should do the same when writing on the GPU as the ParquetWriter now supports Structs in cudf

razajafri:
This may improve performance of writing/reading cache but to do any operation on a table with intervals, we will have to fallback on the CPU as cudf doesn't support intervals atm. I also don't think people generally cache tables with intervals, so unless we see a use case this is a much lower priority than we originally thought.

PR #2675, opened 2021-06-10
Support for Decimals with negative scale for Parquet Cached Batch Serializer

razajafri:
PCBS stores cache in parquet format which is good because it compresses it but the drawback is that we have to handle data that parquet doesn't one example is Decimals with negative scale, which is supported as a Legacy feature in Spark.

This PR deals with storing the unscaled long value associated with Decimals so we can support negative scale. This PR also lays the foundation for possible improvement in performance which is a part of #1143 

### DIFF
--- a/integration_tests/src/main/python/cache_test.py
+++ b/integration_tests/src/main/python/cache_test.py
@@ -21,14 +21,15 @@ from spark_session import with_cpu_session, with_gpu_session
 from join_test import create_df
 from marks import incompat, allow_non_gpu, ignore_order
 
-enableVectorizedConf = [{"spark.sql.inMemoryColumnarStorage.enableVectorizedReader" : "true"},
-                        {"spark.sql.inMemoryColumnarStorage.enableVectorizedReader" : "false"}]
-allowNegativeDecimalConf = {"spark.sql.legacy.allowNegativeScaleOfDecimal" : "true"}
+conf = [{"spark.sql.inMemoryColumnarStorage.enableVectorizedReader": "true",
+         "spark.sql.legacy.allowNegativeScaleOfDecimal": "true"},
+        {"spark.sql.inMemoryColumnarStorage.enableVectorizedReader": "false",
+         "spark.sql.legacy.allowNegativeScaleOfDecimal": "true"}]
 decimal_struct_gen= StructGen([['child'+str(ind), sub_gen] for ind, sub_gen in enumerate(decimal_gens)])
 
-@pytest.mark.parametrize('enableVectorizedConf', enableVectorizedConf, ids=idfn)
+@pytest.mark.parametrize('conf', conf, ids=idfn)
 @allow_non_gpu('CollectLimitExec')
-def test_passing_gpuExpr_as_Expr(enableVectorizedConf):
+def test_passing_gpuExpr_as_Expr(conf):
     assert_gpu_and_cpu_are_equal_collect(
         lambda spark : unary_op_df(spark, string_gen)
             .select(f.col("a")).na.drop()
@@ -36,7 +37,7 @@ def test_passing_gpuExpr_as_Expr(enableVectorizedConf):
             .agg(f.count(f.col("a")).alias("count_a"))
             .orderBy(f.col("count_a").desc(), f.col("a"))
             .cache()
-            .limit(50), enableVectorizedConf)
+            .limit(50), conf)
 
 # creating special cases to just remove -0.0 because of https://github.com/NVIDIA/spark-rapids/issues/84
 # After 3.1.0 is the min Spark version we can drop this
@@ -56,29 +57,25 @@ all_gen = [StringGen(), ByteGen(), ShortGen(), IntegerGen(), LongGen(),
 
 @pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
 @pytest.mark.parametrize('join_type', ['Left', 'Right', 'Inner', 'LeftSemi', 'LeftAnti'], ids=idfn)
-@pytest.mark.parametrize('enableVectorizedConf', enableVectorizedConf, ids=idfn)
+@pytest.mark.parametrize('conf', conf, ids=idfn)
 @ignore_order
-def test_cache_join(data_gen, join_type, enableVectorizedConf):
+def test_cache_join(data_gen, join_type, conf):
     def do_join(spark):
         left, right = create_df(spark, data_gen, 500, 500)
         cached = left.join(right, left.a == right.r_a, join_type).cache()
         cached.count() # populates cache
         return cached
-    conf = enableVectorizedConf.copy()
-    conf.update(allowNegativeDecimalConf)
     assert_gpu_and_cpu_are_equal_collect(do_join, conf = conf)
 
 @pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
 @pytest.mark.parametrize('join_type', ['Left', 'Right', 'Inner', 'LeftSemi', 'LeftAnti'], ids=idfn)
-@pytest.mark.parametrize('enableVectorizedConf', enableVectorizedConf, ids=idfn)
+@pytest.mark.parametrize('conf', conf, ids=idfn)
 # We are OK running everything on CPU until we complete 'https://github.com/NVIDIA/spark-rapids/issues/360'
 # because we have an explicit check in our code that disallows InMemoryTableScan to have anything other than
 # AttributeReference
 @allow_non_gpu(any=True)
 @ignore_order
-def test_cached_join_filter(data_gen, join_type, enableVectorizedConf):
-    conf = enableVectorizedConf.copy()
-    conf.update(allowNegativeDecimalConf)
+def test_cached_join_filter(data_gen, join_type, conf):
     data = data_gen
     def do_join(spark):
         left, right = create_df(spark, data, 500, 500)
@@ -89,12 +86,10 @@ def test_cached_join_filter(data_gen, join_type, enableVectorizedConf):
     assert_gpu_and_cpu_are_equal_collect(do_join, conf = conf)
 
 @pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
-@pytest.mark.parametrize('enableVectorizedConf', enableVectorizedConf, ids=idfn)
+@pytest.mark.parametrize('conf', conf, ids=idfn)
 @pytest.mark.parametrize('join_type', ['Left', 'Right', 'Inner', 'LeftSemi', 'LeftAnti'], ids=idfn)
 @ignore_order
-def test_cache_broadcast_hash_join(data_gen, join_type, enableVectorizedConf):
-    conf = enableVectorizedConf.copy()
-    conf.update(allowNegativeDecimalConf)
+def test_cache_broadcast_hash_join(data_gen, join_type, conf):
     def do_join(spark):
         left, right = create_df(spark, data_gen, 500, 500)
         cached = left.join(right.hint("broadcast"), left.a == right.r_a, join_type).cache()
@@ -108,12 +103,10 @@ shuffled_conf = {"spark.sql.autoBroadcastJoinThreshold": "160",
                  "spark.sql.shuffle.partitions": "2"}
 
 @pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
-@pytest.mark.parametrize('enableVectorizedConf', enableVectorizedConf, ids=idfn)
+@pytest.mark.parametrize('conf', conf, ids=idfn)
 @pytest.mark.parametrize('join_type', ['Left', 'Right', 'Inner', 'LeftSemi', 'LeftAnti'], ids=idfn)
 @ignore_order
-def test_cache_shuffled_hash_join(data_gen, join_type, enableVectorizedConf):
-    conf = enableVectorizedConf.copy()
-    conf.update(allowNegativeDecimalConf)
+def test_cache_shuffled_hash_join(data_gen, join_type, conf):
     def do_join(spark):
         left, right = create_df(spark, data_gen, 50, 500)
         cached = left.join(right, left.a == right.r_a, join_type).cache()
@@ -122,12 +115,10 @@ def test_cache_shuffled_hash_join(data_gen, join_type, enableVectorizedConf):
     assert_gpu_and_cpu_are_equal_collect(do_join, conf = conf)
 
 @pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
-@pytest.mark.parametrize('enableVectorizedConf', enableVectorizedConf, ids=idfn)
+@pytest.mark.parametrize('conf', conf, ids=idfn)
 @pytest.mark.parametrize('join_type', ['Left', 'Right', 'Inner', 'LeftSemi', 'LeftAnti'], ids=idfn)
 @ignore_order
-def test_cache_broadcast_nested_loop_join(data_gen, join_type, enableVectorizedConf):
-    conf = enableVectorizedConf.copy()
-    conf.update(allowNegativeDecimalConf)
+def test_cache_broadcast_nested_loop_join(data_gen, join_type, conf):
     def do_join(spark):
         left, right = create_df(spark, data_gen, 50, 25)
         cached = left.crossJoin(right.hint("broadcast")).cache()
@@ -136,11 +127,9 @@ def test_cache_broadcast_nested_loop_join(data_gen, join_type, enableVectorizedC
     assert_gpu_and_cpu_are_equal_collect(do_join, conf = conf)
 
 @pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
-@pytest.mark.parametrize('enableVectorizedConf', enableVectorizedConf, ids=idfn)
+@pytest.mark.parametrize('conf', conf, ids=idfn)
 @ignore_order
-def test_cache_expand_exec(data_gen, enableVectorizedConf):
-    conf = enableVectorizedConf.copy()
-    conf.update(allowNegativeDecimalConf)
+def test_cache_expand_exec(data_gen, conf):
     def op_df(spark, length=2048, seed=0):
         cached = gen_df(spark, StructGen([
             ('a', data_gen),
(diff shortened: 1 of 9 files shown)


integration_tests/src/main/python/cache_test.py
```diff
@@ -64,7 +64,7 @@ def do_join(spark):
         cached = left.join(right, left.a == right.r_a, join_type).cache()
         cached.count() # populates cache
         return cached
-
+    enableVectorizedConf.update(allowNegativeDecimalConf)
```
revans2, line 67: This actually changes the global `enableVectorizedConf`, so any test that happens to run in the same process after this one runs that also uses  `enableVectorizedConf` will get the modified version.  I don't think we want that.

integration_tests/src/main/python/cache_test.py
```diff
@@ -140,10 +140,11 @@ def op_df(spark, length=2048, seed=0):
     assert_gpu_and_cpu_are_equal_collect(op_df, conf = enableVectorizedConf)
 
 @pytest.mark.parametrize('data_gen', [all_basic_struct_gen, StructGen([['child0', StructGen([['child1', byte_gen]])]]),
-                                      parquet_decimal_struct_gen] + all_gen, ids=idfn)
+                                      decimal_struct_gen] + all_gen, ids=idfn)
 @pytest.mark.parametrize('enableVectorizedConf', enableVectorizedConf, ids=idfn)
 @allow_non_gpu('CollectLimitExec')
 def test_cache_partial_load(data_gen, enableVectorizedConf):
+    enableVectorizedConf.update(allowNegativeDecimalConf)
```
revans2, line 147: Here too same problem

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -350,61 +366,85 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
+    val schemaWithUnambiguousNames = StructType(schema.fields.indices.map(index => {
+      val field = schema.fields(index)
+      StructField(s"_col$index", field.dataType, field.nullable, field.metadata)
+    }))
+    // We want to map the name of the cols we are casting so we can cast them back when reading
+    // back the cache
+    val castMap = new scala.collection.mutable.HashMap[String, DType]()
+    val columns = for (i <- 0 until oldGpuCB.numCols()) yield {
+      val gpuVector = oldGpuCB.column(i).asInstanceOf[GpuColumnVector]
+      gpuVector.dataType() match {
```
revans2, line 384: What about a struct with a negative decimal in it? or an array with one in it?  How do we handle those?

razajafri: After talking to @hyperbolic2346 it turns out the cudf is already writing decimals as unscaled values which is why we support negative scaled decimals on the GPU. I have modified the code to only cast decimals to int and long when the scale < 0 and only for the CPU version of PCBS.

razajafri:
@revans2 thanks for reviewing. I am still working on the case where a dataframe is cached on GPU but read on the CPU. should have something by mid-day

razajafri:
@revans2 Can you please review this again? I have added the nested capability when writing parquet on the GPU as per your original feedback. I wanted to avoid casting Decimals when writing on GPU because we are writing them as INTs to begin with but that would be a problem if a user cached a Dataframe on GPU and tried to read it on CPU.

integration_tests/src/main/python/cache_test.py
```diff
@@ -194,6 +200,25 @@ def write_read_parquet_cached(spark):
...
+@pytest.mark.parametrize('conf', conf, ids=idfn)
+def test_cache_cpu_gpu_mixed(data_gen, conf):
+    def func(spark):
+        df = binary_op_df(spark, data_gen)
+        df.cache().count()
+        enabled = spark.conf.get("spark.rapids.sql.enabled")
+        if (enabled == "true"):
+            spark.conf.set("spark.rapids.sql.enabled", "false")
+        else:
+            spark.conf.set("spark.rapids.sql.enabled", "true")
```
gerashegalov, line 215: nit:
```suggestion
        spark.conf.set("spark.rapids.sql.enabled", not enabled)
```

razajafri: Thanks

razajafri:
build

razajafri:
build

razajafri:
build

razajafri:
@gerashegalov @revans2 do you have any more comments on this PR?

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -350,61 +366,84 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
-    val buffers = new ListBuffer[ParquetCachedBatch]
-    if (splitIndices.nonEmpty) {
-      val splitVectors = new ListBuffer[Array[ColumnVector]]
-      try {
-        for (index <- 0 until gpuCB.numCols()) {
-          splitVectors +=
-              gpuCB.column(index).asInstanceOf[GpuColumnVector].getBase.split(splitIndices: _*)
+    // We want to map the name of the cols we are casting so we can cast them back when reading
+    // back the cache
+    val castMap = new scala.collection.mutable.HashMap[String, DType]()
```
revans2, line 377: This data structure fundamentally can never work for nested types.  You are storing a single DType per column.  What happens if I have a struct with multiple nested Decimal values in it?  What happens if I also have a mixture of LongType and IntTypes in there too?

revans2 (changes requested):

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -26,20 +26,21 @@ import scala.collection.mutable.{ArrayBuffer, ListBuffer}
 
 import ai.rapids.cudf._
 import ai.rapids.cudf.ParquetWriterOptions.StatisticsFrequency
-import com.nvidia.spark.rapids._
+import com.nvidia.spark.rapids.{GpuColumnVector, _}
```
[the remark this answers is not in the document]
gerashegalov: it might be a bug in the IDEA's scala plugin. I've seen the auto import do this

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
razajafri: Ugh, I meant to use Map[String, ArrayBuffer[DType]]. I have simplified the logic to not need a map. PTAL

revans2: Why is this needed at all??  You have the schema of the output data.  You know how you would transform it and it should be consistent. So you should just need to walk the output schema along with the data structure returned from the parquet reader and bit cast it when you see something different.  Or throw an exception if the types are so different that they cannot be bit cast.

razajafri: Yeah, that's exactly what I have done in the latest commit. 

I still  need to work on renaming all the child columns in a nested structure, otherwise the parquet reader barfs

integration_tests/src/main/python/cache_test.py
```diff
@@ -20,22 +20,24 @@
 from spark_session import with_cpu_session, with_gpu_session
 from join_test import create_df
 from marks import incompat, allow_non_gpu, ignore_order
-from parquet_write_test import parquet_decimal_struct_gen
 
-enableVectorizedConf = [{"spark.sql.inMemoryColumnarStorage.enableVectorizedReader" : "true"},
-                        {"spark.sql.inMemoryColumnarStorage.enableVectorizedReader" : "false"}]
+conf = [{"spark.sql.inMemoryColumnarStorage.enableVectorizedReader": "true",
```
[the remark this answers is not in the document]
razajafri: I have addressed your concern I think

razajafri:
@revans2 I have reworked quite a bit of this. PTAL

The changes will help with the #2382

razajafri:
build

integration_tests/src/main/python/cache_test.py
```diff
@@ -51,25 +55,24 @@ def test_passing_gpuExpr_as_Expr(enableVectorizedConf):
...
+@pytest.mark.parametrize('enableVectorizedConf', enableVectorizedConfs, ids=idfn)
 @ignore_order
 def test_cache_join(data_gen, join_type, enableVectorizedConf):
     def do_join(spark):
         left, right = create_df(spark, data_gen, 500, 500)
         cached = left.join(right, left.a == right.r_a, join_type).cache()
         cached.count() # populates cache
         return cached
-
+    enableVectorizedConf.update(allowNegativeScaleDec)
```
revans2, line 70: This will modify the values in the global `enableVectorizedConfs`.

```
conf = enable_vectorized_conf.copy()
conf.update(allow_negative_scale_of_decimal_conf)
assert_gpu_and_cpu_are_equal_collect(do_join, conf = conf)
```

And this goes for all of the places in this file that are using update.  For me personally I think it is a lot cleaner

integration_tests/src/main/python/cache_test.py
```diff
@@ -20,12 +20,16 @@
 from spark_session import with_cpu_session, with_gpu_session
 from join_test import create_df
 from marks import incompat, allow_non_gpu, ignore_order
-from parquet_write_test import parquet_decimal_struct_gen
 
-enableVectorizedConf = [{"spark.sql.inMemoryColumnarStorage.enableVectorizedReader" : "true"},
-                        {"spark.sql.inMemoryColumnarStorage.enableVectorizedReader" : "false"}]
+enableVectorizedConfs = [{"spark.sql.inMemoryColumnarStorage.enableVectorizedReader": "true"},
```
revans2, line 24: nit: python uses snake_case not camelCase

integration_tests/src/main/python/cache_test.py
```diff
@@ -20,12 +20,16 @@
...
+                        {"spark.sql.inMemoryColumnarStorage.enableVectorizedReader": "false"}]
+allowNegativeScaleDec = {"spark.sql.legacy.allowNegativeScaleOfDecimal": "true"}
```
revans2, line 26: This already exists.

https://github.com/NVIDIA/spark-rapids/blob/231f8e30bc4cd117a891888147e637985b3da773/integration_tests/src/main/python/data_gen.py#L876

datagen.py is probably not the most ideal place for it, but it works.

integration_tests/src/main/python/parquet_test.py
```diff
@@ -199,7 +199,7 @@ def test_ts_read_fails_datetime_legacy(gen, spark_tmp_path, ts_write, ts_rebase,
             lambda spark : readParquetCatchException(spark, data_path),
             conf=all_confs)
 
-@pytest.mark.parametrize('parquet_gens', [decimal_gens,
+@pytest.mark.parametrize('parquet_gens', [[byte_gen, short_gen], decimal_gens,
```
revans2, line 202: Why add `byte_gen` and `short_gen` as their own specific test cast to a test that is only about decimals?

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnUtil.scala
```diff
@@ -0,0 +1,108 @@
...
+
+package com.nvidia.spark.rapids
+
+import scala.collection.mutable.{ArrayBuffer, ArrayBuilder}
+
+import ai.rapids.cudf.{ColumnVector, ColumnView, DType}
+
+import org.apache.spark.sql.types.{ArrayType, DataType, StructField, StructType}
+
+object ColumnUtil extends Arm {
```
revans2, line 25: Can we get some java docs to explain what `ColumnUtil` is intended to be used for?  Any Util classes tend to turn into a dumping ground for things that don't fit anywhere else, and it would be nice to make it very clear what should and should not go here.  Because this only holds a single method it really feels like it should be specific to this method for now, or slightly more generic. Perhaps `NestedColumnBitCast`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -659,8 +659,27 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
             val readField = readDataSchema(writeAt)
             if (areNamesEquiv(clippedGroups, readAt, readField.name, isSchemaCaseSensitive)) {
               val origCol = table.getColumn(readAt)
-              val col = if (typeCastingNeeded && precisionList.nonEmpty) {
-                convertDecimal64ToDecimal32Wrapper(origCol, precisionList.dequeue())
+              val col: ColumnVector = if (typeCastingNeeded && precisionList.nonEmpty) {
+                ColumnUtil.ifTrueThenDeepConvertTypeAtoTypeB(origCol, readField.dataType,
```
revans2, line 663: OK Two things.

1.  Why not just have a single API that is like `nestedBitCastAsNeeded(col: ColumnView, sparkType: DataType)` which would walk through the col and match up the structure with `sparkType` as it went and do a nested bit cast when needed.  Then you can manipulate `sparkType` ahead of time to look exactly how you want it to.  The fact that there is a `precisionList` and this code consumes it feels both kind of brittle and confusing to follow.
2. Why do we need this in the GpuParquetScan code at all?  This appears to be specific to the cache serializer. It has to be because we do *NOT* want this type of thing to happen anytime we are reading a normal parquet file. It is not necessarily something that needs to be fixed here. It just feels like ti would be much cleaner if we could read the parquet file normally and then have a stage that did the bit casts after it was already read in, and the bit casting happened in the serializer without leaking into GpuParquetScan itself.

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -217,9 +219,15 @@ private object ParquetCachedBatch {
   def apply(parquetBuff: ParquetBufferConsumer): ParquetCachedBatch = {
     new ParquetCachedBatch(parquetBuff.numRows, parquetBuff.getBuffer)
   }
+
+  def apply(numRows: Int, buff: Array[Byte]): ParquetCachedBatch = {
```
revans2, line 223: If `ParquetCachedBatch` is a case class why do we need this? It should automatically be generated for us.

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -293,6 +301,7 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
       case ArrayType(elementType, _) => isTypeSupportedByParquet(elementType)
       case MapType(keyType, valueType, _) => isTypeSupportedByParquet(keyType) &&
           isTypeSupportedByParquet(valueType)
+      case d: DecimalType if d.scale < 0 => false
```
revans2, line 304: Could we add some scaladocs to `isTypeSupportedByParquet` to clarify that this is really types that parquet supports without any type of modification.

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -350,61 +360,82 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
-        // T01= {splitCol1(1), splitCol2(1),...,splitColn(1)}
-        // ...
-        // T0m= {splitCol1(m), splitCol2(m),...,splitColn(m)}
-        def makeTableForIndex(i: Int): Table = {
-          val columns = splitVectors.indices.map(j => splitVectors(j)(i))
-          new Table(columns: _*)
+    val columns = for (i <- 0 until oldGpuCB.numCols()) yield {
+      val gpuVector = oldGpuCB.column(i).asInstanceOf[GpuColumnVector]
+      var dataType = origSchema(i).dataType
+      val v = ColumnUtil.ifTrueThenDeepConvertTypeAtoTypeB(gpuVector.getBase,
```
revans2, line 374: Just like in the other file I think it would be a lot cleaner to modify the Spark schema to match what you want and then have a high level API that knows how to make them match with bit casting, instead of trying to make the API overly generic.

integration_tests/src/main/python/parquet_test.py
razajafri: The test array is missing DecimalGen, it should be [byte_gen, short_gen, DecimalGen(7, 3)]. There was a bug in the old code which would pop the precisionList even if the element wasn't a Decimal

razajafri:
build

razajafri:
@revans2 have I addressed all your concerns?

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
revans2: My idea what that you take the schema `StructType("a": DecimalType(10, -2), "b": IntegerType)` and manipulate it how you want. `StructType("a", IntegerType, "b" IntegerType)` then pass that into `nestedBitCastAsNeeded`. It would see that the first column "a" is a Decimal Type and know that it needs to be an IntegerType so it would do the bit cast.  The when going back after you read the data from parquet you just pass in the original schema into `nestedBitCastAsNeeded` and it would see that "a" is an integer type, but you want a `DecimalType(10, -2)` and do the bit cast.

razajafri:
build

revans2 (approved):
I am still not thrilled about the `ifTrueThenDeepConvertTypeAtoTypeB` API, and I think we can improve it, but I think it is good enough for now.

razajafri:
build

revans2 (approved):

razajafri:
I am having trouble testing my code locally. I apologize for the repeated CI failure

razajafri:
build

integration_tests/src/main/python/cache_test.py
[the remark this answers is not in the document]
gerashegalov: My understanding you addressed `enableVectorizedConf` but not granular parameters. It's minor but to avoid some repetitiveness Instead of whole lengthy conf parameters we could have parameters to the tune :

```python
@pytest.mark.parametrize('enable_vectorized_conf', ['VECTORIZED_ON', 'VECTORIZED_OFF'], ...) 
def test_passing_gpuExpr_as_Expr(enable_vectorized_conf):
      ...,
      conf = { 'spark.sql.inMemoryColumnarStorage.enableVectorizedReader': enable_vectorized_conf == 'VECTORIZED_ON' }
   )
```

gerashegalov (approved):
I don't fully understand the serializer code yet. I defer to @revans2 and other reviews. Coding style nits are minor.

razajafri:
@revans2 @gerashegalov 
Thanks for reviewing, this PR is now ready. I have replaced `binary_op_df` data_gen with `unary_op_df` and that reduced the data size and made the test run better

razajafri:
build

gerashegalov (approved):

PR #2676, opened 2021-06-10
Profiling tool: Print Job Information in compare mode

nartal1:
This is a simple PR to print job information in compare mode.
There is already a test in AppicationInfoSuite which calls `jobtoStagesSQL`  function and validates the results in Collect mode. 
So didn't add any tests since we call the same the function.


nartal1:
build

pxLi:
build

tgravescs (approved):

PR #2697, opened 2021-06-10
fix GDS spill bug when copying from the batch write buffer

rongou:
When a spilled buffer is pending and still in GPU memory, sometimes we want to copy part of it to the UCX bounce buffer. We shouldn't copy the whole length of the buffer, and should only copy the specified length.


rongou:
build

Issue #2674, opened 2021-06-10
[FEA] add file size check for pre-merge CI run


pxLi:
Do not allow large files (like > 50MiB) get merged to upstream branch.

Add check in pre-merge pipeline.

```bash
if (( $(git ls-tree --full-tree -r -l HEAD | sort -k 4 -n -r | head -1 | awk '{print $4}') > 52428800 )); then exit 1; fi
```

revans2:
I am more than a little upset that we even have a 40 MB file in our repo. It seems ridiculous and we have 2 of them. I guess 50MB is an okay cutoff, but I really would have preferred to see something closer to 1.5MiB.  Could we explicitly allow the existing files that are over 1.5MiB and then set a limit of 1.5 MiB for anything new?

pxLi:
sure, I will add the check in pre-merge CI using pre-commit tool, and make the default cutoff as 1.5MiB

Issue #2644, opened 2021-06-08
[BUG] test event logs should be compressed


rongou:
**Describe the bug**
We've merged a couple test event log files that are pretty big:
```console
ls -lh tools/src/test/resources/spark-events-profiling
total 103M
-rw-rw-r-- 1 rou rou 343K Jun  8 10:52 eventlog_minimal_events
-rw-rw-r-- 1 rou rou  64M Jun  8 10:52 executors_removed_eventlog
-rw-rw-r-- 1 rou rou 1.7M Jun  8 10:52 malformed_json_eventlog
-rw-rw-r-- 1 rou rou 2.2M Jun  8 10:52 rapids_join_eventlog
-rw-rw-r-- 1 rou rou 2.2M Jun  8 10:52 rapids_join_eventlog2
-rw-rw-r-- 1 rou rou 110K Jun  8 10:52 rp_nosql_eventlog
-rw-rw-r-- 1 rou rou 1.7M Jun  8 10:52 rp_sql_eventlog
-rw-rw-r-- 1 rou rou  32M Jun  8 10:52 task_job_failure_eventlog
```
This causes git to complain with a warning:
```console
remote: warning: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.
remote: warning: See http://git.io/iEPt8g for more information.
remote: warning: File tools/src/test/resources/spark-events-profiling/executors_removed_eventlog is 63.56 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
```

**Steps/Code to reproduce bug**
The files are already merged.

**Expected behavior**
We should compress these log files and change the tests to decompress them on the fly.

**Additional context**
We may want to change the git history so these large files don't show up in every `git clone`.

tgravescs:
we removed them, this required force push.  @nartal1 please make a pr to add those tests back in as I removed the files and commented out the tests.

Issue #2420, opened 2021-05-14
[FEA] Enable applicable standard pre-commit hooks


gerashegalov:
Now that we have pre-commit config in place #2403 we can think about some standardization. 
 
https://github.com/pre-commit/pre-commit-hooks contains a number of useful pre-commit hooks that can help this repo enforce standardization across different tools and developers. Such
- [Trailing white space fixer](https://github.com/pre-commit/pre-commit-hooks/blob/master/pre_commit_hooks/trailing_whitespace_fixer.py) 
- [Inconsistent line breaks](https://github.com/pre-commit/pre-commit-hooks/blob/master/pre_commit_hooks/mixed_line_ending.py). E.g  `dist/dependency-reduced-pom.xml` and some csv files use CRLF whereas all the rest just LF
-  TBD ...

revans2:
OK as an aside `dependency-reduced-pom.xml` should never be checked in. It is generated by the shade plugin.

PR #2699, opened 2021-06-10
Add a pre-commit hook to reject large files

gerashegalov:
Prevents large files as in #2644. 

Contributes to #2420, #2674
    
### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,6 +25,6 @@ repos:
     rev: v4.0.1
     hooks:
       - id: check-added-large-files
-        name: Reject files over 50M
-        args: ['--maxkb=51200', '--enforce-all']
+        name: Check for file over 1.5MiB
+        args: ['--maxkb=1500', '--enforce-all']
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,4 +21,8 @@ repos:
         language: script
         pass_filenames: true
         verbose: true
+  - repo: https://github.com/pre-commit/pre-commit-hooks.git
+    rev: v4.0.1
+    hooks:
+      - id: check-added-large-files
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,5 +25,6 @@ repos:
     rev: v4.0.1
     hooks:
       - id: check-added-large-files
+        name: Reject files over 50M
         args: ['--maxkb=51200', '--enforce-all']
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,4 +25,5 @@ repos:
     rev: v4.0.1
     hooks:
       - id: check-added-large-files
+        args: ['--maxkb=51200', '--enforce-all']
 


gerashegalov:
Developer's use case:
```
$ fallocate -l 51M tools/src/test/resources/bigfile.txt
$ git add tools/src/test/resources/bigfile.txt
$ git commit -m test

Reject files over 50M....................................................Failed
- hook id: check-added-large-files
- exit code: 1

tools/src/test/resources/bigfile.txt (52224 KB) exceeds 51200 KB.
```

abellina:
sounds good, I'd argue for a limit that allows our current repo to pass CI without any extra sha restrictions (something like 2.5MB).

gerashegalov:
> sounds good, I'd argue for a limit that allows our current repo to pass CI without any extra sha restrictions (something like 2.5MB).

@abellina unfortunately, 2.5MiB won't do it per @revans2's comment  https://github.com/NVIDIA/spark-rapids/issues/2674#issuecomment-858772026

```bash
$ git ls-files | xargs ls -alhS | head -3
-rw-rw-r-- 1 gshegalov gshegalov   44M Jun  9 11:00 tools/src/test/resources/spark-events-qualification/nds_q86_fail_test
-rw-rw-r-- 1 gshegalov gshegalov   44M Jun  7 17:09 tools/src/test/resources/spark-events-qualification/nds_q86_test
-rw-rw-r-- 1 gshegalov gshegalov  4.6M Jun  7 17:09 tools/src/test/resources/spark-events-qualification/dsAndDf_eventlog
```

revans2:
> sounds good, I'd argue for a limit that allows our current repo to pass CI without any extra sha restrictions (something like 2.5MB).

What files are that big?  

```
$ find . -type f | xargs ls -lSh | head -10
-rw-rw-r-- 1 user user   44M Jun  9 16:10 ./tools/src/test/resources/spark-events-qualification/nds_q86_fail_test
-rw-rw-r-- 1 user user   44M Jun  9 16:10 ./tools/src/test/resources/spark-events-qualification/nds_q86_test
-rw-rw-r-- 1 user user  4.6M Jun  9 16:10 ./tools/src/test/resources/spark-events-qualification/dsAndDf_eventlog
-rw-rw-r-- 1 user user  2.2M Jun  9 16:10 ./tools/src/test/resources/spark-events-profiling/rapids_join_eventlog
-rw-rw-r-- 1 user user  2.2M Jun 10 13:11 ./tools/src/test/resources/spark-events-profiling/rapids_join_eventlog2
-rw-rw-r-- 1 user user  1.7M Jun  9 16:10 ./tools/src/test/resources/spark-events-profiling/malformed_json_eventlog
-rw-rw-r-- 1 user user  1.7M Jun  9 16:10 ./tools/src/test/resources/spark-events-profiling/rp_sql_eventlog
-rw-rw-r-- 1 user user  1.6M Jun 10 13:11 ./tools/src/test/resources/spark-events-qualification/join_missing_sql_end
-rw-r--r-- 1 user user  1.1M Aug 21  2020 ./integration_tests/src/test/resources/Acquisition_2007Q3.txt
-rw-rw-r-- 1 user user 1003K Jun 10 13:11 ./tools/src/test/resources/spark-events-profiling/tasks_executors_fail_compressed_eventlog.zstd
```

It is all of the files that were just added in as a part of the tools tests.  Everything before that was small, and I would argue we should not have allowed the tools files in. We should have flagged it early on and someone should have looked at it and thought do we really need a 44M MiB file checked into the repo.

revans2:
> those files can be compressed as well or we can decide to move them out

That does not fix the problem.  Once they are in the repo unless we do surgery to it, like we just did, they will be downloaded every time.  The point of these are not to fix what was already done. The point of this is to stop it from ever happening again in the future and making the problem worse

revans2 (approved):

abellina:
Never mind (re 2.5MB). I was assuming the output from an earlier comment was all inclusive.

gerashegalov:
build

PR #2700, opened 2021-06-10
Set numSlices and use parallelize to build dataframe for partition-se…

abellina:
…nsitive tests

Set `numSlices` and use the `parallelize` API, to create a dataframe that has a static number of partitions.

Re: #2477

### DIFF
--- a/integration_tests/src/main/python/data_gen.py
+++ b/integration_tests/src/main/python/data_gen.py
@@ -617,7 +617,7 @@ def skip_if_not_utc():
     if (not is_tz_utc()):
         skip_unless_precommit_tests('The java system time zone is not set to UTC')
 
-def gen_df(spark, data_gen, length=2048, seed=0, numSlices=None):
+def gen_df(spark, data_gen, length=2048, seed=0, num_slices=None):
     """Generate a spark dataframe from the given data generators."""
     if isinstance(data_gen, list):
         src = StructGen(data_gen, nullable=False)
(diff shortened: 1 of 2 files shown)


abellina:
With `numSlices=123`:
Screenshot from 2021-06-10 13-04-16

With `numSlices=12`:
Screenshot from 2021-06-10 13-05-32

integration_tests/src/main/python/sort_test.py
```diff
@@ -107,8 +107,12 @@ def test_single_nested_orderby_with_limit(data_gen, order):
 def test_single_sort_in_part(data_gen, order):
     # This outputs the source data frame each time to debug intermittent test
     # failures as documented here: https://github.com/NVIDIA/spark-rapids/issues/2477
+    # Update: we are setting the number of slices with `numSlices` as the error appears to be
```
revans2, line 110: So will there be another PR to update these comments and clean them up if this proves to be correct?

abellina: Was trying to keep some context as to why we set `numSlices`, but yeah thought similar. The test fails sporadically so I'll forget to remove this comment. Maybe just point to the issue as a test comment?

abellina: (which we already do, but tidy up)

revans2: Ya, just say we are doing 12 slices because of ...

abellina: ok edited!

integration_tests/src/main/python/data_gen.py
```diff
@@ -616,7 +617,7 @@ def skip_if_not_utc():
     if (not is_tz_utc()):
         skip_unless_precommit_tests('The java system time zone is not set to UTC')
 
-def gen_df(spark, data_gen, length=2048, seed=0):
+def gen_df(spark, data_gen, length=2048, seed=0, numSlices=None):
```
revans2, line 620: Sorry one more thing should this be `num_slices` to match the python style?

revans2 (approved):

abellina:
build

PR #2702, opened 2021-06-10
Add in support for arrays in BroadcastNestedLoopJoinExec and CartesianProductExec

revans2:
When running the profiling tool with some large applications I noticed that it would do a few big joins using either `CartesianProductExec` or `BroadcastNestedLoopJoinExec` with an array and array_contains as the conditions.  So I enabled it in the check wrote some tests, and this sped it up massively.


revans2:
build

gerashegalov (approved):
LGTM

Issue #2562, opened 2021-06-02
[FEA] Implement KnownNotNull on the GPU


sameerz:
**Is your feature request related to a problem? Please describe.**
Implement KnownNotNull on GPU

**Describe the solution you'd like**
When KnownNotNull is part of the plan, the plan stays on the GPU

**Describe alternatives you've considered**
N/A

**Additional context**

firestarman:
@sameerz  May I know the case/sql for this `KnownNotNull ` ? Since I need a test case for this expression.

firestarman:
I tried several cases and got this expression in logical plan, but it will be removed after optimization.
I will check the the native UDF.

PR #2707, opened 2021-06-11
Support `KnownNotNull` running on GPU

firestarman:
This PR is to support `KnownNotNull` running on GPU.

- [x] Add tests

closes https://github.com/NVIDIA/spark-rapids/issues/2562


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1340,6 +1340,15 @@ object GpuOverrides {
         override def convertToGpu(child: Expression): GpuExpression =
           GpuKnownFloatingPointNormalized(child)
       }),
+    expr[KnownNotNull](
+      "Tag an expression as known to not be null",
+      ExprChecks.unaryProjectNotLambdaInputMatchesOutput(
+        TypeSig.commonCudfTypes + TypeSig.DECIMAL,
+        TypeSig.commonCudfTypes + TypeSig.DECIMAL),
```
revans2, line 1347: Technically the second param should be `TypeSig.all` because Spark supports all types with this, although in reality it is only ever going to use it on byte, short, int, long, float, and double.  All of the other values are not primitive types and as such can be nullable in a UDF.

firestarman: Updated, however does it mean the `KnownNotNull` also supports `NullType`, which I thought should not be allowed for this expression?

firestarman:
build

firestarman:
build

firestarman:
build

revans2 (approved):

PR #2709, opened 2021-06-12
WIP: batch read small buffers spilled through GDS

rongou:
It's relatively inefficient to read and write small spill buffers one by one through GDS. #2295 batches small buffers when writing, this PR is the counter part for reading. When unspill is enabled and a spilled buffer is requested, the whole batched buffer file (default to 8MB) is read back, and each buffer is then unspilled in a background thread.

The caveat is the batched buffers from the same file may not be needed at the same time, so some of them may be spilled and read back repeatedly. An improvement would be to batch buffers with similar a life cycle together, but it's hard to do this with Spark.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
@@ -253,7 +253,7 @@ class RapidsGdsStore(
         } else if (RapidsBufferCatalog.shouldUnspill) {
           withResource(DeviceMemoryBuffer.allocate(batchWriteBufferSize)) { buffer =>
             CuFile.readFileToDeviceMemory(buffer.getAddress, buffer.getLength, path, 0)
-            logDebug(s"Created device buffer for $path 0:$size via GDS")
+            logDebug(s"Created device buffer for $path 0:$batchWriteBufferSize via GDS")
             spilledBuffers.get(path).foreach { id =>
               if (id != this.id) {
                 withResource(catalog.acquireBuffer(id)) {
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
@@ -23,7 +23,7 @@ import java.util.function.BiFunction
 import scala.collection.mutable.ArrayBuffer
 
 import ai.rapids.cudf._
-import com.nvidia.spark.rapids.StorageTier.StorageTier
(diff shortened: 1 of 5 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -230,26 +236,62 @@ class RapidsGdsStore(
...
+                logDebug(s"Ignoring buffer $id")
+            }
+          }
+        }
+      }
+
       override def materializeMemoryBuffer: MemoryBuffer = this.synchronized {
-        closeOnExcept(DeviceMemoryBuffer.allocate(size)) { buffer =>
-          if (isPending) {
+        if (isPending) {
```
abellina, line 271: Could there be some comments in `materializeMemoryBuffer` for each of the three cases?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -230,26 +236,62 @@ class RapidsGdsStore(
...
             logDebug(s"Created device buffer $size from batch write buffer")
-          } else {
+            buffer
+          }
+        } else if (RapidsBufferCatalog.shouldUnspill) {
+          closeOnExcept(DeviceMemoryBuffer.allocate(batchWriteBufferSize)) { buffer =>
+            CuFile.readFileToDeviceMemory(buffer.getAddress, buffer.getLength, path, 0)
+            logDebug(s"Created device buffer for $path 0:$batchWriteBufferSize via GDS")
+            closeOnExcept(DeviceMemoryBuffer.allocate(size)) { d =>
+              d.copyFromMemoryBuffer(0, buffer, fileOffset, size, Cuda.DEFAULT_STREAM)
```
abellina, line 283: so there's a copy here, but there's also another one in the thread. It's not clear what this copy is, and what the other copy is within the thread. Just by looking at this diff. Perhaps I am missing something.

abellina:
Could there be a short description of this PR? I do realize it is in WIP, so I can wait if it's not ready.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
rongou: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
rongou: This copy is returned right away, while the whole batched buffer is unspilled in the background thread.

rongou:
This causes too many OOM errors since the buffers in a file are not likely needed together.

PR #2721, opened 2021-06-15
Check-list for corner cases in testing.

mythrocks:
This is a cursory pass at documenting corner cases for integration tests, for newly added operations. Currently a work in progress.


integration_tests/README.md
```diff
@@ -303,3 +303,66 @@ The marks you care about are all in marks.py
...
+E.g. `COUNT()` on an empty input column yields `0`. `SUM()` yields `0` for the appropriate numeric type.
+
+### 6. Special floating point values
+Apart from `null` values, `FloatType` and `DoubleType` input columns must also include the following special values:
+   * +/- Zero
+   * +/- Infinity
+   * +/- NaN
+
+### 7. Special values in timestamp columns
+Ensure date/timestamp columns include dates before [epoch](https://en.wikipedia.org/wiki/Epoch_(computing)).
```
sameerz, line 368:
```suggestion
Ensure date/timestamp columns include dates before the [epoch](https://en.wikipedia.org/wiki/Epoch_(computing)).
```

integration_tests/README.md
```diff
@@ -303,3 +303,66 @@ The marks you care about are all in marks.py
...
+
+## Guidelines for Testing
+
+When support for a new operator is added to `rapids-plugin-4-spark`, it is recommended that
+the following conditions be covered in its corresponding integration tests:
+
+### 1. Cover all supported data types
+Ensure that tests cover all data types supported by the added operation. An exhaustive list of data types supported in 
+Spark is available [here](https://spark.apache.org/docs/latest/sql-ref-datatypes.html). These include:
+   * Numeric Types (`ByteType`, `ShortType`, `IntegerType`, etc.)
```
abellina, line 315: it would be good to call out decimal on its own. I am honestly not an expert at decimals, but it would be good to figure out if there are specific gotchas with it.

integration_tests/README.md
```diff
@@ -303,3 +303,66 @@ The marks you care about are all in marks.py
...
+```
+[
+    [0,1,2,3],
+    [], <------- Empty, non-null row.
+    [4,5,6,7],
+    ...
+]
+```
+
+### 5. Degenerate cases with "empty" inputs
```
abellina, line 356: Sometimes expressions can take a literal (like`sum(1)`) or a column (like `sum(x)`), this one covers 0, but I'd add to it other values => basically scalars of all the types.

integration_tests/README.md
razajafri: Can we just list out all the types supported by us instead of referring to the Spark doc? That will make this a self-contained doc and we can add types as we support more

integration_tests/README.md
```diff
@@ -303,3 +303,66 @@ The marks you care about are all in marks.py
...
+   * Numeric Types (`ByteType`, `ShortType`, `IntegerType`, etc.)
+   * Strings (`StringType`, `VarcharType`)
+   * Booleans (`BooleanType`)
+   * Chrono Types (`TimestampType`, `DateType`, `Interval`)
+   * Complex Types (`ArrayType`, `StructType`, `MapType`)
+
+`data_gen.py` provides `DataGen` classes that help generate test data in integration tests.
+
+### 2. Nested Data types
+Complex data types (`ArrayType`, `StructType`, `MapType`) warrant extensive testing for various combinations of nesting.
```
razajafri, line 324: This is a good point. We need a way for us to create nesting in our integration tests quickly instead of typing it out but at the same time be careful the nesting isn't too deep as it slows down the tests.

razajafri: I have created a follow-on issue for it. https://github.com/NVIDIA/spark-rapids/issues/2759

integration_tests/README.md
```diff
@@ -303,3 +303,66 @@ The marks you care about are all in marks.py
...
+
+E.g. `COUNT()` on an empty input column yields `0`. `SUM()` yields `0` for the appropriate numeric type.
+
+### 6. Special floating point values
+Apart from `null` values, `FloatType` and `DoubleType` input columns must also include the following special values:
+   * +/- Zero
+   * +/- Infinity
+   * +/- NaN
+
+### 7. Special values in timestamp columns
```
razajafri, line 367: Python uses proleptic gregorian date which assumes Gregorian calendar always existed in past and will always exist in future. When performing an operation that could potentially cause an overflow we should limit the data_gen to stay within the legal limits of python date

abellina:
> Can we just list out all the types supported by us instead of referring to the Spark doc? That will make this a self-contained doc and we can add types as we support more

If it's a checklist we should probably check Spark (if there's a new type we may learn that way). I doubt that will happen often, but it seems appropriate to point developers to the Spark docs as that's where the source of truth in this scenario.

razajafri:
> > Can we just list out all the types supported by us instead of referring to the Spark doc? That will make this a self-contained doc and we can add types as we support more
> 
> If it's a checklist we should probably check Spark (if there's a new type we may learn that way). I doubt that will happen often, but it seems appropriate to point developers to the Spark docs as that's where the source of truth in this scenario.

Makes sense to include the spark doc link for reference. My comment is toward not spelling out each and every type that we support in this doc and instead saying `etc` which you already pointed out for Decimals, I am going an extra step and saying just type them all out.

integration_tests/README.md
mythrocks: I've added a note about using `DateGen` to generate Proleptic Gregorian date values for tests.

integration_tests/README.md
mythrocks: Made mention of `scalar` values for tests.

integration_tests/README.md
mythrocks: Listed out all types, instead of `etc.`.

sameerz (approved):
LGTM, @tgravescs @jlowe @revans2 any feedback?

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
+   * `Array<primitive_type>`
+   * `Array<Array<primitive_type>>`
+   * `Array<Struct<primitive_type>>`
+   * `Struct<Array<primitive_type>>`
+   * `Array<Struct<Array<primitive_type>>>`
+   * `Struct<Array<Struct<primitive_type>>>`
+
+The `ArrayGen` and `StructGen` classes in `data_gen.py` can be configured to support arbitrary nesting.
+
+### 3. Scalar values
```
revans2, line 351: Spark uses the term literal values where as cudf uses scalar.  I think we should include both so it is clear.

### 3. Scalar or Literal values

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
+Operators and expressions that support scalar operands need to be tested with scalar inputs, of all 
+supported types from 1 and 2, above. 
+For instance, `SUM()` supports numeric columns (e.g. `SUM(a + b)`), or scalars (e.g. `SUM(20)`).
+Similarly, `COUNT()` supports the following:
+   * Columns: E.g. `COUNT(a)` to count non-null rows for column `a`
+   * Scalars: E.g. `COUNT(1)` to count all rows (including nulls)
+   * `*`: E.g. `COUNT(*)`, functionally equivalent to `COUNT(1)`
+It is advised that tests be added for all applicable scalar types, for an operator.
+
```
revans2, line 360: Can we add a comment about static folding in Spark?  Something like 


For many operations if all of the inputs are scalar/literal values Apache Spark will evaluate that expression and replace it with a new literal value in the logical planning/optimization stage. So if you type in `5 + 10` Spark will replace it with a literal `15` before the plugin sees it.  This is great for performance, but typically does not really add more test coverage.

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
+### 4. Null rows
```
revans2, line 361: nit: I would just call this Null values.

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
+Ensure that the test data accommodates null values for input columns.
+Null values in input columns are a frequent source of bugs in `rapids-plugin-4-spark`, because of mismatches in null-handling
```
revans2, line 363: nit: we should not call it `rapids-plugin-4-spark` We should use the proper name the Rapids Accelerator for Apache Spark.

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
 For the most part you can ignore this file. It provides the underlying Spark session to operations that need it, but most tests should interact with
 it through `asserts.py`.
+
+## Guidelines for Testing
+
+When support for a new operator is added to `rapids-plugin-4-spark`, it is recommended that
+the following conditions be covered in its corresponding integration tests:
+
+### 1. Cover all supported data types
+Ensure that tests cover all data types supported by the added operation. An exhaustive list of data types supported in
```
jlowe, line 313: Should this explicitly mention negative testing of data types that Spark supports for the operation?  Many times developers will interpret "all supported data types" as testing only the types supported by the GPU operation.  We also want to ensure that any types not supported by the GPU properly fallback to the CPU when executing.

Bonus points for pointing the reader to helper classes or examples that can assist with data type negative testing which is a bit trickier in practice.

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
+
+### 4. Null rows
+Ensure that the test data accommodates null values for input columns.
+Null values in input columns are a frequent source of bugs in `rapids-plugin-4-spark`, because of mismatches in null-handling
+and semantics, between RAPIDS `libcudf` (on which `rapids-plugin-4-spark` relies heavily), and Apache Spark. 
+
+Apart from null rows in columns of primitive types, the following conditions must be covered for nested types:
+   * Null rows at the "top" level for `Array`/`Struct` columns.   E.g. `[ [1,2], [3], ∅, [4,5,6] ]`.
+   * Non-null rows containing null elements in the child column. E.g. `[ [1,2], [3,∅], ∅, [4,∅,6] ]`.
+
```
revans2, line 369: 1) It would be good to point out that the data gen inserts nulls by default, but this can be over-loaded
2) It would be good to point out that null literal values are also important to test.
3) It might be good to point out that for all aggregations, including group by, reductions, and windowing it is good to test where there all of the data being processed are nulls.

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
+     * `DateType`
+     * `Interval`
+   * Complex Types 
+     * `ArrayType`
+     * `StructType`
+     * `MapType`
+
+`data_gen.py` provides `DataGen` classes that help generate test data in integration tests.
+
+### 2. Nested Data types
```
jlowe, line 339: Nit: Inconsistent capitalization of `Data` relative to other headline occurrences in this doc
```suggestion
### 2. Nested data types
```

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
+supported types from 1 and 2, above. 
+For instance, `SUM()` supports numeric columns (e.g. `SUM(a + b)`), or scalars (e.g. `SUM(20)`).
+Similarly, `COUNT()` supports the following:
+   * Columns: E.g. `COUNT(a)` to count non-null rows for column `a`
+   * Scalars: E.g. `COUNT(1)` to count all rows (including nulls)
+   * `*`: E.g. `COUNT(*)`, functionally equivalent to `COUNT(1)`
+It is advised that tests be added for all applicable scalar types, for an operator.
+
+### 4. Null rows
+Ensure that the test data accommodates null values for input columns.
```
jlowe, line 362: Worth mentioning, but isn't everything in this section covered automatically by default in the data generators?  If so this doc should mention it, otherwise devs may create redundant tests not realizing these cases should already be covered with straightforward testing via the existing data generators.

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
+Null values in input columns are a frequent source of bugs in `rapids-plugin-4-spark`, because of mismatches in null-handling
+and semantics, between RAPIDS `libcudf` (on which `rapids-plugin-4-spark` relies heavily), and Apache Spark. 
+
+Apart from null rows in columns of primitive types, the following conditions must be covered for nested types:
```
jlowe, line 366: What's also interesting related to nulls is testing an input at some level (either at the top level or sometimes at some nesting level when using nested types) that's _entirely null_.  We've had some issues with that in the past.

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
+   * Null rows at the "top" level for `Array`/`Struct` columns.   E.g. `[ [1,2], [3], ∅, [4,5,6] ]`.
+   * Non-null rows containing null elements in the child column. E.g. `[ [1,2], [3,∅], ∅, [4,∅,6] ]`.
+
+### 5. Empty rows in `Array` columns
+Operations on `ArrayType` columns must be tested with input columns containing non-null *empty* rows.
```
jlowe, line 371: Definitely worth mentioning, but don't the data generators cover this by default?  If so then we don't want developers creating separate tests when the case will already be covered.

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
+]
+```
+
+### 6. Degenerate cases with "empty" inputs
+Ensure that operations are tested with "empty" input columns (i.e. containing zero rows.)
+
+E.g. `COUNT()` on an empty input column yields `0`. `SUM()` yields `0` for the appropriate numeric type.
+
+### 7. Special floating point values
+Apart from `null` values, `FloatType` and `DoubleType` input columns must also include the following special values:
```
jlowe, line 388: Should mention that the floating point data generators already cover this by default.

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
+   * +/- Zero
+   * +/- Infinity
+   * +/- NaN
+
```
revans2, line 392: I would include that there are multiple different binary representations of `NaN`, and that data_gen for floats and doubles will automatically insert these values into the data frames.

Do we want to give a warning about exact floating point results vs close-enough floating point results?

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
...
+### 8. Special values in timestamp columns
+Ensure date/timestamp columns include dates before the [epoch](https://en.wikipedia.org/wiki/Epoch_(computing)).
+It is advised that `DateGen` class from `data_gen.py` be used to generate valid (proleptic Gregorian calendar)
+dates when testing operators that work on dates.
```
revans2, line 396: We should give a warning about the python date/timestamp ranges and how Spark can produce errors for some extreme ranges.

integration_tests/README.md
```diff
@@ -303,3 +303,94 @@ The marks you care about are all in marks.py
 
 For the most part you can ignore this file. It provides the underlying Spark session to operations that need it, but most tests should interact with
 it through `asserts.py`.
+
+## Guidelines for Testing
+
+When support for a new operator is added to `rapids-plugin-4-spark`, it is recommended that
```
revans2, line 309: or it is extended to support more data types

integration_tests/README.md
mythrocks: Excellent point! I'll mention this.

I'm not familiar with the Spark vernacular. Shall I refer to this optimization as "Static Folding" in the documentation? I'm struggling to find references to link to. (Hive calls this "Constant Folding".)

revans2:
```
  /**
   * Returns true when an expression is a candidate for static evaluation before the query is
   * executed. A typical use case: [[org.apache.spark.sql.catalyst.optimizer.ConstantFolding]]
   *
   * The following conditions are used to determine suitability for constant folding:
   *  - A [[Coalesce]] is foldable if all of its children are foldable
   *  - A [[BinaryExpression]] is foldable if its both left and right child are foldable
   *  - A [[Not]], [[IsNull]], or [[IsNotNull]] is foldable if its child is foldable
   *  - A [[Literal]] is foldable
   *  - A [[Cast]] or [[UnaryMinus]] is foldable if its child is foldable
   */
  def foldable: Boolean = false
```

And the name of the optimization rule is `ConstantFolding` so my guess is that constant folding is probably the better term to use.

integration_tests/README.md
mythrocks: Hi, @revans2. I'm not sure I know what the errors might be, for date/timestamp ranges. :/

revans2: https://github.com/NVIDIA/spark-rapids/blob/5a379fb0cab5b3370f23a051da24918df9a8a157/integration_tests/src/main/python/data_gen.py#L517-L522

and

https://github.com/NVIDIA/spark-rapids/blob/5a379fb0cab5b3370f23a051da24918df9a8a157/integration_tests/src/main/python/data_gen.py#L457-L459

The timestamp issue is that when pyspark tries to convert the python DateTime object to the desired long value, but ends up overflowing it in a few cases.

integration_tests/README.md
mythrocks: Agreed. Fixed. Will upload fix shortly.

integration_tests/README.md
revans2: We do support negative tests.

`assert_gpu_fallback_collect`

integration_tests/README.md
mythrocks: Agreed. Made mention of the `DataGen` classes from `data_gen.py`.

integration_tests/README.md
mythrocks: Agreed. Added.

integration_tests/README.md
mythrocks: Agreed. Mentioned `ArrayGen` from `data_gen.py`.

integration_tests/README.md
mythrocks: Agreed. Mentioned `FloatGen` and `DoubleGen`.

integration_tests/README.md
mythrocks: Fixed. Will push the change shortly.

integration_tests/README.md
mythrocks: Agreed. I've added this description. I've tried to keep it short so as to keep the checklist manageable.

integration_tests/README.md
mythrocks: Agreed. Emended.

integration_tests/README.md
mythrocks: Agreed. Fixed.

integration_tests/README.md
mythrocks: Good one. Added.

integration_tests/README.md
```diff
@@ -303,3 +303,110 @@ The marks you care about are all in marks.py
...
+Ensure that the test data accommodates null values for input columns.
+Null values in input columns are a frequent source of bugs in `rapids-plugin-4-spark`, because of mismatches in null-handling
+and semantics, between RAPIDS `libcudf` (on which `rapids-plugin-4-spark` relies heavily), and Apache Spark. 
+
+Apart from null rows in columns of primitive types, the following conditions must be covered for nested types:
+
+   * Null rows at the "top" level for `Array`/`Struct` columns.   E.g. `[ [1,2], [3], ∅, [4,5,6] ]`.
+   * Non-null rows containing null elements in the child column. E.g. `[ [1,2], [3,∅], ∅, [4,∅,6] ]`.
+   * All null rows at a nested level. E.g. 
+     * All null list rows: `[ ∅, ∅, ∅, ∅ ]`
```
jlowe, line 371: This case applies to the non-nested types as well (e.g.: all null integer values, all null strings, etc.)

integration_tests/README.md
jlowe: Yes, I was thinking this doc should mention `assert_gpu_fallback_collect`.  That's probably sufficient for now, devs can go looking for existing use-cases of that method to find examples.

integration_tests/README.md
mythrocks: Agreed. Added a blurb on bit representations for special values, plus `@approximate_float` result verification.

integration_tests/README.md
mythrocks: Thanks. Fixed.

integration_tests/README.md
mythrocks: Agreed. I've mentioned the "all nulls" case in the general section, in response to @revans2's feedback as well.
I'll push this change shortly.

integration_tests/README.md
mythrocks: Agreed. I've added mention of `assert_gpu_fallback_collect()`.

jlowe (approved):

jlowe:
build

integration_tests/README.md
```diff
@@ -303,3 +303,131 @@ The marks you care about are all in marks.py
...
+   * +/- NaN
+
+Note that the special values for floating point numbers might have different bit representations for the same
+equivalent values. The [Java documentation for longBitsToDouble()](https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#longBitsToDouble-long-)
+describes this with examples. Operations should be tested with multiple bit-representations for these special values.
+
+The `FloatGen` and `DoubleGen` data generators in `integration_tests/src/main/python/data_gen.py` can be configured
+to generate the special float/double values mentioned above.
+
+Note that floating point values generated on the GPU might not match those from the CPU exactly. Any differences will
```
revans2, line 426: This makes it sound like the GPU for arbitrary operations will get the wrong floating point values.  I think we should be clear that for most operations this will not happen and point to https://github.com/NVIDIA/spark-rapids/blob/branch-21.08/docs/compatibility.md#floating-point for more information about when to use `@approximate_float` and when not to.

mythrocks: I should've known better than to editorialize. I've corrected the wording, and pointed to `compatibility.md`.

revans2 (approved):

revans2:
build

mythrocks:
Any objection to this being checked in?

integration_tests/README.md
```diff
@@ -149,7 +149,7 @@ Next you can start to run the tests.
 
 ```scala
 durations.run(new com.nvidia.spark.rapids.JoinsSuite)
-...
+// ...
```
abellina, line 152: is this on purpose?

mythrocks: @abellina: Good eye! Yes, this was intentional. The IDE flags the `...` as bad Scala, because it's in a `scala` block.

abellina (approved):
I had 1 nit and it's rather minor. This looks to be a great start and a useful document as is to refresh our memory when we test.

integration_tests/README.md
abellina: Ah ok. Perhaps just removing the line would be fine? It prints the line as `// ...`, according to github, and it doesn't seem like it adds much.

mythrocks: Gotcha. Will remove.

mythrocks: Fixed. Pushed new version.

abellina (approved):

mythrocks:
build

abellina:
build

revans2 (approved):

jlowe (approved):

mythrocks:
Merged. Thank you for the reviews, all!

Issue #2725, opened 2021-06-16
[BUG] Heartbeat from unknown executor when running with UCX shuffle in local mode


jlowe:
**Describe the bug**
Running with the RapidsShuffleManager while in local mode causes the following exceptions to be printed to the console every few seconds:
```
21/06/16 18:12:28 ERROR Inbox: Ignoring error
java.lang.IllegalStateException: Heartbeat from unknown executor BlockManagerId(driver, 10.28.9.126, 32931, None)
	at com.nvidia.spark.rapids.RapidsShuffleHeartbeatManager.executorHeartbeat(RapidsShuffleHeartbeatManager.scala:84)
	at com.nvidia.spark.rapids.RapidsDriverPlugin.receive(Plugin.scala:149)
	at org.apache.spark.internal.plugin.PluginEndpoint$$anonfun$receiveAndReply$1.applyOrElse(PluginEndpoint.scala:57)
	at org.apache.spark.rpc.netty.Inbox.$anonfun$process$1(Inbox.scala:103)
	at org.apache.spark.rpc.netty.Inbox.safelyCall(Inbox.scala:203)
	at org.apache.spark.rpc.netty.Inbox.process(Inbox.scala:100)
	at org.apache.spark.rpc.netty.MessageLoop.org$apache$spark$rpc$netty$MessageLoop$$receiveLoop(MessageLoop.scala:75)
	at org.apache.spark.rpc.netty.MessageLoop$$anon$1.run(MessageLoop.scala:41)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
21/06/16 18:12:28 ERROR RapidsShuffleHeartbeatEndpoint: Error during heartbeat
java.lang.IllegalStateException: Heartbeat from unknown executor BlockManagerId(driver, 10.28.9.126, 32931, None)
	at com.nvidia.spark.rapids.RapidsShuffleHeartbeatManager.executorHeartbeat(RapidsShuffleHeartbeatManager.scala:84)
	at com.nvidia.spark.rapids.RapidsDriverPlugin.receive(Plugin.scala:149)
[...]

PR #2726, opened 2021-06-16
Fix local mode starting RAPIDS shuffle heartbeats

abellina:
Fixes: #2725 

The issue is with local mode, where a single "executor" is created within the driver with as many cores as provided in the `local[*]` pattern. This is not a setup that should use UCX, therefore the heartbeats do not apply. The fix is to skip the registration of such an executor that resides in the driver.


jlowe (approved):

jlowe:
build

abellina:
Thanks for reporting @jlowe

PR #2728, opened 2021-06-16
Support string `repeat` SQL

### LINKED ISSUE #68: [FEA] support StringRepeat
We should support the `repeat` SQL function.

ttnghia:
This PR implements `repeat` SQL, and closes #68. 

Example:
```
strs = ['abc', null, '',  'xyz']
out  = repeat(strs, 3)
out is ['abcabcabc', null, '',  'xyzxyzxyz']

repeatTimes = [1, 2, 3, 4]
out = repeat(strs, repeatTimes)
out is ['abc', null, '',  'xyzxyzxyzxyz']

```


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2437,6 +2437,17 @@ object GpuOverrides {
               .withPsNote(TypeEnum.STRING, "only a single character is allowed"), TypeSig.STRING),
           ParamCheck("count", TypeSig.lit(TypeEnum.INT), TypeSig.INT))),
       (in, conf, p, r) => new SubstringIndexMeta(in, conf, p, r)),
+    expr[StringRepeat](
+      "StringRepeat operator",
+      ExprChecks.projectNotLambda(TypeSig.STRING, TypeSig.STRING,
+        Seq(ParamCheck("str", TypeSig.STRING, TypeSig.STRING),
```
ttnghia, line 2443: I'm not sure if this is the correct check.

The first input to this function can either be a string scalar or a strings column.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -484,6 +484,35 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+
+  def doColumnar(lhs: GpuScalar, rhs: GpuColumnVector): ColumnVector =
+    throw new UnsupportedOperationException(s"Cannot columnar evaluate expression: $this")
+
+  def doColumnar(strs: GpuColumnVector, repeatTimes: GpuScalar): ColumnVector = {
+    val repeatTimesVal = repeatTimes.getValue.asInstanceOf[Int]
+    strs.getBase.repeatStrings(repeatTimesVal)
+  }
+
+  def doColumnar(numRows: Int, str: GpuScalar, repeatTimes: GpuScalar): ColumnVector = {
```
ttnghia, line 509: I don't know how to call the overloaded version of `repeat_strings(string_scalar, int) => string_scalar`. Here I can only call the version that returns a strings column.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -484,6 +484,35 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+
+  override def right: Expression = repeatTimes
+
+  override def dataType: DataType = strs.dataType
+
+  override def inputTypes: Seq[AbstractDataType] =
+    Seq(TypeCollection(StringType, BinaryType), IntegerType)
+
+  def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector =
+    throw new UnsupportedOperationException(s"Cannot columnar evaluate expression: $this")
```
gerashegalov, line 499: are we deferring this for future tickets? Spark supports column for `repeatTimes`

```
scala> df.show

+---+---+
| _1| _2|
+---+---+
|  a|  0|
|  b|  1|
|  c|  2|
+---+---+

scala> df.createTempView("test")

scala> sql("select repeat(_1, _2) from test").show()

+--------------+
|repeat(_1, _2)|
+--------------+
|              |
|             b|
|            cc|
+--------------+
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -484,6 +484,35 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+
+  def doColumnar(lhs: GpuScalar, rhs: GpuColumnVector): ColumnVector =
+    throw new UnsupportedOperationException(s"Cannot columnar evaluate expression: $this")
+
+  def doColumnar(strs: GpuColumnVector, repeatTimes: GpuScalar): ColumnVector = {
+    val repeatTimesVal = repeatTimes.getValue.asInstanceOf[Int]
```
andygrove, line 505: Rather than use `asInstanceOf` it would be better to use pattern matching so we can throw nicer errors if we get an unexpected type here. Something like:

```scala
repeatTimes.getValue match {
  case repeatTimesVal: Int => strs.getBase.repeatStrings(repeatTimesVal)
  case _ => throw new IllegalStateException(s"Invalid data type for repeatTimes")  
}
```

andygrove: It would be good to show the actual type in the error message as well

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
jlowe: Looking at #8491 which added the JNI bindings for `repeat_strings`, it appears to be exposed under `Scalar`.  Therefore I think we could avoid exploding the input into a full column by doing something like this:
```scala
    withResource(str.getBase.repeatString(repeatTimes.getValue.asInstanceOf[Int])) { repeatScalar =>
      GpuColumnVector.from(repeatScalar, numRows, StringType)
    }
```

jlowe:
@ttnghia the commit needs to be signed off, see the [signoff section in the contributing guide](https://github.com/NVIDIA/spark-rapids/blob/branch-21.08/CONTRIBUTING.md#sign-your-work) for details.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
ttnghia: Wow, I didn't know that thus didn't implement it in cudf. I may need to do that too.

sameerz:
@ttnghia please retarget to branch 21.10 when you get a chance.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -484,6 +484,35 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+case class GpuStringRepeat(strs: Expression, repeatTimes: Expression)
+    extends GpuBinaryExpression with ImplicitCastInputTypes {
+  override def left: Expression = strs
+
+  override def right: Expression = repeatTimes
+
+  override def dataType: DataType = strs.dataType
+
+  override def inputTypes: Seq[AbstractDataType] =
+    Seq(TypeCollection(StringType, BinaryType), IntegerType)
```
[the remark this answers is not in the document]
ttnghia: Copy-past error :smiley:

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -487,6 +487,60 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+  }
+
+  def doColumnar(numRows: Int, input: GpuScalar, repeatTimes: GpuScalar): ColumnVector = {
+    repeatTimes.getValue match {
+      case repeatTimesVal: Int => { // only support Int type for this overload
+        withResource(input.getBase.repeatString(repeatTimesVal)) {
+          repeatedString => ColumnVector.fromScalar(repeatedString, 1)
+        }
+      }
+      case _ => throw new IllegalStateException("Invalid data type for repeatTimes (must be INT32)")
```
ttnghia, line 538: Can somebody confirm me that this pattern matching can handle the null case for `repeatTimes`? Thanks.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2472,6 +2472,16 @@ object GpuOverrides {
               .withPsNote(TypeEnum.STRING, "only a single character is allowed"), TypeSig.STRING),
           ParamCheck("count", TypeSig.lit(TypeEnum.INT), TypeSig.INT))),
       (in, conf, p, r) => new SubstringIndexMeta(in, conf, p, r)),
+    expr[StringRepeat](
+      "StringRepeat operator",
+      ExprChecks.projectNotLambda(TypeSig.STRING, TypeSig.STRING,
+        Seq(ParamCheck("input", TypeSig.lit(TypeEnum.STRING) + TypeSig.STRING, TypeSig.STRING),
+          ParamCheck("repeatTimes", TypeSig.lit(TypeEnum.INT) + TypeSig.INT, TypeSig.INT))),
```
ttnghia, line 2479: I'm not sure if those checks are correct. I want to check if
 * The first input to this function can either be a string scalar or a strings column.
 * The second input can be an integer number or a numeric column of integer numbers.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
jlowe: To check for null you need to check if the scalar is valid or not (i.e.: `repeatTimes.isValid`).  Accessing a scalar value when it is not valid is undefined behavior and will likely assert in the future.

To check the type examine the scalar type, e.g.: `repeatTimes.dataType == IntegerType`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
jlowe: You can examine the generated documentation for clues as to whether you got the type checking setup correctly.  Also adding unit tests for these cases will further prove this.  Adding `TypeSig.lit` appears to be limiting the support to a scalar, so I'd remove that.  Simply allowing strings for that parameter should cause columnar inputs to call your columnar version of `doColumnar` and a scalar input will call the form that takes a scalar.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -487,6 +487,60 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
 
+case class GpuStringRepeat(input: Expression, repeatTimes: Expression)
+    extends GpuBinaryExpression with ImplicitCastInputTypes with NullIntolerant {
+  override def left: Expression = input
+  override def right: Expression = repeatTimes
+  override def dataType: DataType = input.dataType
+  override def inputTypes: Seq[AbstractDataType] = Seq(StringType, IntegerType)
+
+  def doColumnar(lhs: GpuScalar, rhs: GpuColumnVector): ColumnVector =
+    throw new UnsupportedOperationException(s"Cannot columnar evaluate expression: $this")
```
jlowe, line 498: Why does this need to be the case?  If the input string is a scalar but the repeat times are columnar, we can fill a column of the same length with the scalar and make this work, right?  For example, something like this:
```
  withResource(GpuColumnVector.from(lhs, rhs.getRowCount, lhs.dataType)) { replicatedInput =>
    doColumnar(replicatedInput, rhs)
  }
```

It looks like Spark CPU supports this case, e.g.:
```
scala> val df = Seq(1, 2, 3).toDF
df: org.apache.spark.sql.DataFrame = [value: int]

scala> df.withColumn("repeated", expr("repeat('abc', value)")).show
+-----+---------+
|value| repeated|
+-----+---------+
|    1|      abc|
|    2|   abcabc|
|    3|abcabcabc|
+-----+---------+
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -487,6 +487,60 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+    val repeatTimesCV = repeatTimes.getBase
+
+    // Compute the output size to check for overflow.
+    val outputSizes = input.getBase.repeatStringsSizes(repeatTimesCV)
+    if (outputSizes.getTotalSize > Int.MaxValue) {
+      throw new RuntimeException("Output strings have total size exceed maximum allowed size")
+    }
+
+    // Finally repeat the strings using the pre-computed strings' sizes.
+    input.getBase.repeatStrings(repeatTimesCV, outputSizes.getStringSizes)
```
jlowe, line 510: This leaks the column within `outputSizes`.  Use `withResource` to avoid this, e.g.:
```suggestion
    withResource(input.getBase.repeatStringsSizes(repeatTimesCV)) { outputSizes =>
      if (outputSizes.getTotalSize > Int.MaxValue) {
        throw new RuntimeException("Output strings have total size exceed maximum allowed size")
      }

      // Finally repeat the strings using the pre-computed strings' sizes.
      input.getBase.repeatStrings(repeatTimesCV, outputSizes.getStringSizes)
    }
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -487,6 +487,60 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+  }
+
+  def doColumnar(input: GpuColumnVector, repeatTimes: GpuScalar): ColumnVector = {
+    repeatTimes.getValue match {
```
jlowe, line 514: As mentioned below, this needs to check the validity of the scalar before attempting to access its value and should switch on `repeatTimes.dataType`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
revans2: The `TypeSig.lit(..)` means it is adding a limitation that it only supports a literal input value.  `TypeSig.INT` indicates it can support both a literal and a column as input.  Mixing them I think gives preference to the lit version so this code is saying that `repeatTimes` and `input` both can only support a literal input value, which when Spark sees this it typically runs this during compile and replaces the entire thing with a literal that is the output of running the StringRepeat command. From the code it looks like we can drop the `TypeSig.lit` parts entirely.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2472,6 +2472,16 @@ object GpuOverrides {
               .withPsNote(TypeEnum.STRING, "only a single character is allowed"), TypeSig.STRING),
           ParamCheck("count", TypeSig.lit(TypeEnum.INT), TypeSig.INT))),
       (in, conf, p, r) => new SubstringIndexMeta(in, conf, p, r)),
+    expr[StringRepeat](
+      "StringRepeat operator",
```
revans2, line 2476: Could we put in a more user facing description? Spark uses the string
```
Returns the string which repeats the given string value n times
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
revans2: I don't think we have to switch on the data type because we are restricting it to ints, but an assertion that it is the expected type would be good.  We do have to check the validity before we pull the data out.

integration_tests/src/main/python/string_test.py
```diff
@@ -286,6 +286,47 @@ def test_substring():
...
+                'repeat({}, NULL)'.format(s)
+            ))
+
+def test_repeat_scalar_and_column():
+    gen_s = StringGen(nullable=False)
+    gen_r = IntegerGen(min_val=-100, max_val=100, special_cases=[0], nullable=True)
+    (s,) = gen_scalars_for_sql(gen_s, 1)
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark: unary_op_df(spark, gen_r).selectExpr(
+                'repeat(NULL, a)'.format(s),
```
jlowe, line 310:
```suggestion
                'repeat(NULL, a)',
```

integration_tests/src/main/python/string_test.py
```diff
@@ -286,6 +286,47 @@ def test_substring():
                 'SUBSTRING(a, 1, NULL)',
                 'SUBSTRING(a, 0, 0)'))
 
+def test_repeat_scalar_and_scalar():
```
revans2, line 289: You should look at the plan produced for this test, but I don't think it is covering what you want.  Spark will do constant folding, which means for most SQL expressions if all of the inputs are literal values, then it will execute it at compile time (on the CPU) and replace it with a literal value that is the result of the computation. If that is the case we can keep the test, but it is not really doing much and we should at least document it, but I personally would remove it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -487,6 +487,88 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+      input.getBase.repeatStrings(repeatTimesCV, outputSizes.getStringSizes)
+    }
+  }
+
+  def doColumnar(input: GpuColumnVector, repeatTimes: GpuScalar): ColumnVector = {
+    assert(repeatTimes.dataType == IntegerType)
+
+    if (!repeatTimes.isValid) {
+      // If the input scala repeatTimes is invalid, the results should be all nulls.
+      // This is equivalent to having repeatTimes == 0.
```
jlowe, line 525: Is this a true?  Looking at the Spark code, I would expect repeat times <= 0 to produce an _empty_ string, not a null string.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -487,6 +487,88 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+    }
+  }
+
+  def doColumnar(numRows: Int, input: GpuScalar, repeatTimes: GpuScalar): ColumnVector = {
+    assert(input.dataType == StringType && repeatTimes.dataType == IntegerType)
+
+    if (!repeatTimes.isValid) {
+      // If the input scala repeatTimes is invalid, the results should be all nulls.
+      // This is equivalent to having repeatTimes == 0.
+      withResource(input.getBase.repeatString(0)) {
```
revans2, line 554: nit: would it be better to just create the null scalar directly instead of calling `repeatString(0)`?

```
withResource(Scalar.fromNull(DType.STRING)) {
```

If you do want to keep it this way, I think it might be cleaner to combine some of the code together like.

```
val numTimes = if (repeatTimes.isValid) {
  repearTimes.getValue match {
    case repeatTimesVal: Int => repeatTimesVal
    case _ => throw new IllegalStateException("Invalid data type for repeatTimes " +
            "(must be INT32)")
  }
} else {
  // If the input scala repeatTimes is invalid, the results should be all nulls.
  // This is equivalent to having repeatTimes == 0.
  0
}

withResource(input.getBase.repeatString(numTimes)) {
  repeatedString => ColumnVector.fromScalar(repeatedString, numRows)
}
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -487,6 +487,88 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+        repeatedString => ColumnVector.fromScalar(repeatedString, 1)
```
revans2, line 555: The number of items should be numRows, not 1

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -487,6 +487,88 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+      }
+    } else {
+      repeatTimes.getValue match {
+        case repeatTimesVal: Int => { // only support Int type for this overload
+          withResource(input.getBase.repeatString(repeatTimesVal)) {
+            repeatedString => ColumnVector.fromScalar(repeatedString, 1)
```
revans2, line 561: The number of items should be numRows not 1

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -487,6 +487,88 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+
+  def doColumnar(input: GpuColumnVector, repeatTimes: GpuScalar): ColumnVector = {
+    assert(repeatTimes.dataType == IntegerType)
+
+    if (!repeatTimes.isValid) {
+      // If the input scala repeatTimes is invalid, the results should be all nulls.
+      // This is equivalent to having repeatTimes == 0.
+      input.getBase.repeatStrings(0)
+    } else {
+      repeatTimes.getValue match {
```
jlowe, line 528: Like is done for the other types here, I'd just add an `assert(repeatTimes.dataType == IntegerType)` and call `repeatTimes.getBase.getInt`.  Spark will already be checking that the input is IntegerType while creating the logical plan.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -487,6 +487,88 @@ case class GpuInitCap(child: Expression) extends GpuUnaryExpression with Implici
...
+    assert(input.dataType == StringType && repeatTimes.dataType == IntegerType)
+
+    if (!repeatTimes.isValid) {
+      // If the input scala repeatTimes is invalid, the results should be all nulls.
+      // This is equivalent to having repeatTimes == 0.
+      withResource(input.getBase.repeatString(0)) {
+        repeatedString => ColumnVector.fromScalar(repeatedString, 1)
+      }
+    } else {
+      repeatTimes.getValue match {
```
jlowe, line 558: The code above is already asserting on the datatype, so the match can be replaced with `repeatTimes.getBase.getInt`.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
ttnghia: Sorry I was wrong. Yes, repeat times <= 0 will result in empty strings. Will fix this.

jlowe: I thought one of tests would catch this.  Is that not happening?

ttnghia: I have no idea. There is already unit test for that case (the last one below):

```def test_repeat_column_and_scalar():
    gen_s = StringGen(nullable=True)
            lambda spark: unary_op_df(spark, gen_s).selectExpr(
                'repeat(a, -10)',
                'repeat(a, 0)',
                'repeat(a, 10)',
                'repeat(a, NULL)'
            ))
```

After fixing this (to return a column of nulls, instead of calling `repeat(a, 0`), the tests still passed all.

ttnghia: Maybe that last unit test calls the `repeat(column, column)` version?

integration_tests/src/main/python/string_test.py
ttnghia: Do you mean such tests are equivalent to something like `assert('abc' == 'abc')`? Is there any way to prevent constant folding? Otherwise, we will not be able to test for those cases.

revans2: Close to it.  It is not going to be running any of your code.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
revans2: Put in some debugging and find out.

jlowe: Maybe.  It's a red flag that you can change the behavior and tests that are supposed to detect that behavior don't notice.  Like @revans2 indicated, it's best to manually verify the tests do what you think they're going to do, otherwise we can end up writing redundant or useless tests.

It would be good to understand why the test that's supposed to catch this doesn't in practice and fix it.

integration_tests/src/main/python/string_test.py
ttnghia: Verify this: None of the code is run with these scalar-scalar tests. So I'm going to remove these tests.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
ttnghia: @revans2 @jlowe I added `throw new RuntimeException` into all my `doColumnar` functions. The following tests passed, which means these expressions are not executed by my code. So, these expressions are either ignored, or executed somewhere else:
```
def test_repeat_scalar_and_column():
    gen_r = IntegerGen(min_val=-100, max_val=100, special_cases=[0], nullable=True)
            lambda spark: unary_op_df(spark, gen_r).selectExpr(
                'repeat(NULL, a)'
            ))

def test_repeat_column_and_scalar():
    gen_s = StringGen(nullable=True)
            lambda spark: unary_op_df(spark, gen_s).selectExpr(
                'repeat(a, NULL)'
            ))
```

ttnghia: BTW, is there any way to detect, and force NOT to, such compile-time evaluation of expressions?

revans2: It shouldn't be runtime folded because StringRepeat is a regular binary op with

```
  override def foldable: Boolean = left.foldable && right.foldable
```

There is a way to disable it, but we guard against it and https://github.com/NVIDIA/spark-rapids/pull/1751/files makes it so we would fall back to the CPU in many of those cases.  The code there is a very special case where some times some of the code can return a constant so we need to cover it, but it does not need to be that fast.

jlowe: I'm not so sure this is solely a `foldable` thing.  `StringRepeat` is marked as `NullIntolerant` in Spark, and I wouldn't be surprised if the optimizer elided any subexpression marked `NullIntolerant` when any inputs can be computed at query planning time as null.

revans2: The code sure looks like it is just about foldable.

https://github.com/apache/spark/blob/0f538402fb76e4d6182cc881219d53b5fdf73af1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala#L41-L71

jlowe: `StringRepeat` in Spark is marked as `NullIntolerant` so I suspect the query optimizer eliminates any subexpression that it can tell at query planning time is going to receive a NULL.

I think we can make this work by adding a null column to the dataset and then performing a scalar subquery on it (e.g.: `SELECT repeat(stringscol, (SELECT first(nullcol) from dataset)) from dataset`), although I wouldn't be surprised if something like AQE ends up foiling the obfuscation.

revans2: Yup it is a separate optimization. https://github.com/apache/spark/blob/0f538402fb76e4d6182cc881219d53b5fdf73af1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala#L793-L794

I am not sure it is worth spending a lot of time hacking around this to make the test work, because it is highly unlikely that anyone will actually hit it. Who turns off these optimizations anyways.  But if you want to play games like what @jlowe is suggesting you can.

jlowe: > The code sure looks like it is just about foldable.

Yes, foldable can explain this if both sides are foldable, but even if they aren't (and in this case since one side is a column it's probably not foldable) `NullIntolerant` is where I think the trivial NULL projection is occurring.
```
/**
 * Replaces [[Expression Expressions]] that can be statically evaluated with
 * equivalent [[Literal]] values. This rule is more specific with
 * Null value propagation from bottom to top of the expression tree.
 */
object NullPropagation extends Rule[LogicalPlan] {
[..]
      // Non-leaf NullIntolerant expressions will return null, if at least one of its children is
      // a null literal.
      case e: NullIntolerant if e.children.exists(isNullLiteral) =>
        Literal.create(null, e.dataType)
    }
```

tests/src/test/scala/com/nvidia/spark/rapids/StringFunctionSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.nvidia.spark.rapids
+
+class StringFunctionSuite extends SparkQueryCompareTestSuite {
```
ttnghia, line 19: I tried to add the test cases for scalar-scalar input, but it seems that those cases are compiled-time evaluated because my code didn't run.

tests/src/test/scala/com/nvidia/spark/rapids/StringFunctionSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+      "repeat(strings, 0)",
+      "repeat(strings, 10)"
+    )
+  }
+
+  testSparkResultsAreEqual("Test repeat string(s) with strings column and column repeatTimes",
+    nullableStringsIntsDf) {
+    frame => frame.selectExpr(
+      "repeat(strings, ints)",
+      "repeat(strings, -ints)"
```
ttnghia, line 42: The `ints` column only has positive values. I'm not sure if using `-ints` actually means negating these values in an expression?

tests/src/test/scala/com/nvidia/spark/rapids/StringFunctionSuite.scala
jlowe: To test scalar-scalar input, you would need to write a unit test, not something that goes through Spark's optimizer, i.e.: constructing a `GpuStringRepeat` instance directly and invoking `columnarEval` on it with two child expressions that will `columnarEval` to a `GpuScalar`.

tests/src/test/scala/com/nvidia/spark/rapids/StringFunctionSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package com.nvidia.spark.rapids
+
+class StringFunctionSuite extends SparkQueryCompareTestSuite {
+  testSparkResultsAreEqual("Test repeat string(s) with scalar string and column repeatTimes",
+    intsDf) {
+    frame => frame.selectExpr(
+      "repeat(NULL, ints)",
```
jlowe, line 23: Does this actually test anything?  This will be similar to the Python test that passed a NULL literal, and I would expect Spark's optimizer to simplify this to a simple `null` project.  Adding tests for `NULL` inputs will similarly require bypassing Spark like the testing for scalars above.

tests/src/test/scala/com/nvidia/spark/rapids/StringFunctionSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+      "repeat('abc123', ints)"
```
jlowe, line 24: Are these Scala tests adding any value beyond what was tested in PySpark via pytest?  These are all going through the Spark APIs and I don't see how they're adding any value beyond what we could accomplish in PySpark.  Apologies if I'm missing something.

tests/src/test/scala/com/nvidia/spark/rapids/StringFunctionSuite.scala
jlowe: Yes, this will negate the ints input expression.

tests/src/test/scala/com/nvidia/spark/rapids/StringFunctionSuite.scala
ttnghia: Yeah, I have no idea why. My code actually ran that test (I just throw an exception in my function, and this case invokes such throwing in doColumnar(scalar-column)).

tests/src/test/scala/com/nvidia/spark/rapids/StringFunctionSuite.scala
ttnghia: Sorry maybe I'm wrong here when testing using Spark expresison. I'm rewriting the tests with `columnarEval` on `GpuStringRepeat` instance.

tests/src/test/scala/com/nvidia/spark/rapids/StringFunctionSuite.scala
jlowe: If you are just checking if doColumnar(scalar,column) is being called then that will be true due to the second expression being used, `repeat('abc123', ints)` which won't be optimized out.

ttnghia: No, I tested that for one expression at a time. Yeah, consistently inconsistent :smiley:

tests/src/test/scala/com/nvidia/spark/rapids/unit/StringUnitTest.scala
```diff
@@ -0,0 +1,159 @@
...
+          }
+        }
+
+      doTest(-10, HostColumnVector.fromStrings(null, "", "", "", ""))
+      doTest(0, HostColumnVector.fromStrings(null, "", "", "", ""))
+      doTest(2, HostColumnVector.fromStrings(null, "aa", "", "123123", "éáéá"))
+    }
+  }
+
+  test("Test StringRepeat with column strings and column repeatTimes") {
```
jlowe, line 141: Isn't this test already covered by the PySpark tests?

tests/src/test/scala/com/nvidia/spark/rapids/unit/StringUnitTest.scala
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  test("Test StringRepeat with column strings and scalar repeatTimes") {
+    val strsCol = ColumnVector.fromStrings(null, "a", "", "123", "éá")
+
+    withResource(GpuColumnVector.from(strsCol, DataTypes.StringType)) { strGpuColumn =>
+      // Test repeat(strs, NULL).
+      withResource(GpuScalar(null, DataTypes.IntegerType)) { nullInt =>
+        withResource(GpuStringRepeat(null, null).doColumnar(strGpuColumn, nullInt)) { result =>
+          withResource(result.copyToHost()) { hostResult =>
```
jlowe, line 117: Should assert number of rows is as expected, otherwise this could return an empty column and pass the test.

tests/src/test/scala/com/nvidia/spark/rapids/unit/StringUnitTest.scala
```diff
@@ -0,0 +1,159 @@
...
+            (0 until hostResult.getRowCount.asInstanceOf[Int]).foreach { i =>
+              assertResult(hostResult.isNull(0))(true)
```
jlowe, line 119: This is not using `i`, only checking the first row `i` times.

Could simplify this and avoid the host transfer with something like:
```
  assert(strGpuColumn.getRowCount)(result.getRowCount)
  assertResult(result.getNullCount)(result.getRowCount)
```

tests/src/test/scala/com/nvidia/spark/rapids/unit/StringUnitTest.scala
```diff
@@ -0,0 +1,159 @@
...
+
+    withResource(GpuColumnVector.from(intCol, DataTypes.IntegerType)) { intGpuColumn =>
+      // Test repeat(NULL, intVal).
+      withResource(GpuScalar(null, DataTypes.StringType)) { nullStr =>
+        withResource(GpuStringRepeat(null, null).doColumnar(nullStr, intGpuColumn)) { result =>
+          withResource(result.copyToHost()) { hostResult =>
+            (0 until hostResult.getRowCount.asInstanceOf[Int]).foreach { i =>
+              assertResult(hostResult.isNull(0))(true)
+            }
+          }
```
jlowe, line 92: Same missing row count check and always checking first row in the loop issues.

tests/src/test/scala/com/nvidia/spark/rapids/unit/StringUnitTest.scala
```diff
@@ -0,0 +1,159 @@
...
+        }
+      }
+
+      // Test repeat(strs, intVal).
+      val doTest = (intVal: Int, expectedStrs : HostColumnVector) =>
```
jlowe, line 126: Isn't this part of the test already covered by the PySpark tests?

tests/src/test/scala/com/nvidia/spark/rapids/unit/StringUnitTest.scala
ttnghia: It has been covered in PySpark test. I want to do that again here to maximize coverage.

jlowe: Our CI testing already takes too long, why would we want redundant tests if they aren't adding any value?

tests/src/test/scala/com/nvidia/spark/rapids/unit/StringUnitTest.scala
```diff
@@ -0,0 +1,159 @@
...
+
+package com.nvidia.spark.rapids.unit
+
+import ai.rapids.cudf.{ColumnVector, HostColumnVector}
+import com.nvidia.spark.rapids.{GpuUnitTests, GpuColumnVector, GpuScalar, CudfTestHelper}
+import org.apache.spark.sql.rapids.GpuStringRepeat
+import org.apache.spark.sql.functions.{col, lit}
+import org.apache.spark.sql.types.DataTypes
+
+class StringUnitTest extends GpuUnitTests {
```
jlowe, line 25: IMO this should be renamed to `StringRepeatSuite` (and the file renamed accordingly).  We don't want all string tests to be collected in this file.

tests/src/test/scala/com/nvidia/spark/rapids/unit/StringUnitTest.scala
ttnghia: Ok then I'm going to remove the repeated tests :smiley: .

tests/src/test/scala/com/nvidia/spark/rapids/unit/StringUnitTest.scala
ttnghia: Should be `StringRepeatUnitTest`, right? I observed that `*Suite.scala` are all for expr testing. The unit tests are in `*UnitTest.scala` files under the `unit` folder.

jlowe: Sure, that works too.  I'm used to scala tests being called a `Suite` because that's how scalatest refers to them.  "Suite" does not preclude being a unit test, but if that's how other things in this package want to be called, so be it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2484,6 +2484,16 @@ object GpuOverrides {
               .withPsNote(TypeEnum.STRING, "only a single character is allowed"), TypeSig.STRING),
           ParamCheck("count", TypeSig.lit(TypeEnum.INT), TypeSig.INT))),
       (in, conf, p, r) => new SubstringIndexMeta(in, conf, p, r)),
+    expr[StringRepeat](
+      "StringRepeat operator, which repeat the given strings by given number(s) of times",
```
gerashegalov, line 2488:
```suggestion
      "StringRepeat operator that repeats the input string the repeatTimes number of times",
```

tests/src/test/scala/com/nvidia/spark/rapids/unit/StringRepeatUnitTest.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  test("Test StringRepeat with NULL scalar string and column repeatTimes") {
+    val intCol = ColumnVector.fromBoxedInts(-3, null, -1, 0, 1, 2, null)
+
+    withResource(GpuColumnVector.from(intCol, DataTypes.IntegerType)) { intGpuColumn =>
+      // Test repeat(NULL, intVal).
+      withResource(GpuScalar(null, DataTypes.StringType)) { nullStr =>
+        withResource(GpuStringRepeat(null, null).doColumnar(nullStr, intGpuColumn)) { result =>
+          assertResult(intGpuColumn.getRowCount)(result.getRowCount)
+          assertResult(result.getNullCount)(result.getRowCount)
```
jlowe, line 88: Nit: The expected value is the row count (which we already validated above) and the value being tested is the null count, so these should be flipped.
```suggestion
          assertResult(result.getRowCount)(result.getNullCount)
```

tests/src/test/scala/com/nvidia/spark/rapids/unit/StringRepeatUnitTest.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  test("Test StringRepeat with column strings and NULL scalar repeatTimes") {
+    val strsCol = ColumnVector.fromStrings(null, "a", "", "123", "éá")
+
+    withResource(GpuColumnVector.from(strsCol, DataTypes.StringType)) { strGpuColumn =>
+      // Test repeat(strs, NULL).
+      withResource(GpuScalar(null, DataTypes.IntegerType)) { nullInt =>
+        withResource(GpuStringRepeat(null, null).doColumnar(strGpuColumn, nullInt)) { result =>
+          assertResult(strGpuColumn.getRowCount)(result.getRowCount)
+          assertResult(result.getNullCount)(result.getRowCount)
```
jlowe, line 102:
```suggestion
          assertResult(result.getRowCount)(result.getNullCount)
```

jlowe (dismissed):
Looks good to me other than a small assert result nit.

jlowe (dismissed):

jlowe:
build

revans2:
The latest test run failed because of a breaking change in the cudf APIs that was merged in and the corresponding plugin change has not gone in yet.

ttnghia:
build

pxLi:
hi. please check the build log before trigger againm thx!
```
[2021-08-04T16:34:40.039Z] found modified files during mvn verify:
[2021-08-04T16:34:40.039Z]  M docs/configs.md
[2021-08-04T16:34:40.039Z]  M docs/supported_ops.md
[2021-08-04T16:34:40.039Z] [ERROR] Command execution failed.
```
this PR does not include RapidsConf and SupportedOpsDocs update, so it failed the check

ttnghia:
> hi. please check the build log before trigger again thx!
> 
> ```
> [2021-08-04T16:34:40.039Z] found modified files during mvn verify:
> [2021-08-04T16:34:40.039Z]  M docs/configs.md
> [2021-08-04T16:34:40.039Z]  M docs/supported_ops.md
> [2021-08-04T16:34:40.039Z] [ERROR] Command execution failed.
> ```
> 
> this PR did not include RapidsConf and SupportedOpsDocs update, so it failed the check

Oh thanks for letting me know.
@jlowe I hate to ask you to approve the third time but I have to. I also hate github that discards PR approval every time I push.

pxLi:
> > hi. please check the build log before trigger again thx!
> > ```
> > [2021-08-04T16:34:40.039Z] found modified files during mvn verify:
> > [2021-08-04T16:34:40.039Z]  M docs/configs.md
> > [2021-08-04T16:34:40.039Z]  M docs/supported_ops.md
> > [2021-08-04T16:34:40.039Z] [ERROR] Command execution failed.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this PR did not include RapidsConf and SupportedOpsDocs update, so it failed the check
> 
> Oh thanks for letting me know. I have updated the docs.
> @jlowe I hate to ask you to approve the third time but I have to. I also hate github that discards PR approval every time I push. Maybe we need to modify that behavior for better convenience.

Actually I think we should not, otherwise people could push new commits which could do whatever they want to our repo after the initial approval, this could be horrible :P

ttnghia:
> > Oh thanks for letting me know. I have updated the docs.
> > @jlowe I hate to ask you to approve the third time but I have to. I also hate github that discards PR approval every time I push. Maybe we need to modify that behavior for better convenience.
> 
> Actually I think we should not, otherwise people could push new commits which could do whatever they want to our repo after the initial approval, this could be horrible :P

People should know what they are doing, not "whatever they want" like that :smiley:. I observed (from cudf) that after approvals the authors usually push more minor changes related to docs/typos etc. If every new change requires approvals again then it is very inconvenient for both the authors and the reviewers.

pxLi:
> > > Oh thanks for letting me know. I have updated the docs.
> > > @jlowe I hate to ask you to approve the third time but I have to. I also hate github that discards PR approval every time I push. Maybe we need to modify that behavior for better convenience.
> > 
> > 
> > Actually I think we should not, otherwise people could push new commits which could do whatever they want to our repo after the initial approval, this could be horrible :P
> 
> People should know what they are doing, not "whatever they want" like that 😃. I observed (from cudf) that after approvals the authors usually push more minor changes related to docs/typos etc. If every new change requires approvals again then it is very inconvenient for both the authors and the reviewers.
> 
> Of course, if there are new changes in the code then the authors can request re-review anytime.

haha yeah, I mean cudf approval is horrible. For our repo, if PR tests pass pre-merge CI but requires to add some minor changes like comments only, people could just add `[skip ci]` to the title to skip the additional full tests run w/ new commit approval :)

sameerz:
> > > > Oh thanks for letting me know. I have updated the docs.
> > > > @jlowe I hate to ask you to approve the third time but I have to. I also hate github that discards PR approval every time I push. Maybe we need to modify that behavior for better convenience.
> > > 
> > > 
> > > Actually I think we should not, otherwise people could push new commits which could do whatever they want to our repo after the initial approval, this could be horrible :P
> > 
> > 
> > People should know what they are doing, not "whatever they want" like that 😃. I observed (from cudf) that after approvals the authors usually push more minor changes related to docs/typos etc. If every new change requires approvals again then it is very inconvenient for both the authors and the reviewers.
> > Of course, if there are new changes in the code then the authors can request re-review anytime.
> 
> haha yeah, I mean cudf approval is horrible. For our repo, if PR tests pass pre-merge CI but requires to add some minor changes like comments/docs only (not for generated docs like configs.md or supported_ops.md), people could just add `[skip ci]` to the title to skip the additional full tests run w/ new commit approval :)

It may be inconvenient, but a review on a change, even a small one, is worth the trouble to minimize potential problems.  Doing a review on every change is a good habit.  We can still move quickly while doing the reviews.

jlowe (approved):

jlowe:
build

PR #2730, opened 2021-06-17
qualification and profiling tool support rolled and compressed event logs for CSPs and Apache Spark

tgravescs:
The main part of this pr is to support compressed and rolled event logs from the various CSPs and Apache Spark.  This also has some cleanup to consolidate some duplicate code, move some code to common class for dealing with parsing event log paths, compressing some of the test files and a few style type changes.

The Databricks event logs are different then Apache spark, so special handling was added for them.


tgravescs:
build

tgravescs:
build

tgravescs:
I'm guessing this may fail against because I just compressed the existing event log files that were there to make them smaller but 2 are over 1.5m, I may separate those out if that is the case

tgravescs:
closing and will reopen without those files

Issue #2690, opened 2021-06-10
[BUG] Profiling tool doesn't properly read rolled log files


tgravescs:
**Describe the bug**
Profile and qualification tools output comes out empty reading rolled event log log files.  This should cover reading logs from apache spark, EMR, dataproc, microsoft, and Databricks

PR #2732, opened 2021-06-17
qualification and profiling tool support rolled and compressed event logs for CSPs and Apache Spark

tgravescs:
The main part of this pr is to support compressed and rolled event logs from the various CSPs and Apache Spark. This also has some cleanup to consolidate some duplicate code, move some code to common class for dealing with parsing event log paths, compressing some of the test files and a few style type changes.

The Databricks event logs are different then Apache spark, so special handling was added for them.

fixes https://github.com/NVIDIA/spark-rapids/issues/2690


tgravescs:
build

tools/src/main/scala/com/nvidia/spark/rapids/tool/EventLogPathProcessor.scala
```diff
@@ -0,0 +1,258 @@
...
+  // Databricks has the latest events in file named eventlog and then any rolled in format
+  // eventlog-2021-06-14--20-00.gz
+  def isDatabricksEventLogDir(dir: FileStatus, fs: FileSystem): Boolean = {
+    // try to determine if dir structure looks right
+    val dirList = fs.listStatus(dir.getPath)
+    if (dirList.size > 0) {
+      if (dirList.exists(_.getPath.getName.equals("eventlog"))) {
+        if (dirList.size > 1) {
+          dirList.exists(_.getPath.getName
+            .matches("eventlog-([0-9]){4}-([0-9]){2}-([0-9]){2}--([0-9]){2}-([0-9]){2}.*"))
```
gerashegalov, line 75: this will keep recompiling regex. We need to create a pattern object outside the dir list loop. e.g. via `"regex".r`  or java `Pattern.compile`. 

we could also pass a `PathFilter` instance to `listStatus`  accepting:  `name  == eventlog || matches(regex)` . Then the return value of the `isDatabricksEventLogDir` is `res.contains("eventlog") && res.size() > 1`

tools/src/main/scala/com/nvidia/spark/rapids/tool/EventLogPathProcessor.scala
```diff
@@ -0,0 +1,258 @@
...
+        false
+      }
+    } else {
+      false
+    }
+  }
+
+  def getEventLogInfo(pathString: String): Map[EventLogInfo, Long] = {
+    val inputPath = new Path(pathString)
+    val fs = inputPath.getFileSystem(new Configuration())
```
gerashegalov, line 89: we should  wire the hadoop config from Spark Context to here instead of creating a new one.

tools/src/main/scala/com/nvidia/spark/rapids/tool/EventLogPathProcessor.scala
tgravescs: yeah we probably don't even need the regex based on how everything else is done now, I'll update

tgravescs:
build

tgravescs:
investigating test failures, it seems in this env something must be deleted while its open, locally no issues and everything cleaned up

tgravescs:
build

tgravescs:
@gerashegalov @nartal1

tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/ApplicationInfo.scala
```diff
@@ -203,30 +207,62 @@ class ApplicationInfo(
   // Create Spark DataFrame(s) based on ArrayBuffer(s)
   arraybufferToDF()
 
+  private val codecMap = new ConcurrentHashMap[String, CompressionCodec]()
+
+  def openEventLogInternal(log: Path, fs: FileSystem): InputStream = {
+    EventLogFileWriter.codecName(log) match {
+      case c if (c.isDefined && c.get.equals("gz")) =>
+        val in = new BufferedInputStream(fs.open(log))
```
gerashegalov, line 215: BufferInputStream is not needed for GzipInputStream. `in = fs.open(log))`

tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/ApplicationInfo.scala
```diff
@@ -203,30 +207,62 @@ class ApplicationInfo(
...
   def processEvents(): Unit = {
-    logInfo("Parsing Event Log File: " + eventlog.toString)
+    val eventlog = eventLogInfo.eventLog
 
-    val fs = FileSystem.get(eventlog.toUri,new Configuration())
-    var totalNumEvents = 0
+    logInfo("Parsing Event Log: " + eventlog.toString)
 
+    // at this point all paths should be valid event logs or event log dirs
+    val fs = eventlog.getFileSystem(new Configuration())
```
gerashegalov, line 236: prefer to wire through hadoop conf from SparkContext to creating brand-new instances.

gerashegalov (approved):
LGTM, some nits

tgravescs:
thanks, @gerashegalov  I'll incorporate the nits in my next pr.

PR #2736, opened 2021-06-17
Remove trailing spaces and add a commit hook to prevent them

gerashegalov:


jlowe:
Do we really want this?  If this goes in but we don't mandate the use of the precommit hook then we're going to _increase_ the chance that a PR is going to seemingly randomly touch lines unrelated to a PR because some previous PR went in without the precommit hook.  We'd need a CI precommit check to verify the hook is being used to prevent it, and IMO that just makes it _harder_ to contribute to the project since it makes the potential contributor jump through a hoop that is somewhat difficult to justify in practice.  Like a coding style, there's no semantic difference either way, but _unlike_ a coding style, the difference isn't visual because it's whitespace.

If we're really going to do this, I'd personally rather see it done as part of a mandated coding style that would essentially force developers and contributors to use a precommit hook or script to reformat their PRs to pass a draconian CI precommit style check, much like the clang formatting and flake8 enforcement in the cudf project.  Doing just the whitespace change without also a coding style enforcement means we're in this weird middle ground where we either didn't really solve the "issue" because we aren't enforcing future commits or we are enforcing and thus forcing contributors to make changes that seem very arbitrary given the lack of any other coding style enforcement.

gerashegalov:
@jlowe I advocate for this check to be mandated. I wanted to create a separate PR for CI. However, I  see a marginal benefit of having the CI PR go in first over the situation today.

jlowe:
Mandating the removal of trailing whitespace is a form of an enforced coding standard, but in isolation it has the drawback that it has _no visible benefit_.  The code isn't any easier to read after the transform, and I think we can agree the potential disk space savings is not the goal here.

But then I'm wondering, what _is_ the goal?  What tangible benefit do we point to when potential contributors stumbling into this new hurdle ask why the hurdle exists?  From my perspective, this seems like an arbitrary restriction on contributions since I don't see the benefit of enforcing it.  If it was part of enforcing an overall coding standard that increases code readability then that makes sense to me if the project chooses to force a rigid coding standard.

andygrove:
Spark enforces this via scala style:

```xml
<check level="error" class="org.scalastyle.file.WhitespaceEndOfLineChecker" enabled="true"></check>
```

If we are going to enforce this, we should add this scala style check so that `mvn verify` fails on violations.

gerashegalov:
> Spark enforces this via scala style:
> 
> ```
> <check level="error" class="org.scalastyle.file.WhitespaceEndOfLineChecker" enabled="true"></check>
> ```
> 
> If we are going to enforce this, we should add this scala style check so that `mvn verify` fails on violations.

Adding checks / linters to `mvn verify` is a good idea, and add another layer of defense. 

pre-commit hook's advantages 
- it does not depend on the maven goals and lifecycle phases, 
- covers many file extensions out of the box, and is easily configurable for more or less  
- if the developer installs `pre-commit`, the check is much less likely to be skipped than maven verify before posting/updating  PRs

sameerz:
Will this impact any of the documentation PRs (markdown files)?

jlowe:
> Will this impact any of the documentation PRs (markdown files)?

Yes, markdown files are also filtered as shown in the files modified in this PR (e.g.: docs/download.md).  Any contributors would need to ensure their trailing whitespace is filtered or have their submission rejected once the precommit CI starts enforcing this.  I think we all agree it isn't sensible to put this in without also enforcing it in the precommit, but where we disagree is whether this is worth changing at this time.  I still stand by [my previous comment](https://github.com/NVIDIA/spark-rapids/pull/2736#issuecomment-863591292).

gerashegalov:
From my side, I just paused working on this because it's a lower-priority side issue for me, and I did not anticipate such discussion around it. It was standard in my past work.

Once the developer installs pre-commit we do enforce it in the pre-commit config
https://github.com/NVIDIA/spark-rapids/pull/2736/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R30. And I added the scalastyle rule per @andygrove's comment which the developer may equally skip running when submitting the PR.

I was thinking of adding a CI-side pre-merge check just like we do for large files with the error message how to automatically prevent this by installing pre-commit

I disagree that we should lump this issue with other formatting enforcement because this issue is more general and spans multiple languages including md.

sameerz:
Since this is at an impasse and not impacting functionality, can we close without any further action?

Issue #2698, opened 2021-06-10
[FEA] integration test framework should detect unintentionally failed executors


abellina:
There has been a couple of instances I have seen where executors fail in our integration tests. For example: https://github.com/NVIDIA/spark-rapids/issues/1520 and https://github.com/NVIDIA/spark-rapids/issues/2477. 

https://github.com/NVIDIA/spark-rapids/issues/2477 was a test failure that was caused by an executor leaving the cluster. This would be the same for any test that performs partition-local operations (in this case it was sorting within partitions), and then expect all of the rows to match between the CPU and GPU, as Spark uses the number of cores available (# executors active * # cores/executor) to determine how many map partitions to use.

This issue is to add check if we expect a static number of executors, and we haven't explicitly killed an executor, during integration tests. This would help us narrow down issues that are causing executors to go away like memory or potentially the heartbeat failures seen in https://github.com/NVIDIA/spark-rapids/issues/1520.

jlowe:
Seems like we would want to run with dynamic allocation explicitly disabled and task max attempts == 1 for most test setups, that way the test should fail if a task/executor fails.

abellina:
We can turn off dynamic allocation, but we normally (I think exclusively) run without it, because you'd need the external shuffle service for it to be enabled.

On max task failures, yes that can be enabled. And it would mean our tests should fail quickly and likely not mistakenly succeed on a retry when we don't expect (i.e. if we have an XFAIL it is probably coded so that it fails each time). This seems like a good change @jlowe and it should help uncover flaky tests. In this case, we should see a prior failure helping us uncover the real issue.

tgravescs:
dynamic allocation could be on in different CSPs or environments we run in, so if we are relying on it we should be explicit.

PR #2743, opened 2021-06-18
Disable dynamicAllocation and set maxFailures to 1 in integration tests

abellina:

Closes: #2698

This should make tests fail quicker instead of sometimes succeeding on a re-attempt. We have test failure that is triggered when the shape of the cluster changes, to gain a new executor or when an executor is getting removed between cpu and gpu sessions (https://github.com/NVIDIA/spark-rapids/issues/2477).

### DIFF
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -129,7 +129,9 @@ else
     export PYSP_TEST_spark_ui_showConsoleProgress='false'
     export PYSP_TEST_spark_sql_session_timeZone='UTC'
     export PYSP_TEST_spark_sql_shuffle_partitions='12'
+    # prevent cluster shape to change - and fail quicker rather than retry
     export PYSP_TEST_spark_task_maxFailures='1'
+    export PYSP_TEST_spark_dynamicAllocation_enabled='false'
     if ((${#TEST_PARALLEL_OPTS[@]} > 0));
     then
         export PYSP_TEST_spark_rapids_memory_gpu_allocFraction=$MEMORY_FRACTION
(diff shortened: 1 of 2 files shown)


jlowe (approved):

jlowe:
build

integration_tests/run_pyspark_from_build.sh
```diff
@@ -129,6 +129,9 @@ else
     export PYSP_TEST_spark_ui_showConsoleProgress='false'
     export PYSP_TEST_spark_sql_session_timeZone='UTC'
     export PYSP_TEST_spark_sql_shuffle_partitions='12'
+    # prevent cluster shape to change - and fail quicker rather than retry
+    export PYSP_TEST_spark_task_maxFailures='1'
```
tgravescs, line 133: these aren't getting applied when using the spark-submit way to run , correct?  Seems like we should be consistent

tgravescs:
so I was randomly looking at some integration builds and noticed they had task failures.  I'm wonder what all of our builds will fail when we enable this.  I guess we can merge and see, but I wouldn't merge on a friday.

abellina:
Yeah if they have task failures, but the job succeeded (i.e. not XFAIL) then that should be a failed test as far as I can tell. But absolutely agree: 
> but I wouldn't merge on a friday.

abellina:
Moved to draft to prevent accidental merges until Monday.

integration_tests/run_pyspark_from_build.sh
abellina: Nice catch. Are `jenkins/spark-tests.sh` and then `run_pyspark_from_build.sh` the only two places that trigger the tests? If so I'd need to change spark-tests.sh.

tgravescs: I think those are the only 2 in this repo

tgravescs (approved):

abellina:
build

revans2 (approved):

abellina:
Will merge this in now and keep an eye out for new failures.

Issue #2718, opened 2021-06-15
[BUG] CBO falls back to CPU to write to Parquet in some cases


andygrove:
**Describe the bug**
CBO incorrectly forces some writes to fall back to CPU. 

Here is an example from NDS q16.

## Without CBO

![q16-178](https://user-images.githubusercontent.com/934084/122084056-af180780-cdbe-11eb-8901-9dbe51c540f8.png)

## With CBO

![q16-180](https://user-images.githubusercontent.com/934084/122084060-b0e1cb00-cdbe-11eb-9a90-1f0ad6263a97.png)

**Steps/Code to reproduce bug**
Run NDS q16 at SF=3k with AQE, CBO, and decimal support enabled.

**Expected behavior**
CBO should not force the write to happen on CPU.

**Environment details (please complete the following information)**
Standalone.

**Additional context**
None

Issue #2717, opened 2021-06-15
[BUG] CBO forces joins back onto CPU in some cases


andygrove:
**Describe the bug**
With AQE, CBO, and decimal support enabled, NDS queries q24a, q24b, q59, q93 fall back to CPU for a `SortMergeJoin` that feeds a `BroadcastHashJoin`.

Here is an example from q24a. 

## Without CBO

![q24a-178](https://user-images.githubusercontent.com/934084/122076089-ce5f6680-cdb7-11eb-9ceb-b04868127697.png)

## With CBO

![q24a-180](https://user-images.githubusercontent.com/934084/122076104-d15a5700-cdb7-11eb-9c06-9fb235fa685d.png)

**Steps/Code to reproduce bug**
Enable CBO and decimal support. Run against SF=3k dataset.

**Expected behavior**
Joins should not fall back to CPU.

**Environment details (please complete the following information)**
Standalone.

**Additional context**

PR #2744, opened 2021-06-18
Cost-based optimizer: Implement simple cost model that demonstrates benefits with NDS queries

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/2718 and https://github.com/NVIDIA/spark-rapids/issues/2717

Changes in this PR:

- Implements a simple cost model by specifying default CPU and GPU operator costs, and implements special handling for ProjectExec and UnionExec
- Fixes a bug where writes would fall back to the CPU because the optimizer considered that AdaptiveSparkPlanExec was a CPU plan (#2718)
- Fixes a bug where joins were sometimes falling back to CPU (#2717)
- Added trace logging showing the costs calculated by the cost-based optimizer, and the actions it takes. This makes it possible to debug a query and see what the optimizer is doing.

Real-world benefits of this PR with NDS benchmarks:

- 20+ queries no longer fall back to CPU for writes
- q6: Avoids moving to the GPU for a trivial projection followed by a shuffle
- q11 & q74: Avoids moving to the GPU for a trivial projection following a CPU SortMergeJoin. The next operator moves to the GPU anyway, so this isn't of much benefit overall.
- q23a: Avoids moving to the GPU just for a union
- q58: Avoids moving to the GPU for a trivial projection between two CPU SortMergeJoins.

WIP because I can't test everything locally and need time on environments to do full runs at scale after code changes, to confirm no regressions. I also need to figure out how to write good tests for the current capabilites.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
@@ -21,8 +21,8 @@ import scala.collection.mutable.ListBuffer
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.expressions.{Alias, AttributeReference, Expression, GetStructField, WindowFrame, WindowSpecDefinition}
 import org.apache.spark.sql.catalyst.plans.{JoinType, LeftAnti, LeftSemi}
-import org.apache.spark.sql.execution.{GlobalLimitExec, LocalLimitExec, SparkPlan, TakeOrderedAndProjectExec, UnionExec}
-import org.apache.spark.sql.execution.adaptive.{CustomShuffleReaderExec, QueryStageExec}
+import org.apache.spark.sql.execution.{GlobalLimitExec, LocalLimitExec, ProjectExec, SparkPlan, TakeOrderedAndProjectExec, UnionExec}
+import org.apache.spark.sql.execution.adaptive.{AdaptiveSparkPlanExec, CustomShuffleReaderExec, QueryStageExec}
 import org.apache.spark.sql.execution.aggregate.HashAggregateExec
 import org.apache.spark.sql.execution.exchange.BroadcastExchangeExec
 import org.apache.spark.sql.execution.joins.{BroadcastHashJoinExec, BroadcastNestedLoopJoinExec, ShuffledHashJoinExec, SortMergeJoinExec}
@@ -67,7 +67,6 @@ class CostBasedOptimizer extends Optimizer with Logging {
     optimizations
   }
 
-
   /**
    * Walk the plan and determine CPU and GPU costs for each operator and then make decisions
    * about whether operators should run on CPU or GPU.
@@ -87,32 +86,38 @@ class CostBasedOptimizer extends Optimizer with Logging {
       optimizations: ListBuffer[Optimization],
       finalOperator: Boolean): (Double, Double) = {
 
-    // get the CPU and GPU cost of the child plan(s)
-    val childCosts = plan.childPlans
-        .map(child => recursivelyOptimize(
-          conf,
-          cpuCostModel,
-          gpuCostModel,
-          child,
-          optimizations,
-          finalOperator = false))
-
-    val (childCpuCosts, childGpuCosts) = childCosts.unzip
-
     // get the CPU and GPU cost of this operator (excluding cost of children)
     val operatorCpuCost = cpuCostModel.getCost(plan)
     val operatorGpuCost = gpuCostModel.getCost(plan)
 
+    showCosts(plan, "Operator costs", operatorCpuCost, operatorGpuCost)
+
+    // get the CPU and GPU cost of the child plan(s)
+    val childCosts = plan.childPlans
+      .map(child => recursivelyOptimize(
+        conf,
+        cpuCostModel,
+        gpuCostModel,
+        child,
+        optimizations,
+        finalOperator = false))
+
+    val (childCpuCosts, childGpuCosts) = childCosts.unzip
+
     // calculate total (this operator + children)
     val totalCpuCost = operatorCpuCost + childCpuCosts.sum
     var totalGpuCost = operatorGpuCost + childGpuCosts.sum
+
+    showCosts(plan, "Operator + child costs", totalCpuCost, totalGpuCost)
 
     plan.estimatedOutputRows = RowCountPlanVisitor.visit(plan)
 
     // determine how many transitions between CPU and GPU are taking place between
     // the child operators and this operator
     val numTransitions = plan.childPlans
-      .count(_.canThisBeReplaced != plan.canThisBeReplaced)
+      .count(canRunOnGpu(_) != canRunOnGpu(plan))
+
+    showCosts(plan, s"numTransitions=$numTransitions", totalCpuCost, totalGpuCost)
 
     if (numTransitions > 0) {
       // there are transitions between CPU and GPU so we need to calculate the transition costs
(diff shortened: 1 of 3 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -250,9 +290,21 @@ class CpuCostModel(conf: RapidsConf) extends CostModel {
     val rowCount = RowCountPlanVisitor.visit(plan).map(_.toDouble)
       .getOrElse(conf.defaultRowCount.toDouble)
 
-    val operatorCost = plan.conf
-      .getCpuOperatorCost(plan.wrapped.getClass.getSimpleName)
-      .getOrElse(conf.defaultCpuOperatorCost) * rowCount
+    val operatorCost = plan.wrapped match {
+      case _: ProjectExec =>
```
andygrove, line 294: I'm planning on removing the special handling here and implementing some costs in a data structure instead

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -158,39 +162,71 @@ class CostBasedOptimizer extends Optimizer with Logging {
...
       // reset the costs because this section of the plan was not moved to GPU
       totalGpuCost = totalCpuCost
+      showCosts(plan, s"Reset costs (not on GPU / exchange)", totalCpuCost, totalGpuCost)
     }
 
+    showCosts(plan, "END", totalCpuCost, totalGpuCost)
     (totalCpuCost, totalGpuCost)
   }
 
+  private def showCosts(
```
jlowe, line 205: Minor suggestion: `logCosts` may be more descriptive, as I initially assumed `showCosts` would unconditionally print.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -250,9 +286,21 @@ class CpuCostModel(conf: RapidsConf) extends CostModel {
...
+      case _: ProjectExec =>
+        // this is not accurate because CPU projections do have a cost due to appending values
+        // to each row that is produced, but this needs to be a really small number because
+        // GpuProject cost is zero (in our cost model) and we don't want to encourage moving to
+        // the GPU just to do a trivial projection, so we pretend the overhead of a
+        // CPU projection (beyond evaluating the expressions) is also zero
+        0
+      case _: UnionExec =>
+        // union does not further process data produced by its children
+        0
```
jlowe, line 299: Curious that these are not values pre-loaded into the plan that allow a user to override them just as they could override any other values?  As they are coded now, a user cannot change the costs of these operators via configs as they could other operators.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -299,9 +347,18 @@ class GpuCostModel(conf: RapidsConf) extends CostModel {
...
-      .getGpuOperatorCost(plan.wrapped.getClass.getSimpleName)
-      .getOrElse(conf.defaultGpuOperatorCost) * rowCount
+    val operatorCost = plan.wrapped match {
+      case _: ProjectExec =>
+        // The cost of a GPU projection is mostly the cost of evaluating the expressions
+        // to produce the projected columns
+        0
+      case _: UnionExec =>
+        // union does not further process data produced by its children
+        0
```
jlowe, line 357: Replicated code, worth refactoring into a common method so we don't forget to update one of the copies when yet another special-case exec gets added.  Or better yet, poke these values into the conf as defaults up front so there's no special cases to copy. 😺

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1647,6 +1662,7 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
   }
 
   private def getOptionalCost(key: String) = {
-    conf.get(key).map(toDouble(_, key))
+    // user-provided value takes precedence, then look in defaults map
+    conf.get(key).orElse(optimizerDefaults.get(key)).map(toDouble(_, key))
```
jlowe, line 1666: Curious, why are these configs being handled with a separate, special-case map rather than as just as other, internal configs?  Treating these like any other RapidsConf config key means we can get nice docstrings for them which could be helpful for generating documentation later if we decide to make them public.

andygrove: My reasoning was that we will at some point have a large number of these default costs that we wouldn't expect users to have to deal with, assuming we can come up with cost models that work well enough in a broad set of use cases. These are quite specialized settings and should probably be a separate section in the documentation, at least.

Perhaps it does make sense to add them as regular configs and we can use a filter to pull these into a separate documentation section when we make them public.

@revans2 Would be good to get your thoughts on this too.

andygrove: I pushed an update to use regular RapidsConf, while still allowing the user to provide costs even if we haven't defined a config entry for a particular operator or expression.

tests/src/test/scala/com/nvidia/spark/rapids/CostBasedOptimizerSuite.scala
```diff
@@ -57,20 +63,70 @@ class CostBasedOptimizerSuite extends SparkQueryCompareTestSuite
...
 
+  test("Avoid transition to GPU for trivial projection after CPU SMJ") {
+    logError("Avoid transition to GPU for trivial projection after CPU SMJ")
+
+    val conf = DEFAULT_CONF
+      .set(SQLConf.ADAPTIVE_EXECUTION_ENABLED.key, "true")
+      .set(SQLConf.AUTO_BROADCASTJOIN_THRESHOLD.key, "-1")
+      .set(RapidsConf.ENABLE_CAST_STRING_TO_TIMESTAMP.key, "false")
+      .set(RapidsConf.ENABLE_REPLACE_SORTMERGEJOIN.key, "false")
+      .set(RapidsConf.TEST_ALLOWED_NONGPU.key,
```
jlowe, line 74: This mutates `DEFAULT_CONF` and risks leaking test-specific settings into other tests that use `DEFAULT_CONF`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1647,6 +1687,14 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
   }
 
   private def getOptionalCost(key: String) = {
-    conf.get(key).map(toDouble(_, key))
+    // this is a bit convoluted, but the flow is to see if we have a
+    // defined configuration setting (potentially with a default value)
+    // and if not, then we look for the key in the configuration values
+    // anyway so that we can provide costs for any operator or expression,
+    // even if there is no well-defined configuration yet
+    OPTIMIZER_COSTS.get(key)
```
jlowe, line 1695: If we're going to have to always special-case these settings to handle the defaults properly then I'd rather go back to the prior design.  The problem with this new approach is the need to maintain the conf in two places, the RapidsConf entry and the OPTIMIZER_COSTS list.  If someone forgets to update the latter then it will silently not honor the custom default which won't be fun to track down.

tests/src/test/scala/com/nvidia/spark/rapids/CostBasedOptimizerSuite.scala
andygrove: `DEFAULT_CONF` is a `def` not a `val` so a new instance of `SQLConf` is created for each test. I can change this to a method to make this clearer though.

jlowe: oh my apologies for misreading it.  Having the name all caps does make it look like a `val`.

jlowe (approved):

jlowe:
build

Issue #2723, opened 2021-06-16
[BUG] test failures in CI build (observed in UCX job) after starting to use 21.08


abellina:
These tests failed:

- test_broadcast_nested_loop_join
- test_broadcast_nested_loop_join_special_case_count
- test_cartesean_join
- test_cartesean_join_special_case_count
- test_cartesean_join_special_case_group_by

With basic types as well as nested types. The job was changed for last night's run to use 21.08. I had updated it to use UCX 1.11 native libraries and ran successfully earlier.

---

UPDATE:

https://github.com/NVIDIA/spark-rapids/pull/2749 was PRed to revert a change introduced in https://github.com/NVIDIA/spark-rapids/pull/2702 that recently enabled these types in the plugin and the integration tests that failed above. 

The condition for this issue (#2723) to be closed is that @revans2's code should be re-added and tests should pass.

abellina:
I am not sure if this is UCX specific or not.

abellina:
Here's one example of an exception:
```
../../src/main/python/join_test.py::test_broadcast_nested_loop_join[100-String][IGNORE_ORDER({'local': True})] 21/06/16 09:18:56 WARN TaskSetManager: Lost task 1.0 in stage 28146.0 (TID 1147491) (10.136.6.4 executor 3): ai.rapids.cudf.CudaException: an illegal memory access was encountered
  at ai.rapids.cudf.Cuda.memcpyOnStream(Native Method)
  at ai.rapids.cudf.Cuda.memcpy(Cuda.java:472)
  at ai.rapids.cudf.Cuda.memcpy(Cuda.java:288)
  at ai.rapids.cudf.BaseDeviceMemoryBuffer.copyFromHostBuffer(BaseDeviceMemoryBuffer.java:43)
  at ai.rapids.cudf.BaseDeviceMemoryBuffer.copyFromHostBuffer(BaseDeviceMemoryBuffer.java:105)
  at ai.rapids.cudf.HostColumnVector.copyToDevice(HostColumnVector.java:198)
  at ai.rapids.cudf.HostColumnVector$ColumnBuilder.buildAndPutOnDevice(HostColumnVector.java:1268)
  at com.nvidia.spark.rapids.GpuColumnVector$GpuColumnarBatchBuilder.buildAndPutOnDevice(GpuColumnVector.java:400)
[...]

PR #2748, opened 2021-06-18
Revert "Add in support for arrays in BroadcastNestedLoopJoinExec and CartesianProductExec"

abellina:
Reverts NVIDIA/spark-rapids#2702

Due to https://github.com/NVIDIA/spark-rapids/issues/2723, we'd like to revert this change while we investigate a solution.


abellina:
I am going to do this revert manually, since I need to sign-off the revert commit.

PR #2749, opened 2021-06-18
Revert "Add in support for lists in some joins (#2702)"

abellina:
This reverts commit 44a48ca08c1eecf6280fa0d8b83fdc55b0d48e6c. 

Due to #2723, we'd like to revert this change while we investigate a solution.

### DIFF
--- a/docs/supported_ops.md
+++ b/docs/supported_ops.md
@@ -650,7 +650,7 @@ Accelerator supports are described below.
 <td>S</td>
 <td><b>NS</b></td>
 <td><b>NS</b></td>
-<td><em>PS* (missing nested BINARY, CALENDAR, ARRAY, MAP, STRUCT, UDT)</em></td>
+<td><b>NS</b></td>
 <td><b>NS</b></td>
 <td><b>NS</b></td>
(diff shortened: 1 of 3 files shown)


abellina:
build

jlowe (commented):
Revert looks OK, but is there a followup issue against 21.08 to track restoring this functionality?

abellina:
#2723 is the issue tracking it from our side. I was thinking the cudf issue can be filed once we have a repro.

jlowe:
> #2723 is the issue tracking it from our side.

I saw #2723 mostly as tracking the original bug which doesn't necessarily require restoring this revert in order to declare it "fixed," especially if it ends up just being delegated to a cudf issue.  I'm OK with #2723 being the tracking issue as long as it's clear in that issue that it must not be considered closed until the functionality being reverted here is restored.

abellina:
Ok I see what you mean. I'll make it clear in the issue to re-add Bobby's code. It seems like it is the passing criteria for #2723, that the added types and tests pass.

jlowe (approved):

revans2 (approved):

PR #2761, opened 2021-06-21
Update dot graph to include stages and remove some duplication

### LINKED ISSUE #2711: [BUG] reused stages should not appear multiple times in dot
I was looking at some dot graphs, for things like NDS query 14a and I noticed that the reused stages, appeared multiple times in the dot graph.  This can make it difficult to see what is really happening, because I cannot tell what was, and what wasn't reused.  I know that this can also make it difficult to look at for some complicated queries and I would be okay with an enable/disable option, but I think the default should be to show each stage once.

### LINKED ISSUE #2712: [FEA] Display job and stage info in the dot graph for profiling tool
I was looking at the dot output from the profiling tool for some NDS queries and thought it would be really great if I could tell what parts of the graph correspond to a given stage/job along with the ID for that stage/job and possibly some rolled up metrics for them as well.

It feels like you could use a subgraph for each stage/job to have dot draw a box around it and group it. Then at a minimum label it with the stage/job ID and ideally some metrics too (like timing would be great, but so would other metrics).

revans2:
This uses a modified version of the code that Apache Spark uses to generate its SQL visualization.  I updated it to match the style that we currently are using.

This fixes #2711
This fixes #2712

It does not add in job information it could do it, but it is not that critical.

It is not perfect with how it dedupes stages/etc but it is as good as Apache Spark is, because it reuses the code.

This rips out the compare code, that was not being used, and I plan to work on getting a stage to stage comparison setup next so we can more easily see which stages, in large queries both between the CPU and the GPU, and equivalent.  That way I can more quickly see what is showing speedups or not.


revans2:
build

revans2:
build

revans2:
Something odd happened with the previous CI job. It skipped everything and I don't know why.

tgravescs:
@andygrove can you take a look

andygrove (dismissed):
LGTM. I tested locally with some TPC-DS queries and it worked well. It's great having the stage metrics on the diagram.

tgravescs (commented):
overall looks fine to me, can we add more to the GenerateDotSuite to make sure the job/stage info is present?

revans2:
@tgravescs I updated the test and I also added in some more docs to the README to explain how the stages work.

revans2:
build

tgravescs (dismissed):
looks good, can you upmerge to fix the conflict

revans2:
build

andygrove (approved):

PR #2768, opened 2021-06-21
Make window function test results deterministic.

### LINKED ISSUE #2546: [BUG] Build Failure when building from source
Hey, 

I am trying to build the main branch without any modifications to the code using

``mvn verify``

but unfortunately this fails due to this error

``[ERROR] Failed to execute goal on project rapids-4-spark-shims-spark312_2.12: Could not resolve dependencies for project com.nvidia:rapids-4-spark-shims-spark312_2.12:jar:0.5.0: org.apache.spark:spark-sql_2.12:jar:3.1.2-SNAPSHOT was not found in https://oss.sonatype.org/content/repositories/snapshots during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of snapshots-repo has elapsed or updates are forced -> [``

Is there any simple way to fix this?

Thanks!

mythrocks:
Fixes #2546.

The tests under `WindowFunctionSuite` have a couple of avenues for non-determinism:
1. The window spec definitions specify a single order-by column (`dateLong`). Some rows have the same value for `dateLong`. This causes window aggregations to diverge, on occasion. This is reproducible on Pascal (`P100`).
2. The overall output isn't sorted before comparison.

In this commit, `ORDER BY` clauses have been added to both the window-specifications and the output projections, to ensure determinism.

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/WindowFunctionSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/WindowFunctionSuite.scala
@@ -26,20 +26,21 @@ class WindowFunctionSuite extends SparkQueryCompareTestSuite {
 
   def windowAggregationTester(windowSpec: WindowSpec): DataFrame => DataFrame =
     (df : DataFrame) => df.select(
-      sum("dollars").over(windowSpec),
-      min("dollars").over(windowSpec),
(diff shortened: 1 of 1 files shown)


jlowe (approved):

jlowe:
build

gerashegalov (approved):

PR #2792, opened 2021-06-23
Automatically set spark.task.maxFailures and local[*, maxFailures]

### LINKED ISSUE #2772: [BUG] new integration test failures w/ maxFailures=1
**Describe the bug**
related to maxFailures change https://github.com/NVIDIA/spark-rapids/pull/2743,  and should be the sequel to https://github.com/NVIDIA/spark-rapids/issues/2698

intermittently failed ExecutorLostFailure, not-always reproducible. 
Failed Tests (plz add if you see others):
1. join_test.py::test_cartesean_join[100-Array(Null)0][IGNORE_ORDER({'local': True})]
2. window_function_test.py::test_window_aggs_for_rows_lead_lag_on_arrays[partBy:Long-orderBy:Long-orderBy:LongRange(not_null)-agg:Array(Long)-1000][IGNORE_ORDER({'local': True})]

sample executor log,
```
21/06/22 07:29:51 INFO CoarseGrainedExecutorBackend: Got assigned task 331736
21/06/22 07:29:51 INFO CoarseGrainedExecutorBackend: Got assigned task 331737
21/06/22 07:29:51 INFO Executor: Running task 0.0 in stage 44655.0 (TID 331736)
21/06/22 07:29:51 INFO CoarseGrainedExecutorBackend: Got assigned task 331738
21/06/22 07:29:51 INFO Executor: Running task 1.0 in stage 44655.0 (TID 331737)
21/06/22 07:29:51 INFO Executor: Running task 2.0 in stage 44655.0 (TID 331738)
21/06/22 07:29:51 INFO CoarseGrainedExecutorBackend: Got assigned task 331739
21/06/22 07:29:51 INFO Executor: Running task 3.0 in stage 44655.0 (TID 331739)
21/06/22 07:29:51 INFO CoarseGrainedExecutorBackend: Got assigned task 331740
21/06/22 07:29:51 INFO Executor: Running task 5.0 in stage 44655.0 (TID 331740)
21/06/22 07:29:51 INFO CoarseGrainedExecutorBackend: Got assigned task 331741
21/06/22 07:29:51 INFO Executor: Running task 7.0 in stage 44655.0 (TID 331741)
21/06/22 07:29:51 INFO MapOutputTrackerWorker: Updating epoch to 14283 and clearing cache
21/06/22 07:29:51 INFO TorrentBroadcast: Started reading broadcast variable 44525 with 1 pieces (estimated total size 4.0 MiB)
21/06/22 07:29:51 INFO MemoryStore: Block broadcast_44525_piece0 stored as bytes in memory (estimated size 18.6 KiB, free 6.2 GiB)
21/06/22 07:29:51 INFO TorrentBroadcast: Reading broadcast variable 44525 took 3 ms
21/06/22 07:29:51 INFO MemoryStore: Block broadcast_44525 stored as values in memory (estimated size 40.3 KiB, free 6.2 GiB)
21/06/22 07:29:51 INFO MapOutputTrackerWorker: Don't have map outputs for shuffle 14281, fetching them
21/06/22 07:29:51 INFO MapOutputTrackerWorker: Doing the fetch; tracker endpoint = NettyRpcEndpointRef(spark://MapOutputTracker@it-ub16-302-14-z17mn-87d0q:46005)
21/06/22 07:29:51 INFO MapOutputTrackerWorker: Don't have map outputs for shuffle 14281, fetching them
21/06/22 07:29:51 INFO MapOutputTrackerWorker: Don't have map outputs for shuffle 14281, fetching them
21/06/22 07:29:51 INFO MapOutputTrackerWorker: Don't have map outputs for shuffle 14281, fetching them
21/06/22 07:29:51 INFO MapOutputTrackerWorker: Don't have map outputs for shuffle 14281, fetching them
21/06/22 07:29:51 INFO MapOutputTrackerWorker: Don't have map outputs for shuffle 14281, fetching them
21/06/22 07:29:51 INFO MapOutputTrackerWorker: Got the output locations
21/06/22 07:29:51 INFO ShuffleBlockFetcherIterator: Getting 6 (44.2 KiB) non-empty blocks including 6 (44.2 KiB) local and 0 (0.0 B) host-local and 0 (0.0 B) remote blocks
21/06/22 07:29:51 INFO ShuffleBlockFetcherIterator: Getting 6 (18.7 KiB) non-empty blocks including 6 (18.7 KiB) local and 0 (0.0 B) host-local and 0 (0.0 B) remote blocks
21/06/22 07:29:51 INFO ShuffleBlockFetcherIterator: Started 0 remote fetches in 0 ms
21/06/22 07:29:51 INFO ShuffleBlockFetcherIterator: Started 0 remote fetches in 0 ms
21/06/22 07:29:51 INFO ShuffleBlockFetcherIterator: Getting 6 (44.3 KiB) non-empty blocks including 6 (44.3 KiB) local and 0 (0.0 B) host-local and 0 (0.0 B) remote blocks
21/06/22 07:29:51 INFO ShuffleBlockFetcherIterator: Getting 6 (34.3 KiB) non-empty blocks including 6 (34.3 KiB) local and 0 (0.0 B) host-local and 0 (0.0 B) remote blocks
21/06/22 07:29:51 INFO ShuffleBlockFetcherIterator: Getting 6 (33.3 KiB) non-empty blocks including 6 (33.3 KiB) local and 0 (0.0 B) host-local and 0 (0.0 B) remote blocks
21/06/22 07:29:51 INFO ShuffleBlockFetcherIterator: Started 0 remote fetches in 0 ms
21/06/22 07:29:51 INFO ShuffleBlockFetcherIterator: Getting 6 (47.9 KiB) non-empty blocks including 6 (47.9 KiB) local and 0 (0.0 B) host-local and 0 (0.0 B) remote blocks
21/06/22 07:29:51 INFO ShuffleBlockFetcherIterator: Started 0 remote fetches in 0 ms
21/06/22 07:29:51 INFO ShuffleBlockFetcherIterator: Started 0 remote fetches in 0 ms
21/06/22 07:29:51 INFO ShuffleBlockFetcherIterator: Started 0 remote fetches in 0 ms
21/06/22 07:29:51 INFO CodeGenerator: Code generated in 13.560419 ms
21/06/22 07:29:51 INFO CodeGenerator: Code generated in 13.591383 ms
21/06/22 07:29:51 ERROR CoarseGrainedExecutorBackend: RECEIVED SIGNAL
```

driver log,
```
 21/06/22 07:29:51 WARN HeartbeatReceiver: Removing executor 1 with no recent heartbeats: 151811 ms exceeds timeout 120000 ms
15:28:55  21/06/22 07:29:51 ERROR TaskSchedulerImpl: Lost executor 1 on 10.233.122.131: Executor heartbeat timed out after 151811 ms
15:28:55  21/06/22 07:29:51 WARN TaskSetManager: Lost task 0.0 in stage 44655.0 (TID 331736, 10.233.122.131, executor 1): ExecutorLostFailure (executor 1 exited caused by one of the running tasks) Reason: Executor heartbeat timed out after 151811 ms
15:28:55  21/06/22 07:29:51 ERROR TaskSetManager: Task 0 in stage 44655.0 failed 1 times; aborting job
15:28:55  21/06/22 07:29:51 WARN TaskSetManager: Lost task 3.0 in stage 44655.0 (TID 331739, 10.233.122.131, executor 1): ExecutorLostFailure (executor 1 exited caused by one of the running tasks) Reason: Executor heartbeat timed out after 151811 ms
15:28:55  21/06/22 07:29:51 WARN TaskSetManager: Lost task 7.0 in stage 44655.0 (TID 331741, 10.233.122.131, executor 1): ExecutorLostFailure (executor 1 exited caused by one of the running tasks) Reason: Executor heartbeat timed out after 151811 ms
15:28:55  21/06/22 07:29:51 WARN TaskSetManager: Lost task 2.0 in stage 44655.0 (TID 331738, 10.233.122.131, executor 1): ExecutorLostFailure (executor 1 exited caused by one of the running tasks) Reason: Executor heartbeat timed out after 151811 ms
15:28:55  21/06/22 07:29:51 WARN TaskSetManager: Lost task 1.0 in stage 44655.0 (TID 331737, 10.233.122.131, executor 1): ExecutorLostFailure (executor 1 exited caused by one of the running tasks) Reason: Executor heartbeat timed out after 151811 ms
15:28:55  21/06/22 07:29:51 WARN TaskSetManager: Lost task 5.0 in stage 44655.0 (TID 331740, 10.233.122.131, executor 1): ExecutorLostFailure (executor 1 exited caused by one of the running tasks) Reason: Executor heartbeat timed out after 151811 ms

-------------------------------------
15:34:49                  if answer[1] == REFERENCE_TYPE:
15:34:49  >                   raise Py4JJavaError(
15:34:49                          "An error occurred while calling {0}{1}{2}.\n".
15:34:49                          format(target_id, ".", name), value)
15:34:49  E                   py4j.protocol.Py4JJavaError: An error occurred while calling o1160474.collectToPython.
15:34:49  E                   : org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 44655.0 failed 1 times, most recent failure: Lost task 0.0 in stage 44655.0 (TID 331736, 10.233.122.131, executor 1): ExecutorLostFailure (executor 1 exited caused by one of the running tasks) Reason: Executor heartbeat timed out after 151811 ms
15:34:49  E                   Driver stacktrace:
15:34:49  E                   	at org.apache.spark.scheduler.DAGScheduler.failJobAndIndependentStages(DAGScheduler.scala:2059)
15:34:49  E                   	at org.apache.spark.scheduler.DAGScheduler.$anonfun$abortStage$2(DAGScheduler.scala:2008)
15:34:49  E                   	at org.apache.spark.scheduler.DAGScheduler.$anonfun$abortStage$2$adapted(DAGScheduler.scala:2007)
15:34:49  E                   	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
15:34:49  E                   	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
15:34:49  E                   	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
15:34:49  E                   	at org.apache.spark.scheduler.DAGScheduler.abortStage(DAGScheduler.scala:2007)
15:34:49  E                   	at org.apache.spark.scheduler.DAGScheduler.$anonfun$handleTaskSetFailed$1(DAGScheduler.scala:973)
15:34:49  E                   	at org.apache.spark.scheduler.DAGScheduler.$anonfun$handleTaskSetFailed$1$adapted(DAGScheduler.scala:973)
15:34:49  E                   	at scala.Option.foreach(Option.scala:407)
15:34:49  E                   	at org.apache.spark.scheduler.DAGScheduler.handleTaskSetFailed(DAGScheduler.scala:973)
15:34:49  E                   	at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.doOnReceive(DAGScheduler.scala:2239)
15:34:49  E                   	at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.onReceive(DAGScheduler.scala:2188)
15:34:49  E                   	at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.onReceive(DAGScheduler.scala:2177)
15:34:49  E                   	at org.apache.spark.util.EventLoop$$anon$1.run(EventLoop.scala:49)
15:34:49  E                   	at org.apache.spark.scheduler.DAGScheduler.runJob(DAGScheduler.scala:775)
15:34:49  E                   	at org.apache.spark.SparkContext.runJob(SparkContext.scala:2114)
15:34:49  E                   	at org.apache.spark.SparkContext.runJob(SparkContext.scala:2135)
15:34:49  E                   	at org.apache.spark.SparkContext.runJob(SparkContext.scala:2154)
15:34:49  E                   	at org.apache.spark.SparkContext.runJob(SparkContext.scala:2179)
15:34:49  E                   	at org.apache.spark.rdd.RDD.$anonfun$collect$1(RDD.scala:1004)
15:34:49  E                   	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
15:34:49  E                   	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:112)
15:34:49  E                   	at org.apache.spark.rdd.RDD.withScope(RDD.scala:388)
15:34:49  E                   	at org.apache.spark.rdd.RDD.collect(RDD.scala:1003)
15:34:49  E                   	at org.apache.spark.sql.execution.SparkPlan.executeCollect(SparkPlan.scala:390)
15:34:49  E                   	at org.apache.spark.sql.Dataset.$anonfun$collectToPython$1(Dataset.scala:3450)
15:34:49  E                   	at org.apache.spark.sql.Dataset.$anonfun$withAction$1(Dataset.scala:3618)
15:34:49  E                   	at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId$5(SQLExecution.scala:100)
15:34:49  E                   	at org.apache.spark.sql.execution.SQLExecution$.withSQLConfPropagated(SQLExecution.scala:160)
15:34:49  E                   	at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId$1(SQLExecution.scala:87)
15:34:49  E                   	at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:764)
15:34:49  E                   	at org.apache.spark.sql.execution.SQLExecution$.withNewExecutionId(SQLExecution.scala:64)
15:34:49  E                   	at org.apache.spark.sql.Dataset.withAction(Dataset.scala:3616)
15:34:49  E                   	at org.apache.spark.sql.Dataset.collectToPython(Dataset.scala:3447)
15:34:49  E                   	at sun.reflect.GeneratedMethodAccessor85.invoke(Unknown Source)
15:34:49  E                   	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
15:34:49  E                   	at java.lang.reflect.Method.invoke(Method.java:498)
15:34:49  E                   	at py4j.reflection.MethodInvoker.invoke(MethodInvoker.java:244)
15:34:49  E                   	at py4j.reflection.ReflectionEngine.invoke(ReflectionEngine.java:357)
15:34:49  E                   	at py4j.Gateway.invoke(Gateway.java:282)
15:34:49  E                   	at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:132)
15:34:49  E                   	at py4j.commands.CallCommand.execute(CallCommand.java:79)
15:34:49  E                   	at py4j.GatewayConnection.run(GatewayConnection.java:238)
15:34:49  E                   	at java.lang.Thread.run(Thread.java:748)
```

abellina:
…task failures

Closes #2772 

This should alleviate issues with older Spark's by making `spark.task.maxFailures` automatically to 1 for Spark 3.1.1+ and to 4 for prior versions of Spark.

If the user sets `spark.master` or specifies `--master` in the `$SPARK_SUBMIT_FLAGS`, it is assumed they know what they are doing, and so the local mode workaround for https://issues.apache.org/jira/browse/SPARK-2083 is not followed. That said, it still sets `spark.task.maxFailures`.

### DIFF
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -122,7 +122,7 @@ else
     # Spark versions before 3.1.1, this sets the spark.max.taskFailures to 4 to allow for
     # more lineant configuration, else it will set them to 1 as spurious task failures are not expected
     # for Spark 3.1.1+
-    VERSION_STRING=`$SPARK_HOME/bin/pyspark --version 2>&1|awk '/version\ [0-9.]+$/{print $NF}'`
+    VERSION_STRING=`$SPARK_HOME/bin/pyspark --version 2>&1|grep -v Scala|awk '/version\ [0-9.]+/{print $NF}'`
     [[ -z $VERSION_STRING ]] && { echo "Unable to detect the Spark version at $SPARK_HOME"; exit 1; }
     echo "Detected Spark version $VERSION_STRING"
 
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -122,13 +122,11 @@ else
     # Spark versions before 3.1.1, this sets the spark.max.taskFailures to 4 to allow for
     # more lineant configuration, else it will set them to 1 as spurious task failures are not expected
     # for Spark 3.1.1+
-    VERSION_STRING=`$SPARK_HOME/bin/pyspark --version 2>&1|grep version\ |grep -v Scala|awk '{print $5}'`
+    VERSION_STRING=`unzip -q -c $SPARK_HOME/jars/spark-core_*.jar spark-version-info.properties | head -1 | cut -d'=' -f2`
     echo "Detected Spark version is $VERSION_STRING"
     IS_BEFORE_SPARK_311=`python -c "print(1 if '$VERSION_STRING' < '3.1.1' else 0)"`
     SPARK_TASK_MAXFAILURES=1
-    if (( $IS_BEFORE_SPARK_311 )); then
-      SPARK_TASK_MAXFAILURES=4
-    fi
+    [[ "$VERSION_STRING" < "3.1.1" ]] && SPARK_TASK_MAXFAILURES=4
 
     export PYSP_TEST_spark_driver_extraClassPath="${ALL_JARS// /:}"
     export PYSP_TEST_spark_executor_extraClassPath="${ALL_JARS// /:}"
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -122,7 +122,7 @@ else
     # Spark versions before 3.1.1, this sets the spark.max.taskFailures to 4 to allow for
     # more lineant configuration, else it will set them to 1 as spurious task failures are not expected
     # for Spark 3.1.1+
-    VERSION_STRING=`unzip -q -c $SPARK_HOME/jars/spark-core_*.jar spark-version-info.properties | grep ^version= | cut -d'=' -f2`
+    VERSION_STRING=`$SPARK_HOME/bin/pyspark --version 2>&1|awk '/version\ [0-9.]+$/{print $NF}'`
     [[ -z $VERSION_STRING ]] && { echo "Unable to detect the Spark version at $SPARK_HOME"; exit 1; }
     echo "Detected Spark version $VERSION_STRING"
 
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -117,6 +117,11 @@ else
     NUM_LOCAL_EXECS=${NUM_LOCAL_EXECS:-0}
     MB_PER_EXEC=${MB_PER_EXEC:-1024}
     CORES_PER_EXEC=${CORES_PER_EXEC:-1}
+    # Spark 3.1.1 includes https://github.com/apache/spark/pull/31540
+    # which helps with spurious task failures in the tests. If you are running
+    # older versions of Spark, and experience #31540, please set this environment
+    # variable higher (Spark has 4 by default).
+    SPARK_TASK_MAXFAILURES=${SPARK_TASK_MAXFAILURES:-1}
 
     if ((NUM_LOCAL_EXECS > 0)); then
       export PYSP_TEST_spark_master="local-cluster[$NUM_LOCAL_EXECS,$CORES_PER_EXEC,$MB_PER_EXEC]"
@@ -129,9 +134,9 @@ else
     export PYSP_TEST_spark_ui_showConsoleProgress='false'
     export PYSP_TEST_spark_sql_session_timeZone='UTC'
     export PYSP_TEST_spark_sql_shuffle_partitions='12'
-    # prevent cluster shape to change - and fail quicker rather than retry
-    export PYSP_TEST_spark_task_maxFailures='1'
+    # prevent cluster shape to change
     export PYSP_TEST_spark_dynamicAllocation_enabled='false'
+    export PYSP_TEST_spark_task_maxFailures="${SPARK_TASK_MAXFAILURES}"
     if ((${#TEST_PARALLEL_OPTS[@]} > 0));
     then
         export PYSP_TEST_spark_rapids_memory_gpu_allocFraction=$MEMORY_FRACTION
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -118,13 +118,16 @@ else
     MB_PER_EXEC=${MB_PER_EXEC:-1024}
     CORES_PER_EXEC=${CORES_PER_EXEC:-1}
     # Spark 3.1.1 includes https://github.com/apache/spark/pull/31540
-    # which helps with spurious task failures in the tests. If you are running
-    # older versions of Spark, and experience #31540, please set this environment
-    # variable higher (Spark has 4 by default).
-    SPARK_TASK_MAXFAILURES=${SPARK_TASK_MAXFAILURES:-1}
-
-    if ((NUM_LOCAL_EXECS > 0)); then
-      export PYSP_TEST_spark_master="local-cluster[$NUM_LOCAL_EXECS,$CORES_PER_EXEC,$MB_PER_EXEC]"
+    # which helps with spurious task failures as observed in our tests. If you are running
+    # Spark versions before 3.1.1, this sets the spark.max.taskFailures to 4 to allow for
+    # more lineant configuration, else it will set them to 1 as spurious task failures are not expected
+    # for Spark 3.1.1+
+    VERSION_STRING=`$SPARK_HOME/bin/pyspark --version 2>&1|grep version\ |grep -v Scala|awk '{print $5}'`
+    echo "Detected Spark version is $VERSION_STRING"
+    IS_BEFORE_SPARK_311=`python -c "print(1 if '$VERSION_STRING' < '3.1.1' else 0)"`
+    SPARK_TASK_MAXFAILURES=1
+    if (( $IS_BEFORE_SPARK_311 )); then
+      SPARK_TASK_MAXFAILURES=4
     fi
 
     export PYSP_TEST_spark_driver_extraClassPath="${ALL_JARS// /:}"
@@ -136,7 +139,24 @@ else
     export PYSP_TEST_spark_sql_shuffle_partitions='12'
     # prevent cluster shape to change
     export PYSP_TEST_spark_dynamicAllocation_enabled='false'
-    export PYSP_TEST_spark_task_maxFailures="${SPARK_TASK_MAXFAILURES}"
+
+    # Set spark.task.maxFailures for most schedulers.
+    #
+    # Local (non-cluster) mode is the exception and does not work with `spark.task.maxFailures`.
+    # It requires two arguments to the master specification "local[N, K]" where
+    # N is the number of threads, and K is the maxFailures (otherwise this is hardcoded to 1,
+    # see https://issues.apache.org/jira/browse/SPARK-2083).
+    export PYSP_TEST_spark_task_maxFailures="$SPARK_TASK_MAXFAILURES"
+
+    if ((NUM_LOCAL_EXECS > 0)); then
+      export PYSP_TEST_spark_master="local-cluster[$NUM_LOCAL_EXECS,$CORES_PER_EXEC,$MB_PER_EXEC]"
+    else
+      # If a master is not specified, use "local[*, $SPARK_TASK_MAXFAILURES]"
+      if [ -z "${PYSP_TEST_spark_master}" ] && [ "$SPARK_SUBMIT_FLAGS" != *"--master"* ]; then
+        export PYSP_TEST_spark_master="local[*,$SPARK_TASK_MAXFAILURES]"
+      fi
+    fi
+
     if ((${#TEST_PARALLEL_OPTS[@]} > 0));
     then
         export PYSP_TEST_spark_rapids_memory_gpu_allocFraction=$MEMORY_FRACTION
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -122,16 +122,9 @@ else
     # Spark versions before 3.1.1, this sets the spark.max.taskFailures to 4 to allow for
     # more lineant configuration, else it will set them to 1 as spurious task failures are not expected
     # for Spark 3.1.1+
-    set +e
-    VERSION_PROPERTY=`unzip -q -c $SPARK_HOME/jars/spark-core_*.jar spark-version-info.properties | grep ^version=`
-    if [[ $? -eq "0" ]]; then
-      VERSION_STRING=`echo "$VERSION_PROPERTY" | cut -d'=' -f2`
-      echo "Detected Spark version is $VERSION_STRING"
-    else
-      echo "Unable to detect the Spark version from the distribution in $SPARK_HOME"
-      exit 1
-    fi
-    set -e
+    VERSION_STRING=`unzip -q -c $SPARK_HOME/jars/spark-core_*.jar spark-version-info.properties | grep ^version= | cut -d'=' -f2`
+    [[ -z $VERSION_STRING ]] && { echo "Unable to detect the Spark version at $SPARK_HOME"; exit 1; }
+    echo "Detected Spark version $VERSION_STRING"
 
     SPARK_TASK_MAXFAILURES=1
     [[ "$VERSION_STRING" < "3.1.1" ]] && SPARK_TASK_MAXFAILURES=4
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -122,8 +122,17 @@ else
     # Spark versions before 3.1.1, this sets the spark.max.taskFailures to 4 to allow for
     # more lineant configuration, else it will set them to 1 as spurious task failures are not expected
     # for Spark 3.1.1+
-    VERSION_STRING=`unzip -q -c $SPARK_HOME/jars/spark-core_*.jar spark-version-info.properties | head -1 | cut -d'=' -f2`
-    echo "Detected Spark version is $VERSION_STRING"
+    set +e
+    VERSION_PROPERTY=`unzip -q -c $SPARK_HOME/jars/spark-core_*.jar spark-version-info.properties | grep ^version=`
+    if [[ $? -eq "0" ]]; then
+      VERSION_STRING=`echo "$VERSION_PROPERTY" | cut -d'=' -f2`
+      echo "Detected Spark version is $VERSION_STRING"
+    else
+      echo "Unable to detect the Spark version from the distribution in $SPARK_HOME"
+      exit 1
+    fi
+    set -e
+
     SPARK_TASK_MAXFAILURES=1
     [[ "$VERSION_STRING" < "3.1.1" ]] && SPARK_TASK_MAXFAILURES=4
 
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -124,7 +124,6 @@ else
     # for Spark 3.1.1+
     VERSION_STRING=`unzip -q -c $SPARK_HOME/jars/spark-core_*.jar spark-version-info.properties | head -1 | cut -d'=' -f2`
     echo "Detected Spark version is $VERSION_STRING"
-    IS_BEFORE_SPARK_311=`python -c "print(1 if '$VERSION_STRING' < '3.1.1' else 0)"`
     SPARK_TASK_MAXFAILURES=1
     [[ "$VERSION_STRING" < "3.1.1" ]] && SPARK_TASK_MAXFAILURES=4
 
--- a/jenkins/spark-tests.sh
+++ b/jenkins/spark-tests.sh
@@ -66,7 +66,6 @@ BASE_SPARK_SUBMIT_ARGS="$BASE_SPARK_SUBMIT_ARGS \
     --executor-memory 12G \
     --total-executor-cores 6 \
     --conf spark.sql.shuffle.partitions=12 \
-    --conf spark.task.maxFailures=1 \
     --conf spark.dynamicAllocation.enabled=false \
     --conf spark.driver.extraClassPath=${CUDF_JAR}:${RAPIDS_PLUGIN_JAR}:${RAPIDS_UDF_JAR} \
     --conf spark.executor.extraClassPath=${CUDF_JAR}:${RAPIDS_PLUGIN_JAR}:${RAPIDS_UDF_JAR} \
--- a/jenkins/spark-tests.sh
+++ b/jenkins/spark-tests.sh
@@ -60,12 +60,18 @@ export PATH="$SPARK_HOME/bin:$SPARK_HOME/sbin:$PATH"
 tar zxf $SPARK_HOME.tgz -C $ARTF_ROOT && \
     rm -f $SPARK_HOME.tgz
 
+IS_SPARK_311_OR_LATER=0
+[[ "$(printf '%s\n' "3.1.1" "$SPARK_VER" | sort -V | head -n1)" = "3.1.1" ]] && IS_SPARK_311_OR_LATER=1
+
+SPARK_TASK_MAXFAILURES=1
+[[ "$IS_SPARK_311_OR_LATER" -eq "0" ]] && SPARK_TASK_MAXFAILURES=4
 
 BASE_SPARK_SUBMIT_ARGS="$BASE_SPARK_SUBMIT_ARGS \
     --master spark://$HOSTNAME:7077 \
     --executor-memory 12G \
     --total-executor-cores 6 \
     --conf spark.sql.shuffle.partitions=12 \
+    --conf spark.task.maxFailures=$SPARK_TASK_MAXFAILURES \
     --conf spark.dynamicAllocation.enabled=false \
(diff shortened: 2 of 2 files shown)


revans2 (approved):
This look OK, but if we really want to catch these things should we be looking at the version of spark being used and adjust the default automatically?

abellina:
> This look OK, but if we really want to catch these things should we be looking at the version of spark being used and adjust the default automatically?

I tried doing this, but Spark complained with: `Cannot modify the value of a Spark config`. I also tried setting it via the `SparkContext` but it didn't seem to work either (I still kept getting the 4 default failures). It is very likely I missed something.

On the spark version, I can parse it from the `pyspark --version` command, but it is hardly clean (we get a bunch of ascii art on the way, unfortunately). Another option might be to launch an app to find the version. I figured this was both still strict, yet allowed some jobs to be more leniant.

If you think this can be possible I am happy to investigate more.

revans2:
I meant to do this in the shell script, not in python.

abellina:
@revans2 I implemented automatic discovery here: https://github.com/NVIDIA/spark-rapids/pull/2792/commits/d73f1c9759738361ae6dc8ff1e2fc7a58a28ea0c. Ready for another look.

abellina:
build

integration_tests/run_pyspark_from_build.sh
```diff
@@ -117,9 +117,17 @@ else
...
-
-    if ((NUM_LOCAL_EXECS > 0)); then
-      export PYSP_TEST_spark_master="local-cluster[$NUM_LOCAL_EXECS,$CORES_PER_EXEC,$MB_PER_EXEC]"
+    # Spark 3.1.1 includes https://github.com/apache/spark/pull/31540
+    # which helps with spurious task failures as observed in our tests. If you are running
+    # Spark versions before 3.1.1, this sets the spark.max.taskFailures to 4 to allow for
+    # more lineant configuration, else it will set them to 1 as spurious task failures are not expected
+    # for Spark 3.1.1+
+    VERSION_STRING=`$SPARK_HOME/bin/pyspark --version 2>&1|grep version\ |grep -v Scala|awk '{print $5}'`
+    echo "Detected Spark version is $VERSION_STRING"
```
gerashegalov, line 126: this involves some heavy initialization of Spark JVM and takes a few seconds on my machine. How about we read it from the spark core jar directly to the tune of 
```bash
unzip -q -c ~/dist/spark-3.1.1-bin-hadoop3.2/jars/spark-core_*.jar spark-version-info.properties | cut -d= -f2 | head -1
3.1.1
```

integration_tests/run_pyspark_from_build.sh
```diff
@@ -117,9 +117,17 @@ else
...
+    IS_BEFORE_SPARK_311=`python -c "print(1 if '$VERSION_STRING' < '3.1.1' else 0)"`
```
gerashegalov, line 127: we can use string comparison in bash, and don't seem to need this variable
```bash
[[ "$VERSION_STRING" < "3.1.1" ]] && SPARK_TASK_MAXFAILURES=4
```

jenkins/spark-tests.sh
```diff
@@ -66,7 +66,6 @@ BASE_SPARK_SUBMIT_ARGS="$BASE_SPARK_SUBMIT_ARGS \
     --executor-memory 12G \
     --total-executor-cores 6 \
     --conf spark.sql.shuffle.partitions=12 \
-    --conf spark.task.maxFailures=1 \
```
pxLi, line 69: in nightly integration tests, we explicitly set env `SPARK_VER` in pipelines for spark artifacts download, we could set this config based on `SPARK_VER` like https://github.com/NVIDIA/spark-rapids/blob/branch-21.08/jenkins/spark-tests.sh#L100

integration_tests/run_pyspark_from_build.sh
abellina: that is way better than my prior parse/skip/ascii art solution.

integration_tests/run_pyspark_from_build.sh
abellina: So I mistakenly thought that python was doing something smart about versions in that compare (which is an odd assumption come to think of it). I used that because that's our impl for`is_before_spark_311()`. I think either version is technically wrong. How about:

```
python -c "from packaging import version; print(1 if version.parse('3.1.9') < version.parse('3.1.10') else 0)"
```

The above prints "1", whereas:

```
if [[ 3.1.9 < 3.1.10 ]]; then echo "1"; else echo "0"; fi
```

prints "0".

gerashegalov: I did not comment about this because it looked good enough to me. But this is indeed a correct way to do semantic version comparison.

jenkins/spark-tests.sh
gerashegalov: I think it's better to not rely on it directly, so we have the logic that works for both: pipelines and local direct invocations of integration_tests/run_pyspark_from_build.sh

pxLi: +1, we will need update https://github.com/NVIDIA/spark-rapids/blob/branch-21.08/jenkins/spark-tests.sh#L97-L103 to use run_pyspark_from_build.sh here

integration_tests/run_pyspark_from_build.sh
abellina: went for the simpler approach, since all we really care about is less than 3.1.1

jenkins/spark-tests.sh
gerashegalov: @pxLi good catch, do you mind filing an issue because it won't happen in this PR

abellina: @gerashegalov @pxLi I duplicated some logic in spark-tests.sh. Yes, I tried combining, and it's probably OK, but that seems like a different issue. https://github.com/NVIDIA/spark-rapids/pull/2792/commits/94cca6b524243b6f80e9e7f3eb6c03c03c8ed08e

abellina: I did base it off SPARK_VER, as it was before in spark-tests.sh.

pxLi (dismissed):
LGTM. created a following task at: https://github.com/NVIDIA/spark-rapids/issues/2802

gerashegalov (approved):
LGTM

pxLi:
build

abellina:
@revans2 when you get a chance, this could use another look.

integration_tests/run_pyspark_from_build.sh
```diff
@@ -117,10 +117,15 @@ else
...
-
-    if ((NUM_LOCAL_EXECS > 0)); then
-      export PYSP_TEST_spark_master="local-cluster[$NUM_LOCAL_EXECS,$CORES_PER_EXEC,$MB_PER_EXEC]"
-    fi
+    # Spark 3.1.1 includes https://github.com/apache/spark/pull/31540
+    # which helps with spurious task failures as observed in our tests. If you are running
+    # Spark versions before 3.1.1, this sets the spark.max.taskFailures to 4 to allow for
+    # more lineant configuration, else it will set them to 1 as spurious task failures are not expected
+    # for Spark 3.1.1+
+    VERSION_STRING=`unzip -q -c $SPARK_HOME/jars/spark-core_*.jar spark-version-info.properties | head -1 | cut -d'=' -f2`
```
jlowe, line 125: Is it guaranteed that the Spark version is the first thing in the properties file across all possible Spark versions (e.g.: from cloud providers, etc.)?  I think this should be filtering for the `version` property (e.g.: replace `head -1` with `grep '^version='`) and erroring when it cannot find it.

integration_tests/run_pyspark_from_build.sh
```diff
@@ -129,9 +134,26 @@ else
...
+    # Set spark.task.maxFailures for most schedulers.
+    #
+    # Local (non-cluster) mode is the exception and does not work with `spark.task.maxFailures`.
+    # It requires two arguments to the master specification "local[N, K]" where
+    # N is the number of threads, and K is the maxFailures (otherwise this is hardcoded to 1,
+    # see https://issues.apache.org/jira/browse/SPARK-2083).
+    export PYSP_TEST_spark_task_maxFailures="$SPARK_TASK_MAXFAILURES"
+
+    if ((NUM_LOCAL_EXECS > 0)); then
+      export PYSP_TEST_spark_master="local-cluster[$NUM_LOCAL_EXECS,$CORES_PER_EXEC,$MB_PER_EXEC]"
```
revans2, line 149: There is no requirement that this script be used in local mode!!!. It started out that way, but is now used everywhere, and we probably should rename it.  We cannot assume this.

revans2 (changes requested):

integration_tests/run_pyspark_from_build.sh
abellina: Sorry I am confused. `NUM_LOCAL_EXECS > 0` is unrelated to this change, and if this variable is not specified it defaults to 0. Below, I am trying (perhaps failing) to check that the master was set.

integration_tests/run_pyspark_from_build.sh
abellina: It is according to: https://github.com/apache/spark/blob/master/build/spark-build-info#L29. But I can make it stronger with the grep as suggested + error if the version can't be detected this way.

integration_tests/run_pyspark_from_build.sh
revans2: Sorry this script has changed so much since I wrote it and now it is very confusing.

revans2 (approved):

integration_tests/run_pyspark_from_build.sh
jlowe: Yes, it's that way in that particular codebase, but not necessarily in a future Spark version or in some cloud provider's custom version.  Checking for the well-known property name seems prudent here and lower risk than assuming what is in the first line of a property file that is otherwise semantically unordered.

abellina: Please take a look at https://github.com/NVIDIA/spark-rapids/pull/2792/commits/0fdea2c0ff2169172659c6dae2add2e3fe7420a8, which should handle this case.

integration_tests/run_pyspark_from_build.sh
```diff
@@ -117,10 +117,24 @@ else
...
-      export PYSP_TEST_spark_master="local-cluster[$NUM_LOCAL_EXECS,$CORES_PER_EXEC,$MB_PER_EXEC]"
+    # Spark 3.1.1 includes https://github.com/apache/spark/pull/31540
+    # which helps with spurious task failures as observed in our tests. If you are running
+    # Spark versions before 3.1.1, this sets the spark.max.taskFailures to 4 to allow for
+    # more lineant configuration, else it will set them to 1 as spurious task failures are not expected
+    # for Spark 3.1.1+
+    set +e
+    VERSION_PROPERTY=`unzip -q -c $SPARK_HOME/jars/spark-core_*.jar spark-version-info.properties | grep ^version=`
+    if [[ $? -eq "0" ]]; then
+      VERSION_STRING=`echo "$VERSION_PROPERTY" | cut -d'=' -f2`
```
jlowe, line 128: This could be simplified a bit by just letting the pipeline try to extract the value and then check for an empty result, e.g.:
```
VERSION_STRING=`unzip -q -c $SPARK_HOME/jars/spark-core_*.jar spark-version-info.properties | grep ^version= | cut -d'=' -f2`
[[ -z $VERSION_STRING ]] && { echo "Unable to detect the Spark version at $SPARK_HOME"; exit 1; }
echo "Detected Spark version $VERSION_STRING"
```

jlowe (dismissed):

jlowe:
build

abellina:
Hmm, this failed to detect the spark version with:
```
10:21:52  ./run_pyspark_from_build.sh: line 125: unzip: command not found
```

jlowe:
> `unzip: command not found`

Probably a lean Docker image that doesn't have that command.  That makes me lean towards an approach where we ask Spark to print the version.

abellina:
Ok, then something like what I had before could be used. It does take some time, but in the grand scheme of things, it's probably not a huge deal:

```
$ time $SPARK_HOME/bin/pyspark --version 2>&1|awk '/version\ [0-9.]+$/{print $NF}'
3.1.1

real	0m1.120s
user	0m1.919s
sys	0m0.136s
```

abellina:
@gerashegalov fyi

integration_tests/run_pyspark_from_build.sh
```diff
@@ -117,10 +117,17 @@ else
     NUM_LOCAL_EXECS=${NUM_LOCAL_EXECS:-0}
     MB_PER_EXEC=${MB_PER_EXEC:-1024}
     CORES_PER_EXEC=${CORES_PER_EXEC:-1}
+    # Spark 3.1.1 includes https://github.com/apache/spark/pull/31540
+    # which helps with spurious task failures as observed in our tests. If you are running
+    # Spark versions before 3.1.1, this sets the spark.max.taskFailures to 4 to allow for
+    # more lineant configuration, else it will set them to 1 as spurious task failures are not expected
+    # for Spark 3.1.1+
+    VERSION_STRING=`$SPARK_HOME/bin/pyspark --version 2>&1|awk '/version\ [0-9.]+$/{print $NF}'`
```
jlowe, line 125: This fails for version strings with alphabetical characters in it, e.g.: `3.2.0-SNAPSHOT`

abellina: nice catch, I'll remove the EOL check.

jlowe (approved):

jlowe:
build

abellina:
CI job for this build seems happy (so far):

``
14:50:20  Detected Spark version 3.0.1
14:50:20  + export 'PYSP_TEST_spark_master=local[*,4]'
```

**Comment by gerashegalov** [comment]
> > `unzip: command not found`
> 
> Probably a lean Docker image that doesn't have that command. That makes me lean towards an approach where we ask Spark to print the version.

We should at least have a JDK so extracting spark-properties from jar is an option

```bash
$JAVA_HOME/bin/jar xf $SPARK_HOME/jars/spark-core_*.jar spark-version-info.properties && grep version spark-version-info.properties | cut -d= -f2
3.0.3
```

So is just looking at the file name I suggested as another alternative in the original comment. E.g
```bash
VERSION_REGEX='.*/spark-core_.*-(.*).jar'
[[ $(echo $SPARK_HOME/jars/spark-core_*.jar) =~ $VERSION_REGEX ]] && echo ${BASH_REMATCH[1]}
3.1.1
```

integration_tests/run_pyspark_from_build.sh
gerashegalov: @revans2 The default is 0 and does not interfere with the decision to use or not to use local-cluster mode, but if the user sets it to a positive value it's exactly what the user is signalling to the script https://github.com/NVIDIA/spark-rapids/blob/8d3efb8ffba02da69768f7d69bc148da3a01afa9/integration_tests/README.md#spark-execution-mode

abellina:
@gerashegalov In this scenario, since there is CLI util argument maintained by Spark specific for this (`--version`), I think that's a fine choice. I know the format is not perfect, but it has been consistent for the last few versions I've used Spark. I rather just use `--version` at this point and have the ugly regex, than try to fish for a file in the jar, or depend on a filename.

gerashegalov:
@abellina how often has the naming scheme for the spark-core jar changed? https://repo1.maven.org/maven2/org/apache/spark/spark-core_2.10/0.9.1/

abellina:
> @abellina how often has the naming scheme for the spark-core jar changed? https://repo1.maven.org/maven2/org/apache/spark/spark-core_2.10/0.9.1/

If you are using a regular distro it makes perfect sense. But say I have my own custom distribution of Spark, and I decide I want to change the name to match my versioning. Then our integration tests will fail to detect that Spark version, whereas if there is a `pyspark` command (which we need anyway), it should output a version unless they specifically remove support for `--version`.

gerashegalov (approved):

gerashegalov:
> If you are using a regular distro it makes perfect sense. But say I have my own custom distribution of Spark, and I decide I want to change the name to match my versioning. Then our integration tests will fail to detect that Spark version, whereas if there is a `pyspark` command (which we need anyway), it should output a version unless they specifically remove support for `--version`.

It's good enough

abellina:
Thanks everyone for the reviews.

PR #2793, opened 2021-06-23
Report equivilant stages/sql ids as a part of compare

revans2:
For the Profile tool this helps a lot when trying to compare large complicated queries to each other.

This refactored dot generation because I wanted to reused the stage matching part of it, but I needed a better way to do it. I also wanted to be able to output dot graphs as a apart of compare, even if they were not producing a graph that compared the different applications.


tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/ApplicationInfo.scala
```diff
@@ -34,16 +34,165 @@ import org.apache.spark.deploy.history.{EventLogFileReader, EventLogFileWriter}
...
+    simpleString: String,
+    override val children: Seq[SparkPlanInfoWithStage],
+    metadata: scala.Predef.Map[String, String],
+    metrics: Seq[SQLMetricInfo],
+    val stageId: Option[Int]) extends SparkPlanInfo(nodeName, simpleString, children,
+  metadata, metrics) {
+
+  import SparkPlanInfoWithStage._
+
+  def debugEquals(other: Any, depth: Int = 0): Boolean = {
```
revans2, line 56: I left this in, because our plugin is constantly changing and we have not really tested this with databricks or others, so if we run into issues, this is a quick way to track down what parts of a plan need to be normalized for the comparison to work properly.

tools/README.md
```diff
@@ -265,6 +265,50 @@ Compare Executor Information:
 +--------+-----------------+------------+-------------+-----------+------------+-------------+-------------+--------------+------------------+---------------+-------+-------+
 ```
 
+- Matching SQL IDs Across Applications:
+```
+Matching SQL IDs Across Applications:
++-----------------------+-----------------------+
+|app-20210329165943-0103|app-20210329170243-0018|
++-----------------------+-----------------------+
+|0                      |0                      |
```
tgravescs, line 274: would it be helpful to also add time each took?  I guess we can always add more later.

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/CollectInformation.scala
```diff
@@ -125,102 +120,16 @@ class CollectInformation(apps: Seq[ApplicationInfo],
...
-  def printSQLPlanMetrics(shouldGenDot: Boolean, outputDir: String): Unit = {
+  def printSQLPlanMetrics(): Unit = {
     for (app <- apps){
       if (app.allDataFrames.contains(s"sqlMetricsDF_${app.index}") &&
         app.allDataFrames.contains(s"driverAccumDF_${app.index}") &&
         app.allDataFrames.contains(s"taskStageAccumDF_${app.index}") &&
         app.allDataFrames.contains(s"jobDF_${app.index}") &&
         app.allDataFrames.contains(s"sqlDF_${app.index}")) {
         val messageHeader = "\nSQL Plan Metrics for Application:\n"
-        val accums = app.runQuery(app.generateSQLAccums, fileWriter = fileWriter,
```
tgravescs, line 219: so we lost the reuse of generateSQLAccums here, not sure how long that takes but since its sql query probably non-trivial we may want to see if we can generate once and reuse

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/ProfileMain.scala
```diff
@@ -154,12 +158,29 @@ object ProfileMain extends Logging {
       healthCheck.listRemovedExecutors()
       healthCheck.listPossibleUnsupportedSQLPlan()
 
+      if (appArgs.generateDot()) {
+        if (appArgs.compare()) {
+          logWarning("Dot graph does not compare apps")
```
tgravescs, line 163: just printing but still running it,  skip the generation in this case, we could also add validate in the ProfileArgs to just error out if user specifies

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/ProfileMain.scala
```diff
@@ -154,12 +158,29 @@ object ProfileMain extends Logging {
...
+          GenerateDot(app, outputDirectory)
+          val duration = TimeUnit.SECONDS.convert(System.nanoTime() - start, TimeUnit.NANOSECONDS)
+          textFileWriter.write(s"Generated DOT graphs for app ${app.appId} " +
+              s"to $outputDirectory in $duration second(s)\n")
+        }
+      }
+
       if (appArgs.generateTimeline()) {
         if (appArgs.compare()) {
           logWarning("Timeline graph does not compare apps")
```
tgravescs, line 176: same here,

tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/ApplicationInfo.scala
```diff
@@ -34,16 +34,165 @@ import org.apache.spark.deploy.history.{EventLogFileReader, EventLogFileWriter}
...
+        val normalizedChildren = children.map(_.normalizeForStageComparison)
+        // We are going to ignore the simple string because it can contain things in it that
+        // are specific to a given run
+        new SparkPlanInfoWithStage(newName, newName, normalizedChildren, metadata,
+          metrics, stageId)
+    }
+  }
+
+  /**
+   * Walk the three depth first and get all of the stages for each node in the tree
```
tgravescs, line 140: s/three/tree/

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/CollectInformation.scala
revans2: That generic reuse did not work.  It was not caching the data so the only thing it saved on was re-parsing the SQL query into a data frame.

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/ProfileMain.scala
revans2: I did this change because I wanted to have it run.  I don't want to have to run the profiling tool twice, once to get the dot graphs and once to do the comparison. The warning is just there because some people might think that if they are doing a comparison it might try to do what the original dot code was trying to do.

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/ProfileMain.scala
revans2: Same response here. I want to get the graphs out without having to run the profiling tool twice.

tools/README.md
revans2: Ya I thought about that, but I didn't want to make it too complicated to start out with.

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/ProfileMain.scala
tgravescs: ok, makes sense

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/CollectInformation.scala
revans2: If you want me to I can make the data returned a lazy value in ApplicationInfo so it does end up being shared.

tgravescs: ah right, we change the query in generate dot so don't get any stage reuse

tgravescs:
build

tgravescs (approved):

Issue #2739, opened 2021-06-18
[BUG] CBO causes less efficient plan for NDS q84


andygrove:
**Describe the bug**
With the cost-based optimizer enabled, there is a CPU projection between two GPU BroadcastHashJoin operators, which does not make sense.

![cpu-project-between-gpu-bhj](https://user-images.githubusercontent.com/934084/122489261-76358980-cf9c-11eb-8de1-88cacc533dd5.png)

**Steps/Code to reproduce bug**
Decimals enable, AQE on, CBO on. SF=3k.

**Expected behavior**
Projection should stay on GPU.

**Additional context**
None

andygrove:
I tracked the issue down and it was an integer overflow when estimating the size of data that an operator would produce. The estimated row counts were extremely high, which is why I could only reproduce this on a large data set.

PR #2794, opened 2021-06-23
Prevent integer overflow when estimating data sizes in cost-based optimizer

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/2739

The issue was that an integer overflow when estimating data sizes was resulting in negative costs for the CPU cost model.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -279,6 +280,8 @@ trait CostModel {
 
 }
 
+object Cost
+
```
jlowe, line 284:
```suggestion
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -27,6 +27,7 @@ import org.apache.spark.sql.execution.aggregate.HashAggregateExec
 import org.apache.spark.sql.execution.exchange.BroadcastExchangeExec
```
jlowe, line 27: ProjectExec import is unused

jlowe (approved):

jlowe:
build

PR #2795, opened 2021-06-23
Improve CBO debug logging

andygrove:

With AQE enabled, the optimizer will be called multiple times for a single query plan. This PR introduces BEGIN/END logging around each optimization pass to make it easier to understand the output.


jlowe (approved):

jlowe:
build

PR #2804, opened 2021-06-24
[WIP] Support GpuCollectList/GpuCollectSet in groupBy aggregation

sperlingxx:
Current PR is to support `GpuCollectList` and `GpuCollectSet` in groupBy aggregation.

`CollectList` and `CollectSet` are the first two `TypedImperativeAggregate`, which we attempt to provide GPU support. Unlike `DeclativeAggregate` and other `ImperativeAggregate` (such as `PivotFirst`), `TypedImperativeAggregate` will create aggBufferAttributes of BinaryType (for serialization). Meanwhile, in GPU counterparts, we store aggBufferAttributes just as datatype of update expressions (Array[elementType] for collectOps). Therefore, we involve three additional tasks in this PR:
1. Replaces Binary typed aggBufferAttributes with corresponding GPU buffers, which is done in `GpuObjectHashAggregateMeta.convertToGpu`. Although `ObjectHashAggregateExec` doesn't cover all cases of `TypedImperativeAggregate`, it covers the most common usages and we can support another situations later (for instance, aggregation contains multiple distinct, which rewritten by optimizer).
2. Bypasses the type check system through marking `TypeSig.BINARY` as plugin-supported for `ObjectHashAggregateExec`, `HashPartition` and `ShuffleExchangeExec`. And checks the validity of BinaryTypes in the tagForGpu methods.
3. Reworked (simplified) `GpuHashAggregateExec.setupReferences` to make it work with TypedImperativeAggregate in PartialMerge mode.

I labeled this PR as WIP, since I am not quite confident whether the above approaches are appropriate or not.


integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -371,6 +374,107 @@ def test_hash_reduction_pivot_without_nans(data_gen, conf):
             .agg(f.sum('c')),
         conf=conf)
 
+_repeat_agg_column_for_collect_op = [
```
ttnghia, line 377: Why "has_aggregate"? In cudf, collect list/set are only sort_aggregages.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -255,6 +255,50 @@ class CudfMin(ref: Expression) extends CudfAggregate(ref) {
...
+  override lazy val mergeAggregate: Aggregation = Aggregation.mergeLists()
+  override def toString(): String = "CudfMergeLists"
+}
+
+class CudfCollectSet(ref: Expression) extends CudfAggregate(ref) {
+  override lazy val updateReductionAggregate: cudf.ColumnVector => cudf.Scalar =
+    throw new UnsupportedOperationException("CollectSet is not yet supported in reduction")
+  override lazy val mergeReductionAggregate: cudf.ColumnVector => cudf.Scalar =
+    throw new UnsupportedOperationException("CollectSet is not yet supported in reduction")
+  override lazy val updateAggregate: Aggregation = Aggregation.collectSet()
```
ttnghia, line 285: Humnn, is there any way to call `Aggregation.collectList` if the data are partitioned into more than one batch? As we only need collectSet for the intermediate results. Calling collectSet for them is expensive, as that involves unnecessarily executing `drop_list_duplicates` on the intermediate lists.

integration_tests/src/main/python/hash_aggregate_test.py
revans2: @ttnghia This test is not named after how cudf implements something.  What is more `cudf` hides that detail from users so any test outside of cudf itself That is a hidden implementation detail. It is named after how Spark implemented it, and we are trying to get coverage on that operator.

That said CollectList and CollectSet apparently are showing up as `ObjectHashAggregateExec` and there are also some `SortAggregateExec` tests in here too. At some point we should either split this up into separate files or rename it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2740,12 +2744,25 @@ object GpuOverrides {
       "Hash based partitioning",
       // This needs to match what murmur3 supports.
       PartChecks(RepeatingParamCheck("hash_key",
-        (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.STRUCT).nested(),
+        (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.BINARY +
+            TypeSig.STRUCT).nested()
+            .withPsNote(TypeEnum.BINARY, "Marking BINARY as plugin-supported is only to " +
```
revans2, line 2749: What exactly does this mean and how does this help an end user?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2943,16 +2960,29 @@ object GpuOverrides {
...
+      ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.BINARY +
+          TypeSig.ARRAY + TypeSig.STRUCT).nested()
           .withPsNote(TypeEnum.STRUCT, "Round-robin partitioning is not supported for nested " +
               s"structs if ${SQLConf.SORT_BEFORE_REPARTITION.key} is true")
           .withPsNote(TypeEnum.ARRAY, "Round-robin partitioning is not supported if " +
               s"${SQLConf.SORT_BEFORE_REPARTITION.key} is true")
           .withPsNote(TypeEnum.MAP, "Round-robin partitioning is not supported if " +
-              s"${SQLConf.SORT_BEFORE_REPARTITION.key} is true"),
+              s"${SQLConf.SORT_BEFORE_REPARTITION.key} is true")
+          .withPsNote(TypeEnum.BINARY, "Marking BINARY as plugin-supported is only to " +
```
revans2, line 2971: Why do we need a restriction on this?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3004,6 +3034,19 @@ object GpuOverrides {
             .withPsNote(TypeEnum.STRUCT, "not allowed for grouping expressions"),
         TypeSig.all),
       (agg, conf, p, r) => new GpuHashAggregateMeta(agg, conf, p, r)),
+    exec[ObjectHashAggregateExec](
+      "The backend for hash based aggregations",
```
revans2, line 3038: I think we need a clearer description to distinguish it from the regular HashAggregateExec.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -68,6 +68,13 @@ object AggregateUtils {
       }
     }
   }
+
+  def filterNonAggBufBinaryExpressions(expressions: Seq[Expression]): Seq[Expression] = {
+    expressions.filter {
+      case AttributeReference("buf", BinaryType, _, _) => false
```
revans2, line 74: What makes "buf" special?  This is not good enough.

revans2 (commented):
This is a lot of change and I think we need to look into what is the proper way to deal with `ObjectHashAggregate`.  It feels kind of hacked together and I want to get a real design put in place before we push this in.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -724,11 +782,19 @@ case class GpuCollectList(
...
 case class GpuCollectSet(
-    childExpression: Expression,
+    child: Expression,
     mutableAggBufferOffset: Int = 0,
     inputAggBufferOffset: Int = 0)
     extends GpuCollectBase[CollectSetAggregation] {
 
+  override lazy val updateExpressions: Seq[GpuExpression] = new CudfCollectSet(inputBuf) :: Nil
+
+  override lazy val mergeExpressions: Seq[GpuExpression] = new CudfMergeSets(outputBuf) :: Nil
```
ttnghia, line 792: I'm thinking of totally removing `CudfCollectSet` and `CudfMergeSet`. We just call `CudfCollectList` in `updateExpressions`, and `CudfMergeLists` in `mergeExpressions`. During `evaluateExpression`, we call `cudfDropListDuplicates` on the merged lists.

jlowe:
Is this superceded by #2971?

sperlingxx:
> Is this superceded by #2971?

Yes, I closed it.

PR #2805, opened 2021-06-24
Revert "Use UCX Active Messages for all shuffle transfers (#2735)"

abellina:
This reverts commit e0e4833f1d715826506c2cbc0cf457c26c13f99a.

An integration test caught an issue where we are failing to use a UCX `amData` object after the callback returns. This should be handled by passing extra flags to `worker.setAmRecvHandler` (as per https://github.com/openucx/ucx/pull/6735), but unfortunately the version of this function that takes flags is not published to maven.

This was missed because I tested with a custom UCX that had default flags to (which was the direction the PR was going to):
```
    param.flags      = UCP_AM_FLAG_WHOLE_MSG | UCP_AM_FLAG_PERSISTENT_DATA;
```

But the code actually sets this to 0 unless specified by the user.

Once JUCX 1.11 is published with the change to `setAmRecvHandler`, the change would be to bring this back, and also set: `UCP_AM_FLAG_PERSISTENT_DATA` (for the use-after-callback pattern) and `UCP_AM_FLAG_WHOLE_MSG` (to prevent UCX to call us multiple times with eager fragments in the future).

@petro-rudenko fyi.


jlowe (approved):

jlowe:
build

abellina:
Tracking issue for the fix: https://github.com/NVIDIA/spark-rapids/issues/2807.

PR #2810, opened 2021-06-24
Use SecureRandom for UCPListener TCP port choice

abellina:

Please note that once UCX 1.11 is released, we will remove the need for this choice since UCX will delegate to the kernel to set the ephemeral port.

### DIFF
--- a/shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
+++ b/shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
@@ -183,7 +183,6 @@ class UCX(transport: UCXShuffleTransport, executor: BlockManagerId, rapidsConf: 
       worker = context.newWorker(workerParams)
       logInfo(s"UCX Worker created")
       if (rapidsConf.shuffleUcxUseSockaddr) {
-        val secureRandom = new SecureRandom()
         // For now backward endpoints are not used, but need to create
         // an endpoint from connectionHandler in order to use ucpListener connections.
         // With AM this endpoints would be used as replyEp.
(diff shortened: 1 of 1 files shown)


abellina:
Side note: I did check in this scenario that `nextInt` reads from `/dev/urandom` and does not block in any of my tests in Ubuntu or CentOS. 

With the default algorithm `NativePRNG`, `nextInt` ends up calling `implNextBytes` which ends up getting random values from `/dev/urandom`. If you were to call `nextBytes`, then `/dev/random` is picked and can block, unless you used `NativePRNG.NonBlocking`. I did not want to hard specify an algorithm.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -183,6 +183,7 @@ class UCX(transport: UCXShuffleTransport, executor: BlockManagerId, rapidsConf:
       worker = context.newWorker(workerParams)
       logInfo(s"UCX Worker created")
       if (rapidsConf.shuffleUcxUseSockaddr) {
+        val secureRandom = new SecureRandom()
```
jlowe, line 186: Nit: Why not move this within the one block that needs it?  If the port is specified in the conf then this will still construct an unused instance.

jlowe (approved):
Small nit but otherwise lgtm.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
abellina: Fixed!

abellina:
build

gerashegalov (approved):

jlowe (approved):

PR #2816, opened 2021-06-25
Improve some existing collectTime handling

andygrove:

Changes in this PR:

- Introduces an iterator for wrapping a batch iterator and incrementing collectTime
- Refactor some operators to use this new iterator, to reduce code duplication
- Refactor some code to use `withResource` pattern instead of try/finally with manual resource management

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
@@ -272,6 +272,7 @@ case class GpuTopN(
   override lazy val additionalMetrics: Map[String, GpuMetric] = Map(
     NUM_INPUT_ROWS -> createMetric(DEBUG_LEVEL, DESCRIPTION_NUM_INPUT_ROWS),
     NUM_INPUT_BATCHES -> createMetric(DEBUG_LEVEL, DESCRIPTION_NUM_INPUT_BATCHES),
+    COLLECT_TIME -> createNanoTimingMetric(MODERATE_LEVEL, DESCRIPTION_COLLECT_TIME),
     TOTAL_TIME -> createNanoTimingMetric(MODERATE_LEVEL, DESCRIPTION_TOTAL_TIME),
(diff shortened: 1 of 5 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -272,6 +272,7 @@ case class GpuTopN(
   override lazy val additionalMetrics: Map[String, GpuMetric] = Map(
     NUM_INPUT_ROWS -> createMetric(DEBUG_LEVEL, DESCRIPTION_NUM_INPUT_ROWS),
     NUM_INPUT_BATCHES -> createMetric(DEBUG_LEVEL, DESCRIPTION_NUM_INPUT_BATCHES),
+    COLLECT_TIME -> createNanoTimingMetric(MODERATE_LEVEL, DESCRIPTION_COLLECT_TIME),
```
revans2, line 275: Specifically what are you trying to calculate with a collect time here?

I know we have added/used collect time to try and measure the time it takes for CPU code by adding collect_time to columnar to row and row to columnar transitions.  I also know that spark uses something similar for broadcast joins to try and measure the time it takes to get the broadcast.  I can also see it being used for the initial coalesce batch right after a shuffle to try and measure the read time of a shuffle.  What I don't understand is why are we adding it to GpuTopN?

andygrove: I guess I got carried away while working through the list in https://github.com/NVIDIA/spark-rapids/issues/2334 .. I'll revert the GpuTopN changes and just leave the code cleanup changes.

revans2: I love having more metrics, it is just frustrating that Spark isn't better at handling them.  The rest of the changes look great.

revans2 (approved):

andygrove:
build

PR #2821, opened 2021-06-25
Truncate long plan labels and refer to "print-plans"

### LINKED ISSUE #2710: [BUG] dot file generation can go over the limits of dot
**Describe the bug**
I recently hit a bug when trying to build a dot file generated by the tool for NDS query 24.

```
$ dot -Tsvg -O ./profile_output_171/*.dot
Error: ./profile_output_171/app-20210528152635-0280-query-24.dot: syntax error in line 3 scanning a quoted string (missing endquote? longer than 16384?)
String starting:"\lApplication: app-20210528152635-0280\lQuery: 24\l\l== Parsed Logical Plan ==\
```

I had to go in and delete the extra info from the label for the digraph.

It might be nice to detect this and avoid it, or even better come up with a different way to display the plans that would not violate this limit.

gerashegalov:
Properly treat graphviz compiled with the default 16K label max

Closes #2710.

### DIFF
--- a/tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/GenerateDot.scala
+++ b/tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/GenerateDot.scala
@@ -80,7 +80,7 @@ object GenerateDot {
       appId: String): Unit = {
     val graph = SparkPlanGraph(plan.plan, appId, sqlId.toString, physicalPlanString,
       accumIdToStageId, stageIdToStageMetrics)
-    val str = graph.makeDotFile(plan.metrics)
+    val str = graph.makeDotFile(plan.metrics, fileWriter.outputFilePath.getParent.toString)
     fileWriter.write(str)
   }
 }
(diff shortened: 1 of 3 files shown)


tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/GenerateDot.scala
```diff
@@ -292,6 +284,39 @@ object SparkPlanGraph {
...
   }
+
+  def createDotLabel(
+    outputDir: String, 
+    appId: String, 
+    sqlId: String,
+    physicalPlan: String,
+    maxLength: Int = 16384 
+  ): String = {
+    val sqlPlanFile = s"file://$outputDir/planDescriptions-$appId"
```
tgravescs, line 295: outputDir isn't necessarily file://, could be in hdfs for instance, I guess maybe you are thinking they download it anyway so when open it should be?  But if they do that outputdir could be different

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/GenerateDot.scala
```diff
@@ -292,6 +284,39 @@ object SparkPlanGraph {
...
+    val sqlPlanPlaceHolder = "%s"
+    val htmlLineBreak = """<br align="left"/>""" + "\n"
+    val queryLabelFormat = 
+      s"""<<table border="0">
+         |<tr><td>Application: $appId, Query: $sqlId</td></tr>
+         |<tr><td>$sqlPlanPlaceHolder</td></tr>
+         |<tr><td href="$sqlPlanFile">
```
tgravescs, line 302: I would just say see the output from --print-plans and not do the link.

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/GenerateDot.scala
gerashegalov: yes forgot about non-OS-mounted locations.

gerashegalov:
build

tgravescs (approved):

PR #2822, opened 2021-06-25
Rewrite Qualification tool for better performance

tgravescs:
This PR rewrites parts of the qualification tool.  I kept the overall structure very similar so that code could be reused with the Profiling tool. 

overall this is what this does:
- removes requirement to run inside of Spark (ie no more Spark SQL Dataframe operations for aggregations), now you can just run via java.   It still does require the SPARK jars to be present though.
- Changes output to output both a text summary report with only 4 columns and a CSV file with the raw data that has more info
- reorganizes the parsing code to aggregate things as we read the events whenever possible.
- New QualAppInfo class that is specific to Qualification, removed pieces from ApplicationInfo that were Qualification specific.
- Only keep the summary of each app in memory, individual app data is thrown away so we don't require a lot of memory
- refactors event processor and ApplicationInfo to have base classes that could be shared between profiling and qualification tools
- added in another columns for sql ids that had failed jobs in them.
- added field SQL Duration with Potential Problems that is the total duration for any SQL queries that we found problematic - ie UDFS right now

This takes significantly less time and memory now to process event logs. The time it takes is really close to just the time it takes Spark to read and parse the events.


tgravescs:
build

tgravescs:
build

revans2 (commented):
I am not an expert on the Qualifying tool but the changes look good to me.  The only thing that might be interesting to explore is putting in a thread pool, like the Spark history server uses.  That way if there are more than one application to look at for the qualification tool, it can all be done in parallel (up to a set thread pool size).

tgravescs:
test failure because not upmerged

tgravescs:
thanks Bobby, definitely a good idea. I'll do that in a followup

tgravescs:
build

tgravescs:
build

tgravescs:
@revans2  are you ok with this being merged (if so can you approve)?

revans2 (approved):

tgravescs:
build

tools/pom.xml
```diff
@@ -41,6 +41,7 @@
         <dependency>
             <groupId>org.scala-lang</groupId>
             <artifactId>scala-library</artifactId>
+            <scope>compile,runtime</scope>
```
gerashegalov, line 44: I think  the `compile` scope is a [superset](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope) of the `runtime` scope. As written, this is equivalent to just `compile`

tools/src/main/scala/org/apache/spark/sql/rapids/tool/EventProcessorBase.scala
```diff
@@ -0,0 +1,211 @@
...
+  def doSparkListenerResourceProfileAddedReflect(
+      app: T,
+      event: SparkListenerEvent): Boolean = {
+    val rpAddedClass = "org.apache.spark.scheduler.SparkListenerResourceProfileAdded"
+    if (event.getClass.getName.equals(rpAddedClass)) {
+      try {
+        event match {
+          case _: SparkListenerResourceProfileAdded =>
+            doSparkListenerResourceProfileAdded(app,
+              event.asInstanceOf[SparkListenerResourceProfileAdded])
```
gerashegalov, line 103: nit: can avoid asInstanceOf by using a variable in the pattern instead of `_`

gerashegalov (approved):

tools/pom.xml
tgravescs: updated

tools/src/main/scala/org/apache/spark/sql/rapids/tool/EventProcessorBase.scala
tgravescs: updated

tgravescs:
build

gerashegalov (approved):

Issue #2820, opened 2021-06-25
[BUG] Metrics are inconsistent for GpuRowToColumnarToExec


andygrove:
**Describe the bug**
`GpuRowToColumnarToExec` uses `totalTime` to measure the time to create batches on the GPU so should really be renamed to `opTime` or `gpuOpTime`. It would be good to also have `totalTime` to measure the cost of buffering and waiting for the GPU semaphore.

PR #2720, opened 2021-06-15
Add metrics documentation to the tuning guide

andygrove:
Signed-off-by: Andy Grove <andygrove@nvidia.com>

Add metrics documentation to the tuning guide.

PR #2823, opened 2021-06-25
Add semaphoreWaitTime and gpuOpTime for GpuRowToColumnarExec

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/2820

This PR changes the metrics reported for `GpuRowToColumnarExec` to include `semaphoreWaitTime` and `gpuOpTime` (previously reported as `totalTime`) as well as `totalTime`.

# Before

gpu-r2c-before

# After

There are two code paths for row-to-column.

## GeneratedUnsafeRowToCudfRowIterator

gpu-r2c-after-gpu

## GpuRowToColumnConverter

gpu-r2c-after-cpu


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
```diff
@@ -806,6 +817,8 @@ case class GpuRowToColumnarExec(child: SparkPlan, goal: CoalesceSizeGoal)
   }
 
   override lazy val additionalMetrics: Map[String, GpuMetric] = Map(
+    "semaphoreWait" -> createNanoTimingMetric(DEBUG_LEVEL, "Semaphore wait time"),
```
jlowe, line 820: Curious why this isn't defined in `GpuMetric` like the others?  I would expect this to start showing up in other execs that acquire the semaphore.

jlowe (commented):
I was going to say we need to update the tuning guide to discuss these new metrics, but #2720 wasn't merged?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
```diff
@@ -806,6 +817,8 @@ case class GpuRowToColumnarExec(child: SparkPlan, goal: CoalesceSizeGoal)
   }
 
   override lazy val additionalMetrics: Map[String, GpuMetric] = Map(
+    SEMAPHORE_WAIT_TIME -> createNanoTimingMetric(DEBUG_LEVEL, DESCRIPTION_SEMAPHORE_WAIT_TIME),
+    GPU_OP_TIME -> createNanoTimingMetric(MODERATE_LEVEL, DESCRIPTION_GPU_OP_TIME),
     TOTAL_TIME -> createNanoTimingMetric(MODERATE_LEVEL, DESCRIPTION_TOTAL_TIME),
```
revans2, line 822: I am tempted to drop `TOTAL_TIME` to `DEBUG_LEVEL`, but I don't know how you want to use it with benchmarks/etc so I don't know if that is a good idea or not.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSemaphore.scala
```diff
@@ -103,8 +116,11 @@ private final class GpuSemaphore(tasksPerGpu: Int) extends Logging {
   // Map to track which tasks have acquired the semaphore.
   private val activeTasks = new ConcurrentHashMap[Long, MutableInt]
 
-  def acquireIfNecessary(context: TaskContext): Unit = {
-    val nvtxRange = new NvtxRange("Acquire GPU", NvtxColor.RED)
+  def acquireIfNecessary(context: TaskContext, waitMetric: Option[GpuMetric] = None): Unit = {
+    val nvtxRange = waitMetric match {
+      case Some(m) => new NvtxWithMetrics("Acquire GPU", NvtxColor.RED, m)
```
revans2, line 121: We have done this in a few places. Would it be better to just have `NvtxWithMetrics` have a constructor that takes a `Option[GpuMetric]` and hide it internally?

revans2 (dismissed):
Overall it looks good. Just a question and a small nit.

revans2:
On a side not I noticed that data bricks photon has a "cumulative time" metric that fills the same role we have been using for "total time" but feels a bit more accurate in the naming. Just curious if it is worth refactoring/renaming to match?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
andygrove: What I would really like is OP_TIME. I should be able to calculate that as TOTAL_TIME - SEMAPHORE_WAIT_TIME - GPU_OPTIME. I'll look at that next.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSemaphore.scala
andygrove: > On a side not I noticed that data bricks photon has a "cumulative time" metric that fills the same role we have been using for "total time" but feels a bit more accurate in the naming. Just curious if it is worth refactoring/renaming to match?

I like the cumulative name better. I'll create a separate PR to propose renaming this metric.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
jlowe: Wouldn't that just be measuring mostly the time spent in earlier child execs within the stage?  TOTAL_TIME includes time spent fetching inputs from child iterators.  It wouldn't be very op-specific, and thus OP_TIME would be an odd name for it.

andygrove:
> On a side not I noticed that data bricks photon has a "cumulative time" metric that fills the same role we have been using for "total time" but feels a bit more accurate in the naming. Just curious if it is worth refactoring/renaming to match?

I like the cumulative name better. I'll create a separate PR to propose renaming this metric.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
andygrove: Yes, that's a good point. I would need to subtract the cost of the fetches too, but that might be expensive because this is a row-based iterator. I'll take a look and see what the options are.

revans2: This is converting rows on the CPU to columns on the GPU.  There is close to no processing on the CPU beyond fetching data from upstream and putting it into a buffer. If we try to measure the amount of time it takes to convert from UnsafeRow to CudfUnsafeRow, or to just put it into the arrow format in a buffer (depending on the code path we take), we are likely going to spend more time measuring than actually doing the conversion. Unless knowing that number is critically important we would propose that we just lump it all together.

revans2 (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala
andygrove: Ok, I went ahead and changed the level of TOTAL_TIME to DEBUG_LEVEL.

andygrove:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/UnsafeRowToColumnarBatchIterator.java
```diff
@@ -135,13 +143,9 @@ public ColumnarBatch next() {
...
         if (tc != null) {
-          GpuSemaphore$.MODULE$.acquireIfNecessary(tc);
-        }
-        if (totalTime != null) {
-          buildRange = new NvtxWithMetrics("RowToColumnar", NvtxColor.GREEN, totalTime);
-        } else {
-          buildRange = new NvtxRange("RowToColumnar", NvtxColor.GREEN);
+          GpuSemaphore$.MODULE$.acquireIfNecessary(tc, semaphoreWaitTime);
         }
+        buildRange = NvtxWithMetrics.apply("RowToColumnar: build", NvtxColor.GREEN, Option.apply(gpuOpTime));
```
jlowe, line 148:
```suggestion
        buildRange = NvtxWithMetrics.apply("RowToColumnar: build", NvtxColor.GREEN, Option(gpuOpTime));
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/NvtxWithMetrics.scala
```diff
@@ -18,6 +18,15 @@ package com.nvidia.spark.rapids
 
 import ai.rapids.cudf.{NvtxColor, NvtxRange}
 
+object NvtxWithMetrics {
+  def apply(name: String, color: NvtxColor, metric: Option[GpuMetric] = None): NvtxRange = {
```
jlowe, line 22: Curious why this has a default argument?  The caller should be using a regular `NvtxRange` if they don't have a possible metric to pass as a parameter.

sql-plugin/src/main/java/com/nvidia/spark/rapids/UnsafeRowToColumnarBatchIterator.java
andygrove: Because this is Java code, we have to explicity call the `apply` method here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/NvtxWithMetrics.scala
andygrove: This was added in response to https://github.com/NVIDIA/spark-rapids/pull/2823#discussion_r659802704

andygrove: I see what you mean ... I will remove the default.

jlowe (approved):

jlowe:
build

andygrove:
build

revans2 (approved):

PR #2829, opened 2021-06-28
Fix filtering directories on compression extension match

tgravescs:
If you have a directory with a "." in it, then it matches on our filter to check for filename extensions for compressed files.

Change the checks to make sure to only look at file name extensions when its a file.

Note this was caught by our integration tests so no test added here.

Signed-off-by: Thomas Graves <tgraves@apache.org>

### DIFF
--- a/tools/src/main/scala/com/nvidia/spark/rapids/tool/EventLogPathProcessor.scala
+++ b/tools/src/main/scala/com/nvidia/spark/rapids/tool/EventLogPathProcessor.scala
@@ -92,7 +92,7 @@ object EventLogPathProcessor extends Logging {
       val filePath = fileStatus.getPath()
       val fileName = filePath.getName()
 
-      if (!eventLogNameFilter(filePath)) {
+      if (fileStatus.isFile() && !eventLogNameFilter(filePath)) {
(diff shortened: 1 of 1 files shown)


tgravescs:
build

gerashegalov (approved):
LGTM, 
  - consider a unit test for the regression

tools/src/main/scala/com/nvidia/spark/rapids/tool/EventLogPathProcessor.scala
```diff
@@ -112,14 +112,15 @@ object EventLogPathProcessor extends Logging {
         val (validLogs, invalidLogs) = fs.listStatus(inputPath).partition(s => {
             val name = s.getPath().getName()
             (s.isFile ||
-              (s.isDirectory &&
-                (isEventLogDir(name) || isDatabricksEventLogDir(s, fs))))
+              (s.isDirectory && (isEventLogDir(name) || isDatabricksEventLogDir(s, fs))))
```
gerashegalov, line 115: looks like a purely cosmetic change that we could keep out of the bug fix.

tgravescs:
> Note this was caught by our integration tests so no test added here.

Issue #2786, opened 2021-06-23
[BUG][initCap function]There is an issue converting the uppercase character to lowercase on GPU.


smone123:
**Describe the bug**
While execute initCap function on GPU, if the input string has the digital number in it, for example 'spar2Rk' ,  the output result string is not expected. 
The current result is "Spar2Rk", didn't convert the uppercase character to lowercase behind digital number.

**Steps/Code to reproduce bug**
```python
def createDf1():
   print("### CREATE DATAFRAME 1 ####")
   schema = StructType([
                        StructField("byteF", ByteType()),
                        StructField("shortF", ShortType()),
                        StructField("intF", IntegerType()),
                        StructField("longF", LongType()),
                        StructField("floatF", FloatType()),
                        StructField("doubleF", DoubleType()),
                        StructField("booleanF", BooleanType()),
                        StructField("strF", StringType()),
                        StructField("decimalF", DecimalType(5,2)),
                        StructField("timestampF", TimestampType()),
                        StructField("dateF", DateType())
                      ])
   dt = datetime.date(1990, 1, 1)
   tm = datetime.datetime(2020,2,1,12,1,1)
   dcm1 = Decimal('111.11')
   dcm2 = Decimal('222.11')
   data = [
           (10, 700, 1000, 30, None, 3.000013, True, "nVIDIA inc", dcm2, None, dt),
           (10, 500, 2000, 30, None, 3.000013, True, "n2VIDIA inc", dcm2, None, dt),
          ]
   df = spark.createDataFrame(data, schema)
   df.createOrReplaceTempView("test_table1")
   df.show()
if __name__ == "__main__":
   spark = SparkSession.builder.appName("sparktest").getOrCreate()
   createDf1()
   sql_query_line_list=[
                       "SELECT initcap(strF) FROM test_table1",
                       ]
   for sql_query_line in sql_query_line_list:
       # enable CPU
       spark.conf.set("spark.rapids.sql.enabled", "false")
       print("CPU Physical Plan")
       spark.sql(sql_query_line).explain()
       cpu_result = spark.sql(sql_query_line).collect()
       # enable GPU
       spark.conf.set("spark.rapids.sql.enabled", "true")
       print("GPU Physical Plan")
       spark.sql(sql_query_line).explain()
       gpu_result = spark.sql(sql_query_line).collect()
       #sort result
       cpu_result.sort(key=testlib._RowCmp)
       gpu_result.sort(key=testlib._RowCmp)
[...]

Issue #120, opened 2020-06-05
[BUG] GPU InitCap supports too much white space.


revans2:
**Describe the bug**
Spark appears to only support space ' ', where as cudf supports a lot more than that.

**Steps/Code to reproduce bug**
A python integration test is being added for this.

We might just mark this as incompatible by default.

PR #2838, opened 2021-06-29
Replace `toTitle` with `capitalize` for GpuInitCap

firestarman:
Replace `toTitle` with `capitalize` for GpuInitCap to align with the behavior of Spark `InitCap`.

closes https://github.com/NVIDIA/spark-rapids/issues/2786
closes https://github.com/NVIDIA/spark-rapids/issues/120


firestarman:
This PR depends on https://github.com/rapidsai/cudf/pull/8624, so it is WIP, because the doc can not be updated due to build failure.

integration_tests/src/main/python/string_test.py
```diff
@@ -314,9 +314,6 @@ def test_length():
                 'CHAR_LENGTH(a)',
                 'CHARACTER_LENGTH(a)'))
 
-# Once the xfail is fixed this can replace test_initcap_space
-@incompat
-@pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/120')
 def test_initcap():
```
jlowe, line 317: This test should include more characters than just `[aAbB]`.  Failing to test a wider character range in the words is what led to #2786 being missed during testing.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1250,10 +1250,7 @@ object GpuOverrides {
       ExprChecks.unaryProjectNotLambdaInputMatchesOutput(TypeSig.STRING, TypeSig.STRING),
       (a, conf, p, r) => new UnaryExprMeta[InitCap](a, conf, p, r) {
         override def convertToGpu(child: Expression): GpuExpression = GpuInitCap(child)
-      }).incompat(CASE_MODIFICATION_INCOMPAT + " Spark also only sees the space character as " +
-      "a word deliminator, but this will capitalize any character after a non-alphabetic " + 
-      "character.  The behavior will be aligned to match Spark in the future per " +
-      "https://github.com/NVIDIA/spark-rapids/issues/2786."),
+      }).incompat(CASE_MODIFICATION_INCOMPAT),
```
jlowe, line 1253: Is this still true?  `CASE_MODIFICATION_INCOMPAT` refers to https://github.com/rapidsai/cudf/issues/3132 which is a closed issue, and looking at the code it appears characters that change width are handled properly.  There may be some issues with corner-case Unicode characters based on which Unicode standard libcudf is using vs. the JVM is using, but I suspect the width-changing issue is no longer the problem.

firestarman: I spent some time to catch up with the story. 
Yes, There may be some issues caused by the different Unicode standards used by licudf and JVM respectively. Maybe we need another incompat message for it.
But here just removed this incompat message.

integration_tests/src/main/python/string_test.py
firestarman: Updated

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2201,15 +2194,13 @@ object GpuOverrides {
       ExprChecks.unaryProjectNotLambdaInputMatchesOutput(TypeSig.STRING, TypeSig.STRING),
       (a, conf, p, r) => new UnaryExprMeta[Upper](a, conf, p, r) {
         override def convertToGpu(child: Expression): GpuExpression = GpuUpper(child)
-      })
-      .incompat(CASE_MODIFICATION_INCOMPAT),
```
jlowe, line 2205: If there are still some known issues, should this PR be removing them from the incompatible list?  I'm thinking the reason string needs to be updated to document the current known issues rather than point to an old, resolved cudf issue.

integration_tests/src/main/python/string_test.py
```diff
@@ -314,23 +314,12 @@ def test_length():
...
-    gen = mk_str_gen('([aAbB]{0,5}[ \r\n\t]{1,2}){1,5}')
-    assert_gpu_and_cpu_are_equal_collect(
-            lambda spark: unary_op_df(spark, gen).select(
-                f.initcap(f.col('a'))))
-
-@incompat
-def test_initcap_space():
-    # we see a lot more space delim
-    gen = StringGen('([aAbB]{0,5}[ ]{1,2}){1,5}')
+    gen = mk_str_gen('([aAbB1357_@%-ȺéŉŸ]{0,16}[ \r\n\t]{1,2}){1,5}')
```
jlowe, line 322: Related to the other comment about incompat, the fact that we need to carefully limit the input character set shows that this isn't truly a compatible operation.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
firestarman: Got it. Updated
I planned to add this new doc in another PR but yes, seems it is not a good idea to remove the incompat tag here.

integration_tests/src/main/python/string_test.py
firestarman: Yes, may I know any issue for this change ?
I thought we should cover at least some of the characters mentioned in the bug https://github.com/rapidsai/cudf/issues/3132 .

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -433,9 +433,8 @@ object GpuOverrides {
     "when enabling these, there may be extra groups produced for floating point grouping " +
     "keys (e.g. -0.0, and 0.0)"
   val CASE_MODIFICATION_INCOMPAT =
-    "in some cases unicode characters change byte width when changing the case. The GPU string " +
-    "conversion does not support these characters. For a full list of unsupported characters " +
-    "see https://github.com/rapidsai/cudf/issues/3132"
+    "it will not be always 100% compatible with Spark because the unicode standards used by" +
+    " cuDF and JVM may differ."
```
jlowe, line 437: No need to explain it's incompatible, that is already covered elsewhere and redundant to put here.  This should be simplified to something like:
```suggestion
    "the Unicode standards used by cuDF and the JVM may differ, resulting in some " +
    "corner-case characters not changing case correctly."
```

jlowe (commented):
> I thought we should cover at least some of the characters mentioned in the bug

Yes, many characters are now covered.  IIUC the only issue is that there are some corner-case characters that could be mishandled based on characters that were added to the Unicode standard between versions.  If cudf and JVM aren't targeting the same version of the Unicode standard then those characters occurring in the string can cause a mismatch between GPU and CPU output.

firestarman:
build

firestarman:
Filed the bug https://github.com/rapidsai/cudf/issues/8644 to track the failures for some special characters.

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1250,10 +1249,10 @@ object GpuOverrides {
...
       (a, conf, p, r) => new UnaryExprMeta[InitCap](a, conf, p, r) {
         override def convertToGpu(child: Expression): GpuExpression = GpuInitCap(child)
-      }).incompat(CASE_MODIFICATION_INCOMPAT + " Spark also only sees the space character as " +
-      "a word deliminator, but this will capitalize any character after a non-alphabetic " + 
-      "character.  The behavior will be aligned to match Spark in the future per " +
-      "https://github.com/NVIDIA/spark-rapids/issues/2786."),
+      }).incompat("in some cases Unicode characters change byte width when changing the case." +
+      " The GPU string conversion does not support these characters for capitalize now. This" +
+      " will be fixed in the future per https://github.com/rapidsai/cudf/issues/8644. For a" +
+      " full list of unsupported characters see https://github.com/rapidsai/cudf/issues/3132."),
```
jlowe, line 1255: This still refers to changing byte widths which has been fixed.  That's no longer the issue.  Now the issue is that some characters are not converted either due to a bug, like the issue you filed, or because the Unicode standard is different than the JVM.  We should drop the text about byte width changes along with pointing at issue 3132, IMO.

firestarman: Per the fix https://github.com/rapidsai/cudf/pull/8647, seems the byte widths issue still exists for `capitalize` and `title`.
I will update this after the fix being merged.

firestarman:
build

integration_tests/src/main/python/string_test.py
```diff
@@ -320,15 +320,15 @@ def test_initcap():
...
+    gen = mk_str_gen('([aAbB1357ȺéŸ_@%-]{0,15}[ \r\n\t]{1,2}){1,5}')
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark: unary_op_df(spark, gen).select(
                 f.initcap(f.col('a'))))
 
 @incompat
-@pytest.mark.xfail(reason='https://github.com/rapidsai/cudf/issues/8644')
-def test_initcap_width_change():
-    gen = mk_str_gen('ŉ([aAbB13ŉȺéŉŸ]{0,5}){1,5}')
+@pytest.mark.xfail(reason='Spark initcap will not convert ŉ to ʼN')
```
jlowe, line 329: This is fixed by https://github.com/rapidsai/cudf/pull/8647 and should no longer be xfailed.  It's also weird to mark a test as xfailed but then not test the condition that fails?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
jlowe: That PR notes that it is using the same logic as upper and lower, and we can see in the code that [it is checking the number of bytes](https://github.com/rapidsai/cudf/blob/branch-21.08/cpp/src/strings/capitalize.cu#L85) for the new character.  Therefore the only issue remaining should be whether all of the characters in the conversion table match what the JVM expects (i.e.: Unicode version issues).

integration_tests/src/main/python/string_test.py
firestarman: This test is still failing. 
Because of the fix (rapidsai/cudf#8647), GPU initcap will convert the first  `ŉ` to `ʼN`, but CPU initcap will not, as the reason string said.

Log snip from the pre-merge build.
```Bash
[2021-07-05T08:22:24.286Z]   opposite null order not supported
[2021-07-05T08:22:24.286Z] XFAIL ../../src/main/python/string_test.py::test_initcap_special_chars[INCOMPAT]
[2021-07-05T08:22:24.286Z]   Spark initcap will not convert ŉ to ʼN
```

jlowe (approved):

PR #2841, opened 2021-06-29
Add a threadpool to Qualification tool to process logs in parallel

tgravescs:
Add a threadpool and parameters to configure timeout and number of threads (defaults to cores on host/4).
Updated some log messages so it shows path or something useful since now the log messages don't come out in sequential order.

Fixed a bug with printing out the text file when there aren't any apps in QualOutputWriter.

Times to process 105 log files went from about 6.5 minutes to 45 seconds, interestingly enough, processing 210 log files only takes a bit longer at like 50 seconds, it looks like a few of the log files are skewed and take much longer than others.

### DIFF
--- a/tools/src/main/scala/com/nvidia/spark/rapids/tool/qualification/Qualification.scala
+++ b/tools/src/main/scala/com/nvidia/spark/rapids/tool/qualification/Qualification.scala
@@ -49,15 +49,13 @@ class Qualification(outputDir: String, numRows: Int, hadoopConf: Configuration,
   }
 
   def qualifyApps(allPaths: Seq[EventLogInfo]): Seq[QualificationSummaryInfo] = {
-    try {
-      allPaths.foreach(path => threadPool.submit(new QualifyThread(path)))
-    } catch {
-      case e: Exception =>
-        logError("Exception processing event logs", e)
-        threadPool.shutdown()
-        if (!threadPool.awaitTermination(5, TimeUnit.SECONDS)) {
-          threadPool.shutdownNow()
-        }
+    allPaths.foreach { path =>
+      try {
+        threadPool.submit(new QualifyThread(path))
+      } catch {
+        case e: Exception =>
+          logError(s"Unexpected exception submitting log ${path.eventLog.toString}, skipping!", e)
+      }
     }
     // wait for the threads to finish processing the files
     threadPool.shutdown()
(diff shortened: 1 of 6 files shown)


tgravescs:
build

revans2 (approved):

tools/src/main/scala/com/nvidia/spark/rapids/tool/qualification/Qualification.scala
```diff
@@ -28,31 +31,71 @@ import org.apache.spark.sql.rapids.tool.qualification._
...
-    val allAppsSum: ArrayBuffer[QualificationSummaryInfo] = ArrayBuffer[QualificationSummaryInfo]()
+  private val allApps = new ConcurrentLinkedQueue[QualificationSummaryInfo]()
+  // default is 24 hours
+  private val waitTimeInSec = timeout.getOrElse(60 * 60 * 24L)
+
+  private val threadFactory = new ThreadFactoryBuilder()
+    .setDaemon(true).setNameFormat("qualTool" + "-%d").build()
+  logInfo(s"Threadpool size is $nThreads")
+  private val threadPool = Executors.newFixedThreadPool(nThreads, threadFactory)
+    .asInstanceOf[ThreadPoolExecutor]
```
gerashegalov, line 45: nit: redundant cast?

tools/src/main/scala/com/nvidia/spark/rapids/tool/qualification/Qualification.scala
```diff
@@ -28,31 +31,71 @@ import org.apache.spark.sql.rapids.tool.qualification._
...
+          logError(s"Unexpected exception submitting log ${path.eventLog.toString}, skipping!", e)
+      }
+    }
+    // wait for the threads to finish processing the files
+    threadPool.shutdown()
+    if (!threadPool.awaitTermination(waitTimeInSec, TimeUnit.SECONDS)) {
+      logError(s"Processing log files took longer then $waitTimeInSec seconds," +
+        " stopping processing any more event logs")
+      threadPool.shutdownNow()
+    }
```
gerashegalov, line 66: we may get an InterruptedException in awaitTermination, may want to surround it with try/finally and place shutdownNow into finally.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/qualification/QualAppInfo.scala
```diff
@@ -197,18 +197,18 @@ object QualAppInfo extends Logging {
...
+          logWarning(s"Error parsing JSON: ${path.eventLog.toString}")
           None
         case il: IllegalArgumentException =>
-          logWarning(s"Error parsing file: $path", il)
+          logWarning(s"Error parsing file: ${path.eventLog.toString}", il)
           None
         case e: Exception =>
           // catch all exceptions and skip that file
-          logWarning(s"Got unexpected exception processing file: $path", e)
+          logWarning(s"Got unexpected exception processing file: ${path.eventLog.toString}", e)
```
gerashegalov, line 211: nit: toString is redundant

gerashegalov (commented):
nice speedup

PR #2842, opened 2021-06-29
Support STRUCT/STRING for LEAD()/LAG()

mythrocks:
Added support in `LEAD()`/`LAG()` to process `STRING` and `STRUCT` input columns.

Prior to this commit, `LEAD()`/`LAG()`, when invoked on a `STRING` or `STRUCT` column, do not seem to be GPU accelerated:
```
@Expression <WindowExpression> lag(_w0#21, 1, null) windowspecdefinition(_w1#22, _w2#23 ASC NULLS FIRST, specifiedwindowframe(RowFrame, -1, -1)) could run on GPU
        !Expression <Lag> lag(_w0#21, 1, null) cannot run on GPU because expression Lag lag(_w0#21, 1, null) produces an unsupported type StringType; expression AttributeReference _w0#21 produces an unsupported type StringType
```

This commit should enable `STRING` and `STRUCT`, alongside nested variants (e.g. `ARRAY<STRING>`, `ARRAY<ARRAY<STRUCT<STRING>>>`, etc.)


mythrocks:
build

mythrocks:
Note: It appears to have been an oversight when `ARRAY` support was added `LEAD()`/`LAG()`, on the plugin; all nested types should've worked at that point.

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/Spark311Shims.scala
```diff
@@ -204,29 +204,45 @@ class Spark311Shims extends Spark301Shims {
     // Spark 3.1.1-specific LEAD expression, using custom OffsetWindowFunctionMeta.
     GpuOverrides.expr[Lead](
       "Window function that returns N entries ahead of this one",
-      ExprChecks.windowOnly((TypeSig.numeric + TypeSig.BOOLEAN +
-        TypeSig.DATE + TypeSig.TIMESTAMP + TypeSig.ARRAY).nested(), TypeSig.all,
-        Seq(ParamCheck("input", (TypeSig.numeric + TypeSig.BOOLEAN +
-          TypeSig.DATE + TypeSig.TIMESTAMP + TypeSig.ARRAY).nested(), TypeSig.all),
+      ExprChecks.windowOnly(
+        (TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL +
+          TypeSig.ARRAY + TypeSig.STRUCT).nested(),
```
revans2, line 209: This is adding in structs as well as string, but I didn't see any tests for structs, or arrays of structs or structs with arrays in them/etc.

mythrocks: In [window_function_test.py](https://github.com/NVIDIA/spark-rapids/pull/2842/files#diff-1d2ae10fd5085346e0f2055455ac2f881d950ed8215d61e3cb708917ed1d41bcR128), `lead_lag_data_gens` was modified thus:
```python
lead_lag_data_gens = [long_gen, DoubleGen(no_nans=True, special_cases=[]),
        boolean_gen, timestamp_gen, string_gen, DecimalGen(precision=18, scale=3),
        StructGen(children=[
            ['child_int', IntegerGen()],
            ['child_time', DateGen()],
            ['child_string', StringGen()]
        ])]
```
Also, `lead_lag_array_data_gens` generates `ARRAY` versions of the data, from `lead_lag_data_gens`. So the `ARRAY<STRUCT>` test is covered when `test_window_aggs_for_rows_lead_lag_on_arrays` runs.

That said, I should add an `ARRAY` type to the `StructGen` in `lead_lag_data_gens`.

mythrocks: As an aside, it appears that the `Struct` tests seem to be failing at the moment, while they weren't last night. I'll add the `Struct<Array>` test as soon as that's sorted out.

mythrocks: Hmm... This is interesting. It's failing because one of the tests is to specify a default value for `LAG()`. I can't seem to figure out how to specify a `STRUCT` literal in SQL. The following doesn't work:
`LAG(STRUCT(dollars, dateStr), 3, STRUCT(CAST(1 AS INT),'') )`:
```
org.apache.spark.sql.AnalysisException: cannot resolve 'lag(struct(mytable.`dollars`, mytable.`dateStr`), 3, struct(CAST(1 AS INT), ''))' due to data type mismatch: argument 3 requires (struct<dollars:int,dateStr:string> or null) type, however, 'struct(CAST(1 AS INT), '')' is of struct<col1:int,col2:string> type.; line 1 pos 22;
```

mythrocks: I am inclined to use `None` for `STRUCT` column `LAG()` defaults, because `[1, "two"]` isn't a valid `STRUCT` literal.

Another interesting problem to sort out is why `SortOrder` does not support ordering on `STRUCT` columns in windowing contexts:
```
!Expression <WindowSpecDefinition> windowspecdefinition(uid#81L, dateLong#83L ASC NULLS FIRST, _w2#95 ASC NULLS FIRST, specifiedwindowframe(RowFrame, -3, -3)) cannot run on GPU because expression SortOrder _w2#95 ASC NULLS FIRST produces an unsupported type StructType(StructField(dollars,IntegerType,true), StructField(dateStr,StringType,true))
```

gerashegalov: @mythrocks I think it's because `WindowSpecDefinition` https://github.com/NVIDIA/spark-rapids/blob/b3ae804b76c2b6c2fa396c17d7a7be1a8a32b1c3/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala#L874-L878
is configured to deal with nested outputs https://github.com/NVIDIA/spark-rapids/blob/b3ae804b76c2b6c2fa396c17d7a7be1a8a32b1c3/sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala#L803-L805

mythrocks: I'm past that now. 
```scala
object WindowSpecCheck extends ExprChecks {
  val check: TypeSig =
    TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL +
      TypeSig.STRUCT.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL)
```
The new problem I'm wrestling with is whether to allow a `STRUCT` order-by column for `RANGE` queries. For instance, the window function tests currently have the equivalent of the following:
```sql
SELECT MIN(my_struct) OVER (PARTITION BY my_id ORDER BY my struct) ...
```
This translates to a `RANGE` query `(UNBOUNDED, CURRENT ROW)`.
Technically, sorting on a `STRUCT` column can be GPU accelerated, IFF the column isn't further nested. I'm inclined to enable this sorting in window functions, for this narrow case.

I'll update here when I have this going.

mythrocks:
In light of some complications in the tests, I'm moving this PR back to `DRAFT`. I'll update here when this is sorted out.

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/Spark311Shims.scala
mythrocks: P.S. Thanks for reviewing/investigating, @gerashegalov.

mythrocks: I'm bailing on this. The scope is ballooning, and there would need to be changes in the `cudf::rolling_window()` to support this for structs. I'll redo the test to skip `min()`/`max()` operations for structs.

mythrocks:
Fixed the `STRUCT` test failures.

mythrocks:
build

mythrocks:
I'm inclined to leave this change as it currently stands. Attempting to add `STRUCT<ARRAY>` tests results in the following erroneous "failure":
```
>   ???
E   pyspark.sql.utils.IllegalArgumentException: Part of the plan is not columnar class org.apache.spark.sql.execution.SortExec
E   Sort [a#55 ASC NULLS FIRST, b#56 ASC NULLS FIRST, c#57 ASC NULLS FIRST, lead_5_c#61 ASC NULLS FIRST], true, 0
E   +- Exchange rangepartitioning(a#55 ASC NULLS FIRST, b#56 ASC NULLS FIRST, c#57 ASC NULLS FIRST, lead_5_c#61 ASC NULLS FIRST, 12), true, [id=#259]
E      +- GpuColumnarToRow false
E         +- GpuWindow [gpulead(c#57, 5, null) gpuwindowspecdefinition(a#55, b#56 ASC NULLS FIRST, gpuspecifiedwindowframe(RowFrame, 5, 5)) AS lead_5_c#61], [a#55], [b#56 ASC NULLS FIRST], false
E            +- GpuCoalesceBatches batchedbykey(a#55 ASC NULLS FIRST)
E               +- GpuSort [a#55 ASC NULLS FIRST, b#56 ASC NULLS FIRST], false, com.nvidia.spark.rapids.OutOfCoreSort$@33b465ab
E                  +- GpuShuffleCoalesce 1073741824
E                     +- GpuColumnarExchange gpuhashpartitioning(a#55, 12), true, [id=#253]
E                        +- GpuRowToColumnar targetsize(1073741824)
E                           +- Scan ExistingRDD[a#55,b#56,c#57]
```
Note that the window-function for `struct<array>` is executed on GPU. But when the validation comparison attempts complete ordering, it fails to sort the aggregation output (i.e. `struct<array>`).

mythrocks:
(Just realized that I hadn't moved this PR out from `Draft`. Rectified.)

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -801,7 +801,9 @@ object CaseWhenCheck extends ExprChecks {
  * This is specific to WidowSpec, because it does not follow the typical parameter convention.
  */
 object WindowSpecCheck extends ExprChecks {
-  val check: TypeSig = TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL
+  val check: TypeSig =
+    TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL +
+      TypeSig.STRUCT.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL + TypeSig.NULL)
```
revans2, line 806: So we only support single level structs for lead/lag?

revans2 (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
mythrocks: Not at all, sir.

`Struct<Struct>`:
```scala
scala> val window = " partition by uid order by dt ";
scala> sql(s" select  lead(struct(struct(dollars, dateStr)), 3) over ($window) lead_3 from mytable ").explain
GpuColumnarToRow false
+- GpuProject [lead_3#178]
   +- GpuWindow [_w0#179, uid#144L, dt#147, gpulead(_w0#179, 3, null) gpuwindowspecdefinition(uid#144L, dt#147 ASC NULLS FIRST, gpuspecifiedwindowframe(RowFrame, 3, 3)) AS lead_3#178], [uid#144L], [dt#147 ASC NULLS FIRST]
      +- GpuCoalesceBatches batchedbykey(uid#144L ASC NULLS FIRST)
         +- GpuSort [uid#144L ASC NULLS FIRST, dt#147 ASC NULLS FIRST], false, com.nvidia.spark.rapids.OutOfCoreSort$@7342eb53
            +- GpuShuffleCoalesce 2147483647
               +- GpuColumnarExchange gpuhashpartitioning(uid#144L, 200), ENSURE_REQUIREMENTS, [id=#537]
                  +- GpuProject [named_struct(col1, named_struct(dollars, dollars#149, dateStr, dateStr#148)) AS _w0#179, uid#144L, dt#147]
                     +- GpuFileGpuScan orc [uid#144L,dt#147,dateStr#148,dollars#149] Batched: true, DataFilters: [], Format: ORC, Location: InMemoryFileIndex[file:/home/mithunr/workspace/Demos/windowbench/data/gs/1/part-00000.snappy.orc], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<uid:bigint,dt:date,dateStr:string,dollars:int>
```

`Struct<Array>`:
```scala
scala> sql(s" select  lead(struct(array(dollars)), 3) over ($window) lead_3 from mytable ").explain
== Physical Plan ==
GpuColumnarToRow false
+- GpuProject [lead_3#191]
   +- GpuWindow [_w0#192, uid#144L, dt#147, gpulead(_w0#192, 3, null) gpuwindowspecdefinition(uid#144L, dt#147 ASC NULLS FIRST, gpuspecifiedwindowframe(RowFrame, 3, 3)) AS lead_3#191], [uid#144L], [dt#147 ASC NULLS FIRST]
      +- GpuCoalesceBatches batchedbykey(uid#144L ASC NULLS FIRST)
         +- GpuSort [uid#144L ASC NULLS FIRST, dt#147 ASC NULLS FIRST], false, com.nvidia.spark.rapids.OutOfCoreSort$@7342eb53
            +- GpuShuffleCoalesce 2147483647
               +- GpuColumnarExchange gpuhashpartitioning(uid#144L, 200), ENSURE_REQUIREMENTS, [id=#632]
                  +- GpuProject [named_struct(col1, array(dollars#149)) AS _w0#192, uid#144L, dt#147]
                     +- GpuFileGpuScan orc [uid#144L,dt#147,dollars#149] Batched: true, DataFilters: [], Format: ORC, Location: InMemoryFileIndex[file:/home/mithunr/workspace/Demos/windowbench/data/gs/1/part-00000.snappy.orc], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<uid:bigint,dt:date,dollars:int>
```

The issue is the same as mentioned [above](https://github.com/NVIDIA/spark-rapids/pull/2842#issuecomment-871753188): The sorting of the outputs for the result comparison causes the test to fail spuriously, because `Sort` multi-level structs isn't on GPU. :/

revans2: `@ignore_order(local=True)` does the sorting in python on the CPU. It works with multi-level structs, arrays, and I think I made it works with maps too.

revans2: So if we can test it will that let you turn it on?

mythrocks: Yes, sir. I'll add back the test and check with `local=true`. Thanks for the tip.

mythrocks: Just confirmed that `@ignore_order(local=True)` works. 

Added a separate test for `struct<array>`, `array<struct<array>`, `struct<struct<array>>`, etc. I didn't fold this back into `test_multi_types_window_aggs_for_rows_lead_lag` because of the `ORDER BY c` clause, which will punt the operator off the GPU.

mythrocks:
build

revans2 (approved):

mythrocks:
Thanks for the review, @revans2. @gerashegalov, any objections to our checking this in?

gerashegalov (approved):
LGTM

mythrocks:
Thanks for the reviews, chaps. I've merged this now.

PR #2844, opened 2021-06-29
Document Qualification tool Spark requirements

tgravescs:
To run the qualification tool, it requires the Spark 3.x jars.  If someone is trying to run it on a node that doesn't have it document how they can get them.

Also add in a couple --help options I forgot to add when adding the threadpool support.


tgravescs:
build

jlowe (approved):

PR #2845, opened 2021-06-29
Split window exec into multiple stages if needed

### LINKED ISSUE #2688: [FEA] Split window expressions into pre, window, and post operations

As a part of https://github.com/NVIDIA/spark-rapids/pull/2596 we added in a memory optimization for running window operations.  But databricks supports essentially anything in the expressions that can be done in a ProjectExec and more.  This makes it difficult for us to actually determine what is and isn't a supported window operation. It also makes it difficult for us to optimize the performance by sending all of the aggregations to the GPU at once so it does not have to calculate the window offsets multiple times.

To fix this we should have some tools that can walk through an expression tree and split it up into a project that would happen before the window operation, all of the window operations, and then finally a project that would happen after all the window operations.

As a part of this we should also deduplicate window operations (mostly needed for running window rank and dense_rank where we may need to insert in a row_number so we can do it without needing to buffer/split on window boundaries).

revans2:
This fixes #2688 

It does not go all the way and change window functions so that they are combined into fewer calls.  I plan on doing that in a follow on PR because this was large enough already. I tested this on databricks.

One of the big advantages to this is that on databricks we now can now optimized for GpuRunningWindowExec even if there are complicated operations.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
@@ -16,6 +16,7 @@
 
 package com.nvidia.spark.rapids
 
+import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 
 import ai.rapids.cudf.Scalar
@@ -24,7 +25,7 @@ import org.apache.spark.TaskContext
 import org.apache.spark.internal.Logging
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.{Ascending, Attribute, AttributeReference, CurrentRow, Expression, NamedExpression, RowFrame, SortOrder, UnboundedPreceding}
+import org.apache.spark.sql.catalyst.expressions.{Ascending, Attribute, AttributeReference, AttributeSet, CurrentRow, Expression, NamedExpression, RowFrame, SortOrder, UnboundedPreceding}
 import org.apache.spark.sql.catalyst.plans.physical.{AllTuples, ClusteredDistribution, Distribution, Partitioning}
 import org.apache.spark.sql.execution.{SparkPlan, UnaryExecNode}
 import org.apache.spark.sql.execution.window.WindowExec
@@ -78,6 +79,9 @@ abstract class GpuBaseWindowExecMeta[WindowExecType <: SparkPlan] (windowExec: W
   val orderSpec: Seq[BaseExprMeta[SortOrder]] =
     getOrderSpecs.map(GpuOverrides.wrapExpr(_, conf, Some(this)))
 
+  lazy val inputFields: Seq[BaseExprMeta[Attribute]] =
+    windowExec.children.head.output.map(GpuOverrides.wrapExpr(_, conf, Some(this)))
+
   override def tagPlanForGpu(): Unit = {
     // Implementation depends on receiving a `NamedExpression` wrapped WindowExpression.
     windowExpressions.map(meta => meta.wrapped)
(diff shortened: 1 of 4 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -181,6 +217,142 @@ class GpuWindowExecMeta(windowExec: WindowExec,
...
+        GpuAlias(child, attr.name)(attr.exprId, attr.qualifier)
+      case other => other
+    }
+  }
+
+  private def hasGpuWindowFunction(expr: Expression): Boolean = {
+    expr.find {
+      case _: GpuWindowExpression => true
+      case _ => false
+    }.isDefined
```
jlowe, line 247:
```suggestion
    expr.find(_.isInstanceOf[GpuWindowExpression]).isDefined
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -181,6 +217,142 @@ class GpuWindowExecMeta(windowExec: WindowExec,
...
+  private def extractAndSave(expr: Expression,
+      saved: ArrayBuffer[NamedExpression],
+      deduped: mutable.HashMap[Expression, Attribute]): Expression =
+    expr match {
+      // Don't rename an already named expression
+      case ne: NamedExpression =>
+        val missingExprId = Set(ne.exprId) -- saved.map(_.exprId)
+        if (missingExprId.nonEmpty) {
+          saved += ne
+        }
```
jlowe, line 259: Nit: I found the set operations on set-of-1 to be a bit odd.  The following has the benefit of early out:
```suggestion
        if (!saved.exists(_.exprId == ne.exprId)) {
          saved += ne
        }
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -181,6 +217,142 @@ class GpuWindowExecMeta(windowExec: WindowExec,
...
+   * (_tmp1 - _tmp2) as result
+   * </pre>
+   *
+   * This assumes that there is not a window function of another window function, like
+   * `LAG(SUM(a), 2)` which appears to be something all distros split apart into separate
+   * window operations, so we are good.
+   * @param exprs the input expressions to a GpuWindowExec
+   */
+  def splitAndDedup(exprs: Seq[NamedExpression]):
+  (Seq[NamedExpression], Seq[NamedExpression], Seq[NamedExpression]) = {
```
jlowe, line 304: Nit: Instead of a raw tuple consider using a private `case class` to track output parameters.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -181,6 +217,142 @@ class GpuWindowExecMeta(windowExec: WindowExec,
...
+    val shims = ShimLoader.getSparkShims
+
+    exprs.foreach { expr =>
+      if (hasGpuWindowFunction(expr)) {
+        // First pass looks for GpuWindowFunctions and GpuWindowSpecDefinitions to build up
+        // the preProject phase
+        val firstPass = expr.transformDown {
+          case wf: GpuWindowFunction =>
+            // All window functions, including those that are also aggregation functions, are
+            // wrapped in a GPuWindowExpression, so dedup and save their children into the pre
```
jlowe, line 322:
```suggestion
            // wrapped in a GpuWindowExpression, so dedup and save their children into the pre
```

jlowe (commented):
Would be good to have a windowing expert take a look at this as well, but overall it looks good to me.  Only minor comments.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
revans2: I just don't see it for a single API that is going to wrap a return result and then unwrap it immediately.

```
val (pre, windowOps, post) = GpuWindowExec.splitAndDedup(gpuWindowExpressions)
```

vs

```
GpuSplitWindowExpressions(pre, windowOps, post) = GpuWindowExec.splitAndDedup(gpuWindowExpressions)
```

Scala will already hide the names when creating it, and likely unapplying it like I would probably do. If it was going to be passed around a bit more, then I would understand, but I don't see it here.

revans2:
build

jlowe (approved):

revans2:
build

Issue #2504, opened 2021-05-25
[BUG] OOM when running NDS queries with UCX and GDS


rongou:
**Describe the bug**
When running TPC-DS at scale factor 5000 on a Yarn cluster with 8xA100 40 GB GPUs, seeing out-of-memory errors in libcudf code (see full stacktrace below). Probably caused by native code allocating memory while java/scala is actively spilling.

**Steps/Code to reproduce bug**
TPC-DS at scale factor 5000, query 14a/b, 16, 24b.

**Expected behavior**
Should not throw oom errors.

**Environment details (please complete the following information)**
 - Environment location: YARN
 - Spark configuration settings related to the issue
```properties
spark.driver.maxResultSize=2GB
spark.rapids.cudfVersionOverride=true
spark.executor.cores=4
spark.rapids.sql.concurrentGpuTasks=2
spark.executor.resource.gpu.amount=1
spark.task.resource.gpu.amount=0.25
spark.executor.memory=240G
spark.driver.memory=50G
spark.sql.files.maxPartitionBytes=1g
spark.locality.wait=0
spark.shuffle.service.enabled=false
spark.rapids.shuffle.transport.enabled=true
spark.shuffle.manager=com.nvidia.spark.rapids.spark302.RapidsShuffleManager
spark.driver.extraClassPath=$SPARK_RAPIDS_PLUGIN_JAR:$CUDF_JAR:$SPARK_RAPIDS_PLUGIN_INTEGRATION_TEST_JAR
spark.executor.extraClassPath=$SPARK_RAPIDS_PLUGIN_JAR:$CUDF_JAR
spark.executorEnv.UCX_RNDV_SCHEME=put_zcopy
spark.executorEnv.UCX_ERROR_SIGNALS=
spark.executorEnv.UCX_IB_GPU_DIRECT_RDMA=yes
spark.executorEnv.UCX_IB_RX_QUEUE_LEN=1024
spark.rapids.shuffle.ucx.bounceBuffers.size=4MB
spark.rapids.shuffle.ucx.bounceBuffers.device.count=64
spark.rapids.shuffle.ucx.bounceBuffers.host.count=64
spark.rapids.memory.host.spillStorageSize=32G
spark.sql.adaptive.enabled=true
spark.plugins=com.nvidia.spark.SQLPlugin
spark.rapids.memory.pinnedPool.size=8g
spark.rapids.shuffle.maxMetadataSize=1MB
[...]

PR #2846, opened 2021-06-29
default rmm alloc fraction to the max to avoid unnecessary fragmentation

rongou:
With the current default of `0.9` for `spark.rapids.memory.gpu.allocFraction`, we first allocate 90% of GPU memory, and then when we run out of memory additional GPU memories are allocated. Because of how CUDA Unified Virtual Addressing (UVA) is set up, subsequently allocated addresses are discontinuous, causing unnecessary fragmentation.

Helps with #2504 but doesn't completely fixes it. Having a single continuous GPU address space also helps with debugging and visualization. 


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -330,7 +330,7 @@ object RapidsConf {
       s"configured via $RMM_ALLOC_MAX_FRACTION_KEY.")
     .doubleConf
     .checkValue(v => v >= 0 && v <= 1, "The fraction value must be in [0, 1].")
-    .createWithDefault(0.9)
+    .createWithDefault(1)
```
jlowe, line 333: Changing this default means a new warning now appears on startup with the default configs.  For example:
```
21/06/30 14:38:36 WARN GpuDeviceManager: Initial RMM allocation (15568.75 MB) is larger than the adjusted maximum allocation (15136.5 MB), lowering initial allocation to the adjusted maximum allocation.
```
It would be nice to not emit warnings with the default configs on most setups.  We should consider whether this should be a warning or just an info, or maybe change the conditions under which it is a warning.

revans2: I would like to see us to take a deep look at how we are calculating the RMM pool size etc. 

We have several configs associated with this.
 * `spark.rapids.memory.gpu.allocFraction` - 0 to 1 for the percentage of free memory allocated for the RMM pool, used to be percentage of total memory.
 * `spark.rapids.memory.gpu.maxAllocFraction` - 0 to 1 maximum amount of total GPU memory that can be allocated for the RMM pool. (except we also remove from it a reserved amount)
 * `spark.rapids.memory.gpu.minAllocFraction` - 0 to 1 minimum amount of total GPU memory that can be allocated for the RMM pool before we throw an error saying it is too small.
 * `spark.rapids.memory.gpu.reserve` - number of bytes that have to be reserved for cuda/cudf to load kernels etc so we don't crash because RMM has everything allocated.

The code that does the warnings is a bit confusing too. 

* If initial allocation < min allocation throw an exception
* if max allocation < min allocation throw an exception
* if reserve amount >= max allocation throw an exception (not really sure what this is doing at all. If max and min are small but reserve is big that should be OK so long as everything fits.
* adjust max allocation by subtracting reserve amount from it.
* if adjusted max allocation <= initial allocation output a warning.

So to me it feels like we should instead.

* get the reserved amount
* get the amount of available memory on the GPU by taking free and subtracting reserved from it.
* min allocation is calculated from the total memory on the GPU
* if available is <= min allocation throw an exception this is not going to work.
* max allocation is a percentage of the available memory, not just the free memory
* initial allocation is a percentage of the available memory.
* if initial allocation > max allocation output a warning and adjust initial allocation down to max allocation.

We will need to update the docs in the config section accordingly to better explain what is happening.

sameerz:
Please retarget to the 21.10 branch when it is available.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
rongou: OK, changed how initial allocation is calculated to take into account the reserve amount. To me this seems to make sense since the most you can allocate is `free-reserve`.

revans2 (approved):
This looks good to me.

jlowe (approved):

jlowe:
build

PR #2852, opened 2021-06-30
Fix hash aggregate tests leaking configs into other tests

jlowe:
While working on changes to hash aggregate, I noticed that some test-specific configs were leaking into other tests because a reused `SparkConf` instance was being modified directly with test-specific settings.  This fixes the tests to make a copy of the configs before updating it with specific settings.

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/HashAggregatesSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/HashAggregatesSuite.scala
@@ -271,7 +271,7 @@ class HashAggregatesSuite extends SparkQueryCompareTestSuite {
       "avg literals dbl_max",
       longsFromCSVDf,
       maxFloatDiff = 0.0001,
-      conf = floatAggConf.set(RapidsConf.INCOMPATIBLE_OPS.key, "true")) {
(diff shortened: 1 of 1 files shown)


jlowe:
build

revans2 (dismissed):

kuhushukla (approved):

tests/src/test/scala/com/nvidia/spark/rapids/HashAggregatesSuite.scala
```diff
@@ -271,7 +271,7 @@ class HashAggregatesSuite extends SparkQueryCompareTestSuite {
       "avg literals dbl_max",
       longsFromCSVDf,
       maxFloatDiff = 0.0001,
-      conf = floatAggConf.set(RapidsConf.INCOMPATIBLE_OPS.key, "true")) {
+      conf = floatAggConf.clone.set(RapidsConf.INCOMPATIBLE_OPS.key, "true")) {
```
gerashegalov, line 274: we could make this a one line fix by turning val into `def floatAggConf`.  It will also make it less likely for a new test to re-introduce the bug.

jlowe:
build

gerashegalov (approved):

PR #2853, opened 2021-06-30
Tag last batch when coalescing

jlowe:
Updates the coalesce iterator to tag the last batch being emitted.  This allows code using the iterator to detect the last batch from the iterator without relying on `hasNext` which often has the unfortunate side-effect of manifesting another batch in memory.


jlowe:
build

revans2 (approved):

PR #2854, opened 2021-06-30
Qualification Tool to output only the 'k' highest-ranked or 'k' lowest-ranked applications	

tgravescs:
This is adding support to the Qualification tool to output only the 'k' highest-ranked or 'k' lowest-ranked applications.

This is done with 2 different configs, 1 existing one was for numOutputRows, which was changed slightly to only affect number of output rows in the report text file and the other 1 is a new config called --order <asc|desc> which affects the sort order of the report.

with those 2 configs you can accomplish the  'k' highest-ranked or 'k' lowest-ranked applications.


tgravescs:
build

tools/src/main/scala/com/nvidia/spark/rapids/tool/qualification/QualificationArgs.scala
```diff
@@ -71,3 +81,13 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
 
   verify()
 }
+
+object QualificationArgs {
+  def isOrderAsc(order: String): Boolean = {
+    order.toLowerCase.startsWith("asc")
```
nartal1, line 87: Nit: Should we have "equals" instead of "startsWith" here? 
Or the error message could be updated ""Error, the order must either be desc or asc" above. 
I am fine merging the current commit as well as this is minor.

nartal1 (approved):
Just 1 nit. I am fine merging this in.

tools/src/main/scala/com/nvidia/spark/rapids/tool/qualification/QualificationArgs.scala
tgravescs: I intentionally left it as startsWith in case someone actually typed ascending it would still match and work.

Issue #2527, opened 2021-05-27
[FEA] clean up timing metrics and make them consistent


revans2:
For now I am marking this as a task, but it might turn into an epic.

The goal is to expose timing metrics to the end user that are simple to understand and useful.  Along with this we should have a good way to calculate these metrics so they don't become inconsistent or difficult to deal with.

The proposed metrics are

|*NAME*|*LEVEL*|*DESCRIPTION*|
|-----------|------------|----------------------|
|`gpuOpTime`|MODERATE|The amount of time spent with the GPU semaphore held, ideally doing some kind of computation, but not totally a requirement. This will include time spent spilling data `spillWriteTime` to make room, and time spend reading back in spilled data `spillReadTime`. This is not going to be simple to measure directly because with spilling the semaphore can be acquired when trying to read a batch. So this will be calculated from `opTime` and `semaphoreWaitTime`|
|`opTime`|DEBUG|The amount of time spent either waiting for the GPU semaphore or actually doing computation. It should not include any calls to `hasNext`, `next`, or time spent reading data/writing data as a part of shuffle or to get input to computation/write the result of computation.|
|`semaphopreWaitTime`|DEBUG|The amount of time spent by this part of the plan stuck waiting to acquire the GPU semaphore. |
|`spillWriteTime`|DEBUG|The amount of time spent spilling data to make room. This is going to be a little bit tricky because all operations can have this happen, so we almost have to register a global thread local call back so when this does happen we can capture it.|
|`spillReadTime`|DEBUG|The amount of time spent reading back in spilled data. **WE NEED TO DECIDE DO WE INCLUDE ALLOCATING THE DATA ON THE GPU OR NOT** because if we do then it could also overlap with `spillWriteTime`|
|`fetchTime`|DEBUG/MODERATE|Time spent fetching data from upstream. This could be reading data from a task that is feeding it or it could be reading data from a distributed file system/etc.|

Special care will need to be taken with `opTime` and `semaphoreWaitTime` because `gpuOpTime` is calculated from the other two, but has a different metric level from them, and the NoopMetric does not actually store any data, it just throws it all away, so we will have to come up with a good class to wrap the three metrics and do the right thing in all cases.

jlowe:
[...]

PR #2859, opened 2021-07-01
Add spillable batch caching and sort fallback to hash aggregation

jlowe:
This adds out-of-core support for hash aggregation by making all batches that are held while iterating spillable along with a sort-based fallback for groupby aggregations.

The aggregation strategy has changed as part of this PR.  Besides the sort-based fallback, the approach to aggregations before falling back is also different. The overall goal is to maximize the input size and minimize the number of aggregation calls.
- Input batches are coalesced to 2X the target size (or 2GB cudf limit, whichever is smaller) to target an approximate 4X overall non-spillable footprint (input batch + output batch). This matches the 4X target size footprint for joins.
- Instead of aggregating and immediately merging with a single held batch, aggregated batches are made spillable and collected in a queue.  Only after all input batches have been aggregated do we attempt to merge-aggregate them.
- If we are unable to merge the aggregated batches into a single batch due to the target batch limits then a sort-based fallback is used where an out-of-core sorter sorts the batches based on grouping keys.  The resulting batches are sliced at a key boundary so the batches can be independently merge-aggregated without needing to hold the entire output at once.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
@@ -16,7 +16,9 @@
 
 package com.nvidia.spark.rapids
 
-import scala.collection.mutable.ArrayBuffer
+import java.util
+
+import scala.collection.mutable
 
 import ai.rapids.cudf
 import ai.rapids.cudf.{NvtxColor, Scalar}
@@ -24,15 +26,18 @@ import com.nvidia.spark.rapids.GpuMetric._
 import com.nvidia.spark.rapids.RapidsPluginImplicits._
 
 import org.apache.spark.TaskContext
+import org.apache.spark.internal.Logging
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.{Alias, Attribute, AttributeReference, AttributeSeq, AttributeSet, Expression, If, NamedExpression}
+import org.apache.spark.sql.catalyst.expressions.{Alias, Ascending, Attribute, AttributeReference, AttributeSeq, AttributeSet, Expression, If, NamedExpression, NullsFirst}
 import org.apache.spark.sql.catalyst.expressions.aggregate._
+import org.apache.spark.sql.catalyst.expressions.codegen.LazilyGeneratedOrdering
 import org.apache.spark.sql.catalyst.plans.physical.{AllTuples, ClusteredDistribution, Distribution, HashPartitioning, Partitioning, UnspecifiedDistribution}
 import org.apache.spark.sql.catalyst.util.truncatedString
 import org.apache.spark.sql.execution.{ExplainUtils, SortExec, SparkPlan, UnaryExecNode}
 import org.apache.spark.sql.execution.aggregate.{BaseAggregateExec, HashAggregateExec, SortAggregateExec}
 import org.apache.spark.sql.rapids.{CudfAggregate, GpuAggregateExpression}
+import org.apache.spark.sql.rapids.execution.TrampolineUtil
 import org.apache.spark.sql.types.{ArrayType, LongType, MapType, StructType}
 import org.apache.spark.sql.vectorized.{ColumnarBatch, ColumnVector}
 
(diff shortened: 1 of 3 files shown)


jlowe:
Marking this as a draft while collecting some profiles and performing some scale testing.  The sort fallback is being exercised by some of the existing hash aggregation tests that use a tiny batch size to force the fallback.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -514,38 +443,26 @@ case class GpuHashAggregateExec(
...
-            cudf.ColumnVector.concatenate(
-              col1.asInstanceOf[GpuColumnVector].getBase,
-              col2.asInstanceOf[GpuColumnVector].getBase), col1.dataType())
+      val numCols = batchesToConcat.head.numCols
+      (0 until numCols).safeMap { i =>
+        val columnType = batchesToConcat.head.getBatch.column(i).dataType()
+        val columnsToConcat = batchesToConcat.map {
+          _.getBatch.column(i).asInstanceOf[GpuColumnVector].getBase
+        }
+        GpuColumnVector.from(cudf.ColumnVector.concatenate(columnsToConcat: _*), columnType)
```
revans2, line 461: I was running some tests

```
scala> spark.conf.set("spark.sql.shuffle.partitions", "1")

scala> val df = spark.range(4000000000L)
df: org.apache.spark.sql.Dataset[Long] = [id: bigint]

scala> df.registerTempTable("data")
warning: there was one deprecation warning (since 2.0.0); for details, enable `:setting -deprecation' or `:replay -deprecation'

scala> spark.time(spark.sql("SELECT COUNT(DISTINCT id) from data").show)
```

And I got a leak on this line. There were no other decs or incs to the buffer besides this. I also ran into some cases where I had to drop the batch size to be rather small < 1g to get it to not fail. For me that feels like we have a bug somewhere because the max memory usage should be <= 4x the batch size and this was more than that. I include it here because the leak might be related.

revans2 (commented):
Still looking

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -70,437 +75,361 @@ object AggregateUtils {
...
+  private[this] var outOfCoreIter: Option[GpuOutOfCoreSortIterator] = None
+
+  /** Iterator for fetching aggregated batches if a sort-based fallback has occurred */
+  private[this] var sortFallbackIter: Option[Iterator[ColumnarBatch]] = None
+
+  /** Whether a batch is pending for a reduction-only aggregation */
+  private[this] var hasReductionOnlyBatch: Boolean = isReductionOnly
+
+  override def hasNext: Boolean = {
+    // FIXME: totalTime accounting if it stays
```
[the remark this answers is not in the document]
jlowe: Yeah I'm not a fan of it either, which is why I didn't bother to plumb it through yet.  I left this as a marker to figure out what we really want to do.

@andygrove since you've been refactoring metrics lately for CBO, curious what metrics would be desired for this exec.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -70,437 +75,361 @@ object AggregateUtils {
...
-          // replacing only Final hash aggregates, so a Partial or Complete one should not replace
-          willNotWorkOnGpu("Replacing Partial or Complete hash aggregates disabled")
-        }
-      case "complete" =>
-        if (hashAggModes.contains(Partial) || hashAggModes.contains(Final)) {
-          // replacing only Complete hash aggregates, so a Partial or Final one should not replace
-          willNotWorkOnGpu("Replacing Partial or Final hash aggregates disabled")
+  override def next(): ColumnarBatch = {
+    val batch = sortFallbackIter.map(_.next()).getOrElse {
+      // aggregate and merge all pending inputs
```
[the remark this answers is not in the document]
jlowe: Would renaming `mergeAggregatedBatches` to something like `tryMergeAggregatedBatches` resolve that confusion?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
jlowe: Thanks for the example!  I never saw a failure with it, but I was able to reproduce the leak.  The issue is failing to free this when the concatenated result is later merge-aggregated.  I'll post a fix shortly.

revans2: I started to look at the memory usage of hash aggregate to see if it followed our assumptions about total memory usage.

The size of the memory that the hash table allocates is `2 * number_of_rows * size_of(thrust::pair<size_type, size_type>)` in addition to the original input table, and an output table. I am not 100% sure of the size of the output table, but we might want to ask the cudf team about it.

So in my case where I have a single long as the input with a batch size of ~2GiB and the output size is as big as the input size, then the total GPU memory size is going to approximately be. 2GiB + (2 * 2GiB/sizeof(int64_t) * sizeof(thrust::pair<size_type, size_type>)) + 2GiB => 8 GiB.  This works because the target max size is 4x the batch size.  But if I set the batch size to ~1 GiB I also get 8 GiB as the max memory size, because we are assuming that we can double the input size up to 2 GiB.  And I don't think that is true here.

We might need to rethink how we size joins as well.

revans2: For a byte column we would end up with 2 GiB input, 16 GiB of hash table and 256 bytes of output or between 18 and 36x the target batch size.
For a short column we would end up with 2 GiB input, 8 GiB of hash table and 54 KiB of output or between 5 and 10x the target batch size.
For an int column we would end up with 2 GiB input, 4 GiB of hash table and up to 2 GiB of output or between 4 and 8x the target batch size.

I think byte is the worst case we could have, even strings would not be as bad because we would still have the offsets column which would make it close to the int size in the worst case. So I think we need to do a quick look at the input schema and guess what the maximum batch size is we want to not go over the 4x target size limit. I think this is an odd enough corner case that it should not impact too many things.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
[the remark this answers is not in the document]
revans2: Sure. It really is a nit.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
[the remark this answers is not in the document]
andygrove: The `cbIter` iterator could be wrapped in a `com.nvidia.spark.rapids.CollectTimeIterator` to automatically build `collectTime` metrics on each call to `next` and `hasNext`, if we want to measure that.

For CBO specifically, I would generally want to see metrics that we can directly compare to Spark's metrics so that we can produce a cost model comparing the two. For general profiling purposes, it is good to have metrics for the main operations being performed, like aggregate/merge/concat and it looks like we are already doing this.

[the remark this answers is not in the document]
andygrove: I think https://github.com/NVIDIA/spark-rapids/issues/2527 is the best guide for which metrics to add, except instead of `opTime` we would have multiple metrics for the different operations (aggregate/merge/concat) for these more complex operators.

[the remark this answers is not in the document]
revans2: But do we need collect time or a total time for aggregations?  What metrics do we really need from aggregation?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -70,437 +75,362 @@ object AggregateUtils {
...
+      spillCallback = metrics.spillCallback)
+
+    // Finally wrap the key batching iterator with a merge aggregation on the output batches.
+    new Iterator[ColumnarBatch] {
+      override def hasNext: Boolean = keyBatchingIter.hasNext
+
+      override def next(): ColumnarBatch = {
+        // batches coming out of the sort need to be merged
+        withResource(keyBatchingIter.next()) { unmergedBatch =>
+          computeAggregate(GpuColumnVector.extractColumns(unmergedBatch), merge = true)
```
revans2, line 371: It might speed things up a lot if we can tell `computeAggregte` that the data is now sorted by the grouping keys. This is especially true for aggregations that require the data to be sorted.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -70,437 +75,362 @@ object AggregateUtils {
...
-  //       aggregation, because it detected a spill
-  //    d) sort based aggregation is then the mode forward, and not covered in this.
-  override def doExecuteColumnar(): RDD[ColumnarBatch] = {
-    val numOutputRows = gpuLongMetric(NUM_OUTPUT_ROWS)
-    val numOutputBatches = gpuLongMetric(NUM_OUTPUT_BATCHES)
-    val totalTime = gpuLongMetric(TOTAL_TIME)
-    val computeAggTime = gpuLongMetric(AGG_TIME)
-    val concatTime = gpuLongMetric(CONCAT_TIME)
-    val rdd = child.executeColumnar()
+      aggregatedBatches.addLast(mergedBatch)
```
revans2, line 277: nit: can we add a comment here about how we are doing a single merge pass through the queue and even though we appending the results to the end of the queue, we prevent them from taking part in the merge step again.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -70,437 +75,362 @@ object AggregateUtils {
...
+  private def mergePass(): Boolean = {
+    val batchesToConcat: mutable.ArrayBuffer[LazySpillableColumnarBatch] = mutable.ArrayBuffer.empty
+    var wasBatchMerged = false
+    var concatSize: Long = 0L
+    var batchesLeftInPass = aggregatedBatches.size()
+
+    while (batchesLeftInPass > 0) {
+      closeOnExcept(batchesToConcat) { _ =>
+        var isConcatSearchFinished = false
+        // Although tempting to allow the pass to "wrap around" and pick up batches previously
```
revans2, line 252: This is great.  Might be good to mention that this is because of first/last.  It would also be great if we could have a follow on issue for stable sort.  This is another place where that would be helpful and when I looked I didn't see an issue for it. I guess I forgot to file one.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -70,437 +75,362 @@ object AggregateUtils {
...
+    spillCallback: RapidsBuffer.SpillCallback)
+
+/** Utility class to convey information on the aggregation modes being used */
+case class AggregateModeInfo(
+    uniqueModes: Seq[AggregateMode],
+    hasPartialMode: Boolean,
+    hasFinalMode: Boolean,
+    hasCompleteMode: Boolean)
+
+class GpuHashAggregateIterator(
```
revans2, line 95: Could we get a high level comment on this class, or somewhere in this file that documents the data flow? I understand it now, but it took me two hours of reading the code and jumping back and forth to wrap my head around it. And that is even after we had a call last week and came up with a design for this together.  Just a quick note like. 

The data goes through a few stages.  The first stage does an initial aggregation pass across all of the input batches `aggregateInputBatches`. At that point it needs to either combine all of the batches together into a single output batch `mergeAggregatedBatches` or fall back to sorting this intermediate data so the results can be combined and output as multiple batches `buildSortFallbackIterator`.

revans2 (commented):
Just a few nits, besides the issues I was seeing with actual memory usage in some test cases I ran.  Hopefully those are outliers. I really would like to see some performance testing results especially for NDS runs.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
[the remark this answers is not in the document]
andygrove: For CBO, I suppose we really just want the cost of this operator, exclusive of collectTime (unless collectTime is insignificant in which case we can just include it), so that we can estimate the cost of this operator for a given input row size, ignoring other factors that would affect execution time for now. If the individual metrics (aggregate, merge, etc) can be summed to give the total opTime, I think that would work.

For CPU we can use the optimizer operator isolation hacks to try and measure a total time for this operator (exclusive of child totalTime) and use that for the CPU estimate.

abellina (commented):
Just a first pass, overall I don't see anything major. I'll take another crack at it soon.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -70,437 +75,362 @@ object AggregateUtils {
...
+    while (batchesLeftInPass > 0) {
+      closeOnExcept(batchesToConcat) { _ =>
+        var isConcatSearchFinished = false
+        // Although tempting to allow the pass to "wrap around" and pick up batches previously
+        // merged in this pass, it's avoided to prevent changing the order of aggregated batches.
+        while (batchesLeftInPass > 0 && !isConcatSearchFinished) {
+          val candidate = aggregatedBatches.getFirst
+          val potentialSize = concatSize + candidate.deviceMemorySize
+          isConcatSearchFinished = concatSize > 0 && potentialSize > targetBatchSize
+          if (!isConcatSearchFinished) {
```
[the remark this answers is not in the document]
jlowe: Yes, this would occur when there's exactly one batch after the concat search is finished.  For example, we pick up some large first batch unconditionally, because concatSize == 0.  Then on the next batch we find that adding it would exceed the target batch size, so we mark the search finished and exit the loop without adding it to the concat list.  That's when we hit the else clause, because batchesToConcat.length <= 1 and there's no point in trying to concatenate a single batch.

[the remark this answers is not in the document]
abellina: @jlowe clarified this further as I still was confused (my bad). I had missed that the condition in line 266 was `batchesToConcat.length > 1`  not `batchesToConcat.length > 0`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -662,15 +577,15 @@ case class GpuHashAggregateExec(
     // - Partial, PartialMerge mode: we use the inputProjections or distinct update expressions
     //   for Partial and non distinct merge expressions for PartialMerge.
     // - Final mode: we pick the columns in the order as handed to us.
-    val boundInputReferences = if (uniqueModes.contains(PartialMerge)) {
+    val boundInputReferences = if (modeInfo.uniqueModes.contains(PartialMerge)) {
```
abellina, line 580: minor nit, we have  `hasFinalMode` and `hasCompleteMode` and others, can we add a `hasPartialMerge`?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
jlowe: Filed #2884 to track this.

jlowe (commented):
I still need to test on Databricks and run scale tests, but I believe I have addressed the review comments.

jlowe:
build

jlowe:
CI failed with a CPU/GPU mis-match in the test `test_hash_multiple_mode_query[{'spark.rapids.sql.variableFloatAgg.enabled': 'true', 'spark.rapids.sql.hasNans': 'false', 'spark.rapids.sql.castStringToFloat.enabled': 'true', 'spark.rapids.sql.batchSizeBytes': '1000'}-[('a', RepeatSeq(String)), ('b', Integer), ('c', Null)]]` but I cannot reproduce this locally.  Kicking the CI build again to see if this is consistent.

jlowe:
build

jlowe:
So good news is that the CI failure is repeatable, and I was able to reproduce it locally by setting `PYSP_TEST_spark_master="local[8,1]"`.  The problem goes away if I change the code to avoid hinting to cudf that the data is sorted.  I'll dig more into the failure, see if I can get a repro case with just libcudf.

jlowe:
I was able to localize this to a problem in libcudf with sort-hinted aggregations.  I filed https://github.com/rapidsai/cudf/issues/8717

jlowe:
build

abellina (approved):

jlowe:
I think this is ready to be merged.  CI tests are passing, I manually tested this on Databricks, and this was benchmarked on NDS at 3TB scale with no significant performance regression.

@revans2 it would be great if you could give this another look.

revans2 (approved):

PR #2866, opened 2021-07-02
WIP: Leverage SPARK-35881 to remove use of reflection when avoiding redundant transitions in AQE

andygrove:
This PR demonstrates how we can simplify handling of transitions around the final query stage of adaptive queries once [SPARK-35881](https://github.com/apache/spark/pull/33140) is merged and remove the need to use reflection to access private Spark methods. 

This PR currently only implements the improvement with the fix backported to Spark 3.1.3-SNAPSHOT and we would need to make similar changes to other shims.

## Problem

Prior to this PR, the final query stage of an adaptive query is always row-based. We insert a transition inside the query stage to ensure this. If the adaptive plan is nested in a columnar write then we insert another transition to convert back to columnar, so we now have:

```
InsertIntoHadoopFsRelationExec
  GpuRowToColumnarExec
    AdaptiveSparkPlanExec
      GpuColumnarToRowExec    
```

These transitions are redundant so we insert another operator to avoid the transition at runtime.

```
InsertIntoHadoopFsRelationExec
  AvoidAdaptiveTransitionToRowExec
    GpuRowToColumnarExec
      AdaptiveSparkPlanExec
        GpuColumnarToRowExec  
```

The `AvoidAdaptiveTransitionToRowExec` operator uses reflection to access private methods within `AdaptiveSparkPlanExec`. This is necessary because adaptive plans do not support columnar execution prior to SPARK-35881.

## New Approach

With SPARK-35881 supporting columnar execution of adaptive plans, we can now put the transition outside of the adaptive plan:

```
GpuColumnarToRowExec 
  AdaptiveSparkPlanExec
```

When nested in a columnar write, we now see this interim plan:

```
InsertIntoHadoopFsRelationExec
  GpuRowToColumnarExec
    GpuColumnarToRowExec 
      AdaptiveSparkPlanExec
```

The redundant transitions are now easily optimized out in `GpuTransitionOverrides`, leaving us with:

```
InsertIntoHadoopFsRelationExec
  GpuRowToColumnarExec
    AdaptiveSparkPlanExec
```

We need to still have the `GpuRowToColumnarExec` in the plan because `AdaptiveSparkPlanExec.supportsColumnar` is always false (see SPARK-35881 for more details) but this operator is essential a no-op if it is wrapping an adaptive plan where the final stage is columnar.

### DIFF
--- a/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/AvoidAdaptiveTransitionToRow.scala
+++ b/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/AvoidAdaptiveTransitionToRow.scala
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.nvidia.spark.rapids.shims.spark301
+
+import java.lang.reflect.Method
+
+import com.nvidia.spark.rapids.{GpuColumnarToRowExec, GpuExec, GpuRowToColumnarExec}
+
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.catalyst.InternalRow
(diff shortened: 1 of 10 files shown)


shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/AvoidAdaptiveTransitionToRow.scala
```diff
@@ -0,0 +1,84 @@
...
+ * that wants rows (such as CollectTailExec) or from an operation that wants columns (such as
+ * GpuDataWritingCommandExec).
+ *
+ * Spark does not provide a mechanism for executing an adaptive plan and retrieving columnar
+ * results and the internal methods that we need to call are private, so we use reflection to
+ * call them.
+ *
+ * @param child The plan to execute
+ */
+case class AvoidAdaptiveTransitionToRow(child: SparkPlan) extends UnaryExecNode with GpuExec {
```
andygrove, line 51: This is an existing class that is now moved to the shim layer

PR #2873, opened 2021-07-06
Fix infinite loop in MultiFileCloudPartitionReaderBase

jlowe:
While refactoring some code, I stumbled across an infinite loop in the `next()` method for `MultiFileCloudPartitionReaderBase`.  It recurses to itself without changing any of the state that led to the recursion point.


jlowe:
build

wbo4958:
Thx Jason, Great catch. Could we have a unit test for this? I believe we don't have such unit test for this.

tgravescs (approved):

jlowe:
Added a unit test that fails with a StackOverflowError without the fix and passes with it.

jlowe:
build

jlowe:
build

wbo4958 (approved):
Thx Jason

gerashegalov (approved):
LGTM

PR #2875, opened 2021-07-07
Add support for timeParserPolicy=LEGACY

### LINKED ISSUE #2860: [FEA] Minimal support for LEGACY timeParserPolicy
With `spark.sql.legacy.timeParserPolicy` set to `LEGACY` we currently fall back to CPU. This issue is for the initial work of supporting the legacy policy for format strings that are already supported in `CORRECTED` mode and which would work identically in `LEGACY` mode.

We will not support the hybrid Julian+Gregorian calendar that Spark supports in legacy mode so we need to document that and perhaps have a feature flag for enabling legacy support, since we are not fully compatible.

- With `LEGACY` mode enabled, and with supported formats, we should run on the GPU.
- With `LEGACY` mode enabled, and with unsupported formats, we should fall back to CPU.

N/A

**Additional context**
N/A

andygrove:
Closes #2860

This PR adds support for parsing strings to date/timestamp when `spark.sql.legacy.timeParserPolicy=LEGACY` for the following formats:

- `dd-MM-yyyy`
- `dd/MM/yyyy`
- `yyyy/MM/dd`
- `yyyy-MM-dd`
- `yyyy/MM/dd HH:mm:ss`
- `yyyy-MM-dd HH:mm:ss`

We are not 100% compatible with Spark on CPU in all cases so this support is only enabled when `spark.rapids.sql.incompatibleDateFormats.enabled` is also set to true. We have the following limitations when running on the GPU:

- Only 4-digit years are supported
- The proleptic Gregorian calendar is used instead of the hybrid Julian+Gregorian calendar that Spark uses in legacy mode, so we produce different results for dates prior to the Gregorian calendar starting.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -353,7 +353,12 @@ the specified format string will fall into one of three categories:
 - Supported on GPU but may produce different results to Spark
 - Unsupported on GPU
 
-The formats which are supported on GPU and 100% compatible with Spark are :
+The formats which are supported on GPU vary depending on the setting for `timeParserPolicy`.
+
+### CORRECTED and EXCEPTION timeParserPolicy
+
+With timeParserPolicy set to `CORRECTED` or `EXCEPTION` (the default), the following formats are supported
+on the GPU without requiring any additional settings.
 
 - `dd/MM/yyyy`
 - `yyyy/MM`
@@ -366,10 +371,11 @@ The formats which are supported on GPU and 100% compatible with Spark are :
 - `dd-MM`
 - `dd/MM`
 
-Examples of supported formats that may produce different results are:
+Valid Spark date/time formats that do not appear in the list above may also be supported but have not been 
+extensively tested and may produce different results compared to the CPU. Known issues include:
 
-- Trailing characters (including whitespace) may return a non-null value on GPU and Spark will 
-  return null 
+- Valid dates and timestamps followed by trailing characters (including whitespace) may be parsed to non-null 
+  values on GPU where Spark would treat the data as invalid and return null
 
 To attempt to use other formats on the GPU, set
 [`spark.rapids.sql.incompatibleDateFormats.enabled`](configs.md#sql.incompatibleDateFormats.enabled)
@@ -390,6 +396,26 @@ Formats that contain any of the following words are unsupported and will fall ba
 "D", "DD", "DDD", "s", "m", "H", "h", "M", "MMM", "MMMM", "MMMMM", "L", "LLL", "LLLL", "LLLLL",
 "d", "S", "SS", "SSS", "SSSS", "SSSSS", "SSSSSSSSS", "SSSSSSS", "SSSSSSSS"
 ```
+
+### LEGACY timeParserPolicy
+
+With timeParserPolicy set to `LEGACY` and
+[`spark.rapids.sql.incompatibleDateFormats.enabled`](configs.md#sql.incompatibleDateFormats.enabled)
+set to `true`, and `spark.sql.ansi.enabled` set to `false`, the following formats are supported but not 
+guaranteed to produce the same results as the CPU:
+
+- `dd-MM-yyyy`
+- `dd/MM/yyyy`
+- `yyyy/MM/dd`
+- `yyyy-MM-dd`
+- `yyyy/MM/dd HH:mm:ss`
+- `yyyy-MM-dd HH:mm:ss`
+
+LEGACY timeParserPolicy support has the following limitations when running on the GPU:
+
+- Only 4 digit years are supported
+- The proleptic Gregorian calendar is used instead of the hybrid Julian+Gregorian calender 
+  that Spark uses in legacy mode
 
 ## Formatting dates and timestamps as strings
 
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
@@ -645,12 +645,16 @@ object GpuToTimestamp extends Arm {
     val format = LEGACY_COMPATIBLE_FORMATS.find(_.format == sparkFormat)
       .getOrElse(throw new IllegalStateException(s"Unsupported format $sparkFormat"))
 
+    // optimization to apply only the necessary rules depending on whether we are
+    // parsing to a date or timestamp
     val regexReplaceRules = if (format.isTimestamp) {
       FIX_DATES ++ FIX_TIMESTAMPS
     } else {
       FIX_DATES
     }
 
+    // we support date formats using either `-` or `/` to separate year, month, and day and the
+    // regex rules are written with '-' so we need to replace '-' with '/' here as necessary
     val rulesWithSeparator = format.separator match {
       case '/' =>
         regexReplaceRules.map {
@@ -661,6 +665,7 @@ object GpuToTimestamp extends Arm {
         regexReplaceRules
     }
 
+    // apply each rule in turn to the data
     val fixedUp = rulesWithSeparator
       .foldLeft(rejectLeadingNewlineThenStrip(lhs))((cv, regexRule) => {
         withResource(cv) {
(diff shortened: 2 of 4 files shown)


docs/compatibility.md
```diff
@@ -353,7 +353,12 @@ the specified format string will fall into one of three categories:
 - Supported on GPU but may produce different results to Spark
 - Unsupported on GPU
 
-The formats which are supported on GPU and 100% compatible with Spark are :
+The formats which are supported on GPU and 100% compatible with Spark vary depending on the setting
```
revans2, line 356: But from reading below none of the formats are 100% compatible with Spark because even for `CORRECTED` and `EXCEPTION` we do not detect trailing characters.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -398,17 +398,40 @@ abstract class UnixTimeExprMeta[A <: BinaryExpression with TimeZoneAwareExpressi
...
+              if (GpuToTimestamp.LEGACY_COMPATIBLE_FORMATS.exists(_.format == sparkFormat)) {
+                // LEGACY support has a number of issues that mean we cannot guarantee
+                // compatibility with CPU
+                // - we can only support 4 digit years but Spark supports a wider range
+                // - we use a proleptic Gregorian calender but Spark uses a hybrid Julian+Gregorian
+                //   calender in LEGACY mode
+                if (!conf.incompatDateFormats) {
+                  willNotWorkOnGpu(s"LEGACY format '$sparkFormat' on the GPU is not guaranteed " +
+                    s"to produce the same results as Spark on CPU. Set " +
+                    s"spark.rapids.sql.incompatibleDateFormats.enabled=true to force onto GPU.")
```
revans2, line 416: nit. Instead of hard coding the config here can we use the KEY form the RapidsConf just to avoid any possible misspellings?

andygrove:
During manual performance testing, I ran into some behavior that I don't understand yet, so I am changing this to a draft / WIP for now.

andygrove:
> During manual performance testing, I ran into some behavior that I don't understand yet, so I am changing this to a draft / WIP for now.

The issue was that I had ANSI mode enabled when manually testing, and hadn't implemented ANSI support as part of this PR. It is now updated to fall back to CPU if LEGACY + ANSI are both enabled. If we do want to support LEGACY + ANSI together then we can do that as a follow-on issue.

andygrove:
build

docs/compatibility.md
andygrove: This was badly worded. We do claim that some of the formats are 100% compatible. I have made some changes to try and make this clearer.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -444,6 +472,76 @@ object GpuToTimestamp extends Arm {
...
+  val FIX_SINGLE_DIGIT_MONTH: RegexReplace =
+    RegexReplace("(\\A\\d+)-(\\d{1}-)", "\\1-0\\2")
+
+  /** Regex rule to replace "yyyy-mm-d" with "yyyy-mm-dd" */
+  val FIX_SINGLE_DIGIT_DAY_1: RegexReplace =
+    RegexReplace("(\\A\\d+-\\d{2})-(\\d{1}[\\D\\s]+)", "\\1-0\\2")
+
+  /** Regex rule to replace "yyyy-mm-d" with "yyyy-mm-dd" */
+  val FIX_SINGLE_DIGIT_DAY_2: RegexReplace =
+    RegexReplace("(\\A\\d+-\\d{2})-(\\d{1})\\Z", "\\1-0\\2")
```
jlowe, line 507: Similar to the suggestion at https://github.com/NVIDIA/spark-rapids/pull/2890#discussion_r666504699, I think these could be combined, something like:
```suggestion
    RegexReplace(raw"(\A\d+-\d{2})-(\d{1})([\D\s]+|\Z)", raw"\1-0\2\3")
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -444,6 +472,76 @@ object GpuToTimestamp extends Arm {
...
+
+  /** Regex rule to replace "yyyy-mm-dd h-" with "yyyy-mm-dd hh-" */
```
jlowe, line 509: In the comments here and below there are many places where the comment is using `-` when it should be using `:` since that's what the regexp does.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -444,6 +472,76 @@ object GpuToTimestamp extends Arm {
...
+  val FIX_SINGLE_DIGIT_MINUTE: RegexReplace =
+    RegexReplace("(\\A\\d+-\\d{2}-\\d{2}[ T]\\d{2}):(\\d{1}:)", "\\1:0\\2")
+
+  /** Regex rule to replace "yyyy-mm-dd[ T]hh-mm-s" with "yyyy-mm-dd[ T]hh-mm-ss" */
+  val FIX_SINGLE_DIGIT_SECOND_1: RegexReplace =
+    RegexReplace("(\\A\\d+-\\d{2}-\\d{2}[ T]\\d{2}:\\d{2}):(\\d{1}\\D+)", "\\1:0\\2")
+
+  /** Regex rule to replace "yyyy-mm-dd[ T]hh-mm-s" with "yyyy-mm-dd[ T]hh-mm-ss" */
+  val FIX_SINGLE_DIGIT_SECOND_2: RegexReplace =
+    RegexReplace("(\\A\\d+-\\d{2}-\\d{2}[ T]\\d{2}:\\d{2}):(\\d{1})\\Z", "\\1:0\\2")
```
jlowe, line 527: I think these could be combined, something similar to:
```suggestion
    RegexReplace("(\\A\\d+-\\d{2}-\\d{2}[ T]\\d{2}:\\d{2}):(\\d{1}\\D+)", "\\1:0\\2")

  /** Regex rule to replace "yyyy-mm-dd[ T]hh-mm-s" with "yyyy-mm-dd[ T]hh-mm-ss" */
  val FIX_SINGLE_DIGIT_SECOND_2: RegexReplace =
    RegexReplace(raw"(\A\d+-\d{2}-\d{2}[ T]\d{2}:\d{2}):(\d{1})(\D+|\Z)", raw"\1:0\2\3")
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
andygrove: I had to tweak the syntax slightly for cuDF but I was able to combine all three sets of duplicated rules :tada: .. thanks for the pointers

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -444,6 +472,61 @@ object GpuToTimestamp extends Arm {
     "dd/MM"
   )
 
+  // We are compatible with Spark for these formats when the timeParserPolicy is LEGACY. It
+  // is possible that other formats may be supported but these are the only ones that we have
+  // tests for.
+  val LEGACY_COMPATIBLE_FORMATS = Seq(
```
gerashegalov, line 478: since we are doing lookups in this `Seq` can we make it a `Set[LegacyParseFormat]` or even better a `Map[String, LegacyParseFormat]`

andygrove: Thanks. Fixed.

jlowe (approved):

jlowe:
build

gerashegalov (approved):
:rocket:

Issue #1590, opened 2021-01-26
Rapids Shuffle - UcpListener


abellina:
Use `UcpListener` to connect to peers instead of our own connection. This will enable us to exchange the socket address to communicate with UCX. UCX keeps the connection alive, which allows it detect peer disconnection, which should help with reliability of the rapids shuffle.

This also removes TCP code that was added to exchange UCX's WorkerAddress, so it should end up cleaning code required for this exchange.

abellina:
@petro-rudenko has a PR for this already, we need to follow up with removal of management port.

Issue #737, opened 2020-09-11
[FEA] handle peer timeouts in shuffle


abellina:
When UCX starts supporting peer error handling, this task is to pipe that in and cause a `RapidsShuffleFetchFailedException` or a `RapidsShuffleTimeoutException`.

This will help us report to the user, and the `MapOutputTracker` more accurate information at least for debugging/logs. Current logic in Spark indicates that a fetch failure from a peer will cause all output for such peer to be invalidated, but that certainly is not the case with the external shuffle service, but the error handling for non-ESS cases may change in the future, so we'll be prepared if we are implementing `FetchFailedException` as specified.

sameerz:
This depends on UCX 1.10.

PR #2886, opened 2021-07-07
Use UCP Listener for UCX connections and enable peer error handling

abellina:
Closes https://github.com/NVIDIA/spark-rapids/issues/1590
Closes https://github.com/NVIDIA/spark-rapids/issues/2275
Closes https://github.com/NVIDIA/spark-rapids/issues/737

Created this as a draft because I may add more unit tests. But I could use some feedback on the patch. (I can't think of ways of adding tests without refactoring `UCX.scala`, so I am going to hold off).

This switches to use `UCPListener` as our incoming connection port, and enables peer error handling by default.

With this change, if a peer terminates the following is expected (@petro-rudenko please confirm):

1. UCX will invoke the `onError` callback for each of the pending live requests.
2. After the pending requests have been errored, UCX will call an extra error handler for the failed endpoint.

Because of how UCX works right now, a failing endpoint may or may not be of interest. We make an attempt to not initiate endpoints if we detect that we already had an endpoint created due to a peer initiating against us, but this is a race condition and in this PR we simply make one of those two endpoints win and be the default, while the other sticks around.

In the future: https://github.com/openucx/ucx/pull/6859, we'll be able to better detect this (earlier) and reject certain requests.

@jlowe @petro-rudenko


shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -425,36 +300,57 @@ class UCX(transport: UCXShuffleTransport, executor: BlockManagerId, rapidsConf:
...
+   * the response will never arrive, so this cleans up the response handler.
+   *
+   * @param am active message (id, header) that for the response handler to remove
+   */
+  def unregisterResponseHandler(am: UCXActiveMessage): Unit = {
+    amRegistrations.computeIfPresent(am.activeMessageId,
+      (_, msg) => {
+        msg match {
+          case reg: ResponseActiveMessageRegistration =>
+            reg.removeResponseActiveMessageHandler(am.header)
```
jlowe, line 345: Do we need to do anything special if this was the last handler removed?

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -879,33 +673,363 @@ class UCX(transport: UCXShuffleTransport, executor: BlockManagerId, rapidsConf:
...
+          transport.getDirectByteBuffer(size.toInt)
+        }
+      })
+
+      sendControlRequest(ep, responseAm)
+    }
+
+    // called from the progress thread
+    private def handleConnectedPeerFromRequest(executorId: Long, newEp: UcpEndpoint,
+        peerRkeys: Seq[ByteBuffer]): Unit = {
```
jlowe, line 822: peerRkeys is not used.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCXConnection.scala
```diff
@@ -424,113 +426,57 @@ object UCXConnection extends Logging {
...
-    val numRkeys = readBytesFromStream(false, is, 4).getInt()
-
-    val rkeys = new ArrayBuffer[ByteBuffer](numRkeys)
-    (0 until numRkeys).foreach { _ =>
-      val size = readBytesFromStream(false, is, 4).getInt()
-      rkeys.append(readBytesFromStream(true, is, size))
+  def unpackHandshake(buff: ByteBuffer): (Long, Seq[ByteBuffer]) = {
+    val remoteExecutorId = buff.getLong
+    val numRkeys = buff.getInt
+    val rkeys = (0 until numRkeys).map { i =>
```
jlowe, line 442:
```suggestion
    val rkeys = (0 until numRkeys).map { _ =>
```

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleClientSuite.scala
```diff
@@ -703,4 +706,38 @@ class RapidsShuffleClientSuite extends RapidsShuffleTestHelper {
...
+
+  test("on endpoint failure the iterator is notified if it is registered") {
+    when(mockTransaction.getStatus).thenReturn(TransactionStatus.Success)
+    val metaResp = ShuffleMetadata.buildMetaResponse(Seq.empty)
+    when(mockTransaction.releaseMessage()).thenReturn(
+      new RefCountedDirectByteBuffer(metaResp))
+
+    client.registerPeerErrorListener(mockHandler)
+    client.doFetch(Seq(ShuffleBlockBatchId(1,2,3,4)), mockHandler)
+    client.close()
```
jlowe, line 718: Nit: Can use `withResource` here

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleClient.scala
```diff
@@ -101,7 +101,11 @@ class RapidsShuffleClient(
...
     catalog: ShuffleReceivedBufferCatalog = GpuShuffleEnv.getReceivedCatalog)
-      extends Logging with Arm {
+      extends Logging with Arm with AutoCloseable {
+
+  // these are handlers that are interested (live spark tasks) in peer failure handling
+  private val liveHandlers =
+    ConcurrentHashMap.newKeySet[RapidsShuffleFetchHandler]()
 
   object ShuffleClientOps {
     /**
```
jlowe, line 111: resp parameter in scaladoc just below does not match method signature.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
abellina: ugh, good eye. This needs to be fixed.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -85,7 +84,7 @@ class UCX(transport: UCXShuffleTransport, executor: BlockManagerId, rapidsConf:
 
   logInfo(s"UCX context created")
 
-  def getExecutorId: Int = executor.executorId.toInt
+  def getExecutorId: Long = executor.executorId.toLong
```
gerashegalov, line 87: since we are changing this line, this is a pretty inefficient transformation to be a def https://github.com/scala/scala/blob/v2.12.6/src/library/scala/collection/immutable/StringLike.scala#L306

Can we make it a val?

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCXConnection.scala
```diff
@@ -424,113 +426,57 @@ object UCXConnection extends Logging {
...
+      val rkeySize = buff.getInt
+      val rkeySlice = buff.slice()
+      rkeySlice.limit(rkeySize)
+      val rkey = ByteBuffer
+        .allocateDirect(rkeySize)
```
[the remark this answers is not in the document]
abellina: good call, done.

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleClientSuite.scala
```diff
@@ -150,7 +152,8 @@ class RapidsShuffleClientSuite extends RapidsShuffleTestHelper {
       // the transport will receive no pending requests (for buffers) for queuing
       verify(mockTransport, times(0)).queuePending(any())
 
-      assert(response.isClosed)
+      // this is not called since a message implies success
+      verify(mockTransaction, times(0)).releaseMessage()
 
       newMocks()
```
jlowe, line 158:
```suggestion
```

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
jlowe: Yeah, as long as the map can't grow unbounded with empty handler stubs then it should be fine as-is.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleTransport.scala
```diff
@@ -453,12 +459,16 @@ class DirectByteBufferPool(bufferSize: Long) extends Logging {
       new RefCountedDirectByteBuffer(ByteBuffer.allocateDirect(bufferSize.toInt), Option(this))
     } else {
       buff.clear()
+      // reset endianness to BIG_ENDIAN, as it could have changed depending on the consumer
+      // flatbuffers
+      buff.order(ByteOrder.BIG_ENDIAN)
```
gerashegalov, line 464: Can we add a comment why we need BIG_ENDIAN to begin with? Looks like it's the only place referencing BIG_ENDIAN byte order

abellina: This should handle it @gerashegalov (https://github.com/NVIDIA/spark-rapids/pull/2886/commits/2efa3461313d53b343ff4dcc4f9c2d96407891b9)

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleClientSuite.scala
jlowe: Ah, nevermind, this is one of the cases where it's needed.  However a cleaner way to do this is to either make this a parameterized test or factor out the common code and make multiple tests for the multiple cases.  Bonus points: when the test fails for a particular setting, the test failure lets you know which case it was.  As-is, it will just say the test failed and it's left as an exercise to debug which one.  Not mustfix for this PR.

jlowe (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleTransport.scala
abellina: Added this: https://github.com/NVIDIA/spark-rapids/issues/2897 @gerashegalov to potentially just use LITTLE_ENDIAN consistently.

abellina:
I'll make the `newMocks` suggestion, seems simple enough.

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleClientSuite.scala
abellina: done.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCXConnection.scala
```diff
@@ -424,113 +426,53 @@ object UCXConnection extends Logging {
...
-
-    // pack the worker address
-    hsBuff.putInt(workerAddress.capacity)
-    hsBuff.put(workerAddress)
-
-    // send the executor id
-    hsBuff.putInt(localExecutorId)
-
+  def packHandshake(localExecutorId: Long, rkeys: Seq[ByteBuffer]): ByteBuffer = {
+    val size = 8 + 4 + (4 * rkeys.size) + rkeys.map(_.capacity).sum
```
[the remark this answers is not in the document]
abellina: I did this.

abellina:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleServer.scala
```diff
@@ -343,52 +343,59 @@ class RapidsShuffleServer(transport: RapidsShuffleTransport,
...
+                  }
+                case _ =>
+                  // errored or cancelled
+                  logError(s"Error while sending buffers $bufferTx.")
                   bssExec.synchronized {
                     bufferSendState.close()
                     bssExec.notifyAll()
                   }
                 }
-              }
```
jlowe, line 391: Nit: Looks like the wrong closing brace was removed here, as the brace above is mis-indented.

gerashegalov (approved):
LGTM

jlowe (approved):

jlowe:
build

Issue #2879, opened 2021-07-07
[FEA] ORC reader supports reading Struct columns.


firestarman:
cudf has added support for array and struct type in ORC reader by the PR https://github.com/rapidsai/cudf/pull/8599 , so we can update the plugin side accordingly.

sameerz:
We need to make sure when we do nested pruning we either make sure that works when reading structs in ORC or fall back to the CPU.  See issue #463 and issue #576

firestarman:
Supporting struct is in progress.

tgravescs:
reopening because this was pulled back out with: https://github.com/NVIDIA/spark-rapids/pull/3079

tgravescs:
also note test failures https://github.com/NVIDIA/spark-rapids/issues/3059 and predicate pushdown https://github.com/NVIDIA/spark-rapids/issues/1481

firestarman:
This depends on rapidsai/cudf#8910

Issue #1481, opened 2021-01-08
[FEA] ORC Predicate pushdown for Nested fields


nartal1:
Apache Spark added support for ORC predicate pushdown for Nested fields in https://issues.apache.org/jira/browse/SPARK-25557.  We should look if it's worth to update the code in OrcFilters.scala esp method `isSearchableType`.

firestarman:
Could anyone share how to test this feature ?
According to the [comment ](https://github.com/NVIDIA/spark-rapids/issues/576#issuecomment-879877596), we probably only need to add tests for this.

tgravescs:
I suggest looking at the tests in the Spark issue.

firestarman:
> I suggest looking at the tests in the Spark issue.

Thanks, and yes, I did something similar to the Spark tests.

PR #2887, opened 2021-07-08
ORC reader supports struct

firestarman:
Let ORC reader support struct, since cudf has supported them by the PR https://github.com/rapidsai/cudf/pull/8599 .
Also add tests for pruning nested fields.

closes https://github.com/NVIDIA/spark-rapids/issues/2879
closes https://github.com/NVIDIA/spark-rapids/issues/1481


firestarman:
Filed the issue https://github.com/rapidsai/cudf/issues/8704 to track the test failures.

firestarman:
Updated the tests to avoid generating null struct rows, to unblock this PR.

firestarman:
build

integration_tests/src/main/python/orc_test.py
```diff
@@ -45,9 +45,30 @@ def test_basic_read(std_input_path, name, read_func, v1_enabled_list, orc_impl,
             read_func(std_input_path + '/' + name),
             conf=all_confs)
 
-orc_gens_list = [[byte_gen, short_gen, int_gen, long_gen, float_gen, double_gen,
+orc_basic_gens = [byte_gen, short_gen, int_gen, long_gen, float_gen, double_gen,
     string_gen, boolean_gen, DateGen(start=date(1590, 1, 1)),
-    TimestampGen(start=datetime(1590, 1, 1, tzinfo=timezone.utc))],
+    TimestampGen(start=datetime(1590, 1, 1, tzinfo=timezone.utc))]
+
+# Set to true after the issue https://github.com/rapidsai/cudf/issues/8704 is fixed.
```
jlowe, line 52: We can't check this in as-is -- it can silently corrupt data on an ORC read.  Either we need to make the struct support qualified by a separate incompatible config that defaults to off or wait until it is fixed.

jlowe (changes requested):
In general this looks good, but we can't default to turning ORC struct support on if we know it cannot load null struct rows properly.

integration_tests/src/main/python/orc_test.py
firestarman: The issue rapidsai/cudf#8704 is marked as  " v21.08 Release", so  I plan to wait for the fix.

firestarman:
As suggested by @sameerz , we plan to make two separate PRs for list and struct respectively.
https://github.com/NVIDIA/spark-rapids/pull/2946 is for list, and update this one for struct only.

firestarman:
Here is the fix https://github.com/rapidsai/cudf/pull/8819 for the issue rapidsai/cudf#8704

firestarman:
Found a bug when reading data from orc with different column order in read schema, and @wbo4958 is working on it.

firestarman:
build

firestarman:
build

firestarman:
build

firestarman:
build

integration_tests/run_pyspark_from_build.sh
```diff
@@ -102,7 +102,7 @@ else
     RUN_TESTS_COMMAND=("$SCRIPTPATH"/runtests.py
       --rootdir
       "$LOCAL_ROOTDIR"
-      "$LOCAL_ROOTDIR"/src/main/python)
+      "$LOCAL_ROOTDIR"/src/main/python/orc_test.py)
```
jlowe, line 105: Why was this change made?

firestarman: Nice catch and updated.
Uploaded this by mistake.

firestarman:
build

jlowe (approved):

PR #2888, opened 2021-07-08
Type check with the information from RapidsMeta

sperlingxx:

Currently, during type tagging, we get the type information directly from what we have wrapped in the `RapidsMeta`.  It works well for most of expressions and plans.  But for some special instances, such as `TypedImperativeAggregate`,  we need to do type conversion based on the raw type. Taking `TypedImperativeAggregate` as an example, the GPU implementation introduces a different kind of internal aggregation buffer, whose data type is not same as the CPU counterpart.

Therefore,  in this PR, we modified the RapidsMetas and type checks, to do type check with the type information from the corresponding methods of RapidsMeta rather than raw data types of wrapped instances.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -214,14 +214,18 @@ final class TypeSig private(
...
-   * @param expr the expression to check against.
-   * @param name the name of the expression (typically a parameter name)
+   *
+   * @param meta     the meta that gets marked for support or not.
+   * @param exprMeta the meta of expression to check against.
+   * @param name     the name of the expression (typically a parameter name)
    */
-  def tagExprParam(meta: RapidsMeta[_, _, _], expr: Expression, name: String): Unit = {
+  def tagExprParam(meta: RapidsMeta[_, _, _],
+      exprMeta: BaseExprMeta[_],
```
revans2, line 223: nit: Why split this up onto multiple lines? it is not any longer than it was before?

revans2 (commented):
Generally I think this is okay. My main problem is that it is not obvious to the end user that we changed the type.  We have already been bitten by this with decimal and with timestamps where we do not explain to the user that decimal is disabled or that timestamps are disabled because the time zone is not UTC.  I would like to avoid the same problem and ideally tag the types that we changed with a reason why we changed them.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -846,18 +859,14 @@ object CreateNamedStructCheck extends ExprChecks {
...
+        case Seq(nameMeta, valueMeta) =>
+          nameSig.tagExprParam(meta, nameMeta, "name")
+          nameSig.tagExprParam(meta, valueMeta, "value")
       }
-      if (!resultSig.isSupportedByPlugin(origExpr.dataType, meta.conf.decimalTypeEnabled)) {
-        meta.willNotWorkOnGpu(s"unsupported data type in output: ${origExpr.dataType}")
+      val dt = exprMeta.dataType.get
+      if (!resultSig.isSupportedByPlugin(dt, meta.conf.decimalTypeEnabled)) {
+        meta.willNotWorkOnGpu(s"unsupported data type in output: $dt")
       }
```
gerashegalov, line 870: nit
```suggestion
      exprMeta.dataType.foreach { dt => 
        if (!resultSig.isSupportedByPlugin(dt, meta.conf.decimalTypeEnabled)) {
          meta.willNotWorkOnGpu(s"unsupported data type in output: $dt")
        }
      }
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
sperlingxx: I reverted this change.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/CostBasedOptimizer.scala
```diff
@@ -317,7 +317,7 @@ class CpuCostModel(conf: RapidsConf) extends CostModel {
 
       case _: AttributeReference | _: GetStructField =>
         MemoryCostHelper.calculateCost(MemoryCostHelper.estimateGpuMemory(
-          expr.dataType, nullable = false, rowCount), conf.cpuReadMemorySpeed)
+          expr.dataType.get, nullable = false, rowCount), conf.cpuReadMemorySpeed)
```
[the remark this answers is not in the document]
sperlingxx: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1576,7 +1576,7 @@ object GpuOverrides {
           // types were and cannot recover it. As such for now we are going to do what Spark does,
           // but we have to recompute/recheck the temporary precision to be sure it will fit
           // on the GPU.
-          val Seq(leftDataType, rightDataType) = childExprs.map(_.dataType)
+          val Seq(leftDataType, rightDataType) = childExprs.map(_.dataType.get)
```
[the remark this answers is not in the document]
sperlingxx: Done. Thanks!

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1708,7 +1708,7 @@ object GpuOverrides {
       CaseWhenCheck,
       (a, conf, p, r) => new ExprMeta[CaseWhen](a, conf, p, r) {
         override def tagExprForGpu(): Unit = {
-          if (dataType.isInstanceOf[ArrayType]) {
+          if (dataType.get.isInstanceOf[ArrayType]) {
```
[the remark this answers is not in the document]
sperlingxx: Done. Thanks!

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1787,7 +1787,7 @@ object GpuOverrides {
           // effectively calculating an extra digit of precision. Because cudf does not support this
           // right now we actually increase the scale (and corresponding precision) to get an extra
           // decimal place so we can round it in GpuCheckOverflow
-          val Seq(leftDataType, rightDataType) = childExprs.map(_.dataType)
+          val Seq(leftDataType, rightDataType) = childExprs.map(_.dataType.get)
```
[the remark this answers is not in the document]
sperlingxx: Done. Thanks!

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -541,23 +545,23 @@ case class ContextChecks(
...
-        meta.willNotWorkOnGpu(s"expression ${expr.getClass.getSimpleName} $expr " +
-            s" does not have a corresponding dataType.")
-      }
-    } else {
-      if (!outputCheck.isSupportedByPlugin(dt.get, meta.conf.decimalTypeEnabled)) {
-        meta.willNotWorkOnGpu(s"expression ${expr.getClass.getSimpleName} $expr " +
-            s"produces an unsupported type ${dt.get}")
-      }
+    meta.dataType match {
+      case Some(dt: DataType) =>
```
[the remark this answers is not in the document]
sperlingxx: Done. Thanks!

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
sperlingxx: Done. Thanks!

sperlingxx:
> Generally I think this is okay. My main problem is that it is not obvious to the end user that we changed the type. We have already been bitten by this with decimal and with timestamps where we do not explain to the user that decimal is disabled or that timestamps are disabled because the time zone is not UTC. I would like to avoid the same problem and ideally tag the types that we changed with a reason why we changed them.

Hi @revans2, I involved the reasons of type change via introducing a new class `DataTypeMeta`. It serves as a metadata, but it doesn't inherit  `RapidsMeta`.

revans2 (dismissed):
Conceptually it looks fine. I would like to see how it ends up being used in practice, but that can wait for the next PR.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -732,6 +786,50 @@ object ExpressionContext {
...
+    reason: Option[String] = None) {
+
+  lazy val dataType: Option[DataType] = desired match {
+    case Some(dt) => Some(dt)
+    case None => wrapped
+  }
+
+  // typeConverted will only be true if there exists DataType in wrapped expression
+  lazy val typeConverted: Boolean = {
+    dataType.nonEmpty && wrapped.nonEmpty && dataType.get != wrapped.get
```
gerashegalov, line 807: not a blocker, we could write:
```scala
dataType.nonEmpty && dataType == wrapped
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
gerashegalov: Did you forget to push, I actually don't see this reflected in the PR

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
sperlingxx: Yes, I forgot it. And it's done now.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
sperlingxx: Done.

sperlingxx:
build

revans2 (approved):

gerashegalov (approved):
LGTM

PR #2890, opened 2021-07-08
Fix cast to date bug

### LINKED ISSUE #2868: [BUG] to_date produces wrong value on GPU for some corner cases
**Describe the bug**

`to_date` (without a format specified) produces incorrect results for some invalid inputs. For example, the input string `3330-7 39 49` produces `null` on the CPU but produces the date `3330-07-01` on the GPU.

```
CPU: [3330-7 39 49: 1,null]

GPU: [3330-7 39 49: 1,3330-07-01]
```

I was experimenting with some basic fuzzing in the unit test.

```scala
private def generateTimestampStrings(n: Int): Seq[String] = {
  val validChars = "0123456789:-. \t\n"
  val rand = new Random(0) // fixed seed
  val list = new ListBuffer[String]()
  for (_ <- 0 to n) {
    val len = rand.nextInt(32)
    val str = new StringBuilder(len)
    for (_ <- 0 to len) {
      str.append(validChars.charAt(rand.nextInt(validChars.length)))
    }
    list += str.toString
  }
  list
}
```

**Expected behavior**
Results should be consistent between CPU and GPU.

N/A

**Additional context**
N/A

andygrove:
This PR builds on #2875 and closes #2868 

Commit # 4 adds the test and commit # 5 adds the fix.

There was a bug in the regex that sanitizes input for CAST string to date where it was removing trailing characters in some cases, converting invalid inputs into valid inputs.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
@@ -166,12 +166,10 @@ object GpuCast extends Arm {
    */
   def sanitizeStringToDate(input: ColumnVector): ColumnVector = {
     val rules = Seq(
-      /// replace yyyy-m with yyyy-mm
-      RegexReplace("(\\A\\d{4})-(\\d{1})\\Z", "\\1-0\\2"),
-      /// replace yyyy-m- with yyyy-mm-
-      RegexReplace("(\\A\\d{4})-(\\d{1}-)", "\\1-0\\2"),
-      /// replace yyyy-mm-d with yyyy-mm-dd
-      RegexReplace("(\\A\\d{4}-\\d{2})-(\\d{1})([ T]|\\Z)", "\\1-0\\2\\3")
+      // replace yyyy-m with yyyy-mm
+      RegexReplace(raw"(\A\d{4})-(\d{1})(-|\Z)", raw"\1-0\2\3"),
+      // replace yyyy-mm-d with yyyy-mm-dd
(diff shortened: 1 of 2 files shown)


jlowe:
Marking this a draft so it isn't inadvertently committed while #2875 is under review.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -158,6 +159,24 @@ object GpuCast {
...
+  /**
+   * Sanitization step for CAST string to date. Inserts a single zero before any
+   * single digit month or day.
+   */
+  def sanitizeStringToDate(input: ColumnVector): ColumnVector = {
+    val rules = Seq(
+      /// replace yyyy-m with yyyy-mm
+      RegexReplace("(\\A\\d{4})-(\\d{1})\\Z", "\\1-0\\2"),
+      /// replace yyyy-m- with yyyy-mm-
+      RegexReplace("(\\A\\d{4})-(\\d{1}-)", "\\1-0\\2"),
```
jlowe, line 172: Is it possible to combine these?  I'm thinking of something similar to what the regex just below does to handle the different endings, for example (also using `raw` strings for improved legibility):
```suggestion
      RegexReplace(raw"(\A\d{4})-(\d{1})([\-|\Z])", raw"\1-0\2\3"),
```

andygrove: Thanks, I was able to combine the first two. Also switched to raw strings.

jlowe (approved):

andygrove:
This is now rebased.

jlowe (approved):

jlowe:
build

PR #2895, opened 2021-07-09
Add in running window optimization using scan

revans2:
Spark optimizes running windows to have a linear time algorithm.  When discussing this with cudf https://github.com/rapidsai/cudf/issues/8440 it was decided to use scan and segmented_scan (group by scan).  This puts in a framework for this and adds in a few initial implementations.

In performance tests on my local box row_number and count are only slightly faster now than they were under window, but min, max, and sum all show significant performance gains similar to those I showed were possible in https://github.com/rapidsai/cudf/issues/8440

In a large max running window with no partition by I have seen performance improvements of 171x faster cold and 542x faster hot compared to the CPU

```
scala> spark.time(spark.range(0L, Int.MaxValue, 1L, numPartitions=1).select(max("id").over(Window.orderBy("id").rowsBetween(Window.unboundedPreceding, 0)).as("RN"), col("id")).orderBy(desc("RN")).show)
...
Time taken: 3633 ms
scala> spark.time(spark.range(0L, Int.MaxValue, 1L, numPartitions=1).select(max("id").over(Window.orderBy("id").rowsBetween(Window.unboundedPreceding, 0)).as("RN"), col("id")).orderBy(desc("RN")).show)
...
Time taken: 1153 ms
scala> spark.conf.set("spark.rapids.sql.enabled", "false")
scala> spark.time(spark.range(0L, Int.MaxValue, 1L, numPartitions=1).select(max("id").over(Window.orderBy("id").rowsBetween(Window.unboundedPreceding, 0)).as("RN"), col("id")).orderBy(desc("RN")).show)
...
Time taken: 622094 ms
scala> spark.conf.set("spark.rapids.sql.enabled", "false")
scala> spark.time(spark.range(0L, Int.MaxValue, 1L, numPartitions=1).select(max("id").over(Window.orderBy("id").rowsBetween(Window.unboundedPreceding, 0)).as("RN"), col("id")).orderBy(desc("RN")).show)
...
Time taken: 625660 ms
```

This is a special case because when no partition is given the data all goes to a single task, so it needs a single core to process the data. But as you can see it would still take hundreds of CPU cores in the partitioned cast to offset the performance gains.

On the previous GPU code I could not run this because I had to kill it before my GPU overheated.

This is  stepping stone to be able to support rank and dense_rank


revans2:
I tested this on databricks and it works there too.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -658,24 +721,151 @@ class GroupedAggregations extends Arm {
...
+    val allAggs = runningWindowOptimizedData.map {
+      case (func, _) =>
+        func.groupByScan(inputCb)
+    }.toSeq
+
+    // Part by is always ascending with nulls first
+    val sortedGroupingOpts = GroupByOptions.builder()
+        .withKeysSorted(true)
+        .withKeysDescending(Array.fill(partByPositions.length)(false): _*)
+        .withKeysNullSmallest(Array.fill(partByPositions.length)(true): _*)
```
jlowe, line 771: `GroupByOptions` Javadoc states that keys default to ascending and nulls smallest if not specified, so I think this is unnecessary.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -658,24 +721,151 @@ class GroupedAggregations extends Arm {
...
+    doRunningWindowOptimizedAggs(partByPositions, inputCb, outputColumns)
     doRowAggs(boundOrderSpec, orderByPositions, partByPositions, inputCb, outputColumns)
     doRangeAggs(boundOrderSpec, orderByPositions, partByPositions, inputCb, outputColumns)
   }
 }
 
 /**
- * Calculates the results of window operations
+ * Calculates the results of window operations. It assumes that any batching of the data
+ * or fixups after the fact to get the right answer is done out side of this.
```
jlowe, line 849:
```suggestion
 * or fixups after the fact to get the right answer is done outside of this.
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -197,6 +198,43 @@ case class GpuWindowExpression(windowFunction: Expression, windowSpec: GpuWindow
     case other =>
       throw new IllegalStateException(s"${other.getClass} is not a supported window function")
   }
+
+  lazy val optimizedRunningWindow: Option[GpuRunningWindowFunction] = {
+    if (System.getProperty("TEST_RUNNING", "true").toBoolean &&
```
jlowe, line 203: It would be nice to have a comment here explaining why we would default to assuming a test is running and in what context someone would explicitly specify `-DTEST_RUNNING=false` to prevent this logic.  I don't see anything setting this property to use as a reference.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -197,6 +198,43 @@ case class GpuWindowExpression(windowFunction: Expression, windowSpec: GpuWindow
     case other =>
       throw new IllegalStateException(s"${other.getClass} is not a supported window function")
   }
+
+  lazy val optimizedRunningWindow: Option[GpuRunningWindowFunction] = {
```
jlowe, line 202:
```suggestion
  private[this] lazy val optimizedRunningWindow: Option[GpuRunningWindowFunction] = {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
revans2: Oops forgot to remove that. I was using that for performance testing without recompiling things.

revans2:
build

jlowe (approved):

revans2:
With the review work I accidentally checked in a change that makes this require the fix from https://github.com/rapidsai/cudf/pull/8705. I am inclined to wait for it to get merged in, but if others want to merge this in sooner I can revert the small change and do a follow on PR when the cudf change does get merged in.

revans2:
build

revans2:
build

PR #2898, opened 2021-07-09
Add fuzz tests for cast from string to other types

andygrove:
This PR adds tests for casting from string to other types using generated strings. Every test exposed at least one new bug and issues have been filed and are referenced from this PR.

I would like to merge this PR before fixing any of the bugs so that we can enable the relevant test as part of the bug fix PRs.

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
@@ -57,100 +57,113 @@ class CastOpSuite extends GpuExpressionTestSuite {
     for (from <- supportedTypes; to <- supportedTypes) yield (from, to)
   }
 
-  private val BOOL_CHARS = " \t\r\nfalseTRUE01yesNO"
+  private val BOOL_CHARS = " \t\r\nFALSEfalseTRUEtrue01yesYESnoNO"
   private val NUMERIC_CHARS = "inf \t\r\n0123456789.+-eE"
   private val DATE_CHARS = " \t\r\n0123456789:-/TZ"
 
   ignore("Cast from string to boolean using random inputs") {
     // Test ignored due to known issues
     // https://github.com/NVIDIA/spark-rapids/issues/2902
-    castRandomStrings(DataTypes.BooleanType, BOOL_CHARS, maxStringLen = 1)
-    castRandomStrings(DataTypes.BooleanType, BOOL_CHARS, maxStringLen = 3)
-    castRandomStrings(DataTypes.BooleanType, BOOL_CHARS)
+    testCastStringTo(DataTypes.BooleanType, generateRandomStrings(BOOL_CHARS, maxStringLen = 1))
+    testCastStringTo(DataTypes.BooleanType, generateRandomStrings(BOOL_CHARS, maxStringLen = 3))
+    testCastStringTo(DataTypes.BooleanType, generateRandomStrings(BOOL_CHARS))
+  }
+
+  ignore("Cast from string to boolean using hand-picked values") {
+    testCastStringTo(DataTypes.BooleanType, Seq("\n\nN", "False", "FALSE", "false", "FaLsE",
+      "f", "F", "True", "TRUE", "true", "tRuE", "t", "T", "Y", "y", "10", "01", "0", "1"))
   }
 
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
```diff
@@ -56,6 +57,124 @@ class CastOpSuite extends GpuExpressionTestSuite {
     for (from <- supportedTypes; to <- supportedTypes) yield (from, to)
   }
 
+  private val BOOL_CHARS = " \t\r\nfalseTRUE01yesNO"
```
jlowe, line 60: Shouldn't this include `FALSE`, `true`, `YES`, and `no`, or are we not interested in testing mixed-case strings for a boolean value?

tests/src/test/scala/com/nvidia/spark/rapids/FuzzerUtils.scala
```diff
@@ -355,22 +338,26 @@ class EnhancedRandom(protected val r: Random, protected val options: FuzzerOptio
...
+}
+
+object FuzzerOptions {
+  val ALPHABET_CHARS: String = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+  val NUMERIC_CHARS: String = "0123456789"
+  val ALPHANUMERIC_CHARS: String = ALPHABET_CHARS + NUMERIC_CHARS
+  val WHITESPACE_CHARS: String = " \t\r\n"
+  val SPECIAL_CHARS: String =
+    "!@#$%^&*()-+=/?,.<>\\|[]{}~;:`\"'"
+  val ALL_CHARS: String = ALPHABET_CHARS + NUMERIC_CHARS + SPECIAL_CHARS + WHITESPACE_CHARS
```
jlowe, line 358: Are we only considering ASCII here intentionally?  The name `ALL_CHARS` to me implies a wider coverage than it actually is.

jlowe: Also even if we're targeting ASCII, it's missing some ASCII control characters (e.g. characters for ordinal values 0x01, 0x02, etc.)

tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
andygrove: I updated this but also added some tests for hand-picked edge cases since we can't guarantee that the random data will cover valid mixed case versions of "true" and "false", especially if the fuzzer is modified in the future.

tests/src/test/scala/com/nvidia/spark/rapids/FuzzerUtils.scala
andygrove: I've removed these pre-defined sets of characters. The default now is to delegate to Scala's `Random.nextString` unless the fuzzer is created with a specific set of characters, as demonstrated in the CAST fuzzer tests in the PR.

tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
```diff
@@ -56,6 +57,139 @@ class CastOpSuite extends GpuExpressionTestSuite {
...
+    // Test ignored due to known issues
+    // https://github.com/NVIDIA/spark-rapids/issues/2902
+    testCastStringTo(DataTypes.BooleanType,
+      generateRandomStrings(Some(BOOL_CHARS), maxStringLen = 1))
+    testCastStringTo(DataTypes.BooleanType,
+      generateRandomStrings(Some(BOOL_CHARS), maxStringLen = 3))
+    testCastStringTo(DataTypes.BooleanType, generateRandomStrings(Some(BOOL_CHARS)))
+  }
+
+  ignore("Cast from string to boolean using hand-picked values") {
```
jlowe, line 74: Should there be a comment here explaining why this test instance is ignored?  The others have a comment but this one is missing.

jlowe (commented):
Missing a comment on an ignored test but otherwise lgtm.

jlowe (approved):

jlowe:
build

Issue #2902, opened 2021-07-09
[BUG] CAST string to bool produces incorrect results in some cases


andygrove:
**Describe the bug**
Casting the string `\n\nN` to boolean should produce `false` but produces `null` on the GPU. We should be stripping whitespace and performing a case-insensitive comparison.

**Steps/Code to reproduce bug**
Run tests in https://github.com/NVIDIA/spark-rapids/pull/2898

**Expected behavior**
GPU and CPU should produce same results

PR #2903, opened 2021-07-09
Fix cast string to bool

andygrove:
Builds on https://github.com/NVIDIA/spark-rapids/pull/2898 and closes https://github.com/NVIDIA/spark-rapids/issues/2902

In CAST string to bool, we were not removing whitespace and were performing a case-sensitive comparison so we did not recognize `TRUE` as a boolean value.

The bug fix is in this commit: https://github.com/NVIDIA/spark-rapids/pull/2903/commits/3488fb4b90e892aadb65e1a50ff46c33fdfc0f35

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
@@ -655,21 +655,25 @@ case class GpuCast(
 
     // determine which values are valid bool strings
     withResource(ColumnVector.fromStrings(boolStrings: _*)) { boolStrings =>
-      withResource(input.contains(boolStrings)) { validBools =>
-        // in ansi mode, fail if any values are not valid bool strings
-        if (ansiEnabled) {
-          withResource(validBools.all()) { isAllBool =>
-            if (!isAllBool.getBoolean) {
-              throw new IllegalStateException(GpuCast.INVALID_INPUT_MESSAGE)
+      withResource(input.strip()) { stripped =>
+        withResource(stripped.lower()) { lower =>
+          withResource(lower.contains(boolStrings)) { validBools =>
+            // in ansi mode, fail if any values are not valid bool strings
+            if (ansiEnabled) {
+              withResource(validBools.all()) { isAllBool =>
+                if (!isAllBool.getBoolean) {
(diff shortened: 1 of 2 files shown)


jlowe (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -638,21 +638,25 @@ case class GpuCast(
...
     // determine which values are valid bool strings
     withResource(ColumnVector.fromStrings(boolStrings: _*)) { boolStrings =>
-      withResource(input.contains(boolStrings)) { validBools =>
-        // in ansi mode, fail if any values are not valid bool strings
-        if (ansiEnabled) {
-          withResource(validBools.all()) { isAllBool =>
-            if (!isAllBool.getBoolean) {
-              throw new IllegalStateException(GpuCast.INVALID_INPUT_MESSAGE)
+      withResource(input.strip()) { stripped =>
+        withResource(stripped.lower()) { lower =>
```
gerashegalov, line 642: technically, JDK roundtrips the string toUpper->toLower  in order to normalize the string for case-insensitive comparison https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/String.java#L1941 but we should be ok with the target values "true" "false".

jlowe: Spark has [custom code to do the conversion](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala#L75-L77).  Spark isn't using the JDK to do the cast.

gerashegalov: Thanks for pointing this out. Let us say Spark does not use equalsIgnoreCase :)

andygrove:
build

jlowe (approved):
Now that the dependency has been merged and this rebased, is there a reason it's still in draft?

andygrove:
> Now that the dependency has been merged and this rebased, is there a reason it's still in draft?

No. I just forgot to change it last night. Updated.

gerashegalov (approved):

Issue #2757, opened 2021-06-21
[FEA] Profiling tool display input data types


tgravescs:
**Is your feature request related to a problem? Please describe.**
We want the profiling tool to be able to display what input types were read from the spark event log when possible.

it should also indicate if one of the types isn't supported by the rapids plugin.

tgravescs:
so it looks like the datasource v1 does list all of the types read in the metadata in ReadSchema.  This does not look like its the case for datasource v2 where it only puts it in the plan string so it gets truncated. I've only looked at parquet so far but for v1 we should be able to do this and v2 for possibly just a couple of columns

PR #2904, opened 2021-07-09
Qualification and Profiling tool handle Read formats and datatypes

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/2757

Here is the summary of changes:

1) Parse eventlogs for the datasource Read schema. This is different between datasource v1 and datasource v2 readers.  v2 readers do not include the entire schema, it gets truncated. v1 has the entire schema.
2) For profiling tool it just prints out the information it finds -> format, schema, location, filters. This works in compare mode as well.
3) Modify our dist pom file so that on verify it generates a CSV file of the supported read formats and their datatypes. It also looks at some configs that are off by default. This file gets included in the tools jar and is read by the qualification tool only.
4) The starting Score has been changed to be the total task times rather than the sql Duration/app duration. We think this will be more accurate and take out the issue of how many executors and it will handle the case where you have 12 hour job and 3 hours of it is DF. This before was ranked below for instance a 3 second job that was all DF ops.
5) Qualification tool now has a read format score which just looks for any unsupported (configured off by default) formats or data types. If we find any unsupported it will take off from the starting score based on a configurable percent. Default is 20%. so if starting score is 100, the read format part of that score is 20. If we don't find any unsupported formats or data types, it stays 20, if we find all formats unsupported, that goes to 0 and the final score is 80.
6) rankings of summary report changed to use the new score and it's printed to csv file. The read data source format and schema can be printed to CSV file as well but it's off by default because it can be very long.
7) tests updated to the new scoring algorithm.
8) qualification summary info also printed to stdout and made smaller to fit in 80 characters.  I could split this off if you want, I was already modifying  so just included it.


tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -1624,3 +1631,70 @@ object SupportedOpsDocs {
     }
   }
 }
+
+object SupportedOpsForTools {
```
tgravescs, line 1635: @revans2 would be great if you could look at this part since you did Type checking stuff

tools/README.md
```diff
@@ -133,21 +136,20 @@ Note that SQL queries that contain failed jobs are not included.
 
 Sample output in csv:
 ```
-App Name,App ID,Score,Potential Problems,SQL Dataframe Duration,App Duration,Executor CPU Time Percent,App Duration Estimated,SQL Duration with Potential Problems,SQL Ids with Failures
-job1,app-20210507174503-2538,98.13,"",952802,970984,63.14,false,0,""
-job2,app-20210507180116-2539,97.88,"",903845,923419,64.88,false,0,""
-job3,app-20210319151533-1704,97.59,"",737826,756039,33.95,false,0,""
+App Name,App ID,Score,Potential Problems,SQL DF Duration,SQL Dataframe Task Duration,App Duration,Executor CPU Time Percent,App Duration Estimated,SQL Duration with Potential Problems,SQL Ids with Failures,Read Score Percent,ReadFileFormat Score
```
tgravescs, line 139: perhaps here I should also list any formats/datatypes that we found that were not supported.  That maybe easier and shorter for user to see what might be problematic rather then listing entire format or just score

tgravescs: added in a new column for the ones not supported

tgravescs:
build

tools/src/main/scala/org/apache/spark/sql/rapids/tool/qualification/QualAppInfo.scala
```diff
@@ -118,31 +146,75 @@ class QualAppInfo(
...
+    ToolUtils.calculateDurationPercent(totalCpuTime, totalRunTime)
+  }
+
+  private def getAllReadFileFormats: String = {
+    dataSourceInfo.map { ds =>
+      s"${ds.format.toLowerCase()}[${ds.schema}]"
+    }.mkString(":")
+  }
+  
+  // For the read score we looks at all the read formats and datatypes for each
```
nartal1, line 158: looks s/b look

nartal1:
Overall it looks good. I don't have any additional comments as it meets the user requirements. 
Wanted to know if you tried running large number of events logs with this patch and if the total runtime was not affected considerably.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/qualification/QualAppInfo.scala
tgravescs: fixed

tgravescs:
yes ran it over the tpcds results time was 44 seconds so about the same time as before.

tgravescs:
build

tgravescs:
build

nartal1 (approved):

Issue #2907, opened 2021-07-11
[BUG] incorrect expression to detect previously set `--master`


abellina:
The string match expression on the right of: https://github.com/NVIDIA/spark-rapids/pull/2792/files#diff-146d61aa008bf60d3c970d1d76208d85b87070466a6d6021fa63eb7478037a27R154, is not working in our linux CI box (or in my linux desktop). 

Fix is incoming. In some cases, the tests may be forcing a local cluster, even if the intention was to start the job in a standalone cluster.

PR #2908, opened 2021-07-11
Fixes bad string contains in run_pyspark_from_build

abellina:
Fixes: https://github.com/NVIDIA/spark-rapids/issues/2907

Double square brackets are required here for the string match operation, otherwise bash is going to exactly try to match `*` instead of expanding that to be a wildcard.


abellina:
build

pxLi (approved):

Issue #2790, opened 2021-06-23
[BUG] In Comparison mode health check does not show the application id


revans2:
At least in comparison mode there is no way to know what application the task, stage, job, or query was a part of. It really only works if you are looking at a single application.  The health checks are just printed out one after another with no label for which application it is for.

tgravescs:
https://github.com/NVIDIA/spark-rapids/pull/2762 added in the appIndex to the for the failed tables, but it does still look like the removed blockmanagers and removed executors table don't have the app index or id. from app index you can get id and app name.

PR #2914, opened 2021-07-12
Profiling tool: add app index to tables that don't have it

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/2790

This adds the appIndex column to tables that were missing it.  This is to removed block managers, removed executors and then sql plan metrics.  Updated tests to match.


tgravescs:
build

jlowe (approved):

Issue #2916, opened 2021-07-13
[FEA] Support GpuCollectList and GpuCollectSet as TypedImperativeAggregate


sperlingxx:
**Is your feature request related to a problem? Please describe.**
We have supported `GpuCollectList` and `GpuCollectSet` as `GpuAggregateWindowFunction`. However, we don't support them as `TypedImperativeAggregate` (under GroupBy or Reduction context) yet.

**Describe the solution you'd like**
According to the [design doc](https://docs.google.com/document/d/1sMzpv_cB7P-ZIE_-tN2A-A9l6DkVLd4z4d94QgSV3_E), we can split the entire work into several parts:

1. - [x] Extend TypeChecks: enable tagging types with the information from RapidsMeta instead of raw data types (related PR: #2888)
2. - [x] Rework `GpuHashAggregate.setupReferences`(related PR: #2917)
3. - [x] Support `GpuCollectList` and `GpuCollectSet` as ` GpuTypedImperativeAggregate` (related PR: #2971)
4.  - [ ] Support the data conversion of typed imperative aggregation buffers between the CPU layout and the GPU layout.

There are some doubts on how to implement item 4. Fortunately, item 4 doesn't block other items. And the feature can be partially enabled even without item 4.

PR #2917, opened 2021-07-13
Refine GpuHashAggregateExec.setupReference

sperlingxx:

Current PR refined `GpuHashAggregateExec.setupReference` to support `TypedImperativeAggregate`, which is required by #2916 .  The primary work of current PR is reorganizing the code of `boundInputReferences`.

Originally, we workaround `PartialMerge` modes when we construct `boundInputReferences`.  To be specific, the code piece dealing with aggregations which contain both `PartialMerge` and  `Partial` mode is not generic enough. It regards `GpuPivotFirst`  and `GpuAverage` as exceptional cases:
```scala
      case Partial =>
        // Partial with distinct case
        val updateExpressionsCudfAggsDistinct =
          updateExpressionsDistinct.filter(_.isInstanceOf[CudfAggregate])
              .map(_.asInstanceOf[CudfAggregate].ref)
        if (inputProjectionsDistinct.exists(p => !p.isInstanceOf[NamedExpression])) {
          // Case of distinct average we need to evaluate the "GpuCast and GpuIsNotNull" columns.
          // Refer to how input projections are setup for GpuAverage.
          // In the case where we have expressions to evaluate, pick the unique attributes
          // references from them as you only have one column for it before you start evaluating.
          distinctExpressions = inputProjectionsDistinct
          distinctAttributes = inputProjectionsDistinct.flatMap(ref =>
            ref.references.toSeq).distinct
        } else {
          distinctAttributes = updateAttributesDistinct
          distinctExpressions = updateExpressionsCudfAggsDistinct
        }
```

With the new implementation, we avoid all kinds of workaround.  Instead, we fit all kinds of aggregation modes into three categories by the stage of AggregateExec in aggregation stack:
1. The first stage of aggregation stack, in terms of aggregation modes, it may contain aggregate expressions with `Partial` or `Complete` mode. It may also contain no aggregate expressions. For the first aggregation stage, the input projections are necessary, because it consume the outputs of non-Aggregate plans.
2. The final-like stages, including three conditions: 
    * the final (last) stage of aggregation stack
    * the (partial) merge stage merely for `nonDistinctAggAttributes` (the second stage of aggregation stack with one distinct)

   For final-like stages, we just pass through all output attributes of child plan.
3. The third stage of aggregation stack with one distinct,  in terms of aggregation modes, it contains both `Partial` mode (for distinctAgg) and `PartialMerge` mode (for nonDistinctAgg). For this stage, we need to switch the position of distinctAttributes and nonDistinctAttributes to match the output schema of the previous stage:
   *  the schema of the 2nd stage's outputs: _groupingAttributes ++ distinctAttributes ++ nonDistinctAggBufferAttributes_
   *  the schema of the 3rd stage's aggregate expressions: _nonDistinctMergeAggExpressions ++ distinctPartialAggExpressions_

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
@@ -629,7 +629,7 @@ class GpuHashAggregateIterator(
     // boundInputReferences is used to pick out of the input batch the appropriate columns
     // for aggregation.
     //
-    // - PartialMerge with Partial mode: we use the inputProjections or distinct update expressions
+    // - PartialMerge with Partial mode: we use the inputProjections
     //   for Partial and non distinct merge expressions for PartialMerge.
     // - Final or PartialMerge-only mode: we pick the columns in the order as handed to us.
     // - Partial or Complete mode: we use the inputProjections or distinct update expressions.
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
@@ -627,111 +627,81 @@ class GpuHashAggregateIterator(
     val aggModeCudfAggregates =
       AggregateUtils.computeAggModeCudfAggregates(aggregateExpressions, aggBufferAttributes)
 
-    //
-    // expressions to pick input to the aggregate, and finalize the output to the result projection.
-    //
-    // Pick update distinct attributes or input projections for Partial
-    val (distinctAggExpressions, nonDistinctAggExpressions) = aggregateExpressions.partition(
-      _.isDistinct)
-    val updateExpressionsDistinct =
-      distinctAggExpressions.flatMap(
-        _.aggregateFunction.updateExpressions)
-    val updateAttributesDistinct =
-      distinctAggExpressions.flatMap(_.aggregateFunction.aggBufferAttributes)
-    val inputProjectionsDistinct =
-      distinctAggExpressions.flatMap(_.aggregateFunction.inputProjection)
-
-    // Pick merge non-distinct for PartialMerge
-    val mergeExpressionsNonDistinct =
-      nonDistinctAggExpressions
-          .flatMap(_.aggregateFunction.mergeExpressions)
-          .map(_.asInstanceOf[CudfAggregate].ref)
-    val mergeAttributesNonDistinct =
-      nonDistinctAggExpressions.flatMap(
-        _.aggregateFunction.aggBufferAttributes)
-
-    // Partial with no distinct or when modes are empty
-    val inputProjections: Seq[Expression] = groupingExpressions ++ aggregateExpressions
-        .flatMap(_.aggregateFunction.inputProjection)
-
-    var distinctAttributes = Seq[Attribute]()
-    var distinctExpressions = Seq[Expression]()
-    var nonDistinctAttributes = Seq[Attribute]()
-    var nonDistinctExpressions = Seq[Expression]()
-    modeInfo.uniqueModes.foreach {
-      case PartialMerge =>
-        nonDistinctAttributes = mergeAttributesNonDistinct
-        nonDistinctExpressions = mergeExpressionsNonDistinct
-      case Partial =>
-        // Partial with distinct case
-        val updateExpressionsCudfAggsDistinct =
-          updateExpressionsDistinct.filter(_.isInstanceOf[CudfAggregate])
-              .map(_.asInstanceOf[CudfAggregate].ref)
-        if (inputProjectionsDistinct.exists(p => !p.isInstanceOf[NamedExpression])) {
-          // Case of distinct average we need to evaluate the "GpuCast and GpuIsNotNull" columns.
-          // Refer to how input projections are setup for GpuAverage.
-          // In the case where we have expressions to evaluate, pick the unique attributes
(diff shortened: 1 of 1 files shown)


sperlingxx:
build

abellina:
Could you add to the description for this PR around what exactly was changed and why? Please also note this PR is in flight: https://github.com/NVIDIA/spark-rapids/pull/2859, and also makes changes to `setupReferences`.

sperlingxx:
build

sperlingxx:
> Could you add to the description for this PR around what exactly was changed and why? Please also note this PR is in flight: #2859, and also makes changes to `setupReferences`.

I added the description. And I rebased this branch onto latest main branch.

sperlingxx:
build

revans2:
@abellina and @jlowe could both of you take a look at this? You have a lot more experience with the aggregate code than I do. It looks okay to me, but I don't feel 100% qualified to approve it.

abellina:
@revans2 yeap, I am looking.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -627,83 +627,67 @@ class GpuHashAggregateIterator(
...
-
-    // Pick merge non-distinct for PartialMerge
-    val mergeExpressionsNonDistinct =
-      nonDistinctAggExpressions
+    // - PartialMerge with Partial mode: we use the inputProjections or distinct update expressions
+    //   for Partial and non distinct merge expressions for PartialMerge.
+    // - Final or PartialMerge-only mode: we pick the columns in the order as handed to us.
+    // - Partial or Complete mode: we use the inputProjections or distinct update expressions.
+    val boundInputReferences =
+    if (modeInfo.hasPartialMerge && modeInfo.uniqueModes.contains(Partial)) {
```
abellina, line 638:
```suggestion
    if (modeInfo.hasPartialMerge && modeInfo.hasPartialMode) {
```

abellina: bonus points => `hasPartialMerge` => `hasPartialMergeMode`

abellina (changes requested):
Thanks for the added details @sperlingxx. I am mostly curious on the update expressions, as we used to use them before in the PartialMerge + Partial case https://github.com/NVIDIA/spark-rapids/blob/branch-21.08/sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala#L679. I think the comments need to be adjusted but it would be nice to make sure that you found that input projections is all that is needed in that case (perhaps the shapes match?) 

Tagging @kuhushukla  and @nartal1 as they may be more familiar with both distinct and pivot.

Had other minor nits.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -627,83 +627,67 @@ class GpuHashAggregateIterator(
...
+    } else if (modeInfo.hasPartialMode || modeInfo.hasCompleteMode ||
+        modeInfo.uniqueModes.isEmpty) {
+      // The first aggregation stage (including Partial or Complete or no aggExpression),
+      // whose child node is not an AggregateExec. Therefore, input projections are essential.
+      val inputProjections: Seq[Expression] = groupingExpressions ++ aggregateExpressions
+          .flatMap(_.aggregateFunction.inputProjection)
       GpuBindReferences.bindGpuReferences(inputProjections, childAttr)
+    } else {
+      // This branch should NOT be reached.
+      throw new IllegalStateException(s"invalid unique modes: ${modeInfo.uniqueModes}")
```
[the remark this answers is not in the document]
sperlingxx: Fixed

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -627,83 +627,67 @@ class GpuHashAggregateIterator(
...
-        _.aggregateFunction.updateExpressions)
-    val updateAttributesDistinct =
-      distinctAggExpressions.flatMap(_.aggregateFunction.aggBufferAttributes)
-    val inputProjectionsDistinct =
-      distinctAggExpressions.flatMap(_.aggregateFunction.inputProjection)
-
-    // Pick merge non-distinct for PartialMerge
-    val mergeExpressionsNonDistinct =
-      nonDistinctAggExpressions
+    // - PartialMerge with Partial mode: we use the inputProjections or distinct update expressions
```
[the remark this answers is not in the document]
sperlingxx: I updated this part along with comments, in order to verify and clarify "input projections is all that is needed in that case".

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
sperlingxx: We used `modeInfo.uniqueModes.contains(Partial)` instead of `modeInfo.hasPartialMode` here,  because hasPartialMode is set as `hasPartialMerge || uniqueModes.contains(Partial)`. I adjusted `AggregateModeInfo` to make its members more consistent with their names.

abellina (approved):
I had a couple of nits, but they are minor. Approving ahead of time. Thanks for making the tweaks.

sperlingxx:
build

abellina (approved):
LGTM

PR #2919, opened 2021-07-13
Fix bugs in CAST string to integer

### LINKED ISSUE #2899: [BUG] CAST string to integer produces incorrect results in some cases
**Describe the bug**
I added some fuzz testing in https://github.com/NVIDIA/spark-rapids/pull/2898 and it exposed some issues with CAST string to integer types.

The input string `"6965\n23E++2"` should cast to a null value, but the GPU produces the value 6965.

https://github.com/NVIDIA/spark-rapids/pull/2898

andygrove:

Closes #2899

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
@@ -737,17 +737,16 @@ case class GpuCast(
 
     withResource(GpuCast.sanitizeStringToIntegralType(input, ansiEnabled)) { sanitized =>
       withResource(sanitized.isInteger(dType)) { isInt =>
-        if (ansiEnabled) {
-          withResource(isInt.all()) { allInts =>
-            if (!allInts.getBoolean) {
-              throw new NumberFormatException(GpuCast.INVALID_INPUT_MESSAGE)
-            }
-          }
-          sanitized.castTo(dType)
-        } else {
-          withResource(sanitized.castTo(dType)) { parsedInt =>
-            withResource(GpuScalar.from(null, dataType)) { nullVal =>
-              isInt.ifElse(parsedInt, nullVal)
+        withResource(isInt.all()) { allInts =>
+          if (allInts.getBoolean) {
+            sanitized.castTo(dType)
+          } else if (ansiEnabled) {
+            throw new NumberFormatException(GpuCast.INVALID_INPUT_MESSAGE)
+          } else {
+            withResource(sanitized.castTo(dType)) { parsedInt =>
+              withResource(GpuScalar.from(null, dataType)) { nullVal =>
+                isInt.ifElse(parsedInt, nullVal)
+              }
             }
           }
         }
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
@@ -179,18 +179,18 @@ object GpuCast extends Arm {
   def sanitizeStringToIntegralType(input: ColumnVector, ansiEnabled: Boolean): ColumnVector = {
     // Convert any strings containing whitespace to null values. The input is assumed to already
     // have been stripped of leading and trailing whitespace
-    val sanitized = withResource(GpuScalar.from(null, DataTypes.StringType)) { nullVal =>
-      withResource(input.containsRe("\\s")) { hasWhitespace =>
-        withResource(hasWhitespace.any()) { any =>
-          if (any.getBoolean) {
-            if (ansiEnabled) {
-              throw new NumberFormatException(GpuCast.INVALID_INPUT_MESSAGE)
-            } else {
+    val sanitized = withResource(input.containsRe("\\s")) { hasWhitespace =>
+      withResource(hasWhitespace.any()) { any =>
+        if (any.getBoolean) {
+          if (ansiEnabled) {
+            throw new NumberFormatException(GpuCast.INVALID_INPUT_MESSAGE)
+          } else {
+            withResource(GpuScalar.from(null, DataTypes.StringType)) { nullVal =>
               hasWhitespace.ifElse(nullVal, input)
             }
-          } else {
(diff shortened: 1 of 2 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -175,6 +175,59 @@ object GpuCast extends Arm {
       cv.stringReplaceWithBackrefs(rule.search, rule.replace)
     })
   }
+
+  def sanitizeStringToIntegralType(input: ColumnVector, ansiEnabled: Boolean): ColumnVector = {
+    // Convert any strings containing whitespace to null values. The input is assumed to already
+    // have been stripped of leading and trailing whitespace
+    val sanitized = withResource(GpuScalar.from(null, DataTypes.StringType)) { nullVal =>
```
jlowe, line 182: We only need the null val for the I suspect somewhat uncommon case where there's at least one value with whitespace and we're not in ANSI mode.  Therefore we should only create this GPU-synchronizing resource in the case we need it rather than up-front for all cases.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -175,6 +175,59 @@ object GpuCast extends Arm {
...
+          withResource(sanitized.stringContains(dot)) { hasDot =>
+            // only do the decimal sanitization if any strings do contain dot
+            withResource(hasDot.any(DType.BOOL8)) { anyDot =>
+              if (anyDot.getBoolean) {
+                // Special handling for strings that have no numeric value before the dot, such
+                // as "." and ".1" because extractsRe returns null for the capture group
+                // for these values and it also returns null for invalid inputs so we need this
+                // explicit check
+                withResource(sanitized.matchesRe("^[+\\-]?\\.[0-9]*$")) { startsWithDot =>
+                  withResource(sanitized.extractRe("^([+\\-]?[0-9]*)\\.([0-9]*)?$")) { table =>
```
jlowe, line 214: It looks like there are two capture groups here, but we're only looking at one of them.  It would be nice to avoid manifesting the second group if we don't need it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -175,6 +175,59 @@ object GpuCast extends Arm {
...
+                    withResource(Scalar.fromString("0")) { zero =>
+                      withResource(startsWithDot.ifElse(zero, table.getColumn(0).incRefCount())) {
```
jlowe, line 216: There's an refcount increment here, but who is responsible for closing the extra reference?

andygrove: Thanks for catching that. The `incRefCount` was in the original code but now that we're passing this into the `ifElse` it was no longer needed.

andygrove:
@jlowe This is read for re-review. Thanks.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -681,46 +734,19 @@ case class GpuCast(
...
-          withResource(cleaned.castTo(dType)) { parsedInt =>
-            withResource(GpuScalar.from(null, dataType)) { nullVal =>
-              isInt.ifElse(parsedInt, nullVal)
+    withResource(GpuCast.sanitizeStringToIntegralType(input, ansiEnabled)) { sanitized =>
+      withResource(sanitized.isInteger(dType)) { isInt =>
+        withResource(isInt.all()) { allInts =>
+          if (allInts.getBoolean) {
+            sanitized.castTo(dType)
+          } else if (ansiEnabled) {
+            throw new NumberFormatException(GpuCast.INVALID_INPUT_MESSAGE)
```
jlowe, line 744: Are null inputs already covered for ANSI mode?  `all()` skips nulls.

andygrove: Well that opened quite the can of worms. I have added tests for ANSI mode and fixed null handling and some other issues that came up.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -175,6 +175,70 @@ object GpuCast extends Arm {
...
+        }
+      }
+    }
+
+    if (ansiEnabled) {
+      // ansi mode only supports simple integers, so no exponents, decimal places, or Inf
+      val regex = "^[+\\-]?[0-9]+$"
+      withResource(sanitized.matchesRe(regex)) { isNumeric =>
+        withResource(isNumeric.not()) { notNumeric =>
+          withResource(notNumeric.any()) { any =>
```
jlowe, line 203: Saw the most recent commit changed from `!all()` to this which is less efficient (due to the need to manifest `not()`).  `not()` shouldn't change whether something's null or not and `any()`, like `all()`, ignores nulls, so I'm curious how this fixed anything.  Also are there issues when `notNumeric` is entirely a null batch?

jlowe: I think I sorta answered my own question here a bit -- one main difference is that when the batch is entirely null, `any()` will do the right thing whereas I suspect `all()` will not.  `all()` does _not_ return true if the batch is empty or entirely null.  That means my suggestion at https://github.com/NVIDIA/spark-rapids/pull/2933#discussion_r669977924 probably has the same issue and needs to be reverted to the original behavior.

andygrove: Right, I found this very confusing and spent a lot of time wondering why my null tests were failing. I am starting to think that I should build some convenience methods for some of these operations that we repeat in a lot of the casting code as well. I will review the other change.

jlowe (approved):

jlowe:
build

PR #2922, opened 2021-07-13
Improve UCX assertion to show the failed assertion

andygrove:

I had an error in my setup and got an `assertion failed` with no additional context. This PR simply shows what the failed assertion was, saving a bit of time in debugging.

## Before:

```
Caused by: java.lang.AssertionError: assertion failed
	at scala.Predef$.assert(Predef.scala:208)
	at org.apache.spark.sql.rapids.GpuShuffleEnv$.setRapidsShuffleManager(GpuShuffleEnv.scala:91)
```
## After:

```
Caused by: java.lang.AssertionError: assertion failed: com.nvidia.spark.rapids.spark301.RapidsShuffleManager == com.nvidia.spark.rapids.spark311.RapidsShuffleManager
	at scala.Predef$.assert(Predef.scala:223)
	at org.apache.spark.sql.rapids.GpuShuffleEnv$.setRapidsShuffleManager(GpuShuffleEnv.scala:92)
```

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuShuffleEnv.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuShuffleEnv.scala
@@ -88,7 +88,8 @@ object GpuShuffleEnv extends Logging {
     managerOpt: Option[RapidsShuffleInternalManagerBase] = None): Unit = {
     if (managerOpt.isDefined) {
       val manager = managerOpt.get
-      assert(manager.getClass.getCanonicalName == GpuShuffleEnv.RAPIDS_SHUFFLE_CLASS)
+      assert(manager.getClass.getCanonicalName == GpuShuffleEnv.RAPIDS_SHUFFLE_CLASS,
+        s"${manager.getClass.getCanonicalName} != ${GpuShuffleEnv.RAPIDS_SHUFFLE_CLASS}")
       logInfo("RapidsShuffleManager is initialized")
     }
     mgr = managerOpt
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuShuffleEnv.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuShuffleEnv.scala
@@ -89,7 +89,7 @@ object GpuShuffleEnv extends Logging {
     if (managerOpt.isDefined) {
       val manager = managerOpt.get
       assert(manager.getClass.getCanonicalName == GpuShuffleEnv.RAPIDS_SHUFFLE_CLASS,
-        s"${manager.getClass.getCanonicalName} != ${GpuShuffleEnv.RAPIDS_SHUFFLE_CLASS}")
+        s"${manager.getClass.getCanonicalName} == ${GpuShuffleEnv.RAPIDS_SHUFFLE_CLASS}")
       logInfo("RapidsShuffleManager is initialized")
     }
     mgr = managerOpt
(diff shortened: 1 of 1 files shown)


abellina (approved):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuShuffleEnv.scala
```diff
@@ -88,7 +88,8 @@ object GpuShuffleEnv extends Logging {
     managerOpt: Option[RapidsShuffleInternalManagerBase] = None): Unit = {
     if (managerOpt.isDefined) {
       val manager = managerOpt.get
-      assert(manager.getClass.getCanonicalName == GpuShuffleEnv.RAPIDS_SHUFFLE_CLASS)
+      assert(manager.getClass.getCanonicalName == GpuShuffleEnv.RAPIDS_SHUFFLE_CLASS,
```
jlowe, line 91: If you got this in a bad setup, maybe this should be a `require` instead of an `assert` so it fires outside of unit tests as well?  We'd want an improved message that users can understand for that case.

andygrove: I do see this assertion when running from Spark shell.

```
$ ./spark-shell-ucx.sh 
21/07/14 14:43:05 WARN Utils: Your hostname, nvworkstation resolves to a loopback address: 127.0.1.1; using 192.168.0.10 instead (on interface eno1)
21/07/14 14:43:05 WARN Utils: Set SPARK_LOCAL_IP if you need to bind to another address
21/07/14 14:43:06 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Using Spark's default log4j profile: org/apache/spark/log4j-defaults.properties
...
21/07/14 14:43:10 ERROR SparkContext: Error initializing SparkContext.
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	...
Caused by: java.lang.AssertionError: assertion failed
	at scala.Predef$.assert(Predef.scala:208)
	at org.apache.spark.sql.rapids.GpuShuffleEnv$.setRapidsShuffleManager(GpuShuffleEnv.scala:91)
	at org.apache.spark.sql.rapids.RapidsShuffleInternalManagerBase.<init>(RapidsShuffleInternalManagerBase.scala:233)
	at org.apache.spark.sql.rapids.shims.spark301.RapidsShuffleInternalManager.<init>(RapidsShuffleInternalManager.scala:29)
	at com.nvidia.spark.rapids.spark301.RapidsShuffleManager.<init>(RapidsShuffleManager.scala:25)
	... 69 more
```

jlowe: If this is something the user can trigger then we should use `require` rather than `assert` since the former is about checking inputs rather than sanity of the program state.  Ultimately it's more about the type of exception that will be thrown, but this is also a check we don't wan to drop if someone decides to compile with assertions off.

The obtuse stack trace as seen by the user also drives home the point about making a better message when this error does trigger, regardless of whether it's an assert or require.

abellina (approved):

jlowe (approved):

jlowe:
build

andygrove:
Tests against 3.1.3-SNAPSHOT failed with

```
java.lang.NoSuchMethodError: org.apache.spark.sql.execution.datasources.parquet.ParquetFilters.<init>(Lorg/apache/parquet/schema/MessageType;ZZZZIZ)V
[2021-07-16T20:40:17.896Z] 	at com.nvidia.spark.rapids.GpuParquetFileFilterHandler.filterBlocks(GpuParquetScan.scala:269)
[2021-07-16T20:40:17.896Z] 	at com.nvidia.spark.rapids.GpuParquetMultiFilePartitionReaderFactory.$anonfun$buildBaseColumnarReaderForCoalescing$1(GpuParquetScan.scala:365)
```

pxLi:
build

Issue #2807, opened 2021-06-24
[BUG] Use UCP_AM_FLAG_WHOLE_MSG and UCP_AM_FLAG_PERSISTENT_DATA for receive handlers


abellina:
For receive handlers that want to call `amData.receive` after the UCX callback completes (i.e. rndv/buffers) we must pass `UCP_AM_FLAG_PERSISTENT_DATA` in `setAmRecvHandler`. The flags argument to this function is getting added in JUCX 1.11-rc1, and we don't have an official jar published with it yet (other than SNAPSHOT). We also should set `UCP_AM_FLAG_WHOLE_MSG` to prevent fragment-sized multiple callbacks for eager messages.

The plan is to move to JUCX 1.11-SNAPSHOT and test it works, then PR that. There should be a 1.11-rc1 jar published (officially) next week, and we should be able to move to that, irrespective of when the non-RC 1.11 jar is published. Note: there's a 1.11.0 JAR published currently, but it was published by mistake => so I'd imagine the official one would be 1.11.0.1 or 1.11.1.

abellina:
This, unfortunately, is blocked until UCX 1.11 native is released as well. UCX 1.10.1 is not able to work with GPU buffers, so we can't move to Active Messages yet for all buffers. UCX 1.11 RC1 was tagged so we are going towards that version.

Issue #2925, opened 2021-07-13
[DOC] Update documentation, LICENSE, and dockerfiles for UCX 1.11.0


abellina:
For 21.08 we are looking to set our minimum version to UCX 1.11.0. This issue is to change our documentation to state this, dockerfiles should also be updated to fetch the UCX 1.11 artifacts.

abellina:
There's a diff in the LICENSE file in the UCX repo for 1.11.0-rc4 that we need to add prior to tagging our release of 0.28:

https://github.com/openucx/ucx/commit/49a75241607872b1e103a341bfa47343a7fc3f33

Issue #2892, opened 2021-07-09
[BUG] UCX error `Message truncated` observed with UCX 1.11 RC in Q77 NDS


abellina:
We are seeing an error in Q77 NDS at 3TB:
```
 jucx_common_def.cc:257  UCX  ERROR JUCX: request error: Message truncated
```
I used these UCX settings:

```
spark.executorEnv.UCX_RNDV_SCHEME=put_zcopy
spark.executorEnv.UCX_ERROR_SIGNALS=
spark.executorEnv.UCX_IB_GPU_DIRECT_RDMA=yes
spark.executorEnv.UCX_IB_RX_QUEUE_LEN=1024
```
This error only happens with the latest 21.08 nightly (rapids-4-spark_2.12-21.08.0-20210708.152651-40.jar) build and UCX 1.11 RC (https://github.com/openucx/ucx/releases/tag/v1.11.0-rc3).

Reverting UCX to 1.10.1 (https://github.com/openucx/ucx/releases/tag/v1.10.1) works. So this appears to be a regression with UCX 1.11, or at least a bad interplay between JUCX 1.11.0 (as opposed to JUCX 1.11.0-RC3) and the native bits. 

UCX 1.11 allows logging, and the logs are just showing:

```
ucp_request.c:524  UCX  DEBUG message truncated: recv_length 8230 offset 0 buffer_size 4480
 jucx_common_def.cc:257  UCX  ERROR JUCX: request error: Message truncated
```

I am checking the other side of the connection, and I don't see 8230 bytes sent, so I think this is a fragment size (i.e. we are potentially falling back to fragment based copies for such tiny buffers).

abellina:
@petro-rudenko @yosefe. Any ideas of any behavior change related to this?

abellina:
With JUCX 1.11.0-rc3 this issue goes away. There's a separate issue @petro-rudenko and I are trying to debug that prevents me to simply update to the JUCX 1.11.0-rc3 jar, but the truncation issue goes away for Q77.

abellina:
The issue can also be seen with JUCX 1.11.0-rc3. For now, we have verbose logs, and are looking into the root cause.

abellina:
With GPU active messages I am not able to reproduce this. The error above is specific to UCX tag-based messages, which are in use in the currently merged code.

I am going to move forward towards active messages after re-checking all of NDS.

FYI @petro-rudenko

abellina:
[...]

PR #2926, opened 2021-07-13
GPU Active Messages for all buffer types

abellina:
This undoes [this revert](https://github.com/NVIDIA/spark-rapids/pull/2805).

The change is pretty much the same, but I had to fix diffs as there were conflicts with the [UCPListener change](https://github.com/NVIDIA/spark-rapids/pull/2886).

I upgrade UCX to 1.11.0-rc3 in this change also, and handle:

> Once JUCX 1.11 is published with the change to setAmRecvHandler, the change would be to bring this back, and also set: UCP_AM_FLAG_PERSISTENT_DATA (for the use-after-callback pattern) and UCP_AM_FLAG_WHOLE_MSG (to prevent UCX to call us multiple times with eager fragments in the future).

I also tested https://github.com/NVIDIA/spark-rapids/issues/2892, and I can see that the issue is not reproducible with plain active messages (we no longer use tags).

Updating the docs for UCX 1.11.0 is tracked here: https://github.com/NVIDIA/spark-rapids/issues/2925. I am hoping we can fix the docs once 1.11.0 is not a release candidate.

Closes https://github.com/NVIDIA/spark-rapids/issues/2807.


jlowe (approved):

jlowe:
build

Issue #2912, opened 2021-07-12
[BUG] Profiling compare mode fails when comparing spark 2 eventlog to spark 3 event log


tgravescs:
**Describe the bug**
I think the issue is with the resourceProfileid column

Reproduce:
$SPARK_HOME/bin/spark-submit  --master local[*] --driver-memory 30g --class com.nvidia.spark.rapids.tool.profiling.ProfileMain  ./rapids-4-spark-tools_2.12-21.08.0-SNAPSHOT.jar  -c  tools/src/test/resources/spark-events-profiling/tasks_executors_fail_compressed_eventlog.zstd tools/src/test/resources/spark-events-profiling/spark2-eventlog.zstd

```
21/07/12 13:36:35 INFO ToolTextFileWriter: Output location: rapids_4_spark_profile/rapids_4_spark_tools_output.log
Exception in thread "main" org.apache.spark.sql.AnalysisException: Union can only be performed on tables with the same number of columns, but the first table has 12 columns and the second table has 11 columns;
'Sort ['appIndex ASC NULLS FIRST], true
+- 'Distinct
   +- 'Union false, false
      :- Project [1 AS appIndex#1097, resourceProfileId#449, numExecutors#1095L, totalCores#448 AS executorCores#1098, maxMem#396L, maxOnHeapMem#397L, maxOffHeapMem#398L, executorMemory#367L, numGpusPerExecutor#368, executorOffHeap#369L, taskCpu#370, taskGpu#371]
      :  +- Join Inner, (maxId#1096 = executorId#393)
      :     :- Join Inner, (resourceProfileId#449 = id#365)
      :     :  :- SubqueryAlias t
      :     :  :  +- Aggregate [resourceProfileId#449, totalCores#448], [resourceProfileId#449, totalCores#448, count(executorId#446) AS numExecutors#1095L, max(executorId#446) AS maxId#1096]
      :     :  :     +- SubqueryAlias executorsdf_1
      :     :  :        +- LocalRelation [executorID#446, host#447, totalCores#448, resourceProfileId#449]
      :     :  +- SubqueryAlias rp
      :     :     +- SubqueryAlias resourceprofilesdf_1
      :     :        +- LocalRelation [id#365, executorCores#366, executorMemory#367L, numGpusPerExecutor#368, executorOffHeap#369L, taskCpu#370, taskGpu#371]
      :     +- SubqueryAlias bm
      :        +- SubqueryAlias blockmanagersdf_1
[...]

PR #2932, opened 2021-07-14
Profiling tool: Fix comparing spark2 and spark3 event logs

tgravescs:
Profiling compare mode fails when comparing spark 2 eventlog to spark 3 event log, the issue is spark2 is missing the resource profile id column.  Just make it so the column is always there.

fixes https://github.com/NVIDIA/spark-rapids/issues/2912


tgravescs:
build

jlowe (approved):

PR #2933, opened 2021-07-14
Improve CAST string to float implementation to handle more edge cases

### LINKED ISSUE #2900: [BUG] CAST string to float/double produces incorrect results in some cases
**Describe the bug**
Casting string to float produces incorrect results in some cases. Examples:

|Input|CPU|GPU|
|------|---|---|
| `0260995.3E\r\t` | `null` | `260995.3` |
| `e` | `null` | `0` |
| `-1e\n` | `null` | `-1.0` |
| `5f ` | 5.0 | null | 
	
Discovered by running the fuzz tests in https://github.com/NVIDIA/spark-rapids/pull/2898

**Expected behavior**
CPU and GPU should produce the same results

andygrove:

Closes #2900


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -175,6 +175,64 @@ object GpuCast extends Arm {
...
+        val withoutWhitespace = withResource(stripped.containsRe("(\r|\n)")) {
+            hasVerticalWhitespace =>
+          hasVerticalWhitespace.ifElse(nullString, stripped)
+        }
+        // replace all possible versions of infinity with Inf
+        withResource(withoutWhitespace) { _ =>
+          withResource(withoutWhitespace.stringReplaceWithBackrefs(
+            "([iI][nN][fF])([iI][nN][iI][tT][yY])?", "Inf")) { inf =>
+            // replace +Inf with Inf
+            withResource(inf.stringReplaceWithBackrefs("([+])(Inf)", "Inf")) {
```
jlowe, line 214: Why is this using a regex with backrefs?  I don't see any need for backrefs since we aren't using any captured groups in the replacement.  Isn't this simply a standard replace of `+Inf` with `Inf`?  That would be much faster.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -175,6 +175,64 @@ object GpuCast extends Arm {
...
+        ")" +
+        "|(Inf)" +                  // Infinity
+        "|([nN][aA][nN])" +         // NaN
+      ")" +
+      "$"                           // end of line
+
+    withResource(input.lstrip()) { stripped =>
+      withResource(GpuScalar.from(null, DataTypes.StringType)) { nullString =>
+        // filter out strings containing breaking whitespace
+        val withoutWhitespace = withResource(stripped.containsRe("(\r|\n)")) {
```
jlowe, line 205: Are the parentheses needed here?  Also this may be more efficient as a `contains` using a column containing the two scalar values since it doesn't need a full regex engine for that.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -175,6 +175,64 @@ object GpuCast extends Arm {
...
+    // This regex gets applied after the transformation to normalize use of Inf and is
+    // just strict enough to filter out known edge cases that would result in incorrect
+    // values. We further filter out invalid values using the cuDF is_float method.
+    val VALID_FLOAT_REGEX =
+      "^" +                         // start of line
+      "[+\\-]?" +                   // optional + or - at start of string
+      "(" +
+        "(" +
+          "(" +
+            "([0-9]+)|" +           // digits, OR
```
jlowe, line 190: Nit: It seems like there's many parentheses in this regex that aren't needed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -175,6 +175,64 @@ object GpuCast extends Arm {
       cv.stringReplaceWithBackrefs(rule.search, rule.replace)
     })
   }
+
+  def sanitizeStringToFloat(input: ColumnVector): ColumnVector = {
+
+    // This regex gets applied after the transformation to normalize use of Inf and is
+    // just strict enough to filter out known edge cases that would result in incorrect
+    // values. We further filter out invalid values using the cuDF is_float method.
```
jlowe, line 183:
```suggestion
    // values. We further filter out invalid values using the cuDF isFloat method.
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -175,6 +175,64 @@ object GpuCast extends Arm {
...
+            withResource(inf.stringReplaceWithBackrefs("([+])(Inf)", "Inf")) {
+                infWithoutPlus =>
+              // filter out any strings that are not valid floating point numbers
+              val floatOrNull = withResource(infWithoutPlus.matchesRe(VALID_FLOAT_REGEX)) {
+                  looksLikeFloat =>
+                looksLikeFloat.ifElse(infWithoutPlus, nullString)
+              }
+              // strip floating-point designator 'f' or 'd' but don't strip the 'f' from 'Inf'
+              withResource(floatOrNull) { _ =>
+                withResource(floatOrNull.matchesRe("^[+\\-]?Inf$")) { isInf =>
```
jlowe, line 223: There isn't a regex that can strip the trailing designator without needing to check for `Inf` separately first?  I think we just want a string regex replace of `(^n)[fFdD]` with `\1`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -737,59 +795,41 @@ case class GpuCast(
...
-      withResource(withPosNegInfinityReplaced.matchesRe(NAN_REGEX)) { isNan =>
+      withResource(sanitized.matchesRe(NAN_REGEX)) { isNan =>
         // now check if the values are floats
-        withResource(withPosNegInfinityReplaced.isFloat()) { isFloat =>
+        withResource(sanitized.isFloat) { isFloat =>
           if (ansiEnabled) {
             withResource(isNan.not()) { notNan =>
               withResource(isFloat.not()) { notFloat =>
                 withResource(notFloat.and(notNan)) { notFloatAndNotNan =>
                   withResource(notFloatAndNotNan.any()) { notNanAndNotFloat =>
```
jlowe, line 823: I think this can be simplified with something like:
```suggestion
            withResource(isNan.or(isFloat)) { nanOrFloat =>
              withResource(nanOrFloat.all()) { allNanOrFloat =>
                if (!allNanOrFloat.getBoolean) {
```

andygrove:
Thanks @jlowe this is ready for another review.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
andygrove: I was able to remove some of the parenthesis. I couldn't remove them from `([0-9]+)` on line190 though because cuDF complained with `invalid regex pattern: nothing to repeat at position 16`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -175,6 +175,64 @@ object GpuCast extends Arm {
...
+        val withoutWhitespace = withResource(ColumnVector.fromStrings("\r", "\n")) {
+            verticalWhitespace =>
+          withResource(stripped.contains(verticalWhitespace)) {
+            _.ifElse(nullString, stripped)
+          }
+        }
+        // replace all possible versions of infinity with Inf
+        val inf = withResource(withoutWhitespace) { _ =>
+          withoutWhitespace.stringReplaceWithBackrefs(
+            "([iI][nN][fF])([iI][nN][iI][tT][yY])?", "Inf")
```
jlowe, line 214: Does this need a capture group?  This is a case where we really want to just call `stringReplaceRegex` but apparently that doesn't exist in the cudf Java bindings even though libcudf supports it. 😞  Still, I think this becomes a cheaper regex without the capture group.  Wouldn't this be equivalent?
```suggestion
            "[iI][nN][fF])([iI][nN][iI][tT][yY]", "Inf")
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -175,6 +175,64 @@ object GpuCast extends Arm {
...
+          }
+        }
+        // filter out any strings that are not valid floating point numbers according
+        // to the regex pattern
+        val floatOrNull = withResource(infWithoutPlus.matchesRe(VALID_FLOAT_REGEX)) {
+            _.ifElse(infWithoutPlus, nullString)
+        }
+        // strip floating-point designator 'f' or 'd' but don't strip the 'f' from 'Inf'
+        withResource(floatOrNull) {
+          _.stringReplaceWithBackrefs("(Inf|-Inf|[nN][aA][nN]|[^InNaAfFdD]*)([fFdD]?)", "\\1")
```
jlowe, line 231: I'm not sure why this regex needs to be so complicated.  We've already filtered the strings to things that have to look like floats, and as the comment above says, we want to strip trailing `f` or `d` as long as it's not part of `Inf`.  Looking at the `VALID_FLOAT_REGEX` definition above, I don't see the harm in making this a very straightforward regex of essentially "drop trailing f's and d's as long as they aren't preceded by an n".
```suggestion
          _.stringReplaceWithBackrefs("([^n])[fFdD]?$", "\\1")
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -737,59 +795,37 @@ case class GpuCast(
...
+    // 1. convert the different infinities to "Inf"/"-Inf" which is the only variation cudf
+    // understands
+    // 2. identify the nans
+    // 3. identify the floats. "nan", "null" and letters are not considered floats
+    // 4. if ansi is enabled we want to throw and exception if the string is neither float nor nan
+    // 5. convert everything thats not floats to null
+    // 6. set the indices where we originally had nans to Float.NaN
+    //
+    // NOTE Limitation: "1.7976931348623159E308" and "-1.7976931348623159E308" are not considered
+    // Inf even though spark does
```
jlowe, line 807:
```suggestion
    // Inf even though Spark does
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
andygrove: I did try variations on this and couldn't find anything that worked. The challenge is that we need to retain the `n` in the capture group so that we retain the `Inf` to pass to cuDF. Perhaps I am just missing some regex trick here though.

jlowe: > The challenge is that we need to retain the n in the capture group

But the regex example I showed above shouldn't hit at all on `Inf` because it's looking for anything _except_ an `n` character to capture before it finds a case-insensitive `f` or `d`.  And what it does capture it emits, so I'm totally confused how this could be doing anything at all on `Inf` or how it could eat an `n` character which should alone disqualify the expression from matching.

jlowe: Maybe try dropping the `?` which I think is unnecessary.  We want to strip any `f` or `d` character at the end of the string if it's not preceded by `n`.  If there's no `f` or `d` at the end then we don't want it to match at all and emit the original string.
```
stringReplaceWithBackrefs("([^n])[fFdD]$", "\\1")
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
andygrove: I committed this, then saw tests fail and realized I had misread this. There are two groups here - (Inf) and (inity) and the second grouping is optional so that we replace all variations of Inf and Infinity with the standardized version,

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
andygrove: Thanks. That works perfectly. I'm confused now why I couldn't get this to work yesterday.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
jlowe: Oops, I made the same mistake when reading it.  Funny how the parentheses can disappear in the middle. 😅

jlowe: Since we don't care about actually capturing the groups for backreference, I think these groups should be marked as non-capturing to hopefully use a cheaper regex kernel, i.e.: something like:
```
"(?:[iI][nN][fF])(?:[iI][nN][iI][tT][yY])?"
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -176,6 +176,78 @@ object GpuCast extends Arm {
...
+        val withoutWhitespace = withResource(ColumnVector.fromStrings("\r", "\n")) {
+            verticalWhitespace =>
+          withResource(stripped.contains(verticalWhitespace)) {
+            _.ifElse(nullString, stripped)
+          }
+        }
+        // replace all possible versions of "Inf" and "Infinity" with "Inf"
+        val inf = withResource(withoutWhitespace) { _ =>
+            withoutWhitespace.stringReplaceWithBackrefs(
+          "([iI][nN][fF])" + "([iI][nN][iI][tT][yY])?", "Inf")
```
jlowe, line 214: These should be non-capturing groups since we don't need the captures. Hopefully this hints us into a cheaper regex kernel, as this really should be a regex replace rather than a backref replace.

jlowe (approved):

andygrove:
build

andygrove:
Build failed with unrelated test failure:

```
[2021-07-19T16:37:26.722Z] - Cast from string to date using random inputs with valid year prefix *** FAILED ***
[2021-07-19T16:37:26.722Z]   Mismatch casting string [2021-012] to DateType. CPU: 2021-12-01; GPU: null (CastOpSuite.scala:235)
```

I am investigating.

andygrove:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -176,6 +176,78 @@ object GpuCast extends Arm {
...
+        }
+        // filter out any strings that are not valid floating point numbers according
+        // to the regex pattern
+        val floatOrNull = withResource(infWithoutPlus) { _ =>
+          withResource(infWithoutPlus.matchesRe(VALID_FLOAT_REGEX)) { isFloat =>
+            if (ansiEnabled) {
+              withResource(isFloat.all()) { allMatch =>
+                // Check that all non-null values are valid floats. Note that isFloat will be false
+                // if all rows are null so we need to check for that condition.
+                if (!allMatch.getBoolean &&
```
jlowe, line 232: So this is technically invalid.  `allMatch.getBoolean` is not valid to call if `allMatch.isValid` returns false.  I think we can avoid the separate, explicit null count check and change this to be something like:
```
if (allMatch.isValid && !allMatch.getBoolean) {
```

We can do a similar technique to restore my original `any()` to `all()` suggestion I made earlier and avoid manifesting the `not()` intermediates.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -774,59 +846,37 @@ case class GpuCast(
...
-                    if (notNanAndNotFloat.getBoolean()) {
-                      throw new NumberFormatException(GpuCast.INVALID_FLOAT_CAST_MSG)
-                    }
-                  }
+            withResource(isNan.or(isFloat)) { nanOrFloat =>
+              withResource(nanOrFloat.all()) { allNanOrFloat =>
+                // Check that all non-null values are valid floats or NaN. Note that
+                // allNanOrFloat will be false if all rows are null so we need to check
+                // for that condition.
+                if (!allNanOrFloat.getBoolean && sanitized.getNullCount != sanitized.getRowCount) {
```
jlowe, line 873: Same issue and suggestion regarding `allNanOrFloat.isValid` here.

jlowe (approved):

jlowe:
build

PR #2934, opened 2021-07-14
Qualification tool: Add filter based on appName

nartal1:
This PR includes basic implementation of filtering by partially reading eventlogs.
Additional filtering features can be added easily as we capture the required information in this PR.
Added filtering based on appName.


tools/src/main/scala/com/nvidia/spark/rapids/tool/qualification/QualificationMain.scala
```diff
@@ -64,4 +72,8 @@ object QualificationMain extends Logging {
     val res = qual.qualifyApps(eventLogInfos)
     (0, res)
   }
+
+  def argsContainsAppFilters(appArgs:QualificationArgs): Boolean = {
```
nartal1, line 76: We can add more checks as we add new filter features.

nartal1:
Will resolve the conflicts and update README.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -0,0 +1,95 @@
...
+    qualFilter.shutdown()
+    if (!qualFilter.awaitTermination(waitTimeInSec, TimeUnit.SECONDS)) {
+      logError(s"Processing log files took longer then $waitTimeInSec seconds," +
+          " stopping processing any more event logs")
+      qualFilter.shutdownNow()
+    }
+    var eventlog = filteredeventLogs.asScala.map(x => x.eventlog).toSeq
+
+    // This will be required to do the actual filtering
+    val allSumfilteredEventLogs = filteredeventLogs.asScala.map(x => (x.appInfo, x.eventlog)).toSeq
```
tgravescs, line 71: why do the map, just do asScala and then use that below.  This might be cleaner like this:

```
  // This will be required to do the actual filtering
    val apps = filteredeventLogs.asScala

    if (appArgs.applicationName.isSupplied && filterAppName.nonEmpty) {
      val filtered = apps.filter { app =>
        val appNameOpt = app.appInfo.map(_.appName)
        if (appNameOpt.isDefined) {
          appNameOpt.get.equals(filterAppName)
        } else {
          // in complete log file
          false
        }
      }
      filtered.map(_.eventlog).toSeq
    } else {
      apps.map(x => x.eventlog).toSeq
    }
```

tools/src/main/scala/org/apache/spark/sql/rapids/tool/FilterAppInfo.scala
```diff
@@ -0,0 +1,57 @@
...
+package org.apache.spark.sql.rapids.tool
+
+import com.nvidia.spark.rapids.tool.EventLogInfo
+import org.apache.hadoop.conf.Configuration
+
+import org.apache.spark.scheduler.{SparkListenerApplicationStart, SparkListenerEvent}
+
+case class ApplicationStartInfo(
+    appName: String,
+    appId: Option[String],
```
tgravescs, line 26: appId doesn't seem to be needed?

nartal1:
build

tgravescs (approved):

PR #2935, opened 2021-07-15
Read the complete batch before returning when selectedAttributes is empty

### LINKED ISSUE #2891: [BUG] Discrepancy in getting count before and after caching
**Describe the bug**
There is a different count reported before caching a DF and after caching a DF

```
scala> val data = spark.read.parquet("/data/perf_test")
data: org.apache.spark.sql.DataFrame = [id: string, byte_col: tinyint ... 18 more fields]

scala> val dataAll = data.selectExpr("byte_col", "short_col", "integer_col", "long_col", "float_col", 
     |     "double_col", "boolean_col", "timestamp_col", "dec_32_col", "dec_64_col", "struct_col")
dataAll: org.apache.spark.sql.DataFrame = [byte_col: tinyint, short_col: smallint ... 9 more fields]

scala> dataAll.count
res0: Long = 600000000                                                          

scala> dataAll.filter("byte_col = true").count
res1: Long = 4723659                                                            

scala> dataAll.filter("byte_col = true").cache().count
res2: Long = 360                                                                

scala> dataAll.filter("byte_col = true").count
res3: Long = 360
```

razajafri:
When the `selectedAttributes` was empty, there was an optimization that would return an empty batch of the same number of rows, but that wasn't playing well with how we are doing a row count and the count was always returned as `0` in the GPU case, but in the CPU case, it was returning a different number corresponding to the number of partitions. 

There was also a bug in converting `CatchedBatch`es to Columnar/InternalRow, the code was only reading the first batch before moving on to the next partition. 

fixes #2891 

### DIFF
--- a/integration_tests/src/main/python/cache_test.py
+++ b/integration_tests/src/main/python/cache_test.py
@@ -279,3 +279,25 @@ def test_cache_additional_types(enable_vectorized, with_x_session, select_expr):
 
     # NOTE: we aren't comparing cpu and gpu results, we are comparing the cached and non-cached results.
     assert_equal(reg_result, cached_result)
+
+@pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
+@pytest.mark.parametrize('with_x_session', [with_gpu_session, with_cpu_session])
+@pytest.mark.parametrize('enable_vectorized_conf', enable_vectorized_confs, ids=idfn)
+@ignore_order
+def test_cache_count(data_gen, with_x_session, enable_vectorized_conf):
+    def with_cache(cached):
+        def helper(spark):
+            df = unary_op_df(spark, data_gen)
+            if cached:
+                df.cache().count()
+            return df.count()
+        return helper
+
+    conf = enable_vectorized_conf.copy()
+    conf.update(allow_negative_scale_of_decimal_conf)
+    conf.update({"spark.rapids.sql.test.batchsize": "100"})
+    reg_result = with_x_session(with_cache(False), conf)
+    cached_result = with_x_session(with_cache(True), conf)
+
+    assert_equal(reg_result, cached_result)
+
--- a/integration_tests/src/main/python/cache_test.py
+++ b/integration_tests/src/main/python/cache_test.py
@@ -56,7 +56,7 @@ all_gen = [StringGen(), ByteGen(), ShortGen(), IntegerGen(), LongGen(),
            pytest.param(DoubleGen(special_cases=double_special_cases), marks=[incompat]),
            BooleanGen(), DateGen(), TimestampGen()] + decimal_gens
 
-@pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
+@pytest.mark.parametrize('data_gen', [BooleanGen()], ids=idfn)
 @pytest.mark.parametrize('join_type', ['Left', 'Right', 'Inner', 'LeftSemi', 'LeftAnti'], ids=idfn)
 @pytest.mark.parametrize('enable_vectorized_conf', enable_vectorized_confs, ids=idfn)
 @ignore_order
--- a/integration_tests/src/main/python/cache_test.py
+++ b/integration_tests/src/main/python/cache_test.py
@@ -280,7 +280,7 @@ def test_cache_additional_types(enable_vectorized, with_x_session, select_expr):
     # NOTE: we aren't comparing cpu and gpu results, we are comparing the cached and non-cached results.
     assert_equal(reg_result, cached_result)
 
-def test_function_on_cached_df(with_x_session, func, data_gen, enable_vectorized_conf):
+def function_to_test_on_cached_df(with_x_session, func, data_gen, test_conf):
     def with_cache(cached):
         def helper(spark):
             df = unary_op_df(spark, data_gen)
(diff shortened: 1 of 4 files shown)


razajafri:
build

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -1056,6 +1079,7 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
       sharedConf: Broadcast[Map[String, String]]) {
 
     val conf: SQLConf = getConfFromMap(sharedConf)
+    val bytesAllowedPerBatch = new RapidsConf(conf).bytesAllowedPerBatch
```
jlowe, line 1082: No need to build an entire `RapidsConf` just to get one conf setting from it.
```suggestion
    val bytesAllowedPerBatch = RapidsConf.TEST_CACHE_BATCH_SIZE.get(conf)
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -928,6 +928,14 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(false)
 
+  val _2GB: Long = 2L * 1024 * 1024 * 1024
+  val APPROX_PAR_META_DATA: Int = 10 * 1024 * 1024 // we are estimating 10MB
```
jlowe, line 932: Rather than pollute the public members of this class with essentially temporary values, I'd rather see something like this:
```
private val TEST_CACHE_BATCH_SIZE_DEFAULT = {
  val _2gb = 2L * 1024 * 1024 * 1024
  val approx_par_meta_data = 10 * 1024 * 1024
  _2gb - approx_par_meta_data
}

**Comment by jlowe** [comment]
> This PR makes multiple changes to the way we were handling case when selectedAttributes was empty.

Can you elaborate a bit more in the PR description what the changes are?  This essentially says, "stuff changed when selectedAttributes is empty" which doesn't tell me anything I didn't already know from the PR headline.

File: sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala [hunk 3]
```diff
@@ -928,6 +928,14 @@ object RapidsConf {
...
+  val TEST_CACHE_BATCH_SIZE = conf("spark.rapids.sql.test.batchsize")
```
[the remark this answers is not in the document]
razajafri: I just didn't think of using that because I was using this for testing purposes, but I think it makes more sense to use the existing conf

integration_tests/src/main/python/cache_test.py
```diff
@@ -279,3 +279,40 @@ def helper(spark):
...
+    test_function_on_cached_df(with_x_session, lambda df: df.count(), data_gen, enable_vectorized_conf)
+
+@pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
+@pytest.mark.parametrize('with_x_session', [with_cpu_session, with_gpu_session])
+@pytest.mark.parametrize('enable_vectorized_conf', enable_vectorized_confs, ids=idfn)
+@ignore_order
+# This tests the cached and uncached values returned by collect on the CPU and GPU.
+# When running on the GPU with the DefaultCachedBatchSerializer, to project the results Spark adds a ColumnarToRowExec
+# to be able to show the results which will cause this test to throw an exception as it's not on the GPU so we have to
+# add that case to the `allowed` list. As of now there is no way for us to limit the scope of allow_non_gpu based on a
```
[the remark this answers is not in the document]
razajafri: So this is how I understand it, when `ParquetCachedBatchSerializer` isn't being used, the `InMemoryTableScanExec` is not on the GPU. Now if we set the `spark.sql.inMemoryColumnarStorage.enableVectorizedReader` to true, that causes the plan to use the columnar option of the serializer, that means that the output from the `InMemoryTableScanExec` is going to be columnar, therefore it needs to be converted to rows before we can collect it.

razajafri:
build

integration_tests/src/main/python/cache_test.py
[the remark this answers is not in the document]
jlowe: I'm still confused here because just about every test uses `collect()` to grab the results, but I don't see most tests needing to exclude `ColumnarToRowExec` in order to perform the collect without a failure.  Do we really understand this, or was it just a workaround to get the test to pass?  Why don't other tests need this when they collect?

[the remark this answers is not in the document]
razajafri: Our tests don't cache, this is strictly for `InMemoryTableScanExec`. This is how I understand this. I would love to know what @revans2 thinks about my explanation

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -324,7 +324,7 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
       conf: SQLConf): RDD[CachedBatch] = {
 
     val rapidsConf = new RapidsConf(conf)
-    val bytesAllowedPerBatch = rapidsConf.bytesAllowedPerBatch
+    val bytesAllowedPerBatch = rapidsConf.gpuTargetBatchSizeBytes - APPROX_PAR_META_DATA
```
jlowe, line 327: Do we need to max this with `1` or something to prevent it from going negative?  This comment applies to the other places where this computation is done below.

razajafri: Good point. Let me add a min batch size in the code

shims/spark311/src/main/scala/com/nvidia/spark/rapids/shims/spark311/ParquetCachedBatchSerializer.scala
```diff
@@ -987,14 +978,27 @@ class ParquetCachedBatchSerializer extends CachedBatchSerializer with Arm {
...
+       */
+      def nextBatch: Boolean = {
+        try {
           for (vector <- columnVectors) {
             vector.reset()
           }
           columnarBatch.setNumRows(0)
-          if (rowsReturned >= totalRowCount) return false
+          if (rowsReturned >= totalRowCount) {
+            parquetFileReader.close()
```
jlowe, line 999: This has the problem that unless the iterator is drained completely, `parquetFileReader` is not guaranteed to be closed.

Instead of peppering `catch` clauses everywhere, I suggest making `CurrentBatchIterator` closeable and then install a task completion handler at the higher level that guarantees the iterator is always closed when the task completes, closing any other instances as you move between instances of the iterator along the way.

razajafri: OK, but then we still need to handle the case when/if an exception is thrown anywhere in `CurrentBatchIterator` don't we? May be we need to wrap this in the `CloseableIterator` in addition to this "peppering"? Or does the task completion handler take care of this even if there is an exception?

jlowe: No, we're basically delegating to the user of the iterator to ensure it's closed even in exceptions.  Think of it as the iterator sort of wrapping the parquet reader, the reader will be closed either when the iterator is drained or when the caller closes the iterator, and it's the responsibility of the iterator user to ensure it's closed even in exceptional conditions.

> Or does the task completion handler take care of this even if there is an exception?

Exactly.  The exception will bubble up and kill the task, causing the completion handler to be invoked.  Or worse-case if someone up the call stack eats the exception, the task will eventually complete in some form and ensure this is not indefinitely leaked.

razajafri:
build

razajafri:
I have found a problem that will cause the nightly build to fail. Let me fix that before this is merged.

razajafri:
build

razajafri:
Not sure whats going on with the CI but other PRs are also failing the CI @pxLi are you aware of this?

pxLi:
build

integration_tests/src/main/python/cache_test.py
```diff
@@ -56,7 +56,7 @@ def test_passing_gpuExpr_as_Expr(enable_vectorized_conf):
            pytest.param(DoubleGen(special_cases=double_special_cases), marks=[incompat]),
            BooleanGen(), DateGen(), TimestampGen()] + decimal_gens
 
-@pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
+@pytest.mark.parametrize('data_gen', [BooleanGen()], ids=idfn)
```
jlowe, line 59: Why would we stop testing most of the data types for this test?

razajafri:
Build

jlowe (approved):

PR #2940, opened 2021-07-15
qualification tool: add filtering by app start time

tgravescs:
Resolved story:
 As a user, I want to select log files for applications whose application start event occurred within a specific time range so that I can select only jobs I've run, e.g., in the last week, month, or quarter 

Note I also fixed a few bugs in here with the while loop going over lines.  one with counting events and a second when exception thrown dealing with incrementing i.


tgravescs:
build

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -76,7 +77,18 @@ class AppFilterImpl(
         if (appNameOpt.isDefined) {
           appNameOpt.get.equals(filterAppName)
         } else {
-          // in complete log file
+          // incomplete log file
+          false
+        }
+      }
+      filtered.map(_.eventlog).toSeq
+    } else if (appArgs.startAppTime.isSupplied) {
```
nartal1, line 85: Should we support both the cases? i.e If applciationName is given along with time range in the commandline. 
 --application-name "appName" --start-app-time "1w"
If AND is not supported, then we can document that. I will follow the same in my follow on PR's for limit.

tgravescs: yeah that is a good point we talked about this, it is officially a P1 so perhaps lets do taht as a follow on once we have all the filtering features in and make sure we have tests that will cover all the cases.

PR #2947, opened 2021-07-16
Qualification tool: Filter based on timestamp in event logs

nartal1:
User story: 

1. As a user, I want to select at most k log files overall, keeping either the newest or oldest k logs (based on the timestamp inside the event log)
2.  As a user, I want to select at most k log files for each unique application name, keeping either the newest or oldest k logs so that I can ensure that I process every distinct application I have run even if I process only a small subset of logs


nartal1:
build

pxLi:
build

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -18,16 +18,16 @@ package org.apache.spark.sql.rapids.tool
 
 import java.util.Calendar
 import java.util.concurrent.{ConcurrentLinkedQueue, Executors, ThreadPoolExecutor, TimeUnit}
-
```
tgravescs, line 21: need to fix these, build failing with scalastyle

tools/README.md
```diff
@@ -196,12 +196,23 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
...
-  -f, --filter-criteria  <arg>      Filter newest or oldest N eventlogs for
-                                    processing.eg: 100-newest (for processing
-                                    newest 100 event logs). eg: 100-oldest (for
-                                    processing oldest 100 event logs)
+  -f, --filter-criteria  <arg>      Filter newest or oldest N eventlogs based on 
+                                    timestamp of file for processing. eg: 100-newest 
+                                    (for processing newest 100 event logs). eg: 100-oldest
+                                    (for processing oldest 100 event logs). Filesystem
+                                    based filtering happens before any application based filtering.
+                                    Application based filter-criteria are:
```
tgravescs, line 204: I'm debating on the names here.  I would think most people probably want to filter based on the app time so I wonder if we use the 100-newest to be app time and then make like a 100-newest-filesystem for filesystem time.  I hate to change the config but don't think many people probably use it yet.

tools/src/main/scala/com/nvidia/spark/rapids/tool/EventLogPathProcessor.scala
```diff
@@ -183,11 +183,16 @@ object EventLogPathProcessor extends Logging {
...
       matched.take(numberofEventLogs)
-    }.getOrElse(matchedLogs)
-
+    } else {
+      matchedLogs
+    }
     filteredLogs.keys.toSeq
   }
 
+  def filterByAppName(filterNLogs: Option[String]): Boolean = {
```
tgravescs, line 192: the name here seems wrong since its either by app name or by app time, perhaps use something like filerByAppCriteria

tools/src/main/scala/com/nvidia/spark/rapids/tool/qualification/QualificationArgs.scala
```diff
@@ -42,10 +42,18 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
           " eg: s3a://<BUCKET>/eventlog1 /path/to/eventlog2")
   val filterCriteria: ScallopOption[String] =
     opt[String](required = false,
-      descr = "Filter newest or oldest N eventlogs for processing." +
+      descr = "Filter newest or oldest N eventlogs based on timestamp of file for processing." +
```
tgravescs, line 45: I think we should reword the description to have the first line state filter by filesystem timestamp, app timstamp, or app unique....   Then explain each one.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -18,16 +18,16 @@ package org.apache.spark.sql.rapids.tool
...
 import com.google.common.util.concurrent.ThreadFactoryBuilder
 import com.nvidia.spark.rapids.tool.EventLogInfo
+import com.nvidia.spark.rapids.tool.EventLogPathProcessor.logError
 import com.nvidia.spark.rapids.tool.qualification.QualificationArgs
 import org.apache.hadoop.conf.Configuration
-
 import org.apache.spark.internal.Logging
 
+import scala.collection.mutable.LinkedHashMap
```
tgravescs, line 29: this should go above with other scala imports

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -98,7 +100,38 @@ class AppFilterImpl(
...
+        val numberofEventLogs = filteredInfo(0).toInt
+        val criteria = filteredInfo(1)
+        val filtered = if (criteria.equals("oldest")) {
+          appTimeFiltered.toSeq.sortBy(_.appInfo.get.startTime).take(numberofEventLogs)
+        } else {
+          appTimeFiltered.toSeq.sortBy(_.appInfo.get.startTime).reverse.take(numberofEventLogs)
+        }
+        filtered
+      } else if (filterCriteria.endsWith("-per-app-name")) {
+        val distinctAppNameMap = apps.groupBy(_.appInfo.get.appName)
```
tgravescs, line 115: so this one isn't using the other filters above? I would think the start time would at least apply, I think we could even have the app name apply as well and just of those it would take the newest.  We may want to document this as well just saying this is applied after the others

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -302,6 +302,36 @@ class AppFilterSuite extends FunSuite {
       "3w", 2)
   }
 
+  test("full app name exact and 2-oldest-overall no match from filename") {
```
tgravescs, line 305: name here isn't exact so change description.

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -302,6 +302,36 @@ class AppFilterSuite extends FunSuite {
...
+    testFileSystemTimeAndStartAndAppFull(appsFullWithFsToTest, "2-oldest-overall", "nds", "nomatch",
+      "3w", 0)
+  }
+
+  test("full app name exact and 2-oldest-overall") {
+    testFileSystemTimeAndStartAndAppFull(appsFullWithFsToTest, "2-oldest-overall", "nds", "nds",
+      "3w", 2)
```
tgravescs, line 312: this isn't verifying oldest vs newest

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -302,6 +302,36 @@ class AppFilterSuite extends FunSuite {
...
+  }
+
+  test("full app name exact and 2-newest-overall") {
+    testFileSystemTimeAndStartAndAppFull(appsFullWithFsToTest, "2-newest-overall", "ndsweeks2",
+      "nds", "3w", 1)
+  }
+
+  test("full app name exact and 1-newest-per-app-name") {
+    testFileSystemTimeAndStartAndAppFull(appsFullWithFsToTest, "1-newest-per-app-name", "nds",
+      "nds", "3w", 3)
```
tgravescs, line 322: we don't validate we got the newest one

tgravescs: overall I'd like to see more specific tests for this to make sure we are getting oldest and newest, also should have some without other filters in it to make sure it works standalone.

tools/README.md
nartal1: Yes, I also don't think many people are using this config yet. I am fine changing if you think it makes more sense.
"100-newest" -  per application based on app time.
"100-newest-overall" -  based on app time.
"100-newest-filesystem" -  based on filesystem time.

willb: I agree that filtering based on filesystem metadata isn’t likely to be particularly useful (and `find(1)` will do that in any case, so it doesn’t necessarily need to be built into a tool).

In the prototype, we had `--limit X` (to choose the newest `X` logs) and `--per-app-limit X` (to choose at most the `X` newest logs per application name).  We also had a `--oldest` flag to override the ordering.  I thought this was a pretty sensible UX and in general prefer multiple options to configure multiple aspects of functionality instead of having a string to parse as an argument.

For time-based filtering, there are a couple of interesting cases:  

- since a certain (human-readable) time
- between a range of human-readable times
- in the last `k` days (or hours / weeks / months)

I think that a reasonable command-line UX for these would look like this:

- `--since 7/1/2021`
- `--since 7/1/2021 --before 7/15/2021`
- `--in-last 5d`

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
nartal1: created another tests which verifies the same by comparing the result appNames with expected appNames.

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
nartal1: I have added the tests without other filters. PTAL.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
nartal1: Thanks! It was unintentional. I had missed updating this.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala

nartal1:
build

tools/README.md
nartal1: Thanks @willb . 
We do have flag for : the last `k` days(mins/hours/weeks/months)  - start-app-time . The input is not the date format though. It is :  --start-app-time 2d (past 2days), 3w(past 3 weeks) and so on.

tools/README.md
```diff
@@ -196,12 +196,29 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
...
-                                    processing oldest 100 event logs)
+  -f, --filter-criteria  <arg>      Filter newest or oldest N eventlogs based on filesystem
+                                    timestamp, application start timestamp or unique application
+                                    name. eg: 100-newest-filesystem (for processing newest 100 event
+                                    logs based on filesystem timestamp). eg: 100-oldest-filesystem
+                                    (for processing oldest 100 event logs  based on filesystem
+                                    timestamp). Filesystem based filtering happens before any
+                                    application based filtering.
+                                    For application based filtering, the order in which filters are
+                                    applied is: application-name, start-app-time, filter-criteria
```
tgravescs, line 207: nit, add .

tools/README.md
```diff
@@ -196,12 +196,29 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
                                     of a selection criterion. i.e Select all
                                     event logs except the ones which have
                                     application name as the input string.
-  -f, --filter-criteria  <arg>      Filter newest or oldest N eventlogs for
-                                    processing.eg: 100-newest (for processing
-                                    newest 100 event logs). eg: 100-oldest (for
-                                    processing oldest 100 event logs)
+  -f, --filter-criteria  <arg>      Filter newest or oldest N eventlogs based on filesystem
```
tgravescs, line 199: lets change the order and put filesystem information last

tools/src/main/scala/com/nvidia/spark/rapids/tool/qualification/QualificationArgs.scala
```diff
@@ -97,8 +108,10 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
   }
 
   validate(filterCriteria) {
-    case crit if (crit.endsWith("-newest") || crit.endsWith("-oldest")) => Right(Unit)
-    case _ => Left("Error, the filter criteria must end with either -newest or -oldest")
+    case crit if (crit.endsWith("-filesystem") || crit.endsWith("-oldest")
```
tgravescs, line 111: we should verify the entire string newest-filesystem, oldest-per-app-name, etc, otherwise we need to make sure its correct later.

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -272,34 +272,64 @@ class AppFilterSuite extends FunSuite {
...
+    testFileSystemTimeAndStartAndAppFull(appsFullWithFsToTest, "10-oldest-filesystem",
+      "nds86", "app-nds86", "3w", 2)
+  }
+
+  test("full app name exact and 2-oldest by app time no match from filename") {
+    testFileSystemTimeAndStartAndAppFull(appsFullWithFsToTest, "2-oldest",
+      "nds", "nomatch", "3w", 0)
+  }
+
+  test("full app name exact and 2-oldest by app time") {
```
tgravescs, line 310: nit - name of test is still wrong here, we aren't specifying the full app name

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -339,4 +369,64 @@ class AppFilterSuite extends FunSuite {
...
+    val expected = Array("nds86", "nds86", "ndsweeks2")
+    testAppStartTimeAndPerAppName(appsFullWithFsToTest, "3-oldest", 3, expected)
+  }
+
+  test("standalone 10-oldest based on app time") {
+    val expected = Array("ndshours18", "nds86", "nds86", "ndsweeks2")
+    testAppStartTimeAndPerAppName(appsFullWithFsToTest, "10-oldest", 4, expected)
+  }
+
+  private def testAppStartTimeAndPerAppName(
```
tgravescs, line 398: this isn't testing app start time (--start-app-time) lets update name

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -339,4 +369,64 @@ class AppFilterSuite extends FunSuite {
       }
     }
   }
+
+  test("standalone 1-oldest-per-app-name") {
+    val expected = Array("ndshours18", "ndsweeks2", "nds86")
+    testAppStartTimeAndPerAppName(appsFullWithFsToTest, "1-oldest-per-app-name", 3, expected)
```
tgravescs, line 375: I'd prefer to see a list of apps. with more then just one that has 2 of the same app name.  one should have > 2 and another should have at least 2.

Checking expected is nds86 doesn't test oldest or newest. the appid should have unique name: local-16261043003${app.uniqueId} so lets check that to make sure we got the right nds86 app

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -272,34 +272,64 @@ class AppFilterSuite extends FunSuite {
...
-      "3w", 0)
+  test("full 2-oldest-filesystem no match from app name") {
+    testFileSystemTimeAndStartAndAppFull(appsFullWithFsToTest, "2-oldest-filesystem",
+      "nomatch", "nds", "3w", 0)
   }
 
-  test("full app name exact and 10-oldest and 3w") {
-    testFileSystemTimeAndStartAndAppFull(appsFullWithFsToTest, "10-oldest", "nds86", "app-nds86",
-      "3w", 2)
+  test("full app name exact and 10-oldest-filesystem and 3w") {
```
tgravescs, line 300: we should also add negative tests for invalid parameters names passed in

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
nartal1: Updated the tests where we compare both appId and appName. So we can be certain the oldest/newest is the right one as expected.

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
nartal1: Updated it. Not sure if it is still confusing :D

tools/src/main/scala/com/nvidia/spark/rapids/tool/qualification/QualificationArgs.scala
nartal1: done

tools/README.md
tgravescs: don't see that this got done?

nartal1:
build

tgravescs (commented):
I think only minor thing would be missed the comment about changing the help/usage output to put the filesystem based options last since we think they will be used least

tools/README.md
nartal1: Sorry I had updated the order in this line but hadn't changed the order below in the examples section. Done now.

tgravescs (approved):

tgravescs:
build

PR #2948, opened 2021-07-16
Remove unnecessary copies of ParquetCachedBatchSerializer

### LINKED ISSUE #2905: Remove Spark version specific Serializers unless they need to override behavior
Right now we have ParquetCachedBatchSerializer for every version of Spark since 3.1.1 e.g 312 311db and 311cdh. We need to come up with a way to avoid this unnecessary class.

razajafri:
We decided to remove the extra serializers and have a single version of it in the 311 shims. The package name will be changed to better suit whatever strategy we come up as a part of supporting Spark 3.2 

Fixes #2905 


razajafri:
build

jlowe (commented):
Looks like a spark320 version was missed?

razajafri:
> Looks like a spark320 version was missed?

Sorry, I left it out for the draft, but forgot to remove it post our meeting

razajafri:
build

jlowe (approved):

jlowe (changes requested):
Actually the documentation does need to be changed here.  The cache docs refer to a table showing different classes for each Spark version, and those are being deleted in this PR.

jlowe (approved):

razajafri:
build

jlowe (changes requested):
It looks like jenkins/spark-tests.sh also needs to be updated since it tries to pick the shim-specific class.

razajafri:
build

jenkins/spark-tests.sh
```diff
@@ -104,7 +104,7 @@ spark-submit $BASE_SPARK_SUBMIT_ARGS $CUDF_UDF_TEST_ARGS --jars $RAPIDS_TEST_JAR
 #only run cache tests with our serializer in nightly test for Spark version >= 3.1.1
 if [[ "$IS_SPARK_311_OR_LATER" -eq "1" ]]; then
   SHIM_PACKAGE=$(echo ${SPARK_VER} | sed 's/\.//g' | sed 's/-SNAPSHOT//')
```
jlowe, line 106: This is no longer needed.
```suggestion
```

razajafri:
build

jlowe (approved):

PR #2950, opened 2021-07-16
Fix shutdown bugs in the RAPIDS Shuffle Manager

abellina:

While debugging another issue, I noticed these were causing NPEs in the logs.

- the `stop()` function in the shuffle manager can be called multiple times, so I am protecting against calling `.close()` multiple times in things like the bounce buffers
- the worker and context can be null at various times during shut down, protecting against that too and fixing the order of shutdown in UCX.scala so it doesn't happen.

### DIFF
--- a/shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
+++ b/shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
@@ -626,13 +626,21 @@ class UCX(transport: UCXShuffleTransport, executor: BlockManagerId, rapidsConf: 
     endpointManager.getConnection(peerExecutorId, peerMgmtHost, peerMgmtPort)
   }
 
-  def onWorkerThreadAsync(task: () => Unit): Unit = {
-    workerTasks.add(task)
+  private def signalWorker(): Unit = {
     if (rapidsConf.shuffleUcxUseWakeup) {
       withResource(new NvtxRange("UCX Signal", NvtxColor.RED)) { _ =>
-        worker.signal()
+        try {
+          worker.signal()
+        } catch {
+          case e: Throwable if isShuttingDown => // ignore these
+        }
       }
     }
+  }
+
+  def onWorkerThreadAsync(task: () => Unit): Unit = {
+    workerTasks.add(task)
+    signalWorker()
   }
 
   /**
@@ -695,6 +703,7 @@ class UCX(transport: UCXShuffleTransport, executor: BlockManagerId, rapidsConf: 
 
   override def close(): Unit = {
     onWorkerThreadAsync(() => {
+      endpointManager.close()
       amRegistrations.forEach { (activeMessageId, _) =>
         logDebug(s"Removing Active Message registration for " +
           s"${TransportUtils.toHex(activeMessageId)}")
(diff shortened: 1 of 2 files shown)


shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -626,15 +626,23 @@ class UCX(transport: UCXShuffleTransport, executor: BlockManagerId, rapidsConf:
...
-  def onWorkerThreadAsync(task: () => Unit): Unit = {
-    workerTasks.add(task)
+  private def signalWorker(): Unit = {
     if (rapidsConf.shuffleUcxUseWakeup) {
       withResource(new NvtxRange("UCX Signal", NvtxColor.RED)) { _ =>
-        worker.signal()
+        try {
+          worker.signal()
+        } catch {
+          case e: Throwable if isShuttingDown => // ignore these
```
jlowe, line 635: Normally I would expect an NPE to be fixed by adding some check somewhere to make sure we don't deference something that can be null.  I'm normally not a fan of fixing NPE's by catch-alls and ignoring, as it can submarine all sorts of other issues accidentally.

What was the cause of the NPE?  Is it fixed or are we just catching and ignoring as a "fix"?

abellina: The NPEs are caused by double-closing in UCX. We close once, and a nativeId `Long` is set to `null`, but then if there's a second call to close, the `null` is normally sent to a JNI call which expects `long` primitives. This is producing a `NullPointerException`, as java can't convert the `Long` null to `long`.

So that's what I am trying to fix, but you are right it could be done better. Given that `signal` can be called from other threads (literally anyone trying to queue up a send or a receive ends up there) and `close` is called from the progress thread when exiting, we really need a lock between the `close` and `signal` functions for `worker`.

abellina: @jlowe, pushed this (https://github.com/NVIDIA/spark-rapids/pull/2950/commits/e921d038adf86eaf4e666b2647ae8472bd3b9739) to handle the above.

shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala
```diff
@@ -190,8 +192,11 @@ class UCX(transport: UCXShuffleTransport, executor: BlockManagerId, rapidsConf:
         }
       }
 
-      logDebug("Exiting UCX progress thread.")
-      Seq(endpointManager, worker, context).safeClose()
+      worker.synchronized {
+        logDebug("Exiting UCX progress thread.")
+        Seq(endpointManager, worker, context).safeClose()
+        worker = null
```
jlowe, line 198: I see we're synchronizing up above implying multiple threads could be racing through here. This smashes `worker` to null yet the above synchronize block isn't checking for a null worker before trying to use it, and many other places aren't checking or expecting a null `worker` before they try to use it.

If we really want to go down this route, `worker` would need to be `volatile` and code would need to double-check after acquiring the lock that it's still not null.

jlowe (approved):

jlowe:
build

pxLi:
build

PR #2952, opened 2021-07-16
Query fuzzer manual test

andygrove:
This PR adds a manual test that generates random queries against random in-memory DataFrames and executes them on CPU and GPU and fails if there are any differences. The Spark configuration is also randomly generated.

This is a manual test because it runs in a loop until it finds a difference.

It is consistently failing fairly quickly at the moment due to known bugs.


tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzer.scala
```diff
@@ -0,0 +1,240 @@
...
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.sql.functions.col
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.types.{DataTypes, DecimalType}
+
+class QueryFuzzer(val seed: Long) {
+
+  val rand = new Random(seed)
+
+  val idGen = new AtomicLong(0)
```
jlowe, line 32: Curious on the choice to use an `AtomicLong` here.  If we have threads racing through this code, how can we get a deterministic plan for a given seed?  If we don't have racing threads, why do we need `AtomicLong`?

tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzer.scala
```diff
@@ -0,0 +1,240 @@
...
+    value(rand.nextInt(value.length))
+  }
+
+  def randomOperator(spark: SparkSession, depth: Int, maxDepth: Int): DataFrame = {
+
+    // if we have reached maximum depth then create a leaf node
+    if (depth == maxDepth) {
+      return generateDataSource(spark)
+
+      //TODO Add support for reading from CSV and Parquet data sources
```
jlowe, line 70: Not sure this comment needs to be here.  It's an implementation detail and arguably should be something `generateDataSource` should do.  Also farming from a single, static file is probably not very 'fuzzy', may be better to emit fuzzed values to a temporary Parquet or CSV file and feed that as input.

tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzer.scala
```diff
@@ -0,0 +1,240 @@
...
+      val tf = randomElement(validTransformations)
+      // apply the transformation
+      tf.transform(df)
+    }
+  }
+
+  /**
+   * Generate a random in-memory DataFrame.
+   */
+  def generateDataSource(spark: SparkSession) = {
```
jlowe, line 107:
```suggestion
  def generateDataSource(spark: SparkSession): DataFrame = {
```

tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzer.scala
```diff
@@ -0,0 +1,240 @@
...
+      DataTypes.StringType,
+      new DecimalType(10,2),
+      DataTypes.DateType,
+      DataTypes.TimestampType,
+      DataTypes.BooleanType,
+      DataTypes.DoubleType,
+      DataTypes.FloatType)
+    val dataTypes = (0 until numFields).map(_ => randomElement(supportedTypes))
+    val schema = FuzzerUtils.createSchema(dataTypes)
+    val options = FuzzerOptions(validStringChars = Some(" \t\r\n0123456789.+-/:aidfnT"),
```
jlowe, line 119: Curious why this tests such a limited character set -- no multi-byte Unicode characters for example.

tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzer.scala
```diff
@@ -0,0 +1,240 @@
...
+      maxStringLen = 12)
+    val df = FuzzerUtils.generateDataFrame(spark, schema, rowCount = 256, options, seed)
+      .repartition(1 + rand.nextInt(4))
```
jlowe, line 122: Why force a repartition here?  Repartition is already an operation that can randomly appear in the plan, so curious why we don't let the RNG decide if we should or should not repartition the data.  Otherwise seems like we're limiting the testing to dataframes that always are repartitioned before further processing.

tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzer.scala
```diff
@@ -0,0 +1,240 @@
...
+
+case class Filter(ctx: FuzzContext) extends Transformation {
+
+  override def canTransform(df: DataFrame): Boolean = {
+    df.schema.fields.exists(_.dataType == DataTypes.IntegerType)
+  }
+
+  override def transform(df: DataFrame): DataFrame = {
+    val intCols = df.schema.fields.filter(_.dataType == DataTypes.IntegerType)
+    df.filter(col(intCols(0).name).gt(col(intCols(1).name)))
```
jlowe, line 166: I assume the plan is to make this a bit more sophisticated in the future, with RNG-selected operations, columns, and column ordering so it's not always lower-indexed columns being compared against higher-indexed columns.

tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzer.scala
```diff
@@ -0,0 +1,240 @@
...
+  }
+}
+
+case class SortRandomColumns(ctx: FuzzContext) extends Transformation {
+
+  override def canTransform(df: DataFrame): Boolean = df.schema.fields.nonEmpty
+
+  override def transform(df: DataFrame): DataFrame = {
+    val numSortColumns = 1 + ctx.fuzzer.rand.nextInt(2)
```
jlowe, line 175: Why hardcode 2 instead of the number of columns in the dataframe?

tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzer.scala
```diff
@@ -0,0 +1,240 @@
...
+    val sortColumns = (0 until numSortColumns)
+      .map(_ => df.columns(ctx.fuzzer.rand.nextInt(df.columns.length)))
```
jlowe, line 177: I expected this to be a random shuffle of the dataframe columns and select the first N where N is a random between 0 and column count.

tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzer.scala
```diff
@@ -0,0 +1,240 @@
...
+
+  override def canTransform(df: DataFrame): Boolean = true
+
+  override def transform(df: DataFrame): DataFrame = {
+    df.repartition(1 + ctx.fuzzer.rand.nextInt(4),
+      col(df.columns(ctx.fuzzer.rand.nextInt(df.columns.length))))
+  }
+}
+
+case class Join(ctx: FuzzContext) extends Transformation {
```
jlowe, line 223: Not referenced.

tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzerSuite.scala
```diff
@@ -0,0 +1,118 @@
...
+
+    (tryCpu, tryGpu) match {
+      case (Success(cpu), Success(gpu)) =>
+        compareResults(sort = true, maxFloatDiff = 0.0001, cpu, gpu)
+      case (Success(_), Failure(gpu)) =>
+        fail(s"CPU run succeeded. GPU run failed.", gpu)
+      case (Failure(cpu), Success(_)) =>
+        fail(s"GPU run succeeded. CPU run failed.", cpu)
+      case (Failure(cpu), Failure(gpu)) =>
+        println(s"Query failed both on CPU and GPU:\nCPU: $cpu\nGPU: $gpu")
```
jlowe, line 65: I would expect this to show the seed so it's easy to grep the important info on one line.

tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzer.scala
andygrove: I was doing this to force these DataFrames onto the GPU early, but this is a good point. I will remove this.

tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzer.scala
andygrove: Yes, all of this is just a slightly cleaned-up version of the demo that I hacked together and is just a starting point.

tests/src/test/scala/com/nvidia/spark/rapids/QueryFuzzer.scala
andygrove: This was really just for convenience. I can implement a generator method in the class instead.

PR #2953, opened 2021-07-16
Add in support for rank and dense_rank

revans2:
This adds in support for the rank and dense_rank window operations.  This end up being a rather large change because it needed to refactor several parts of the windowing code to let rank work one batch at a time on queries.  I think those changes are good overall, because with them we should be able to get average working as a batched rolling window function as well, but I held off on doing that because this is already big enough.

I have not tested this with nested types like structs in the order by column.  In some cases it should be able to work, but I don't know if it will work with structs in structs. @rwlee are there limitations on this?

This depends on https://github.com/rapidsai/cudf/pull/8652

I did also modify some of the window tests for array lead/lag and such.  I removed some of the batching size options because they were taking a very long time to run and not adding more coverage. lead/lag do not have special batching fixup code so the batch size is not that important, but I left them as the smaller batch size so that at least the split batch by key code is being tested.
This helps with the performance, especially when I am adding in a number of new tests.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
@@ -750,11 +750,7 @@ class GroupedAggregations extends Arm {
                     // because for windows that only happens if the first values are nulls. So we
                     // will then call replace nulls as needed to fix that up. Typically the
                     // replacement policy is preceding.
-                    if (replacePolicy.isDefined) {
-                      scanned.replaceNulls(replacePolicy.get)
-                    } else {
-                      scanned.incRefCount()
-                    }
+                    replacePolicy.map(scanned.replaceNulls).getOrElse(scanned.incRefCount())
                 }
           }
           func.scanCombine(isRunningBatched, replacedCols)
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
@@ -21,7 +21,8 @@ import java.util.concurrent.TimeUnit
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 
-import ai.rapids.cudf.{Aggregation, AggregationOnColumn, AggregationOverWindow, DType, GroupByOptions, NullPolicy, NvtxColor, ReplacePolicy, ReplacePolicyWithColumn, Scalar, ScanType, Table, WindowOptions}
+import ai.rapids.cudf
+import ai.rapids.cudf.{Aggregation, AggregationOverWindow, DType, GroupByOptions, NullPolicy, NvtxColor, ReplacePolicy, ReplacePolicyWithColumn, Scalar, ScanType, Table, WindowOptions}
 
 import org.apache.spark.TaskContext
 import org.apache.spark.internal.Logging
@@ -32,7 +33,7 @@ import org.apache.spark.sql.catalyst.plans.physical.{AllTuples, ClusteredDistrib
 import org.apache.spark.sql.execution.{SparkPlan, UnaryExecNode}
 import org.apache.spark.sql.execution.window.WindowExec
 import org.apache.spark.sql.rapids.GpuAggregateExpression
-import org.apache.spark.sql.types.{ArrayType, BooleanType, ByteType, CalendarIntervalType, DataType, IntegerType, LongType, ShortType, StructType}
+import org.apache.spark.sql.types.{ArrayType, ByteType, CalendarIntervalType, DataType, IntegerType, LongType, ShortType, StructType}
 import org.apache.spark.sql.vectorized.{ColumnarBatch, ColumnVector}
 import org.apache.spark.unsafe.types.CalendarInterval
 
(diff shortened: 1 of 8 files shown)


revans2:
I should add that even if the underlying scan/group by scan support more nesting to the structs the plugin code would not work in those cases. I can make it work, but it would require a follow on issue to make it work.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -715,131 +716,254 @@ class GroupedAggregations extends Arm {
...
+                    // For scans when nulls are excluded then each input row that has a null in it
+                    // the output row also has a null in it. Typically this is not what we want,
+                    // because for windows that only happens if the first values are nulls. So we
+                    // will then call replace nulls as needed to fix that up. Typically the
+                    // replacement policy is preceding.
+                    if (replacePolicy.isDefined) {
+                      scanned.replaceNulls(replacePolicy.get)
+                    } else {
+                      scanned.incRefCount()
+                    }
```
jlowe, line 757: Can simplify this with something like:
```suggestion
                    replacePolicy.map(scanned.replaceNulls).getOrElse(scanned.incRefCount())
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -784,121 +853,446 @@ class SumBinaryFixer extends BatchedRunningWindowFixer with Arm with Logging {
...
+    rowNumFixer.fixUp(samePartitionMask, sameOrderMask, initialRowNum).close()
+    logDebug(s"rank: updateState from $previousRank to...")
+    previousRank.foreach(_.close)
+    previousRank = Some(ret.getScalarElement(ret.getRowCount.toInt - 1))
+    logDebug(s"rank/row: ... $previousRank $previousRow")
+    ret
+  }
+
+  override def close(): Unit = {
+    previousRank.foreach(_.close())
```
jlowe, line 1019: Should this set `previousRank = None` afterwards?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -784,121 +853,446 @@ class SumBinaryFixer extends BatchedRunningWindowFixer with Arm with Logging {
...
+    logDebug(s"dense rank: updateState from $previousRank to...")
+    previousRank.foreach(_.close)
+    previousRank = Some(ret.getScalarElement(ret.getRowCount.toInt - 1))
+    logDebug(s"dense rank: ... $previousRank")
+
+    ret
+  }
+
+  override def close(): Unit = {
+    previousRank.foreach(_.close())
```
jlowe, line 1115: Similar should this be updated to None comment here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -784,121 +853,446 @@ class SumBinaryFixer extends BatchedRunningWindowFixer with Arm with Logging {
...
+ * a running batch, then we drop the row number part because it is just not needed.
+ * @param children the order by columns.
+ */
+case class GpuRank(children: Seq[Expression]) extends GpuRunningWindowFunction
+    with GpuBatchedRunningWindowWithFixer {
   override def nullable: Boolean = false
   override def dataType: DataType = IntegerType
 
-  override def children: Seq[Expression] = Nil
+  // RUNNING WINDOW ONLY - RANK DOES NOT WORK FOR ANYTHING ELSE IN SPARK
```
jlowe, line 1172: Didn't expect an all-caps comment here which looks almost like a TODO.  Should this be a `@note` in the class description?

This applies to a couple of other all-caps comments below as well.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -784,121 +853,446 @@ class SumBinaryFixer extends BatchedRunningWindowFixer with Arm with Logging {
...
+
+  override def groupByScanInputProjection(isRunningBatched: Boolean): Seq[Expression] = {
+    // The requirement is that each input projection has a corresponding aggregation
+    // associated with it. This also fits with how rank works in cudf, where the input is also
+    // a single column. If there are multiple order by columns we wrap them in a struct.
+    // This is not ideal from a memory standpoint, and in the future we might be able to fix this
+    // with a ColumnView, but for now with how the java cudf APIs work it would be hard to work
```
jlowe, line 1179: Nit:
```suggestion
    // with a ColumnView, but for now with how the Java cudf APIs work it would be hard to work
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -784,121 +853,446 @@ class SumBinaryFixer extends BatchedRunningWindowFixer with Arm with Logging {
...
+  override def dataType: DataType = IntegerType
+
+  // RUNNING WINDOW ONLY - DENSE RANK DOES NOT WORK FOR ANYTHING ELSE IN SPARK
+
+  override def groupByScanInputProjection(isRunningBatched: Boolean): Seq[Expression] = {
+    // The requirement is that each input projection has a corresponding aggregation
+    // associated with it. This also fits with how rank works in cudf, where the input is also
+    // a single column. If there are multiple order by columns we wrap them in a struct.
+    // This is not ideal from a memory standpoint, and in the future we might be able to fix this
+    // with a ColumnView, but for now with how the java cudf APIs work it would be hard to work
```
jlowe, line 1245: Nit:
```suggestion
    // with a ColumnView, but for now with how the Java cudf APIs work it would be hard to work
```

jlowe (commented):
Looks great overall, just some minor nits and questions.  I'm definitely not an expert on the windowing code, so it'd be nice to have someone more familiar look as well.

jlowe (approved):

rwlee:
https://github.com/rapidsai/cudf/pull/8652 is merged into the 21.08 release

jlowe:
build

Issue #2957, opened 2021-07-16
[BUG] Builds failing due to breaking changes in SPARK-36034


andygrove:
**Describe the bug**

There is a [breaking change](https://github.com/apache/spark/commit/ba7117224c797e3ade64a281b1a165cf5040c541) in Spark branch-3.1 

```
java.lang.NoSuchMethodError: org.apache.spark.sql.execution.datasources.parquet.ParquetFilters.<init>(Lorg/apache/parquet/schema/MessageType;ZZZZIZ)V
[2021-07-16T20:40:17.896Z] 	at com.nvidia.spark.rapids.GpuParquetFileFilterHandler.filterBlocks(GpuParquetScan.scala:269)
[2021-07-16T20:40:17.896Z] 	at com.nvidia.spark.rapids.GpuParquetMultiFilePartitionReaderFactory.$anonfun$buildBaseColumnarReaderForCoalescing$1(GpuParquetScan.scala:365)
```

**Steps/Code to reproduce bug**
We are seeing this in CI builds

pxLi:
always reproducible w/ `mvn test -Pspark313tests,snapshot-shims` use the 3.1.X-SNAPSHOT shim test

pxLi:
https://github.com/apache/spark/commit/e41ecef61b3f40b51d2d63056c826ef35b920791 also pushed to 3.0 branch, so it failed our 3.0.4-snapshot test, going to add shim to 304 as a following-up of https://github.com/NVIDIA/spark-rapids/pull/2961

PR #2961, opened 2021-07-19
Fix ParquetFilters issue

wbo4958:
Spark has added datetime rebase mode into the constructor of ParquetFilters,
see SPARK-36034, which causes the build failed.

This change is adding getParquetFilters in the shim layers to match
different Spark version.

Here is an issue to track this, see https://github.com/NVIDIA/spark-rapids/issues/2957

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
@@ -22,6 +22,7 @@ import java.nio.ByteBuffer
 import org.apache.arrow.memory.ReferenceManager
 import org.apache.arrow.vector.ValueVector
 import org.apache.hadoop.fs.Path
+import org.apache.parquet.schema.MessageType
 
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.{SparkSession, SparkSessionExtensions}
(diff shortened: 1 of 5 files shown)


wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
```diff
@@ -22,14 +22,15 @@ import java.nio.ByteBuffer
...
+import org.apache.parquet.schema.MessageType
 
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.{SparkSession, SparkSessionExtensions}
 import org.apache.spark.sql.catalyst.{InternalRow, TableIdentifier}
 import org.apache.spark.sql.catalyst.analysis.Resolver
 import org.apache.spark.sql.catalyst.catalog.{CatalogTable, SessionCatalog}
 import org.apache.spark.sql.catalyst.encoders.ExpressionEncoder
-import org.apache.spark.sql.catalyst.expressions.{Alias, Expression, ExprId, NullOrdering, SortDirection, SortOrder}
+import org.apache.spark.sql.catalyst.expressions.{Alias, ExprId, Expression, NullOrdering, SortDirection, SortOrder}
```
pxLi, line 33: Thanks for looking into this!

scalastyle error from CI
error file=/home/jenkins/agent/workspace/jenkins-rapids_premerge-github-2118/sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala message=Expression should come before ExprId. line=33 column=49

pxLi:
also the change only affect 3.1.3+ (https://github.com/apache/spark/commit/ba7117224c797e3ade64a281b1a165cf5040c541), could just add the change to 313shim+ only

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
wbo4958: Thx, Done

wbo4958:
I checked spark 3.2-SNAPSHOT has contained that PR, so it should affect 3.2

pxLi:
> I checked spark 3.2-SNAPSHOT has contained that PR, so it should affect 3.2

My bad, 320shims actually extends Spark311Shims. This one looks good~

firestarman:
LGTM

pxLi (approved):

pxLi:
LGTM, going to merge this to unblock other CI runs. If someone have more comments we can have following PRs to address

Issue #2959, opened 2021-07-18
[BUG] Resolve common code source incompatibility with supported Spark versions


gerashegalov:
**Describe the bug**
The default build configuration builds unshimmed code such as the `sql-plugin` module [against Spark 3.0.1](https://github.com/NVIDIA/spark-rapids/blob/9998174a389376237e7d92a7099a3b4de5844ca2/pom.xml#L234). Thus, the source-incompatibility when it arises in the later versions is only caught at runtime. While we are working on reworking the build of the shims to eliminate this issue we can manually reproduce and fix these issues in Shims even using the current build. 

**Steps/Code to reproduce bug**
Iterate the list of supported Spark versions and run
```
mvn  -DskipTests -Dspark.version=3.x.y -Psparkxy  clean verify
```

*Spark 3.1.2*
```bash
mvn clean package -Dspark.version=3.1.2 -Pspark312
[ERROR] [Error] /home/gshegalov/gits/NVIDIA/spark-rapids/sql-plugin/src/main/scala/com/nvidia/spark/rapids/DateUtils.scala:24: value SQLDate is not a member of object org.apache.spark.sql.catalyst.util.DateTimeUtils
[ERROR] [Error] /home/gshegalov/gits/NVIDIA/spark-rapids/sql-plugin/src/main/scala/com/nvidia/spark/rapids/DateUtils.scala:95: not found: type SQLDate
[WARNING] [Warn] /home/gshegalov/gits/NVIDIA/spark-rapids/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala:234: value ENABLE_JOB_SUMMARY in class ParquetOutputFormat is deprecated: see corresponding Javadoc for more information.
[ERROR] [Error] /home/gshegalov/gits/NVIDIA/spark-rapids/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala:537: wrong number of arguments for pattern org.apache.spark.sql.catalyst.expressions.UnaryMinus(child: org.apache.spark.sql.catalyst.expressions.Expression,failOnError: Boolean)
[...]

PR #2967, opened 2021-07-19
Resolve hidden source incompatibility between Spark30x and Spark31x  Shims

gerashegalov:
This PR addresses the compile errors encountered during build with
```bash
mvn clean compile -Dspark.version=3.1.1 
```

Contributes to #2959

### DIFF
--- a/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
+++ b/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
@@ -51,7 +51,7 @@ import org.apache.spark.sql.execution.joins.{BroadcastHashJoinExec, BroadcastNes
 import org.apache.spark.sql.execution.python.{AggregateInPandasExec, ArrowEvalPythonExec, FlatMapGroupsInPandasExec, MapInPandasExec, WindowInPandasExec}
 import org.apache.spark.sql.execution.window.WindowExecBase
 import org.apache.spark.sql.internal.SQLConf
-import org.apache.spark.sql.rapids.{GpuFileSourceScanExec, GpuStringReplace, GpuTimeSub, ShuffleManagerShimBase}
+import org.apache.spark.sql.rapids.{GpuFileSourceScanExec, GpuStringReplace, GpuTimeSub, GpuUnaryMinus, ShuffleManagerShimBase}
 import org.apache.spark.sql.rapids.execution.{GpuBroadcastExchangeExecBase, GpuBroadcastNestedLoopJoinExecBase, GpuShuffleExchangeExecBase}
 import org.apache.spark.sql.rapids.execution.python.{GpuAggregateInPandasExecMeta, GpuArrowEvalPythonExec, GpuFlatMapGroupsInPandasExecMeta, GpuMapInPandasExecMeta, GpuPythonUDF, GpuWindowInPandasExecMetaBase}
 import org.apache.spark.sql.rapids.shims.spark301.{GpuSchemaUtils, ShuffleManagerShim}
@@ -652,4 +652,11 @@ abstract class SparkBaseShims extends SparkShims {
   override def hasAliasQuoteFix: Boolean = false
 
   override def hasCastFloatTimestampUpcast: Boolean = false
+
+  override def toGpuExpression(exp: Expression): Expression = {
+    exp match {
+      case UnaryMinus(child) => GpuUnaryMinus(child)
+      case _ => exp
+    }
+  }
 }
--- a/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
+++ b/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
@@ -51,7 +51,7 @@ import org.apache.spark.sql.execution.joins.{BroadcastHashJoinExec, BroadcastNes
 import org.apache.spark.sql.execution.python.{AggregateInPandasExec, ArrowEvalPythonExec, FlatMapGroupsInPandasExec, MapInPandasExec, WindowInPandasExec}
 import org.apache.spark.sql.execution.window.WindowExecBase
 import org.apache.spark.sql.internal.SQLConf
-import org.apache.spark.sql.rapids.{GpuFileSourceScanExec, GpuStringReplace, GpuTimeSub, GpuUnaryMinus, ShuffleManagerShimBase}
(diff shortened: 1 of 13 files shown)


gerashegalov:
build

shims/spark311/src/main/scala/org/apache/spark/sql/rapids/shims/spark311/ShuffleManagerShim.scala
```diff
@@ -33,4 +34,10 @@ class ShuffleManagerShim extends ShuffleManagerShimBase {
     shuffleManager.getReader(
       handle, startMapIndex, endMapIndex, startPartition, endPartition, context, metrics)
   }
+
+  override def toGpu(x: ShufflePartitionSpec): GpuShufflePartitionSpec = x match {
```
andygrove, line 38: nit: Can we change the argument name to something more meaningful, perhaps `spec` or `partitionSpec`?

andygrove (dismissed):
LGTM other than a minor nit

shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
```diff
@@ -652,4 +652,11 @@ abstract class SparkBaseShims extends SparkShims {
   override def hasAliasQuoteFix: Boolean = false
 
   override def hasCastFloatTimestampUpcast: Boolean = false
+
+  override def toGpuExpression(exp: Expression): Expression = {
```
jlowe, line 656: The name of this method is super confusing.  The name implies it will take an expression and convert it to a GPU expression, but then apparently only does it for the very specific `UnaryMinus` expression and even then doesn't recursively try to translate any of the child expressions.  Either the name is too generic or this doesn't do what we need it to do.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DateUtils.scala
```diff
@@ -21,12 +21,14 @@ import java.time.LocalDate
...
 
 import org.apache.spark.sql.catalyst.util.DateTimeUtils
-import org.apache.spark.sql.catalyst.util.DateTimeUtils.{localDateToDays, SQLDate}
+import org.apache.spark.sql.catalyst.util.DateTimeUtils.localDateToDays
 
 /**
  * Class for helper functions for Date
  */
 object DateUtils {
+  type RapidsSQLDate = Int
```
jlowe, line 30: Is there any point to this typedef?  It's never used except in this file as a return type.  Spark apparently just reverted to using Int directly, which makes me wonder why we wouldn't follow their lead.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/ShuffledBatchRDD.scala
```diff
@@ -175,22 +178,25 @@ class ShuffledBatchRDD(
...
-          context,
-          sqlMetricsReporter)
-        val blocksByAddress = shim.getMapSizesByExecutorId(
-          dependency.shuffleHandle.shuffleId, 0, Int.MaxValue, reducerIndex, reducerIndex + 1)
-        val partitionSize = blocksByAddress.flatMap(_._2)
-            .filter(tuple => tuple._3 >= startMapIndex && tuple._3 < endMapIndex)
-            .map(_._2).sum
-        (reader, partitionSize)
+      case prps: PartialReducerPartitionSpec =>
+        ShimLoader.getSparkShims.getShuffleManagerShims.toGpu(prps) match {
```
jlowe, line 182: Seems like we could avoid shimming this entirely (at least for now) by avoiding the `unapply` on `ParitalReducerPartitionSpec`.  Instead we could reference `prps.startMapIndex` etc. in the use below.

shims/spark301/src/main/scala/org/apache/spark/sql/rapids/shims/spark300/ShuffleManagerShim.scala
```diff
@@ -17,7 +17,8 @@ package org.apache.spark.sql.rapids.shims.spark301
```
jlowe, line 17: Happened to notice this file is in the spark301 package and shim but in a spark300 directory?!  Can clean this up in a separate PR.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExpression.scala
```diff
@@ -532,10 +532,13 @@ case class GpuSpecifiedWindowFrame(
...
 
-  private def boundarySql(expr: Expression): String = expr match {
-    case e: GpuSpecialFrameBoundary => e.sql
-    case UnaryMinus(n) => n.sql + " PRECEDING"
-    case e: Expression => e.sql + " FOLLOWING"
+  private def boundarySql(expr: Expression): String = {
+    val gpuExp = ShimLoader.getSparkShims.toGpuExpression(expr)
+    gpuExp match {
+      case e: GpuSpecialFrameBoundary => e.sql
+      case GpuUnaryMinus(n) => n.sql + " PRECEDING"
```
jlowe, line 539: We could avoid shimming this entirely (at least for now) by avoiding the `unapply` and doing something like this instead:
```suggestion
      case u: UnaryMinus => u.child.sql + " PRECEDING"
```

jlowe (changes requested):

shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
gerashegalov: I anticipated there will be more of this kind, and thought of using a single template for all of them instead of adding specific methods all the time

jlowe: this still doesn't do what we want.  The name implies it will convert an expression to a GPU expression, like what GpuOverrides does, but it does not do that at all.  IMO we shouldn't be shimming this in the first place.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DateUtils.scala
gerashegalov: I considered this as well, and favored staying close to what we had, but  reasoning for simply switching to Int makes sense to me as well.

gerashegalov:
Another push pending undoing markdown changes

gerashegalov:
build

jlowe (approved):

andygrove (approved):

PR #2971, opened 2021-07-20
Support GpuCollectList and GpuCollectSet as TypedImperativeAggregate

sperlingxx:

Current PR is to support GpuCollectList and GpuCollectSet as TypedImperativeAggregate, which is the task 3 of #2916. In this PR, we also introduce TypedImperativeAggExprMeta and  GpuNoHashAggregateMeta to provide a general support for TypedImperativeAggregate functions.

In addition, Aggregate stacks with TypedImperativeAggregate functions may lead to unexpected crash if the stack falls back to CPU partially, because GPU data types are inconsistent with CPU counterparts. This problem will be fixed in the task 4 of #2916. To avoid this kind of unexpected crash in current, we bring up the "associated fallback" mechanism in this PR, which only affects Aggregate plans containing TypedImperativeAggregate functions.

The "associated fallback" falls back all stages of an Aggregate (logical plan) to CPU once we need to fall back any stage of the plan. The "associated fallback" will be triggered on each final stage of Aggregate which contains TypedImperativeAggregate functions. It traverses the plan tree to collect all stages of current Aggregate (logical plan), and to determine whether to fallback them entirely or not.  In addition, the "associated fallback" also works when AQE is on.


sperlingxx:
build

sperlingxx:
build

jlowe:
#2916 was postponed to the 21.10 release.  Given we're in burndown, I think this should be retargeted to branch-21.10 when that is available (hopefully soon).

pxLi:
Yes, I am still waiting for cudf to get their CI ready for 21.10 (planned on Jul 22). After that I can start working on setup CICD for our plugin 21.10.0

sperlingxx:
build

sperlingxx:
> #2916 was postponed to the 21.10 release. Given we're in burndown, I think this should be retargeted to branch-21.10 when that is available (hopefully soon).

Re-targeted to the new branch.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -773,15 +773,25 @@ class GpuHashAggregateIterator(
...
-            // cast will be cheap if type matches, only does refCount++ in that case
-            closeOnExcept(result.getColumn(i).castTo(rapidsType)) { castedCol =>
-              resCols += GpuColumnVector.from(castedCol, dataTypes(i))
+          val resCols = mutable.ArrayBuffer.empty[ColumnVector]
+          closeOnExcept(resCols) { resCols =>
+            (0 until result.getNumberOfColumns).foldLeft(resCols) { case (ret, i) =>
+              val column = result.getColumn(i)
+              val rapidsType = GpuColumnVector.getRapidsType(dataTypes(i))
+              // extra type conversion check for nested types
+              if ((rapidsType.equals(DType.LIST) || rapidsType.equals(DType.STRUCT)) &&
```
abellina, line 782: Why not call `typeConversionAllowed` for all columns? (i.e. not need to special case LIST and STRUCT)?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -45,6 +45,12 @@ class GpuSortMeta(
     parent: Option[RapidsMeta[_, _, _]],
     rule: DataFromReplacementRule)
   extends SparkPlanMeta[SortExec](sort, conf, parent, rule) {
+
+  // Uses output attributes of child plan because SortExec will not change the attributes,
+  // and we need to propagate possible type conversions on the output attributes of
+  // GpuSortAggregateExec.
+  override protected val useOutputAttributesOfChild: Boolean = true
```
abellina, line 52: I am confused why this PR changes the behavior of `GpuSortMeta`. Can you elaborate?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuShuffleExchangeExec.scala
```diff
@@ -49,6 +49,10 @@ class GpuShuffleMeta(
   override val childParts: scala.Seq[PartMeta[_]] =
     Seq(GpuOverrides.wrapPart(shuffle.outputPartitioning, conf, Some(this)))
 
+  // propagate possible type conversions on the output attributes of map-side plans to
+  // reduce-side counterparts
+  override protected val useOutputAttributesOfChild: Boolean = true
```
abellina, line 54: I am also confused why this PR changes the behavior of GpuShuffleMeta. Can you elaborate? An example in the PR description would be nice.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -927,6 +937,179 @@ abstract class GpuBaseAggregateMeta[INPUT <: SparkPlan](
...
+        case ar: AttributeReference if desiredInputAggBufTypes.contains(ar.exprId) =>
+          retMeta.overrideDataType(desiredInputAggBufTypes(ar.exprId))
+        case _ =>
+      }
+    }
+  }
+}
+
+object GpuNoHashAggregateMeta {
+  private val logicalIdToMetas =
```
abellina, line 1078: Is this going to continue to grow as the application runs? It seems like so because this is a static class.

I think the main reason for this is so that a half aggregate isn't currently doable for the typed imperative case. Is that right? Did you try other ways of handling it?

abellina (changes requested):
@sperlingxx first pass through your changes.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
sperlingxx: For non-nested types, it may happen type casting here, such as casting INT to LONG. Therefore, the check of `typeConversionAllowed` may fail when the type conversion is necessary.  For nested types, no type conversion is necessary (available), which indicates the check is safe.  What's more, we didn't match any children types in `GpuColumnVector.getRapidsType`. So, we check whether they match or not via `typeConversionAllowed`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -927,6 +937,179 @@ abstract class GpuBaseAggregateMeta[INPUT <: SparkPlan](
   }
 }
 
+/**
+ * Base class for metadata around `SortAggregateExec` and `ObjectHashAggregateExec`, which may
+ * contain TypedImperativeAggregate functions in aggregate expressions.
+ */
+abstract class GpuNoHashAggregateMeta[INPUT <: SparkPlan](
```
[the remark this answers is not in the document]
sperlingxx: Because `GpuSortAggregateMeta` also inherits from it. When `spark.sql.execution.useObjectHashAggregateExec` is set to False, Spark catalyst will plan a `SortAggregateExec` instead of `ObjectAggregateExec` for Aggregate (logical plan) with TypedImperativeAggregate functions.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -927,6 +937,179 @@ abstract class GpuBaseAggregateMeta[INPUT <: SparkPlan](
...
+        conf.gpuTargetBatchSizeBytes)
+    } else {
+      super.convertToGpu()
+    }
+  }
+
+  /**
+   * The method replaces data types of aggregation buffers created by TypedImperativeAggregate
+   * functions with the actual data types used in the GPU runtime. And this method can adapt all
+   * Aggregation modes, including PartialMerge.
```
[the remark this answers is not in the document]
sperlingxx: I apologize for making you confused. The comment here is not clear enough. This is method only gets called for Partial and PartialMerge. I will remove this line.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -692,14 +693,35 @@ abstract class SparkPlanMeta[INPUT <: SparkPlan](plan: INPUT,
...
+      if (wrapped.children.length != 1) {
+        throw new IllegalArgumentException("useOutputAttributesOfChild ONLY works on UnaryPlan")
+      }
+      // We will check whether the child plan can be replaced or not. We only pass through the
+      // outputAttributes of child plan when it is GPU enabled. Otherwise, we should fetch the
+      // outputAttributes from wrapped plan, because type overriding of RapidsMeta is specialized
+      // for the GPU runtime.
+      //
+      // We can safely call childPlan.canThisBeReplaced here, because outputAttributes is called
+      // when tagSelfForGpu. And the tagging of child plan has been done in the moment.
```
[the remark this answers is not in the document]
sperlingxx: It is right. Thanks for correcting my writings!

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -692,14 +693,35 @@ abstract class SparkPlanMeta[INPUT <: SparkPlan](plan: INPUT,
         case (ar, _) =>
           ar
       }
+    case None if useOutputAttributesOfChild =>
+      if (wrapped.children.length != 1) {
```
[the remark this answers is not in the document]
sperlingxx: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -573,6 +573,13 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(true)
 
+  val ENABLE_TYPED_IMPERATIVE_AGGREGATE = conf("spark.rapids.sql.typedImperativeAggregate.enabled")
```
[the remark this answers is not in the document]
sperlingxx: This setting was added before the introduction of "associated fallback". With this setting, we can disable TypedImperativeAggregate by default, to get rid of unexpected crash with default configuration.
Perhaps we can remove this option, since we rewrite all partial-accelerated cases in current.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
sperlingxx: Yes, here is a potential risk. And I reworked this part with a different approach.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuShuffleExchangeExec.scala
sperlingxx: In most situations, at least one `ShuffleExchangeExec`  will be inserted for each Aggregate logical plan. ShuffleExchange for aggregate takes the internal aggregation buffers along with grouping keys. As we override the data type of aggregation buffer created by `TypedImperativeAggregate` functions,  we also need to apply the same overriding on ShuffleExhange for those buffers.

For instance, we have query ```select a, collect_list(b) from table group by a;```, which physical plan on CPU looks like:
```
ObjectHashAggregate(keys=[a#10], functions=[collect_list(b#11, 0, 0)], output=[a#10, collect_list(b)#17])
+- Exchange hashpartitioning(a#10, 200), true, [id=#13]
   +- ObjectHashAggregate(keys=[a#10], functions=[partial_collect_list(b#11, 0, 0)], output=[a#10, buf#21])
      +- LocalTableScan [a#10, b#11]
```
Among the outputs of the partial aggregate, `buf#21`, the attribute referring to aggregation buffer of CollectList, whose data type in CPU runtime(BinaryType) is different from GPU runtime(Array of ElementType). We need to override it in AggregateMeta. And we also need to propagate the overriding through ShuffleExchange (to the final stage).

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
sperlingxx: The reason of changing `GpuSortMeta` is similar to `GpuShuffleMeta`. But the change here is specialized for `SortAggregateExec` with TypedImperativeAggregate functions, which will appear when we disable the configuration:
`spark.sql.execution.useObjectHashAggregateExec`.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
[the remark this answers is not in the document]
abellina: Ok, if you know that the aggregate does the right thing in all cases and won't crash then I would remove the extra settings.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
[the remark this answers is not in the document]
sperlingxx: I've removed the setting.

abellina:
Thanks @sperlingxx. I am sorry for the delay, I'll take a look again today.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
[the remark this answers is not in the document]
revans2: I also think we can come up with a better name, even a long name like. `GpuTypeImperativeSupportedAggregateExecMeta`

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -375,6 +375,125 @@ def test_hash_reduction_pivot_without_nans(data_gen, conf):
...
+]
+
+@approximate_float
+@ignore_order(local=True)
+@incompat
+@pytest.mark.parametrize('data_gen', _gen_data_for_collect_op, ids=idfn)
+@pytest.mark.parametrize('use_obj_hash_agg', [True, False], ids=idfn)
+def test_hash_groupby_collect_list(data_gen, use_obj_hash_agg):
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: gen_df(spark, data_gen, length=100)
```
revans2, line 407: why only 100 entries?

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -375,6 +375,125 @@ def test_hash_reduction_pivot_without_nans(data_gen, conf):
...
+    # test collect_ops with other distinct aggregations
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: gen_df(spark, data_gen, length=100)
+            .groupby('a')
+            .agg(f.sort_array(f.collect_list('b')),
+                 f.sort_array(f.collect_set('b')),
+                 f.countDistinct('c'),
+                 f.count('c')))
+
+    # test distinct collect with other aggregations
```
revans2, line 436: Why use both SQL and Dataframe to do the same thing in the same test?

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -375,6 +375,125 @@ def test_hash_reduction_pivot_without_nans(data_gen, conf):
...
+                    count(c)
+            from tbl group by a"""
+    assert_gpu_and_cpu_are_equal_sql(
+        df_fun=lambda spark: gen_df(spark, data_gen, length=100),
+        table_name="tbl", sql=sql)
+
+# Queries with multiple distinct aggregations will fallback to CPU if they also contain
+# collect aggregations. Because Spark optimizer will insert expressions like `If` and `First`
+# when rewriting distinct aggregates, while `GpuIf` and `GpuFirst` doesn't support the datatype
+# of collect aggregations (ArrayType).
```
revans2, line 450: Can you dd in references to the issue to support these for GpuIF and GpuFirst? If they do not exist, then could you please file them?

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -375,6 +375,125 @@ def test_hash_reduction_pivot_without_nans(data_gen, conf):
...
+@incompat
+@pytest.mark.parametrize('data_gen', _gen_data_for_collect_op, ids=idfn)
+def test_hash_groupby_collect_with_multi_distinct_fallback(data_gen):
+    sql = """select a,
+                    sort_array(collect_list(b)),
+                    sort_array(collect_set(b)),
+                    count(distinct b),
+                    count(distinct c)
+            from tbl group by a"""
+    assert_gpu_and_cpu_are_equal_sql(
```
revans2, line 465: Typically when we have a fallback test we want an assertion that verifies part of the code actually did fall back like `assert_gpu_sql_fallback_collect`

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -375,6 +375,125 @@ def test_hash_reduction_pivot_without_nans(data_gen, conf):
...
+@ignore_order(local=True)
+@allow_non_gpu('ObjectHashAggregateExec', 'ShuffleExchangeExec',
+               'HashPartitioning', 'SortArray', 'Alias', 'Literal',
+               'Count', 'CollectList', 'CollectSet', 'AggregateExpression')
+@incompat
+@pytest.mark.parametrize('data_gen', _gen_data_for_collect_op, ids=idfn)
+@pytest.mark.parametrize('conf', [_nans_float_conf_partial, _nans_float_conf_final], ids=idfn)
+@pytest.mark.parametrize('aqe_enabled', ['true', 'false'], ids=idfn)
+def test_hash_groupby_collect_partial_replace_fallback(data_gen, conf, aqe_enabled):
+    conf.update({'spark.sql.adaptive.enabled': aqe_enabled})
```
revans2, line 480: Please always copy conf before doing an update.  We have seen issues with global values being modified by tests doing this and it is just good practice.

revans2 (commented):
This mostly looks good.  My main problem with this is

> In addition, Aggregate stacks with TypedImperativeAggregate functions may lead to unexpected crash if the stack falls back to CPU partially, because GPU data types are inconsistent with CPU counterparts. This problem will be fixed in the task 4 of #2916. To avoid this kind of unexpected crash in current, we bring up the "associated fallback" mechanism in this PR, which only affects Aggregate plans containing TypedImperativeAggregate functions.

We cannot have CollectList and CollectSet on by default if there are chances that we can crash.

integration_tests/src/main/python/hash_aggregate_test.py
sperlingxx: It just follows other tests on hash aggregate. Perhaps for test efficiency?

integration_tests/src/main/python/hash_aggregate_test.py
sperlingxx: There exists a small difference between the Query by Dataframe and the Query by SQL. In Query by SQL, there is a distinct collect_list `collect_list(distinct b)`, which can not be represented via Dataframe API.

sperlingxx:
> This mostly looks good. My main problem with this is
> 
> > In addition, Aggregate stacks with TypedImperativeAggregate functions may lead to unexpected crash if the stack falls back to CPU partially, because GPU data types are inconsistent with CPU counterparts. This problem will be fixed in the task 4 of #2916. To avoid this kind of unexpected crash in current, we bring up the "associated fallback" mechanism in this PR, which only affects Aggregate plans containing TypedImperativeAggregate functions.
> 
> We cannot have CollectList and CollectSet on by default if there are chances that we can crash.

Yes, and I believe we can get rid of the potential crashes with "associated fallback".

integration_tests/src/main/python/hash_aggregate_test.py
sperlingxx: Fixed.

integration_tests/src/main/python/hash_aggregate_test.py
sperlingxx: I added the check of fallback capture.

integration_tests/src/main/python/hash_aggregate_test.py
sperlingxx: Yes, I filed the issue.

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
```diff
@@ -486,16 +486,32 @@ private static DType toRapidsOrNull(DataType type) {
       } else {
         return DecimalUtil.createCudfDecimal(dt.precision(), dt.scale());
       }
+    } else if (supportNestedType) {
```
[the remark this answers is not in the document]
sperlingxx: I agree. So, I reverted this change.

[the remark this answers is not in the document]
sperlingxx: "As a side note, are we seeing issues with this? Are we collecting a list/struct and the types are not correct?"

For CollectList and CollectSet, I believe the types will always correct. But I am not sure whether there are some aggregations which we plan to support in future producing the inconsistent nested types.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -927,6 +942,185 @@ abstract class GpuBaseAggregateMeta[INPUT <: SparkPlan](
...
+   * Recursively collect all PlanMetas of input LogicalPlan. At the same time, and check whether
+   * existing plans which can NOT be replaced among them. If any, return true as the label of
+   * the entire fallback.
+   */
+  private def recursiveCheckForFallback(
+      currentMeta: SparkPlanMeta[_],
+      logical: LogicalPlan,
+      metaOfAllStages: mutable.ListBuffer[GpuBaseAggregateMeta[_]]): Boolean = {
+    currentMeta match {
+      case aggMeta: GpuBaseAggregateMeta[_] if aggMeta.agg.logicalLink.contains(logical) =>
```
revans2, line 1111: This looks like it searches the entire query plan, but just looking at the first part of the query.  If any of them cannot be replaced, then none of them will be replaced. So what happens if there are multiple aggregations that take place in a single query? It looks like we are now at a point where even one aggregation that falls back to the CPU will cause all of them to fall back after it. Even if they are not TypeImperative aggregates.

sperlingxx: I believe it will only fall back AggregateExecs created from same logical plan as the target stage. I added a test to verify that no other aggregations will be affected if there exists more than one Aggregation plans: `test_hash_groupby_collect_partial_replace_fallback_with_other_agg`.

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -375,6 +375,140 @@ def test_hash_reduction_pivot_without_nans(data_gen, conf):
...
+            .agg(f.sort_array(f.collect_list('b')), f.sort_array(f.collect_set('b'))),
+        conf=local_conf)
+    assert_gpu_fallback_collect(
+        lambda spark: gen_df(spark, data_gen, length=100)
+            .groupby('a')
+            .agg(f.sort_array(f.collect_list('b')), f.sort_array(f.collect_set('b'))),
+        cpu_fallback_class_name='ObjectHashAggregateExec',
+        conf=local_conf)
+    # test with single Distinct
+    assert_gpu_and_cpu_are_equal_collect(
```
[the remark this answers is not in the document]
sperlingxx: Refined.

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -375,6 +375,140 @@ def test_hash_reduction_pivot_without_nans(data_gen, conf):
...
+def test_hash_groupby_collect_partial_replace_fallback(data_gen, conf, aqe_enabled):
+    local_conf = conf.copy()
+    local_conf.update({'spark.sql.adaptive.enabled': aqe_enabled})
+    # test without Distinct
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: gen_df(spark, data_gen, length=100)
+            .groupby('a')
+            .agg(f.sort_array(f.collect_list('b')), f.sort_array(f.collect_set('b'))),
+        conf=local_conf)
+    assert_gpu_fallback_collect(
```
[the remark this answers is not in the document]
sperlingxx: Refined.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -927,6 +942,185 @@ abstract class GpuBaseAggregateMeta[INPUT <: SparkPlan](
...
+    // Avoid duplicated check and fallback.
+    val checked = meta.agg.getTagValue[Boolean](entireAggFallbackCheck).contains(true)
+    if (checked) return
+
+    meta.agg.setTagValue(entireAggFallbackCheck, true)
+    val logicalPlan = meta.agg.logicalLink.get
+    val stageMetas = mutable.ListBuffer[GpuBaseAggregateMeta[_]]()
+    // Go through all Aggregate stages to check whether all stages is GPU supported. If not,
+    // we fall back all GPU supported stages to CPU.
+    if (recursiveCheckForFallback(meta, logicalPlan, stageMetas)) {
```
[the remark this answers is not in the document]
sperlingxx: To adapt AQE, I take advantage of  `gpuSupportedTag` which was introduced by @andygrove. I added the line
```scala
wrapped.getTagValue(gpuSupportedTag).foreach(_.foreach(willNotWorkOnGpu))
```
in `GpuTypedImperativeSupportedAggregateExecMeta.tagPlanForGpu` to retrieve the information about GPU support which is captured and cached during `GpuQueryStagePrepOverrides`.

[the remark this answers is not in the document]
revans2: I understand, but does that run over the entire plan at some point, or is it just sections of the plan.  If it is the entire plan it would be good to explain that in a comment, because otherwise it looks like we have cases where we can crash.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -927,6 +942,185 @@ abstract class GpuBaseAggregateMeta[INPUT <: SparkPlan](
...
+object GpuTypedImperativeSupportedAggregateExecMeta {
+
+  private val entireAggFallbackCheck = TreeNodeTag[Boolean](
+    "rapids.gpu.checkAndFallbackAggregateExecEntirely")
+
+  private def checkAndFallbackEntirely(
+      meta: GpuTypedImperativeSupportedAggregateExecMeta[_]): Unit = {
+    // We only run the check for final stages which contain TypedImperativeAggregate.
+    val needToCheck = meta.agg.aggregateExpressions.exists(e =>
+      (e.mode == Final || e.mode == Complete) &&
```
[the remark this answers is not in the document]
revans2: We should also test this on databricks if you have not done so already.

revans2 (dismissed):
It looks good. I mostly want to be sure that we have run the tests on databricks. And it would be nice to have some of the comments in the fallback code updated to explain how it works with AQE so it is simpler to follow.

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -375,6 +377,163 @@ def test_hash_reduction_pivot_without_nans(data_gen, conf):
...
+@approximate_float
+@ignore_order(local=True)
+@incompat
+@pytest.mark.parametrize('data_gen', _gen_data_for_collect_op, ids=idfn)
+@pytest.mark.parametrize('use_obj_hash_agg', [True, False], ids=idfn)
+def test_hash_groupby_collect_list(data_gen, use_obj_hash_agg):
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: gen_df(spark, data_gen, length=100)
+            .groupby('a')
+            .agg(f.sort_array(f.collect_list('b')), f.count('b')),
```
abellina, line 411: It seems every `collect_*` function is wrapped in a `sort_array`. Is that on purpose? Could a comment be added somewhere on why? Especially because we have `@ignore_order` so I was curious.

abellina (commented):
I had at this point the comment on the tests, but otherwise it is looking good so far.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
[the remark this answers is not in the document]
sperlingxx: I made up some comments on this section.

integration_tests/src/main/python/hash_aggregate_test.py
sperlingxx: It is because `@ignore_order` only ensures the order between rows, while in these cases we also need to take care of the orders of each Array produced by collect ops.
And I added this comment to the test file.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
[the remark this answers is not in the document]
sperlingxx: I tested it on Databricks (https://blossom.nvidia.com/sw-gpu-spark-jenkins/view/Testing/job/lc-db/19/execution/node/60/log/). Everything looks fine.

integration_tests/src/main/python/hash_aggregate_test.py
abellina: Makes sense. I’d extend ignore_order to do the sorting after the collect for the array case. The reason being you would be able to test the aggregate in another way a user is likely to invoke.

I’m ok if you want to do that as a follow up also.

abellina (approved):

integration_tests/src/main/python/hash_aggregate_test.py
revans2: I would be fine with an `@ignore_array_order` or something like that. I'd rather not have `@ignore_order` cover both.

revans2: But do it as a follow on issue if we do it at all.

revans2 (approved):

Issue #2965, opened 2021-07-19
[BUG] test_round_robin_sort_fallback failed with ValueError: 'a_1' is not in list


jlowe:
In the nightly Dataproc integration tests for 21.08, test_round_robin_sort_fallback failed:
```
[2021-07-19T12:07:52.837Z] =================================== FAILURES ===================================
[2021-07-19T12:07:52.837Z] [31m[1m_ test_round_robin_sort_fallback[[('a', Struct(('a_1', Struct(('a_1_1', Integer),('a_1_2', Float),('a_1_3', Double))),('b_1', Long)))]] _[0m
[2021-07-19T12:07:52.837Z] 
[2021-07-19T12:07:52.837Z] self = Row(a=None, extra=1), item = 'a_1'
[2021-07-19T12:07:52.837Z] 
[2021-07-19T12:07:52.837Z]     [94mdef[39;49;00m [92m__getitem__[39;49;00m([96mself[39;49;00m, item):
[2021-07-19T12:07:52.837Z]         [94mif[39;49;00m [96misinstance[39;49;00m(item, ([96mint[39;49;00m, [96mslice[39;49;00m)):
[2021-07-19T12:07:52.837Z]             [94mreturn[39;49;00m [96msuper[39;49;00m(Row, [96mself[39;49;00m).[92m__getitem__[39;49;00m(item)
[2021-07-19T12:07:52.837Z]         [94mtry[39;49;00m:
[2021-07-19T12:07:52.837Z]             [90m# it will be slow when it has many fields,[39;49;00m
[2021-07-19T12:07:52.837Z]             [90m# but this will not be used in normal cases[39;49;00m
[2021-07-19T12:07:52.837Z] >           idx = [96mself[39;49;00m.__fields__.index(item)
[2021-07-19T12:07:52.837Z] [1m[31mE           ValueError: 'a_1' is not in list[0m
[2021-07-19T12:07:52.837Z] 
[2021-07-19T12:07:52.837Z] [1m[31m/hadoop/yarn/nm-local-dir/usercache/root/appcache/application_1626687042683_0001/container_e01_1626687042683_0001_01_000001/pyspark.zip/pyspark/sql/types.py[0m:1558: ValueError
[2021-07-19T12:07:52.837Z] 
[2021-07-19T12:07:52.837Z] [33mDuring handling of the above exception, another exception occurred:[0m
[2021-07-19T12:07:52.837Z] 
[...]

PR #2973, opened 2021-07-20
Assert that CPU and GPU row fields match when present

jlowe:
While investigating #2965, I noticed that the assert code was checking if both the CPU and GPU had fields but was not verifying that the two had the _same fields_.  This adds a sanity check that the CPU and GPU fields match before trying to lookup values by the field names.


jlowe:
build

kuhushukla (approved):

Issue #2972, opened 2021-07-20
[BUG] When using timeout option(-t) of qualification tool, it does not print anything in output after timeout.


viadea:
**Describe the bug**
When using timeout option(-t) of qualification tool, it does not print anything in output after timeout.

```
ERROR Qualification: Processing log files took longer then 20 seconds, stopping processing any more event logs
INFO ToolTextFileWriter: CSV output location: /tmp/testqualification/rapids_4_spark_qualification_output/rapids_4_spark_qualification_output.csv
==========================================================
|App ID|App Duration|SQL DF Duration|Problematic Duration|
==========================================================
==========================================================
```
**Steps/Code to reproduce bug**
Please provide a list of steps or a code sample to reproduce the issue.
Avoid posting private or sensitive data.

```
export SPARK_HOME=/xxx/spark
java -cp /Users/xxx/github/spark-rapids/tools/target/rapids-4-spark-tools_2.12-21.08.0-SNAPSHOT.jar:$SPARK_HOME/jars/* com.nvidia.spark.rapids.tool.qualification.QualificationMain \
--num-threads 8 \
--order desc \
-o /tmp/testqualification/ \
-t 20 \
/Users/xxx/eventlog/tpcds/49/49
```

**Expected behavior**
I expect after timeout, the output of the qualification tool can print the output for whatever logs processed.

**Environment details (please complete the following information)**
 - Environment location: [Standalone, YARN, Kubernetes, Cloud(specify cloud provider)]
 - Spark configuration settings related to the issue

21.08 snapshot qualification tool.
Spark 3.1.1

**Additional context**

viadea:
Close this issue as per my tests. This is just slower execution of the qualification tool than last version.

tgravescs:
using hthis to look into why its slower

tgravescs:
the problem is here: https://github.com/NVIDIA/spark-rapids/commit/bb8baabe65c0f93e6e0b6f4d50b140c1393413c3#diff-8190fc4fb61da36bdcb2aaff10beba7ed17a5bd08a5157d4b981765b9f36e605R92

Where we changed a foreach over list to be a while. The while loop is just 4-5x slower.

I'm looking at changing back to a List operation

PR #2974, opened 2021-07-20
Qualification tool: fix performance regression

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/2972

It turns out this while loop is 4-5x slower. One particular event log took about 4.5 seconds to process before and when we switched to use while it took slower to 20 seconds.  

In this case we use find{} instead to loop through all elements until we hit an event that says we are done. The done is really only used for the app filtering logic when we want to stop processing the event log after the app start event.


tgravescs:
build

nartal1 (approved):

tgravescs:
integration test failed which has nothing to do with this change since this is in tools, even at that doesn't show what failed.

tgravescs:
build

Issue #2589, opened 2021-06-04
[BUG] RapidsShuffleHeartbeatManager needs to remove executors that are stale


abellina:
We only add executors in `RapidsShuffleHeartbeatManager` but we never remove them.

We need to keep track of the last time we saw a heartbeat, and as we get more registrations check if any executors are stale, and not include in executor update messages. We don't need to tell already registered executors about the loss of a peer, because UCX should be able to figure that out on its own.

We also need to update the logic on how the "new executors" for a peer is decided. Currently it's based on registration order, and it will need to be time based I believe.

abellina:
This PR https://github.com/NVIDIA/spark-rapids/pull/2587 is related and improves things slightly for 21.06.

PR #2977, opened 2021-07-20
Expire executors from the RAPIDS shuffle heartbeat manager on timeout

abellina:
Signed-off-by: Alessandro Bellina <abellina@gmail.com>

Closes https://github.com/NVIDIA/spark-rapids/issues/2589

This adds a timeout to the RAPIDS Shuffle Manager heartbeat manager that when triggered it will remove the executor from the set of executors peers should care about.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -994,10 +994,20 @@ object RapidsConf {
...
+      .doc("Shuffle early start heartbeat interval (milliseconds). " +
+        "Executors will send a heartbeat RPC message to the driver at this interval")
       .integerConf
       .createWithDefault(5000)
 
+  val SHUFFLE_TRANSPORT_EARLY_START_HEARTBEAT_TIMEOUT =
+    conf("spark.rapids.shuffle.transport.earlyStart.heartbeatTimeout")
+      .doc(s"Shuffle early start heartbeat timeout (milliseconds). " +
+        s"Executors that don't heartbeat within this timeout will be considered stale. " +
+        s"This timeout must be set higher than: " +
```
jlowe, line 1006:
```suggestion
        s"This timeout must be higher than the value for " +
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsShuffleHeartbeatManager.scala
```diff
@@ -79,31 +115,71 @@ class RapidsShuffleHeartbeatManager extends Logging {
...
   }
+
+  // the timeout amount has elapsed, assume this executor is dead
+  private def isStaleHeartbeat(hbMillis: Long, currentTime: Long): Boolean = {
+    (currentTime - hbMillis) > heartbeatTimeoutMillis
+  }
+
+  // must be called holding the object lock
+  private def removeDeadExecutors(currentTime: Long,
+                                  heartbeatingExec: Option[BlockManagerId]): Unit = {
```
jlowe, line 161: I'm confused about this parameter.  Why are we special-casing the scenario where the current heartbeat is for an executor that is stale?  We need to gracefully handle the case where an executor heartbeats and is already considered stale, so I don't see the need to special case this.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsShuffleHeartbeatManager.scala
```diff
@@ -79,31 +115,71 @@ class RapidsShuffleHeartbeatManager extends Logging {
    * @return `RapidsExecutorUpdateMsg` with new executors, since the last heartbeat was received.
    */
   def executorHeartbeat(id: BlockManagerId): RapidsExecutorUpdateMsg = synchronized {
-    val lastRegistration = lastRegistrationSeen.get(id)
+    val lastRegistration = executorRegistrations.get(id)
     if (lastRegistration == null) {
       throw new IllegalStateException(s"Heartbeat from unknown executor $id")
```
jlowe, line 120: Speaking of gracefully handling executors that heartbeat after becoming stale, this doesn't seem like it's gracefully handling it.  Note that the default timeout is only 10 seconds, and a dropped TCP packet could result in a minute or more before it is recovered.  Also when running on an overloaded machine or heavily GC'ing JVM, I could easily see it taking more than 10 seconds, so I would fully expect this to happen in the real world.

abellina: Yes, thought about this but I didn't go for the change, thanks for raising. I believe the stale executor could rejoin with a new order. Just need to make sure that the suddenly reappearing executor doesn't cause issues at the executor level trying to connect again.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsShuffleHeartbeatManager.scala
abellina: I did that because if the current executor is telling me it is not stale, I thought it was wasteful to remove and re-add. Yes the graceful rejoin prompted me to do that. I am planning on handling that part as suggested. The other comment made me think, if we have several executors all GCing and missing heartbeats, then the re-add is going to happen anyway for most of those.. so this isn't probably saving is much.

jlowe: Exactly.  If there's a lot of executors then the chance that _this exact executor heartbeat_ is when an executor would have been declared stale is exceedingly small.  Since we have to gracefully handle the case where the executor is already considered stale when it heartbeats, I personally would simply make trimming stale executors the first thing any registration or heartbeat handler code does before it proceeds.

abellina:
thanks for the comments @jlowe, added https://github.com/NVIDIA/spark-rapids/pull/2977/commits/64ec75570f139e7c974197d2d62aa5aaf4b91d3a, to handle them. In the shuffle manager we use `getConnection`, which handles several calls to that function, so returning an executor again after a timeout is harmless.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
abellina: Handled: https://github.com/NVIDIA/spark-rapids/pull/2977/commits/f129279129d60553d1041fb5a0f44734a63caf73

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsShuffleHeartbeatManager.scala
```diff
@@ -79,30 +115,67 @@ class RapidsShuffleHeartbeatManager extends Logging {
...
+          leastRecentHeartbeat.remove(e.id)
+        } else {
+          aliveExecutors.append(e)
         }
-      } else {
-        // We are iterating backwards and have found the last index previously inspected
-        // for this peer. We can stop since all peers below this have been sent to this peer.
-        done = true
       }
+      executors.clear()
```
jlowe, line 176: Nit: Not seeing the point of clearing an ArrayBuffer we're about to make garbage anyway

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsShuffleHeartbeatManager.scala
```diff
@@ -79,30 +115,67 @@ class RapidsShuffleHeartbeatManager extends Logging {
    * @return `RapidsExecutorUpdateMsg` with new executors, since the last heartbeat was received.
    */
   def executorHeartbeat(id: BlockManagerId): RapidsExecutorUpdateMsg = synchronized {
-    val lastRegistration = lastRegistrationSeen.get(id)
+    removeDeadExecutors(getCurrentTimeMillis)
+    val lastRegistration = executorRegistrations.get(id)
     if (lastRegistration == null) {
-      throw new IllegalStateException(s"Heartbeat from unknown executor $id")
+      logDebug(s"Heartbeat from unknown executor $id")
+      registerExecutor(id)
```
jlowe, line 122: Are executors robust in terms of being notified of executors they already know about (including themselves?!). Glancing at the code, it looks like they will blindly attempt to connect to each executor reported in the message.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsShuffleHeartbeatManager.scala
```diff
@@ -79,30 +115,67 @@ class RapidsShuffleHeartbeatManager extends Logging {
...
+    val leastRecentHb = leastRecentHeartbeat.peek() // look at the executor that is lagging most
+    if (leastRecentHb != null &&
+        isStaleHeartbeat(
+          executorRegistrations.get(leastRecentHb).lastHeartbeatMillis.getValue, currentTime)) {
+      // make a new buffer of alive executors and replace the old one
+      val aliveExecutors = new ArrayBuffer[ExecutorRegistration]()
+      executors.foreach { e =>
+        if (isStaleHeartbeat(e.lastHeartbeatMillis.getValue, currentTime)) {
+          logDebug(s"Stale exec, removing $e")
+          executorRegistrations.remove(e.id)
```
jlowe, line 170: This removes the stale executor from the driver, but how are the other executors notified to update their maps?

abellina: We don't. The assumption is that UCP Listener will detect the timeout eventually and tear things down.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsShuffleHeartbeatManager.scala
abellina: On `registerExecutor` we get the collection of executors to send to the registrant _before_ adding ourselves. So it doesn't happen, since we ended up in `registerExecutor` because we are not in the list. That said, I should make this more robust. The transport should refuse (and blow up) if it is trying to connect to itself, there is no reason to do that. Let me know if you'd like to see that in this PR.

The heartbeat endpoint, on update, ends up calling `https://github.com/NVIDIA/spark-rapids/blob/branch-21.08/shuffle-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/ucx/UCX.scala#L787`. This protects against existing connections. This is the same function we call when we are initializing the `RapidsShuffleIterator` (i.e. after the early wakeup happens).

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsShuffleHeartbeatManager.scala
abellina: Sorry one more thing, the `RapidsShuffleHeartbeatEndpoint` is stateless, so it's really all kept within the transport.

jlowe (approved):

jlowe:
build

PR #2982, opened 2021-07-21
Add FAQ entry for timezone error

jlowe:
This adds a FAQ entry to describe what the cudf error `Failed to open the timezone file` means and a common method for users to resolve it.  Improving the error message in cudf is being tracked by https://github.com/rapidsai/cudf/issues/8795.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -360,6 +360,14 @@ different because of compression. Users can turn on
 the GPU try to replicate more closely what the output ordering would have been if sort were used,
 like on the CPU.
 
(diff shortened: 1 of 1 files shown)


sameerz (commented):
Any reason to call out the tzdata package for Ubuntu as opposed to other platforms?

jlowe:
> Any reason to call out the tzdata package for Ubuntu as opposed to other platforms?

Because that was simple for me to verify.  Looks like CentOS/Fedora also uses a package named `tzdata` so I'll change it to say that's typically the package name.

jlowe:
build

sameerz (approved):

PR #2987, opened 2021-07-21
Update documentation for unsupported edge cases when casting from string to timestamp

### LINKED ISSUE #2889: [BUG] CAST string to timestamp produces null for some valid timestamps
**Describe the bug**
Casting the following strings to timestamp produces null values instead of valid timestamps.

```
"\nT2\n", // today, 02:00:00.0
"\nT\n", // today, midnight
"\n:\n", // today, midnight
"\n1:\n", // today, 01:00:00.0
"\n5:24\n", // today, 05:24:00.0
"\n:2\n", // today, 2 minutes after midnight
"\n1::2\n" // today, 01:00:02.0
```

I will check in the fuzzer used to generate these values as part of the PR that fixes the issue.

**Expected behavior**
We should produce the same values as Spark.

**Additional context**
None

andygrove:

Related to #2889

I started down the path of trying to support all the edge cases, and also looked at having a regex to detect the edge cases and throw an error, but even that seems to be a fair bit of work, so perhaps we are better off just documenting this.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -541,7 +541,12 @@ Casting from string to timestamp currently has the following limitations.
 
 - <a name="Footnote1"></a>[1] The timestamp portion must be complete in terms of hours, minutes, seconds, and
  milliseconds, with 2 digits each for hours, minutes, and seconds, and 6 digits for milliseconds. 
- Only timezone 'Z' (UTC) is supported. Casting unsupported formats will result in null values. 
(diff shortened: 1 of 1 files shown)


jlowe:
> Closes #2889

Does it?  The bug exists before and after this PR, so I would argue it does not.  I think this is an improvement, but I'd like to see #2889 remain open to track the lacking functionality.

jlowe (approved):

jlowe:
build

PR #3009, opened 2021-07-23
Propagate child output ordering in GpuCoalesceBatches

### LINKED ISSUE #3002: [BUG] writing parquet with partitionBy() loses sort order
**Describe the bug**
I'm trying to read from partitioned parquet, sort within partitions, and then write each partition back out.  Everything works great up until the last step where I call partitionBy().  The resulting file is no longer fully sorted.  Omitting the partitionBy yields the correct result, but without the partitioning.

Run the following from a spark shell

```scala
val df = sc.parallelize((1L to 10000000L)).map{i => ("a", f"$i%010d", i)}.toDF("partkey", "val", "val2")
df.repartition(1, $"partkey").sortWithinPartitions($"partkey", $"val", $"val2").write.mode("overwrite").partitionBy("partkey").parquet("file:///tmp/foosort")
df.repartition(1, $"partkey").sortWithinPartitions($"partkey", $"val", $"val2").write.mode("overwrite").parquet("file:///tmp/foosort2")
```

**Expected behavior**
After running the above, foosort will be partially sorted, but not completely.  foosort2 is fully sorted as expected.

This happens both standalone and running via YARN.  I'm using branch 21.08 from around 6/18/2021, with spark 3.2.0-SNAPSHOT.  I set spark.sql.adaptive.enabled=false
 

**Additional context**
The problem seems to go away if I change the "val" column to an Int.  I also notice that the plans in the spark GUI differ.  Without the partitionBy the last stage does GpuColumnarExchange->GpuShuffleCoalesce->GpuSort.  With partitionBy there are extra steps GpuColumnarExchange->GpuShuffleCoalesce->GpuSort->GpuCoalesceBatches->GpuProject->GpuSort.

jlowe:
Fixes #3002.

`GpuCoalesceBatches` was failing to propagate the child output ordering.  This can force the plan to re-sort the data on a subset of the keys that the plan had already sorted on, and without forcing a stable sort can cause inadvertent reordering of data.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
@@ -554,7 +554,7 @@ case class GpuCoalesceBatches(child: SparkPlan, goal: CoalesceGoal)
     case batchingGoal: BatchedByKey =>
       batchingGoal.order
     case _ =>
-      super.outputOrdering
+      child.outputOrdering
   }
 
   override def doExecuteColumnar(): RDD[ColumnarBatch] = {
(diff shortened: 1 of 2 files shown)


jlowe:
build

abellina (approved):

PR #3011, opened 2021-07-23
Support MapType in joins

### LINKED ISSUE #2133: [FEA] Join support for passing MapType columns along when not join keys

andygrove:
Closes #2133 

This adds MapType support for sort-merge join, hash join, and broadcast join.

I didn't add it for broadcast nested-loop join because it looks like we don't support other complex types for that one.

### DIFF
--- a/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
+++ b/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
@@ -231,19 +231,31 @@ abstract class SparkBaseShims extends SparkShims {
       GpuOverrides.exec[SortMergeJoinExec](
         "Sort merge join, replacing with shuffled hash join",
         ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.ARRAY +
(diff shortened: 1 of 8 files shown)


shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
```diff
@@ -231,20 +231,20 @@ abstract class SparkBaseShims extends SparkShims {
       GpuOverrides.exec[SortMergeJoinExec](
         "Sort merge join, replacing with shuffled hash join",
         ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL + TypeSig.ARRAY +
-            TypeSig.STRUCT).nested(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL
-        ), TypeSig.all),
+            TypeSig.STRUCT + TypeSig.MAP).nested(TypeSig.commonCudfTypes + TypeSig.NULL +
+          TypeSig.DECIMAL), TypeSig.all),
```
jlowe, line 235: I would expect psnotes for array, struct, and map stating these types are not supported as join keys.  Same comment applies to other execs with the same limitations.

jlowe (approved):

jlowe:
build

Issue #3007, opened 2021-07-23
[BUG] data mess up reading from ORC


wbo4958:
Assume the file schema of an ORC file is 
```
Type: struct<name:string,number:int,english:float,math:int,history:float>
```

Below reading code will result data mess-up
``` scala
      val schema = StructType(Array(
        StructField("number", IntegerType),
        StructField("name", StringType),
      ))
     val df = spark.read.schema(schema).orc("xxxxxx")
     df.show()
```

while below code works fine
``` scala
      val schema = StructType(Array(
        StructField("name", StringType),
        StructField("number", IntegerType)
      ))
     val df = spark.read.schema(schema).orc("xxxxxx")
     df.show()
```

That's because we are messing up the orc data with the read schema when re-constructing the orc files.

we're reading data according to the columns' sequence in orc file, while we are writing the read schema into the footer.
So, as long as the read schema is not following the sequence of orc file schema, then issue will happen.

This bug can be repro in latest 21.08-SNAPTHOT and in 21.06 release. I didnt' test the before release. It seems this bug always exist.

tgravescs:
full repro case:

```
// create a file like schema above
case class Testing(name: String, number: Int, english: Float, math: Int, history: Float)

val x = sc.parallelize(Array(Testing("three", 23, 3, 3, 4)))
val df = spark.createDataFrame(x)
df.write.orc("testing.orc")
val dfread = spark.read.orc("testing.orc")
dfread.printSchema()
root
 |-- name: string (nullable = true)
 |-- number: integer (nullable = true)
 |-- english: float (nullable = true)
 |-- math: integer (nullable = true)
 |-- history: float (nullable = true)

// now read it

import org.apache.spark.sql.types._
val schema = StructType(Array(
        StructField("number", IntegerType),
        StructField("name", StringType),
      ))
val df = spark.read.schema(schema).orc("testing.orc")
df.show()
df.printSchema
```

Spark seems to just flip the columns, so it doesn't force the schema read onto the column, it just dictates the order.

spark: 

```
scala> df.show()
+------+-----+
|number| name|
+------+-----+
|    23|three|
+------+-----+

scala> df.printSchema
root
 |-- number: integer (nullable = true)
 |-- name: string (nullable = true)
```


With the rapids plugin enabled we get:

```
scala> df.printSchema()
root
 |-- number: integer (nullable = true)
 |-- name: string (nullable = true)
scala> df.show()
+------+----+
|number|name|
+------+----+
[...]

PR #3015, opened 2021-07-23
Fix ORC read error when read schema reorders file schema columns

wbo4958:
The orc reader reads the needed columns according to the column order of
original orc file, but we are writing the file schema using reading schema
order. If the reading schema order is not following the file schema order
then the re-constructed orc file buffer will be mangled.

See the issue https://github.com/NVIDIA/spark-rapids/issues/3007

### DIFF
--- a/integration_tests/src/main/python/orc_test.py
+++ b/integration_tests/src/main/python/orc_test.py
@@ -264,6 +264,48 @@ def test_missing_column_names(spark_tmp_table_factory, reader_confs):
         lambda spark : spark.sql("SELECT _col3,_col2 FROM {}".format(table_name)),
         all_confs)
 
+def setup_orc_file_with_column_names(spark, table_name):
+    drop_query = "DROP TABLE IF EXISTS {}".format(table_name)
+    create_query = "CREATE TABLE `{}` (`c_1` INT, `c_2` STRING, `c_3` ARRAY<INT>) USING orc".format(table_name)
+    insert_query = "INSERT INTO {} VALUES(13, '155', array(2020))".format(table_name)
+    spark.sql(drop_query).collect
+    spark.sql(create_query).collect
+    spark.sql(insert_query).collect
+
+@pytest.mark.parametrize('reader_confs', reader_opt_confs, ids=idfn)
+def test_disorder_read_schema(spark_tmp_table_factory, reader_confs):
+    table_name = spark_tmp_table_factory.get()
+    with_cpu_session(lambda spark : setup_orc_file_with_column_names(spark, table_name))
+    all_confs = reader_confs.copy()
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark : spark.sql("SELECT c_2,c_1 FROM {}".format(table_name)),
+        all_confs)
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark : spark.sql("SELECT c_3,c_1 FROM {}".format(table_name)),
+        all_confs)
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark : spark.sql("SELECT c_3,c_2 FROM {}".format(table_name)),
+        all_confs)
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark : spark.sql("SELECT c_1,c_3,c_2 FROM {}".format(table_name)),
+        all_confs)
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark : spark.sql("SELECT c_1,c_2,c_3 FROM {}".format(table_name)),
+        all_confs)
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark : spark.sql("SELECT c_2,c_1,c_3 FROM {}".format(table_name)),
+        all_confs)
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark : spark.sql("SELECT c_2,c_3,c_1 FROM {}".format(table_name)),
+        all_confs)
+    assert_gpu_and_cpu_are_equal_collect(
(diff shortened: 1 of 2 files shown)


wbo4958:
This PR is not so good, since it will change code again when we try to support union/map for orc reading.

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala
```diff
@@ -909,20 +909,66 @@ private case class GpuOrcFileFilterHandler(
...
           nextOutputColumnId += 1
-        } else {
-          result(i) = -1
         }
       }
-      result
+
+      //  Recursively update columnMapping and idMapping according to the TypeDescription id
+      // For now, we are only supporting struct and list nested type
+      def updateMapping(prefix: String, schema: TypeDescription, isRoot: Boolean = false): Unit = {
```
jlowe, line 944: It's been a while since I dug into ORC, so apologies if I'm missing something basic.  I'm not understanding why we need a recursive method.  IIUC all of the columns, whether they're nested or not, will appear in `fileIncluded`.  We're updating a flat-mapping of columns in `columnMapping` and `idMapping`, so that seems to be the case.

If we're just building a mapping back from the new column to the old column and keeping the existing mapping of old column ID to new column ID, why wouldn't we keep the existing function logic and simply update `idMapping` when we update `columnMapping`?  If we know enough to setup the forward-mapping for the column then we know how to set the reverse mapping as well.  Given it's flat-mapped on the input, I don't see why we need to special-case nesting here.

integration_tests/src/main/python/orc_test.py
```diff
@@ -264,6 +264,48 @@ def test_missing_column_names(spark_tmp_table_factory, reader_confs):
...
+def setup_orc_file_with_column_names(spark, table_name):
+    drop_query = "DROP TABLE IF EXISTS {}".format(table_name)
+    create_query = "CREATE TABLE `{}` (`c_1` INT, `c_2` STRING, `c_3` ARRAY<INT>) USING orc".format(table_name)
+    insert_query = "INSERT INTO {} VALUES(13, '155', array(2020))".format(table_name)
+    spark.sql(drop_query).collect
+    spark.sql(create_query).collect
+    spark.sql(insert_query).collect
+
+@pytest.mark.parametrize('reader_confs', reader_opt_confs, ids=idfn)
+def test_disorder_read_schema(spark_tmp_table_factory, reader_confs):
```
tgravescs, line 276: can you add same test in parquet, I tested manually and it doesn't appear to have same issue.  I guess it can be done separately too.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala
wbo4958: the columnMapping should store the new columns' order following by **read schema** and it will be used when [setting the new footer](https://github.com/NVIDIA/spark-rapids/blob/branch-21.08/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala#L1031). But the original function logic always makes columnMapping store the old columns' order, which means we are pruning the columns according to the original orc file schema instead of read schema.

Given the below example,

Orc file schema: **struct<a:Int, b:string>**, the read scheam: **struct<b:string, a:int>**, then the fileIncluded=[true, true, true].

and the columnMapping in the original function will be **[0, 1, 2]**, which will follow the original ORC file schemal order, while the columnMapping with this PR will be **[0, 2, 1]**, which will follow the read schema order.

The recursive function is taking **read schema** as the parameter, and it will recursively lookup and set the mapping from the column id of read schema to the column id in the orc file schema.
 
I may not quite understand your question correctly, if that, sorry for that. @jlowe

jlowe: Thanks, @wbo4958.  I mistakenly read this as only creating a back-mapping with the new `idMapping` array and missed that it was also potentially reordering the existing `columnMapping`.

jlowe (approved):
This looks OK to me, but I don't think it will address @firestarman's desire to satisfy nested schema pruning.  I see that `checkSchemaCompatibility` will trim the top-level schema for columns that we don't need, but I think that needs to be a recursive function to potentially prune unnecessary child fields out of structs underneath the top-level schema.  As it is coded now, it takes the entire top-level column as-is, which I think means any struct columns will always have all of their child fields even if the read schema only wants a portion of the fields.

firestarman:
> This looks OK to me, but I don't think it will address @firestarman's desire to satisfy nested schema pruning. I see that `checkSchemaCompatibility` will trim the top-level schema for columns that we don't need, but I think that needs to be a recursive function to potentially prune unnecessary child fields out of structs underneath the top-level schema. As it is coded now, it takes the entire top-level column as-is, which I think means any struct columns will always have all of their child fields even if the read schema only wants a portion of the fields.

No, it will not.
As you mentioned, at least we need to update `checkSchemaCompatibility` accordingly to support nested column pruning. It will come in another PR.

wbo4958:
> This looks OK to me, but I don't think it will address @firestarman's desire to satisfy nested schema pruning. I see that `checkSchemaCompatibility` will trim the top-level schema for columns that we don't need, but I think that needs to be a recursive function to potentially prune unnecessary child fields out of structs underneath the top-level schema. As it is coded now, it takes the entire top-level column as-is, which I think means any struct columns will always have all of their child fields even if the read schema only wants a portion of the fields.

Yeah, we need to re-work checkSchemaCompatibility, I just have a simple test for nested-column-prune by just returning the read schema, and the dumped orc file has pruned the necessary columns.

Issue #3021, opened 2021-07-26
[BUG] When lots of NULLs in partition-column of the window, GPU run has task skew and is much slower than CPU.


viadea:
**Describe the bug**
When lots of NULLs in partition-column of the window, GPU run has task skew and is much slower than CPU.

**Steps/Code to reproduce bug**
Please provide a list of steps or a code sample to reproduce the issue.
Avoid posting private or sensitive data.

Just use NDS dataset to reproduce because the table `store_sales` has lots of NULLs in column `ss_customer_sk`.
```
val df = spark.read.parquet("/xxx/xxx/xxx/xxx_100G_parquet/store_sales")
df.createOrReplaceTempView("df")
spark.sql("""
select ss_customer_sk,avg(avg_price) as avg_price
from
(
 SELECT ss_customer_sk ,avg(ss_sales_price) OVER (PARTITION BY ss_customer_sk order by ss_sold_date_sk ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW ) as avg_price
    FROM df
) group by ss_customer_sk order by 2 desc limit 10
""").collect
```
GPU run: 6mins
CPU run: 30s
This is because in GPU run, here are several skewed tasks(16 out of 200 tasks) took around 6mins while other tasks only took 3s.

If we filter out the NULLs:
```
spark.conf.set("spark.sql.files.maxPartitionBytes","1g")
val df = spark.read.parquet("/xxx/xxx/data/xxx_100G_parquet/store_sales")
df.createOrReplaceTempView("df")
spark.sql("""
select ss_customer_sk,avg(avg_price) as avg_price
from
(
 SELECT ss_customer_sk ,avg(ss_sales_price) OVER (PARTITION BY ss_customer_sk order by ss_sold_date_sk ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW ) as avg_price
    FROM df
    where ss_customer_sk is not null
) group by ss_customer_sk order by 2 desc limit 10
""").collect
```
GPU run: 7s
CPU run: 30s

**Expected behavior**

GPU runs faster even though there are lots of NULLs in partition column of window.

**Environment details (please complete the following information)**
 - Environment location: [Standalone, YARN, Kubernetes, Cloud(specify cloud provider)]
 - Spark configuration settings related to the issue

Spark 3.1.1 Standalone
21.08 snapshot jars(2021-07-26 version)

**Additional context**

viadea:
[...]

PR #3022, opened 2021-07-26
Add in docs to describe window performance

revans2:
Right now this is a document only change, but I really would be interested to hear from others if we think we should include some configs to allow users to selectively disable window operations. The problem is that the slowness is based off of the window size, and this is data dependent, as you can see from #3021. That means we cannot detect ahead of time if we are going to see slowness. I will file a follow on issue/epic for optimizations we know we can do.

### DIFF
--- a/docs/tuning-guide.md
+++ b/docs/tuning-guide.md
@@ -339,17 +339,25 @@ Apache Spark supports a few optimizations for different windows patterns. Genera
 buffers all the data for a partition by key in memory and then loops through the rows looking for
 boundaries. When it finds a boundary change, it will then calculate the aggregation on that window.
 This ends up being `O(N^2)` where N is the size of the window. In a few cases in can improve on
-that.  These optimizations include.
+that. These optimizations include.
 
   * Lead/Lag. In this case Spark keeps an offset pointer and can output the result from the
-    buffered data in linear time.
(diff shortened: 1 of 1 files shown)


jlowe (approved):
Docs look good to me.

> I really would be interested to hear from others if we think we should include some configs to allow users to selectively disable window operations.

It might be handy in practice to have window configs to control the various window types.  That way if a user discovers a performance problem or bug in windowing then they can easily work around that case without forcing all window operations off the GPU.  If it's easy I say let's do it, but if it's quite a bit of work then maybe we wait until we more clearly see the need.  It seems to come down to how likely do we think a query will have multiple types of windows where disabling one window type but leaving another enabled is better than blanket disabling `GpuWindowExec`.

jlowe:
build

revans2:
build

jlowe:
> I am not sure that this is what we want either.

Yeah so then I'm thinking maybe we hold off a bit until we get clearer guidance from real-world use as to what we really need.

jlowe (approved):

revans2:
Had some offline review comments that I tried to clarify.

revans2:
build

revans2:
I filed https://github.com/NVIDIA/spark-rapids/issues/3023 to try and track some possible performance optimizations.

PR #3024, opened 2021-07-26
Check validity of any() or all() results that could be null

jlowe:
While reviewing the legacy timestamp casting changes, I noticed that `any()` and `all()` can return a null scalar result, but there are places in the plugin code where it blindly assumes the value cannot be null.  This updates those locations to check for validity of the scalar before checking the value, as it's undefined behavior to access a null scalar value.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
@@ -251,7 +251,7 @@ object GpuCast extends Arm {
     // have been stripped of leading and trailing whitespace
     val sanitized = withResource(input.containsRe("\\s")) { hasWhitespace =>
       withResource(hasWhitespace.any()) { any =>
-        if (any.getBoolean) {
+        if (any.isValid && any.getBoolean) {
           if (ansiEnabled) {
             throw new NumberFormatException(GpuCast.INVALID_INPUT_MESSAGE)
           } else {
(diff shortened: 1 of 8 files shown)


jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -622,7 +622,7 @@ case class GpuCast(
     def throwIfAny(cv: ColumnView): Unit = {
       withResource(cv) { cv =>
         withResource(cv.any()) { isAny =>
-          if (isAny.getBoolean) {
+          if (isAny.isValid && isAny.getBoolean) {
```
revans2, line 625: Are there any tests that hit this, or any of the others?  All nulls is not that common, but I am a bit concerned about some of the cast logic and I am not sure if all nulls in an ANSI cast is doing exactly what we want without a test for it.

jlowe: @andygrove did add tests for nulls which found some issues back when he fixed some `any`/`all` usage in #2933.  I'll try to make a test for this.  Note that `any`/`all` will also return a null scalar if the input column is empty in case this ever gets invoked on a degenerate batch.

jlowe: I added some tests that exercise these code paths, and I used a custom cudf that asserts in the Scalar.get* methods to verify they fail without the changes and pass afterwards.

jlowe:
build

revans2 (approved):

PR #3028, opened 2021-07-27
Enable parallel build for pre-merge job to reduce overall duration [skip ci]

zhanga5:
and upload logs to github for failed stage only. It will run `mvn verify` and `unit tests` stages in parallel. If any one failed will abort another if not finished yet.


zhanga5:
build

jenkins/spark-premerge-build-ut.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
pxLi, line 1: can we just put the content into spark-premerge-build.sh and use the flag/ENV to control if run?  so people do not need to find 2 places to make changes

jenkins/Jenkinsfile-blossom.premerge
```diff
@@ -151,37 +154,74 @@ pipeline {
...
+                    }
+
+                    steps {
+                        script {
+                            unstash "source_tree"
+
+                            container('gpu') {
+                                timeout(time: 2, unit: 'HOURS') {
+                                    sh "sleep 60"
+                                    sh "exit 1"
```
pxLi, line 217: for testing?

jenkins/Jenkinsfile-blossom.premerge
```diff
@@ -151,37 +154,74 @@ pipeline {
...
+                            customWorkspace "${CUSTOM_WORKSPACE}-ut" // Use different workspace to avoid conflict with IT
+                        }
+                    }
+
+                    steps {
+                        script {
+                            unstash "source_tree"
+
+                            container('gpu') {
+                                timeout(time: 2, unit: 'HOURS') {
```
pxLi, line 215: could give it the enough timeout maybe same timeout as IT stage (4 hours), so people will not have to increase the timeout before adding new UTs

jenkins/Jenkinsfile-blossom.premerge
zhanga5: yes, that's right. Here try to test upload logs for failed stage. Will remove it since it wouldn't take effect until merged

jenkins/Jenkinsfile-blossom.premerge
zhanga5: OK, will refine it

jenkins/spark-premerge-build-ut.sh
zhanga5: here just try to keep it simple. So people can easily run IT or UT separately. I can't think of any case to make change 2 places at the same time.

BTW, `. jenkins/version-def.sh` would be removed as it's not required by UT

pxLi: I don't think people would like to use this script to run test locally, it was for pre-merge only

zhanga5: OK, then I would suggest just move to Jenkinsfile as it's quite simple. And it would be easier to run different version UT in parallel if necessary in the future

jenkins/Jenkinsfile-blossom.premerge
```diff
@@ -151,37 +153,72 @@ pipeline {
...
+                                          execPattern           : '**/target/jacoco.exec',
+                                          classPattern          : 'target/jacoco_classes/',
+                                          sourcePattern         : 'shuffle-plugin/src/main/scala/,udf-compiler/src/main/scala/,sql-plugin/src/main/java/,sql-plugin/src/main/scala/,shims/spark311/src/main/scala/,shims/spark301db/src/main/scala/,shims/spark301/src/main/scala/,shims/spark302/src/main/scala/,shims/spark303/src/main/scala/,shims/spark304/src/main/scala/,shims/spark312/src/main/scala/,shims/spark313/src/main/scala/',
+                                          sourceInclusionPattern: '**/*.java,**/*.scala'
+                                    ])
+                                }
+                            }
+                        }
+                    }
+                } // end of Integration Test stage
```
pxLi, line 191: nit: also the comment

jenkins/Jenkinsfile-blossom.premerge
```diff
@@ -151,37 +153,72 @@ pipeline {
...
-                            sh "$PREMERGE_SCRIPT"
-                            step([$class                : 'JacocoPublisher',
-                                  execPattern           : '**/target/jacoco.exec',
-                                  classPattern          : 'target/jacoco_classes/',
-                                  sourcePattern         : 'shuffle-plugin/src/main/scala/,udf-compiler/src/main/scala/,sql-plugin/src/main/java/,sql-plugin/src/main/scala/,shims/spark311/src/main/scala/,shims/spark301db/src/main/scala/,shims/spark301/src/main/scala/,shims/spark302/src/main/scala/,shims/spark303/src/main/scala/,shims/spark304/src/main/scala/,shims/spark312/src/main/scala/,shims/spark313/src/main/scala/',
-                                  sourceInclusionPattern: '**/*.java,**/*.scala'
-                            ])
+            failFast true
+            parallel {
+                stage('Integration Test') {
```
pxLi, line 159: nit: can we rename it as something like `mvn verify`, since it actually does more than just integration test

jenkins/Jenkinsfile-blossom.premerge
```diff
@@ -151,37 +153,72 @@ pipeline {
...
+                        }
+                    }
+
+                    steps {
+                        script {
+                            unstash "source_tree"
+
+                            container('gpu') {
+                                timeout(time: 4, unit: 'HOURS') {
+                                    sh "RUN_UT_ONLY=yes $PREMERGE_SCRIPT"
```
pxLi, line 215: this line does not match `if [[ ! -z "$RUN_UT_ONLY" ]]; then`

jenkins/Jenkinsfile-blossom.premerge
```diff
@@ -197,9 +234,9 @@ pipeline {
                 } else {
                     // upload log only in case of build failure
                     def guardWords = ["gitlab.*?\\.com", "urm.*?\\.com"]
-                    def logPattern = "### BEGIN OF TEST LOG ###.*### END OF TEST LOG ###"
+                    guardWords.add("nvidia-smi(?s)(.*?)(?=jenkins/version-def.sh)") // hide GPU info
```
pxLi, line 237: looks like this will filter out the UT log part w/ new commit?

jenkins/spark-premerge-build.sh
```diff
@@ -19,12 +19,16 @@ set -ex
 
 nvidia-smi
 
-function on_exit {
-    echo '### END OF TEST LOG ###'
-}
-trap on_exit EXIT
+if [[ ! -z "$RUN_UT_ONLY" ]]; then
```
pxLi, line 22: I think we could be better to add the full branches here like,
1. VERIFY_ONLY || ALL
2. UT_ONLY || ALL

So this is more people to modify and keep it simplicity if need to bring it back to sequential stages

jenkins/Jenkinsfile-blossom.premerge
zhanga5: OK, will rename it

jenkins/Jenkinsfile-blossom.premerge
zhanga5: not sure, what's convention in this project. To avoid confusing with yes/no, true/false (and case sensitive), usually would test whether it's empty or not

jenkins/Jenkinsfile-blossom.premerge
zhanga5: nice catch! Will fix it

jenkins/spark-premerge-build.sh
zhanga5: I'm trying to minimize the change. If need cover above more cases, I would suggest make this script parameterized, such as `xxx-build.sh [TYPE]`, e.g, `xxx-build.sh UT`, `xxx-build.sh IT` and `all` by default

pxLi: > To avoid confusing with yes/no, true/false (and case sensitive), usually would test whether it's empty or not

sure, but this this one. I think to test if empty maybe not enough, `RUN_UT_ONLY=yes` people would obviously considering to disable that they should set it as RUN_UT_ONLY=no, but this won't work

pxLi: talked offline. If adding arg, please make sure its `VERIFY` instead of `IT` as above

pxLi:
convert to draft to avoid accidental merge. We will instantly do following tests against this repo

jenkins/Jenkinsfile-blossom.premerge
```diff
@@ -151,37 +153,72 @@ pipeline {
...
+                            customWorkspace "${CUSTOM_WORKSPACE}"
                         }
                     }
-                }
-            }
+
+                    steps {
+                        script {
+                            container('gpu') {
+                                // TODO: improve resource management
```
tgravescs, line 178: what does this mean?

tgravescs (commented):
so to be clear, this is running mvn verify in parallel with the unit tests, correct. So basically build and integration tests run in one and unit tests run in another?   can you update description to clarify.

jenkins/Jenkinsfile-blossom.premerge
pxLi: this TODO was added by me earlier, since we added lockable resource and parallel run, we can remove this one cc @zhanga5

zhanga5: OK, will remove it

zhanga5:
> so to be clear, this is running mvn verify in parallel with the unit tests, correct. So basically build and integration tests run in one and unit tests run in another? can you update description to clarify.

yes, that's right. Updated description and added more comments to Jenkinsfile

pxLi (approved):
LGTM.
Alex has verified in forked repo, we are going trigger a PR build right after the merge.

pxLi:
build

PR #3032, opened 2021-07-27
Add disorder read schema test case for Parquet

wbo4958:

This PR is a following PR for https://github.com/NVIDIA/spark-rapids/issues/3007. We've found when the column order of read schema is not matching with column order in the file schema for ORC reading, then the data read will be mangled.  And there is PR https://github.com/NVIDIA/spark-rapids/pull/3015 to fix this issue.

So make this PR to check if there is the same issue for Parquet.


wbo4958:
build

tgravescs (approved):

PR #3034, opened 2021-07-27
 Support Sort on nested struct 

### LINKED ISSUE #2878: [FEA] Support Sort on nested struct
A clear and concise description of what the problem is. Ex. I wish the RAPIDS Accelerator for Apache Spark would [...]
Support Sort on nested struct.

Currently the Sort can only work on nested struct:
```
!Exec <SortExec> cannot run on GPU because not all expressions can be replaced
  !Expression <SortOrder> col#74 ASC NULLS FIRST cannot run on GPU because expression AttributeReference col#74 produces an unsupported type StructType(StructField(name,StructType(StructField(firstname,StringType,true)...
```

GpuSort should work on nested struct.

**Additional context**
Add any other context, code examples, or references to existing implementations about the feature request here.

sperlingxx:
Closes #2878

### DIFF
--- a/integration_tests/src/main/python/sort_test.py
+++ b/integration_tests/src/main/python/sort_test.py
@@ -40,8 +40,7 @@ def test_single_orderby(data_gen, order):
 @pytest.mark.parametrize('stable_sort', ['STABLE', 'OUTOFCORE'])
 @pytest.mark.parametrize('data_gen', [
     pytest.param(all_basic_struct_gen),
-    pytest.param(StructGen([['child0', all_basic_struct_gen]]),
-        marks=pytest.mark.xfail(reason='second-level structs are not supported')),
+    pytest.param(StructGen([['child0', all_basic_struct_gen]])),
     pytest.param(ArrayGen(string_gen),
         marks=pytest.mark.xfail(reason="arrays are not supported")),
     pytest.param(MapGen(StringGen(pattern='key_[0-9]', nullable=False), simple_string_to_string_map_gen),
@@ -78,8 +77,7 @@ def test_single_orderby_with_limit(data_gen, order):
 
 @pytest.mark.parametrize('data_gen', [
     pytest.param(all_basic_struct_gen),
-    pytest.param(StructGen([['child0', all_basic_struct_gen]]),
-                 marks=pytest.mark.xfail(reason='second-level structs are not supported')),
+    pytest.param(StructGen([['child0', all_basic_struct_gen]])),
     pytest.param(ArrayGen(string_gen),
                  marks=pytest.mark.xfail(reason="arrays are not supported")),
     pytest.param(MapGen(StringGen(pattern='key_[0-9]', nullable=False), simple_string_to_string_map_gen),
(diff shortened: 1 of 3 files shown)


sperlingxx:
build

integration_tests/src/main/python/sort_test.py
```diff
@@ -228,12 +229,12 @@ def test_large_orderby(data_gen, stable_sort):
 # This is similar to test_large_orderby, but here we want to test some types
 # that are not being sorted on, but are going along with it
 @pytest.mark.parametrize('data_gen', [byte_gen,
-    string_gen,
-    float_gen,
-    date_gen,
-    timestamp_gen,
-    decimal_gen_default,
-    StructGen([('child1', byte_gen)]),
+    # string_gen,
```
firestarman, line 232: If they were no longer needed, better to remove them.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1916,11 +1916,11 @@ object GpuOverrides {
     expr[SortOrder](
       "Sort order",
       ExprChecks.projectOnly(
-        pluginSupportedOrderableSig,
+        pluginSupportedOrderableSig + TypeSig.STRUCT.nested(),
```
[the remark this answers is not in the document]
revans2: That check already exists below on line 1932

integration_tests/src/main/python/sort_test.py
revans2: Agree if we don't think we need these any more then we should delete them instead of commenting them out. But that said are we removing this because of performance? If so then you could refactor it so there is a single test with one column for each type we want to test and do a single join. Was it commented out just because you were testing? It would be nice to know why these tests are being removed so we can make a proper decision on how to deal with it.

integration_tests/src/main/python/sort_test.py
```diff
@@ -40,8 +40,7 @@ def test_single_orderby(data_gen, order):
 @pytest.mark.parametrize('stable_sort', ['STABLE', 'OUTOFCORE'])
 @pytest.mark.parametrize('data_gen', [
     pytest.param(all_basic_struct_gen),
-    pytest.param(StructGen([['child0', all_basic_struct_gen]]),
```
revans2, line 43: In the original code we didn't add any tests that would check that we fall back to the CPU on structs that have a mismatch between the null ordering, it would be nice if you could add it here.

integration_tests/src/main/python/sort_test.py
sperlingxx: These tests are necessary. I just forgot to uncomment them after the debugging work on Array of Byte. I am sorry for making you confused  :(

sperlingxx:
build

integration_tests/src/main/python/sort_test.py
sperlingxx: I've added a case to test fallback due to non-default null ordering.

jlowe (approved):

revans2 (approved):

PR #3039, opened 2021-07-27
Update docs for google analytics

Salonijain27:
Updates _config.yml file and adds _include/analytics.html file to set up google analytics tracker

### DIFF
--- a/_include/analytics.html
+++ b/_include/analytics.html
@@ -1,0 +1,11 @@
+<head>
+  <!-- Global site tag (gtag.js) - Google Analytics -->
+  <script async src="https://www.googletagmanager.com/gtag/js?id=UA-203084925-1"></script>
+  <script>
+    window.dataLayer = window.dataLayer || [];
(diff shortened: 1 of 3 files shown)


_include/analytics.html
```diff
@@ -0,0 +1,11 @@
+<head>
+  <!-- Global site tag (gtag.js) - Google Analytics -->
+  <script async src="https://www.googletagmanager.com/gtag/js?id=UA-203084925-1"></script>
```
sameerz, line 3: Please confirm the UA and GA tags are connected.

sameerz (approved):

_include/analytics.html
Salonijain27: Updated the code to include only GA tags

sameerz:
One more request - can you retarget this to 21.08, and then we can cherry pick this into the gh-pages branch?

Salonijain27:
Re-creating the PR with 21.08 as its base

Issue #3040, opened 2021-07-27
[BUG] prevent going above `vm.max_map_count` OS limit with UCX


abellina:
We saw an issue where q72 at 3TB with 8 GPUs was causing requests with thousands (~7K) of small blocks (sized around ~1KB), leading to us running above `max_map_count`.

Normally, the `RapidsShuffleServer` queues all copies it needs to do on the CUDA stream and then waits for a single CUDA synchronize call. If there are several requests that successfully acquired a bounce buffer, they all get handled, and then they all wait for that CUDA synchronize call. The problem isn't the waiting, but mostly the fact that each waiting copy as obtained a reference for a `RapidsBuffer`. If the `RapidsBuffer` is in the disk, it also means we are holding on to the `mmap`ed region as well.

Users can change `vm.max_map_count` to get more breathing room, they can also change the `spark.sql.shuffle.partitions` to a degree, but in reality we should handle it better.

One way to handle this is to coalesce the blocks written to disk (bigger files). We can then `mmap` larger regions (but keeping in mind that we don't want to `mmap` huge chunks either), additionally we can add better handling when an `mmap` failure does happen, so we can handle exceptions better, which the current code doesn't do well.

abellina:
On an issue like this, we are also leaving the other side waiting for the timeout. Adding the comment here, I may have to split that to a different issue, but we really ought to reply to the peer with an error message.

sameerz:
Adding the needs triage label to see if this needs to be raised to a P1.

PR #3049, opened 2021-07-27
Handle mmap exception more gracefully in RapidsShuffleServer

abellina:

Improves handling of a potential `IOException` when attempting to `mmap` in a system without resources, with many small shuffle blocks, or with low settings for `vm.max_map_count`. This is an improvement related to https://github.com/NVIDIA/spark-rapids/issues/3040, but it is not the full solution.

The root of the problem is that the spilled blocks can be many, and that each block is likely to get `mmap`ed when read or transmitted. When we have many small blocks, the access pattern in `RapidsShuffleServer` can create issues as documented in https://github.com/NVIDIA/spark-rapids/issues/3040.

With this code, I can get q72 at 3TB to fail to `mmap` when there is a surge of requests, but reattempt the `mmap` successfully. Note that this doesn't prevent other parts of the system to fail when we are close to the OS limits.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
@@ -151,7 +151,13 @@ class BufferSendState(
     }
   }
 
-  def tryGetBufferToSend(): MemoryBuffer = synchronized {
+  /**
+   * Prepares and returns a `MemoryBuffer` that can be used in a send.
+   * @return - a memory buffer slice backed by either a host or device bounce buffer, depending on
+   *         the tier location for buffers we are sending.
+   * @throws `RapidsShuffleSendPrepareException` when copies to the bounce buffer fail.
+   */
+  def getBufferToSend(): MemoryBuffer = synchronized {
     require(acquiredBuffs.isEmpty,
       "Called next without calling `releaseAcquiredToCatalog` first")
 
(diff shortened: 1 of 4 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
```diff
@@ -148,7 +151,7 @@ class BufferSendState(
     }
   }
 
-  def next(): MemoryBuffer = synchronized {
+  def tryGetBufferToSend(): MemoryBuffer = synchronized {
```
jlowe, line 154: Nit: This name is confusing.  Normally a method with `try` or `maybe` in the name returns a boolean, `Option`, or some other indicator that it failed.  If the method simply throws an exception, well there's lots of methods that can throw exceptions but we don't put `try` in the method name.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/shuffle/BufferSendState.scala
```diff
@@ -185,11 +188,32 @@ class BufferSendState(
...
+              blockRange.rangeSize(), serverStream)
+            buffOffset += blockRange.rangeSize()
+          }
+          needsCleanup = false
+        } catch {
+          case ioe: IOException =>
+            val ex = new RapidsShuffleSendPrepareException(
+              s"Error while copying to bounce buffer for executor ${peerExecutorId} and " +
+                  s"header ${TransportUtils.toHex(peerBufferReceiveHeader)}")
+            ex.addSuppressed(ioe)
```
jlowe, line 209: Why is this a suppressed exception instead of a caused exception?  Suppressed exceptions are ones that occur while processing another exception, but the RapidsShuffleSendPrepareException is caused by the IOException.

sql-plugin/src/main/scala/org/apache/spark/shuffle/RapidsShuffleExceptions.scala
```diff
@@ -31,3 +31,10 @@ class RapidsShuffleFetchFailedException(
 }
 
 class RapidsShuffleTimeoutException(message: String) extends Exception(message)
+
+/**
+ * Internal exception thrown by `BufferSendState` in case where it detects
+ * an `IOException` when copying buffers into a bounce buffer that it is preparing.
+ * @param message - string describing the issue
+ */
+class RapidsShuffleSendPrepareException(message: String) extends Exception(message)
```
jlowe, line 40: This class should support a cause exception and delegate it to the underlying `Exception`.

abellina:
Thanks @jlowe. I added a couple of commits that should address the review comments.

jlowe (approved):

jlowe:
build

PR #3056, opened 2021-07-28
Generate supported operations doc for each version of Spark supported by the plugin

### LINKED ISSUE #1977: [DOC] Supported operations doc does not include features tied to specific Spark versions
The supported operations document that is automatically generated from the code does not include features that require a minimum Spark version beyond the base supported version.  For example, the ParquetCachedBatchSerializer requires Spark 3.1 and is not covered in the document to describe which Spark types are supported by that feature.

razajafri:
The code from `dist/pom.xml` that generates the documentation has been moved to each shim layer which generates its own `supported_ops.md` and `configs.md`

Thank you @tgravescs for helping with Databricks. Can you PTAL. 

Fixes #1977


jenkins/databricks/build.sh
```diff
@@ -108,6 +109,14 @@ mvn -B install:install-file \
    -Dversion=$SPARK_VERSION_TO_INSTALL_DATABRICKS_JARS \
    -Dpackaging=jar
 
+mvn -B install:install-file \
```
tgravescs, line 112: can you add a comment saying required for SPARK_VERSION

shims/spark311cdh/src/main/scala/com/nvidia/spark/rapids/shims/spark311cdh/SparkShimServiceProvider.scala
```diff
@@ -20,7 +20,7 @@ import com.nvidia.spark.rapids.{ClouderaShimVersion, SparkShims}
 
 object SparkShimServiceProvider {
   val VERSION = ClouderaShimVersion(3, 1, 1, "3.1.7270")
-  val VERSIONNAMES = Seq(s"3.1.1.3.1.7270")
+  val VERSIONNAMES = Seq(s"3.1.1.3.1.7270.0-253")
```
tgravescs, line 23: why this change?

tgravescs (commented):
how do we reference all the difference versions of the docs configs/supported ops from top level docs?  I probably missed it but didn't open all of them up.

docs/configs.md
```diff
@@ -1,381 +0,0 @@
----
```
jlowe, line 1: There are a lot of existing hyperlinks to this document, and I don't think this should be removed.  Instead there can be a set of supplemental configs documents covering the shims that we can link to from this common document.  I think we would want the same for the supported_ops document, since we don't currently have plans to change what types are supported for a common operation based on which Spark version is being used.

sameerz:
Please modify so the docs generated include a top level config.md and supported_ops.md which are parent level nav pages for the version specific configs and supported_ops md.  Otherwise each page is being generated in our gh-pages at the top level with the same name. 

Screen Shot 2021-07-28 at 10 04 52 AM

sameerz (changes requested):
Please modify so that the gh-pages renders properly.

shims/spark311cdh/src/main/scala/com/nvidia/spark/rapids/shims/spark311cdh/SparkShimServiceProvider.scala
razajafri: Without this change, the Shim for CDH is never loaded as the version in the pom.xml is `3.1.1..3.1.7270.0-253` which doesn't match `3.1.1.3.1.7270` so `SparkShimServiceProvider.VERSIONNAMES.contains(version)` returns false. 

I don't know how it was working previously

docs/configs.md
razajafri: I wrongly thought git would forward the symlinks to the actual file. 

I have made the change now to generate the docs in two places, `docs` and `docs/SPARK_DEFAULT_VERSION`.

jlowe: But we don't want almost full copies of these docs in each shim, right?  Especially if many places are going to link to the "common" doc, how will users know which version of the doc they're supposed to reference?  Having duplicates like this is very confusing, IMO, and ideally the shims should only contain the deltas from the common document with the common document indicating Spark-version specific extensions are covered in these shim docs (ideally with hyperlinks).

razajafri: If the shims-docs contains deltas, won't that be even more confusing for the user? Like e.g. if a user is looking at the default version, he will have to go to the "version-specific" doc to see if anything is different in his version of spark. IMHO the way it is right now avoids that problem, as the user can just go to the "version-specific" doc and know exactly what execs are accelerated.

We just have to handle the linking problem. Right now we can modify the incoming links to point to the default version i.e. "3.0.1".

jlowe: > Right now we can modify the incoming links to point to the default version i.e. "3.0.1".

But then there's no indication to the user that the document they're looking at does not apply to their Spark version.

As I mentioned before, we have a core set of configs and supported operators and types, that should be a common document.  It's easy to link to and there's no redundant duplication.  Shim docs can have any extra operators and types that apply to that Spark version.  Arguably this shouldn't be shim-specific either, otherwise we get a bunch of redundant docs for ParquetCacheBatchSerializer, for example.  I think of it as a "base Spark version" thing rather than something shim-specific, but I'm curious what others think.  Personally I don't want to see 8 copies of the docs and all the churn on those files every time someone touches a typecheck or exec meta.

revans2: This is not a simple problem.  Because of how the data is stored to generate these docs we need a version of Spark to run this against (or at least compile the code against before we generate the docs).  That means unless you log onto a data bricks server that config is not being updated. It also means that there is no good way to pick through the different parts of things and generate "deltas". We override one map entry with another in many cases, how do we indicate that in the docs? How do we store it in the original code so we can indicate it in the docs?

razajafri: >But then there's no indication to the user that the document they're looking at does not apply to their Spark version.

How is that different from today? Let's say the user is on Spark 3.1.1 today and he clicks on the link, unless I am missing something, he is taken to the default version without even mentioning that the doc is only relevant for version 3.0.1? 

This is obviously a more complicated problem than we originally thought, at least I did. Can we agree to a solution that will at least document everything. The next step can be to make it better. I will leave it to others to decide how much time we should spend on this problem. 

@revans2 I was thinking we could possibly just document the `commonExecs` instead of documenting the `exec` from GpuOverrides, that way we can only get the execs that are overridden in that specific version only for the "deltas", but again, this will be more work and I would love to do the "deltas" stuff, if it's even possible, as a follow-on. 

Thoughts?

revans2: If we are going to rewrite things lets get a design first on what we actually want this to look like. The we can have a discussion on how to make it happen, and what limitations there are with github pages that might prevent us from doing specific things. So the first thing I would ask for you @razajafri is to come up with some kind of a proposal on how we want this to look, and how it will interact with the other doc pages. Then once we all agree on that we can start to change the code.

razajafri:
This PR is no longer valid as their will be a design discussion before we proceed

Issue #3060, opened 2021-07-28
[BUG] ORC read can corrupt data when specified schema does not match file schema ordering


wbo4958:
@firestarman has found the same issue with https://github.com/NVIDIA/spark-rapids/issues/3007 for the special schema which can't be pruned. I can repro it with the below code. 

``` scala
      val df = Seq(Testing(1, "hello", 2021)).toDF
      df.printSchema()
      // root
      // |-- _col1: integer (nullable = false)
      // |-- _col2: string (nullable = true)
      // |-- _col3: long (nullable = false)
      df.show()
      // +-----+-----+-----+
      //|_col1|_col2|_col3|
      //+-----+-----+-----+
      //|    1|hello| 2021|
      //+-----+-----+-----+
      df.write.mode("overwrite").orc(resource1)

      val schema = StructType(
        Seq(
          StructField("_col2", StringType),
          StructField("_col3", LongType),
          StructField("_col1", IntegerType),
          ))
      val dfRead = spark.read.schema(schema).orc(resource1)
      dfRead.show()
```

The GPU output is
``` console
+-----+-----+-----+
|_col2|_col3|_col1|
+-----+-----+-----+
|     |    1|    5|
+-----+-----+-----+
```

while the CPU output is

``` console
+-----+-----+-----+
|_col2|_col3|_col1|
+-----+-----+-----+
|    1| null| 2021|
+-----+-----+-----+
```

Looks like there also is an issue for CPU reading ORC

PR #3062, opened 2021-07-28
Fix ORC read corruption when specified schema does not match file order

wbo4958:
This PR first fixes the orc data reading mess up for the schema which can't be pruned
second, adds the dis-order read schema unit tests for both Parquet and Orc.

Please refer to issue #3060. 

With this PR, GPU can produce the correct result but The CPU itself has produced the wrong result. So this PR will bring the in-compatible behavior between GPU and CPU.


wbo4958:
build

wbo4958:
build

tests/src/test/scala/com/nvidia/spark/rapids/OrcScanSuite.scala
```diff
@@ -33,4 +34,66 @@ class OrcScanSuite extends SparkQueryCompareTestSuite {
...
+      StructField("c1_int", IntegerType))))) { frame => frame }
+
+  testSparkResultsAreEqual("schema-can-prune dis-order read schema 5",
+    frameFromOrcWithSchema("schema-can-prune.orc", StructType(Seq(
+      StructField("c3_long", LongType),
+      StructField("c1_int", IntegerType))))) { frame => frame }
+
+  /**
+   * We can't compare the results from CPU and GPU, since CPU will get in-correct result
+   * see https://github.com/NVIDIA/spark-rapids/issues/3060
```
jlowe, line 73: IIUC the "can't be pruned" case is when the column names appear to be the same as the case when ORC doesn't provide column names (which was true for old versions of ORC in the early Hive days).  In those cases, I don't think Spark can fix this because there aren't any column names to match up if the specified schema doesn't match the file order.

jlowe (approved):
@wbo4958 can you elaborate in the PR description about the binary files that were added?

tests/src/test/scala/com/nvidia/spark/rapids/OrcScanSuite.scala
wbo4958: Yes, you're correct. So if it's not a Spark bug, Should I merge this PR? @jlowe

firestarman: Per your test case, even without this fix, the CPU and GPU still produce different outputs. So personally I prefer to merge it.

jlowe: Agreed, this seems to not break anything and improves behavior in a corner case.

Issue #3066, opened 2021-07-28
[BUG] Profiling tool generate dot file fails to convert


tgravescs:
**Describe the bug**

Running profiling tool to generate the dot graph with: 

`$SPARK_HOME/bin/spark-submit --master local[*] --driver-memory 30g --class com.nvidia.spark.rapids.tool.profiling.ProfileMain ./tools/target/rapids-4-spark-tools_2.12-21.08.0-SNAPSHOT.jar   -g application_1616746343401_0025
`
Then try to convert it to pdf or svg both fail:

```
$ dot -Tpdf  application_1616746343401_0025-query-0.dot > app.pdf
Error: not well-formed (invalid token) in line 95
...
                                    :           :              +- GpuFileGpuScan parquet default.insight_baiduapp_follow_history[cuid#5,uid#6,third_id#7,type#8,ext#13,event_day#14] Batched: true, DataFilters: [isnotnull(type#8), isnotnull(cuid#5), NOT (cuid#5 = ), (type#8 = ugc), (NOT (cast(ext#13[fake] a..., Format: Parquet, Location: InMemoryFileIndex[hdfs://bdbl-rpm-1106-57451.bdbl:9000/user/turing/spark/bbs/work/fengchao/ubs_da..., PartitionFilters: [isnotnull(event_day#14), (event_day#14 = 20201118)], PushedFilters: [IsNotNull(type), IsNotNull(cuid), Not(EqualTo(cuid,)), EqualTo(type,ugc)], ReadSchema: struct<cuid:string,uid:string,third_id:string,type:string,ext:map<string,string> ...
in label of graph G
```

tgravescs:
the issue is with label generation introduce dhere: https://github.com/NVIDIA/spark-rapids/commit/13b740c560914f7ffe39b1d0dc1368f915a4b700#diff-8dc61637872038b82e8dda72007efa65d71ff93aae90b312c7185abd320dbe87

It looks like certain characters in the plan string can mess it up.

tgravescs:
`ReadSchema: struct<_c1:string,_c2:string><br align="left"/>`

causes issues

PR #3068, opened 2021-07-28
Profiling tool generate dot file fails on unescaped html characters

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/3066

If the html in the label generated has less than or greater than signs like 
`ReadSchema: struct<_c1:string,_c2:string><br align="left"/>
`

The dot graph generation will fail.  Here we escape any html characters in the physical plan.


tgravescs:
build

gerashegalov (approved):
LGTM

PR #3069, opened 2021-07-28
Fall back to the CPU on window partition by struct or array

### LINKED ISSUE #3065: [BUG] window exec tries to do too much on the GPU
**Describe the bug**
Window exec does not fall back to the CPU when the partition by keys are nested. At least in the case of structs cudf does not complain about this and produces an incorrect result.

revans2:
This fixes #3065 

I will try and file an issue against cudf also because they just returned the wrong answer instead of failing on structs as a window key.


revans2:
build

jlowe (approved):

PR #3074, opened 2021-07-28
Partial support for time windows

### LINKED ISSUE #2943: [FEA] Support PreciseTimestampConversion when using windowing function
A clear and concise description of what the problem is. Ex. I wish the RAPIDS Accelerator for Apache Spark would [...]

I wish the RAPIDS Accelerator for Apache Spark would support PreciseTimestampConversion operator.

Below is an example to reproduce:
```
import datetime
from pyspark.sql import functions as F
from pyspark.sql.window import Window

df1 = sc.parallelize([(17,"2017-03-10 15:27:18.000000Z"), (13,"2017-03-15 12:27:18.000000Z"), (25,"2017-03-18 11:27:18.000000Z")]).toDF(["dollars", "datestring"])
df2 = df1.withColumn('timestampGMT', df1.datestring.cast('timestamp'))
df2.show(truncate=False)

w = (Window.partitionBy(F.window("timestampGMT", "7 days")))
df4 = df2.withColumn('rolling_average', F.avg("dollars").over(w))
df4.show(truncate=False)
df4.explain()
```
It will fallback on CPU with below messages:
```
              !NOT_FOUND <PreciseTimestampConversion> precisetimestampconversion(((((CASE WHEN (cast(CEIL((cast((precisetimestampconversion(cast(datestring#158 as timestamp), TimestampType, LongType) - 0) as double) / 6.048E11)) as double) = (cast((precisetimestampconversion(cast(datestring#158 as timestamp), TimestampType, LongType) - 0) as double) / 6.048E11)) THEN (CEIL((cast((precisetimestampconversion(cast(datestring#158 as timestamp), TimestampType, LongType) - 0) as double) / 6.048E11)) + 1) ELSE CEIL((cast((precisetimestampconversion(cast(datestring#158 as timestamp), TimestampType, LongType) - 0) as double) / 6.048E11)) END + 0) - 1) * 604800000000) + 604800000000), LongType, TimestampType) cannot run on GPU because no GPU enabled version of expression class org.apache.spark.sql.catalyst.expressions.PreciseTimestampConversion could be found
```

Above query can run on GPU and not fallback.

**Additional context**
Add any other context, code examples, or references to existing implementations about the feature request here.

revans2:
This fixes #2943 

Spark supports time windows, and technically this lets us also support them, but without support for grouping by a struct of (timestamp, timestamp) or partitioning by this on window operations it is not going to do a lot for actual customers.

Time windows are supported by the `window` [function](https://github.com/apache/spark/blob/eb4d1c03326cf84d5588023d17744f1d81faf21f/sql/core/src/main/scala/org/apache/spark/sql/functions.scala#L3496-L3549). This essentially does a bunch of math operations along with an expand to produce a struct column with the timestamp bucketed into a number of different time window buckets.  You can then group by these buckets or do window operations on them to get either tumbling or sliding window aggregations. This adds some tests that are in the context of those more complete operations, because just creating the bucket, though interesting, is not that useful on its own.  When we do support groupby and window operations on structs then this can be extended.

### DIFF
--- a/integration_tests/src/main/python/data_gen.py
+++ b/integration_tests/src/main/python/data_gen.py
@@ -868,9 +868,9 @@ date_n_time_gens = [date_gen, timestamp_gen]
 
 boolean_gens = [boolean_gen]
 
-single_level_array_gens = [ArrayGen(sub_gen) for sub_gen in all_basic_gens + decimal_gens + [null_gen]]
+single_level_array_gens = [ArrayGen(sub_gen) for sub_gen in all_basic_gens + decimal_gens]
 
-single_level_array_gens_no_decimal = [ArrayGen(sub_gen) for sub_gen in all_basic_gens + [null_gen]]
+single_level_array_gens_no_decimal = [ArrayGen(sub_gen) for sub_gen in all_basic_gens]
 
 map_string_string_gen = [MapGen(StringGen(pattern='key_[0-9]', nullable=False), StringGen())]
 
--- a/integration_tests/src/main/python/time_window_test.py
+++ b/integration_tests/src/main/python/time_window_test.py
@@ -29,7 +29,7 @@ _restricted_ts_gen = TimestampGen(start=_restricted_start, end=_restricted_end)
 
 # Once we support grouping by a struct (even single level) this should go away
 # https://github.com/NVIDIA/spark-rapids/issues/2877
-# Shuffle falls back to CPU becasue it is in between two CPU hash/sort aggregates
+# Shuffle falls back to CPU because it is in between two CPU hash/sort aggregates
 @allow_non_gpu('HashAggregateExec', 'SortAggregateExec', 'AggregateExpression', 'Max', 'Alias', 'ShuffleExchangeExec', 'HashPartitioning')
 @pytest.mark.parametrize('data_gen', integral_gens + [string_gen], ids=idfn)
 @ignore_order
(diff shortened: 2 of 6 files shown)


revans2:
build

integration_tests/src/main/python/time_window_test.py
```diff
@@ -0,0 +1,98 @@
...
+from pyspark.sql.window import Window
+
+# do it over a day so we have more chance of overlapping values
+_restricted_start = datetime(2020, 1, 1, tzinfo=timezone.utc)
+_restricted_end = datetime(2020, 1, 2, tzinfo=timezone.utc)
+_restricted_ts_gen = TimestampGen(start=_restricted_start, end=_restricted_end)
+
+# Once we support grouping by a struct (even single level) this should go away
+# https://github.com/NVIDIA/spark-rapids/issues/2877
+# Shuffle falls back to CPU becasue it is in between two CPU hash/sort aggregates
```
jlowe, line 32:
```suggestion
# Shuffle falls back to CPU because it is in between two CPU hash/sort aggregates
```

This suggestion also applies to all the copies below.

integration_tests/src/main/python/time_window_test.py
```diff
@@ -0,0 +1,98 @@
...
+@allow_non_gpu('HashAggregateExec', 'SortAggregateExec', 'AggregateExpression', 'GetArrayItem', 'Literal', 'Max', 'Alias', 'ShuffleExchangeExec', 'HashPartitioning')
+@pytest.mark.parametrize('data_gen', integral_gens + [string_gen], ids=idfn)
+@ignore_order
+def test_grouped_tumbling_window_array(data_gen):
+    row_gen = StructGen([['ts', _restricted_ts_gen],['data', ArrayGen(data_gen)]], nullable=False)
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : gen_df(spark, row_gen).groupBy(f.window('ts', '5 hour')).agg(f.max(f.col("data")[3]).alias("max_data")))
+
+# Warning. On Sliding windows is it easy to make lots of overlapping windows. This can make the Spark code generation
+# have some real problems and even crash some times when trying to JIT it. This problem only happnes on the CPU
```
jlowe, line 54:
```suggestion
# have some real problems and even crash some times when trying to JIT it. This problem only happens on the CPU
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3047,7 +3059,10 @@ object GpuOverrides {
       (sort, conf, p, r) => new GpuSortMeta(sort, conf, p, r)),
     exec[ExpandExec](
       "The backend for the expand operator",
-      ExecChecks(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL, TypeSig.all),
+      ExecChecks(
+        (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL +
+            TypeSig.STRUCT + TypeSig.ARRAY).nested(),
```
jlowe, line 3064: If we're in the process of adding nested struct and array support, does adding MapType support make sense?

revans2:
build

revans2:
@jlowe I think I have addressed all of your review comments.

integration_tests/src/main/python/data_gen.py
```diff
@@ -868,9 +868,9 @@ def gen_scalars_for_sql(data_gen, count, seed=0, force_no_nulls=False):
 
 boolean_gens = [boolean_gen]
 
-single_level_array_gens = [ArrayGen(sub_gen) for sub_gen in all_basic_gens + decimal_gens + [null_gen]]
```
revans2, line 871: FYI null_gen was already a part of all_basic_gens so we would double up a few tests.

jlowe (approved):

revans2:
build

revans2:
build got stuck and I aborted it. Not sure what happened. The logs didn't show any errors beyond `Could not connect to ... to send interrupt signal to process` which makes me think it was a CI system issue, but I am not 100% sure.

PR #3089, opened 2021-07-29
Add filter node in the query plan for conditional joins

jlowe:
This refactors the implementation of conditional inner joins.  The post-join filter is now an explicit node in the query plan rather than an implementation detail of the join node.  This has the advantage of adding visibility into how many rows were manifested from the join and subsequently filtered.

Despite joins no longer using the conditional expression after this PR, I left the condition parameter in place for now.  When AST conditional join support is added to the plugin, we can use this condition parameter to convey an expression that should be translated to an AST and processed directly as part of the join rather than as a post-filter.


jlowe:
build

revans2 (approved):

PR #3099, opened 2021-07-30
Decimal_128 type checks

revans2:
This updates the type checks to be able to support 128-bit decimal. It mostly updates makes a distinction between decimal 64 support and decimal 128 support.  It also updates how we report both decimal and timestamp incompatibility in the docs and decimal incompatibility in the query explain.  The former is probably the biggest change and I would really like feedback on how it looks. This does cause a lot of merge conflicts so it might be best to hold off on this until 21.08 is really done.


revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -415,7 +505,12 @@ object TypeSig {
   val DATE: TypeSig = new TypeSig(TypeEnum.ValueSet(TypeEnum.DATE))
   val TIMESTAMP: TypeSig = new TypeSig(TypeEnum.ValueSet(TypeEnum.TIMESTAMP))
   val STRING: TypeSig = new TypeSig(TypeEnum.ValueSet(TypeEnum.STRING))
-  val DECIMAL: TypeSig = new TypeSig(TypeEnum.ValueSet(TypeEnum.DECIMAL))
+  val DECIMAL_64: TypeSig =
+    new TypeSig(TypeEnum.ValueSet(TypeEnum.DECIMAL), DType.DECIMAL64_MAX_PRECISION)
+  val DECIMAL_128_NO_OVERFLOW: TypeSig =
+    new TypeSig(TypeEnum.ValueSet(TypeEnum.DECIMAL), DecimalType.MAX_PRECISION - 1)
+  val DECIMAL_128_FULL: TypeSig =
```
jlowe, line 512: I'm a bit confused on the semantics for these, and I think a comment could help.  For example, is DECIMAL_128_NO_OVERFLOW covering cases where we need to check and prevent silent overflow or where we don't need to check?

revans2: I thought about removing it entirely until we are ready to start using it. But then `DECIMAL_128_FULL` felt a little odd.  I will clean it up.

jlowe (approved):

jlowe:
build

kuhushukla (approved):

revans2:
build

jlowe (dismissed):

jlowe (approved):

razajafri (dismissed):

revans2:
build

revans2:
Hopefully this is the last time I am going to need to upmerge. If the build passes before something else gets merged in that causes a conflict I will merge this.

jlowe (approved):

PR #2223, opened 2021-04-21
Adds subquery aggregate tests from SPARK-31620

abellina:
Closes https://github.com/NVIDIA/spark-rapids/issues/1898.

This is being added to incorporate sub-query aggregate tests added in SPARK-31620 (Spark 3.0.0). These later showed failures in SPARK-32031 (3.1.0), but I can't reproduce the failure with our gpu aggregate.

The tests are being added so we exercise the sub-query aggs going forward.

PR #3105, opened 2021-07-30
Add filter in query plan for conditional nested loop and cartesian joins

jlowe:
This is the same type of change as in #3089 applied to broadcast nested loop and cartesian joins.  As in the other PR, the join condition parameter was left as a placeholder until the AST joins are leveraged.

### DIFF
--- a/integration_tests/src/main/python/join_test.py
+++ b/integration_tests/src/main/python/join_test.py
@@ -241,7 +241,7 @@ def test_broadcast_join_right_table_with_job_group(data_gen, join_type):
 @ignore_order(local=True)
 @pytest.mark.parametrize('data_gen', all_gen + single_level_array_gens, ids=idfn)
 @pytest.mark.parametrize('batch_size', ['100', '1g'], ids=idfn) # set the batch size so we can test multiple stream batches
-def test_cartesean_join(data_gen, batch_size):
+def test_cartesian_join(data_gen, batch_size):
     def do_join(spark):
         left, right = create_df(spark, data_gen, 50, 25)
         return left.crossJoin(right)
@@ -256,7 +256,7 @@ def test_cartesean_join(data_gen, batch_size):
     reason='https://github.com/NVIDIA/spark-rapids/issues/334')
 @pytest.mark.parametrize('data_gen', all_gen + single_level_array_gens, ids=idfn)
 @pytest.mark.parametrize('batch_size', ['100', '1g'], ids=idfn) # set the batch size so we can test multiple stream batches
-def test_cartesean_join_special_case_count(data_gen, batch_size):
+def test_cartesian_join_special_case_count(data_gen, batch_size):
     def do_join(spark):
         left, right = create_df(spark, data_gen, 500, 250)
         return left.crossJoin(right).selectExpr('COUNT(*)')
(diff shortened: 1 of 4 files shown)


jlowe:
build

revans2 (approved):

revans2:
The build failure is odd. It looks like a mismatch with cudf because it looks like it crashed/exited on the very first test that touched cudf.

jlowe:
> The build failure is odd.

Agreed.  It looks like cudf somehow is not in the classpath or is missing the properties file?
```
21/07/30 18:10:38.066 ScalaTest-main-running-HashAggregatesSuite ERROR RapidsExecutorPlugin: Exception in the executor plugin
com.nvidia.spark.rapids.CudfVersionMismatchException: Could not find cudf version in rapids4spark-version-info.properties
	at com.nvidia.spark.rapids.RapidsExecutorPlugin.$anonfun$checkCudfVersion$2(Plugin.scala:223)
	at scala.Option.getOrElse(Option.scala:138)
	at com.nvidia.spark.rapids.RapidsExecutorPlugin.checkCudfVersion(Plugin.scala:222)
	at com.nvidia.spark.rapids.RapidsExecutorPlugin.init(Plugin.scala:190)
	at org.apache.spark.internal.plugin.ExecutorPluginContainer.$anonfun$executorPlugins$1(PluginContainer.scala:111)
[...]
```

jlowe:
build

jlowe:
> It looks like cudf somehow is not in the classpath or is missing the properties file?

Sorry, I misread the stack exception.  This is a case of the _plugin_ properties file being present but somehow not containing the `cudf_version` key.   Very bizarre.

@pxLi it looks like a unit test run and `mvn verify` were running in parallel during the build?  The unit tests require the shaded jar to be built before the tests can run.  I'm not sure if the racing builds could lead to the error I saw above in premerge build #2223, but it seems to have gotten past it in #2224.

pxLi:
> > It looks like cudf somehow is not in the classpath or is missing the properties file?
> 
> Sorry, I misread the stack exception. This is a case of the _plugin_ properties file being present but somehow not containing the `cudf_version` key. Very bizarre.
> 
> @pxLi it looks like a unit test run and `mvn verify` were running in parallel during the build? The unit tests require the shaded jar to be built before the tests can run. I'm not sure if the racing builds could lead to the error I saw above in premerge build 2223, but it seems to have gotten past it in 2224.

hmm, this looks weird. The shaded ja should be built w/ the first mvn test run. 
Also we found some ephemeral NFS issue around that moment. I will keep monitoring this to see if this could occur again

jlowe:
> The shaded ja should be built w/ the first mvn test run.

`mvn test` doesn't produce the shaded jar, you need to run `mvn package` to do that.

pxLi:
> > The shaded ja should be built w/ the first mvn test run.
> 
> `mvn test` doesn't produce the shaded jar, you need to run `mvn package` to do that.

I may misunderstood, the test fail should be related to the blossom NFS issue and SRE is working on fixing it.
CMIW for the shaded jars, looks like our UTs can pass w/o the specific rapids.shade.packages. If this is required, I could enable it before the UT starts

PR #3109, opened 2021-08-02
Cudf agg type changes

### LINKED ISSUE #3101: [FEA] update cudf and plugin for separated aggregations
cudf is in the process of tagging aggregations for what operations they support using multiple inheritance.  We did an initial round of this for rolling window aggregations, and now there is work for some of the other aggregations.  https://github.com/rapidsai/cudf/pull/8906

The rolling window code was difficult to make work because java does not really have multiple inheritance and we have sub class/wrapper classes of these aggregations for different operations, like windows vs groupby/etc. I think it would just be cleaner if we completely redid it so that all of our aggregation classes are separate based off of the type of aggregation you want to do, scan vs reduction vs groupby scan ....  In the Spark code we already have to separate them almost entirely so lets just make it a cleaner more java like separation.  Or we remove the window separation and just keep it all based off of dynamic checking.

revans2:
This fixes #3101 

This depends on https://github.com/rapidsai/cudf/pull/8919

CUDF has been in the process of trying to make using the wrong aggregation with the wrong API a compile error instead of a runtime error.   The first set of changes were for rolling windows and I ended up using generics and interfaces to try and replicate the same thing. This didn't work out as well as I had hoped and the code to use them ended up being more cumbersome than I wanted.  This is an attempt to try and fix that, but I am open to other suggestions.  This is a breaking change so this and https://github.com/rapidsai/cudf/pull/8919 need to go in together.


abellina (dismissed):
Changes look good to me and the api makes it clearer what's available in each situation.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -412,7 +412,7 @@ trait GpuWindowBaseExec extends UnaryExecNode with GpuExec {
  * groups those two together so we can have a complete picture of how to perform these types of
  * aggregations.
  */
-case class AggAndReplace(agg: Aggregation, nullReplacePolicy: Option[ReplacePolicy])
+case class AggAndReplace[T](agg: T, nullReplacePolicy: Option[ReplacePolicy])
```
jlowe, line 415: Nit: At first I expected this to have some kind of "is an aggregation" constraint, but then realized in the new cudf refactor there's no way to implement that.  Obviously not a must-have.

jlowe (approved):

revans2:
@jlowe could you take another look. I had to upmerge which included fixing some issues with the new collect_list and collect_set group by aggregations.

jlowe (dismissed):

revans2:
build

revans2:
It looks like the plugin build pulled in an old version of the cudf jar. I restarted the build and we are trying again

revans2:
build

revans2:
Nope I just forgot to check in the fix after I upmerged. I feel dumb. @jlowe please take another look, for real this time.

jlowe (approved):

Issue #3042, opened 2021-07-27
[FEA] Qualification tool: Add conjunction and disjunction filters.


nartal1:
**Is your feature request related to a problem? Please describe.**
These 2 filters to be added:

1. Conjunction -  As a user, I want to be able to apply multiple log filtering criteria and process only logs for which **all** are satisfied, so that, for example, I could select only logs for applications whose name included "NDS," that were submitted within the last week, **and** that were submitted by user "benchmarks".

2. Disjunction - As a user, I want to be able to apply multiple log filtering criteria and process only logs for which **any** are satisfied, so that, for example, I could select only logs for applications whose name included "NDS," that were submitted within the last week, **or** that were submitted by user "benchmarks"

PR #3112, opened 2021-08-02
Qualification tool: Add conjunction and disjunction filters

nartal1:
This PR  closes https://github.com/NVIDIA/spark-rapids/issues/3042

### DIFF
--- a/docs/additional-functionality/qualification-profiling-tools.md
+++ b/docs/additional-functionality/qualification-profiling-tools.md
@@ -198,6 +198,14 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
                                     use ~APPLICATION_NAME. i.e Select all event
                                     logs except the ones which have application
                                     name as the input string.
+      --any                         Apply multiple event log filtering criteria and process
+                                    only logs for which any condition is satisfied.
+                                    Example: <Filter1> <Filter2> <Filter3> --any -> result
+                                    is <Filter1> OR <Filter2> OR <Filter3>
+      --all                         Apply multiple event log filtering criteria and process
+                                    only logs for which all conditions are satisfied.
+                                    Example: <Filter1> <Filter2> <Filter3> --all -> result
+                                    is <Filter1> OR <Filter2> OR <Filter3>. Default is all=true.                              
   -f, --filter-criteria  <arg>      Filter newest or oldest N eventlogs based on
                                     application start timestamp, unique
                                     application name or filesystem timestamp.
@@ -224,15 +232,6 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
                                     filesystem timestamp). 100-oldest-filesystem
                                     (for processing oldest 100 event logsbased
                                     on filesystem timestamp).
-  -l, --logic-filter <arg>          Filter event logs where any or all filter criteria
-                                    based on application names. Default is all.
-                                    Example: <Filter1> <Filter2> <Filter3>
-                                    --logicFilter any -> processes logs for which  +
-                                    any condition is satisfied. i.e result is <Filter1> OR
-                                    <Filter2> OR <Filter3>.
-                                    <Filter1> <Filter2> <Filter3> --logicFilter all ->
-                                    processes logs for which all are conditions are satisfied.
-                                    i.e result is <Filter1> AND <Filter2> AND <Filter3>
   -m, --match-event-logs  <arg>     Filter event logs whose filenames contain
                                     the input string. Filesystem based filtering
                                     happens before any application based
--- a/docs/additional-functionality/qualification-profiling-tools.md
+++ b/docs/additional-functionality/qualification-profiling-tools.md
@@ -205,7 +205,7 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
       --all                         Apply multiple event log filtering criteria and process
                                     only logs for which all conditions are satisfied.
                                     Example: <Filter1> <Filter2> <Filter3> --all -> result
-                                    is <Filter1> OR <Filter2> OR <Filter3>. Default is all=true.                              
+                                    is <Filter1> AND <Filter2> AND <Filter3>. Default is all=true.
   -f, --filter-criteria  <arg>      Filter newest or oldest N eventlogs based on
                                     application start timestamp, unique
                                     application name or filesystem timestamp.
--- a/docs/additional-functionality/qualification-profiling-tools.md
+++ b/docs/additional-functionality/qualification-profiling-tools.md
@@ -224,6 +224,15 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
                                     filesystem timestamp). 100-oldest-filesystem
                                     (for processing oldest 100 event logsbased
                                     on filesystem timestamp).
+  -l, --logic-filter <arg>          Filter event logs where any or all filter criteria
+                                    based on application names. Default is all.
+                                    Example: <Filter1> <Filter2> <Filter3>
+                                    --logicFilter any -> processes logs for which  +
+                                    any condition is satisfied. i.e result is <Filter1> OR
+                                    <Filter2> OR <Filter3>.
+                                    <Filter1> <Filter2> <Filter3> --logicFilter all ->
+                                    processes logs for which all are conditions are satisfied.
+                                    i.e result is <Filter1> AND <Filter2> AND <Filter3>
   -m, --match-event-logs  <arg>     Filter event logs whose filenames contain
                                     the input string. Filesystem based filtering
                                     happens before any application based
--- a/tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
+++ b/tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
@@ -75,7 +75,12 @@ class AppFilterImpl(
     val filterAppName = appArgs.applicationName.getOrElse("")
     val filterCriteria = appArgs.filterCriteria.getOrElse("")
     val userName = appArgs.userName.getOrElse("")
-    val logicFilter = appArgs.logicFilter.getOrElse("all")
+    //val logicFilter = appArgs.logicFilter.getOrElse("all")
+    val logicFilter = if (appArgs.any()) {
+      "any"
+    } else {
+      "all"
+    }
 
     val appNameFiltered = if (appArgs.applicationName.isSupplied && filterAppName.nonEmpty) {
       val filtered = if (filterAppName.startsWith(NEGATE)) {
@@ -92,7 +97,7 @@ class AppFilterImpl(
    // If `any` is specified as logicFilter config, then store the value in variable which is used
    // later for final computation.
    val appNameFilteredForDisjunction = if (appArgs.applicationName.isSupplied &&
-       filterAppName.nonEmpty && appArgs.logicFilter.isSupplied
+       filterAppName.nonEmpty && appArgs.any()
        && logicFilter.contains("any")) {
       appNameFiltered
     } else {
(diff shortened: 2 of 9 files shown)


docs/additional-functionality/qualification-profiling-tools.md
```diff
@@ -224,6 +224,15 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
                                     filesystem timestamp). 100-oldest-filesystem
                                     (for processing oldest 100 event logsbased
                                     on filesystem timestamp).
+  -l, --logic-filter <arg>          Filter event logs where any or all filter criteria
```
willb, line 227: Can we just call these options `--any` and `--all`?  `logic-filter` is likely to confuse users since it is unidiomatic and also not exactly correct (it's not a logic filter; it's an operation to combine several filters).  It's also a lot of keystrokes.

nartal1: Thanks @willb for your feedback. Updated the code.

nartal1:
build

docs/additional-functionality/qualification-profiling-tools.md
```diff
@@ -198,6 +198,14 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
...
                                     logs except the ones which have application
                                     name as the input string.
+      --any                         Apply multiple event log filtering criteria and process
+                                    only logs for which any condition is satisfied.
+                                    Example: <Filter1> <Filter2> <Filter3> --any -> result
+                                    is <Filter1> OR <Filter2> OR <Filter3>
+      --all                         Apply multiple event log filtering criteria and process
+                                    only logs for which all conditions are satisfied.
+                                    Example: <Filter1> <Filter2> <Filter3> --all -> result
+                                    is <Filter1> OR <Filter2> OR <Filter3>. Default is all=true.
```
andygrove, line 208: Shouldn't this be `<Filter1> AND <Filter2> AND <Filter3>` ?

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -75,6 +75,12 @@ class AppFilterImpl(
     val filterAppName = appArgs.applicationName.getOrElse("")
     val filterCriteria = appArgs.filterCriteria.getOrElse("")
     val userName = appArgs.userName.getOrElse("")
+    //val logicFilter = appArgs.logicFilter.getOrElse("all")
```
andygrove, line 78: Can we remove the commented-out code?

docs/additional-functionality/qualification-profiling-tools.md
nartal1: Yes. Copied from above but forgot to update here. Thanks for noticing this.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -528,4 +528,146 @@ class AppFilterSuite extends FunSuite {
...
+
+  def startTimePeriod(startPeriod: String): Array[String] = {
+    Array("--start-app-time", startPeriod)
+  }
+
+  def userName(name: String): Array[String] = {
+    Array("--user-name", name)
+  }
+
+  def logicFilter(lFilter: String): Array[String] = {
```
tgravescs, line 636: not used, remove

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -528,4 +528,146 @@ class AppFilterSuite extends FunSuite {
       }
     }
   }
+
+  case class TestConjunctionAndDisjunction(
+      fileName: String, fsTime: Long, appName: String,
```
tgravescs, line 533: not really testing fsTime and by code we aren't applying any/all to it properly

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -528,4 +528,146 @@ class AppFilterSuite extends FunSuite {
...
+    TestConjunctionAndDisjunction("app-nds86-1", msDaysAgo(3), "nds86",
+      msDaysAgo(4), 1, "user1"),
+    TestConjunctionAndDisjunction("app-nds86-2", msDaysAgo(6), "nds86",
+      msWeeksAgo(1), 2, "user2"),
+    TestConjunctionAndDisjunction("app-nds86-3", msDaysAgo(18), "nds86",
+      msWeeksAgo(3), 3, "user3"))
+
+  test("Test disjunction all filters") {
+    testConjunctionAndDisjunction(appsNameConjunctionAndDisjunctionToTest,
+      filterCriteria("10-newest") ++ filterAppName("nds") ++
```
tgravescs, line 554: nds app name matches everthing make it more unique

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -87,8 +92,26 @@ class AppFilterImpl(
     } else {
       apps
     }
+
+   // If `any` is specified as logicFilter config, then store the value in variable which is used
+   // later for final computation.
+   val appNameFilteredForDisjunction = if (appArgs.applicationName.isSupplied &&
+       filterAppName.nonEmpty && appArgs.any()
+       && logicFilter.contains("any")) {
```
andygrove, line 100: What condition would lead to `appArgs.any()` being true but `logicFilter` not being set to "any"?

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -87,8 +92,26 @@ class AppFilterImpl(
...
     val userNameFiltered = if (appArgs.userName.isSupplied && userName.nonEmpty) {
-      val filtered = appNameFiltered.filter { app =>
+      val appNamelogicFiltered = if (appArgs.any() &&
+          logicFilter.contains("any")) {
+        apps
+      } else {
+        appNameFiltered
+      }
+
+      val filtered = appNamelogicFiltered.filter { app =>
```
andygrove, line 114: This could be simplified by replacing 

```scala
val filtered = appNamelogicFiltered.filter { app =>
  if (app.appInfo.isDefined) {
    app.appInfo.get.userName.contains(userName)
  } else {
    false
  }
}
filtered
```

with 

```scala
appNamelogicFiltered.filter(_.appInfo.exists(_.userName.contains(userName)))
```

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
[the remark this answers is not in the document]
andygrove: I think the `appArgs.applicationName.isSupplied` check is redundant here because we really just care about `filterAppName.nonEmpty` and it can only be non-empty if supplied? The same applies to some other similar checks.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -99,9 +122,16 @@ class AppFilterImpl(
...
     val appTimeFiltered = if (appArgs.startAppTime.isSupplied) {
       val msTimeToFilter = AppFilterImpl.parseAppTimePeriodArgs(appArgs)
-      val filtered = userNameFiltered.filter { app =>
+      val logicFiltered = if (appArgs.any()) {
+        apps
+      } else {
+        userNameFiltered
+      }
+
+      val filtered = logicFiltered.filter { app =>
```
andygrove, line 134: This can be simplified by replacing

```scala
val filtered = logicFiltered.filter { app =>
  val appStartOpt = app.appInfo.map(_.startTime)
  if (appStartOpt.isDefined) {
    appStartOpt.get >= msTimeToFilter
  } else {
    false
  }
}
filtered
```

with

```scala
logicFiltered.filter(_.appInfo.exists(_.startTime >= msTimeToFilter))
```

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -75,6 +75,11 @@ class AppFilterImpl(
     val filterAppName = appArgs.applicationName.getOrElse("")
     val filterCriteria = appArgs.filterCriteria.getOrElse("")
     val userName = appArgs.userName.getOrElse("")
+    val logicFilter = if (appArgs.any()) {
```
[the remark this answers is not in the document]
nartal1: We have 2 file based filters 1. based on fileName match and 2. N-newest/oldest. I don't think we need to make changes to the file based filter code. I will add test which will test N-newest-filesystem along with appFilters with `any` and `all`

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
nartal1: I will add tests for fsTime along with appFilters with `any/all`

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
nartal1: nds app name matches with 5 out of 7 input.  Rest 2 are `Nds`. It is not doing Regex match in this case.  Could you please clarify what you mean by more unique.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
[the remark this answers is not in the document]
tgravescs: yes so why don't we need to make changes.  The file based ones apply first right now and then the other user, app name, etc..  So if you specify any wouldn't you want it to match either file name or on an application name?  But if we don't change it will first filter only those filename and then apply any within those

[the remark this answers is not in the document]
nartal1: My understanding was we always do file based filters first followed by app based filters(i.e result of file based filter is input to app filters). But what you mentioned above makes sense for `any` case. Thanks for the use case. I will make the changes to reflect both file and app based filters.

tools/src/test/scala/com/nvidia/spark/rapids/tool/profiling/ApplicationInfoSuite.scala
```diff
@@ -379,7 +379,7 @@ class ApplicationInfoSuite extends FunSuite with Logging {
       s"$qualLogDir/dataset_eventlog"
     ))
 
-    val result = EventLogPathProcessor.processAllPaths(appArgs.filterCriteria.toOption,
+    val (result, allEvents) = EventLogPathProcessor.processAllPaths(appArgs.filterCriteria.toOption,
```
tgravescs, line 382: change allEvents to _ if not used, same with below allLogs to _ when not used

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
tgravescs: thats fine, I missed the Nds

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
nartal1: Thanks! This is easy to read.

tgravescs (approved):

tgravescs:
build

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
nartal1: It's redundant. Removed `logicFilter`.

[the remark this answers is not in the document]
nartal1: Yes, you are right. Not sure why I had both checks :) . Removed the redundant check.

andygrove (approved):

PR #3113, opened 2021-08-02
Fix order of operations when using mkString in typeConversionInfo

abellina:
Prior to this, the explain was broken when doing type conversions (this was found while reviewing: https://github.com/NVIDIA/spark-rapids/pull/2971). 

```
      *Exec <ObjectHashAggregateExec> will run on GPU
 ; T; h; e;  ; d; a; t; a;  ; t; y; p; e;  ; o; f;  ; f; o; l; l; o; w; i; n; g;  ; e; x; p; r; e; s; s; i; o; n; s;  ; w; i; l; l;  ; b; e;  ; c; o; n; v; e; r; t; e; d;  ; i; n;  ; G; P; U;  ; r; u; n; t; i; m; e; :; 
; S; e; t; (; C; o; n; v; e; r; t; e; d;  ; D; a; t; a; T; y; p; e;  ; o; f;  ; b; u; f;  ; f; r; o; m;  ; B; i; n; a; r; y; T; y; p; e;  ; t; o;  ; A; r; r; a; y; T; y; p; e; (; I; n; t; e; g; e; r; T; y; p; e; ,; f; a; l; s; e; ); ,;  ; b; e; c; a; u; s; e;  ; C; o; n; v; e; r; t; e; d;  ; B; i; n; a; r; y; T; y; p; e;  ; t; o;  ; A; r; r; a; y; T; y; p; e; (; I; n; t; e; g; e; r; T; y; p; e; ,; f; a; l; s; e; ); ,;  ; b; e; c; a; u; s; e;  ; )
```

This tweaks the explain so it's in one line, and does some minor fixes. The original function duplicated the message, but I think the code I have here is safe since the `typeMeta` passed looks to be at the same expression level, and contains the same data as caller. It would be good to get confirmation I understood this correctly.
```
           *Exec <ObjectHashAggregateExec> will run on GPU. The data type of following expressions will be converted in GPU runtime: buf#10: Converted BinaryType to ArrayType(IntegerType,false)
```

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
@@ -311,14 +311,14 @@ abstract class RapidsMeta[INPUT <: BASE, BASE, OUTPUT <: BASE](
         (cannotRunOnGpuBecauseOfSparkPlan || shouldThisBeRemoved) => "could " + replaceMessage
     case Some(v) if v.isEmpty => "will " + replaceMessage
     case Some(v) =>
-      noReplacementPossibleMessage(v mkString "; ")
+      noReplacementPossibleMessage(v.mkString("; "))
   }
 
   private def willBeRemovedInfo: String = shouldBeRemovedReasons match {
     case None => ""
     case Some(v) if v.isEmpty => ""
     case Some(v) =>
-      val reasons = v mkString "; "
+      val reasons = v.mkString("; ")
       s" but is going to be removed because $reasons"
   }
 
@@ -327,7 +327,7 @@ abstract class RapidsMeta[INPUT <: BASE, BASE, OUTPUT <: BASE](
     case Some(v) if v.isEmpty => ""
     case Some(v) =>
       "The data type of following expressions will be converted in GPU runtime: " +
-          (v mkString "; ")
+          (v.mkString("; "))
   }
 
   /**
@@ -811,7 +811,8 @@ class DataTypeMeta(
   def reasonForConversion: String = {
     val reasonMsg = (if (typeConverted) reason else None)
         .map(r => s", because $r").getOrElse("")
-    s"Converted ${wrapped.get} to ${dataType.get}" + reasonMsg
+    s"Converted ${wrapped.getOrElse("N/A")} to " +
+      s"${dataType.getOrElse("N/A")}" + reasonMsg
   }
 }
 
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
@@ -190,10 +190,9 @@ abstract class RapidsMeta[INPUT <: BASE, BASE, OUTPUT <: BASE](
   /**
    * Call this method to record information about type conversions via DataTypeMeta.
    */
-  final def addConvertedDataType(name: String, typeMeta: DataTypeMeta): Unit = {
-    val reason = s"Converted DataType of $name from ${typeMeta.wrapped.get} to " +
-        s"${typeMeta.dataType.get}, because ${typeMeta.reasonForConversion}"
-    typeConversionReasons.get.add(reason)
+  final def addConvertedDataType(expression: Expression, typeMeta: DataTypeMeta): Unit = {
+    typeConversionReasons.get.add(
+      s"$expression: ${typeMeta.reasonForConversion}")
   }
 
   /**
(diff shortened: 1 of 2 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -327,8 +326,8 @@ abstract class RapidsMeta[INPUT <: BASE, BASE, OUTPUT <: BASE](
     case None => ""
     case Some(v) if v.isEmpty => ""
     case Some(v) =>
-      " The data type of following expressions will be converted in GPU runtime:\n" +
-          v mkString "; "
+      "The data type of following expressions will be converted in GPU runtime: " +
+          (v mkString "; ")
```
andygrove, line 330: I know we don't have coding standards for things like this, but I think it would be better to have `v.mkString("; ")` here rather than `(v mkString "; ")` as recommended here - https://docs.scala-lang.org/style/method-invocation.html

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -809,8 +809,9 @@ class DataTypeMeta(
    * Returns the reason for conversion if exists
    */
   def reasonForConversion: String = {
-    val reasonMsg = (if (typeConverted) reason else None).getOrElse("")
-    s"Converted ${wrapped.get} to ${dataType.get}, because $reasonMsg"
+    val reasonMsg = (if (typeConverted) reason else None)
+        .map(r => s", because $r").getOrElse("")
+    s"Converted ${wrapped.get} to ${dataType.get}" + reasonMsg
```
andygrove, line 814: Can `wrapped` ever be `None` when this method is called?

abellina: Hmm, I am not quite sure, but good catch. I gave them a default value to prevent an issue here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
abellina: should be fixed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -327,7 +327,7 @@ abstract class RapidsMeta[INPUT <: BASE, BASE, OUTPUT <: BASE](
     case Some(v) if v.isEmpty => ""
     case Some(v) =>
       "The data type of following expressions will be converted in GPU runtime: " +
-          (v mkString "; ")
+          (v.mkString("; "))
```
andygrove, line 330: The outer parentheses should no longer be required because we're explicitly invoking mkString on v now?

abellina: Good eye, fixed.

abellina:
build

abellina:
@andygrove do you have any other comments for this PR?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -311,23 +310,23 @@ abstract class RapidsMeta[INPUT <: BASE, BASE, OUTPUT <: BASE](
...
       s" but is going to be removed because $reasons"
   }
 
   private def typeConversionInfo: String = typeConversionReasons match {
     case None => ""
     case Some(v) if v.isEmpty => ""
     case Some(v) =>
-      " The data type of following expressions will be converted in GPU runtime:\n" +
-          v mkString "; "
+      "The data type of following expressions will be converted in GPU runtime: " +
```
razajafri, line 328: string interpolator can be used here?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -828,8 +828,10 @@ class DataTypeMeta(
    * Returns the reason for conversion if exists
    */
   def reasonForConversion: String = {
-    val reasonMsg = (if (typeConverted) reason else None).getOrElse("")
-    s"Converted ${wrapped.get} to ${dataType.get}, because $reasonMsg"
+    val reasonMsg = (if (typeConverted) reason else None)
+        .map(r => s", because $r").getOrElse("")
+    s"Converted ${wrapped.getOrElse("N/A")} to " +
+      s"${dataType.getOrElse("N/A")}" + reasonMsg
```
razajafri, line 834: nit: same here we can use $reasonMsg inside the string

andygrove:
> @andygrove do you have any other comments for this PR?

Sorry for the delay. LGTM.

andygrove (approved):

PR #3144, opened 2021-08-04
Optimize AQE with Spark 3.2+ to avoid redundant transitions [databricks]

andygrove:
Spark PR https://github.com/apache/spark/pull/33624 added the capability to create adaptive plans with `supportsColumnar=true`.

This PR leverages this new capability to avoid redundant columnar-to-row / row-to-columnar transitions when possible.

Here is an example plan. Note that the only transition is at the root.

```
GpuColumnarToRow false
+- Execute GpuInsertIntoHadoopFsRelationCommand file:/tmp/AdaptiveQueryExecSuite$/AvoidTransitionOutput.parquet, false, com.nvidia.spark.rapids.GpuParquetFileFormat@1b632442, [path=/tmp/AdaptiveQueryExecSuite$/AvoidTransitionOutput.parquet], Overwrite, [a]
   +- AdaptiveSparkPlan isFinalPlan=true supportsColumnar=true
      +- == Final Plan ==
         GpuFileGpuScan parquet [a#7] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex[file:/tmp/AdaptiveQueryExecSuite$/AvoidTransitionInput.parquet], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<a:int>
      +- == Initial Plan ==
         FileScan parquet [a#7] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex[file:/tmp/AdaptiveQueryExecSuite$/AvoidTransitionInput.parquet], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<a:int>
```

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
@@ -16,14 +16,8 @@
 
 package com.nvidia.spark.rapids
 
-import java.lang.reflect.Method
-
 import scala.annotation.tailrec
 
-import com.nvidia.spark.rapids.shims.v2.ShimUnaryExecNode
-
-import org.apache.spark.rdd.RDD
-import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.{Ascending, Attribute, AttributeReference, Expression, InputFileBlockLength, InputFileBlockStart, InputFileName, SortOrder}
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution._
@@ -34,8 +28,7 @@ import org.apache.spark.sql.execution.datasources.v2.DataSourceV2ScanExecBase
 import org.apache.spark.sql.execution.exchange.{BroadcastExchangeLike, Exchange, ReusedExchangeExec}
 import org.apache.spark.sql.execution.joins.{BroadcastHashJoinExec, BroadcastNestedLoopJoinExec}
 import org.apache.spark.sql.rapids.{GpuDataSourceScanExec, GpuFileSourceScanExec, GpuInputFileBlockLength, GpuInputFileBlockStart, GpuInputFileName, GpuShuffleEnv}
-import org.apache.spark.sql.rapids.execution.{GpuBroadcastExchangeExec, GpuBroadcastToCpuExec, GpuCustomShuffleReaderExec, GpuHashJoin, GpuShuffleExchangeExecBase}
-import org.apache.spark.sql.vectorized.ColumnarBatch
+import org.apache.spark.sql.rapids.execution.{GpuBroadcastExchangeExec, GpuBroadcastExchangeExecBase, GpuBroadcastToCpuExec, GpuCustomShuffleReaderExec, GpuHashJoin, GpuShuffleExchangeExecBase}
 
 /**
  * Rules that run after the row to columnar and columnar to row transitions have been inserted.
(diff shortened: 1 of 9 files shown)


revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -82,48 +82,58 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
...
-      // because we need to return an operator that implements `BroadcastExchangeLike` or
-      // `ShuffleExchangeLike`. The coalesce step gets added back into the plan later on, in a
-      // future query stage that reads the output from this query stage. This is handled in the
-      // case clauses below.
-      e.withNewChildren(e.children.map(c => optimizeAdaptiveTransitions(c, Some(e))))
+      case GpuBringBackToHost(child) if parent.isEmpty =>
+        // This is hacky but we need to remove the GpuBringBackToHost from the final
+        // query stage, if there is one. It gets inserted by
+        // GpuTransitionOverrides.insertColumnarFromGpu around columnar adaptive
+        // plans when we are writing to Parquet on the GPU. It would be nice to avoid
```
jlowe, line 90: This isn't specific to Parquet, correct?  I assume this would apply to any format being written by the GPU.  Wonder if there are other plans not involving writing that would trigger it as well.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -82,48 +82,58 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
...
+        // inserting it in the first place but we just don't have enough context
+        // at the time GpuTransitionOverrides is applying rules.
+        child
+
+      // HostColumnarToGpu(RowToColumnarExec(..)) => GpuRowToColumnarExec(..)
+      case HostColumnarToGpu(r2c: RowToColumnarExec, goal) =>
+        optimizeAdaptiveTransitions(r2c.child, Some(r2c)) match {
+          case a: AdaptiveSparkPlanExec =>
+            // we hit this case when we have an adaptive plan wrapped in a write
+            // to Parquet on the GPU
```
jlowe, line 100: Specific to Parquet?

jlowe (commented):
Small comment nits but otherwise lgtm.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
andygrove: True. This is for columnar writes in general. I will update the comments.

tests/src/test/scala/com/nvidia/spark/rapids/AdaptiveQueryExecSuite.scala
```diff
@@ -342,16 +343,22 @@ class AdaptiveQueryExecSuite
       // the read should be an adaptive plan
       val adaptiveSparkPlanExec = TestUtils.findOperator(writeCommand.get,
         _.isInstanceOf[AdaptiveSparkPlanExec])
-          .get.asInstanceOf[AdaptiveSparkPlanExec]
+        .get.asInstanceOf[AdaptiveSparkPlanExec]
 
-      val transition = adaptiveSparkPlanExec
+      if (cmpSparkVersion(3, 2, 0) < 0) {
```
jlowe, line 348: This is likely to cause problems with at least the upcoming Databricks 8.3 shim, as that's an example of a shim that is _partially_ based on 3.2 code.  Since it's already shipped, I do not expect it to contain the fix that we're looking for with this check here.

andygrove: I have removed the version check and added a `supportsColumnarAdaptivePlans` method to the shim layer

jlowe (dismissed):

andygrove:
build

jlowe (commented):
It looks like this touches code that applies to all shims except the Databricks shims.  Do we need to update those as well?

andygrove:
build

andygrove:
build

andygrove:
There are a number of integration test failures on this branch so this needs more work. I am moving it back to draft for now.

andygrove:
build

andygrove:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -82,48 +75,61 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
...
-      // individual query with no parent and we need to remove these operators in this case
-      // because we need to return an operator that implements `BroadcastExchangeLike` or
-      // `ShuffleExchangeLike`. The coalesce step gets added back into the plan later on, in a
-      // future query stage that reads the output from this query stage. This is handled in the
-      // case clauses below.
-      e.withNewChildren(e.children.map(c => optimizeAdaptiveTransitions(c, Some(e))))
+      // adaptive plan final query stage with columnar output
+      case r2c @ RowToColumnarExec(child) if parent.isEmpty =>
+        val optimizedChild = optimizeAdaptiveTransitions(child, Some(r2c))
+        GpuRowToColumnarExec(optimizedChild, TargetSize(rapidsConf.gpuTargetBatchSizeBytes))
```
jlowe, line 106: Do we need to preserve a tag here as was done earlier?  If not, adding some comments to the case where we do preserve the tagging and why we only need to do it there would help.

andygrove: Honestly, I have no idea. It would seem consistent to do it here as well so will push that change but we have no tests that are affected one way or the other. Maybe @sperlingxx can help guide us here. The same question applies in the  `Spark30XShim` code in the `columnarAdaptivePlan` method.

sql-plugin/src/main/301until320-nondb/scala/com/nvidia/spark/rapids/shims/v2/Spark30XShims.scala
```diff
@@ -142,4 +142,14 @@ trait Spark30XShims extends SparkShims {
...
   override def getLegacyStatisticalAggregate(): Boolean = true
+
+  override def supportsColumnarAdaptivePlans: Boolean = false
+
+  override def columnarAdaptivePlan(a: AdaptiveSparkPlanExec, goal: CoalesceSizeGoal): SparkPlan = {
+    // When the input is an adaptive plan we do not get to see the GPU version until
+    // the plan is executed and sometimes the plan will have a GpuColumnarToRowExec as the
+    // final operator and we can bypass this to keep the data columnar by inserting
+    // the [[GpuAvoidAdaptiveTransitionToRow]] operator here
+    AvoidAdaptiveTransitionToRow(GpuRowToColumnarExec(a, goal))
```
andygrove, line 153: We are not checking for preProcessing tags here and I'm not sure if we should be in this case.

jlowe: Do we need to update the Databricks shims with the same change?  I see AvoidAdaptiveTransitionToRow class declared in the 301until320 v2 shim, but it seems to only be called in the non-db cases.  We don't support AQE on Databricks, but if that problem gets solved it would be nice to not be surprised with all the things we could have updated for AQE in Databricks shims but did not.

jlowe (commented):
Need to resolve merge conflict and a question about Databricks shims, but otherwise lgtm.

andygrove:
build

jlowe:
Curious why this was force-pushed?

jlowe (approved):

andygrove:
The shims have been refactored significantly recently and one of the shim
classes modified in the original PR had been deleted so I thought it would
be cleaner to re-apply the changes on top. Should I have persevered with
merging and fixing the conflicts in this case?

jlowe:
> Should I have persevered with merging and fixing the conflicts in this case?

Yes.  Then the reviewer can see the changes since the last review (even if that is a complete re-write).  The reviewer can always see the net change via the files view, but being able to see the changes since the last review often can be helpful.  By force-pushing, the comments become disconnected and the reviewer loses the ability to know what has changed since the last review.  Force-pushing should be a last resort once a PR starts being reviewed.

PR #3155, opened 2021-08-05
Simple nested additions v2

revans2:
This is the second set of changes trying to add more support for nested types when all we have to do is enable and test them. This mostly covers expressions. Sadly I couldn't add a lot for structs but I did manage to add a bit more for decimal and lists and maps.


revans2:
build

jlowe (approved):

Issue #1459, opened 2021-01-06
[FEA] Support union for nested types


revans2:
**Is your feature request related to a problem? Please describe.**
Union only works for non-nested types right now. It would be nice to support it for nested types as well. This is mostly blocked on scalar support for nested types. When we do get around to implementing it we should be sure to test with [SPARK-32376](https://issues.apache.org/jira/browse/SPARK-32376) where a struct can have null columns added to it to make the two sides match. I am not sure if it will require any code changes on our part though.

kuhushukla:
Thanks for the feature request!

Did some simple struct tests and adding the nested types to expr checks seems to work and match cpu result. But again, it was a preliminary test.
@revans2 
>This is mostly blocked on scalar support for nested types
> 
Could you elaborate what do you mean by this?  Which test should exercise this limitation and can we do without it? Thank you.

revans2:
> This is mostly blocked on scalar support for nested types

In Spark 3.1.0 and above when it sees a union of `Struct<A: Int, B: Array<String>>` and `Struct<A:Int, C: Long>` the output will look like `Struct<A:Int, B: Array<String>, C: Long>`.  To make this work the first data frame will add a `C` to the Struct that is always nulls, and the second data frame will add a `B` to the Struct that is always null.  The way that works is to insert a project that picks apart the original struct and puts it back together with the new scalar null value inserted in.  This requires us to be able to create a scalar `null` for an `Array<String>` and expand that out into a full column.

It is not directly in union that the problem show up, but it is a secondary effect that happens afterwards.

razajafri:
@sameerz we won't be able to support the union of structs in cases where the struct can have a null value. I have added tests for all the cases and xfailed the ones that we don't support ATM. 

This is still dependent on the cudf support for Scalar Structs

sameerz:
Update: at this point we support union of structs and union of nested structs.  We still need to support union of of lists and maps.

PR #3164, opened 2021-08-06
Support Union on Map types

rwlee:
Related to #1459 -- Map support but not List support.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3031,7 +3031,7 @@ object GpuOverrides {
       (shuffle, conf, p, r) => new GpuShuffleMeta(shuffle, conf, p, r)),
     exec[UnionExec](
       "The backend for the union operator",
-      ExecChecks(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64 +
+      ExecChecks(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64 + TypeSig.MAP +
```
revans2, line 3034: This is great. Could we add in arrays too? If we don't do it now we are just going to have to add it again in a separate PR soon.

```
(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64 + 
    TypeSig.MAP + TypeSig.STRUCT + TypeSig.ARRAY).nested()
```

revans2 (dismissed):

revans2 (dismissed):
My comment still applies. I would like to see us add array support here too, not because this is wrong, but because we are going to want to add it soon anyways and doing it now should not be harder then doing the original map work.  But I am approving this just in case you don't want to because this does match the requirements.

rwlee:
build

docs/supported_ops.md
```diff
@@ -424,9 +424,9 @@ Accelerator supports are described below.
 <td>S</td>
 <td><b>NS</b></td>
 <td><b>NS</b></td>
-<td><b>NS</b></td>
-<td><b>NS</b></td>
-<td><em>PS<br/>unionByName will not optionally impute nulls for missing struct fields  when the column is a struct and there are non-overlapping fields;<br/>max nested DECIMAL precision of 18;<br/>UTC is only supported TZ for nested TIMESTAMP;<br/>missing nested BINARY, CALENDAR, ARRAY, UDT</em></td>
+<td><em>PS<br/>max nested DECIMAL precision of 18;<br/>UTC is only supported TZ for nested TIMESTAMP;<br/>missing nested BINARY, CALENDAR, ARRAY, UDT</em></td>
```
revans2, line 427: Looks like we don't support arrays of arrays? Is that what you expected?

rwlee: Array support split off into another branch for simplicity and because of nested type support concerns (need to test+support maps of arrays and structs containing arrays or else nested arrays can't be supported)

revans2 (approved):

revans2:
build

rwlee:
build

saktikalyan:
build

saktikalyan:
build

sameerz:
build

sameerz:
build

PR #3173, opened 2021-08-09
Profile tool: Fix reporting app contains Dataset

tgravescs:
Found a bug where we aren't reporting the plan contains dataset properly. The issue here is timing when we set it and when we discover the dataset, which happens later. 


tgravescs:
build

tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/ApplicationInfo.scala
```diff
@@ -233,7 +233,6 @@ class ApplicationInfo(
   val accumIdToStageId: mutable.HashMap[Long, Int] = new mutable.HashMap[Long, Int]()
   var taskEnd: ArrayBuffer[TaskCase] = ArrayBuffer[TaskCase]()
   var unsupportedSQLplan: ArrayBuffer[UnsupportedSQLPlan] = ArrayBuffer[UnsupportedSQLPlan]()
-  var datasetSQL: ArrayBuffer[DatasetSQLCase] = ArrayBuffer[DatasetSQLCase]()
```
revans2, line 236: Shouldn't the case class `DatasetSQLCase` also be removed? It looks like it is only used here.

tgravescs: yes it should thanks for catching

tgravescs:
build

revans2 (approved):

nartal1 (approved):

PR #3179, opened 2021-08-10
Support GpuFirst/GpuLast on more data types

### LINKED ISSUE #3097: [FEA] Support GpuFirst/GpuLast on more data types
It is necessary to support GpuFirst/GpuLast on more data types (like: ArrayType, DecimalTypes). And we are able to provide the support without any change on libcudf.

sperlingxx:

Closes #3097

### DIFF
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -719,6 +719,7 @@ def test_first_last_reductions_extra_types(data_gen):
                 'last(a)'),
             conf = allow_negative_scale_of_decimal_conf)
 
+# TODO: https://github.com/NVIDIA/spark-rapids/issues/3221
 @allow_non_gpu('HashAggregateExec', 'SortAggregateExec',
                'ShuffleExchangeExec', 'HashPartitioning',
                'AggregateExpression', 'Alias', 'First', 'Last')
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -466,7 +466,7 @@ def test_hash_groupby_single_distinct_collect(data_gen):
 @approximate_float
 @ignore_order(local=True)
 @pytest.mark.parametrize('data_gen', _gen_data_for_collect_op, ids=idfn)
-def test_hash_groupby_collect_with_multi_distinct_fallback(data_gen):
+def test_hash_groupby_collect_with_multi_distinct(data_gen):
     def spark_fn(spark_session):
         return gen_df(spark_session, data_gen, length=100).groupby('a').agg(
             f.sort_array(f.collect_list('b')),
@@ -688,7 +688,6 @@ def test_count_distinct_with_nan_floats(data_gen):
         _no_nans_float_conf)
 
 # TODO: Literal tests
-# TODO: First and Last tests
 
 # REDUCTIONS
 
@@ -698,6 +697,7 @@ non_nan_all_basic_gens = [byte_gen, short_gen, int_gen, long_gen,
         FloatGen(no_nans=True, special_cases=[]), DoubleGen(no_nans=True, special_cases=[]),
         string_gen, boolean_gen, date_gen, timestamp_gen]
 
+_nested_gens = array_gens_sample + struct_gens_sample + map_gens_sample
 
 @pytest.mark.parametrize('data_gen', decimal_gens, ids=idfn)
 def test_first_last_reductions_extra_types(data_gen):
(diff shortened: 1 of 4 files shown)


sperlingxx:
build

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -774,6 +774,15 @@ def test_arithmetic_reductions(data_gen):
...
             conf = _no_nans_float_conf)
 
+@ignore_order(local=True)
+@pytest.mark.parametrize('data_gen',
+                         all_gen + [ArrayGen(g) for g in all_gen] + struct_gens_sample, ids=idfn)
+def test_groupby_first_last(data_gen):
+    gen_fn = [('a', RepeatSeqGen(LongGen(), length=20)), ('b', data_gen)]
+    agg_fn = lambda df: df.groupBy('a').agg(
+            f.first('b'), f.last('b'), f.first('b', True), f.last('b', True))
+    assert_gpu_and_cpu_are_equal_collect(lambda spark : agg_fn(gen_df(spark, gen_fn)))
```
revans2, line 784: nit: first and last are not deterministic when they are run in a real distributed setup.  There are times when this test is run that way. It would probably be best to do a `.coalesce(1)` on the input data so that the input order is deterministic for those cases.

razajafri: Just for my understanding, will sort also make the results deterministic in those cases? 

@sperlingxx should we also add a test for Array[Struct]??

revans2: The issue is that at least half of the aggregation happens after a shuffle, and the order that upstream task data shows up in a down stream task is not deterministic.

If I have data like
|a|b|
|-|-|
|1|1|
|1|2|
|1|3|
|1|4|

and two tasks, then the data likely would be divided up so that you get

Task 1
|a|b|
|-|-|
|1|1|
|1|2|

Task 2
|a|b|
|-|-|
|1|3|
|1|4|

The result of group by a first(b) from each of these tasks would be

Task 1
|a|b|
|-|-|
|1|1|

Task 2
|a|b|
|-|-|
|1|3|

Now when Task 3 starts that is going to pull in all of the data for `a == 1` it will start fetching from several upstream tasks in parallel.  If Task 1 shows up first you get data like.
|a|b|
|-|-|
|1|1|
|1|3|

and the file result for `first(b)` is 1, but if Task 2 wins and shows up first you get data like
|a|b|
|-|-|
|1|3|
|1|1|

and `first(b)` ends up being 3. 

Sorting the data after that makes no difference. 1 != 3  We get away with this most of the time on local mode tests because in the Spark shuffle there is an optimization for local reads That puts them first in the queue, and it really is a queue so the order is maintained. As soon as you go to a distributed setup you can get different answers from one run to the next, even on the CPU.  By doing a coalesce(1) we are only going to have a single task so it removes the ambiguity.

sperlingxx:
build

sperlingxx:
build

pxLi:
NFS lock issue, let me ask blossom to fix. Please retrigger later

pxLi:
build

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -463,17 +463,8 @@ def test_hash_groupby_single_distinct_collect(data_gen):
...
-# aggregations (ArrayType).
-# TODO: support GPUFirst on ArrayType https://github.com/NVIDIA/spark-rapids/issues/3097
 @approximate_float
 @ignore_order(local=True)
-@allow_non_gpu('SortAggregateExec',
-               'SortArray', 'Alias', 'Literal', 'First', 'If', 'EqualTo', 'Count',
-               'CollectList', 'CollectSet', 'AggregateExpression')
-@incompat
 @pytest.mark.parametrize('data_gen', _gen_data_for_collect_op, ids=idfn)
 def test_hash_groupby_collect_with_multi_distinct_fallback(data_gen):
```
revans2, line 469: Can we change the name of the test? It is not fallback back any more.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2119,10 +2119,13 @@ object GpuOverrides {
     expr[First](
       "first aggregate operator",
       ExprChecks.aggNotWindow(
-        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64,
+        (TypeSig.psNote(TypeEnum.STRUCT, "only supports in GroupBy aggregation") +
+          TypeSig.psNote(TypeEnum.ARRAY, "only supports in GroupBy aggregation") +
```
revans2, line 2123: I'm not sure why we need a partial support note here?  Does STRUCT and ARRAY not work for first/last on a reduction?  If so then we need tests to verify that we will fall back to the CPU correctly in those cases, and a follow on issue to fix it in cudf.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2119,10 +2119,13 @@ object GpuOverrides {
     expr[First](
       "first aggregate operator",
       ExprChecks.aggNotWindow(
-        TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64,
+        (TypeSig.psNote(TypeEnum.STRUCT, "only supports in GroupBy aggregation") +
```
revans2, line 2122: If we support arrays and structs then we should also support maps.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala

integration_tests/src/main/python/hash_aggregate_test.py
sperlingxx: Fixed.

sperlingxx:
build

revans2 (approved):

PR #3183, opened 2021-08-10
Add TypeSig checks for join keys and other special cases

### LINKED ISSUE #2102: [DOC] Supported ops page differentiate type supported in key vs passed along
## Report incorrect documentation

The supported ops page doesn't seem to indicate if a type is supposed as the main key (like joining on) vs just being able to be passed along as another columns.  Perhaps we should add a note for this.

andygrove:

Closes #2102 

This PR allows us to provide TypeSig signatures for join keys and generates appropriate documentation in support_ops.md

This is implemented for joins and WindowExec.

### DIFF
--- a/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
+++ b/shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
@@ -234,8 +234,8 @@ abstract class SparkBaseShims extends SparkShims {
             TypeSig.STRUCT + TypeSig.MAP)
           .nested(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.STRUCT +
           TypeSig.DECIMAL_64),
-          Map("leftKeys" -> (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64),
-            "rightKeys" -> (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64)),
+          Map("leftKeys" -> TypeSig.joinKeyTypes,
+            "rightKeys" -> TypeSig.joinKeyTypes),
           TypeSig.all),
         (join, conf, p, r) => new GpuSortMergeJoinMeta(join, conf, p, r)),
       GpuOverrides.exec[BroadcastHashJoinExec](
@@ -244,8 +244,8 @@ abstract class SparkBaseShims extends SparkShims {
             TypeSig.ARRAY + TypeSig.STRUCT + TypeSig.MAP)
           .nested(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.STRUCT +
             TypeSig.DECIMAL_64),
-          Map("leftKeys" -> (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64),
-            "rightKeys" -> (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64)),
+          Map("leftKeys" -> TypeSig.joinKeyTypes,
+            "rightKeys" -> TypeSig.joinKeyTypes),
           TypeSig.all),
         (join, conf, p, r) => new GpuBroadcastHashJoinMeta(join, conf, p, r)),
       GpuOverrides.exec[ShuffledHashJoinExec](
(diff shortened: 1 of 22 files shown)


shims/spark301/src/main/scala/com/nvidia/spark/rapids/shims/spark301/SparkBaseShims.scala
```diff
@@ -232,31 +232,31 @@ abstract class SparkBaseShims extends SparkShims {
         "Sort merge join, replacing with shuffled hash join",
         ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64 + TypeSig.ARRAY +
             TypeSig.STRUCT + TypeSig.MAP)
-          .withPsNote(TypeEnum.ARRAY, "Cannot be used as join key")
-          .withPsNote(TypeEnum.STRUCT, "Cannot be used as join key")
-          .withPsNote(TypeEnum.MAP, "Cannot be used as join key")
```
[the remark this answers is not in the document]
andygrove: Prior to this PR we relied on the "ps notes" to document the fact that we don't support map/struct/array as join keys, but we now have specific lines in the supported_ops table for `leftKeys` and `rightKeys` for the join operators, showing which types are supported, so we no longer need these ps notes.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -807,19 +811,39 @@ class ExecChecks private(
...
+
+    extendedChecks.foreach {
+      case (fieldName, typeSig) =>
+        val fieldMeta = meta.namedChildExprs(fieldName)
+          .flatMap(_.typeMeta.dataType)
+          .zipWithIndex
+          .map(t => StructField(s"c${t._2}", t._1))
+        tagUnsupportedTypes(meta, typeSig, allowDecimal, fieldMeta,
+          s"unsupported data types in '$fieldName': %s")
+    }
```
revans2, line 823: This is great. I just had another thought. It would be good as an internal sanity check that we also verify that there is a check for each key in `namedChildExprs`.  We already will throw an exception if `fieldName` is not in `namedChildExprs`. We should just make sure that both are tied to each other.

revans2 (dismissed):
I just had one more thing to add to my wish list for this. I think it looks great. The next step would be too apply it to other areas that also need this type of check.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
andygrove: I added a check for this and manually tested by introducing a typo in the key of one of the hash maps and got the error `GpuBroadcastHashJoinMeta is missing ExecChecks for rightKeys`.

jlowe (approved):

jlowe:
build

shims/spark311db/src/main/scala/com/nvidia/spark/rapids/shims/spark311db/Spark311dbShims.scala
```diff
@@ -170,33 +170,32 @@ class Spark311dbShims extends Spark311Shims {
...
         ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64 + TypeSig.ARRAY +
             TypeSig.STRUCT + TypeSig.MAP)
-          .withPsNote(TypeEnum.ARRAY, "Cannot be used as join key")
-          .withPsNote(TypeEnum.STRUCT, "Cannot be used as join key")
-          .withPsNote(TypeEnum.MAP, "Cannot be used as join key")
           .nested(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.STRUCT +
-          TypeSig.DECIMAL_64)
-          , TypeSig.all),
+          TypeSig.DECIMAL_64),
+          Map("leftKeys" -> (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64),
```
revans2, line 186: It looks like as part of an upmerge or whatever that you missed that structs, even nested structs, can be used as join keys and now we have tests failing because of it.

andygrove:
build

revans2 (dismissed):

revans2 (approved):

pxLi:
build

Issue #3111, opened 2021-08-02
[FEA] Support conditional nested loop joins


jlowe:
cudf added Java support for conditional nested loop joins in https://github.com/rapidsai/cudf/pull/8888, and this can be leveraged for GPU support Spark inequality joins (i.e.: BroadcastNestedLoopJoin and CrossJoin with conditions).  The Catalyst condition will need to be converted into an equivalent AST condition.

This covers the pure inequality join portion of #288.

The following conditional nested loop joins are more straightforward, where we can simply apply a condition to the normal nested loop join:
- [x] LeftOuter with BuildRight
- [x] LeftSemi with BuildRight
- [x] LeftAnti with BuildRight
- [x] RightOuter with BuildLeft

The following conditional nested loop joins are more involved to implement in a batched fashion, we need to track which rows in the build side have matched as we join with the stream side in batches.  If it is an outer join, we emit the matched join rows for each batch, but for semi/anti join we simply track the rows that matched.  Once we have exhausted the stream side, we either need to emit the null join rows for the build side rows that were left unmatched in the case of outer joins or simply emit the rows that were matched/unmatched for the semi/anti join case, respectively.
- [ ] LeftOuter with BuildLeft
- [ ] RightOuter with BuildRight
- [ ] FullOuter
- [ ] LeftSemi with BuildLeft
- [ ] LeftAnti with BuildLeft

jlowe:
The join types left to be implemented are tracked in #3269.  Closing this as fixed by #3184 and #3242.

PR #3184, opened 2021-08-10
Support broadcast nested loop join for LeftSemi and LeftAnti

jlowe:
Part of #3111.

Adds GPU support for LeftSemi and LeftAnti broadcast nested loop joins for conditions that can be supported by cudf AST expressions.  This is dependent upon https://github.com/rapidsai/cudf/pull/9002 for the new conditional join method names.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExec.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExec.scala
@@ -96,7 +96,7 @@ class GpuBroadcastNestedLoopJoinMeta(
       case _: InnerLike => false
       case LeftSemi | LeftAnti =>
         conditionMeta.foreach(_.tagForAst())
-        val isAst = conditionMeta.exists(_.canThisBeAst)
+        val isAst = conditionMeta.forall(_.canThisBeAst)
         assert(isAst, s"Non-AST condition in ${join.joinType}")
         isAst
       case _ => throw new IllegalStateException(s"${join.joinType} nested loop join not supported")
@@ -324,8 +324,10 @@ object GpuBroadcastNestedLoopJoinExecBase extends Arm {
       // A nested loop join with no condition directly in the join is treated like a cross join.
       new CrossJoinIterator(builtBatch, stream, targetSize, buildSide, joinTime, totalTime)
     } else {
-      val compiledAst = boundCondition.get.convertToAst(numFirstTableColumns)
-          .asInstanceOf[ast.Expression].compile()
+      val compiledAst = boundCondition.get.convertToAst(numFirstTableColumns) match {
+        case e: ast.Expression => e.compile()
+        case e => new ast.UnaryExpression(ast.UnaryOperator.IDENTITY, e).compile()
+      }
       joinType match {
         case LeftAnti | LeftSemi =>
           assert(buildSide == GpuBuildRight)
@@ -364,6 +366,8 @@ object GpuBroadcastNestedLoopJoinExecBase extends Arm {
         }
         numOutputRows += ret.numRows()
         numOutputBatches += 1
+        // grab the semaphore for downstream processing
+        GpuSemaphore.acquireIfNecessary(TaskContext.get())
         ret
       })
     }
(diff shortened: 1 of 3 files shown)


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExec.scala
```diff
@@ -79,32 +89,46 @@ class GpuBroadcastNestedLoopJoinMeta(
...
+      condition.map(c => GpuFilterExec(c, joinExec)).getOrElse(joinExec)
+    }
   }
 }
 
-/**
- * An iterator that does a cross join against a stream of batches.
- */
-class CrossJoinIterator(
+/** Base class for the join iterators based on a nested loop join */
```
revans2, line 119: nit: Is this new with the formatting? Generally I thought the multi-line format was preferred for classes. This is more for my info so I can keep things consistent.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExec.scala
```diff
@@ -207,29 +203,143 @@ class CrossJoinIterator(
...
+      boundCondition: Option[GpuExpression],
+      spillCallback: SpillCallback,
       numOutputRows: GpuMetric,
       joinOutputRows: GpuMetric,
       numOutputBatches: GpuMetric,
       joinTime: GpuMetric,
       totalTime: GpuMetric): Iterator[ColumnarBatch] = {
-    val joinIterator =
+    val joinIterator = if (boundCondition.isEmpty) {
+      // A nested loop join with no condition directly in the join is treated like a cross join.
```
revans2, line 324: I don't think this is true.  It looks like if there is no condition, then the bound condition always returns true.

https://github.com/apache/spark/blob/c888bad6a12b45f3eda8d898bdd90405985ee05c/sql/core/src/main/scala/org/apache/spark/sql/execution/joins/BroadcastNestedLoopJoinExec.scala#L109

Which means if I do something stupid like. 

```
df.join(df2, Seq(), "leftAnti").show
```

I will get a cross join as output instead of nothing, which I should have gotten.

We should either fall back to the CPU in very odd cases like this, or do what Spark does and force the conditional to be `GpuLiteral(True)`.

revans2: Actually I tested the join types and this is true for everything except left_semi and left_anti. All of the others do end up being a cross join. So the check should probably just be for semi or anti.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExec.scala
jlowe: It's more my preference.  I prefer to keep one-liner comments as one line, regardless of whether it's a class, variable, method, etc. to save on the vertical whitespace.  I can turn this back into a multi-line comment.

revans2: I don't care. If the style checker says it is okay then it is okay with me. I just was curious.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExec.scala
jlowe: Well that ended up being a fun journey.  Not only was lacking condition expression not being handled correctly for semi/anti, the count-only handling was also incorrect because it was specific to cross/inner.  I've updated the code to handle the degenerate semi/anti cases when the condition is empty as well as handling the count-only special case for those.  I ended up refactoring the code a bit into sub-methods to keep `doExecuteColumnar` from becoming a large, unwieldy function.

I also fixed an issue where `getRowCountAndClose` was releasing the GPU semaphore but never re-acquiring it before emitting the next columnar batch.

revans2: So the comment and code is identical to what was here before. It looks like you worked around it all at a higher level, which is fine, but it would be nice to have the comments here or on the method explain what is and is not supported just to be careful

jlowe: Sorry, I meant to update the comment as part of the refactor.  I tweaked the comments and added an assert.

revans2 (dismissed):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExec.scala
```diff
@@ -350,30 +466,95 @@ abstract class GpuBroadcastNestedLoopJoinExecBase(
...
+              joinTime, totalTime)
+          }
+      }
+    }
+  }
 
-      lazy val buildCount: Int = computeBuildRowCount(broadcastRelation, buildTime, buildDataSize)
+  /** Special-case handling of an unconditional join that just needs to output a row count. */
+  private def doUnconditionalJoinRowCount(
+      broadcastRelation: Broadcast[SerializeConcatHostBuffersDeserializeBatch],
```
revans2, line 542: Trailing comma here that messes with scala style

jlowe:
build

revans2 (approved):

PR #3200, opened 2021-08-11
Enable a smoke test for UCX in pre-merge

abellina:
This PR runs `test_hash_grpby_sum` with UCX enabled, as UCX is [now installed in the docker image](https://github.com/NVIDIA/spark-rapids/pull/3218).

This adds ~1-2 minutes to the build from the logs. It starts the smoke test from the `mvn_verify` function so it could be merged without disturbing the jenkins file. The `mvn_verify` and now `rapids_shuffle_smoke_test` assume spark 3.0.1 for the pre-merge build. There will be a follow on PR to clean up how `rapids_shuffle_smoke_test` gets invoked, though right now I piggy back from `mvn_verify` binaries.

I added a variable so the correct shuffle manager shim can be picked up (`SHUFFLE_SPARK_SHIM`). This should be kept in sync with `SPARK_VER`.

There is a `local-cluster` issue I noted which happens even without the shuffle manager and will file a follow up, as it seems unrelated. I am inclined to let this use standalone mode for now. I could use some more eyes on it to see if it makes sense.

Also I chose a simple test, it does use the UCX shuffle. I can enable others in case it's not going to change that 1-2 minute time addition too much, or pick something else. Just pick the first thing that I knew would use UCX.


abellina:
build

abellina:
build

abellina:
build

abellina:
build

abellina:
This seems to mostly be working, but I am getting an issue in `local-cluster` mode where executors get created after the tests (and I mean 100 executors or so). The logs show UCX is utilized, but something is happening later. Note this issue with the runaway executors is with or without the manager, so likely a pytest config issue on my side.

abellina:
build

abellina:
build

abellina:
build

abellina:
Waiting on: https://github.com/NVIDIA/spark-rapids/pull/3218 to be merged and image pushed.

abellina:
build

abellina:
build

abellina:
Ok last version does work with standalone mode. The `local-cluster` issue I noted happens even without the shuffle manager and will file a follow up. I am inclined to let this use standalone mode for now. I could use some more eyes on it to see if it makes sense.

This adds ~1-2 minutes to the build from the logs. It starts the smoke test from the `mvn_verify` function so it could be merged without disturbing the jenkins file. The `mvn_verify` and now `rapids_shuffle_smoke_test` assume spark 3.0.1 for the pre-merge build. I added a variable so the correct shuffle manager shim can be picked up (`SHUFFLE_SPARK_SHIM`). This should be kept in sync with `SPARK_VER`.

abellina:
Also I chose a simple test, it does use the UCX shuffle. I can enable others in case it's not going to change that 1-2 minute time addition too much, or pick something else. Just pick the first thing that I knew would use UCX.

jenkins/version-def.sh
```diff
@@ -31,6 +31,7 @@ CUDA_CLASSIFIER=${CUDA_CLASSIFIER:-"cuda11"}
 PROJECT_VER=${PROJECT_VER:-"21.10.0-SNAPSHOT"}
 PROJECT_TEST_VER=${PROJECT_TEST_VER:-"21.10.0-SNAPSHOT"}
 SPARK_VER=${SPARK_VER:-"3.0.1"}
+SHUFFLE_SPARK_SHIM=${SHUFFLE_SPARK_SHIM:-"spark301"}
```
jlowe, line 34: If we want to try to keep this in sync automatically in case someone only sets `SPARK_VER` we could try something like this.  Should work for Apache Spark versions but probably doesn't do so well with others like EMR, Databricks, etc.  However it'd be nice to work for most common cases rather than requiring users to remember to set both all the time.
```suggestion
SHUFFLE_SPARK_SHIM=${SHUFFLE_SPARK_SHIM:-spark${SPARK_VER//./}}
```

jenkins/spark-premerge-build.sh
```diff
@@ -63,8 +63,40 @@ mvn_verify() {
...
+      PYSP_TEST_spark_executor_cores=1 \
+      SPARK_SUBMIT_FLAGS="--conf spark.executorEnv.UCX_ERROR_SIGNALS=" \
+      PYSP_TEST_spark_shuffle_manager=com.nvidia.spark.rapids.$SHUFFLE_SPARK_SHIM.RapidsShuffleManager \
+      PYSP_TEST_spark_rapids_memory_gpu_minAllocFraction=0 \
+      PYSP_TEST_spark_rapids_memory_gpu_maxAllocFraction=0.1 \
+      PYSP_TEST_spark_rapids_memory_gpu_allocFraction=0.1 \
+      ./integration_tests/run_pyspark_from_build.sh -m shuffle_test
+
+    $SPARK_HOME/sbin/spark-daemon.sh stop org.apache.spark.deploy.worker.Worker 1
+    $SPARK_HOME/sbin/stop-master.sh
```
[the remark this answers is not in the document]
abellina: yes, I can't think of a reason to complicate the script to catch the error code and shut the processes down, if everything is going down anyway.

jenkins/version-def.sh
abellina: that is much better for the default value. I'll add a comment also.

jlowe (dismissed):

jlowe:
build

jenkins/version-def.sh
```diff
@@ -31,6 +31,11 @@ CUDA_CLASSIFIER=${CUDA_CLASSIFIER:-"cuda11"}
 PROJECT_VER=${PROJECT_VER:-"21.10.0-SNAPSHOT"}
 PROJECT_TEST_VER=${PROJECT_TEST_VER:-"21.10.0-SNAPSHOT"}
 SPARK_VER=${SPARK_VER:-"3.0.1"}
+# Make a best attempt to set the default value for the shuffle shim.
+# Note that SPARK_VER for non-Apache Spark flavors (i.e. databricks,
+# cloudera, and others) may not be a simple as just the version number, so
+# this variable should be set accordingly.
+SHUFFLE_SPARK_SHIM=${SHUFFLE_SPARK_SHIM:-spark${SPARK_VER//./}}
```
pxLi, line 38: no csp version as default is reasonable. can we add default to cover spark snapshot versions here? maybe something like 

`${SHUFFLE_SPARK_SHIM:-spark${SPARK_VER//[.|\-SNAPSHOT]/}}`

abellina: I am not an expert at bash pattern matching, but searching around yielded this:
`${SHUFFLE_SPARK_SHIM:-spark${SPARK_VER//?(.|\-SNAPSHOT)/}}`

the pattern you provided matches the letters S,N,A,P,S,H,O,T in any order, the pattern above specifically looks for `-SNAPSHOT` and trims it in as one unit.

abellina:
build

pxLi (approved):

zhanga5 (approved):

pxLi:
```
java.lang.ClassNotFoundException: com.nvidia.spark.rapids.spark3.0.1.RapidsShuffleManager
```
Looks like the pattern matching does not work consistently in different bash versions

abellina:
> Looks like the pattern matching does not work consistently in different shell envs

I tested this with the command line, instead of within bash, so I was able to reproduce the issue locally. I got it to work with 2 statements, but was not able to with a single statement.

abellina:
build

jlowe (approved):

Issue #2731, opened 2021-06-17
Premerge build takes too long (3 hours)


tgravescs:
**Describe the bug**
At the end of 21.06 the premerge build was taking around 3 hours. This is too long to move quickly, especially if you have prs depending on other prs.    

I know part of the time is because we run the test on a bunch of various spark versions.

We should investigate to see if there are ways to make this run in a more timely manner.

Perhaps we move more spark versions to just run nightly, perhaps there is stuff we can parallelize more, etc..

GaryShen2008:
Is it possible to skip the xfail cases to make it faster?

tgravescs:
pytest has the skip functionality, we could go through and see if we can convert any to that, but we don't want to blanket do that,  some we want to explicitly show up as xfail.

revans2:
I downloaded a recent log file and pulled out what maven reports for timing, and then summed things up.

| stage | time | percent total|
|--------|--------|-----------------|
|rapids-4-spark-api-validation | 00:00:19| 0.2%|
|rapids-4-spark-integration-tests_2.12| **01:07:22**| **41.7%**|
|rapids-4-spark-tests-next-spark_2.12| 00:01:17| 0.8%|
|RAPIDS Accelerator for Apache Spark Distribution| 00:00:16|0.2%|
|RAPIDS Accelerator for Apache Spark Root Project| 00:00:23|0.2%|
|RAPIDS Accelerator for Apache Spark Scala UDF Plugin|  00:04:13| 2.6%|
|RAPIDS Accelerator for Apache Spark Shuffle Plugin| 00:00:24|0.2%|
|RAPIDS Accelerator for Apache Spark SQL Plugin Shim Aggregator| 00:00:06|0.1%|
|RAPIDS Accelerator for Apache Spark SQL Plugin Shims| 00:00:07|0.1%|
|RAPIDS Accelerator for Apache Spark SQL Plugin Spark 3.0.1 EMR Shim| 00:00:21|0.2%|
|RAPIDS Accelerator for Apache Spark SQL Plugin Spark 3.0.1 Shim| 00:00:34|0.4%|
|RAPIDS Accelerator for Apache Spark SQL Plugin Spark 3.0.2 Shim| 00:00:22|0.2%|
|RAPIDS Accelerator for Apache Spark SQL Plugin Spark 3.0.3 Shim| 00:00:45|0.5%|
[...]

PR #3204, opened 2021-08-12
Decrease parallelism for pre-merge integration tests

pxLi:

We realize some ephemeral OOM kills for xdist process in pre-merge, lets decrease the parallelism for now.

And We have another issue to try split integration tests run in different stages/pods https://github.com/NVIDIA/spark-rapids/issues/2731


zhanga5 (approved):

pxLi:
this setup cost 15 more mins to finish the integration tests. And the peak mem usage was ~47GB down from ~57GB

Issue #3229, opened 2021-08-13
[FEA] Improve CreateMap to support multiple key and value expressions


andygrove:
**Is your feature request related to a problem? Please describe.**
Issue https://github.com/NVIDIA/spark-rapids/issues/3014 adds basic support for CreateMap from two expressions. This issue is to extend support for multiple expressions e.g. `select map(key1, value1, key2, value2) from foo`.

**Describe the solution you'd like**
We either need to implement this in cuDF or in the plugin. We could concatenate the columns into a single pair of columns before calling cuDF.

**Describe alternatives you've considered**
N/A

**Additional context**
This is part of #44

PR #3230, opened 2021-08-13
Initial support for CreateMap on GPU

### LINKED ISSUE #3014: [FEA] Add initial support for CreateMap
I would like to be able to create Map types on the GPU. For example:

```scala
spark.sql("select map(name,name) as newmap from df1")
```

Create a `GpuCreateMap` expression.

None

**Additional context**
This is part of https://github.com/NVIDIA/spark-rapids/issues/44

andygrove:

Closes #3014 

There is a follow-on issue https://github.com/NVIDIA/spark-rapids/issues/3229 for supporting multiple key-value pairs.

### DIFF
--- a/integration_tests/src/main/python/map_test.py
+++ b/integration_tests/src/main/python/map_test.py
@@ -38,35 +38,13 @@ def test_simple_get_map_value(data_gen):
                 'a["NOT_FOUND"]',
                 'a["key_5"]'))
 
-def test_map_expr_project_string_to_string():
-    data_gen = [('a', StringGen(nullable=False)), ('b', StringGen(nullable=False))]
+@pytest.mark.parametrize('key_gen', [StringGen(nullable=False), IntegerGen(nullable=False), basic_struct_gen], ids=idfn)
+@pytest.mark.parametrize('value_gen', [StringGen(nullable=False), IntegerGen(nullable=False), basic_struct_gen], ids=idfn)
+def test_single_entry_map(key_gen, value_gen):
+    data_gen = [('a', key_gen), ('b', value_gen)]
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : gen_df(spark, data_gen).selectExpr(
-                "map(a, b) as m1"))
-
-def test_map_expr_project_int_to_int():
-    data_gen = [('a', IntegerGen(nullable=False)), ('b', IntegerGen(nullable=False))]
-    assert_gpu_and_cpu_are_equal_collect(
-            lambda spark : gen_df(spark, data_gen).selectExpr(
-                "map(a, b) as m2"))
-
-def test_map_expr_project_string_to_int():
-    data_gen = [('a', StringGen(nullable=False)), ('b', IntegerGen(nullable=False))]
-    assert_gpu_and_cpu_are_equal_collect(
-            lambda spark : gen_df(spark, data_gen).selectExpr(
-                "map(a, b) as m1"))
-
-def test_map_expr_project_int_to_string():
-    data_gen = [('a', IntegerGen(nullable=False)), ('b', StringGen(nullable=False))]
-    assert_gpu_and_cpu_are_equal_collect(
-            lambda spark : gen_df(spark, data_gen).selectExpr(
-                "map(a, b) as m1"))
-
-def test_map_expr_project_struct_to_struct():
-    data_gen = [('a', basic_struct_gen), ('b', basic_struct_gen)]
-    assert_gpu_and_cpu_are_equal_collect(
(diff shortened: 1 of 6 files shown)


integration_tests/src/main/python/map_test.py
```diff
@@ -32,6 +38,44 @@ def test_simple_get_map_value(data_gen):
...
+            lambda spark : gen_df(spark, data_gen).selectExpr(
+                "map(a, b) as m2"))
+
+def test_map_expr_project_string_to_int():
+    data_gen = [('a', StringGen(nullable=False)), ('b', IntegerGen(nullable=False))]
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : gen_df(spark, data_gen).selectExpr(
+                "map(a, b) as m1"))
+
+def test_map_expr_project_int_to_string():
```
revans2, line 59: nit: I would prefer to see this parametrize. That way we can add more types simply.

```

@pytest.mark.parametrize('key_type', [StringGen, IntegerGen], ids=idfn)
@pytest.mark.parametrize('value_type', [StringGen, IntegerGen], ids=idfn)
def test_single_entry_map(key_type, value_type):
    data_gen = [('a', key_gen(nullable=False)), ('b', value_gen(nullable=False))]
            lambda spark : gen_df(spark, data_gen).selectExpr(
                "map(a, b) as m1"))
```

integration_tests/src/main/python/map_test.py
```diff
@@ -32,6 +38,44 @@ def test_simple_get_map_value(data_gen):
...
+            lambda spark : gen_df(spark, data_gen).selectExpr(
+                "map(a, b) as m1"))
+
+@allow_non_gpu('ProjectExec,Alias,CreateMap')
+# until https://github.com/NVIDIA/spark-rapids/issues/3229 is implemented
+def test_map_expr_multiple_pairs():
+    data_gen = [('a', StringGen(nullable=False)), ('b', StringGen(nullable=False))]
+    assert_gpu_fallback_collect(
+            lambda spark : gen_df(spark, data_gen).selectExpr(
+                "map(a, b, b, a) as m1"), 'ProjectExec')
```
revans2, line 77: What happens if "a == b" for a given row?  

Spark has a config for what should happen.

https://github.com/apache/spark/blob/f620996142ba312f7e52f75476b1b18be667ffdf/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L2969-L2982

Sadly neither of those we are going to be able to support out of the box. I think we need to file a follow on issue so that we can figure out how to dedupe maps the proper way, with cudf's help. In the short term we might need to have a config indicating that it is okay to enable this because we are not doing duplicate checks, but if the keys are all scalar values we can determine duplicates up front and avoid the entire issue.

integration_tests/src/main/python/map_test.py
```diff
@@ -32,6 +38,44 @@ def test_simple_get_map_value(data_gen):
                 'a["NOT_FOUND"]',
                 'a["key_5"]'))
```
revans2, line 40: I think a very common case is going to be scalar keys (especially strings). Could we please test this case too?

andygrove: I spent some time looking into this and we do not support this yet. I filed https://github.com/NVIDIA/spark-rapids/issues/3245 to implement it.

integration_tests/src/main/python/map_test.py
andygrove: https://github.com/NVIDIA/spark-rapids/issues/3229 is the follow-on issue for supporting multiple key-value pairs and where we would need to tackle the duplicate issue.

revans2: It can be a separate issue after that or as a part of it. I think most of the time people will use literal values for the keys. If that is the case, then we can still enable this for those cases without much difficulty, but we would need to fall back to the CPU for other cases, or have a config to say I know what I am doing.

integration_tests/src/main/python/map_test.py
revans2: Sorry about the confusion. I think I put the comment in a bad place.  What I think is going to be common is something like.

```
map("key1", valueColumn)
```

This is what I want a test for. From the code it looks like it should just work all the time, but it would be nice to have an explicit test for this.

andygrove: That makes more sense. I added the test and it did expose an issue, which is now fixed.

revans2 (approved):

andygrove:
build

revans2 (approved):

revans2:
build

PR #3231, opened 2021-08-13
Fix some resource leaks in GpuCast and RapidsShuffleServerSuite

jlowe:
Noticed some warnings about leaked resources in the detailed test log, and upon investigating I found a few leaks.  In the case of GpuCast, a column vector was being leaked if we threw an exception due to ANSI mode checks.  For RapidsShuffleServerSuite, the `getBufferSendState` method returns an instance that must be closed but the tests were failing to do so.


jlowe:
Note that viewing the diff with whitespace changes hidden should help with the GpuCast change.

jlowe:
build

abellina (approved):

revans2 (approved):

PR #3234, opened 2021-08-14
Parquet support for MapType

razajafri:
This PR adds support for Maps and nested Maps for parquet.

### DIFF
--- a/integration_tests/src/main/python/parquet_write_test.py
+++ b/integration_tests/src/main/python/parquet_write_test.py
@@ -45,17 +45,15 @@ parquet_struct_gen = [StructGen([['child'+str(ind), sub_gen] for ind, sub_gen in
 parquet_array_gen = [ArrayGen(sub_gen, max_length=10) for sub_gen in parquet_basic_gen + parquet_struct_gen] + \
                     [ArrayGen(ArrayGen(sub_gen, max_length=10), max_length=10) for sub_gen in parquet_basic_gen + parquet_struct_gen]
 
-parquet_write_gens_list0 = parquet_basic_gen + parquet_struct_gen + parquet_array_gen + parquet_decimal_gens
-parquet_map_gens = map_gens_sample + [MapGen(StructGen([['child0', StringGen()], ['child1', StringGen()]], nullable=False), FloatGen())]
-# splitting it in two lists as the data is too big for a single parquet file in our unit tests
-parquet_write_gens_list = [parquet_write_gens_list0, parquet_map_gens]
+parquet_map_gens = map_gens_sample + [MapGen(StructGen([['child0', StringGen()], ['child1', StringGen()]], nullable=False), FloatGen()), MapGen(StructGen([['child0', StringGen(nullable=True)]], nullable=False), StringGen())]
+parquet_write_gens_list = [parquet_basic_gen + parquet_struct_gen + parquet_array_gen + parquet_decimal_gens + parquet_map_gens]
 parquet_ts_write_options = ['INT96', 'TIMESTAMP_MICROS', 'TIMESTAMP_MILLIS']
 
 @pytest.mark.parametrize('parquet_gens', parquet_write_gens_list, ids=idfn)
 @pytest.mark.parametrize('reader_confs', reader_opt_confs)
 @pytest.mark.parametrize('v1_enabled_list', ["", "parquet"])
 @pytest.mark.parametrize('ts_type', parquet_ts_write_options)
-def test_write_round_trip(spark_tmp_path, parquet_gens, v1_enabled_list, ts_type,
+def test_write_round_trip1(spark_tmp_path, parquet_gens, v1_enabled_list, ts_type,
                                   reader_confs):
     gen_list = [('_c' + str(i), gen) for i, gen in enumerate(parquet_gens)]
     data_path = spark_tmp_path + '/PARQUET_DATA'
--- a/integration_tests/src/main/python/parquet_write_test.py
+++ b/integration_tests/src/main/python/parquet_write_test.py
@@ -20,6 +20,7 @@ from data_gen import *
 from marks import *
 from pyspark.sql.types import *
 from spark_session import with_cpu_session, with_gpu_session
+import pyspark.sql.functions as f
 import random
 
 # test with original parquet file reader, the multi-file parallel reader for cloud, and coalesce file reader for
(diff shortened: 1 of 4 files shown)


razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -105,26 +105,60 @@ object GpuParquetFileFormat {
...
+    }
+  }
+
+  def parquetWriterOptionsFromSchema[T <: NestedBuilder[_, _], V <: ParquetColumnWriterOptions](
+      builder: ParquetColumnWriterOptions.NestedBuilder[T, V],
+      schema: StructType,
+      writeInt96: Boolean,
+      nullable: Boolean = true): T = {
     // TODO once https://github.com/rapidsai/cudf/issues/7654 is fixed go back to actually
     // setting if the output is nullable or not everywhere we have hard-coded nullable=true
```
jlowe, line 139: Has this been addressed?  Looks like the code is removing the hardcoded `true` nullable parameters.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2941,6 +2941,7 @@ object GpuOverrides {
       ExecChecks((TypeSig.commonCudfTypes +
         TypeSig.DECIMAL_64.withPsNote(TypeEnum.DECIMAL, "Only supported for Parquet") +
         TypeSig.STRUCT.withPsNote(TypeEnum.STRUCT, "Only supported for Parquet") +
+          TypeSig.MAP.withPsNote(TypeEnum.MAP, "Only supported for Parquet") +
```
revans2, line 2944: nit: Indentation appears to be off compared to the other lines.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
revans2: It has not been addressed. Cudf does not want to do it because of performance :). So it is unlikely to ever be addressed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -105,26 +105,60 @@ object GpuParquetFileFormat {
...
+      case _ =>
+        new ParquetColumnWriterOptions(name, nullable)
+    }
+  }
+
+  def parquetWriterOptionsFromSchema[T <: NestedBuilder[_, _], V <: ParquetColumnWriterOptions](
+      builder: ParquetColumnWriterOptions.NestedBuilder[T, V],
+      schema: StructType,
+      writeInt96: Boolean,
+      nullable: Boolean = true): T = {
```
revans2, line 137: I really dislike having a parameter for nullable separate from the schema.  The schema should have nullable in it. If it is a `nullableOverride`, then name it like that. We need comments explaining how this is supposed to be used and what it does. Because from what I can see there are bugs in this code. `Map<Array<String>, String>` with the current code says not only that the key to the Map needs to not be nullable, but also that the Strings within the Array cannot be nullable, which I don't think is true.  But I could be wrong, because the recursion appears to call into `getParquetColumnWriterOption` for map keys and values only and honestly trying to trace this through everywhere with the separate functions that appear to be 99% the same is really confusing to me.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -105,26 +105,60 @@ object GpuParquetFileFormat {
     }
   }
 
-  def parquetWriterOptionsFromSchema[T <: NestedBuilder[_, _], V <: ParquetColumnWriterOptions]
-  (builder: ParquetColumnWriterOptions.NestedBuilder[T, V],
-   schema: StructType, writeInt96: Boolean): T = {
+  def getParquetColumnWriterOption(
```
revans2, line 108: Why exactly do we need a separate API here just for map keys and values?  I really don't like the existing code for arrays that pulls out the element type to wrap it in a struct just so we can have a little bit of code reuse. To me it feels very much like we are doing the recursion at the wrong level, but I have not taken the time to think it through.

integration_tests/src/main/python/parquet_write_test.py
```diff
@@ -34,20 +34,21 @@
...
+                      StructGen([['child0', MapGen(StringGen(nullable=False), StringGen())], ['child1', IntegerGen()]])]
 parquet_array_gen = [ArrayGen(sub_gen, max_length=10) for sub_gen in parquet_basic_gen + parquet_struct_gen] + \
                     [ArrayGen(ArrayGen(sub_gen, max_length=10), max_length=10) for sub_gen in parquet_basic_gen + parquet_struct_gen]
-parquet_write_gens_list = [parquet_basic_gen + parquet_struct_gen + parquet_array_gen +
-                           [decimal_gen_default,
-                           decimal_gen_scale_precision, decimal_gen_same_scale_precision, decimal_gen_64bit]]
 
+parquet_write_gens_list0 = parquet_basic_gen + parquet_struct_gen + parquet_array_gen + parquet_decimal_gens
+parquet_map_gens = map_gens_sample + [MapGen(StructGen([['child0', StringGen()], ['child1', StringGen()]], nullable=False), FloatGen())]
+# splitting it in two lists as the data is too big for a single parquet file in our unit tests
```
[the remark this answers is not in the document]
razajafri: There was an OOM but I am not able to reproduce it. I will revert to a single gen

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -105,29 +105,60 @@ object GpuParquetFileFormat {
...
           builder.withListColumn(
-            parquetWriterOptionsFromSchema(listBuilder(field.name),
-              StructType(Array(StructField(field.name, a.elementType, true))), writeInt96)
-              .build())
+            parquetWriterOptionsFromField(
+              listBuilder(name, nullable),
+              a.elementType,
+              name,
+              writeInt96,
+              true).build())
```
revans2, line 132: Why is this true when the others are nullable? A comment would be good.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -105,29 +105,60 @@ object GpuParquetFileFormat {
...
+        case m: MapType =>
+          builder.withMapColumn(
+            mapColumn(name,
+              parquetWriterOptionsFromField(
+                ParquetWriterOptions.builder(),
+                m.keyType,
+                "key",
+                writeInt96,
+                false).build().getChildColumnOptions()(0),
```
revans2, line 141: I am a little concerned with this in relation to https://github.com/rapidsai/cudf/issues/7654. The problem is even if the key column contains no nulls, if it has a validity buffer, that is enough to cause problems, and I don't know enough to guarantee that it will not have one.  Have we tested to see if we set this to true if Spark has issues with reading it?

As an alternative we could take the hit ourselves and write a cudf function that will walk through an input column view and prune off unneeded validity buffers.  That would let us call that before writing to parquet/etc. And would not add any extra overhead to the core cudf.

integration_tests/src/main/python/parquet_write_test.py
```diff
@@ -34,20 +34,19 @@
...
-                      StructGen([['child0', StructGen([[ 'child1', byte_gen]])]])]
+                      StructGen([['child0', StructGen([[ 'child1', byte_gen]])]]),
+                      StructGen([['child0', MapGen(StringGen(nullable=False), StringGen())], ['child1', IntegerGen()]])]
 parquet_array_gen = [ArrayGen(sub_gen, max_length=10) for sub_gen in parquet_basic_gen + parquet_struct_gen] + \
                     [ArrayGen(ArrayGen(sub_gen, max_length=10), max_length=10) for sub_gen in parquet_basic_gen + parquet_struct_gen]
-parquet_write_gens_list = [parquet_basic_gen + parquet_struct_gen + parquet_array_gen +
-                           [decimal_gen_default,
-                           decimal_gen_scale_precision, decimal_gen_same_scale_precision, decimal_gen_64bit]]
 
+parquet_map_gens = map_gens_sample + [MapGen(StructGen([['child0', StringGen()], ['child1', StringGen()]], nullable=False), FloatGen()), MapGen(StructGen([['child0', StringGen(nullable=True)]], nullable=False), StringGen())]
```
jlowe, line 48: `map_gens_sample` includes maps using a `TimestampGen` that isn't qualified and thus can produce some very old timestamps.  Earlier in this file there's code that is carefully limiting timestamp generation ranges because it can cause issues, so don't we need to worry about that here as well?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
razajafri: It's working fine after I set the key column to be nullable. 

I will take on the alternative as a follow-on if I can get to it this sprint because I know its come up many times.

razajafri: I apologize, I had put the nullable in the wrong place, its confusing when we have multiple nesting. Upon testing it again, spark doesn't like it and is throwing an exception like @revans2 suspected. 
```
E                     File "/home/rjafri/spark-3.1.1-bin-hadoop3.2/python/lib/pyspark.zip/pyspark/sql/types.py", line 1367, in verify_map
E                     File "/home/rjafri/spark-3.1.1-bin-hadoop3.2/python/lib/pyspark.zip/pyspark/sql/types.py", line 1408, in verify
E                     File "/home/rjafri/spark-3.1.1-bin-hadoop3.2/python/lib/pyspark.zip/pyspark/sql/types.py", line 1278, in verify_nullability
E                   ValueError: key of map field _c57: This field is not nullable, but got None

```
Are you OK with getting this in while I work on the cudf function?

revans2: No I am not.

razajafri: OK, I am working on a cudf PR. The changes will be in the TableJni.cpp, we can later move it to the core cudf-cpp if there is a need. Is that OK?

razajafri:
I have pushed a PR for the cudf side for removing null_mask if there are no nulls in a table [here](https://github.com/rapidsai/cudf/pull/9061). 

I am not exactly sure how to test this out as I haven't yet been successful in creating a column from python tests that can create a null_mask even though the column doesn't have nulls. We might need a scala test for it

revans2:
> I have pushed a PR for the cudf side for removing null_mask if there are no nulls in a table [here](https://github.com/rapidsai/cudf/pull/9061).
> 
> I am not exactly sure how to test this out as I haven't yet been successful in creating a column from python tests that can create a null_mask even though the column doesn't have nulls. We might need a scala test for it

The simplest way I know of is to create a column with nulls in it and then filter out all of the nulls.  copy_if_else is not smart enough to know that you are filtering on nulls so it will always still allocate a validity buffer.

ttnghia:
> I have pushed a PR for the cudf side for removing null_mask if there are no nulls in a table [here](https://github.com/rapidsai/cudf/pull/9061).
Maybe my comment here is too late since that PR has been merged: Why do we need to remove the null mask? Why don't we just check for null count and ignore the null mask if there is not any null? As per Arrow spec, the null mask can be optional with all-valid bit.

revans2:
> Maybe my comment here is too late since that PR has been merged: Why do we need to remove the null mask? Why don't we just check for null count and ignore the null mask if there is not any null? As per Arrow specs, the null mask can be optional with all-valid bit.

I filed an issue against cudf for that, but because of performance reasons they didn't want to actually check for a null count, but instead the presence of the validity buffer. This is here as a work around for that.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
razajafri: let me know if the comment for passing `nullable` param to the listBuilder and others is enough. We already documented why we are hard-coding nullable to true

razajafri:
I apologize for force-pushing, I had to because git was confused about the number of commits after I rebased

razajafri:
build

razajafri:
build

jlowe (approved):
> I apologize for force-pushing, I had to because git was confused about the number of commits after I rebased

Why was the PR being rebased?  That's typically going to require a force-push on its own.  Once a PR is posted for review, normally one would only add merge commits to bring a PR up to date rather than rebase the branch.

Issue #3177, opened 2021-08-10
[FEA] Support decimal type in ORC reader


firestarman:
cuDF has added decimal type support in ORC reader by the PR https://github.com/rapidsai/cudf/pull/7970.

So we can enable the decimal support on plugin side now. Maybe just updating the type signature can have this done.

https://github.com/NVIDIA/spark-rapids/issues/3192 is for ORC writer

sameerz:
Hi @firestarman can we split this into two issues, one for decimal reading, and one for writing?  Let's prioritize the reading work for 21.10 as P1.

firestarman:
> Hi @firestarman can we split this into two issues, one for decimal reading, and one for writing? Let's prioritize the reading work for 21.10 as P1.

Sure, and done.

PR #3239, opened 2021-08-17
Support decimal type in orc reader

firestarman:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3177

### DIFF
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
@@ -629,17 +629,6 @@ public class GpuColumnVector extends GpuColumnVectorBase {
    */
   static boolean typeConversionAllowed(ColumnView cv, DataType colType) {
     DType dt = cv.getType();
-    // We should honor the actual decimal type used for the cudf column first, then the precision.
-    // Instead of always inferring a decimal type from the precision of Spark. Because both
-    // DECIMAL32 and DECIMAL64 can support precisions <= MAX_INT_DIGITS. Cudf may use either one
-    // for this case. e.g. ORC reader will always read decimal as DECIMAL64, even the precision
-    // is suitable for DECIMAL32.
-    if (colType instanceof DecimalType) {
-      int p = ((DecimalType) colType).precision();
-      return DType.DTypeEnum.DECIMAL32.equals(dt.getTypeId()) && p <= Decimal.MAX_INT_DIGITS() ||
-             DType.DTypeEnum.DECIMAL64.equals(dt.getTypeId()) && p <= Decimal.MAX_LONG_DIGITS();
-    }
-
     if (!dt.isNestedType()) {
       return getNonNestedRapidsType(colType).equals(dt);
     }
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
@@ -630,7 +630,7 @@ public class GpuColumnVector extends GpuColumnVectorBase {
   static boolean typeConversionAllowed(ColumnView cv, DataType colType) {
     DType dt = cv.getType();
     // We should honor the actual decimal type used for the cudf column first, then the precision.
-    // Instead of always inferring a deciaml type from the precision of Spark. Because both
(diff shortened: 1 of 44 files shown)


firestarman:
build

firestarman:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
```diff
@@ -629,6 +629,17 @@ public static ColumnarBatch from(Table table, DataType[] colTypes) {
    */
   static boolean typeConversionAllowed(ColumnView cv, DataType colType) {
     DType dt = cv.getType();
+    // We should honor the actual decimal type used for the cudf column first, then the precision.
+    // Instead of always inferring a deciaml type from the precision of Spark. Because both
```
jlowe, line 633:
```suggestion
    // Instead of always inferring a decimal type from the precision of Spark. Because both
```

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
```diff
@@ -629,6 +629,17 @@ public static ColumnarBatch from(Table table, DataType[] colTypes) {
...
+    // DECIMAL32 and DECIMAL64 can support precisions <= MAX_INT_DIGITS. Cudf may use either one
+    // for this case. e.g. ORC reader will always read decimal as DECIMAL64, even the precision
```
jlowe, line 635: Should the plugin's ORC reader be downcasting columns to `DECIMAL32` in the case of excessively-precise decimals?  That could save space on operations downstream from the query.

firestarman:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
firestarman: Yes, it will be better to cast columns to `DECIMAL32` when the precision allows. However I think this is an improvement since it is not required for the functionality. 

I thought this for a while. Seems we need to iterate the columns (including nested columns) to find out the ones matching the case and do the casting. And compond columns need to be rebuilt with new cast children. The whole process looks like a rule for space optimization, which can be used in more places than just the ORC reader.
How about implementing it as an util method, instead of integrating it with ORC reader in another PR ?

sperlingxx:
LGTM

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
jlowe: I think we should be consistent with how the decimals are loaded.  The Parquet reader is already converting DECIMAL64 columns into DECIMAL32 columns on read when appropriate, so I think we should do the same here.  The code to convert is already written and can be refactored from the Parquet code as a utility method.

jlowe: The Parquet code for this is part of `ParquetPartitionReaderBase#evolveSchemaIfNeededAndClose`

firestarman:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
firestarman: Got it. Updated.

firestarman:
build

revans2 (approved):
Just a nit for me

jlowe (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala
```diff
@@ -381,6 +381,55 @@ trait OrcCommonFunctions extends OrcCodecWritingHelper {
...
+      t.getCategory match {
+        case TypeDescription.Category.DECIMAL => Seq(t.getPrecision)
+        case c if !c.isPrimitive => getPrecisionsList(t.getChildren.asScala)
+        case _ => Seq.empty[Int]
+      }
+    }
+  }
+
+  /**
+   * Cast columns with precision that can be stored in an int to DECIMAL32, to save space.
```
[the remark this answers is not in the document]
firestarman: Will add it in a following PR to avoid running the premerge again for just doc udpate to get new approvals.

PR #3242, opened 2021-08-17
Support for LeftOuter/BuildRight and RightOuter/BuildLeft nested loop joins

jlowe:
Part of #3111.

This adds GPU support for LeftOuter with BuildRight and RightOuter with BuildLeft broadcast nested loop joins.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
@@ -1,0 +1,311 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.nvidia.spark.rapids
+
+import scala.collection.mutable
+
+import ai.rapids.cudf.{GatherMap, NvtxColor}
+import com.nvidia.spark.rapids.RapidsBuffer.SpillCallback
+
+import org.apache.spark.TaskContext
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.vectorized.ColumnarBatch
+
+/**
+ * Base class for iterators producing the results of a join.
+ * @param gatherNvtxName name to use for the NVTX range when producing the join gather maps
+ * @param targetSize configured target batch size in bytes
+ * @param joinTime metric to record GPU time spent in join
+ * @param totalTime metric to record total time in the iterator
+ */
+abstract class AbstractGpuJoinIterator(
+    gatherNvtxName: String,
+    targetSize: Long,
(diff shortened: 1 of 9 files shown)


jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -0,0 +1,311 @@
...
+    val wasInitialJoin = isInitialJoin
+    isInitialJoin = false
+    if (pendingSplits.nonEmpty || stream.hasNext) {
+      val cb = if (pendingSplits.nonEmpty) {
+        withResource(pendingSplits.dequeue()) {
+          _.getColumnarBatch()
+        }
+      } else {
+        val batch = withResource(stream.next()) { lazyBatch =>
+          // TODO: Worth having a releaseBatch method for LazySpillableBatch?
```
revans2, line 185: Are you planning on doing this?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -192,104 +188,58 @@ class HashJoinIterator(
...
-      }
 
-      if (ret.isDefined) {
-        // We are about to return something. We got everything we need from it so now let it spill
-        // if there is more to be gathered later on.
-        gathererStore.foreach(_.allowSpilling())
-      }
-      ret
+  override def computeNumJoinRows(cb: ColumnarBatch): Long = {
+    // TODO: Replace this estimate with exact join row counts using the corresponding cudf APIs
```
revans2, line 214: Is there a follow on issue for this, or should it be done now?

revans2 (commented):
It looks good, just curious about the TODOs in the code

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
jlowe: I left this here to see if you thought that would be a good idea or not.  Seems a bit wasteful to walk through a column batch to increment refcounts only to turn around and decrement them all immediately again.  But it's probably not a significant performance problem either.  If you want me to add this here or part of a followup, happy to do so.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
jlowe: The Java APIs do not exist yet, so it can't be done now.  I'll update this TODO to point to the cudf issue tracking the Java bindings for this.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
revans2: I thought it might have been. For me it is all about cleanliness of the code, because like you said the performance difference is minimal. I think a `release` would make it simpler to follow what is happening.

jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
jlowe: I added a `batchRelease` method to lazy spillable.  I made it act semantically like a "get batch and close without closing the resulting batch".  That means for spill-only it does _not_ end up closing the underlying resource, since that seems to be the whole point of that class.  Let me know if this makes sense, or if the semantics are more confusing than the incref approach that was there before.

revans2 (dismissed):

jlowe:
Had to resolve a conflict with yesterday's AST API update.

jlowe:
build

revans2 (approved):

sameerz:
build

Issue #3244, opened 2021-08-17
[BUG] join_test LeftAnti failing on Databricks


tgravescs:
**Describe the bug**
 FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-String]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Byte]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Short]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Integer]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Long]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Boolean]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Date]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Timestamp]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Null]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Float][INCOMPAT]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Double][INCOMPAT]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Decimal(18,0)]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Decimal(7,3)]
16:54:32  FAILED ../../src/main/python/join_test.py::test_broadcast_nested_loop_join_condition_missing_count[LeftAnti-Decimal(7,7)]
[...]

PR #3246, opened 2021-08-17
Disable test_broadcast_nested_loop_join_condition_missing_count on Databricks

jlowe:
Relates to #3244 

The new test_broadcast_nested_loop_join_condition_missing_count test is failing due to what appears to be a bug in hash aggregate code.  It looks like we're trying to pull out an input reference from a batch that has no columns.  Thus this looks more like a latent bug in hash aggregate on Databricks that was exposed in by the new join code rather than a bug in the new join code itself.

To unblock the build while more investigation into the hash aggregate issue occurs, this xfails the test on Databricks and references #3244.


jlowe:
build

tgravescs (approved):

Issue #3250, opened 2021-08-18
[FEA] Make CreateMap duplicate key handling compatible with Spark and enable CreateMap by default


andygrove:
**Is your feature request related to a problem? Please describe.**
In https://github.com/NVIDIA/spark-rapids/issues/3229 we added support for CreateMap with multiple key-value pairs e.g. `map(key1, value1, key2, value2)`. We had to add a config flag to enable this option because our handling of duplicate values is not compatible with Spark.

**Describe the solution you'd like**
It would be nice if we could make duplicate handling compatible with Spark and enable CreateMap by default. See conversation at https://github.com/NVIDIA/spark-rapids/pull/3230#discussion_r689641323 for more information.

**Describe alternatives you've considered**
None.

**Additional context**
N/A

revans2:
I closed rapidsai/cudf#9123 because after discussion on the issue we determined that we can do this today. It might not be the most optimal, but it should work.  The following pseudo-ish code should do it.

```
withResource(extractKeysFrom(mapCol)) { keyListView =>
    val dedupeCountCol = withResource(keyListView.dropDuplicates()) { dupsDropped =>
        dupsDropped.countElements()
    }
    val sameLengthCol = withResource(dedupeCountCol) { dedupeCountCol =>
        withResource(keyListView.countElements()) { origCount =>
           origCount.equals(dedupeCountCol)
        }
    }
    withResource(sameLengthCol) { sameLengthCol =>
        withResource(sameLengthCol.all()) { allSame =>
             allSame.getBoolean()
        }
    }
}
```

PR #3251, opened 2021-08-18
CreateMap support for multiple key-value pairs

andygrove:
Builds on https://github.com/NVIDIA/spark-rapids/pull/3230.

Closes https://github.com/NVIDIA/spark-rapids/issues/3229.

There is a follow-on issue https://github.com/NVIDIA/spark-rapids/issues/3250 for making duplicate handling compatible with Spark so that we can remove the config option and enable CreateMap by default.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -637,6 +637,14 @@ object RapidsConf {
       .booleanConf
       .createWithDefault(false)
 
+  val ENABLE_CREATE_MAP = conf("spark.rapids.sql.createMap.enabled")
+    .doc("When set to true, support the CreateMap expression on the GPU with multiple " +
+      "key-value pairs. The GPU version does not detect duplicate keys or make any guarantees " +
+      "about which key wins if there are duplicates. CreateMap is always supported on the GPU" +
```
revans2, line 643: missing a space at the end of this line.

integration_tests/src/main/python/map_test.py
```diff
@@ -32,6 +38,39 @@ def test_simple_get_map_value(data_gen):
                 'a["NOT_FOUND"]',
                 'a["key_5"]'))
 
+@pytest.mark.parametrize('key_gen', [StringGen(nullable=False), IntegerGen(nullable=False), basic_struct_gen], ids=idfn)
+@pytest.mark.parametrize('value_gen', [StringGen(nullable=False), IntegerGen(nullable=False), basic_struct_gen], ids=idfn)
```
revans2, line 42: Why no nulls on the values?

integration_tests/src/main/python/map_test.py
```diff
@@ -32,6 +38,39 @@ def test_simple_get_map_value(data_gen):
...
+                'map(a, b) as map2'))
+
+def test_map_expr_no_pairs():
+    data_gen = [('a', StringGen(nullable=False)), ('b', StringGen(nullable=False))]
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : gen_df(spark, data_gen).selectExpr(
+                'map() as m1'))
+
+def test_map_expr_multiple_pairs():
+    data_gen = [('a', StringGen(nullable=False)), ('b', StringGen(nullable=False))]
```
revans2, line 57: We should add some comments here that because of the high number of values possible in StringGen that in this test duplicate keys never show up.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -983,6 +983,32 @@ object WindowSpecCheck extends ExprChecks {
   }
 }
 
+object CreateMapCheck extends ExprChecks {
+
+  // Spark supports all types except for Map for key and value (Map is not supported
+  // even in nested types)
+  val keyValueSig: TypeSig = (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64 +
```
revans2, line 990: So we are not going to support a map as a value in a map? I understand the key, but why also the value?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -983,6 +983,32 @@ object WindowSpecCheck extends ExprChecks {
...
+    TypeSig.ARRAY + TypeSig.STRUCT).nested()
+
+  override def tag(meta: RapidsMeta[_, _, _]): Unit = {
+    if (meta.childExprs.length > 2 && !meta.conf.isCreateMapEnabled) {
```
revans2, line 994: Do we want to check if all of the values are literal values, and if so allow it through, assuming there are no duplicates?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -82,6 +82,65 @@ case class GpuCreateArray(children: Seq[Expression], useStringTypeWhenEmpty: Boo
...
+      children.indices.foreach { index =>
+        columns(index) = GpuExpressionsUtils.columnarEvalToColumn(children(index), batch).getBase
+      }
+
+      if (children.length == 2) {
+        withResource(ColumnVector.makeStruct(columns: _*)) { struct =>
+          GpuColumnVector.from(ColumnVector.makeList(numRows, DType.STRUCT, struct), dataType)
+        }
+      } else {
+        val structs = Range(0, columns.length, 2)
```
revans2, line 114: Does this code not work if `children.length == 2`? It looks like it would and then we could remove the conditional and the extra code.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
andygrove: Sure.  Could you give me some pointers on how to check for duplicates using cuDF? I assume I would have to concat the key columns first but I am not sure where to go from there.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
andygrove: Yes, this code works in both cases. I will remove the special case.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
andygrove: Never mind ... these are literal values in the expression tree ... got it

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
andygrove: Thanks. I had misread the Spark documentation around this. I will add separate signatures for key and value.

integration_tests/src/main/python/map_test.py
ttnghia: And also the keys?

integration_tests/src/main/python/map_test.py
ttnghia: So is there a way/parameter to force generating duplicates?

revans2: We would have to force StringGen for both a and b to be much more restrictive.

```
 data_gen = [('a', StringGen('duplicate', nullable=False)), ('b', StringGen('duplicate', nullable=False))]
```

Now the only value that will ever be in the 'a' or 'b' column is 'duplicate'. Not that we want to do that, because the test would fail with the CPU run throwing an exception that there are duplicate keys.

integration_tests/src/main/python/map_test.py
andygrove: Spark doesn't allow nulls for the keys. The tests have been updated to allow nulls in the values.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -637,6 +637,15 @@ object RapidsConf {
       .booleanConf
       .createWithDefault(false)
 
+  val ENABLE_CREATE_MAP = conf("spark.rapids.sql.createMap.enabled")
+    .doc("When set to true, support the CreateMap expression on the GPU with multiple " +
```
revans2, line 641: nit: Could we try to reword this a bit? I'm not sure we need all of the details, but I am fine if others disagree.

```
The GPU enabled version of the CreateMap expression, `map` SQL function, does not detect duplicate keys in all 
cases and does not guarantee which key wins if there are duplicates. When this config is set to true `CreateMap`
will be enabled to run on the GPU even in cases where there might be duplicate keys.
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -1103,20 +1108,34 @@ object CreateMapCheck extends ExprChecks {
...
       meta.willNotWorkOnGpu(s"this is not supported in the $context context")
     } else {
-      if (meta.childExprs.length != 2) {
-        // See https://github.com/NVIDIA/spark-rapids/issues/3229
-        meta.willNotWorkOnGpu("CreateMap only supports two expressions on GPU")
+      // if there are more than two key-value pairs then there is the possibility of duplicate keys
+      if (meta.childExprs.length > 2) {
+        // check for duplicate keys if the keys are literal values
+        val keyExprs = meta.childExprs.indices.filter(_ % 2 == 0).map(meta.childExprs)
+        if (keyExprs.forall(_.wrapped.isInstanceOf[Literal])) {
```
revans2, line 1115: Could you use `GpuOverrides.extractLit` instead? It handles cases where ti might be wrapped in an alias, which we have seen before.

revans2 (dismissed):
Just a nit

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -1112,7 +1112,8 @@ object CreateMapCheck extends ExprChecks {
       if (meta.childExprs.length > 2) {
         // check for duplicate keys if the keys are literal values
         val keyExprs = meta.childExprs.indices.filter(_ % 2 == 0).map(meta.childExprs)
-        if (keyExprs.forall(_.wrapped.isInstanceOf[Literal])) {
+        if (keyExprs.forall(e => GpuOverrides.extractLit(
+            e.wrapped.asInstanceOf[Expression]).isDefined)) {
           val keys = keyExprs.map(_.wrapped.asInstanceOf[Literal].value)
```
revans2, line 1117: Sorry I meant to use it in both places.  Something more like

```
val litKeys = keyExprs.map(e => GPuOverrides.extractLit(e.wrapped))
if (litKeys.forall(_.isDefined)) {
  val keys = litKeys.map(_.get.value)
...
```

revans2 (approved):

andygrove:
build

Issue #3224, opened 2021-08-12
[FEA] Add maven compile/package plugin executions, one for each supported Spark dependency version


gerashegalov:
**Is your feature request related to a problem? Please describe.**
This feature contributes to #3223. Each version-independent module. i.e. all the source code outside the shim parent module, should produce a separate specifier-differentiated artifact  using a specific spark artifact version as a dependency

**Describe the solution you'd like**
1. Add non-default profiles, 1 per shim version, consistently named across modules such that they can be activated consistently across modules. Define properties that can activated related profiles at the same time, e.g. -Dreleased31x activates all 3.1.x, reuse existing properties if possible.
2. In each profile configure scala compile plugin execution with the spark.version overridden correspondingly, it should use a dedicated class output directory that is then consumed by a packaging plugin to create an artifact such as `sql-plugin/target/rapids-4-spark-sql_2.12-21.10.0-SNAPSHOT-spark312.jar` if we take sql-plugin module as example. Compilation fixed is necessary to make source build against the target version.
3. If there is a dependency from on plugin submodule A to another B this should now use specifiers to make sure that the dependency is consistent with respect to the Spark version.
3. To make sure that the current build is not disrupted the main artifact (without the shim specifier) is still manufactured with spark301.version 

**Describe alternatives you've considered**
 CI  scripts can be run multiple times, once per Spark version . For each supported release 3.x.y of Spark run the full build with 
mvn -Dspark.version=3.x.y …
After each iteration extract the spark-rapids uber jar into a separate directory. This solution has a number of drawbacks:
- on the developer machines, we might accidentally install "wrong" versions in the local repo. 
- manual steps to remember
- need to make sure that each module actually allows overriding spark versions.

Breakdown of tasks:

- [x] Initial build framework - Spark 3.0.1 and spark3.0.2
- [x] Add other Apache Spark build framework and copy shims - Spark 3.0.x, 3.1.x, cdh, emr - https://github.com/NVIDIA/spark-rapids/pull/3277
- [x] Add build framework for databricks 3.0.1 and 3.1.1 shims
- [x] Add build framework for Spark 3.2.0
- [ ] Optionally add build framework for databricks 8.3 runtime (3.1.2+)

tgravescs:
[...]

PR #3257, opened 2021-08-19
Add maven compile/package plugin executions for Spark302 and Spark301

tgravescs:
start of https://github.com/NVIDIA/spark-rapids/issues/3224

Initial pr that aims to leave existing build the way it is but adds in some profiles to build all the source against a particular version of Spark. In this case added profiles for spark301 and spark302. You are meant to only build against a single version of Spark at a time.  Eventually we will have build scripts to build each version independently and install them and then the dist package will pull the installed versions by classifier.

Currently this is setup to build with 2 profiles: -Prelease302,with-classifier

The with-classifier profile changes the dependencies to look for the classifier version of jars and the release302 set the versions to build spark 302.  it was done this way to minimize the code needed in each pom. The with-classifier profile can be reused against all the spark versions.

This creates jars with a classifier in each of the submodules, so the sql jar looks like:

You can also specify the property instead of the profiles like:  -Dbuildver={302/301} and that will activate the profiles automatically.

rapids-4-spark-sql_2.12-21.10.0-SNAPSHOT-spark302.jar
rapids-4-spark-sql_2.12-21.10.0-SNAPSHOT-spark302tests.jar

The dist jar is the only one that stays without classifier.  We can discuss the integration test jar more as needed, right now we only deploy one version which is ok as long as we don't have spark dependent api which it likely doesn't because it was using public api for most things.  So I can easily remove the classifier version there if we decide its not needed.

the dist/pom.xml was a bit weird where I had to duplicate all dependencies with profile to get it to pull them into final jar properly for some reason. I'll investigate this more, but for now copying it like this pulls everything and contents are the same between the version built with classifier and version now (except for the shim layers not pulled in obviously)

We could go down to 1 profile but would just require duplicating code. I was looking for a way to activate multiple profiles with a property but didn't see a good way to do it with duplicating code again.  

The other change here is to copy all the shim layer spark301 code into the spark302 code.  I diffed them and the only differences is in the package names and copyrights.  Once we have all the builds using the new system we can commonize the code possible from the shims.


tgravescs:
build

tgravescs:
build

tgravescs:
build looks like it was just terminated, rekicking

tgravescs:
build

tgravescs:
it seems to be trying to build dist before other modules, which doesn't happen for me locally, investigating.

tgravescs:
build

dist/pom.xml
```diff
@@ -29,40 +29,150 @@
...
+               <groupId>org.apache.spark</groupId>
+               <artifactId>spark-sql_${scala.binary.version}</artifactId>
+               <scope>provided</scope>
+            </dependency>
+            <dependency>
+              <!-- required for conf generation script -->
+              <groupId>org.apache.spark</groupId>
+              <artifactId>spark-hive_${scala.binary.version}</artifactId>
+              <scope>provided</scope>
+            </dependency>
```
jlowe, line 70: Seems like we should be able to move these out of the profiles and into a common `dependencies` section since they're not changing between profiles

dist/pom.xml
```diff
@@ -29,40 +29,150 @@
...
+                                <executable>bash</executable>
+                                <commandlineArgs>-c 'export MODIFIED=$(git status --porcelain | grep "^ M"); [[ -z $MODIFIED ]] &amp;&amp; exit 0 || { echo -e "found modified files during mvn verify:\n$MODIFIED"; exit 1;}'</commandlineArgs>
+                              </configuration>
+                        </execution>
+                    </executions>
+                </plugin>
+            </plugins>
+        </build>
+        <activation>
+            <activeByDefault>false</activeByDefault>
```
jlowe, line 171: Nit: Necessary?  It's missing for the `with-classifier` profile, and we should be consistent.

integration_tests/pom.xml
```diff
@@ -67,21 +67,69 @@
...
+                    <groupId>com.nvidia</groupId>
+                    <artifactId>rapids-4-spark-tests_${scala.binary.version}</artifactId>
+                    <version>${project.version}</version>
+                    <classifier>${spark.version.classifier}tests</classifier>
+                    <type>test-jar</type>
+                    <scope>test</scope>
+                </dependency>
+                <dependency>
+                    <groupId>com.nvidia</groupId>
+                    <artifactId>rapids-4-spark-udf-examples_${scala.binary.version}</artifactId>
```
jlowe, line 107: While we are technically using Spark Catalyst classes in this module, like `UDF1`, All of the APIs should be marked as stable.  This module is showing examples of UDF code users could write against Spark, so I would expect those APIs to be stable.  Therefore we could theoretically remove the classifier on this module and just compile it once against the minspec Spark version, for example.  This is a jar we're likely to publish separately, and it would be nice not to have a Spark classifier on it if it's unnecessary.

dist/pom.xml
```diff
@@ -29,40 +29,150 @@
...
+            <dependency>
+              <!-- required for conf generation script -->
+              <groupId>org.apache.spark</groupId>
+              <artifactId>spark-hive_${scala.binary.version}</artifactId>
+              <scope>provided</scope>
+            </dependency>
+        </dependencies>
+    </profile>
+    <profile>
+        <id>with-classifier</id>
```
jlowe, line 74: I wonder if we can get down to one thing on the command-line that activates multiple profiles automatically, e.g.: use a property and then have behavior like this based on that property:

- `with-classifier` automatically turns on if the property is defined regardless of the value
- `release301` only turns on if the property is defined and equal to `301`
- `release302` only turns on if the property is defined and equal to `302`
- etc.

shims/pom.xml
```diff
@@ -32,22 +32,82 @@
...
+                <module>spark312</module>
+                <module>spark311cdh</module>
+                <module>aggregator</module>
+            </modules>
+        </profile>
+        <profile>
+            <id>release301</id>
+            <modules>
+                <module>spark301</module>
+                <module>aggregator</module>
```
jlowe, line 55: Can this be moved out to a top-level `modules` regardless of profile?  Seems we would always want to traverse that child module.

shuffle-plugin/pom.xml
```diff
@@ -65,6 +65,32 @@
...
 
+    <profiles>
+       <profile>
+            <id>with-classifier</id>
+            <dependencies>
+                <dependency>
+                    <groupId>com.nvidia</groupId>
+                    <artifactId>rapids-4-spark-sql_${scala.binary.version}</artifactId>
+                    <version>${project.version}</version>
+                    <classifier>${spark.version.classifier}</classifier>
```
jlowe, line 76: We don't need to worry about the base dependencies of this module also trying to pull in rapids-4-spark-sql without a classifier?  I expected to see a `default` vs `with-classifier` profile setup like the other poms.  This comment applies to a lot of other poms doing the same thing below.

dist/pom.xml
tgravescs: I made a comment on this in pr description, I agree seems like we should but it wasn't working for some reason. I'm going to investigate more but I've run into that a couple places where it doesn't seem to properly override. In this case it wasn't pulling all the dependencies. I'm going to investigate more but was hoping it wouldn't block this.

dist/pom.xml
tgravescs: if you know how to do turns on regardless of value let me know. I attempted something but it didn't work.  Was planning on trying  again, couldn't find docs on it

tgravescs: oh you know what I think I see how to now

dist/pom.xml
tgravescs: I'll remove

dist/pom.xml
jlowe: Yep, definitely not a blocker.  Just surprised this isn't working since we rely on the "profile adds dependencies to base dependencies" behavior elsewhere.

shuffle-plugin/pom.xml
tgravescs: will make it consistent

tgravescs:
addressed review comments, you can now just build with mvn  -Dbuildver={302/301} and it will activate with-classifier and the release302/301.

tgravescs:
build

tgravescs:
build

jlowe:
I think this is close.  Looks like some code needs to be updated in the new 302 copies of code, hopefully we can come up with a better way to do this to significantly reduce the amount of duplicated code.

Also seeing this in the build log which would be good to fix:
```
13:23:22  [WARNING] 
13:23:22  [WARNING] Some problems were encountered while building the effective model for com.nvidia:rapids-4-spark-tools_2.12:jar:21.10.0-SNAPSHOT
13:23:22  [WARNING] 'build.plugins.plugin.version' for org.scala-tools:maven-scala-plugin is missing. @ line 84, column 21
13:23:22  [WARNING] 
13:23:22  [WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
13:23:22  [WARNING] 
13:23:22  [WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
13:23:22  [WARNING] 
```

jlowe:
Hmm, looks like the tools module issue is unrelated, I'll fix in a separate PR.

abellina:
Last CI failure seems relevant. Otherwise LGTM and the process to add a new shim is straightforward. +1 on finding a way to remove duplication of code, but as far as I understand that's a separate stage and it could break the current build if done wholesale.

tgravescs:
yeah its weird, its like the 3.2 shim isn't picking up the dependency on sql plugin it all works locally for me though.

tgravescs:
build

tgravescs:
build

tgravescs:
Note, I'm trying to debug the pre-merge build here so I'll end up reverting these last few commits. The failure doesn't make any sense and can't reproduce locally.

tgravescs:
build

tgravescs:
ok, the issue is some 301 shim layer code got checked in and I need to copy to 302.

tgravescs:
build

pxLi:
build

pxLi:
last CI failure was due to one java gateway (gw1) crashed, passed on retry

tgravescs:
@jlowe could you take another look?  I think the main updates since you looked last were to shim layer where upmerged and copied to the 302 shim

jlowe (approved):

PR #3259, opened 2021-08-19
Generate docs for AST from checks

### LINKED ISSUE #3172: [DOC] Generated documentation for AST expression support
With the addition of AST expression support in #3165, we need the ability to generate documentation describing what is supported by an AST expression.  We will also have to consider how best to present this in a way that doesn't confuse users with respect to normal GPU expression support.

revans2:
This fixes #3172 

Please take a close look. I think I matched everything correctly, but it would be good to have a second pair of eyes check it over.

This updates explain for operations that require AST so that the expressions that would not work with AST are shown in the standard place and not repeated in the reason why the parent operator cannot work.

This drops the lambda expression context entirely.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
@@ -540,7 +540,7 @@ abstract class SparkPlanMeta[INPUT <: SparkPlan](plan: INPUT,
 
   def tagForExplain(): Unit = {
     if (!canThisBeReplaced) {
-      childExprs.foreach(_.recursiveSparkPlanPreventsRunningOnGpu)
+      childExprs.foreach(_.recursiveSparkPlanPreventsRunningOnGpu())
       childParts.foreach(_.recursiveSparkPlanPreventsRunningOnGpu())
       childScans.foreach(_.recursiveSparkPlanPreventsRunningOnGpu())
       childDataWriteCmds.foreach(_.recursiveSparkPlanPreventsRunningOnGpu())
@@ -555,6 +555,12 @@ abstract class SparkPlanMeta[INPUT <: SparkPlan](plan: INPUT,
   }
 
   def requireAstForGpuOn(exprMeta: BaseExprMeta[_]): Unit = {
+    // willNotWorkOnGpu does not deduplicate reasons. Most of the time that is fine
+    // but here we want to avoid adding the reason twice, because this method can be
+    // called multiple times, and also the reason can automatically be added in if
+    // a child expression would not work in the non-AST case either.
+    // So only add it if canExprTreeBeReplaced changed after requiring that the
+    // given expression is AST-able.
     val previousExprReplaceVal = canExprTreeBeReplaced
     exprMeta.requireAstForGpu()
     val newExprReplaceVal = canExprTreeBeReplaced
@@ -780,7 +786,9 @@ object ProjectExprContext extends ExpressionContext {
  * not. This part is not done automatically.
  */
 object AstExprContext extends ExpressionContext  {
-  override def toString: String = "ast"
+  override def toString: String = "AST"
+
+  val notSupportedMsg = "this expression does not support AST"
 }
 object GroupByAggExprContext extends ExpressionContext  {
   override def toString: String = "aggregation"
@@ -819,7 +827,6 @@ object ExpressionContext {
 
   def getRegularOperatorContext(meta: RapidsMeta[_, _, _]): ExpressionContext = meta.wrapped match {
     case _: Expression if meta.parent.isDefined => getRegularOperatorContext(meta.parent.get)
-    // TODO do we want to look for joins/etc here and check the join type???
     case _ => ProjectExprContext
   }
 }
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
@@ -20,7 +20,7 @@ import java.time.ZoneId
 
 import scala.collection.mutable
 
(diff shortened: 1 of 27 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -554,6 +554,16 @@ abstract class SparkPlanMeta[INPUT <: SparkPlan](plan: INPUT,
     childPlans.foreach(_.tagForExplain())
   }
 
+  def requireAstForGpuOn(exprMeta: BaseExprMeta[_]): Unit = {
+    val previousExprReplaceVal = canExprTreeBeReplaced
+    exprMeta.requireAstForGpu()
+    val newExprReplaceVal = canExprTreeBeReplaced
```
jlowe, line 560: This code is a little confusing, I think a comment would help.  I'm assuming this is covering the case where something can be translated to an AST but not translated in the general sense?  Otherwise I would expect this to just call `canExprTreeBeReplaced` after requiring AST.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -801,8 +818,8 @@ object ExpressionContext {
   }
 
   def getRegularOperatorContext(meta: RapidsMeta[_, _, _]): ExpressionContext = meta.wrapped match {
-    case _: LambdaFunction => LambdaExprContext
     case _: Expression if meta.parent.isDefined => getRegularOperatorContext(meta.parent.get)
+    // TODO do we want to look for joins/etc here and check the join type???
```
jlowe, line 822: I'm not sure I follow how the join type matters to the AST expression. It's certainly critical for the join to know whether the condition expression can be translated to an AST, but I'm not following why the expr meta needs to check for joins here and why expression behavior would change based on join type.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -937,6 +980,15 @@ case class ExprChecksImpl(contexts: Map[ExpressionContext, ContextChecks])
         (expContext, check.support(dataType))
     }
   }
+
+  override def tagAst(exprMeta: BaseExprMeta[_]): Unit = {
+    val checks = contexts.get(AstExprContext)
+    if (checks.isEmpty) {
+      exprMeta.willNotWorkInAst(s"this is not supported by AST")
```
jlowe, line 987: Nit: Sometimes we use "this is not supported by AST" and sometimes "AST is not supported".  We should be consistent.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -1195,20 +1270,28 @@ class CastChecks extends ExprChecks {
     case TypeEnum.UDT => (udtChecks, sparkUdtSig)
   }
 
+  override def tagAst(meta: BaseExprMeta[_]): Unit = {
+    meta.willNotWorkInAst("AST is not support by cast")
```
jlowe, line 1274: "is not support by".  This probably should use the same consistent message per comment above, whichever it is.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -1660,9 +1770,21 @@ object SupportedOpsDocs {
...
+    println("on the CPU and an appropriate set of GPU kernels are selected to do those")
+    println("operations. For example `a >= b + 1`. Would result in calling a GPU kernel to add")
+    println("`1` to `b`, followed by another kernel that is called to compare `a` to that result.")
+    println("The interpretation is happening on the CPU, and the GPU is used to do the processing.")
+    println("For AST the interpretation for some reason cannot happen on the CPU and instead must")
+    println("be done in the GPU kernel itself. An example of this is conditional joins. If you")
+    println("want to join on `A.a >= B.b + 1` where `A` and `B` are separate tables or data")
+    println("frames, the `+` and `>=` operations cannot run as separate independent kernels")
+    println("because it is done on a combination of rows in both `A` and `B`. Instead part of the")
+    println("plan that Spark developed is turned into an abstract syntax tree and set to the GPU")
```
jlowe, line 1785:
```suggestion
    println("plan that Spark developed is turned into an abstract syntax tree and sent to the GPU")
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
```diff
@@ -38,16 +38,25 @@ class GpuProjectExecMeta(
...
     val gpuExprs = childExprs.map(_.convertToGpu()).toList
     val gpuChild = childPlans.head.convertIfNeeded()
     if (conf.isProjectAstEnabled) {
-      childExprs.foreach(_.tagForAst())
       if (childExprs.forall(_.canThisBeAst)) {
         return GpuProjectAstExec(gpuExprs, gpuChild)
       }
+      // explain AST because this is optional and it is sometimes hard to debug
+      val exp = conf.explain
+      if (!exp.equalsIgnoreCase("NONE")) {
```
jlowe, line 53: Nit: We should probably put the logic for interpreting the various values of the `explain` conf in `RapidsConf` as separate, parameterless predicate functions rather than pepper the logic everywhere the conf is used.  Might be good for a followup issue.

jlowe (commented):
Comments from the first pass, still need to go through in greater detail.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
revans2: Sorry this was left over from me trying to decide if we try to auto detect the `AstContext` or treat it totally separately. I went with totally separately.

revans2:
build

PR #3079, opened 2021-07-29
Remove the struct support in ORC reader

firestarman:
We do this because of the issue https://github.com/NVIDIA/spark-rapids/issues/3059, which seems not to be easy to fix in a short time.

closes https://github.com/NVIDIA/spark-rapids/issues/3059 

Signed-off-by: Firestarman <firestarmanllc@gmail.com>

PR #3262, opened 2021-08-20
Re-enable the struct support for the ORC reader.

firestarman:
This fixes https://github.com/NVIDIA/spark-rapids/issues/2879 and https://github.com/NVIDIA/spark-rapids/issues/1481.

It also adds tests for the nested predicate pushdown, and the support for nested column pruning.

The old relevant PRs:
  https://github.com/NVIDIA/spark-rapids/pull/3079
  https://github.com/NVIDIA/spark-rapids/pull/2887


firestarman:
build

firestarman:
There is still an issue in CUDF for a coner case when reading a struct column, and here is the fix https://github.com/rapidsai/cudf/pull/9060.

integration_tests/src/main/python/orc_test.py
```diff
@@ -323,3 +342,34 @@ def test_missing_column_names_filter(spark_tmp_table_factory, reader_confs):
...
+    all_confs.update({'spark.sql.sources.useV1SourceList': v1_enabled_list,
+        'spark.sql.optimizer.nestedSchemaPruning.enabled': nested_enabled})
+    # This is a hack to get the type in a slightly less verbose way
+    rs = StructGen(read_schema, nullable=False).data_type
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : spark.read.schema(rs).orc(data_path),
+            conf=all_confs)
+
+
+# This is for the corner case of reading only a strcut column that has no nulls.
```
jlowe, line 365:
```suggestion
# This is for the corner case of reading only a struct column that has no nulls.
```

integration_tests/src/main/python/orc_test.py
```diff
@@ -323,3 +342,34 @@ def test_missing_column_names_filter(spark_tmp_table_factory, reader_confs):
...
+# Then there will be no streams in a stripe connecting to this column(Its ROW_INDEX
```
jlowe, line 366:
```suggestion
# Then there will be no streams in a stripe connecting to this column (Its ROW_INDEX
```

integration_tests/src/main/python/orc_test.py
```diff
@@ -323,3 +342,34 @@ def test_missing_column_names_filter(spark_tmp_table_factory, reader_confs):
...
+# streams have been pruned by the Plugin.), and CUDF throws an exception for such case.
+@pytest.mark.xfail(reason='https://github.com/rapidsai/cudf/issues/8878')
```
jlowe, line 368: Don't we need to wait until the cudf issue is fixed before checking this in?  The test below shows there's nothing particularly fancy being done here to trigger the failure.

firestarman: OK, let's wait for the fix being merged.

firestarman:
Move to draft because of the open blocking issue https://github.com/rapidsai/cudf/pull/9060.

firestarman:
build

firestarman:
build

integration_tests/src/main/python/orc_test.py
firestarman: Updated.

integration_tests/src/main/python/orc_test.py
firestarman: Nice catch. Updated.

jlowe (commented):
Looks good to me, just waiting on the cudf issue.

firestarman:
build

firestarman:
The fix rapidsai/cudf#9060 has been merged.

jlowe (approved):

integration_tests/src/main/python/orc_test.py
```diff
@@ -323,3 +342,34 @@ def test_missing_column_names_filter(spark_tmp_table_factory, reader_confs):
...
+    rs = StructGen(read_schema, nullable=False).data_type
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : spark.read.schema(rs).orc(data_path),
+            conf=all_confs)
+
+
+# This is for the corner case of reading only a struct column that has no nulls.
+# Then there will be no streams in a stripe connecting to this column (Its ROW_INDEX
+# streams have been pruned by the Plugin.), and CUDF throws an exception for such case.
+# Here is the tracking issue: 'https://github.com/rapidsai/cudf/issues/8878'.
```
revans2, line 368: https://github.com/rapidsai/cudf/issues/8878 is closed and the test is not marked as xfail or trying to fall back to the CPU.  Could you update the comment so it explains that cudf used to fail, and this is to ensure the corner case remains tested.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala
```diff
@@ -393,6 +393,9 @@ trait OrcCommonFunctions extends OrcCodecWritingHelper {
 
   /**
    * Cast columns with precision that can be stored in an int to DECIMAL32, to save space.
+   * Besides the plugin makes the assumption that if the precision is small enough to fit
+   * in a DECIMAL32, then CUDF has it stored as a DECIMAL32. Getting this wrong may lead
```
revans2, line 397: nit: This is not 100% accurate.  In the plugin we enforce that if the precision is small enough to fit in a DECIMAL32 then it will be stored in a DECIMAL32. We don't assume that CUDF has stored it that way, as we tell CUDF what to do and what types to use.  It just happens that in some cases, like ORC and Parquet we don't have the ability to tell them what to do, so we can end up needing to fix it after the fact.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala
```diff
@@ -1187,19 +1190,48 @@ private case class GpuOrcFileFilterHandler(
...
+     *
+     * @param fileSchema input file's ORC schema
+     * @param readSchema ORC schema for what will be read
+     * @return true if they are compatible, otherwise false
+     */
+    private def isSchemaCompatible(
+        fileSchema: TypeDescription,
+        readSchema: TypeDescription): Boolean = {
+      fileSchema == readSchema ||
+        fileSchema != null && readSchema != null &&
```
revans2, line 1226: nit: It would be nice to include some parentheses so that there is no confusion about the order of operations.

```
(fileSchema == readSchema) ||
    (fileSchema != null && readSchema != null && ...)
```

revans2 (approved):

revans2:
All the code looks fine, just a few minor nits that I can live without.

PR #3241, opened 2021-08-17
Split pytest by 'slow_test' tag and run from different k8s pods to reduce premerge job duration

zhanga5:
part of https://github.com/NVIDIA/spark-rapids/issues/2731

split pytests by 'slow_test' tag in order to balance test duration and memory consumption from two running pods in parallel after identify two .py files that required long duration or big memory. Now premerge overall duration could be reduced to ~1:30h from ~2:30h

PR #3264, opened 2021-08-20
Add parallel support to nightly spark standalone tests

### LINKED ISSUE #2802: Update jenkins/start-tests.sh to use integration_tests/run_pyspark_from_build.sh
As title, sequel to https://github.com/NVIDIA/spark-rapids/pull/2792

pxLi:

fix #2802

To speed our nightly spark standalone integrations tests,
spark 3.0.x total time: ~3h 40m to ~1h 15m
spark 3.1.x total time: ~4h to ~1h 35m (include extra ParquetCachedBatchSerializer cache_test)

I am still doing more verification on other scenarios, submit first to collect feedbacks, thanks!

This requires to be enabled in nightly pipelines' Jenkinsfile separately

### DIFF
--- a/jenkins/spark-tests.sh
+++ b/jenkins/spark-tests.sh
@@ -87,7 +87,7 @@ export BASE_SPARK_SUBMIT_ARGS="$BASE_SPARK_SUBMIT_ARGS \
 --conf spark.executor.extraJavaOptions=-Duser.timezone=UTC \
 --conf spark.sql.session.timeZone=UTC"
 
-export SEQ_CONF="--executor-memory 12G \
+export SEQ_CONF="--executor-memory 16G \
 --total-executor-cores 6"
 
 # currently we hardcode the parallelism and configs based on our CI node's hardware specs,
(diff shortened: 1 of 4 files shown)


jenkins/spark-tests.sh
```diff
@@ -98,14 +77,105 @@ jps
...
+        return $CODE
+        ;;
+    esac
+}
+export -f run_test
+
+# integration tests
+if [[ $PARALLEL_TEST == "true" ]] && [ -x "$(command -v parallel)" ]; then
+  # put most time-consuming tests at the head of queue
+  time_consuming_tests="join_test.py generate_expr_test.py"
```
jlowe, line 162: I thought we were tagging the source to indicate they were slow as in #3241.  Curious why this isn't leveraging that, e.g.: `time_consuming_tests=$(grep -rl pytest.mark.slow_test "$SCRIPT_PATH"/src/main/python

pxLi: `slow_tag` there was for pre-merge parallel split (either time-consuming or high memory usage), mostly to balance test time of two parallel test stages (the tag naming looks ambiguous)
In nightly cases, I want pick time-consuming ones only, and per my tests join_test (3000s) + generate_expr_test (1800s) were the only 2 tests that spent consistently over 15 mins, so I manually put them here to avoid that they were in the middle or tail of task queue

pxLi:
build

jenkins/spark-tests.sh
jlowe: Should we split the test cases in `join_test` and `generate_expr_test` so they spread more evenly even without using the `parallel` hack?

jlowe (approved):
Looks OK to me other than we may want a followup to split some of these expensive test cases to make it easier for pytest to make better decisions about running them in parallel on its own.

jenkins/spark-tests.sh
pxLi: Yes, split in test-cases level that would be better. I have tested w/ a few non-serious cases-level split but not getting much benefits from it, and this could make it harder for developer to manage the test scenarios. I would like to get back to this if current setup does not fulfill our efficiency requirement

pxLi:
> Looks OK to me other than we may want a followup to split some of these expensive test cases to make it easier for pytest to make better decisions about running them in parallel on its own.

Thanks! I filed an issue  https://github.com/NVIDIA/spark-rapids/issues/3279 to track the followup

PR #3273, opened 2021-08-20
Very basic support for ArrayTransform

revans2:
This is a first step at trying to support ArrayTransform.  There is a lot more that I should be able to do with supporting the array index column, and other columns using explode and pos_explode behind the scenes. But before I got into the details of that I wanted to make sure that this looked like I was going in the right direction.  For now I have this as a draft but if things look good I will replace the TODOs with follow on issues and possibly merge it in.


revans2:
build

revans2:
This has some bugs in it, so I am going to close it for now and open a new one when I have them all worked out.

PR #3274, opened 2021-08-23
Add m2 cache to fast premerge build

zhanga5:
Sometimes premerge build took much longer time than normal because of unstable network connection. This PR enables leveraging pre-created m2 cache to minimize such impact to fast premerge build. The pre-created m2 cache were removed snapshot or plugin related jars/xml in order to eliminate potential side effects. Please refer to job 'update_premerge_m2_cache' on Blossom about building m2 tarball details. (if necessary this pre-created m2 cache could be updated regularly)

### DIFF
--- a/jenkins/spark-premerge-build.sh
+++ b/jenkins/spark-premerge-build.sh
@@ -111,6 +111,13 @@ MVN_GET_CMD="mvn org.apache.maven.plugins:maven-dependency-plugin:2.8:get -B \
 
 rm -rf $ARTF_ROOT && mkdir -p $ARTF_ROOT
 
+# If possible create '~/.m2' cache from pre-created m2 tarball to minimize the impact of unstable network connection.
+# Please refer to job 'update_premerge_m2_cache' on Blossom about building m2 tarball details.
+M2_CACHE_TAR=${M2_CACHE_TAR:-"/home/jenkins/agent/m2_cache/premerge_m2_cache.tar"}
+if [ -s "$M2_CACHE_TAR" ] ; then
+    tar xvf $M2_CACHE_TAR -C ~/
+fi
+
 # Download a full version of spark
 $MVN_GET_CMD \
     -DgroupId=org.apache -DartifactId=spark -Dversion=$SPARK_VER -Dclassifier=bin-hadoop3.2 -Dpackaging=tgz


zhanga5:
build

tgravescs:
> m2 cache were removed snapshot or plugin related jars/xml in order to eliminate potential side effects.

Does it also remove cudf jars?

jenkins/spark-premerge-build.sh
```diff
@@ -111,6 +111,13 @@ MVN_GET_CMD="mvn org.apache.maven.plugins:maven-dependency-plugin:2.8:get -B \
 
 rm -rf $ARTF_ROOT && mkdir -p $ARTF_ROOT
 
+# If possible create '~/.m2' cache from pre-created m2 tarball to minimize the impact of unstable network connection.
+# Please refer to job 'update_premerge_m2_cache' on Blossom about building m2 tarball details.
+M2_CACHE_TAR=${M2_CACHE_TAR:-"/home/jenkins/agent/m2_cache/premerge_m2_cache.tar"}
```
jlowe, line 116: Where can we find the contents of this tar archive?  Is there a jenkins job responsible for creating the artifact?  It would be nice for the comments here to point people in the right direction to learn more about it in case it needs to be fixed or updated as dependencies change.

zhanga5:
> > m2 cache were removed snapshot or plugin related jars/xml in order to eliminate potential side effects.
> 
> Does it also remove cudf jars?

yes, it's removed including every file from `~/.m2/repository/ai/`

jenkins/spark-premerge-build.sh
zhanga5: yes, there's Jenkins job `update_premerge_m2_cache` on Blossom. Since it's internal, how about add some description to the document about the content?

jlowe: If we can't link to it directly then mentioning the job name is fine.  The danger of being too detailed here is that someone could update the Blossom job, invalidating the description here, and they forget to update it.  I think mentioning the Blossom job name is fine.

jlowe (approved):

jenkins/spark-premerge-build.sh
zhanga5: I see. Let met add some comments to the job on Blossom and corresponding Jenkinsfile in order to minimize such issue.

zhanga5: Blossom job and Jenkinsfile are updated. Please refer to internal Blossom system

Issue #2945, opened 2021-07-16
Improve the over-estimating size when doing ORC COALESCING reading


wbo4958:
With https://github.com/NVIDIA/spark-rapids/pull/2909, ORC has supported COALESCING reading.

Before coalescing small files, we need to estimate an initial size to allocate a HostMemoryBuffer to store the HEADER, STRIPES, and FOOTER. From the testing result on non-partitioned 5000 files total 1.3G, there is overestimating for the initial estimated size.

```
stderr:21/07/16 09:08:38 INFO MultiFileOrcPartitionReader: ORC Coalescing reading estimates the initTotalSize: 63929078, and the true size: 63160978
stderr:21/07/16 09:08:38 INFO MultiFileOrcPartitionReader: ORC Coalescing reading estimates the initTotalSize: 63535693, and the true size: 62768848
stderr:21/07/16 09:08:38 INFO MultiFileOrcPartitionReader: ORC Coalescing reading estimates the initTotalSize: 63159114, and the true size: 62391504
stderr:21/07/16 09:08:38 INFO MultiFileOrcPartitionReader: ORC Coalescing reading estimates the initTotalSize: 63055310, and the true size: 62287543
stderr:21/07/16 09:08:38 INFO MultiFileOrcPartitionReader: ORC Coalescing reading estimates the initTotalSize: 64388818, and the true size: 63620556
stderr:21/07/16 09:08:39 INFO MultiFileOrcPartitionReader: ORC Coalescing reading estimates the initTotalSize: 63271377, and the true size: 62502777
stderr:21/07/16 09:08:39 INFO MultiFileOrcPartitionReader: ORC Coalescing reading estimates the initTotalSize: 63699389, and the true size: 62931986
stderr:21/07/16 09:08:39 INFO MultiFileOrcPartitionReader: ORC Coalescing reading estimates the initTotalSize: 63398089, and the true size: 62630814
stderr:21/07/16 09:08:41 INFO MultiFileOrcPartitionReader: ORC Coalescing reading estimates the initTotalSize: 62961599, and the true size: 62193810
stderr:21/07/16 09:08:41 INFO MultiFileOrcPartitionReader: ORC Coalescing reading estimates the initTotalSize: 62870696, and the true size: 62103385
stderr:21/07/16 09:08:42 INFO MultiFileOrcPartitionReader: ORC Coalescing reading estimates the initTotalSize: 62784883, and the true size: 62017554
[...]

PR #3275, opened 2021-08-23
Improve over-estimating for ORC coalescing reading

wbo4958:
This PR is for issue #2945 .

Let's assume we're coalescing 2 orc files with 2 stripes for each file in a Spark Task.

orc-coalesce

The original initial estimated size like below,

``` console
size_of (“ORC”) + 
size_of stripe1 (DATA + StripeFooter) +
size_of stripe2 (DATA + StripeFooter) + 
size_of stripe3 (DATA + StripeFooter) + 
size_of stripe4 (DATA + StripeFooter) +
size_of FOOTER1 +
size_of FOOTER2  +
256 (Max size of PostScript) + 
1 (Length of PostScript) +
128 * 1024 (Add in a bit of fudging in case the whole file is being consumed and
                   our codec version isn't as efficient as the original writer's codec.)
```

We have added all ORC file's FOOTER size (size_of FOOTER1 + size_of FOOTER2) in order to estimate the StripeInformation size in FOOTER. It seems a little over-estimating since FOOTER size includes not only StripeInformation but also  Schema/ColumnStatistics ... 

This PR uses a fixed worst-case StripeInformation size instead of using ORC files' FOOTER size.

Below is the comparing test on 5000 orc files total 1.3G

```
over-estimating = initialized estimated size - the final true size
```

| spark task number |  upstream over-estimating size  | with this PR over-estimating size|
|---|---|---|
|  12 |  14.1M | 1.6M |
| 162 | 32.7M | 20.6M|

**Please be note**, we have added `128 * 1024` as the fudging. If we reduce this size, the over-estimating can only be `1.6M - 12 * (128*1024) =  102K` and `20.6M - 162 * (128*1024) = 358.4K`

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala
@@ -1602,6 +1602,19 @@ class MultiFileOrcPartitionReader(
   implicit def toOrcExtraInfo(in: ExtraInfo): OrcExtraInfo =
     in.asInstanceOf[OrcExtraInfo]
 
+  // Estimate the size of StripeInformation with the worst case.
+  // The serialized size may be different because of the different values.
+  // Here set most of values to "Long.MaxValue" to get the worst case.
+  lazy val sizeOfStripeInformation = {
+    OrcProto.StripeInformation.newBuilder()
+      .setOffset(Long.MaxValue)
+      .setIndexLength(0) // Index stream is pruned
+      .setDataLength(Long.MaxValue)
+      .setFooterLength(Int.MaxValue) // StripeFooter size should be small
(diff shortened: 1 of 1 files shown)


wbo4958:
build

pxLi:
build

wbo4958:
@tgravescs Hi Tom, Could you help to review this PR?

tgravescs:
what is the logic behind the 128 * 1024  size?

tgravescs (approved):
overall looks fine, just curious if we can improve or document on the 128*1024.

wbo4958:
> overall looks fine, just curious if we can improve or document on the 128*1024.

@tgravescs I don't know, it's from the very beginning. From the comments, looks like the different codec between the original file and the current one we use may consume different buffer.

PR #3276, opened 2021-08-23
Use child type in some places to make it more clear

revans2:
We used "nested" types in a number of places to refer to a "child" type. This was confusing and this tries to fix it. If needed I can add in a definition of a nested type and a child type to the docs too.


revans2:
build

tgravescs:
child makes it more clear to me but should get others feedback and keep things consistent in our terminology.  @sameerz

jlowe (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -144,18 +144,18 @@ final class TypeSig private(
    */
   def withAllLit(): TypeSig = {
     // don't need to combine initialTypes with litOnlyTypes because litOnly should be a subset
-    new TypeSig(initialTypes, maxAllowedDecimalPrecision, nestedTypes, initialTypes, notes)
+    new TypeSig(initialTypes, maxAllowedDecimalPrecision, childTypes, initialTypes, notes)
   }
 
   /**
-   * Combine two type signatures together.  Base types and nested types will be the union of
+   * Combine two type signatures together.  Base types and child types will be the union of
```
ttnghia, line 151:
```suggestion
   * Combine two type signatures together. Base types and child types will be the union of
```

ttnghia:
> child makes it more clear to me but should get others feedback and keep things consistent in our terminology. @sameerz

Does `childType` here really mean a subtype or a type of some children (column)?

revans2:
> Does `childType` here really mean a subtype or a type of some children (columns) or something similar?

Type of child columns. Not child type as in the `IntegerType` class is a child of the `NumericType` class, which is also a child of the `AtomicType` class in Spark.

revans2:
build

jlowe (dismissed):

revans2:
build

jlowe (dismissed):

revans2:
build

revans2:
build

jlowe (approved):

sameerz:
build

pxLi:
build

Issue #3223, opened 2021-08-12
[FEA] Rework the shim layer to robustly handle ABI and API incompatibilities across Spark releases


gerashegalov:
**Is your feature request related to a problem? Please describe.**

Apache Spark uses [semantic versioning major.feature.patch](https://spark.apache.org/versioning-policy.html), e.g. 3.1.2. 
- Most compatibility promises are moot for the Plugin because our integration points are in the developer API and internal API land, not the Public APIs this language pertains to. 
- Even if the compatibility guarantees were applicable we need to deal with breaking changes because we support multiple feature release lines 3.0.1+, 3.1.1+, and the future 3.2.x. In practice, incompatibilities at the API and ABI level may occur much more often in internal APIs.
- Per documentation and simple bytecode comparison, even if there is compatibility, there is often no ABI compatibility, just API (source) compatibility requiring recompilation. Thus our current implementation relying on the version-independent compilation using 3.0.1 as a dependency runs into runtime failures with the other Spark versions that is only detected with tests.

**Describe the solution you'd like**
A solution that catches 99% of the issues at build time and is robust at run time.

**Describe alternatives you've considered**
Rely on the existing solution that is based on test coverage to identify the issues at run time

**Additional context**
Prototype branches such as 
- https://github.com/gerashegalov/spark-rapids/tree/reshim-source-deps-classifier-and-parallel-classloader-spark320-2
- https://github.com/gerashegalov/spark-rapids/tree/reshim-source-deps-classifier-and-parallel-classloader-spark320-refactor-on-the-fly

Tasks:
- [x] Build one spark version at a time  - https://github.com/NVIDIA/spark-rapids/issues/3224
- [x]  Add shimming against incompatibility "inherited" via changes to catalyst TreeNode derivatives  - https://github.com/NVIDIA/spark-rapids/pull/3323
- [x] Classloader changes - https://github.com/NVIDIA/spark-rapids/pull/3381
- [x] Build changes for combining classifier versions of Spark into single dist package - https://github.com/NVIDIA/spark-rapids/issues/3223#issue-969324171
- [x] https://github.com/NVIDIA/spark-rapids/issues/3338
- [x] #3495
- [x] ParquetCachedBatchSerializer move to common place for all spark 3.1+ builds - https://github.com/NVIDIA/spark-rapids/issues/3314
- [x] Fix any deploy targets and update jenkins builds
- [x] deduplicate shim layer code 
[...]

PR #3277, opened 2021-08-23
Build changes for Spark 3.0.3, 3.0.4, 3.1.1, 3.1.2, 3.1.3, 3.1.1cdh and 3.0.1emr

tgravescs:
part of https://github.com/NVIDIA/spark-rapids/issues/3224

This is a lot of copied code. All the shim code for each of these was copied into the individual shims.  

Much of the core code from like SparkBaseShims was copied into the 3.0.x versions and for 3.1.1 I split that code out into its own SparkBaseShims with the changes from Spark311Shims and that got copied to the 3.1.x code, except for the cases it was overrident like ParquetFilters. The files should diff pretty well.

The other changes were to add build profiles.  For the shims/aggregator and shims, I created another profile called releaseOne that handles all the dependencies so we didn't need individual ones.
Unfortunately for cdh to build I had to make specific changes to various pom files to deal with dependencies there.

I also made the tests-spark310+ only run for Spark 3.1.1 right now because it directly references the shim version for 3.1.1  ParquetCachedBatchSerializer.  We want to move ParquetCachedBatchSerializer out of shims but can't do it until we switch the build system. For now it seems like just running it against 3.1.1 should be fine and fix when we move it around.

I've run all the unit tests for all these versions but have not run all the integration tests on all the final versions after upmerging, I'll kick some of those off.


tgravescs:
build

tgravescs:
ran mvn verify and had doc updates,

tgravescs:
build

pom.xml
```diff
@@ -401,9 +499,13 @@
                   <artifactId>hive-storage-api</artifactId>
                 </exclusion>
                 <exclusion>
-                  <groupId> com.esotericsoftware</groupId>
+                  <groupId>com.esotericsoftware</groupId>
                   <artifactId>kryo-shaded</artifactId>
                 </exclusion>
+                <exclusion>
+                  <groupId>com.google.guava</groupId>
```
tgravescs, line 506: I need to make this conditional for just the new builds

tgravescs:
build

sql-plugin/pom.xml
```diff
@@ -91,11 +84,70 @@
...
+                </dependency>
+                <dependency>
+                    <groupId>org.apache.spark</groupId>
+                    <artifactId>spark-hive_${scala.binary.version}</artifactId>
+                    <version>${spark311cdh.version}</version>
+                </dependency>
+                <dependency>
+                    <groupId>org.apache.curator</groupId>
+                    <artifactId>curator-recipes</artifactId>
+                    <version>4.3.0.7.2.7.0-184</version>
```
jlowe, line 132: By pulling in a different dependency version here, is there any concern with conflicts? Does it need to be shaded as we have done in other cases where we pull in a different dependency version? This comment applies to other places where the curator-recipes dependency was overridden as well.

jlowe (commented):
It would be good to verify the contents of the distribution jar before/after this PR with the default build.

pom.xml
```diff
@@ -216,36 +314,24 @@
             <properties>
                 <spark.test.version>${spark311cdh.version}</spark.test.version>
             </properties>
-            <modules>
-                <module>tests-spark310+</module>
-            </modules>
         </profile>
         <profile>
             <id>spark312tests</id>
```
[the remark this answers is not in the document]
tgravescs: no, but I didn't want to remove anything until we decided to switch over.

shims/spark312/src/main/scala/com/nvidia/spark/rapids/shims/spark312/ParquetCachedBatchSerializer.scala
```diff
@@ -0,0 +1,1521 @@
+/*
```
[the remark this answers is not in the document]
tgravescs: so I agree, the plan is to move these out into someplace for spark 3.1+ but wasn't going to do that here, was going to do that separately.  I should probably file issues for the separate tasks.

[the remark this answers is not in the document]
tgravescs: Note this also depends on if separate module or just separate source directory inside of sql-plugin and when we want to do that vs potentially breaking the old build.

sql-plugin/pom.xml
```diff
@@ -44,18 +44,11 @@
...
-        </dependency>
-        <dependency>
-            <groupId>org.apache.spark</groupId>
-            <artifactId>spark-sql_${scala.binary.version}</artifactId>
-        </dependency>
         <dependency>
           <groupId>org.apache.orc</groupId>
           <artifactId>orc-core</artifactId>
           <classifier>${orc.classifier}</classifier>
+          <version>1.5.8</version>
```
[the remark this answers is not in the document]
tgravescs: shouldn't even be needed, I think its left over from me debugging/figuring something out,

shims/spark301emr/src/main/scala/com/nvidia/spark/rapids/shims/spark301emr/GpuBroadcastNestedLoopJoinExec.scala
```diff
@@ -0,0 +1,43 @@
...
+import com.nvidia.spark.rapids.GpuBuildSide
+
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.catalyst.plans.JoinType
+import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.execution.joins.BroadcastNestedLoopJoinExec
+import org.apache.spark.sql.rapids.execution._
+
+/**
+ *  Spark 3.1 changed packages of BuildLeft, BuildRight, BuildSide
```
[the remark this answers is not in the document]
tgravescs: I'll clean them up

shims/spark301emr/src/main/scala/com/nvidia/spark/rapids/shims/spark301emr/GpuParquetScan.scala
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.connector.read.PartitionReaderFactory
+import org.apache.spark.sql.execution.datasources.PartitioningAwareFileIndex
+import org.apache.spark.sql.execution.datasources.v2.FileScan
+import org.apache.spark.sql.sources.Filter
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.util.CaseInsensitiveStringMap
+
+// FileScan changed in Spark 3.1.0 so need to compile in Shim
```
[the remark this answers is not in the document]
tgravescs: I'll clean them up

shims/spark301emr/src/main/scala/com/nvidia/spark/rapids/shims/spark301emr/GpuOrcScan.scala
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.connector.read.PartitionReaderFactory
+import org.apache.spark.sql.execution.datasources.PartitioningAwareFileIndex
+import org.apache.spark.sql.execution.datasources.v2.FileScan
+import org.apache.spark.sql.sources.Filter
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.util.CaseInsensitiveStringMap
+
+// FileScan changed in Spark 3.1.0 so need to compile in Shim
```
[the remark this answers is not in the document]
tgravescs: I'll clean them up

sql-plugin/pom.xml
tgravescs: good question, I started just copying what they did in the shim layer so I'll check on this.

tgravescs: I went ahead and changed all these to be provided as cdh environment should have these. 
 I also added a task for this in https://github.com/NVIDIA/spark-rapids/issues/3223 for us to check when we start combining these to make sure it doesn't leak/conflict.

tgravescs: I also checked the final dist jar for cdh to make sure nothing else was in there and its good now

tgravescs:
I think I addressed all the comments. I checked the final dist jar against one we push nightly and the only diffs I found were the extra shim files that we have now.

tgravescs:
build

shims/spark311cdh/pom.xml
```diff
@@ -96,6 +91,7 @@
             <groupId>org.apache.curator</groupId>
             <artifactId>curator-recipes</artifactId>
             <version>4.3.0.7.2.7.0-184</version>
+	        <scope>provided</scope>
```
jlowe, line 94: Nit: Indentation is off here.

udf-examples/pom.xml
```diff
@@ -115,6 +115,40 @@
...
+            <dependency>
+                <groupId>org.apache.spark</groupId>
+                <artifactId>spark-sql_${scala.binary.version}</artifactId>
+                <version>${spark311cdh.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>org.apache.curator</groupId>
+                        <artifactId>curator-recipes</artifactId>
+                    </exclusion>
+                </exclusions>
```
jlowe, line 136: If we're pulling in the CDH version of the Spark jars, why would we need to override a dependency and replace it with something else?  Are the dependencies wrong on their jar?

udf-compiler/pom.xml
```diff
@@ -141,6 +134,13 @@
                     <groupId>org.apache.curator</groupId>
                     <artifactId>curator-recipes</artifactId>
                     <version>4.3.0.7.2.7.0-184</version>
+                    <scope>provided</scope>
+                </dependency>
+                <dependency>
+                    <groupId>com.google.guava</groupId>
+                    <artifactId>guava</artifactId>
+                    <version>28.0-jre</version>
+                    <scope>provided</scope>
```
jlowe: Guava is notorious for being incompatible between the versions.  If the UDF compiled needs Guava we'll probably need to shade it to guarantee it gets the version it needs rather than hope the one it finds in the Spark env is compatible.

tgravescs: changed to test scope and commented, asking Cloudera about it

udf-examples/pom.xml
tgravescs: I assume so, this is what they had in the shim layer that they wrote.. I'll follow up with them to see if we can remove later

tgravescs:
build

udf-compiler/pom.xml
tgravescs: note it was just unit tests that needed it (thus the change to test scope - which is what I should have had it at in th efirst place)

jlowe (approved):

PR #3283, opened 2021-08-24
Fix ClassCastException for unsupported TypedImperativeAggregate functions

### LINKED ISSUE #3253: [BUG] ClassCastException for unsupported TypedImperativeAggregate functions
**Describe the bug**

If we attempt to use any `TypedImperativeAggregate` functions that we do not support yet (such as approx percentile) then the plugin throws a ClassCastException.

```
Welcome to
      ____              __
     / __/__  ___ _____/ /__
    _\ \/ _ \/ _ `/ __/  '_/
   /___/ .__/\_,_/_/ /_/\_\   version 3.1.2
      /_/
         
Using Scala version 2.12.10 (OpenJDK 64-Bit Server VM, Java 1.8.0_292)
Type in expressions to have them evaluated.
Type :help for more information.

scala> val df1 = (600 to 1000).toDF("n")
df1: org.apache.spark.sql.DataFrame = [n: int]

scala> val df2 = df1.agg(expr("approx_percentile(n, array(0.25, 0.5, 0.75))").as("approx_percentiles"))
df2: org.apache.spark.sql.DataFrame = [approx_percentiles: array<int>]

scala> df2.show
java.lang.ClassCastException: com.nvidia.spark.rapids.RuleNotFoundExprMeta cannot be cast to com.nvidia.spark.rapids.TypedImperativeAggExprMeta
  at com.nvidia.spark.rapids.GpuTypedImperativeSupportedAggregateExecMeta.$anonfun$overrideAggBufTypes$1(aggregate.scala:1047)
  at com.nvidia.spark.rapids.GpuTypedImperativeSupportedAggregateExecMeta.$anonfun$overrideAggBufTypes$1$adapted(aggregate.scala:1043)
```

**Expected behavior**
We should fall back to CPU

- Standalone / local mode

**Additional context**
None

sperlingxx:

Fixes #3253 

The ClassCastException occurred during overriding the buffer types of TypedImperativeAggregate functions. The code for overriding assumed that  `TypedImperativeAggreagte` must be wrapped by `TypedImperativeAggExprMeta`. But there are exceptions: for  `TypedImperativeAggreagte` without GPU implementation , they are wrapped by `RuleNotFoundExprMeta`.

### DIFF
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -547,6 +547,20 @@ def test_hash_groupby_collect_partial_replace_fallback_with_other_agg(conf, aqe_
     group by k1""",
         conf=local_conf)
 
+@ignore_order(local=True)
+@allow_non_gpu('ObjectHashAggregateExec', 'ShuffleExchangeExec',
+               'HashAggregateExec', 'HashPartitioning',
+               'ApproximatePercentile', 'Alias', 'Literal', 'AggregateExpression')
+def test_hash_groupby_typed_imperative_agg_without_gpu_implementation_fallback():
+    assert_cpu_and_gpu_are_equal_sql_with_capture(
+        lambda spark: gen_df(spark, [('k', RepeatSeqGen(LongGen(), length=20)),
+                                     ('v', LongRangeGen())], length=100),
+        exist_classes='ApproximatePercentile,ObjectHashAggregateExec',
+        non_exist_classes='GpuApproximatePercentile,GpuObjectHashAggregateExec',
+        table_name='table',
+        sql="""select k,
+        approx_percentile(v, array(0.25, 0.5, 0.75)) from table group by k""")
+
 @approximate_float
 @ignore_order
 @incompat
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
@@ -1071,19 +1071,13 @@ abstract class GpuTypedImperativeSupportedAggregateExecMeta[INPUT <: SparkPlan](
     val desiredInputAggBufTypes = mutable.HashMap.empty[ExprId, DataType]
     // Collects exprId from TypedImperativeAggBufferAttributes, and maps them to the data type
     // of `TypedImperativeAggExprMeta.aggBufferAttribute`.
-    agg.aggregateExpressions.zipWithIndex.foreach {
-      case (expr, i) if expr.aggregateFunction.isInstanceOf[TypedImperativeAggregate[_]] =>
+    aggregateExpressions.map(_.childExprs.head).foreach {
+      case aggMeta: TypedImperativeAggExprMeta[_] =>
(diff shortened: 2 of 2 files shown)


sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -1071,19 +1071,13 @@ abstract class GpuTypedImperativeSupportedAggregateExecMeta[INPUT <: SparkPlan](
     val desiredInputAggBufTypes = mutable.HashMap.empty[ExprId, DataType]
     // Collects exprId from TypedImperativeAggBufferAttributes, and maps them to the data type
     // of `TypedImperativeAggExprMeta.aggBufferAttribute`.
-    agg.aggregateExpressions.zipWithIndex.foreach {
-      case (expr, i) if expr.aggregateFunction.isInstanceOf[TypedImperativeAggregate[_]] =>
+    aggregateExpressions.map(_.childExprs.head).foreach {
+      case aggMeta: TypedImperativeAggExprMeta[_] =>
```
abellina, line 1075: Nit, this removes the need for the `asInstanceOf` for `wrapped`:

```
case aggMeta: TypedImperativeAggExprMeta[TypedImperativeAggregate[_]] =>
```

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -547,6 +547,20 @@ def test_hash_groupby_collect_partial_replace_fallback_with_other_agg(conf, aqe_
     group by k1""",
         conf=local_conf)
 
+@ignore_order(local=True)
```
abellina, line 550: I believe `approx_percentile` is coming, and when it does show up we will fail this test and then whoever implements the function needs to figure out how to address, it would be better if we force the aggregate function to not be on the GPU. Could you explicitly turn off an already working typed imperative aggregate?

abellina: (i.e. `spark.rapids.sql.expression.CollectList=false`, or you should be able to get away with `spark.rapids.sql.expression.ApproximatePercentile=false`)

ttnghia: Newbie question: Why do we allow_non_gpu for these aggs?

abellina: The plugin can be placed in a mode that throws if some parts of the plan can't be replaced (`spark.rapids.sql.test.enabled`). The allow list (`spark.rapids.sql.test.allowedNonGpu`) specifies expressions that are OK even in this mode, as they are pertinent to the test.

revans2: The `spark.rapids.sql.expression.ApproximatePercentile=false` is not going to work for actually testing this. We have to have a `TypedImperativeAggregate` that does not exist in the GpuOverrides mapping. Because even with that config it would still create a `TypedImperativeAggExprMeta` to wrap it. It would just be turned off.

I would propose we go with an API that is less likely we are going to support any time soon. Specifically a udaf, but that is really hard to do inside the python tests.  As it is I am fine with this for now, but looking at the list of TypedImperativeAggregate functions there are not a lot to choose from.

revans2 (approved):

PR #3288, opened 2021-08-24
Use cudf to compute exact hash join output row sizes

jlowe:
Removes the hash join code to estimate the join size and replaces it with the cudf output join size APIs.  This also removes the OOM catch-and-retry logic since it theoretically should no longer be necessary since we are no longer producing an estimate but instead an exact amount.

Posting as a draft for the following reasons:

- [ ] Depends on fix for https://github.com/rapidsai/cudf/issues/9095
- [ ] Depends on fix for https://github.com/rapidsai/cudf/issues/9092
- [ ] Need to validate performance gain by removing estimation code

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
@@ -49,6 +49,12 @@ abstract class AbstractGpuJoinIterator(
   protected def hasNextStreamBatch: Boolean
 
   /**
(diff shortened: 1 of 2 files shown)


revans2 (approved):
From the look of things it appears that the builtHash does not out live a single batch being returned. The rest of the changes look good.

jlowe:
> From the look of things it appears that the builtHash does not out live a single batch being returned.

Yeah, the built hash isn't spillable, and the code tries to make everything spillable before it produces a batch.  Therefore I thought it prudent to throw away the built hash to avoid a potential OOM.

PR #3289, opened 2021-08-24
Support for ArrayTransform

revans2:
This adds in full support for array transform.  The same code can probably be mostly reused to be able to support TransformValue, which transforms map values.  It also begins to setup a basic framework to be able to support lambda functions in the plugin.  There is probably more work and generalization that needs to happen, but it provides a starting point.


revans2:
build

revans2:
build

PR #3299, opened 2021-08-25
Insert buffer converters for TypedImperativeAggregate

sperlingxx:
Current PR addresses the last task of #2916, supporting the aggregation buffer conversion between CPU and GPU format for TypedImperativeAggregate functions. With the ability to insert buffer converters, we can handle TypedImperativeAggregate functions running across CPU and CPU in runtime. It indicates that we don't need to fallback the entire Aggregate stack to CPU once one stage need to fallback when the Aggregate contains TypedImperativeAggregate functions.

The general idea is to create buffer converters and bind them to certain physical plans during `preColumnarTransitions`. And integrate these buffer converters into RowToColumnar/ColumnarToRow transitions as pre-processing/post-processing projections during `postColumnarTransitions`. And this idea works even when AQE is on. To adapt AQE, we leverage TreeNodeTag to cache temporary information, including: buffer converters, and some meta data. 

For better understanding, let's walk through the entire procedure of inserting buffer converters:
1.  binding buffer converters to certain plans (preColumnarTransitions)
  1.1 collecting all stages of Aggregation which contains TypedImperativeAggregate functions   
       The binding procedure is triggered in `GpuTypedImperativeSupportedAggregateExecMeta.tagPlanForGpu` if wrapped plan 
       is the final stage. At first, we collect all stages as what we do for associated fallback.
  1.2 filtering stages who need buffer converters to fill the data gap with their child
  1.3 creating buffer converters with filtered stages
We add two new interfaces on GpuTypedImperativeSupportedAggregateExecMeta: `createCpuToGpuBufferConverter` and `createGpuToCpuBufferConverter`.
  1.4 binding buffer converters into certain plans
The plans who carry the buffer converters are the CPU plans (can not be replaced) located right before/after the potential R2C/C2R transitions.
2. materializing RowToColumnar/ColumnarToRow transitions with buffer converters (postColumnarTransitions)
We add extra field preTransitions/postTransitions to GpuRowToColumnar/GpuColumnarToRow, in order to insert projections like buffer converters for TypedImperativeAggregate.


sperlingxx:
build

shims/spark313/src/main/scala/com/nvidia/spark/rapids/shims/spark313/SparkBaseShims.scala
```diff
@@ -24,16 +24,16 @@ import org.apache.arrow.memory.ReferenceManager
 import org.apache.arrow.vector.ValueVector
 import org.apache.hadoop.fs.Path
 import org.apache.parquet.schema.MessageType
-
 import org.apache.spark.SparkEnv
+
```
revans2, line 28: Same formatting issue

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
```diff
@@ -293,13 +298,23 @@ abstract class GpuColumnarToRowExecParent(child: SparkPlan, val exportColumnarRd
...
+    val rdata = if (exportColumnarRdd) {
       // If we are exporting columnar rdd we need an easy way for the code that walks the
       // RDDs to know where the columnar to row transition is happening.
       GpuColumnToRowMapPartitionsRDD.mapPartitions(cdata, f)
     } else {
       cdata.mapPartitions(f)
     }
+
+    postTransition match {
+      case Some(transformations) =>
```
revans2, line 310: So what happens if transformations is empty?  Would it be cleaner to just have transformations not be optional but just be a Sequence, and go off of an empty sequence instead of None?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3188,21 +3201,39 @@ object GpuOverrides {
...
-          if (!exec.child.supportsColumnar) {
-            willNotWorkOnGpu(
-              "Unable to replace CustomShuffleReader due to child not being columnar")
+        new SparkPlanMeta[CustomShuffleReaderExec](exec, conf, p, r) {
+          override def tagPlanForGpu(): Unit = {
+            if (!exec.child.supportsColumnar) {
+              willNotWorkOnGpu(
+                "Unable to replace CustomShuffleReader due to child not being columnar")
+            }
+            val shuffleEx = exec.child.asInstanceOf[ShuffleQueryStageExec].plan
```
revans2, line 3212: How do we know that the child is a `ShuffleQueryStageExec`? That should only ever happen when AQE is enabled.  And even then I don't think it is a 100% guarantee.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -891,3 +896,56 @@ case class GpuCollectSet(
...
+case class CpuToGpuCollectBufferTransition(
+    override val child: Expression,
+    private val elementType: DataType) extends CpuToGpuBufferTransition {
+
+  private lazy val row = new UnsafeRow(1)
+
+  override def dataType: DataType = ArrayType(elementType, containsNull = false)
+
+  override protected def nullSafeEval(input: Any): ArrayData = {
+    val bytes = input.asInstanceOf[Array[Byte]]
```
revans2, line 930: Some comments here might be good to explain what is happening and point to the code that Spark uses to serialize out the data.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -891,3 +896,56 @@ case class GpuCollectSet(
...
+    GpuToCpuCollectBufferTransition(child)
+  }
+}
+
+case class GpuToCpuCollectBufferTransition(
+    override val child: Expression) extends GpuToCpuBufferTransition {
+
+  private lazy val projection = UnsafeProjection.create(Array(child.dataType))
+
+  override protected def nullSafeEval(input: Any): Array[Byte] = {
```
revans2, line 947: Again some comments might be nice to point to the Spark code that is doing the serialization/deserialization so we can show how we match.

revans2 (changes requested):
I am a little concerned that this code appears to assume AQE is on all the time. It also looks like we cannotl have a TypeImerativeAggregation that does not use transitions. For percentile approximate it is going to use a very different algorithm and only supporting transitions is going to be very hard. I would prefer to keep the old code if we have to pick just one way to do it.

tgravescs:
it looks like this is missing the databricks shim updates for GpuColumnarToRowTransitionExec. If you are touching all the shims please think about the databricks ones as well.

tgravescs:
I'll put up pr to fix

tgravescs:
Sorry it looks like this may have just been in my branch with various build changes, apologize

PR #3301, opened 2021-08-25
Approx Percentile

andygrove:

Depends on https://github.com/rapidsai/cudf/pull/9094

Status:

- [x] Get one test working end to end (even if results are not 100% correct)
- [x] Add casts so final percentiles are in same data type as input
- [x] GpuOverride type checking
- [x] Add config to enable feature
- [x] GpuOverride fall back to CPU if used as a reduction
- [x] PySpark integration tests
- [x] Decide on translation (if any) from Spark delta to cuDF delta
- [x] Documentation


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -253,6 +253,31 @@ class CudfMin(ref: Expression) extends CudfAggregate(ref) {
...
+  private lazy val accuracy = accuracyExpression.value.toString.toInt
+
+  override lazy val updateReductionAggregate: ColumnVector => Scalar =
+    throw new UnsupportedOperationException()
+  override lazy val mergeReductionAggregate: ColumnVector => Scalar =
+    throw new UnsupportedOperationException()
+  override val updateAggregate: GroupByAggregation = GroupByAggregation.createTDigest(accuracy)
+  override val mergeAggregate: GroupByAggregation = GroupByAggregation.mergeTDigest(accuracy)
+  override def toString(): String = "CudfTDigest"
+  override def dataType: DataType = ArrayType(new StructType(Array(
```
ttnghia, line 271: What is the different of using `ArrayType(new StructType(` vs `ArrayType(StructType(`?

andygrove: The `new` is actually redundant here so can be removed. Scala requires `new` for regular classes but does not require it for case classes and `StructType` is a case class.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,142 @@
...
+
+  // Mark as lazy to avoid being initialized when creating a GpuApproximatePercentile.
+  override lazy val initialValues: Seq[GpuExpression] = throw new UnsupportedOperationException
+
+  override lazy val updateExpressions: Seq[Expression] =
+    new CudfTDigest(inputBuf,
+      percentageExpression,
+      accuracyExpression) :: Nil
+
+  override lazy val mergeExpressions: Seq[GpuExpression] =
```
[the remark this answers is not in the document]
ttnghia: Sorry I'm a bit confused. Here are the original methods from `GpuAggregateFunctions`:
```
// these are values that spark calls initial because it uses
  // them to initialize the aggregation buffer, and returns them in case
  // of an empty aggregate when there are no expressions,
  // here we copy them but with the gpu equivalent
  val initialValues: Seq[GpuExpression]

  // update: first half of the aggregation (count = count)
  val updateExpressions: Seq[Expression]

  // merge: second half of the aggregation (count = sum). Also use to merge multiple batches.
  val mergeExpressions: Seq[GpuExpression]

  // mostly likely a pass through (count => sum we merged above).
  // average has a more interesting expression to compute the division of sum/count
  val evaluateExpression: Expression
```

So we are using mixed `GpuExpression` and `Expression`. Why?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,147 @@
...
+      containsNull = false) else child.dataType
+  }
+
+  override def dataType: DataType = internalDataType
+
+  override def children: Seq[Expression] = Seq(child, percentageExpression, accuracyExpression)
+}
+
+object GpuApproximatePercentile {
+  def apply(child: Expression, percentageExpression: GpuLiteral): GpuApproximatePercentile = {
```
revans2, line 113: If this is not called by the plugin, then lets remove it. Spark adds these types of things to make the SQL translation simpler, but we don't have to deal with that part.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,147 @@
...
+    val expr = child.asInstanceOf[GpuExpression]
+    withResource(expr.columnarEval(batch).asInstanceOf[GpuColumnVector]) { cv =>
+      withResource(cv.getBase.approxPercentile(percentiles)) { percentiles =>
+        if (finalDataType == DataTypes.DoubleType) {
+          GpuColumnVector.from(percentiles.incRefCount(), dataType)
+        } else {
+          // cast cuDF Array[Double] to Array[finalDataType]
+          withResource(percentiles.getChildColumnView(0)) { childView =>
+            withResource(recursiveDoColumnar(childView, DataTypes.DoubleType, finalDataType,
+                ansiMode = SQLConf.get.ansiEnabled, legacyCastToString = false)) { childCv =>
```
revans2, line 134: I think ansiMode is always going to be false here. The original spark code didn't check ansiMode at all, and in theory we could just cast it ourselves with relatively little effort for everything except DecimalType which we still are not going to do great at, but at least when we do "fix" this in cast it would also be fixed here.

```
    val result = child.dataType match {
      case DateType => doubleResult.map(_.toInt)
      case TimestampType => doubleResult.map(_.toLong)
      case ByteType => doubleResult.map(_.toByte)
      case ShortType => doubleResult.map(_.toShort)
      case IntegerType => doubleResult.map(_.toInt)
      case LongType => doubleResult.map(_.toLong)
      case FloatType => doubleResult.map(_.toFloat)
      case DoubleType => doubleResult
      case _: DecimalType => doubleResult.map(Decimal(_))
      case other: DataType =>
        throw new UnsupportedOperationException(s"Unexpected data type ${other.catalogString}")
    }
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2792,6 +2792,51 @@ object GpuOverrides {
           aggBuffer.copy(dataType = c.dataType)(aggBuffer.exprId, aggBuffer.qualifier)
         }
       }),
+    expr[ApproximatePercentile](
+      "Approximate percentile",
+      ExprChecks.fullAggAndProject(
```
revans2, line 2797: This should not work for project, and it is not supported for reduce yet, so we probably should add a new category to ExprChecks that only enables this for regular group by aggregations, not reductions, and not window operations, and definitely not project.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2792,6 +2792,51 @@ object GpuOverrides {
...
+        // note that output can be single number or array depending on whether percentiles param
+        // is a single number or an array
+        TypeSig.numeric + TypeSig.DECIMAL_64 +
+          TypeSig.ARRAY.nested(TypeSig.numeric + TypeSig.DECIMAL_64),
+        TypeSig.ARRAY.nested(TypeSig.all),
```
revans2, line 2802: This is not correct.

The `inputTypes` for ApproximatePercentile are

```
  override def inputTypes: Seq[AbstractDataType] = {
    // Support NumericType, DateType and TimestampType since their internal types are all numeric,
    // and can be easily cast to double for processing.
    Seq(TypeCollection(NumericType, DateType, TimestampType),
      TypeCollection(DoubleType, ArrayType(DoubleType, containsNull = false)), IntegralType)
  }
```

The output might or might not be in an array, depending on the percentages passed in.

```
ExprChecks.groupByAggOnly(
    TypeSig.gpuNumeric + TypeSig.ARRAY.nested(TypeSig.gpuNumeric),
    TypeSig.numeric + TypeSig.DATE + TypeSig.TIMESTAMP + TypeSig.ARRAY.nested(TypeSig.numeric + TypeSig.DATE + TypeSig.TIMESTAMP),
    Seq(
      ParamCheck("input", TypeSig.gpuNumeric, TypeSig.numeric + TypeSig.DATE + TypeSig.TIMESTAMP),
      ParamCheck("percentage", TypeSig.DOUBLE + TypeSig.ARRAY.nested(TypeSig.DOUBLE), TypeSig.DOUBLE + TypeSig.ARRAY.nested(TypeSig.DOUBLE)),
      ParamCheck("accuracy", TypeSig.INT, TypeSig.INT))),
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2792,6 +2792,51 @@ object GpuOverrides {
...
+          ParamCheck("input", TypeSig.all, TypeSig.all),
+          ParamCheck("percentage",
+            TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 +
+              TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL_64),
+            TypeSig.all),
+          ParamCheck("accuracy", TypeSig.commonCudfTypes + TypeSig.DECIMAL_64, TypeSig.all)
+        )),
+      (c, conf, p, r) => new TypedImperativeAggExprMeta[ApproximatePercentile](c, conf, p, r) {
+        override def tagExprForGpu(): Unit = {
+          if (childExprs.length < 2 || childExprs.length > 3) {
```
revans2, line 2813: This is uneeded. None of the child expressions are optional or a sequence so the length of childExprs will always be 3.  You can assert on that just to be sure if you want to.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3143,7 +3188,7 @@ object GpuOverrides {
       "The backend for hash based aggregations supporting TypedImperativeAggregate functions",
       ExecChecks(
         (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64 +
-          TypeSig.MAP + TypeSig.ARRAY + TypeSig.STRUCT)
+          TypeSig.MAP + TypeSig.ARRAY.nested(TypeSig.STRUCT + TypeSig.DOUBLE) + TypeSig.STRUCT)
```
revans2, line 3191: This is confusing because anything nested in a TypeSig applies to all nest-able types in the TypeSig.  Could you just update the whole thing to the following?

```
(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64 +
  TypeSig.MAP + TypeSig.ARRAY + TypeSig.STRUCT).nested()
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2792,6 +2792,51 @@ object GpuOverrides {
...
+            TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 +
+              TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL_64),
+            TypeSig.all),
+          ParamCheck("accuracy", TypeSig.commonCudfTypes + TypeSig.DECIMAL_64, TypeSig.all)
+        )),
+      (c, conf, p, r) => new TypedImperativeAggExprMeta[ApproximatePercentile](c, conf, p, r) {
+        override def tagExprForGpu(): Unit = {
+          if (childExprs.length < 2 || childExprs.length > 3) {
+            willNotWorkOnGpu("wrong number of arguments")
+          } else if (!childExprs.tail.forall(_.wrapped.isInstanceOf[Literal])) {
```
revans2, line 2815: Probably want to check the value of the literal here too. It looks like we cannot support nulls for any of these. Not really sure what Spark would do if we put a null in for them so you might want to test that too.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -253,6 +253,39 @@ class CudfMin(ref: Expression) extends CudfAggregate(ref) {
...
+class CudfTDigest(
+    ref: Expression,
+    percentileExpr: GpuLiteral,
+    accuracyExpression: GpuLiteral)
+  extends CudfAggregate(ref) {
+
+  // map Spark delta to cuDF delta - this is experimental for now and
+  // will not guarantee that we have better results than Spark although
+  // it did help with one of the tests
+  private lazy val accuracy = accuracyExpression.value.toString.toInt match {
```
revans2, line 265: We know the type of value is always an int so just cast it.

`accuracyExpression.value.asInstanceOf[Int]`. Or better yet you already are doing a match. Just combine it.

``` 
private lazy val accuracy = accuracyExpression.value match {
    case delta: Int if delta >= 10000 => 10000
    ...
}
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
andygrove: I've updated the signatures but I don't see a `groupByAggOnly` method on ExprChecks

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
andygrove: I have updated this now so will resolve this conversation

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,147 @@
...
+    accuracyExpression: GpuLiteral)
+  extends GpuAggregateFunction with GpuAggregateWindowFunction {
+
+  override val inputProjection: Seq[Expression] = Seq(child)
+
+  // Attributes of fields in the aggregation buffer.
+  override def aggBufferAttributes: Seq[AttributeReference] = outputBuf :: Nil
+
+  // Mark as lazy to avoid being initialized when creating a GpuApproximatePercentile.
+  override lazy val initialValues: Seq[GpuExpression] = throw new UnsupportedOperationException
```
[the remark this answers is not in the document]
andygrove: Okay. I will work on writing a test case to trigger this exception and then implement something here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,147 @@
...
+  private def makeArray(v: Any)  = v match {
+    // Rule ImplicitTypeCasts can cast other numeric types to double
+    case null => (false, null)
+    case num: Double => (false, Array(num))
+    case arrayData: ArrayData => (true, arrayData.toDoubleArray())
+  }
+
+
+  // The result type is the same as the input type.
+  private lazy val internalDataType: DataType = {
```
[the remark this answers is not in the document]
andygrove: Sorry, I don't understand this one. I copied this code from Spark (3.1). Could you expand on this?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * @param child child expression that can produce column value with `child.eval()`
+ * @param percentageExpression Expression that represents a single percentage value or
+ *                             an array of percentage values. Each percentage value must be between
+ *                             0.0 and 1.0.
+ * @param accuracyExpression Integer literal expression of approximation accuracy. Higher value
+ *                           yields better accuracy, the default value is
+ *                           DEFAULT_PERCENTILE_ACCURACY.
+ */
+case class GpuApproximatePercentile (
```
ttnghia, line 54: I see that most other aggregate classes are put into `AggregateFunctions.scala`. Should we put this class into it to? In the future, we may organize the aggregates into a folder by a similar way as in Spark.

docs/compatibility.md
```diff
@@ -560,3 +560,11 @@ double quotes around strings in JSON data, whereas Spark allows single quotes ar
 data.  The RAPIDS Spark `get_json_object` operation on the GPU will return `None` in PySpark or
 `Null` in Scala when trying to match a string surrounded by single quotes.  This behavior will be
 updated in a future release to more closely match Spark.
+
+## Approximate Percentile
+
+The `approximate_percentile` operation is now supported on the GPU. The GPU implementation uses
```
revans2, line 566: nit: I think we should reword this a bit.

The `approximate_percentile` operation on the GPU uses ...

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -25,6 +25,10 @@
 import pyspark.sql.functions as f
 from spark_session import is_before_spark_311, with_cpu_session
 
+_approx_percentile_conf = { 'spark.rapids.approxPercentileEnabled': 'true',
+                            'spark.rapids.sql.test.allowedNonGpu': 'SortExec'
```
revans2, line 29: Why is sorting off of the GPU?

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -1088,3 +1092,116 @@ def do_it(spark):
...
+        0.05)
+
+# @ignore_order(local=True)
+# def test_hash_groupby_approx_percentile_decimal():
+#     compare_percentile_approx(
+#         lambda spark: gen_df(spark, [('k', StringGen(nullable=False)),
+#                                      ('v', DecimalGen(nullable=False))], length=100),
+#         [0.05, 0.25, 0.5, 0.75, 0.95])
+#
+# @ignore_order(local=True)
```
revans2, line 1138: why are these commented out?

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -1088,3 +1092,116 @@ def do_it(spark):
...
+# def test_hash_groupby_approx_percentile_decimal_nullable():
+#     compare_percentile_approx(
+#         lambda spark: gen_df(spark, [('k', StringGen(nullable=False)),
+#                                      ('v', DecimalGen(nullable=True))], length=100),
+#         [0.05, 0.25, 0.5, 0.75, 0.95])
+
+def compare_percentile_approx(df_fun, percentiles):
```
revans2, line 1145: Some comments here explaining what this is doing would be nice.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
revans2: This is large enough that I think it can stand on its own. I am fine if we want to put the aggregations into a separate aggregations package like Spark does. We have not done a great job with breaking up a single large rapids package.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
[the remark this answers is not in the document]
revans2: Any update on this?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,171 @@
...
+  protected final lazy val outputBuf: AttributeReference =
+    inputBuf.copy("outputBuf", CudfTDigest.dataType)(inputBuf.exprId, inputBuf.qualifier)
+
+  override def nullable: Boolean = false
+
+  // Mark as lazy so that percentageExpression is not evaluated during tree transformation.
+  private lazy val (returnPercentileArray, percentiles) =
+    percentageExpression match {
+      case GpuLiteral(v, _) => makeArray(v)
+      case _ => makeArray(percentageExpression.eval())
```
revans2, line 99: I think we can clean this code up. `percentageExpression` is guaranteed to be a GpuLiteral from the input type.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+  private def makeArray(v: Any): (Boolean, Either[Double, Array[Double]])  = v match {
+    // Rule ImplicitTypeCasts can cast other numeric types to double
+    case null => (false, Right(Array()))
+    case num: Double => (false, Left(num))
+    case arrayData: ArrayData => (true, Right(arrayData.toDoubleArray()))
```
revans2, line 106: I'm not sure this is guaranteed to be an array data, but your tests appear to indicate that it will be.

Just to avoid code warnings can we have a default case fro this with an exception?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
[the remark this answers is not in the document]
revans2: I am just saying that the name `internalDataType` is a bit confusing. But you are right it came from spark so it is fine.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,171 @@
...
+    val expr = child.asInstanceOf[GpuExpression]
+    withResource(expr.columnarEval(batch).asInstanceOf[GpuColumnVector]) { cv =>
+
+      percentiles match {
+        case Left(p) =>
+          // For the scalar case, we still pass cuDF an array of percentiles
+          // (containing a single item) and then extract the child column from the resulting
+          // array and return that (after converting from Double to finalDataType
+          withResource(cv.getBase.approxPercentile(Array(p))) { percentiles =>
+            withResource(percentiles.getChildColumnView(0)) { childView =>
```
revans2, line 135: What happens if there is a null at the list level? Is that guaranteed to never happen? If it did we could end up skipping that row in the child data column entirely.  I personally would rather see us call `extractListElement` which is guaranteed to work in all cases, even if this is technically safe to do.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,171 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.nvidia.spark.rapids
+
+import com.nvidia.spark.rapids.GpuCast.{recursiveDoColumnar}
```
revans2, line 19: I would prefer for us to keep the GpuCast in the call to recursiveDoColumnar otherwise it makes the code much harder to read and understand what is happening.

That or make the public facing API something like doCast.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3014,6 +3014,48 @@ object GpuOverrides extends Logging {
 
         override val supportBufferConversion: Boolean = true
       }),
+    expr[ApproximatePercentile](
+      "Approximate percentile",
+      ExprChecks.aggNotReduction(
```
revans2, line 3019: aggNotReduction says that this works as a window aggregation too, but that is not true. We might need to make a new API. I think that @ttnghia had to do the same thing for stddev so you might want to coordinate.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3014,6 +3014,48 @@ object GpuOverrides extends Logging {
...
+
+        override def tagExprForGpu(): Unit = {
+          if (!conf.isApproxPercentileEnabled) {
+            willNotWorkOnGpu(s"approx_percentile on GPU is disabled. " +
+              s"Set ${RapidsConf.ENABLE_APPROX_PERCENTILE}=true to enable it")
+          } else if (!childExprs.tail.forall(_.wrapped match {
+            case lit: Literal => lit.value != null
+            case _ => false
+          })) {
+            willNotWorkOnGpu("approx_percentile on GPU only supports non-null literal " +
```
revans2, line 3041: Can we have a follow on issue for the null percentages?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3014,6 +3014,48 @@ object GpuOverrides extends Logging {
...
+              "expressions for percentiles and accuracy")
```
revans2, line 3042: What about an array with nulls in it?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3367,7 +3409,8 @@ object GpuOverrides extends Logging {
       ExecChecks(
         (TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64 +
           TypeSig.MAP + TypeSig.ARRAY + TypeSig.STRUCT)
-            .nested()
```
revans2, line 3370: This change feels like we have regressed.  And empty nested means everything in the original type. The new code drops Maps from what is allowed as a nested value.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1280,6 +1280,14 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(true)
 
+  val ENABLE_APPROX_PERCENTILE = conf("spark.rapids.approxPercentileEnabled")
```
revans2, line 1283: The framework already generates a config for each operator based off of the name of the Spark operation.  You can have it disabled by default by adding something like

```
        .disabledByDefault("...")
```

in the meta for it.  I think in this case the config is `spark.rapids.sql.expression.ApproximatePercentile`  You could probably drop this config and all of the checks for it and replace it with the one that the framework automatically generates and checks.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -319,6 +318,45 @@ class CudfMin(ref: Expression) extends CudfAggregate(ref) {
   override def toString(): String = "CudfMin"
 }
 
+class CudfTDigest(
```
revans2, line 321: If this is not going to be used anywhere except in GpuApproximatePercentile it might be good to move it there.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -319,6 +318,45 @@ class CudfMin(ref: Expression) extends CudfAggregate(ref) {
...
+    ref: Expression,
+    percentileExpr: GpuLiteral,
+    accuracyExpression: GpuLiteral)
+  extends CudfAggregate(ref) {
+
+  // Map Spark delta to cuDF delta
+  private lazy val accuracy = accuracyExpression.value match {
+    case delta: Int => delta.max(1000)
```
revans2, line 329: nit: when we have this figured out we should document what we are doing.

integration_tests/src/main/python/hash_aggregate_test.py
andygrove: I have filed https://github.com/NVIDIA/spark-rapids/issues/3606 for supporting decimal types. I do not yet understand why these are failing. I have updated this PR to fall back to CPU in this case now and have added a fallback test.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
andygrove: We now fall back to CPUI for arrays containing nulls and a test has been added.

andygrove: Do we need a follow-on issue to support arrays with nulls on GPUs?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
andygrove: Done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.rapids.{CudfTDigest, GpuAggregateFunction}
+import org.apache.spark.sql.types.{ArrayType, DataType, DataTypes}
+import org.apache.spark.sql.vectorized.ColumnarBatch
+
+case class GpuApproximatePercentile (
+    child: Expression,
+    percentageExpression: GpuLiteral,
+    accuracyExpression: GpuLiteral)
+  extends GpuAggregateFunction with GpuAggregateWindowFunction {
```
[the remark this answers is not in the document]
andygrove: I removed this. Do we need a follow-on issue for supporting this as a window function?

integration_tests/src/main/python/hash_aggregate_test.py
andygrove: This is removed now.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3034,12 +3035,28 @@ object GpuOverrides extends Logging {
...
+            // check if the percentile expression can be supported on GPU
+            childExprs(1).wrapped match {
+              case lit: Literal => lit.value match {
+                case null =>
+                  willNotWorkOnGpu(
+                    "approx_percentile on GPU only supports non-null literal percentiles")
+                case a: GenericArrayData if a.array.isEmpty =>
+                  willNotWorkOnGpu(
+                    "approx_percentile on GPU does not support empty percentiles arrays")
+                case a: GenericArrayData if a.array.contains(null) =>
```
revans2, line 3056: We ran into issues not that long ago with UnsafeArrayData instead of GenericArrayData in a literal.  That is because the ArrayData was the result of running a foldable method instead of just generating the literal directly.   Would be good to work with just ArrayData and avoid using `.array` for it because UnsafeArrayData does not support `.array`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
revans2: It is just a really odd corner case. I am find with it as is right now.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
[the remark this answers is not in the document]
revans2: No one is asking for it for a window function right now. We can hold off on it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
revans2: still want to see this happen.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
ttnghia: Oh yeah, I almost forget about this. I'll add `groupByOnly` check.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,210 @@
...
+    "approx_percentile does not support reduction")
+
+  // the update expression will create a t-digest (List[Struct[Double, Double])
+  override lazy val updateExpressions: Seq[Expression] =
+    new CudfTDigest(inputBuf,
+      percentageExpression,
+      accuracyExpression) :: Nil
+
+  // the merge expression will merge t-digests
+  override lazy val mergeExpressions: Seq[GpuExpression] =
```
abellina, line 78:
```suggestion
  override lazy val mergeExpressions: Seq[Expression] =
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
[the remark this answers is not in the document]
andygrove: This is updated now to use `Seq[Expression]` in both cases

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -1074,6 +1076,108 @@ def do_it(spark):
         return df.groupBy('a').agg(f.min(df.b[1]["a"]))
     assert_gpu_and_cpu_are_equal_collect(do_it)
 
+@ignore_order(local=True)
+def test_hash_groupby_approx_percentile_long_repeated_keys():
+    compare_percentile_approx(
+        lambda spark: gen_df(spark, [('k', RepeatSeqGen(LongGen(), length=20)),
+                                     ('v', LongRangeGen())], length=100),
+        [0.05, 0.25, 0.5, 0.75, 0.95])
```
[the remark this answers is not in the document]
andygrove: Percentiles have to be between 0 and 1 and Spark checks this before the plan gets to us.

```
pyspark.sql.utils.AnalysisException: cannot resolve 'percentile(t.`v`, CAST(-0.05BD AS DOUBLE), 1L)' due to data type mismatch: Percentage(s) must be between 0.0 and 1.0, but got cast(-0.05 as double); line 1 pos 10;
```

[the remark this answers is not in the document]
andygrove: I think the null handling could have more tests though so I'll work on adding more there

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -1074,6 +1076,108 @@ def do_it(spark):
         return df.groupBy('a').agg(f.min(df.b[1]["a"]))
     assert_gpu_and_cpu_are_equal_collect(do_it)
 
+@ignore_order(local=True)
```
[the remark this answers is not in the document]
andygrove: We do have tests in scala `ApproximatePercentileSuite` to check that we fall back to CPU entirely for certain cases. See `empty percentile array fall back to CPU`, `array with nulls fall back to CPU`. and `fall back to CPU for reduction`

integration_tests/src/main/python/hash_aggregate_test.py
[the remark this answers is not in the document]
andygrove: @abellina Could you review the null tests in `ApproximatePercentileSuite` and see if those are adequate?

andygrove:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,220 @@
...
+ */
+case class ApproxPercentileFromTDigestExpr(
+    child: Expression,
+    percentiles: Either[Double, Array[Double]],
+    finalDataType: DataType)
+  extends GpuExpression {
+
+  override def columnarEval(batch: ColumnarBatch): Any = {
+    val expr = child.asInstanceOf[GpuExpression]
+    withResource(expr.columnarEval(batch).asInstanceOf[GpuColumnVector]) { cv =>
```
revans2, line 138: I don't think this is guaranteed to be a `GpuColumnVector`. Can we use `GpuExpressionsUtils.columnarEvalToColumn`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,220 @@
...
+      percentiles match {
+        case Left(p) =>
+          // For the scalar case, we still pass cuDF an array of percentiles
+          // (containing a single item) and then extract the child column from the resulting
+          // array and return that (after converting from Double to finalDataType)
+          withResource(cv.getBase.approxPercentile(Array(p))) { percentiles =>
+            withResource(percentiles.extractListElement(0)) { childView =>
+              withResource(doCast(childView, DataTypes.DoubleType, finalDataType,
+                  ansiMode = SQLConf.get.ansiEnabled, legacyCastToString = false,
```
revans2, line 147: I don't think we want to worry about ANSI mode here.  The Spark code is not thinking about it, and is just casting the value to a Double.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuApproximatePercentile.scala
```diff
@@ -0,0 +1,220 @@
...
+
+        case Right(p) =>
+          // array case - cast cuDF Array[Double] to Array[finalDataType]
+          withResource(cv.getBase.approxPercentile(p)) { percentiles =>
+            if (finalDataType == DataTypes.DoubleType) {
+              GpuColumnVector.from(percentiles.incRefCount(), dataType)
+            } else {
+              withResource(percentiles.getChildColumnView(0)) { childView =>
+                withResource(doCast(childView, DataTypes.DoubleType, finalDataType,
+                    ansiMode = SQLConf.get.ansiEnabled, legacyCastToString = false,
```
revans2, line 162: Same here I don't think we should worry about ANSI mode, and instead just set it to false.

revans2 (approved):

andygrove:
build

integration_tests/src/main/python/hash_aggregate_test.py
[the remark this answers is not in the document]
abellina: That looks good, thanks

abellina (approved):

andygrove:
build failed with connection error, will try again

andygrove:
build

PR #3303, opened 2021-08-25
Update to cudf conditional join change that removes null equality argument

jlowe:
Updates the plugin code to reflect the breaking cudf API change in https://github.com/rapidsai/cudf/pull/9096.  Now that operators can be null-aware and there's separate `EQUAL` vs. `NULL_EQUAL`, it's no longer necessary to specify at the top level whether nulls compare as equal.


revans2 (dismissed):
Did you want to look into adding support for AND and OR too? It looks like it is covered by the cudf fix, but I understand if we want to keep them separate.

jlowe:
build

jlowe:
build

revans2 (approved):

PR #3317, opened 2021-08-26
Adds in support for the transform_values SQL function

revans2:
The majority of the reason for adding this in is to help generalize the lambda code. I didn't want to try and support this for transform_keys, because duplicate key checking is not supported right now.


jlowe:
> I wonder that why do we ever need to update configs.md and supported_ops.md in every PR?

More than once we have found bugs in the review of PRs by examining the diff to the generated documentation.  The developer thought they were adding the appropriate checks for certain types, but the generated documentation proved otherwise.  In practice we have not seen very many PRs get hung up on the generated documentation as why they aren't being merged.

jlowe (approved):

PR #3330, opened 2021-08-27
Support `int96RebaseModeInWrite` and `int96RebaseModeInRead`

razajafri:
While writing/reading a parquet file a user can specify a different config for reading/writing int96 values vs datetime values in general. 

This PR takes the value set by users to `int96RebaseModeInWrite` and `int96RebaseModeInRead` and matches what Apache spark does. 


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -87,6 +87,22 @@ object GpuParquetFileFormat {
       TrampolineUtil.dataTypeExistsRecursively(field.dataType, _.isInstanceOf[DateType])
     }
 
+    ShimLoader.getSparkShims.getSparkShimVersion.toString() match {
```
jlowe, line 90: Main code should never check the specific shim version like this.  It's very fragile and likely to break if there were other versions that had strings like "3.0.1-something".  Put this logic directly in the shim instead rather than making the common code check the shim version before calling the shim.

This comment applies everywhere in the PR where non-shim code is switching on the shim version string.

jlowe (commented):
I was expecting a shim function that would check INT96 timestamp rebase node, and in Spark < 3.1.1 it would check the old datetime config and in Spark >= 3.1.1 it would check the new INT96-specific configs.

razajafri:
> I was expecting a shim function that would check INT96 timestamp rebase node, and in Spark < 3.1.1 it would check the old datetime config and in Spark >= 3.1.1 it would check the new INT96-specific configs.

Ironically, that was my first implementation. The reason why I didn't like that was because we will be returning the value of `dateTimeRebase` mode instead of `int96RebaseMode` when asked for it, its even worse when asking for `int96RebaseModeWrite.key`, but I can see that it's still better than running into a RTE when/if we ever support another version of `3.0.x`. I will make the change.

jlowe:
> The reason why I didn't like that was because we will be returning the value of dateTimeRebase mode instead of int96RebaseMode when asked for it

But that is exactly what Spark < 3.1.1 is doing today.  The shim function name would indicate we're trying to get the int96 rebase mode, and on Spark >= 3.1.1 it would check the Int96-specific config, and on Spark < 3.1.1 it would check the same config the Spark code is checking in that situation.

razajafri:
I have made some unnecessary changes to the SparkBaseShim in every version of 311. Will revert in a bit

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -298,13 +315,14 @@ class GpuParquetWriter(
...
     dateTimeRebaseException: Boolean,
+    int96RebaseException: Boolean,
     context: TaskAttemptContext)
   extends ColumnarOutputWriter(path, context, dataSchema, "Parquet") {
 
   val outputTimestampType = conf.get(SQLConf.PARQUET_OUTPUT_TIMESTAMP_TYPE.key)
 
   override def scanTableBeforeWrite(table: Table): Unit = {
-    if (dateTimeRebaseException) {
+    if (dateTimeRebaseException || int96RebaseException) {
```
jlowe, line 325: Why does this class take two booleans for rebase exceptions but treat them as if they were a single one?  The INT96 config is specific to the case when INT96 is being used, and I don't see this checking for that specific type. It looks like if a rebase is needed on a date column then this config could trigger an exception despite the other config (that should apply to date columns) indicates it should not throw.

jlowe (commented):
The Databricks shims will also need to be updated accordingly.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
razajafri: `int96RebaseException` is taking the `parquetOutputTimestampType` into account. But I agree that I have to separate the exception checking for both of these booleans because if INT96 is set to `CORRECTED` it shouldn't check the timestamp columns for exceptions. Let me fix that.

razajafri:
build

pxLi:
build

pxLi:
build

pxLi:
NFS issue, reported to SRE

pxLi:
build

sql-plugin/src/main/scala/com/nvidia/spark/RebaseHelper.scala
```diff
@@ -37,7 +37,16 @@ object RebaseHelper extends Arm {
...
+    } else {
+      false
+    }
+  }
+
+  private[this] def isTimeRebaseNeeded(column: ColumnVector,
+      startDay: Int,
+      startTs: Long): Boolean = {
+    val dtype = column.getType
+    if (dtype.isTimestampType) {
```
jlowe, line 49: This would allow `TIMESTAMP_DAYS`.
```suggestion
    if (dtype.hasTimeResolution) {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -202,8 +214,13 @@ class GpuParquetFileFormat extends ColumnarFileFormat with Logging {
 
     val conf = ContextUtil.getConfiguration(job)
 
+    val outputTimestampType = sparkSession.sessionState.conf.parquetOutputTimestampType
     val dateTimeRebaseException = "EXCEPTION".equals(
         sparkSession.sqlContext.getConf(ShimLoader.getSparkShims.parquetRebaseWriteKey))
+    val int96RebaseException =
```
jlowe, line 220: Rather than calling this int96RebaseException, it might be simpler to evaluate it as timestampRebaseException and check for (INT96 && EXCEPTION) || (!INT96 && dateTimeRebaseException).

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -298,15 +315,25 @@ class GpuParquetWriter(
...
+      // if col is a day
+      if (dateTimeRebaseException && col.getType == DType.TIMESTAMP_DAYS) {
+        if (RebaseHelper.isDateRebaseNeeded(col)) {
+          throw DataSourceUtils.newRebaseExceptionInWrite("Parquet")
+        }
+      }
+      // if col is a time
+      else if (col.getType.hasTimeResolution && (int96RebaseException ||
+          outputTimestampType != ParquetOutputTimestampType.INT96 && dateTimeRebaseException)) {
+        if (RebaseHelper.isTimeRebaseNeeded(col)) {
```
jlowe, line 336: Related to the previous comments, the rebase helper method already checks the column types.  If we make the boolean track whether or not we need to check timestamp rebasing, this then would simplify to:
```scala
      else if (timestampRebaseException && RebaseHelper.isTimeRebaseNeeded(col)) {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -223,6 +238,21 @@ object GpuParquetPartitionReaderFactoryBase {
...
+  def isCorrectedInt96RebaseMode(
+      lookupFileMeta: String => String,
+      isCorrectedModeConfig: Boolean): Boolean = {
+    // If there is no version, we return the mode specified by the config.
+    Option(lookupFileMeta(SPARK_VERSION_METADATA_KEY)).map { version =>
+      // Files written by Spark 3.0 and earlier follow the legacy hybrid calendar and we need to
+      // rebase the INT96 timestamp values.
+      // Files written by Spark 3.1 and latter may also need the rebase if they were written with
+      // the "LEGACY" rebase mode.
+      version >= "3.1.0" && lookupFileMeta(SPARK_LEGACY_INT96) == null
```
jlowe, line 253: Doesn't this need to also behave like `isCorrectedRebaseMode` if the recorded Spark version is < "3.1.0"?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -1149,10 +1208,19 @@ class MultiFileCloudParquetPartitionReader(
...
       }
       closeOnExcept(table) { _ =>
-        if (!isCorrectRebaseMode) {
-          (0 until table.getNumberOfColumns).foreach { i =>
-            if (RebaseHelper.isDateTimeRebaseNeededRead(table.getColumn(i))) {
-              throw RebaseHelper.newRebaseExceptionInRead("Parquet")
+        (0 until table.getNumberOfColumns).foreach { i =>
+          val col = table.getColumn(i)
+          // if col is a day
+          if (!isCorrectRebaseMode && col.getType == DType.TIMESTAMP_DAYS) {
```
jlowe, line 1214: Same comments as the other copies of this code.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -902,10 +944,19 @@ class MultiFileParquetPartitionReader(
...
 
     closeOnExcept(table) { _ =>
-      if (!extraInfo.isCorrectedRebaseMode) {
-        (0 until table.getNumberOfColumns).foreach { i =>
-          if (RebaseHelper.isDateTimeRebaseNeededRead(table.getColumn(i))) {
-            throw RebaseHelper.newRebaseExceptionInRead("Parquet")
+      (0 until table.getNumberOfColumns).foreach { i =>
+        val col = table.getColumn(i)
+        // if col is a day
+        if (!extraInfo.isCorrectedRebaseMode && col.getType == DType.TIMESTAMP_DAYS) {
```
jlowe, line 950: Same comment about unnecessary type checks here and just below.  Would be nice to factor out this as a utility rather than copy+paste the code.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -1291,10 +1360,19 @@ class ParquetPartitionReader(
...
         }
         closeOnExcept(table) { _ =>
-          if (!isCorrectedRebaseMode) {
-            (0 until table.getNumberOfColumns).foreach { i =>
-              if (RebaseHelper.isDateTimeRebaseNeededRead(table.getColumn(i))) {
-                throw RebaseHelper.newRebaseExceptionInRead("Parquet")
+          (0 until table.getNumberOfColumns).foreach { i =>
+            val col = table.getColumn(i)
+            // if col is a day
+            if (!isCorrectedRebaseMode && col.getType == DType.TIMESTAMP_DAYS) {
```
jlowe, line 1366: Same comments as the other copies of this code.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
razajafri: Yes, and it will because `int96RebaseMode` is based on the `dateTimeRebaseMode` if Spark version < 3.1.0

jlowe: Sorry, maybe I'm missing something here.  It looks like this method is being unconditionally called, with the result stored in a variable `isCorrectedInt96RebaseForThisFile`, to determine the setting for this file.  If this is called on a file written by Spark 3.0 then this will always return false because it will fail the first `version >= "3.1.0"` clause.  For files written by Spark versions >= 3.0 and < 3.1, I think this needs to be:
```
  (version >= "3.1.0" && lookupFileMeta(SPARK_LEGACY_INT96) == null)
    || (version >= "3.0.0" && lookupFileMeta(SPARK_LEGACY_DATETIME == null)
```
or maybe a bit more elaborately for better readability:
```
  if (version >= "3.0.0") {
    // Starting with Spark 3.1, INT96 has a separate metadata key from datetime conversions
    val metaKey = if (version >= "3.1.0") SPARK_LEGACY_INT96 else SPARK_LEGACY_DATETIME
    lookupFileMeta(metaKey) == null
  } else {
    false
  }
```
This would then not always return false for files written by Spark 3.0.0 since it would lookup the old datetime metadata key being used for INT96 on those files.

razajafri:
build

razajafri:
build

razajafri:
build

pxLi:
@razajafri hi, I saw you triggered premerge twice but not abort the previous build, any reason?

razajafri:
> @razajafri hi, I saw you triggered premerge twice but not abort the previous build, may I ask why? each build would take 2GPUs

Sorry, that was not the intention. I must have hit enter twice.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -114,6 +114,25 @@ object GpuParquetScanBase {
...
+      isCorrectedInt96Rebase: Boolean,
+      isCorrectedDateTimeRebase: Boolean): Unit = {
+    (0 until table.getNumberOfColumns).foreach { i =>
+      val col = table.getColumn(i)
+      // if col is a day
+      if (!isCorrectedDateTimeRebase && RebaseHelper.isDateRebaseNeeded(col)) {
+        throw DataSourceUtils.newRebaseExceptionInRead("Parquet")
+      }
+      // if col is a time
+      else if (!isCorrectedInt96Rebase || !isCorrectedDateTimeRebase) {
```
jlowe, line 128: This logic looks incorrect.  If INT96 config is specified and says don't throw but the legacy datetime config says throw, we still shouldn't throw on a timestamp column **if** it's INT96.  I think this method needs to take two boolean parameters, one to indicate whether we're checking for a rebase on dates and another that says whether we're checking for a rebase on timestamps.  The caller is responsible for passing the proper boolean, as this code is not checking whether we're writing or reading INT96 here.

I'm also not seeing how the config settings are factoring into this code path.  It looks like the parameters passed are simply what's read from the file, regardless of the read or write config settings specified by the user.

sql-plugin/src/main/scala/com/nvidia/spark/RebaseHelper.scala
```diff
@@ -23,9 +23,9 @@ import org.apache.spark.sql.catalyst.util.RebaseDateTime
 import org.apache.spark.sql.rapids.execution.TrampolineUtil
 
 object RebaseHelper extends Arm {
-  private[this] def isDateTimeRebaseNeeded(column: ColumnVector,
-                                                startDay: Int,
-                                                startTs: Long): Boolean = {
+  private[this] def isDateRebaseNeeded(column: ColumnVector,
+      startDay: Int,
+      startTs: Long): Boolean = {
```
jlowe, line 28: `startTs` is unused.

sql-plugin/src/main/scala/com/nvidia/spark/RebaseHelper.scala
```diff
@@ -37,7 +37,16 @@ object RebaseHelper extends Arm {
...
         }
       }
-    } else if (dtype.isTimestampType) {
+    } else {
+      false
+    }
+  }
+
+  private[this] def isTimeRebaseNeeded(column: ColumnVector,
+      startDay: Int,
```
jlowe, line 46: `startDay` is unused.

revans2:
> @razajafri hi, I saw you triggered premerge twice but not abort the previous build, may I ask why? each build would take 2GPUs

If it is that critical that we abort the first job when a second one for the same PR is launched, then we should make it an official part of the tool.  Developers are not going to notice this in all cases. I know I have triggered a build at the same time someone else does and neither of us notice until a long time later.  @pxLi If there is not an issue already can we please file one to add this in as a feature to the pre-merge build?

razajafri:
build

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -217,10 +217,14 @@ class GpuParquetFileFormat extends ColumnarFileFormat with Logging {
     val outputTimestampType = sparkSession.sessionState.conf.parquetOutputTimestampType
     val dateTimeRebaseException = "EXCEPTION".equals(
       sparkSession.sqlContext.getConf(ShimLoader.getSparkShims.parquetRebaseWriteKey))
+    // prior to spark 311 int96 don't check for rebase exception
+    // https://github.com/apache/spark/blob/068465d016447ef0dbf7974b1a3f992040f4d64d/sql/core/src/
+    // main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetWriteSupport.scala#L195
+    val isBeforeSpark311 = ShimLoader.getSparkShims.getSparkShimVersion.toString < "3.1.1"
```
jlowe, line 223: This has similar problems like I raised at https://github.com/NVIDIA/spark-rapids/pull/3330#discussion_r698554846.  A particular vendor flavor of Spark could have cherry-picked this behavior, so this needs to be a call _into_ the shim to check for the feature we're looking for rather than a direct check on the shim version string.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -252,6 +312,19 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
...
   private val rebaseMode = ShimLoader.getSparkShims.parquetRebaseRead(sqlConf)
   private val isCorrectedRebase = "CORRECTED" == rebaseMode
+  val int96RebaseMode = ShimLoader.getSparkShims.int96ParquetRebaseRead(sqlConf)
+  private val isInt96CorrectedRebase = "CORRECTED" == int96RebaseMode
+
+
+  def isParquetTimeInInt96(parquetType: Type): Boolean = {
+    parquetType match {
+      case p:PrimitiveType
+        if p.getPrimitiveTypeName == PrimitiveTypeName.INT96 => true
```
jlowe, line 322:
```suggestion
      case p:PrimitiveType =>
        p.getPrimitiveTypeName == PrimitiveTypeName.INT96
```

pxLi:
> > @razajafri hi, I saw you triggered premerge twice but not abort the previous build, may I ask why? each build would take 2GPUs
> 
> If it is that critical that we abort the first job when a second one for the same PR is launched, then we should make it an official part of the tool. Developers are not going to notice this in all cases. I know I have triggered a build at the same time someone else does and neither of us notice until a long time later. @pxLi If there is not an issue already can we please file one to add this in as a feature to the pre-merge build?

@revans2 sorry for the late response. The reasons we let developer to decide if they need to abort manually,
1. People may just want to trigger twice to see the consistency of tests they added
2. For 2 triggered builds for the same/diff commits, we do not know which one should be kept (probably the first one since it has been running for some time)
3. From CICD side, it's hardly to know if we should kill a trigger, if the developer want the build result in some cases

If this a must have, we can limit usage (one PR, at most one running build), and abort all previous builds if exist

jlowe:
> If this a must have, we can limit usage (one PR, at most one running build), and abort all previous builds if exist

Please implement this.  I don't know of anyone who kicks off two builds on purpose.

> People may just want to trigger twice to see the consistency of tests they added

They can run these serially.  No compelling reason to allow them to take up even more CI resources concurrently for a single PR .

> For 2 triggered builds for the same/diff commits, we do not know which one should be kept

Always the latter one.  Any former build for the same PR should be killed.

> it's hardly to know if we should kill a trigger, if the developer want the build result in some cases

I haven't heard a compelling use-case for a developer _needing_ the ability to run more than one CI concurrently for a PR, and there's clearly the expectation and desire from developers that a subsequent build supercedes a previous one.

pxLi:
> > If this a must have, we can limit usage (one PR, at most one running build), and abort all previous builds if exist
> 
> Please implement this. I don't know of anyone who kicks off two builds on purpose.
> 
> > People may just want to trigger twice to see the consistency of tests they added
> 
> They can run these serially. No compelling reason to allow them to take up even more CI resources concurrently for a single PR .
> 
> > For 2 triggered builds for the same/diff commits, we do not know which one should be kept
> 
> Always the latter one. Any former build for the same PR should be killed.
> 
> > it's hardly to know if we should kill a trigger, if the developer want the build result in some cases
> 
> I haven't heard a compelling use-case for a developer _needing_ the ability to run more than one CI concurrently for a PR, and there's clearly the expectation and desire from developers that a subsequent build supercedes a previous one.

@jlowe sounds OK. I filed a ticket https://github.com/NVIDIA/spark-rapids/issues/3486 here and will work on this later

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -202,8 +214,18 @@ class GpuParquetFileFormat extends ColumnarFileFormat with Logging {
...
     val conf = ContextUtil.getConfiguration(job)
 
+    val outputTimestampType = sparkSession.sessionState.conf.parquetOutputTimestampType
     val dateTimeRebaseException = "EXCEPTION".equals(
-        sparkSession.sqlContext.getConf(ShimLoader.getSparkShims.parquetRebaseWriteKey))
+      sparkSession.sqlContext.getConf(ShimLoader.getSparkShims.parquetRebaseWriteKey))
+    // prior to spark 311 int96 don't check for rebase exception
+    // https://github.com/apache/spark/blob/068465d016447ef0dbf7974b1a3f992040f4d64d/sql/core/src/
+    // main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetWriteSupport.scala#L195
+    val isBeforeSpark311 = ShimLoader.getSparkShims.isBeforeSpark311
```
jlowe, line 223: This misses the point of my previous comment.  Some Spark platforms could be a _hybrid of Spark changes_ such that a single version check is not appropriate.  As in this case -- we really don't care what version of Spark it is, we're trying to understand whether INT96 has its own set of configs or not.  That's the predicate method we need, a way to tell if INT96 has been separated from datetime.  That's going to be more reliable than checking the Spark version number on Spark platforms that have cherry-picked changes from future Spark versions beyond what version they are actually reporting.

razajafri:
build

jlowe:
Are the Databricks shims covered with this change?  May want to test with the `[databricks]` tag in the PR summary.

razajafri:
build

razajafri:
build

jlowe (dismissed):

razajafri:
build

razajafri:
build

jlowe (approved):

razajafri:
Looks like a failure unrelated to my change in DB, root causing it

razajafri:
Skipping `test_no_fallback_when_ansi_enabled` until a resolution to #3611

razajafri:
build

razajafri:
build

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -1160,6 +1160,7 @@ def do_it(spark):
 
 
 @pytest.mark.parametrize('data_gen', _no_overflow_ansi_gens, ids=idfn)
+@ignore_order
```
jlowe, line 1163:
```suggestion
@ignore_order(local=True)
```

razajafri:
build

razajafri:
@revans2 or @jlowe can you please +1 this and merge before we see any more merge conflicts?

jlowe (approved):

tgravescs:
this pr is not up to date with latest moves of of shim files, reverted https://github.com/NVIDIA/spark-rapids/pull/3627

PR #3335, opened 2021-08-30
Support Databricks 3.1.1 in new build profiles

tgravescs:
This unfortunately required a bunch more copy of files.

relates to https://github.com/NVIDIA/spark-rapids/issues/3224

1) Copy files just for existing shim layer for both 301db and 311db.
2) Needed to move the *Pandas* files into shim layer due to databricks differences
3) A few other functions needed new shim interfaces
4) A few places ShuffleQueryStageExec and others use different parameter numbers so made checks more generic
5) Installed a lot more databricks jars into the .m2 and updated pom dependencies to use those jars rather then the spark 3.x ones and pom file.  Note we don't have their pom files so dependencies aren't pulled directly. Currently thi si sonly for databricks 3.1.1, 3.0.1 databricks will be done separately.
6) removed 301emr from build since we don't test it, I can pull this out if people want.
7) Added TaskContextTrampoline in databricks 3.0.1 to be able to get to the memory manager. 3.1.1 has that api public.


tgravescs:
build

tgravescs:
running build looks like some shim change went in, I need update

tgravescs:
actually it looks like the change didn't update the databricks shim

tgravescs:
build

pom.xml
```diff
@@ -167,6 +173,28 @@
                 <spark.version>${spark311.version}</spark.version>
                 <spark.test.version>${spark311.version}</spark.test.version>
             </properties>
+            <modules>
+                <module>api_validation</module>
```
jlowe, line 177: We're not building versions of these for Databricks?

pom.xml
```diff
@@ -315,24 +349,36 @@
             <properties>
                 <spark.test.version>${spark301.version}</spark.test.version>
             </properties>
+            <modules>
+                <module>tests</module>
```
jlowe, line 353: `tests` is already a module that's included unconditionally?

pom.xml
```diff
@@ -424,57 +481,56 @@
...
             </dependency>
             <dependency>
-                <groupId>org.scala-lang</groupId>
-                <artifactId>scala-library</artifactId>
-                <version>${scala.version}</version>
-                <scope>provided</scope>
+              <groupId>org.slf4j</groupId>
+              <artifactId>jcl-over-slf4j</artifactId>
+              <version>${slf4j.version}</version>
+              <!-- runtime scope is appropriate, but causes SBT build problems -->
```
jlowe, line 509: Do we need the same comment for `jul-to-slf4j`?

shims/pom.xml
```diff
@@ -42,7 +42,6 @@
                 <activeByDefault>true</activeByDefault>
             </activation>
             <modules>
-                <module>spark301emr</module>
```
jlowe, line 45: Does nothing build spark301emr now?  I'd rather rip out the code if we aren't even going to regularly verify it remains buildable.

shims/spark301db/src/main/scala/org/apache/spark/rapids/shims/spark301db/TaskContextTrampoline.scala
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package org.apache.spark.rapids.shims.spark301db
+
+import org.apache.spark.TaskContext
+import org.apache.spark.memory.TaskMemoryManager
+
+object TaskContextTrampoline {
+
+  def getTaskMemoryManager(): TaskMemoryManager = {
```
jlowe, line 24: Why isn't this in TrampolineUtil?

tests/pom.xml
```diff
@@ -164,6 +164,149 @@
...
+                </dependency>
+                <dependency>
+                    <groupId>org.apache.spark</groupId>
+                    <artifactId>spark-core_${scala.binary.version}</artifactId>
+                    <version>${spark311db.version}</version>
+                    <scope>provided</scope>
+                </dependency>
+                <dependency>
+                    <groupId>org.apache.spark</groupId>
+                    <artifactId>spark-network-common_${scala.binary.version}</artifactId>
```
jlowe, line 190: This seems like a second-level dependency which we shouldn't have to list here.  If we do depend on it directly, why isn't this or others listed for the other profiles?

jenkins/databricks/build.sh
```diff
@@ -61,7 +64,11 @@ RAPIDS_BUILT_JAR=rapids-4-spark_$SCALA_VERSION-$SPARK_PLUGIN_JAR_VERSION.jar
 RAPIDS_UDF_JAR=rapids-4-spark-udf-examples_$SCALA_VERSION-$SPARK_PLUGIN_JAR_VERSION.jar
 
 echo "Scala version is: $SCALA_VERSION"
-mvn -B -P${BUILD_PROFILES} clean package -DskipTests || true
+if [[ -z $INSTALL_DEPS_ONLY ]]
+then
+    echo "Running maven to pull Spark dependencies"
+    mvn -B -P${BUILD_PROFILES} clean package -DskipTests || true
```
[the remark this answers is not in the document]
tgravescs: this is how the existing build (before these changes works) it runs this to pull the base 3.X dependencies, installs the databricks jars and then runs again.  I just left it the way it was.  I can certainly look to make it more efficient but if we are switching to the new build way it doesn't do this anymore.

pom.xml
tgravescs: I removed them because we don't validate our api vs databricks version so why build it?

pom.xml
tgravescs: thanks for catching it, I moved to under some profiles and then moved it back and didn't remove these

shims/spark301db/src/main/scala/org/apache/spark/rapids/shims/spark301db/TaskContextTrampoline.scala
tgravescs: Since it was just needed by this one shim, I put it in here, I can move it to common if we prefer?

shims/pom.xml
tgravescs: no, nothing builds it. I figured if we don't have their jars no reason to build it.  I left it thinking it might be useful for EMR folks.  I can rip it out or we can talk to them to get more details on how they build?

jlowe: IMO either we build it or we remove it.  If we leave it unbuilt, I guarantee it will rot over time to the point it no longer builds.  If it's useful to the EMR team then we should have them pull in the changes to their repo where they build the EMR shim.

shims/spark301db/src/main/scala/org/apache/spark/rapids/shims/spark301db/TaskContextTrampoline.scala
jlowe: Yes, there may be other cases in the future where we need to get the TaskMemoryManager

shims/pom.xml
tgravescs: ok, I'll just remove it

jlowe: We might want to do that as a separate PR rather than as part of adding Databricks.  This is already a really big PR.

tests/pom.xml
tgravescs: copy and paste error. I cleaned up, but I had lost a bunch of changes due to data bricks instance shutdown on me, thought I remade but missed some. There are a few others that go along with this.

shims/pom.xml
tgravescs: ok

tgravescs:
build

integration_tests/pom.xml
```diff
@@ -162,8 +162,72 @@
                     <groupId>org.apache.curator</groupId>
                     <artifactId>curator-recipes</artifactId>
                     <version>4.3.0.7.2.7.0-184</version>
+                </dependency>
+            </dependencies>
+        </profile>
+    <profile>
```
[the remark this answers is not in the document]
tgravescs: fixed

jenkins/databricks/build.sh
```diff
@@ -74,7 +81,75 @@ CATALYSTJAR=----workspace_${SPARK_MAJOR_VERSION_STRING}--sql--catalyst--catalyst
...
+   -DartifactId=spark-annotation_$SCALA_VERSION \
+   -Dversion=$SPARK_VERSION_TO_INSTALL_DATABRICKS_JARS \
+   -Dpackaging=jar
+
+
+if [[ -z $INSTALL_DEPS_ONLY ]]
+then
+
+# this install pom file from Spache Spark 3.1.1, whereas new build doesn't
+mvn dependency:get  -DgroupId=org.apache.spark -DartifactId=spark-core_$SCALA_VERSION -Dversion=$BASE_SPARK_VERSION
```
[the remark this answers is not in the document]
tgravescs: updated

tgravescs:
Updated and reran builds on databricks and apache.
I rechecked the databricks built jar with this for default build vs existing build and only differences are due to removed spark311 shim files and moved *Pandas* files.

tests/pom.xml
```diff
@@ -164,6 +164,109 @@
...
+                    <version>${spark311db.version}</version>
+                    <scope>provided</scope>
+                </dependency>
+                <dependency>
+                    <groupId>com.esotericsoftware.kryo</groupId>
+                    <artifactId>kryo-shaded-db</artifactId>
+                    <version>${spark311db.version}</version>
+                    <scope>provided</scope>
+                </dependency>
+		<!--
```
tgravescs, line 230: need to remove

tgravescs:
build

pom.xml
```diff
@@ -111,20 +122,9 @@
                 <spark.version>${spark302.version}</spark.version>
                 <spark.test.version>${spark302.version}</spark.test.version>
             </properties>
-        </profile>
-        <profile>
-            <id>release301emr</id>
```
jlowe, line 116: I thought we weren't going to pull the EMR shim out in this PR?

tgravescs: ok put it all back so the default build now contains it.

tgravescs:
build

jlowe (approved):

tgravescs:
error on tearing down one of the tests, trying to reproduce to see if related to this or something happened during build:
>  _ ERROR at teardown of test_write_sql_save_table

tgravescs:
build

PR #3341, opened 2021-08-30
ArrayMax and ArrayMin support plus a few other simple ops

revans2:
This is a first step at doing some array aggregations.  It implements array_min and array_max.  In the next step I hope to be able to do pattern matching on the higher order functions in aggregation and be able to implement a subset of it using a combination of `transform` which is already implemented and some more generalized array aggregation code.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuListUtils.scala
```diff
@@ -0,0 +1,54 @@
...
+   * Replace the data column in a LIST column. This will keep the same offsets and validity
+   * of the listColumn.  This returns a view so it is the responsibility of the caller to keep
+   * both listCol and newDataCol alive longer than the returned ColumnView.
+   * @param listCol the list column to use as a template
+   * @param newDataCol the new data column.
+   * @return a new ColumnView.
+   * @throws IllegalArgumentException if data column does not match the original data column in
+   *                                  size.
+   */
+  def replaceListDataColumn(
```
jlowe, line 37: Given this is intended to be a general utility, should this have something like "asView" in the name to help drive home the dangerousness of this method?  If someone assumes this creates a column instead of a column view it could get ugly to debug.  Or are you convinced that other APIs that require a ColumnVector instead of a ColumnView would prevent this mistake in practice?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -680,6 +680,29 @@ object GpuOverrides {
   def isAnyStringLit(expressions: Seq[Expression]): Boolean =
     expressions.exists(isStringLit)
 
+  def anyOf(dataType: DataType, fn: DataType => Boolean): Boolean = {
```
jlowe, line 683: I think this already exists as `TrampolineUtil.dataTypeExistsRecursively`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2440,6 +2461,102 @@ object GpuOverrides {
...
+      ExprChecks.unaryProject(
+        TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 + TypeSig.NULL,
+        TypeSig.orderable,
+        TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 + TypeSig.NULL),
+        TypeSig.ARRAY.nested(TypeSig.orderable)),
+      (in, conf, p, r) => new UnaryExprMeta[ArrayMin](in, conf, p, r) {
+        override def tagExprForGpu(): Unit = {
+          if (isOrContainsFloatingPoint(in.dataType)) {
+            if (!conf.isFloatAggEnabled) {
+              willNotWorkOnGpu("the GPU will aggregate floating point values in" +
```
jlowe, line 2515: Nit: This is repeated below and almost identical to the text used for `Sum` and `Average` aggregates.  Should this be directly reused each time we want to reference this property rather than copy+paste?

revans2:
Moved this into draft because I am seeing some illegal memory access errors when running the tests in parallel, but not when running the tests one at a time. I need to spend some time debugging this.

revans2:
It looks like min and max on strings when doing a sorted aggregation can produce the IllegalMemoryAccess problem.  I will try to reproduce it in cudf.

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2504,26 +2482,16 @@ object GpuOverrides {
...
-                  " more than once as part of the same query.  To enable this anyways set" +
-                  s" ${RapidsConf.ENABLE_FLOAT_AGG} to true.")
-            }
-            if (conf.hasNans) {
-              willNotWorkOnGpu("Min aggregation on floating point columns that can contain NaNs" +
-                  " will compute incorrect results. If it is known that there are no NaNs, set" +
-                  s" ${RapidsConf.HAS_NANS} to false.")
-            }
-          }
+          checkAndTagFloatNanAgg("Min", in.dataType, conf, this)
```
jlowe, line 2494: Looks like there's a missing call to `checkAndTagFloatAgg` here?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2532,26 +2500,16 @@ object GpuOverrides {
...
-                  " more than once as part of the same query.  To enable this anyways set" +
-                  s" ${RapidsConf.ENABLE_FLOAT_AGG} to true.")
-            }
-            if (conf.hasNans) {
-              willNotWorkOnGpu("Max aggregation on floating point columns that can contain NaNs" +
-                  " will compute incorrect results. If it is known that there are no NaNs, set" +
-                  s" ${RapidsConf.HAS_NANS} to false.")
-            }
-          }
+          checkAndTagFloatNanAgg("Max", in.dataType, conf, this)
```
jlowe, line 2512: Missing `checkAndTagFloatAgg` call here?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuListUtils.scala
```diff
@@ -31,16 +31,17 @@ object GpuListUtils extends Arm {
...
    */
-  def replaceListDataColumn(
+  def replaceListDataColumnAsView(
       listCol: ColumnView,
       newDataCol: ColumnView): ColumnView = {
     assert(DType.LIST.equals(listCol.getType))
     withResource(listCol.getChildColumnView(0)) { dataCol =>
       if (dataCol.getRowCount != newDataCol.getRowCount) {
-        throw new IllegalArgumentException("Mismatch in the number of rows in the data columns")
+        throw new scala.IllegalArgumentException(
```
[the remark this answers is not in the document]
revans2: The compiler output a warning for the scala docs because it could not tell which one it was. So for consistency I made them both the same.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
revans2: No that was a bug in the old code. min and max always produce the correct answer (assuming no NaNs) because the order of operations does not matter for them in floating point.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
revans2: Same here I fixed a bug I had before that I caught because I combined the checks and I had to think why the regular min and max aggregations didn't do the check I added for the array versions.

jlowe (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuListUtils.scala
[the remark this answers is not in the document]
revans2: As a side note. The scala one just maps to the java one so scala is just confusing itself for no good reason.

Issue #3047, opened 2021-07-27
[FEA]Profiling tool: Structured output format


nartal1:
**Is your feature request related to a problem? Please describe.**
As a user, I want to save the output of the profiling tool(for "Job + Stage level aggregated task metrics" and "SQL level aggregated task metrics" 2 sections) as structured output format csv/parquet so that I can analyze using custom queries.

**Additional context**
**Acceptance criteria:** Able to run customized queries on aggregated task metrics for all processed applications.

PR #3342, opened 2021-08-30
Profiling tool add CSV output option and add new combined mode

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/3047

Add an option to the profiling tool to output csv files in addition to the text output. 

While doing this I also got a request to run the profiling tool on 10,000+ event logs at a time and output a single table.  To support that I made some changes for memory management and added a new combined mode.  The combined mode is the same as compare where all applications get put into a single table, except for it doesn't output the 2 tables that actually try to compare sqlids and job ids between the applications.   The main scale issue here was with the memory needed to keep all the raw application info in memory.  I changed it so that it processes each application and then just outputs a summary information which is much smaller.  for the new combined mode it processes each application separately in the thread pool and then at the end combines all the summary information into single tables.   I processed  11618 apps in 20 minutes using the combined mode. I did require quite a bit of memory still - around 120GB. I didn't try with less.   The combined mode options outputs files into a /combined/ subdirectory.

To support outputting to csv I changed most of the functions to be getters and then we print the information separately. ProfileOutputWriter was extended to know how to output text and csv format.


tgravescs:
build

tgravescs:
build

tgravescs:
Note, I ran over the same 11618 apps with 50GB memory and it took 31 minutes.

tgravescs:
looks like my branch isn't up to date, will upmerge

tgravescs:
build

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/CollectInformation.scala
```diff
@@ -330,4 +223,53 @@ object CollectInformation extends Logging {
...
+      processedProps: HashMap[String, ArrayBuffer[String]],
+      numApps: Int): Unit = {
+    val inter = processedProps.keys.toSeq.intersect(newRapidsRelated.keys.toSeq)
+    val existDiff = processedProps.keys.toSeq.diff(inter)
+    val newDiff = newRapidsRelated.keys.toSeq.diff(inter)
+
+    // first update intersecting
+    inter.foreach { k =>
+      val appVals = processedProps.getOrElse(k, ArrayBuffer[String]())
+      appVals += newRapidsRelated.getOrElse(k, "null")
```
nartal1, line 256: Sorry, I don't understanding this.  Could you please explain what does `appVals` here and below in existDiff used for? 
I see appVals being created locally but that is not being updated in any HashMap.

docs/additional-functionality/qualification-profiling-tools.md
```diff
@@ -327,15 +327,18 @@ The profiling tool generates information which can be used for debugging and pro
 ### Profiling tool functions
 
 Below is the information the profiling tool reports, see [Profile tool Detailed Output and Examples](#profile-tool-detailed-output-and-examples) for more information
-and examples. It has 2 main modes of operation: collection and compare.  Collection mode is when no other options are specified it simply collects information
-on each application individually and outputs a file per application. Compare mode will combine all the applications information in the same tables into a single file.
-It can also optionally output the SQL plan, the SQL graphs, and a timeline graph for each application when in collection mode.
+and examples. It has 3 modes of operation: collection, combined, and compare.  Collection mode is when no other options are specified it simply collects information
+on each application individually and outputs a file per application. Combined mode is collection mode but them combines all the applications into single tables
```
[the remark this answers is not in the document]
tgravescs: actually I meant to have "tables" because it outputs multiple tables.  Them should definitely be then.

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/CollectInformation.scala
tgravescs: so this is updating the values in processedProps itself.  appVals gets the Arraybuffer of properties that match the key.  ie key that has multiple values.  We adde any new values to that for that certain key.

This is because the table is backwards from other tables. Here the rows keys are the property keys and the app values are the columns.  So the ArrayBuffer here is the app values for each key.

We have to handle multiple cases to make sure all the array buffers for each key get updated with the same number of entries.  
1) key already existed
2) this app doesn't contain a key that other apps had
3) new key that wasn't in the other apps we processed

nartal1: Thanks for the explanation.  It makes sense. Didn't realize we were updating processedProps itself.

tgravescs:
build

PR #3344, opened 2021-08-30
Move classes out of Apache Spark packages

### LINKED ISSUE #3235: Classes directly in Apache Spark packages
Most classes are either under the com.nvidia.spark.rapids, org.apache.spark.sql.rapids, or some other Spark package prefix that ends in `.rapids`, but there are a few classes in the plugin that appear directly within an Apache Spark package:
- org.apache.spark.sql.catalyst.CudfUnsafeRow
- org.apache.spark.shuffle.RapidsShuffleExceptions.scala

Ideally these classes should be moved to an existing `*.rapids` package.

jbrennan333:
Fixes #3235

### DIFF
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/catalyst/CudfUnsafeRow.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/catalyst/CudfUnsafeRow.java
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
-package org.apache.spark.sql.catalyst;
+package com.nvidia.spark.rapids.catalyst;
 
 import com.nvidia.spark.rapids.GpuColumnVector;
+import org.apache.spark.sql.catalyst.InternalRow;
 import org.apache.spark.sql.catalyst.expressions.Attribute;
 import org.apache.spark.sql.catalyst.expressions.SpecializedGettersReader;
 import org.apache.spark.sql.catalyst.util.ArrayData;
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/CudfUnsafeRow.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/CudfUnsafeRow.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2021, NVIDIA CORPORATION.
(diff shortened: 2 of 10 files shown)


jbrennan333:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/catalyst/CudfUnsafeRow.java
```diff
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
-package org.apache.spark.sql.catalyst;
+package com.nvidia.spark.rapids.catalyst;
```
jlowe, line 17: Nit: Why isn't this just under com.nvidia.spark.rapids?  The `com.nvidia.spark.rapids.catalyst` package doesn't exist before this PR, and arguably many things under the former would be candidates for the latter since they deal with Catalyst.

sql-plugin/src/main/java/com/nvidia/spark/rapids/catalyst/CudfUnsafeRow.java
```diff
@@ -14,9 +14,10 @@
...
 
 import com.nvidia.spark.rapids.GpuColumnVector;
+import org.apache.spark.sql.catalyst.InternalRow;
```
jlowe, line 20: There's supposed to be a whitespace between the Spark imports.  See https://github.com/NVIDIA/spark-rapids/blob/branch-21.10/CONTRIBUTING.md#code-formatting and the sample code formatting file.

sql-plugin/src/main/java/com/nvidia/spark/rapids/catalyst/CudfUnsafeRow.java
jbrennan333: Agreed.  I will move it to com.nvidia.spark.rapids

sql-plugin/src/main/java/com/nvidia/spark/rapids/catalyst/CudfUnsafeRow.java
jbrennan333: Thanks!  This helped me realize I was not using the correct profile in IntelliJ.

jlowe (approved):

jlowe:
build

PR #3361, opened 2021-09-01
Fix databricks 3.1.1 arrow dependency version

tgravescs:
So this was previous bug with arrow version number in hte spark311db shim layer.  It referenced 0.15.1 but should have been 2.0.0. Previously the Databricks 3.1.1 shim used the arrow functionality from the Spark 3.1.1 shim which had the correct arrow dependency and the dependency in the spark311db shim was never used. Once we changed to build everything per shim this broke and I didn't catch the failing test because other tests were failing.

The datasource arrow test now passing, I'm in progress of running them all as well.

### DIFF
--- a/shims/spark311db/pom.xml
+++ b/shims/spark311db/pom.xml
@@ -105,7 +105,7 @@
             </activation>
             <properties>
                 <parquet.version>1.10.1</parquet.version>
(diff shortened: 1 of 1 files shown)


tgravescs:
build

jlowe (approved):

Issue #3333, opened 2021-08-29
[BUG]Mortgage ETL input_file_name is not correct when using CPU's CsvScan


GaryShen2008:
**Describe the bug**
When reading CVS without settting "--conf spark.rapids.sql.csv.read.long.enabled=true", the "quarter" column in Mortgage ETL will show empty value when selecting multiple columns, but it can be shown correctly when only selecting itself.
Related bug: https://github.com/NVIDIA/spark-rapids-examples/issues/15

**Steps/Code to reproduce bug**
The code is [XGBoostETL.scala](https://github.com/NVIDIA/spark-rapids-examples/blob/branch-21.08/examples/mortgage/scala/src/com/nvidia/spark/examples/mortgage/XGBoostETL.scala)

spark-shell --master spark://master_ip:7077 --conf spark.plugins=com.nvidia.spark.SQLPlugin --executor-memory 10G --executor-cores 5 --driver-memory 4G --conf spark.cores.max=5 --conf spark.executor.resource.gpu.amount=1 --conf spark.task.cpus=1 --conf spark.rapids.sql.incompatibleDateFormats.enabled=true

``` scala
scala> :paste

import org.apache.spark.sql.expressions.Window
import org.apache.spark.sql.functions._
import org.apache.spark.sql.types._
import org.apache.spark.sql.{Column, DataFrame, SparkSession}

object GetQuarterFromCsvFileName {
  // The format is path/TYPE_yyyy\QQ.txt followed by a (_index)* where index is a single digit number [0-9]
  // i.e. mortgage/perf/Performance_2003Q4.txt_0_1
  // So we strip off the .txt and everything after it
  // and then take everything after the last remaining _
  def apply(): Column = substring_index(
    substring_index(input_file_name(), ".", 1), "_", -1)
}

object CsvReader {

  def readPerformance(spark: SparkSession, paths: Seq[String], optionsMap: Map[String, String]): DataFrame = {
    val performanceSchema = StructType(Array(
      StructField("loan_id", LongType),
      StructField("monthly_reporting_period", StringType),
      StructField("servicer", StringType),
      StructField("interest_rate", DoubleType),
      StructField("current_actual_upb", DoubleType),
      StructField("loan_age", DoubleType),
      StructField("remaining_months_to_legal_maturity", DoubleType),
      StructField("adj_remaining_months_to_maturity", DoubleType),
      StructField("maturity_date", StringType),
      StructField("msa", DoubleType),
[...]

PR #3366, opened 2021-09-02
Fix input_file_xxx issue when FileScan is running on CPU

wbo4958:
This PR tries to force plans ( from SparkPlan (with the first input_file_xxx) to FileScan ) to run on CPU, Please see the detail from issue https://github.com/NVIDIA/spark-rapids/issues/3333

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -3459,7 +3459,6 @@ case class GpuOverrides() extends Rule[SparkPlan] with Logging {
       } else {
         Seq.empty
       }
-      InputFileBlockRule.apply(wrap)
       wrap.runAfterTagRules()
       if (conf.shouldExplain) {
         wrap.tagForExplain()
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -3459,6 +3459,7 @@ case class GpuOverrides() extends Rule[SparkPlan] with Logging {
       } else {
         Seq.empty
       }
+      InputFileBlockRule.apply(wrap)
       wrap.runAfterTagRules()
       if (conf.shouldExplain) {
         wrap.tagForExplain()
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
@@ -275,23 +275,14 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
     case _ => false
   }
 
-  /**
-   * Because we cannot change the executors in spark itself we need to try and account for
-   * the ones that might have issues with coalesce here.
-   */
-  private def disableCoalesceUntilInput(exec: Expression): Boolean = exec match {
-    case _: InputFileName => true
-    case _: InputFileBlockStart => true
-    case _: InputFileBlockLength => true
-    case e => e.children.exists(disableCoalesceUntilInput)
-  }
+
 
   /**
    * Because we cannot change the executors in spark itself we need to try and account for
    * the ones that might have issues with coalesce here.
    */
   private def disableCoalesceUntilInput(plan: SparkPlan): Boolean = {
-    plan.expressions.exists(disableCoalesceUntilInput)
+    plan.expressions.exists(GpuTransitionOverrides.checkHasInputExpressions)
   }
 
   private def disableScanUntilInput(exec: Expression): Boolean = {
@@ -573,6 +564,18 @@ object GpuTransitionOverrides {
         }
       case _ => plan
     }
+  }
+
+  /**
+   * Check the Expression is or has Input expressions.
+   * @param exec expression to check
+   * @return true or false
+   */
+  def checkHasInputExpressions(exec: Expression): Boolean = exec match {
+    case _: InputFileName => true
+    case _: InputFileBlockStart => true
+    case _: InputFileBlockLength => true
+    case e => e.children.exists(checkHasInputExpressions)
   }
 }
 
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
@@ -282,7 +282,7 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
    * the ones that might have issues with coalesce here.
    */
   private def disableCoalesceUntilInput(plan: SparkPlan): Boolean = {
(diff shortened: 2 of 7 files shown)


wbo4958:
build

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3459,6 +3459,7 @@ case class GpuOverrides() extends Rule[SparkPlan] with Logging {
       } else {
         Seq.empty
       }
+      InputFileBlockRule.apply(wrap)
```
revans2, line 3462: We probably want to do the same thing with runAfterTagRules. And possibly just make this a part of those rules. They are doing essentially the same kind of thing.  Make sure that two different parts of the plan are both on the CPU or both on the GPU.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/InputFileBlockRule.scala
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * InputFileBlockRule is to prevent the SparkPlans
+ * [SparkPlan (with first input_file_xxx expression), FileScan) to run on GPU
+ *
+ * See https://github.com/NVIDIA/spark-rapids/issues/3333
+ */
+object InputFileBlockRule {
+
+  private def checkHasInputFileExpressions(exec: Expression): Boolean = exec match {
+    case _: InputFileName => true
```
revans2, line 34: With this we now have two different ways of checking for this situation and there is nothing forcing the two together.  At a minimum can we add in comments here and on `GpuExpression.disableCoalesceUntilInput` pointing to each other so we know to try and keep the two in sync? It would be even better if we could come up with a way through tags or something that would make it very difficult for someone to add in a new method that needs this and not update both places.  Although I don't think more methods are coming, so the comments are good enough.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -156,6 +156,12 @@ abstract class RapidsMeta[INPUT <: BASE, BASE, OUTPUT <: BASE](
     childDataWriteCmds.foreach(_.recursiveSparkPlanRemoved())
   }
 
+  final def inputFilePreventsRunningOnGpu(): Unit = {
+    if (canThisBeReplaced) {
+      willNotWorkOnGpu("Removed by InputFileBlockRule")
```
revans2, line 161: nit: A little longer explanation might be good.

revans2 (dismissed):
Generally it looks good. I just have a few nits.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
wbo4958: Hi @revans2, This PR provides the rule to find the spark plan chain first, and then tag them not to run on GPU. runAfterTagRules seems can't satisfy the requirement.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/InputFileBlockRule.scala
wbo4958: I suppose you were meaning `GpuTransitionOverrides.disableCoalesceUntilInput`. I just rename and put the same function in the GpuTransitionOverrides.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
wbo4958: Done

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -574,6 +565,18 @@ object GpuTransitionOverrides {
       case _ => plan
     }
   }
+
+  /**
+   * Check the Expression is or has Input expressions.
+   * @param exec expression to check
+   * @return true or false
+   */
+  def checkHasInputExpressions(exec: Expression): Boolean = exec match {
```
revans2, line 574: nit: can we rename this `checkHasInputFileExpressions`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/InputFileBlockRule.scala
revans2: `GpuExpression.disableCoalesceUntilInput` is called from `GpuExec.disableCoalesceUntilInput` which in turn is called from `GpuTransitionOverrides.insertCoalesce` to figure out if it should be disabled or not for instances of `GpuExec`. For things that are not instances of `GpuExec` the code calls into `GpuTransitionOverrides.disableCoalesceUntilInput`, which was updated to call into `GpuTransitionOverrides.checkHasInputExpressions`.  It would be nice to have a way to keep them in sync so we only have to update things in one place. Now that I look more closely there is no good way to do a comment for this.  I think we could clean up the code more to remove duplication between the two paths, but I think this is fine for now.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
revans2: In order to do the transitions correctly we have rules about what you can, and cannot look at when tagging a part of the plan.

https://github.com/NVIDIA/spark-rapids/blob/f774e1a231416aab7cacc86f06010ef797fc19c7/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala#L293-L297

This makes it so one rule cannot change things in a way that will violate the goals of other rules.

`runAfterTagRules` is just a single location to run all of the rules that do not follow these requirements, so we can be 100% sure that they will not interfere with each other. 

https://github.com/NVIDIA/spark-rapids/blob/f774e1a231416aab7cacc86f06010ef797fc19c7/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala#L627-L643

For example right now `InputFileBlockRule` runs before `fixUpExchangeOverhead` in `runAfterTagRules`. If for some reason it ran after `fixUpExchangeOverhead` instead that would be a bug because it could force things onto the CPU that `fixUpExchangeOverhead` already looked at.

If we put them all together in one place we can reason about what the order should be and work to avoid these bugs early on.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
wbo4958: Done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
wbo4958: Thx for the explanation. Done

wbo4958:
build

revans2 (approved):

PR #3373, opened 2021-09-02
Support `stddev` and `variance` aggregations families [databricks]

### LINKED ISSUE #1601: [FEA] Support AggregationFunction StddevSamp
Provide GPU override of Spark AggregationFunction `StddevSamp`.

It seems that cuDF AggregationOp `Aggregation.standardDeviation` do the same job as `StddevSamp`. So I think, as other GPU overrides of AggregationFunction, we can implement `GpuStddevSamp` on the basis of `Aggregation.standardDeviation`.

ttnghia:
This PR adds `stddev` and `variance` aggregations families (`std`, `stddev`, `stddev_pop`, `stddev_samp`, `variance`, `var_pop` and `var_samp`). Note that `std` and `sttdev` are just aliases of `stddev_samp` and `variance` is just alias of `var_samp`.

Closes #1601.

Design doc: https://docs.google.com/document/d/1k0eCLr5juGRbGw9Ld4zwgysuwP66dnuJs5SeLBmgSbc/edit?usp=sharing


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -891,3 +932,118 @@ case class GpuCollectSet(
...
+ */
+abstract class GpuM2(child: Expression)
+  extends GpuAggregateFunction with ImplicitCastInputTypes {
+
+  // TODO: remove comment below
+  // dataType must be a structs column, and nullable of the output must be false
+  // the struct field `mean` and `m2` can be nullable (but are not always nullable)
+  // so I'm no sure if set `nullable = true` is correct here.
+
+  override def dataType: DataType = StructType(Array(
```
andygrove, line 947: nit: this struct type is defined in two places and it would be good to put it in a constant

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3014,6 +3014,22 @@ object GpuOverrides extends Logging {
 
         override val supportBufferConversion: Boolean = true
       }),
+    expr[StddevPop](
+      "Aggregation computing population standard deviation",
+      ExprChecks.aggNotReduction(
+        TypeSig.DOUBLE, TypeSig.DOUBLE,
+        Seq(ParamCheck("input", TypeSig.gpuNumeric, TypeSig.numeric))),
```
revans2, line 3021: Technically the only value supported by StddevPop as input in spark is Double

https://github.com/apache/spark/blob/adbea252db0f61ed3aaa45ea20b704bad6c47408/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CentralMomentAgg.scala#L58

Could you please update the type signature here and on StddevSamp to only show DOUBLE as the supported input type by both us and Spark?  This should also make testing a lot simpler because you don't need to worry about all of the different types.  Spark will cast them ahead of time to a Double for us.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -418,6 +418,63 @@ class CudfLastExcludeNulls(ref: Expression) extends CudfFirstLastBase(ref) {
...
+    throw new UnsupportedOperationException("M2 aggregation is not yet supported in reduction")
+
+  override lazy val updateAggregate: GroupByAggregationOnColumn =
+    GroupByAggregation.M2().onColumn(getOrdinal(ref))
+  override lazy val mergeAggregate: GroupByAggregationOnColumn =
+    GroupByAggregation.mergeM2().onColumn(getOrdinal(ref))
+
+  override def toString(): String = "CudfM2"
+  override def dataType: DataType = DoubleType
+  override def nullable: Boolean = true // TODO: Check nullable
```
revans2, line 455: It is false.  Technically for all of these they are not nullable.

https://github.com/apache/spark/blob/adbea252db0f61ed3aaa45ea20b704bad6c47408/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CentralMomentAgg.scala#L60-L62

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -1049,3 +1106,111 @@ case class GpuToCpuCollectBufferTransition(
     projection.apply(InternalRow.apply(arrayData)).getBytes
   }
 }
+
+/**
+ * Base class for overriding standard deviation and variance aggregations.
+ * This class is a GPU-based equivalent of `CentralMomentAgg` aggregation class in Spark.
```
revans2, line 1112: nit: it is not 100% the same because it is only doing the M2 portion of it. It is missing m3 and m4, which is not needed in the aggregations that we support.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
ttnghia: Hah I thought that if `TypeSig` is not double then Spark will throw an exception instead if the input is not double (such as int). Thanks for letting me know this.

revans2: There are different classes in Spark that indicate if type coercion is allowed or not. `ImplicitCastInputTypes` tells Spark that it can cast the inputs to try and match the desired input. There are also several rules about what types can automatically be cast and which cannot `TypeCoersion.scala`.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
ttnghia: Then what happens if the ouput from libcudf is nullable? For example, libcudf M2 outputs null if the input column contains all nulls.

revans2: Then this is a difference between spark and cudf, that we need to understand to make things work properly. 

Spark starts off with 0.0 as the value for everything.

https://github.com/apache/spark/blob/adbea252db0f61ed3aaa45ea20b704bad6c47408/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CentralMomentAgg.scala#L77

And then modifies it when it sees a non-null values.

revans2: So that means we either need to test null outputs from cudf, or replace them with 0.0

abellina: Could use `GpuCoalesce` to accomplish this, like: https://github.com/NVIDIA/spark-rapids/blob/branch-21.10/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala#L754. 

This can happen in postUpdate / postMerge.

docs/configs.md
```diff
@@ -319,6 +319,8 @@ Name | SQL Function(s) | Description | Default Value | Notes
 <a name="sql.expression.Max"></a>spark.rapids.sql.expression.Max|`max`|Max aggregate operator|true|None|
 <a name="sql.expression.Min"></a>spark.rapids.sql.expression.Min|`min`|Min aggregate operator|true|None|
 <a name="sql.expression.PivotFirst"></a>spark.rapids.sql.expression.PivotFirst| |PivotFirst operator|true|None|
+<a name="sql.expression.StddevPop"></a>spark.rapids.sql.expression.StddevPop|`stddev_pop`|Aggregation computing population standard deviation|true|None|
```
ttnghia, line 322: Please ignore all docs files for now.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -1061,3 +1118,149 @@ case class GpuToCpuCollectBufferTransition(
...
+case class GpuStddevSamp(child: Expression) extends GpuM2(child) {
+  override lazy val evaluateExpression: Expression = {
+    // stddev_samp = sqrt(m2 / (n - 1.0)).
+    val stddevSamp =
+      GpuSqrt(GpuDivide(bufferM2, GpuSubtract(bufferN, GpuLiteral(1.0), failOnError = false),
+        failOnErrorOverride = false))
+
+    // Set nulls for the rows where n == 0 and n == 1.
+    GpuIf(GpuEqualTo(bufferN, GpuLiteral(1.0)), GpuLiteral(null, DoubleType),
+      GpuIf(GpuEqualTo(bufferN, GpuLiteral(0.0)), GpuLiteral(null, DoubleType), stddevSamp))
```
ttnghia, line 1233: I just realize that Spark may set `null` or `NaN` for the case when `n === 1` (https://github.com/apache/spark/blob/adbea252db0f61ed3aaa45ea20b704bad6c47408/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CentralMomentAgg.scala#L198). The option is set by:
```
nullOnDivideByZero: Boolean = !SQLConf.get.legacyStatisticalAggregate
```
Should we do exactly the same?

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -1088,3 +1089,44 @@ def do_it(spark):
         df = two_col_df(spark, StringGen('k{1,5}'), ArrayGen(MapGen(StringGen('a{1,5}', nullable=False), StringGen('[ab]{1,5}'))))
         return df.groupBy('a').agg(f.min(df.b[1]["a"]))
     assert_gpu_and_cpu_are_equal_collect(do_it)
+
+
+# Tests for standard deviation and variance aggregations.
```
abellina, line 1094: we should add the Stddev* and Variance* aggs to the ANSI fallback tests here (and need make sure they do fallback).

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -418,6 +418,63 @@ class CudfLastExcludeNulls(ref: Expression) extends CudfFirstLastBase(ref) {
   override val offset: Int = -1
 }
 
+/**
+ * This class is only used by the M2 class aggregates, do not confuse this with GpuAverage.
+ * In the future, this aggregate class should be removed and the mean values should be
+ * generated in the output of libcudf's M2 aggregate.
+ */
+class CudfMean(ref: Expression) extends CudfAggregate(ref) {
+  @transient val rapidsAvgType: DType = GpuColumnVector.getNonNestedRapidsType(dataType)
```
abellina, line 427: seems like this one is not used not sure if I had it there before, but please check if it can be removed.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
revans2: Yes we need to do the same.

However [that change](https://issues.apache.org/jira/browse/SPARK-13860) went into Spark 3.1.0. It was a bug fix so we need to decide if we just put the change in everywhere or if we want it shimmed so we ignore that config in Spark 3.0.x but not in 3.1.0+.  I personally am fine either way.

ttnghia: Let have it here instead of shim for simplicity and time sensitivity. We can always go back and work for a shim.

revans2 (commented):
Overall it is looking good.

abellina (changes requested):
Just some high level comments for now, will try your failing test.

integration_tests/src/main/python/hash_aggregate_test.py
revans2: No we don't because all the only type is supports is double. So it will just work.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
ttnghia: It is used: https://github.com/NVIDIA/spark-rapids/pull/3373/files#diff-7f3472a1577f44fa821a6156d356a32dcb0bdd4edd0873570dba59c31601528cR1150

integration_tests/src/main/python/hash_aggregate_test.py
abellina: > No we don't because all the only type is supports is double. So it will just work.

I stand corrected, everything is doubles with this agg!

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
abellina: No I meant `rapidsAvgType`, I don't see that referenced anywhere.

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -1088,3 +1100,66 @@ def do_it(spark):
...
+        'stddev(b),' +
+        'stddev_pop(b),' +
+        'stddev_samp(b),' +
+        'variance(b),' +
+        'var_pop(b),' +
+        'var_samp(b)' +
+        ' from data_table group by a',
+        conf=conf)
+
+# This test also includes the test above thus let remove the test above when this one can be enabled.
```
abellina, line 1125: normally we have 2 spaces between tests (python style nit)

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -1088,3 +1100,66 @@ def do_it(spark):
...
+@pytest.mark.xfail(reason='GPU-casting directly from decimal to double is not yet supported')
```
abellina, line 1126: please put a link to the issue here, like other xfails https://github.com/NVIDIA/spark-rapids/issues/3610

abellina (changes requested):
couple of nits, will look at the Gpu aggs more closely next.

Also need to upmerge, the signature of `AggExprMeta` changed a bit.

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -1169,3 +1182,69 @@ def do_it(spark):
...
+        'variance(b),' +
+        'var_pop(b),' +
+        'var_samp(b)' +
+        ' from data_table group by a',
+        conf=conf)
+
+
+# xfail because GPU-casting directly from decimal to double is not yet supported.
+# This test also includes the test above thus let remove the test above when this one can be enabled.
+@pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/3610')
```
abellina, line 1210: Bobby's fix went in, can we remove the xfail for this test?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -1061,3 +1116,167 @@ case class GpuToCpuCollectBufferTransition(
...
+abstract class GpuM2(child: Expression, nullOnDivideByZero: Boolean)
+  extends GpuAggregateFunction with ImplicitCastInputTypes with Serializable {
+
+  override def children: Seq[Expression] = Seq(child)
+  override def dataType: DataType = DoubleType
+  override def nullable: Boolean = true
+  override def inputTypes: Seq[AbstractDataType] = Seq(NumericType)
+
+  protected def divideByZeroEvalResult: Expression = {
+    if (nullOnDivideByZero) GpuLiteral(null, DoubleType)
```
abellina, line 1134: nit, suggestion:
```
  protected def divideByZeroEvalResult: Expression = {
    GpuLiteral(if (nullOnDivideByZero) null else Double.NaN, DoubleType)
  }
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -1061,3 +1116,167 @@ case class GpuToCpuCollectBufferTransition(
...
+  override lazy val postUpdate: Seq[Expression] = {
+    val bufferCountAsInt = bufferN.copy(dataType = IntegerType)(bufferN.exprId, bufferN.qualifier)
+    val bufferAvgNoNulls = GpuCoalesce(Seq(bufferAvg, GpuLiteral(0.0, DoubleType)))
+    val bufferM2NoNulls  = GpuCoalesce(Seq(bufferM2, GpuLiteral(0.0, DoubleType)))
+    GpuCast(bufferCountAsInt, DoubleType) :: bufferAvgNoNulls :: bufferM2NoNulls :: Nil
+  }
+
+  // Before merging we have 3 columns and we need to combine them into a structs column.
+  // This is because we are going to do the merging using libcudf's native MERGE_M2 aggregate,
+  // which only accept one column in the input.
```
abellina, line 1176:
```suggestion
  // which only accepts one column in the input.
```

abellina:
The impls for the aggs look correct to me.

integration_tests/src/main/python/hash_aggregate_test.py
ttnghia: I did that and a test failed. Investigating...

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -187,9 +186,10 @@ object AggregateUtils {
  */
 case class BoundCudfAggregate(
     aggExpression: GpuAggregateExpression,
-    boundCudfAggregate: Seq[CudfAggregate])
+    boundCudfAggregate: Seq[CudfAggregate],
```
abellina, line 189: `boundUpdateAggregate` and `boundMergeAggregate` .

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -173,7 +172,7 @@ object AggregateUtils {
  *
  * For example, the `GpuM2` aggregate can have either a `CudfM2` or a `CudfMergeM2`
```
abellina, line 173: should revisit this comment, since `BoundCudfAggregate` change a bit.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -187,9 +186,10 @@ object AggregateUtils {
  */
 case class BoundCudfAggregate(
     aggExpression: GpuAggregateExpression,
-    boundCudfAggregate: Seq[CudfAggregate])
+    boundCudfAggregate: Seq[CudfAggregate],
+    boundMergeAggs: Seq[CudfAggregate])
 
-/** Utility class to hold all of the metrics related to hash aggregation */
+  /** Utility class to hold all of the metrics related to hash aggregation */
```
abellina, line 192:
```suggestion
/** Utility class to hold all of the metrics related to hash aggregation */
```

abellina:
To me this is looking good, am having a hard time running the integration tests locally, that's my hold up.

abellina (approved):
Tests are passing locally. They add about 3 minutes to the tests locally. Thanks @ttnghia!

abellina:
I am going to have this test against databricks.. fyi

ttnghia:
build

ttnghia:
build

abellina:
build

abellina:
@ttnghia I missed some databricks overrides, I added them here: https://github.com/abellina/spark-rapids/commit/99feef7bccdef81243e127137c613a833de19937

ttnghia:
build

ttnghia:
build

ttnghia:
build

ttnghia:
build

ttnghia:
build

shims/spark320/src/main/scala/com/nvidia/spark/rapids/shims/spark320/Spark320Shims.scala
```diff
@@ -1007,4 +1007,9 @@ class Spark320Shims extends Spark32XShims {
     kryo.register(classOf[SerializeBatchDeserializeHostBuffer],
       new KryoJavaSerializer())
   }
+
+  override def getCentralMomentDivideByZeroEvalResult(): Expression = {
+    val nullOnDivideByZero: Boolean = !SQLConf.get.legacyStatisticalAggregate
```
abellina, line 1012: so `nullOnDivideByZero` is coming from the config at execution time, and not from the aggregate function at plan time, which is different than Spark. Theoretically this should be fine, and it tests correctly in a shell against a distributed setup. I am just a little worried and wanted another pair of eyes to confirm (@revans2, @jlowe)

abellina: +1 and thanks for the feedback @revans2.

abellina (approved):
@ttnghia can you file a follow up issue to pull `nullOnDivideByZero` from the aggregate expression, instead of `SQLConfig` at execution time?

ttnghia:
> @ttnghia can you file a follow up issue to pull `nullOnDivideByZero` from the aggregate expression, instead of `SQLConfig` at execution time?

Done.

ttnghia:
@revans2 @andygrove @jlowe Do you have any other comments/suggestions? If not, I'll merge this at the end of today.

revans2 (approved):

abellina:
I ran q17 q39a and q39b at 3000 scale. Good news is that all output matches and all are faster than before. I tested this with simple row counts and then with a join by all keys and diff of non-key columns, followed by a filter for diffs > epsilon (I used 1e-15).

- q17 9.5 seconds => 7.5 seconds, 
- q39a  13.9 seconds => 8 seconds,
- q38b 13 seconds => 6.9 seconds.

q17 seems not interesting at this scale, where stddev is null for all rows with the op on the cpu or gpu. q39a and q39b are interesting and are computing the "coefficient of variation" (`stddev/mean`), which was non null and matched within an epsilon of 1e-15. So I think that's pretty good.

Issue #3376, opened 2021-09-02
[BUG] Unit test failures in Spark 3.2 shim build


gerashegalov:
**Describe the bug**
Umbrella issue to fix unit tests in the 3.2 shim

**Steps/Code to reproduce bug**
```
$ mvn clean package -Dbuildver=320 -Dskip  
```

**Expected behavior**
no unit test failures

**Environment details (please complete the following information)**
local

**Additional context**
#3323 

Tasks:

## HashAggregatesSuite

Fixed by https://github.com/NVIDIA/spark-rapids/pull/3403

- [x]  WITH DECIMALS: Avg Distinct with filter *** FAILED ***

## CastOpSuite

See https://github.com/NVIDIA/spark-rapids/issues/3382 for more detail. 

Fixed by https://github.com/NVIDIA/spark-rapids/pull/3439 and https://github.com/NVIDIA/spark-rapids/pull/3531

- [x]  Cast from string to date using random inputs *** FAILED ***
- [x]  Cast from string to date using random inputs with valid year prefix *** FAILED ***
- [x]  Cast from string to timestamp *** FAILED ***
- [x]  Test all supported casts with in- [ ] range values *** FAILED ***

## ExpandExecSuite

Fixed by https://github.com/NVIDIA/spark-rapids/pull/3403

- [x]  IGNORE ORDER, WITH DECIMALS: group with aggregates *** FAILED ***

## WindowFunctionSuite

Filed as https://github.com/NVIDIA/spark-rapids/issues/3415

Fixed by https://github.com/NVIDIA/spark-rapids/pull/3547

- [x]  WITH DECIMALS: [Window] [RANGE] [ ASC] [- [ ] 2 DAYS, 3 DAYS]  *** FAILED ***
- [x]  WITH DECIMALS: [Window] [RANGE] [DESC] [- [ ] 2 DAYS, 3 DAYS]  *** FAILED ***
- [x]  WITH DECIMALS: [Window] [RANGE] [ ASC] [- [ ] 2 DAYS, CURRENT ROW]  *** FAILED ***
- [x]  WITH DECIMALS: [Window] [RANGE] [DESC] [- [ ] 2 DAYS, CURRENT ROW]  *** FAILED ***
- [x]  WITH DECIMALS: [Window] [RANGE] [ ASC] [CURRENT ROW, 3 DAYS]  *** FAILED ***
[...]

PR #3377, opened 2021-09-03
Fix AnsiCastOpSuite failures with Spark 3.2

andygrove:
This PR updates an assertion in AnsiCastOpSuite and fixes all 21 test failures when running against Spark 3.2

This partially addresses https://github.com/NVIDIA/spark-rapids/issues/3376


shims/spark320/src/main/scala/com/nvidia/spark/rapids/shims/spark320/Spark320Shims.scala
```diff
@@ -805,4 +808,30 @@ class Spark320Shims extends Spark32XShims {
...
+        predicate: SparkPlan => Boolean,
+        accum: ListBuffer[SparkPlan]): Seq[SparkPlan] = {
+      plan match {
+        case _ if predicate(plan) =>
+          accum += plan
+          plan.children.flatMap(p => recurse(p, predicate, accum)).headOption
+        case a: AdaptiveSparkPlanExec => recurse(a.executedPlan, predicate, accum)
+        case qs: BroadcastQueryStageExec => recurse(qs.broadcast, predicate, accum)
+        case qs: ShuffleQueryStageExec => recurse(qs.shuffle, predicate, accum)
+        case c: CommandResultExec => recurse(c.commandPhysicalPlan, predicate, accum)
```
andygrove, line 825: `CommandResultExec` is new in Spark 3.2

gerashegalov: For the sake of keeping Shims small can we hack something like this instead of doing a pattern match:
```scala
    plan.productIterator.foreach { planMember =>
        if (planMember.isInstanceOf[SparkPlan]) {
          recurse(planMember.asInstanceOf[SparkPlan], predicate, accum)
        }
      }
```

revans2: Each of the items called out in the pattern hides part of the rest of the plan in different ways. If it were generic it would not need to be a part of the shim at all.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
```diff
@@ -252,6 +252,11 @@ trait SparkShims {
   def isCustomReaderExec(x: SparkPlan): Boolean
 
   def aqeShuffleReaderExec: ExecRule[_ <: SparkPlan]
+
+  /** Return list of operators in the plan that match the predicate */
+  def findOperators(plan: SparkPlan, predicate: SparkPlan => Boolean): Seq[SparkPlan]
+
+  def shouldAssertIsOnTheGpu(plan: SparkPlan): Boolean
```
revans2, line 259: Can we get a little more explanation about this? It is both obvious what it is doing, and very non-obvious because it is combined with other common patterns.  We could also rename it so it includes extra, or something like that.

sql-plugin/src/main/spark30+all/scala/com/nvidia/spark/rapids/shims/v2/Spark30XShims.scala
```diff
@@ -96,4 +98,25 @@ trait Spark30XShims extends SparkShims {
...
         TypeSig.STRUCT + TypeSig.MAP).nested(), TypeSig.all),
     (exec, conf, p, r) => new GpuCustomShuffleReaderMeta(exec, conf, p, r))
+
+  override def findOperators(plan: SparkPlan, predicate: SparkPlan => Boolean): Seq[SparkPlan] = {
+    def recurse(
+        plan: SparkPlan,
+        predicate: SparkPlan => Boolean,
+        accum: ListBuffer[SparkPlan]): Seq[SparkPlan] = {
+      plan match {
+        case _ if predicate(plan) =>
```
revans2, line 108: I think we have to separate out the predicate matching from the recursion.  If the predicate matches for a `BroadcastQueryStageExec` this will not recurse properly. And this goes for the other implementations too.

shims/spark320/src/main/scala/com/nvidia/spark/rapids/shims/spark320/Spark320Shims.scala
andygrove: Sure, I can update this

andygrove: Sorry, misread the comments here. Ignore my previous comment,

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
andygrove: I pushed some javadocs. I am happy to change the name too but could not find anything I was happy with so far.

revans2 (dismissed):
Looks good to me, but I want to be sure that @gerashegalov is also okay with it because he reviewed this too.

revans2 (approved):

revans2:
build

shims/spark320/src/main/scala/com/nvidia/spark/rapids/shims/spark320/Spark320Shims.scala
```diff
@@ -805,4 +808,29 @@ class Spark320Shims extends Spark32XShims {
...
+        case other => other.children.flatMap(p => recurse(p, predicate, accum)).headOption
+      }
+      accum
+    }
+    recurse(plan, predicate, new ListBuffer[SparkPlan]())
+  }
+
+  override def shouldAssertIsOnTheGpu(plan: SparkPlan): Boolean = plan match {
```
gerashegalov, line 832: I am tempted to suggest trading off type safety for such small auxiliary functions and just do string comparison in the shared code instead of instanceof checks in the future.

gerashegalov (approved):
LGTM, but preferably merge after #3381

andygrove:
I am looking into the test failures

andygrove:
build

andygrove:
build

revans2 (commented):
I don't know the new build system well enough to approve this right now. I see shim changes for spark320 and 301+-nondb, but what about databricks?

andygrove:
This PR does not build on Databricks. I will have a fix up later today.

andygrove:
I pushed a fix and this now builds on Databricks 7.3

andygrove:
build

revans2 (approved):
Just FYI the build has been having some issues because of a mirror of public maven repos.

pxLi:
build

Issue #3232, opened 2021-08-14
[FEA] create the plugin package capable of storing conflicting multiple versions of same named classes


gerashegalov:
**Is your feature request related to a problem? Please describe.**
This feature contributes to #3223. Spark does not provide link compatibility. Therefore identical classes compiled against different spark versions may result in incompatible bytecode. We need to save multiple copies of such classes in the plugin jar. Standard way of using ASM-based shading/relocation [does not work well with Scala](https://contributors.scala-lang.org/t/scala-signature-layout/3327/4). 

**Describe the solution you'd like**

This issue proposes an approach similar to  [ParallelWorldClassLoader](https://github.com/openjdk/jdk/blob/jdk8-b120/jaxws/src/share/jaxws_classes/com/sun/istack/internal/tools/ParallelWorldClassLoader.java). Instead of overriding `findClass` in Spark we can lean on `MutableURLClassLoader` in conjunction with [JarURLConnection JAR URLs](https://docs.oracle.com/javase/7/docs/api/java/net/JarURLConnection.html)

The package will consist of three types of areas
1. a few publicly documented classes in the conventional layout 
2. a large fraction of classes whose bytecode is identical under all supported Spark versions 
3. a smaller fraction of classes that differ under one of the supported Spark vesions

```bash
$ jar tvf rapids-4-spark_2.12.jar
com/nvidia/spark/SQLPlugin.class
spark3xx-common/com/nvidia/spark/rapids/CastExprMeta.class
spark301/org/apache/spark/sql/rapids/GpuUnaryMinus.class    
spark311/org/apache/spark/sql/rapids/GpuUnaryMinus.class
spark320/org/apache/spark/sql/rapids/GpuUnaryMinus.class
```

So each shim can see a consistent parallel world without conflicts by referencing only one conflicting directory. 

E.g., Spark 3.2.0 Shim will use 
```
jar:file:/home/spark/rapids-4-spark_2.12-21.10.jar!/spark3xx-common/
jar:file:/home/spark/rapids-4-spark_2.12-21.10.jar!/spark320/
```

Spark 3.1.1 will use
```
jar:file:/home/spark/rapids-4-spark_2.12-21.10.jar!/spark3xx-common/
jar:file:/home/spark/rapids-4-spark_2.12-21.10.jar!/spark311/
```

**Describe alternatives you've considered**
- shading
- JDK ParallelWorldClassLoader
- replicating copies of the same under different package names.

**Additional context**
[...]

PR #3381, opened 2021-09-03
Implement per-shim parallel world jar classloader

gerashegalov:

Contributes to #3232. Use MutableURLClassLoader in conjunction with JarURLConnection JAR URLs to create "parallel worlds" for each shim in a single jar file.

Assumes a package layout consisting of three types of areas

- a few publicly documented classes in the conventional layout
- a large fraction of classes whose bytecode is identical under all supported Spark versions
- a smaller fraction of classes that differ under one of the supported Spark versions, aka "parallel worlds" in the JDK's com.sun.istack.internal.tools.ParallelWorldClassLoader terminology

```
$ jar tvf rapids-4-spark_2.12.jar
com/nvidia/spark/SQLPlugin.class
spark3xx-common/com/nvidia/spark/rapids/CastExprMeta.class
spark301/org/apache/spark/sql/rapids/GpuUnaryMinus.class    
spark311/org/apache/spark/sql/rapids/GpuUnaryMinus.class
spark320/org/apache/spark/sql/rapids/GpuUnaryMinus.class
```


gerashegalov:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
```diff
@@ -16,53 +16,168 @@
...
+
+  def shimId: String = shimIdFromPackageName(shimProviderClass)
+
+  def getRapidsShuffleManagerClass: String = {
+    findShimProvider()
+    s"com.nvidia.spark.rapids.$shimId.RapidsShuffleManager"
+  }
+
+  def getRapidsShuffleInternal: String = {
+    findShimProvider()
```
tgravescs, line 85: a bit odd we don't get return value , should at least comment why its called here

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
```diff
@@ -16,53 +16,168 @@
...
+  @volatile private var shimURL: URL = _
+  @volatile private var pluginClassLoader: ClassLoader = _
+
+  // REPL-only logic
+  @volatile private var tmpClassLoader: MutableURLClassLoader = _
+
+  def shimId: String = shimIdFromPackageName(shimProviderClass)
+
+  def getRapidsShuffleManagerClass: String = {
+    findShimProvider()
```
tgravescs, line 80: same here comment why called since not getting output and not using anything from It that is obvious

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
```diff
@@ -16,53 +16,168 @@
...
+      .map(scala.io.Source.fromURL)
+      .flatMap(_.getLines())
+
+    val shimServiceProviderOpt = serviceProviderList.flatMap { shimServiceProviderStr =>
+      val mask = shimIdFromPackageName(shimServiceProviderStr)
+      try {
+        val shimURL = new java.net.URL(s"${shimRootURL.toString}$mask/")
+        val shimClassLoader = new MutableURLClassLoader(Array(shimURL, shimCommonURL),
+          getClass.getClassLoader)
+        // can't use ServiceLoader with parallel world layout
```
tgravescs, line 139: what does this mean?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
```diff
@@ -16,53 +16,168 @@
...
+  private def shimIdFromPackageName(shimServiceProvider: SparkShimServiceProvider) = {
+    shimServiceProvider.getClass.getPackage.toString.split('.').last
+  }
+
+  private def shimIdFromPackageName(shimServiceProviderStr: String) = {
+    shimServiceProviderStr.split('.').takeRight(2).head
+  }
+
+  private def findShimProvider(): String = {
+    // TODO restore support for shim provider override
```
tgravescs, line 171: just make sure we do this because Cloudera uses regularly

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
```diff
@@ -76,7 +191,39 @@ object ShimLoader extends Logging {
...
+    shuffleClass.getConstructor(classOf[SparkConf], java.lang.Boolean.TYPE)
+        .newInstance(conf, java.lang.Boolean.valueOf(isDriver))
+        .asInstanceOf[VisibleShuffleManager]
+  }
+
+  // avoid cached constructors
+  private def instantiateClass[T](cls: Class[T]): T = {
+    logDebug(s"Instantiate ${cls.getName} using classloader " + cls.getClassLoader)
+    cls.getClassLoader match {
+      case m: MutableURLClassLoader => logDebug("GERA_DEBUG urls " + m.getURLs.mkString("\n"))
```
tgravescs, line 223: remove GERA

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -16,20 +16,20 @@
...
 
-import java.util
 import java.util.Properties
 import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference}
 
 import scala.collection.JavaConverters._
 import scala.util.Try
 
 import com.nvidia.spark.rapids.python.PythonWorkerSemaphore
+import java.util
```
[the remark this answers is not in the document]
gerashegalov: I am getting rid of it. IDEA import optimization is always confused by this

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
gerashegalov: Even if I pass the modified classloader to the  ServiceLoader.load invocation it expects META-INF/services (which is ok) and provider classes in the normal root / entry (not ok)

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
gerashegalov: need to double-check if we still need it to begin with

gerashegalov:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
gerashegalov: is there a typical classname they use here?

gerashegalov:
build

gerashegalov:
build

sql-plugin/src/main/scala/com/nvidia/spark/SQLPlugin.scala
```diff
@@ -26,6 +26,9 @@ import org.apache.spark.internal.Logging
  * To enable this plugin, set the config "spark.plugins" to `com.nvidia.spark.SQLPlugin`
  */
 class SQLPlugin extends SparkPlugin with Logging {
-  override def driverPlugin(): DriverPlugin = new RapidsDriverPlugin
-  override def executorPlugin(): ExecutorPlugin = new RapidsExecutorPlugin
+  override def driverPlugin(): DriverPlugin =
+    ShimLoader.newInstanceOf("com.nvidia.spark.rapids.RapidsDriverPlugin")
```
revans2, line 30: I am not thrilled with a reflection like API when the number of classes is small and we could have a dedicated API for each. Is there a reason this needs to be reflection based?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -151,13 +143,15 @@ class RapidsDriverPlugin extends DriverPlugin with Logging {
...
   }
 
-  override def init(sc: SparkContext, pluginContext: PluginContext): util.Map[String, String] = {
+  override def init(
+    sc: SparkContext, pluginContext: PluginContext): java.util.Map[String, String] = {
     val sparkConf = pluginContext.conf
     RapidsPluginUtils.fixupConfigs(sparkConf)
     val conf = new RapidsConf(sparkConf)
-    if (conf.shimsProviderOverride.isDefined) {
+    if (conf.shimsProviderOverride.isDefined) { // TODO test it, probably not working yet
```
revans2, line 151: Have you tested this yet?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SQLExecPlugin.scala
```diff
@@ -0,0 +1,40 @@
...
+ * Extension point to enable GPU SQL processing.
+ */
+class SQLExecPlugin extends (SparkSessionExtensions => Unit) with Logging {
+  override def apply(extensions: SparkSessionExtensions): Unit = {
+    extensions.injectColumnar(columnarOverrides)
+    extensions.injectQueryStagePrepRule(queryStagePrepOverrides)
+  }
+
+  private def columnarOverrides(sparkSession: SparkSession): ColumnarRule = {
+    ShimLoader.newInstanceOf("com.nvidia.spark.rapids.ColumnarOverrideRules")
```
revans2, line 34: Again not thrilled by a reflection like API. I am fine with it to get things working, but would prefer to have something more type safe/etc.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
```diff
@@ -16,53 +16,181 @@
...
+  @volatile private var pluginClassLoader: ClassLoader = _
+
+  // REPL-only logic
+  @volatile private var tmpClassLoader: MutableURLClassLoader = _
+
+  def shimId: String = shimIdFromPackageName(shimProviderClass)
+
+  def getRapidsShuffleManagerClass: String = {
+    findShimProvider()
+    s"com.nvidia.spark.rapids.$shimId.RapidsShuffleManager"
```
revans2, line 85: If we control the package names, why are we making this more complicated for ourselves and having shim versions in the package names? Is this just for the transition and will go away once we fully change over? If so could you add comments like you did in the pom.xml to that effect?  If not why not?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
```diff
@@ -76,7 +204,40 @@ object ShimLoader extends Logging {
...
+  def newInstanceOf[T](className: String): T = {
+    val loader = getShimClassLoader()
+    logDebug(s"Loading $className using $loader with the parent loader ${loader.getParent}")
+    instantiateClass(loader.loadClass(className)).asInstanceOf[T]
+  }
+
+  def newInternalShuffleManager(conf: SparkConf, isDriver: Boolean): VisibleShuffleManager = {
+    val shuffleClassLoader = getShimClassLoader()
+    val shuffleClassName =
+      s"org.apache.spark.sql.rapids.shims.${shimId}.RapidsShuffleInternalManager"
```
revans2, line 225: Why not use getRapidsShuffleInternal

sql-plugin/src/main/scala/com/nvidia/spark/SQLPlugin.scala
gerashegalov: The purpose is to prevent the caller's classloader being used at the call site. The first entry point in the code we can control must be reflection-based unless and until we can push the classloader changes upstream to Apache Spark. Otherwise, we will end up with even more reflection calls.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
gerashegalov: not yet

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SQLExecPlugin.scala
gerashegalov: it's a necessary "evil"

sql-plugin/src/main/scala/com/nvidia/spark/SQLPlugin.scala
revans2: Then can we try to limit all of the reflection to be self contained in ShimLoader?  having newInstanceOf be private forces people to put the reflection code in the ShimLoader and helps us keep it all in once place.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
gerashegalov: +1 on adding more comments. 
The gist: 
1. we are preserving the current naming scheme. 
2. we can't use the same shuffle manager class name like we can make for other classes because it's a spark setting processed before the plugin is loaded whereas we can't create a static class satisfying the superset of the API across Spark versions. It's explained in the design doc. Unless in the future version of Spark we can have a configurable that defers ShuffleManager loading to the driver/executor plugin which we do under the covers in this PR

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
gerashegalov: good catch

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
tgravescs: still needed?

tgravescs:
I'm assuming you haven't tested on Databricks unless I'm out of date, I pulled this branch into my branch with dist packaging changes and it fails.  I have fixed it on my branch if we want to wait but we should probably put this and that in close together.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
gerashegalov: yes, commenting

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
gerashegalov: actually hard to test before I can pull @tgravescs's change

gerashegalov:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
tgravescs: --conf "spark.rapids.shims-provider-override=com.nvidia.spark.rapids.shims.spark311cdh.SparkShimServiceProvider"

gerashegalov: shimId being part of the package name even for an override is what I hoped for. If we start *requiring* this restriction going forward. it will be easy to accommodate after #3411

pom.xml
```diff
@@ -243,7 +251,8 @@
                                 <phase>generate-sources</phase>
                                 <configuration>
                                     <sources>
-                                        <source>${project.basedir}/src/main/spark30+all/scala</source>
+                                        <source>${project.basedir}/src/main/spark301+311-all/scala</source>
```
tgravescs, line 254: would it be easier for others to understand just to use words?  301until311 (exclusive)  301thru312 (inclusive)

tgravescs: note not a blocker, just throwing the idea out there, we can always rename later.

tgravescs (approved):
overall looks fine for now, the overrides we will need to fix once https://github.com/NVIDIA/spark-rapids/pull/3411 is merged

pom.xml
gerashegalov: I like the idea, it looks more readable to me except I lean more towards `301to312` instead of informal `thru`  for brevity, and Scala range syntax analogy. Depending if there will be more blocking comments I'll incorporate the suggestion.

abellina:
I tried this code with spark 3.1.1. I was able to get a UCX shuffle with the changes, so we know it works. Early start worked as well, and executors registered when I expected it. However, I noticed a late print of the following message:

```
2021-09-08 21:37:29,900 WARN  RapidsShuffleInternalManager:69 - Rapids Shuffle Plugin enabled. Transport enabled (remote fetches will use com.nvidia.spark.rapids.shuffle.ucx.UCXShuffleTransport. To disable the RAPIDS Shuffle Manager set `spark.rapids.shuffle.enabled` to false
```

Which happened when I needed a shuffle (it seemed). This was on the shell, I had a dataframe that would shuffle, and I asked for `collect`, so the driver is not initializing this early as it used to. Normally this happens at initialization, near the "here's how you turn off the SQL plugin" piece. In the big picture, it seems like a minor thing that can be corrected later, I don't think it is a show stopper for this PR, since the query still late-initialized the shuffle manager, and produced a plan with the correct shuffle handles for the UCX shuffle to work.

gerashegalov:
build

pom.xml
```diff
@@ -448,9 +444,9 @@
                                 <phase>generate-sources</phase>
                                 <configuration>
                                     <sources>
-                                        <source>${project.basedir}/src/main/spark30+all/scala</source>
-                                        <source>${project.basedir}/src/main/spark31+all/scala</source>
-                                        <source>${project.basedir}/src/main/spark31+apache/scala</source>
+                                        <source>${project.basedir}/src/main/301until320-all/scala</source>
+                                        <source>${project.basedir}/src/main/311+-all/scala</source>
+                                        <source>${project.basedir}/src/main/311+-apache/scala</source>
```
tgravescs, line 449: I assume 311+- is includes for 311 thru all?

shims/spark304/src/main/scala/com/nvidia/spark/rapids/shims/spark304/RapidsShuffleInternalManager.scala
```diff
@@ -17,8 +17,9 @@
 package org.apache.spark.sql.rapids.shims.spark304
 
 import org.apache.spark.{SparkConf, TaskContext}
+
```
tgravescs, line 20: style issue

pom.xml
gerashegalov: correct, adding more explicit wording to the doc

gerashegalov:
build

tgravescs (dismissed):

shims/spark304/src/main/scala/com/nvidia/spark/rapids/shims/spark304/RapidsShuffleInternalManager.scala
tgravescs: this still looks like issue?

tgravescs: oh this is because https://github.com/NVIDIA/spark-rapids/issues/3388# 
I actually fixed these in https://github.com/NVIDIA/spark-rapids/pull/3411 so I guess its ok here

gerashegalov:
build

tgravescs (approved):

PR #3387, opened 2021-09-03
Fallback to the CPU for IGNORE NULLS on lead and lag

### LINKED ISSUE #1923: [FEA] Fall back to the CPU when LEAD/LAG wants to IGNORE NULLS
Support `[ IGNORE NULLS | RESPECT NULLS ]` for `LEAD`/`LAG`  was added here: https://github.com/apache/spark/commit/3e9821edfd . Need to evaluate and add tests.

revans2:
This fixes #1923


revans2:
build

revans2:
It should be noted that RESPECT NULLS is the default and we already support that

integration_tests/src/main/python/window_function_test.py
```diff
@@ -609,6 +610,34 @@ def do_it(spark):
...
+
+    assert_gpu_sql_fallback_collect(
+        lambda spark: gen_df(spark, data_gen),
+        'Lead',
+        "window_agg_table",
+        '''
+        SELECT
+            LEAD(d, 5) IGNORE NULLS OVER (PARTITION by a ORDER BY b,c) lead_d_5,
+            LEAD(d, 2, d_default) IGNORE NULLS OVER (PARTITION by a ORDER BY b,c) lead_d_2_default,
+            LAG(d, 5) IGNORE NULLS OVER (PARTITION by a ORDER BY b,c) lag_d_5,
```
jlowe, line 636: Should there be separate queriess for LEAD vs. LAG?  By putting both expressions in the same query with the same partitioning and ordering, don't we risk the test passing just because one of the expressions gets the logic right and kicks the operation off the GPU but the other expression tagging logic could still be incorrect?

revans2: Yup I was being lazy...

revans2:
build

integration_tests/src/main/python/window_function_test.py
```diff
@@ -631,10 +631,32 @@ def test_window_aggs_lead_lag_ignore_nulls_fallback(a_gen, b_gen, c_gen, d_gen):
...
+@pytest.mark.parametrize('c_gen', [LongRangeGen()], ids=meta_idfn('orderBy:'))
+@pytest.mark.parametrize('b_gen', [long_gen], ids=meta_idfn('orderBy:'))
+@pytest.mark.parametrize('a_gen', [long_gen], ids=meta_idfn('partBy:'))
+def test_window_aggs_lag_ignore_nulls_fallback(a_gen, b_gen, c_gen, d_gen):
+    data_gen = [
+            ('a', RepeatSeqGen(a_gen, length=20)),
+            ('b', b_gen),
+            ('c', c_gen),
+            ('d', d_gen),
+            ('d_default', d_gen)]
```
jlowe, line 651: Now seems to be unused? Also applies to the copy above.

jlowe (approved):
Small nit about some apparently unused data being generated but otherwise lgtm.

jlowe (approved):

jlowe:
build

Issue #3379, opened 2021-09-03
[BUG] q93 failed w/ UCX


pxLi:
**Describe the bug**
benchmark q93 failed w/ UCX on egx machine. Remove `--conf spark.shuffle.manager` the test works fine

```
16:03:51  21/09/03 08:06:51 INFO MapOutputTrackerMasterEndpoint: Asked to send map output locations for shuffle 2 to 10.136.6.4:48304
16:03:51  21/09/03 08:06:51 INFO MapOutputTrackerMasterEndpoint: Asked to send map output locations for shuffle 2 to 10.136.6.4:48308
16:03:51  21/09/03 08:06:51 INFO MapOutputTrackerMasterEndpoint: Asked to send map output locations for shuffle 2 to 10.136.6.4:48302
16:03:51  21/09/03 08:06:51 INFO MapOutputTrackerMasterEndpoint: Asked to send map output locations for shuffle 2 to 10.136.6.4:48306
16:03:51  21/09/03 08:06:51 INFO TaskSetManager: Starting task 16.0 in stage 33.0 (TID 306, 10.136.6.4, executor 0, partition 16, NODE_LOCAL, 7318 bytes)
16:03:51  21/09/03 08:06:51 WARN TaskSetManager: Lost task 11.0 in stage 33.0 (TID 301, 10.136.6.4, executor 0): scala.MatchError: com.nvidia.spark.rapids.GpuPackedTableColumn@33daca7c (of class com.nvidia.spark.rapids.GpuPackedTableColumn)
16:03:51  	at com.nvidia.spark.rapids.GpuColumnarBatchSerializerInstance$$anon$1.$anonfun$writeValue$2(GpuColumnarBatchSerializer.scala:82)
16:03:51  	at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:158)
16:03:51  	at com.nvidia.spark.rapids.GpuColumnarBatchSerializerInstance$$anon$1.writeValue(GpuColumnarBatchSerializer.scala:81)
16:03:51  	at org.apache.spark.storage.DiskBlockObjectWriter.write(DiskBlockObjectWriter.scala:249)
16:03:51  	at org.apache.spark.shuffle.sort.BypassMergeSortShuffleWriter.write(BypassMergeSortShuffleWriter.java:158)
16:03:51  	at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59)
[...]

PR #3323, opened 2021-08-27
Enable optional Spark 3.2.0 shim build

gerashegalov:
Signed-off-by: Gera Shegalov <gera@apache.org>

This PR contributes to #3223 
- enables spark 3.2.0 build via `-Dbuildver=320`
- prepares deduplication of identical classes once all shims follow spark 3.2.0 pattern
- sets stage for the parallel world classloader   

Issues to file:

Compilation:
```
ShuffledBatchRDD.scala:141: match may not be exhaustive.
It would fail on the following input: CoalescedMapperPartitionSpec(_, _, _)
[WARNING] [Warn] /home/gshegalov/gits/NVIDIA/spark-rapids/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/ShuffledBatchRDD.scala:165: match may not be exhaustive.
It would fail on the following input: CoalescedMapperPartitionSpec(_, _, _)
[WARNING] [Warn] GpuMultiFileReader.scala:87: method getAllStatistics in class FileSystem is deprecated: see corresponding Javadoc for more information.

[WARNING] [Warn] sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala:604: method getDecimalMetadata in class PrimitiveType is deprecated: see corresponding Javadoc for more information.

[WARNING] [Warn] sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/rapids/GpuPartitioningUtils.scala:121: method isDirectory in class FileSystem is deprecated: see corresponding Javadoc for more information.
```

```
Replace  semanticEqauls with SPARK-35742 canonicalize
```

```
// https://issues.apache.org/jira/browse/SPARK-33890
  protected def doEval(
      srcString: org.apache.spark.unsafe.types.UTF8String,
      trimString: org.apache.spark.unsafe.types.UTF8String
  ): org.apache.spark.unsafe.types.UTF8String = {
    throw new UnsupportedOperationException("TODO: Columnar only message!")
  }

  protected def doEval(
      srcString: org.apache.spark.unsafe.types.UTF8String
  ): org.apache.spark.unsafe.types.UTF8String = {
    throw new UnsupportedOperationException("TODO: Columnar only message!")
  }
```

failing unit tests:
``` 
~/gits/NVIDIA/spark-rapids$ grep -a '*** FAILED' /tmp/build320.log 
- WITH DECIMALS: Avg Distinct with filter *** FAILED ***
- Cast from string to date using random inputs *** FAILED ***
[...]

PR #3389, opened 2021-09-03
Bring back setRapidsShuffleManager in the driver side

abellina:

This should fix the failure reported here: #3379. 

This was an inadvertent change introduced in https://github.com/NVIDIA/spark-rapids/pull/3323, that effectively breaks the UCX shuffle manager in HEAD.


abellina:
build

tgravescs (approved):

PR #3390, opened 2021-09-03
Remove shim-specific copies of ParquetCachedBatchSerializer

### LINKED ISSUE #3314: [FEA] Create non-shim specific version of ParquetCachedBatchSerializer
Today for user to specify ParquetCachedBatchSerializer, they have to specify the shim version:
--conf spark.sql.cache.serializer=com.nvidia.spark.rapids.shims.spark311.ParquetCachedBatchSerializer

Ideally we don't want user to have to say spark version and they just specify something like com.nvidia.spark.rapids.ParquetCachedBatchSerializer and we load the proper version.

Note with the 3.2 build work going on, we have created versions for each shim temporarily as well.

razajafri:
This PR removes the shim-specific copies of the ParquetCachedBatchSerializer. 

After building with the default profile. This was tested with the v2 version of the serializer against Spark-3.1.1 and Spark-3.1.2 for running test_cache.py and they both were successful. 

fixes #3314


tgravescs:
it looks like the no-buildver-default profile from the pom.xml is missing from the spark313 shim (missed in https://github.com/NVIDIA/spark-rapids/pull/3323/)

razajafri:
build

gerashegalov:
This should wait on #3381 and dist jar change

gerashegalov:
@tgravescs is working on #3411 which might be difficult to upmerge, so I think this PR still should wait.

tgravescs:
taking this over, going to upmerge and then re-enable tests if everything looks good.

PR #3392, opened 2021-09-06
Support transforming BinaryType between Row and Columnar

### LINKED ISSUE #3367: [FEA] Support GpuRowToColumnConverter on BinaryType
Currently, `GpuRowToColumnarExec` doesn't support BinaryType because there is no `GpuRowToColumnConverter` for  BinaryType. Meanwhile, `GpuRowToColumnarExec` on BinaryType occurs in certain situations, such as:
```
      +- SortAggregate(key=[a#354L, c#356L], functions=[merge_collect_list(b#355, 0, 0), merge_collect_set(b#355, 0, 0)], output=[a#354L, c#356L, buf#374, buf#376])
         +- GpuColumnarToRow false
            +- GpuSort [a#354L ASC NULLS FIRST, c#356L ASC NULLS FIRST], false, com.nvidia.spark.rapids.OutOfCoreSort$@27ad8c1a
               +- GpuShuffleCoalesce 2147483647
                  +- GpuColumnarExchange gpuhashpartitioning(a#354L, c#356L, 12), ENSURE_REQUIREMENTS, [id=#344]
                        +- SortAggregate(key=[a#354L, c#356L], functions=[partial_collect_list(b#355, 0, 0), partial_collect_set(b#355, 0, 0)], output=[a#354L, c#356L, buf#374, buf#376])
```
When setting `spark.sql.execution.useObjectHashAggregateExec` True,  the `GpuExchange` here will not fallback to CPU because additional GpuExec (`GpuSort`) is inserted on the top of it.  Meanwhile,  buffer converters for TypedImperativeAgg will not be inserted because both Merge and Update stage are on CPU.  Therefore, the `GpuRowToColumnarExec` here is trying to convert row data of BinaryType to GPU.

sperlingxx:

Closes #3367

Current PR is to enable GpuRowToColumnar and GpuColumnarToRow with BinaryType. During transforming,  BinaryType will be mapped to DType.LIST (with DType.INT8 as child type).

### DIFF
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
@@ -251,6 +251,8 @@ public class GpuColumnVector extends GpuColumnVectorBase {
         children[i] = convertFrom(fields[i].dataType(), fields[i].nullable());
       }
       return new HostColumnVector.StructType(nullable, children);
+    } else if (spark instanceof BinaryType) {
+      return new HostColumnVector.ListType(nullable, convertFrom(DataTypes.ByteType, false));
     } else {
       // Only works for basic types
       return new HostColumnVector.BasicType(nullable, getNonNestedRapidsType(spark));
@@ -472,8 +474,10 @@ public class GpuColumnVector extends GpuColumnVectorBase {
       return DType.TIMESTAMP_DAYS;
     } else if (type instanceof TimestampType) {
       return DType.TIMESTAMP_MICROSECONDS;
-    } else if (type instanceof StringType || type instanceof BinaryType) {
+    } else if (type instanceof StringType) {
       return DType.STRING;
+    } else if (type instanceof BinaryType) {
+      return DType.LIST;
     } else if (type instanceof NullType) {
       // INT8 is used for both in this case
       return DType.INT8;
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
@@ -472,7 +472,7 @@ public class GpuColumnVector extends GpuColumnVectorBase {
       return DType.TIMESTAMP_DAYS;
     } else if (type instanceof TimestampType) {
       return DType.TIMESTAMP_MICROSECONDS;
-    } else if (type instanceof StringType) {
+    } else if (type instanceof StringType || type instanceof BinaryType) {
       return DType.STRING;
     } else if (type instanceof NullType) {
       // INT8 is used for both in this case
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
@@ -179,7 +179,7 @@ public class RapidsHostColumnVectorCore extends ColumnVector {
 
   @Override
   public final byte[] getBinary(int rowId) {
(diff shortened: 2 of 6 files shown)


sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
```diff
@@ -472,7 +472,7 @@ private static DType toRapidsOrNull(DataType type) {
       return DType.TIMESTAMP_DAYS;
     } else if (type instanceof TimestampType) {
       return DType.TIMESTAMP_MICROSECONDS;
-    } else if (type instanceof StringType) {
+    } else if (type instanceof StringType || type instanceof BinaryType) {
```
revans2, line 475: We have a decision to make. Is BinaryType the same as a DType.STRING, or is it the same as a DType.LIST of DType.INT8 where the bytes are not nullable?  Internally the data layout will be identical.  But to me Binary can store things that are technically illegal to be stored in a String.  String type data is encoded as UTF8.  Some characters require a multi-byte encoding and it is technically an error to only store the first bytes of a multi-byte encoded string.  I personally wold rather store it as a LIST of INT8.  But I am open to arguments as to why it is better to store it as a STRING instead.

ttnghia: `DType.LIST` of `DType.INT8` (CHAR if available?) makes more sense to me, as in C++ people frequently use `vector<char>` to store binary.

sperlingxx:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
sperlingxx: I agree with @ttnghia. And I changed the underlying DType of BinaryType to LIST.

sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
```diff
@@ -22,6 +22,7 @@
 
 import org.apache.spark.sql.types.ArrayType;
 import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
```
ttnghia, line 25: Are they related?

sql-plugin/src/main/java/com/nvidia/spark/rapids/GpuColumnVector.java
```diff
@@ -474,6 +476,8 @@ private static DType toRapidsOrNull(DataType type) {
       return DType.TIMESTAMP_MICROSECONDS;
     } else if (type instanceof StringType) {
       return DType.STRING;
+    } else if (type instanceof BinaryType) {
+      return DType.LIST;
```
revans2, line 480: nested types for this should return null.

revans2 (approved):
Just one nit.

sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
sperlingxx: `org.apache.spark.sql.types.DataType` refers to the abstract Scala class: [DataType](https://github.com/apache/spark/blob/HEAD/sql/catalyst/src/main/scala/org/apache/spark/sql/types/DataType.scala), who is the parent of all specific data types.
`org.apache.spark.sql.types.DataTypes` refers to a Java class: [DataTypes](https://github.com/apache/spark/blob/HEAD/sql/catalyst/src/main/java/org/apache/spark/sql/types/DataTypes.java), who holds the singletons representing all data types.

Issue #3058, opened 2021-07-28
[BUG] GPU ORC reader complains errors when specifying columns that do not exist in file schema.


firestarman:
**Describe the bug**
Spark ORC reader produces nulls for the columns that do not exist in file schema, but GPU ORC reader fails to read.
GPU:

```
scala> val rs = StructType(Seq(StructField("_c0",StructType(Seq(StructField("child0",ByteType,true), StructField("no_child",ByteType,true)))), StructField("no_int", IntegerType)))
scala> spark.read.schema(rs).orc("/data/tmp/test.log").show

21/07/28 08:28:06 ERROR Executor: Exception in task 0.0 in stage 0.0 (TID 0)
org.apache.spark.sql.execution.QueryExecutionException: Incompatible schemas for ORC file at file:///data/tmp/test.log
 file schema: struct<_c0:struct<child0:tinyint,child1:int,child2:bigint,child3:double,child4:boolean,child5:date,child6:timestamp>>
 read schema: struct<_c0:struct<child0:tinyint,no_child:tinyint>,no_int:int>
	at com.nvidia.spark.rapids.GpuOrcFileFilterHandler$GpuOrcPartitionReaderUtils.$anonfun$checkSchemaCompatibility$1(GpuOrcScan.scala:1099)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
	at com.nvidia.spark.rapids.GpuOrcFileFilterHandler$GpuOrcPartitionReaderUtils.checkSchemaCompatibility(GpuOrcScan.scala:1092)
	at com.nvidia.spark.rapids.GpuOrcFileFilterHandler$GpuOrcPartitionReaderUtils.getOrcPartitionReaderContext(GpuOrcScan.scala:894)
	at com.nvidia.spark.rapids.GpuOrcFileFilterHandler.$anonfun$filterStripes$3(GpuOrcScan.scala:760)
```

CPU:
```
scala> spark.read.schema(rs).orc("/data/tmp/test.log").show
+------------+------+
|         _c0|no_int|
+------------+------+
|        null|  null|
| {-21, null}|  null|
| {112, null}|  null|
| {-50, null}|  null|
| {-49, null}|  null|
| {-55, null}|  null|
|{null, null}|  null|
[...]

PR #3393, opened 2021-09-06
Fill with null columns for the names exist only in read schema in ORC reader

firestarman:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3058

### DIFF
--- a/integration_tests/src/main/python/orc_test.py
+++ b/integration_tests/src/main/python/orc_test.py
@@ -88,6 +88,8 @@ orc_gens_list = [orc_basic_gens,
     pytest.param([date_gen], marks=pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/131')),
     pytest.param([timestamp_gen], marks=pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/131'))]
 
+flattened_orc_gens = orc_basic_gens + orc_array_gens_sample + orc_struct_gens_sample
+
 @allow_non_gpu('FileSourceScanExec')
 @pytest.mark.parametrize('read_func', [read_orc_df, read_orc_sql])
 @pytest.mark.parametrize('disable_conf', ['spark.rapids.sql.format.orc.enabled', 'spark.rapids.sql.format.orc.read.enabled'])
@@ -377,3 +379,26 @@ def test_read_struct_without_stream(spark_tmp_path):
             lambda spark : unary_op_df(spark, data_gen, 10).write.orc(data_path))
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : spark.read.orc(data_path))
+
+
+@pytest.mark.parametrize('orc_gen', flattened_orc_gens, ids=idfn)
+@pytest.mark.parametrize('reader_confs', reader_opt_confs, ids=idfn)
+@pytest.mark.parametrize('v1_enabled_list', ["", "orc"])
+def test_read_with_more_columns(spark_tmp_path, orc_gen, reader_confs, v1_enabled_list):
+    struct_gen = StructGen([('nested_col', orc_gen)])
+    # Map is not supported yet.
+    gen_list = [("top_pri", orc_gen), ("top_st", struct_gen),
+                ("top_ar", ArrayGen(struct_gen, max_length=10))]
+    data_path = spark_tmp_path + '/ORC_DATA'
+    with_cpu_session(
+            lambda spark : gen_df(spark, gen_list).write.orc(data_path))
+    all_confs = reader_confs.copy()
+    all_confs.update({'spark.sql.sources.useV1SourceList': v1_enabled_list})
+    # This is a hack to get the type in a slightly less verbose way
+    extra_struct_gen = StructGen([('nested_col', orc_gen), ("extra_nested_col", orc_gen)])
+    extra_gen_list = [("top_pri", orc_gen), ("top_ar", ArrayGen(extra_struct_gen, max_length=10)),
+        ("top_st", extra_struct_gen), ("extra_top", orc_gen)]
+    rs = StructGen(extra_gen_list, nullable=False).data_type
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : spark.read.schema(rs).orc(data_path),
+            conf=all_confs)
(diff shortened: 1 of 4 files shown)


firestarman:
build

pxLi:
build

firestarman:
build

integration_tests/src/main/python/orc_test.py
```diff
@@ -377,3 +379,26 @@ def test_read_struct_without_stream(spark_tmp_path):
...
+    with_cpu_session(
+            lambda spark : gen_df(spark, gen_list).write.orc(data_path))
+    all_confs = reader_confs.copy()
+    all_confs.update({'spark.sql.sources.useV1SourceList': v1_enabled_list})
+    # This is a hack to get the type in a slightly less verbose way
+    extra_struct_gen = StructGen([('nested_col', orc_gen), ("extra_nested_col", orc_gen)])
+    extra_gen_list = [("top_pri", orc_gen), ("top_ar", ArrayGen(extra_struct_gen, max_length=10)),
+        ("top_st", extra_struct_gen), ("extra_top", orc_gen)]
+    rs = StructGen(extra_gen_list, nullable=False).data_type
+    assert_gpu_and_cpu_are_equal_collect(
```
wbo4958, line 402: what about the case that the reader schema is `top_pri,  extra_top,  top_ar, top_st`

wbo4958: and  `top_st.extra_nested_col, top_st.nested_col` ?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala
```diff
@@ -1474,17 +1438,11 @@ class MultiFileCloudOrcPartitionReader(
         val batchSizeBytes = GpuColumnVector.getTotalDeviceMemoryUsed(table)
         logDebug(s"GPU batch size: $batchSizeBytes bytes")
         maxDeviceMemory = max(batchSizeBytes, maxDeviceMemory)
-        val numColumns = table.getNumberOfColumns
-        if (readDataSchema.length != numColumns) {
-          throw new QueryExecutionException(s"Expected ${readDataSchema.length} columns " +
```
wbo4958, line 1479: Looks like the similar exception is not thrown in the  evolveSchemaIfNeededAndClose in the production version.

firestarman: Yes, it is removed intentionally, because it is no longer correct after supporting the feature mentioned in issue #3058.
The final output batch consists of two types of columns. Some are read from ORC files, and the other (columns filled with nulls) are generated by our plugin, to align with the behavior of Spark ORC reader.

integration_tests/src/main/python/orc_test.py
firestarman: Yes, I think so.

firestarman: The output will always follow the column order in the `readSchema`.

res-life: NIT: I think replacing extra_top with non_existing is more meaningful. Anyway, Looks good.

integration_tests/src/main/python/orc_test.py
```diff
@@ -377,3 +379,26 @@ def test_read_struct_without_stream(spark_tmp_path):
             lambda spark : unary_op_df(spark, data_gen, 10).write.orc(data_path))
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : spark.read.orc(data_path))
+
+
+@pytest.mark.parametrize('orc_gen', flattened_orc_gens, ids=idfn)
+@pytest.mark.parametrize('reader_confs', reader_opt_confs, ids=idfn)
+@pytest.mark.parametrize('v1_enabled_list', ["", "orc"])
+def test_read_with_more_columns(spark_tmp_path, orc_gen, reader_confs, v1_enabled_list):
```
revans2, line 387: Spark has the config `spark.sql.caseSensitive` which is used by ORC to decide if columns can be matched up in case sensitive or insensitive way. The default is case insensitive.  Could we have some tests to verify that things like `TOP_AR` and `top_ar` match by default?  Could we also violate the request of Spark and set this to be case sensitive and verify that they do not match as expected?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SchemaUtils.scala
```diff
@@ -0,0 +1,232 @@
...
+    case ArrayType(et, nullable) =>
+      ArrayType(replaceCharVarcharWithString(et), nullable)
+    case MapType(kt, vt, nullable) =>
+      MapType(replaceCharVarcharWithString(kt), replaceCharVarcharWithString(vt), nullable)
+    case StructType(fields) =>
+      StructType(fields.map { field =>
+        field.copy(dataType = replaceCharVarcharWithString(field.dataType))
+      })
+    case _: CharType => StringType
+    case _: VarcharType => StringType
```
revans2, line 46: VarcharType was introduced by https://issues.apache.org/jira/browse/SPARK-33480 in Spark 3.1.0.  This code is going to need to be updated either so parts of it are under the shim layer, or we don't deal with VarcharType.

integration_tests/src/main/python/orc_test.py
firestarman: Updated. Thanks for review.

integration_tests/src/main/python/orc_test.py
firestarman: Great idea. 
Found a bug during testing, and fixed it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SchemaUtils.scala
firestarman: Nice catch.
CharType(len: Int) is also introduced in Spark 3.1.0. I chose to remove the code, since GPU ORC reader does not support the two types yet. 
I will add shims if needed in the future.

firestarman:
build

integration_tests/src/main/python/orc_test.py
firestarman: Updated the names in read schema with some upper chars, it should cover the case-sensitive things.

firestarman:
build

firestarman:
build

wbo4958:
LGTM

revans2 (approved):
This looks good to me.

PR #3397, opened 2021-09-07
Add tests for GpuInSet

### LINKED ISSUE #2764: Check if casting needed for Set and InSet after SPARK-35316
https://github.com/apache/spark/commit/cfde117c6f

allows set and inset to have types that are slightly different for the needle and the haystack.  They eliminate the cast in cases where java can implicitly cast the values. This allows for predicate push down in some cases where it could not work before, but I don't think that cudf is that smart, so we might need to fix it up ourselves in some cases.  But we need to test it.

sperlingxx:

Closes #2764

After tests, I found that current implementation of `GpuInSet` can fully adapt SPARK-35316. If I understand correctly, the datatype of needles is the same as haystack,  since we build the needles from child's datatype.  The potential problem is what if some values in needles can not cast to child's datatype, such as: `select cast(a as int) in (1L, 2L, 3147483647L)`. The lit value `3147483647L` can not be casted to INT32 due to out of range.  It seems we don't need to worry about that, because these outliers will be removed in [`UnwrapCastInBinaryComparsion`](https://github.com/apache/spark/blob/branch-3.2/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/UnwrapCastInBinaryComparison.scala#L242).

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/FilterExprSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/FilterExprSuite.scala
@@ -49,9 +49,10 @@ class FilterExprSuite extends SparkQueryCompareTestSuite {
       val createDF = (ss: SparkSession) => ss.read.parquet(path)
       val fun = (df: DataFrame) => df
           // In
-          .filter(col("ints").isin(90L, 95L, Int.MaxValue))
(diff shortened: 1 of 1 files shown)


sperlingxx:
build

sperlingxx:
build

revans2 (approved):

PR #3399, opened 2021-09-07
Fix collect_list test so it covers nested types properly

revans2:
The tests for collect_list had a small bug in them that made it so they didn't actually test with arrays, structs, and maps.  This fixes that and updates the type checking to allow it to work for all of the aggregation operators.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -2909,11 +2909,11 @@ object GpuOverrides {
       "Collect a list of non-unique elements, not supported in reduction",
       // GpuCollectList is not yet supported in Reduction context.
       ExprChecks.aggNotReduction(
-        TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 +
+        TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 + TypeSig.NULL +
             TypeSig.STRUCT + TypeSig.ARRAY + TypeSig.MAP),
         TypeSig.ARRAY.nested(TypeSig.all),
         Seq(ParamCheck("input",
-          (TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 +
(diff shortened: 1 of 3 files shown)


revans2:
build

jlowe (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2909,11 +2909,11 @@ object GpuOverrides {
       "Collect a list of non-unique elements, not supported in reduction",
       // GpuCollectList is not yet supported in Reduction context.
       ExprChecks.aggNotReduction(
-        TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 +
+        TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 + TypeSig.NULL +
             TypeSig.STRUCT + TypeSig.ARRAY + TypeSig.MAP),
         TypeSig.ARRAY.nested(TypeSig.all),
         Seq(ParamCheck("input",
-          (TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 +
+          (TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 + TypeSig.NULL +
```
ttnghia, line 2916: What is `TypeSig.NULL`? Does that mean a nullable column?

jlowe: It is Spark's `NullType` which is the datatype representing NULL values.  This is a separate concept from a nullable type.  For example, you might be parsing CSV data with an inferred schema, and Spark could infer the schema is `NullType` because it only contains null values and thus it couldn't deduce a specific type beyond that.

revans2: NullType really only shows up as a corner case of SQL, but we do some abusive things in our tests. If I run `select null as a_col from table` what is the type of `a_col`? For Spark it is a NullType because it could not infer any other type from it.  If I did `select CAST(null as int) as a_col from table` now spark knows it is an int.

Issue #3400, opened 2021-09-07
[BUG] Canonicalized GPU plans sometimes not consistent when using Spark 3.2


andygrove:
**Describe the bug**
The following test suites have one or more failures due to `canonicalizationMatchesCpu=true != canonicalizationMatchesGpu=false` when running against Spark 3.2.0-SNAPSHOT

- HashAggregateSuite
- ExpandExecSuite
- ScalarSuubquerySuite

**Steps/Code to reproduce bug**
Run tests with Spark 3.2

**Expected behavior**
Tests should pass

**Additional context**

The issue seems to be related to a change in how partial aggregate expressions are represented in the plan, possibly related to a change in how filter expressions are handled. When running `ExpandExecSuite` against Spark 3.0.1, we see the following functions (with normalized `exprId`).

```
functions=[
  partial_gpucount(if ((1 = none#3)) none#1 else null), 
  partial_gpucount(if ((2 = none#3)) none#2 else null), 
  partial_gpufirst(if ((0 = none#3)) none#4L else null, true),
  ...
]
```

We see the following when running against Spark 3.2.0 and the `exprId` is not getting normalized.

```
functions=[
  partial_wrappedaggfunction(gpucount(none#355), wrappedaggfunction(gpucount(none#2), (2 = none#3))),
  partial_wrappedaggfunction(gpucount(none#354), wrappedaggfunction(gpucount(none#1), (1 = none#3))),
  partial_wrappedaggfunction(gpufirst(none#359L) ignore nulls, ....
]

```

andygrove:
Full plan with Spark 3.0.1

```
GpuColumnarToRow false
+- GpuHashAggregate(keys=[none#0], functions=[gpucount(if ((1 = none#0)) none#0 else null), gpucount(if ((2 = none#0)) none#0 else null), gpufirst(if ((0 = none#0)) none#0L else null, true), gpufirst(if ((0 = none#0)) none#0L else null, true), gpufirst(if ((0 = none#0)) none#0L else null, true)], output=[none#0, #0L, #1L, #2L, #3L, #4L])
   +- GpuShuffleCoalesce 2147483647
      +- GpuColumnarExchange gpuhashpartitioning(none#0, 2), true, [id=#233]
[...]

PR #3403, opened 2021-09-07
Fix canonicalization regression with Spark 3.2

andygrove:

Partially address #3400 

This PR extends the `exprId` normalization logic to also apply to aggregate expressions where `mode=Partial` and fixes two test failures (HashAggregatesSuite and ExpandExecSuite) when running against Spark 3.2.

See #3400 for more detail on why the behavior is different with Spark 3.2

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
@@ -149,12 +149,13 @@ case class GpuAggregateExpression(origAggregateFunction: GpuAggregateFunction,
   // We compute the same thing regardless of our final result.
   override lazy val canonicalized: Expression = {
     val normalizedAggFunc = mode match {
-      // For PartialMerge or Final mode, the input to the `aggregateFunction` is aggregate buffers,
-      // and the actual children of `aggregateFunction` is not used, here we normalize the expr id.
-      case PartialMerge | Final => aggregateFunction.transform {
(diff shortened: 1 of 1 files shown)


revans2 (approved):
This looks correct to me.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -149,12 +149,13 @@ case class GpuAggregateExpression(origAggregateFunction: GpuAggregateFunction,
...
-      // and the actual children of `aggregateFunction` is not used, here we normalize the expr id.
-      case PartialMerge | Final => aggregateFunction.transform {
+      // For Partial, PartialMerge, or Final mode, the input to the `aggregateFunction` is
+      // aggregate buffers, and the actual children of `aggregateFunction` is not used,
+      // here we normalize the expr id.
+      case Partial | PartialMerge | Final => aggregateFunction.transform {
         case a: AttributeReference => a.withExprId(ExprId(0))
       }
-      case Partial | Complete => aggregateFunction
+      case Complete => aggregateFunction
```
abellina, line 158: What happens if we normalize all? (i.e. no match).

andygrove: I did wonder about that as well but I really don't know. Prior to this PR, our code matches Spark, so there seems to be a reason not to apply this in all cases.

Issue #3401, opened 2021-09-07
[BUG] Qualification tool failed with java.lang.ArrayIndexOutOfBoundsException


viadea:
**Describe the bug**
Qualification tool failed with java.lang.ArrayIndexOutOfBoundsException:

```
java.lang.ArrayIndexOutOfBoundsException: 1
       at org.apache.spark.sql.rapids.tool.qualification.QualAppInfo$.$anonfun$parseReadSchemaForNestedTypes$5(QualAppInfo.scala:377)
       at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:237)
       at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
       at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
       at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
       at scala.collection.TraversableLike.map(TraversableLike.scala:237)
       at scala.collection.TraversableLike.map$(TraversableLike.scala:230)
       at scala.collection.AbstractTraversable.map(Traversable.scala:108)
       at org.apache.spark.sql.rapids.tool.qualification.QualAppInfo$.parseReadSchemaForNestedTypes(QualAppInfo.scala:377)
       at org.apache.spark.sql.rapids.tool.qualification.QualAppInfo.reportComplexTypes(QualAppInfo.scala:191)
       at org.apache.spark.sql.rapids.tool.qualification.QualAppInfo.$anonfun$aggregateStats$1(QualAppInfo.scala:219)
       at scala.Option.map(Option.scala:163)
       at org.apache.spark.sql.rapids.tool.qualification.QualAppInfo.aggregateStats(QualAppInfo.scala:198)
       at com.nvidia.spark.rapids.tool.qualification.Qualification.com$nvidia$spark$rapids$tool$qualification$Qualification$$qualifyApp(Qualification.scala:102)
       at com.nvidia.spark.rapids.tool.qualification.Qualification$QualifyThread.run(Qualification.scala:50)
       at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
       at java.util.concurrent.FutureTask.run(FutureTask.java:266)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[...]

PR #3405, opened 2021-09-07
Qualification tool: Filter empty strings from Read Schema

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3401. 

We see this error when  ReadSchema is not present in the eventlog for a job. 
In these cases we assign ""(empty string) for  read schema. We did not have test case for this where  the eventlog could contain some valid read schemas and an empty schema. Had test cases which tested above separately. Updated the tests to include this case.

### DIFF
--- a/tools/src/main/scala/org/apache/spark/sql/rapids/tool/qualification/QualAppInfo.scala
+++ b/tools/src/main/scala/org/apache/spark/sql/rapids/tool/qualification/QualAppInfo.scala
@@ -367,13 +367,30 @@ object QualAppInfo extends Logging {
     // Check if it has types
     val incompleteTypes = incompleteSchema.map { x =>
       if (x.contains("...") && x.contains(":")) {
-        x.split(":", 2)(1).split("\\.\\.\\.")(0)
+        val schemaTypes = x.split(":", 2)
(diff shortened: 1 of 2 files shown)


nartal1:
build

tools/src/main/scala/org/apache/spark/sql/rapids/tool/qualification/QualAppInfo.scala
```diff
@@ -351,8 +351,7 @@ object QualAppInfo extends Logging {
       if (angleBracketsCount == 0 && parenthesesCount == 0 && char.equals(',')) {
         individualSchema += tempStringBuilder.toString
         tempStringBuilder.setLength(0)
-      }
-      else {
+      } else {
         tempStringBuilder.append(char);
       }
     }
```
tgravescs, line 357: down below in the split 

> x.split(":", 2)(1).split("\\.\\.\\.")(0)
>     val completeTypes = completeSchema.map(x => x.split(":", 2)(1))

we should check to make sure there are 2 elements afte rthe split before blindingly accessing (1)

tgravescs (approved):

PR #3407, opened 2021-09-08
Add ml submodule

wjxiz1992:
- This PR add initial support for PCA algorithm in Spark ML.
- This module contains both native code, JNI code and scala code.
- The usage of new class could be found in PCASuite.scala

Co-authored-by: Rong Ou <rou@nvidia.com>


wjxiz1992:
build

ml/src/main/scala/org/apache/spark/ml/linalg/CUBLAS.scala
```diff
@@ -0,0 +1,66 @@
...
+    jniCUBLAS.dspr(v.size, v.values, U)
+  }
+
+  /**
+   * C := B.transpose * B
+   *
+   * @param B the matrix B that will be left multiplied by its transpose. Size of m x n.
+   * @param C the resulting matrix C. Size of n x n.
+   */
+  def gemm(B: DenseMatrix, C: DenseMatrix, deviceID: Int): Unit = {
```
GaryShen2008, line 50: We'd better add more comments to explain the parameters for all these functions.

wjxiz1992:
build

ml/src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -0,0 +1,240 @@
...
+   *       `PCA.fit()`.
+   */
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    val outputSchema = transformSchema(dataset.schema, logging = true)
+
+    // TODO(rongou): make this faster and re-enable.
+    //    if (getUseGemm) {
+    //      val transformed = dataset.toDF().rdd.mapPartitions(iterator => {
+    //        val gpuID = TaskContext.get().resources()("gpu").addresses(0)
+    //        val partition = iterator.toList
```
GaryShen2008, line 161: Why it's commented out?

ml/pom.xml
```diff
@@ -0,0 +1,346 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.nvidia</groupId>
+        <artifactId>rapids-4-spark-parent</artifactId>
+        <version>21.10.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>rapids-4-spark-ml_2.12</artifactId>
+    <name>RAPIDS Accelerator for Apache Spark ML</name>
+    <description>The RAPIDS cuML library for Apache Spark</description>
+    <version>21.10.0-SNAPSHOT</version>
```
gerashegalov, line 28: We have it everywhere but technically we can just inherit version from the parent pom

ml/pom.xml
```diff
@@ -0,0 +1,346 @@
...
+
+    <properties>
+        <java.version>1.8</java.version>
+        <spark.version>3.1.2</spark.version>
```
gerashegalov, line 32: why hardcode 3.1.2?

ml/pom.xml
```diff
@@ -0,0 +1,346 @@
...
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-mllib_${scala.binary.version}</artifactId>
+            <version>${spark.version}</version>
+        </dependency>
+
+        <!-- Test -->
+        <dependency>
+            <groupId>org.scalatest</groupId>
+            <artifactId>scalatest_${scala.binary.version}</artifactId>
+            <version>3.2.9</version>
```
gerashegalov, line 61: we should inherit dependency versions from the parent profiles/dependency management

ml/pom.xml
wjxiz1992: Removed.

ml/pom.xml
wjxiz1992: We are using some latest (>3.1.0) code from Spark ML side, and test over 3.1.0 and 3.1.2. The code cannot be compiled with versions such as 3.0.1. @rongou  Do we have plan to support earlier version? Please correct me if I miss anything.

ml/pom.xml
wjxiz1992: Thanks for the suggestion, I agree. But I tried to use version from parent and the compilation will fail. 
Most of the depencies are copied from Spark ML module to make sure it can run under the ML framework.
Currently this module is like an independent module though it's in the project. We haven't cover all versions for Spark and for the initial drop, can we firstly make it compilable (though for specific versions)?

ml/README.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+## Build
+The artifact jar for this submodule will not be included in the plugin jar of the whole spark-rapids project. User can build it directly in the _project root path_:
+```
+mvn -pl ml -am clean package
+```
+Then `rapids-4-spark-ml_2.12-21.10.0-SNAPSHOT.jar` will be generated under `ml/target` folder.
+
+_Note_: This module contains both native and Java/Scala code. When compiling native code to get the essential library, `cmake`, `ninja`, and `gcc` are all required.
```
tgravescs, line 28: perhaps more instructions on how to build native part. do I run a separate command, does it run as part of the mvn?  If part of mvn what happens if I don't have these installed?

ml/pom.xml
tgravescs: does it require spark 3.1.2 to run then I assume?  If so it should be documented in readme and really it should only build when building 3.1.2 so the top level pom file can move the module underneath just the 3.1.2 profile.

ml/pom.xml
```diff
@@ -0,0 +1,343 @@
...
+        <version>21.10.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>rapids-4-spark-ml_2.12</artifactId>
+    <name>RAPIDS Accelerator for Apache Spark ML</name>
+    <description>The RAPIDS cuML library for Apache Spark</description>
+
+    <properties>
+        <java.version>1.8</java.version>
+        <spark.version>3.1.2</spark.version>
+        <scala.version>2.12.14</scala.version>
```
tgravescs, line 32: similar hear, you require newer Scala version? otherwise inherit from parent which is 2.12.8

ml/pom.xml
```diff
@@ -0,0 +1,343 @@
...
+                    <include>LICENSE</include>
+                </includes>
+            </resource>
+            <resource>
+                <directory>${project.build.directory}/native-deps/</directory>
+            </resource>
+        </resources>
+
+        <plugins>
+            <!-- This plugin dumps the test classpath into a file -->
```
tgravescs, line 119: for what purpose?

ml/pom.xml
tgravescs: note, I assume its for 3.1.2+ just still means you should pick up the spark version being built but just not build for anything <

ml/native/src/cublas_jni.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+#include <cublas_v2.h>
+#include <cuda_runtime.h>
+#include <jni.h>
+
+extern "C" {
+
+JNIEXPORT void JNICALL Java_com_nvidia_spark_ml_linalg_JniCUBLAS_dspr(JNIEnv* env, jclass, jint n, jdoubleArray x,
+                                                                      jdoubleArray A) {
+  jclass jlexception = env->FindClass("java/lang/Exception");
```
jlowe, line 25: Does this need to set the GPU device ID like the other native methods?

ml/pom.xml
```diff
@@ -0,0 +1,343 @@
...
+    xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.nvidia</groupId>
+        <artifactId>rapids-4-spark-parent</artifactId>
+        <version>21.10.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>rapids-4-spark-ml_2.12</artifactId>
+    <name>RAPIDS Accelerator for Apache Spark ML</name>
+    <description>The RAPIDS cuML library for Apache Spark</description>
```
jlowe, line 27: What is the official name of this?  Elsewhere this is referred to as RAPIDS ML not cuML, and the docs don't ever mention "RAPIDS Accelerator for Apache Spark ML".  We should be consistent with the naming between here and the docs.

ml/pom.xml
```diff
@@ -0,0 +1,343 @@
...
+                            <goal>doc-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <scalaVersion>${scala.version}</scalaVersion>
+                    <checkMultipleScalaVersions>true</checkMultipleScalaVersions>
+                    <failOnMultipleScalaVersions>true</failOnMultipleScalaVersions>
+                    <recompileMode>incremental</recompileMode>
+                    <args>
```
jlowe, line 169: Why isn't this just inheriting the plugin configuration from the parent pom?  Ideally we shouldn't be compiling modules differently.  Was an override here really necessary?  If so please document why for future maintainers.

ml/src/main/java/com/nvidia/spark/ml/linalg/JniCUBLAS.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    Path temp;
+    try (InputStream resource = this.getClass().getClassLoader().getResourceAsStream(
+        String.format("%s/%s/libcublas_jni.so", osArch, osName))) {
+      assert resource != null;
+      Files.copy(resource, temp = Files.createTempFile("libcublas_jni.so", "",
+              PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rwxr-x---"))),
+          StandardCopyOption.REPLACE_EXISTING);
+      temp.toFile().deleteOnExit();
```
jlowe, line 46: In cudf we delete the file as soon as the library is loaded to minimize the chance of the file leaking if the JVM crashes hard and cannot proces the on exit tasks.

ml/src/main/scala/org/apache/spark/ml/feature/RapidsPCA.scala
```diff
@@ -0,0 +1,240 @@
...
+
+  /** @group getParam */
+  def getUseGemm: Boolean = $(useGemm)
+
+  /**
+   * The GPU id to use.
+   *
+   * @group param
+   */
+  private[ml] final val gpuId: IntParam = new IntParam(this, "gpuId", "the GPU id to use")
```
jlowe, line 59: Curious, how is this set in a distributed fashion?  For example, I'm running on a node with 4 GPUs and 4 executors.  How does each instance get the appropriately assigned ID?
```suggestion
  private[ml] final val gpuId: IntParam = new IntParam(this, "gpuId", "the GPU ID to use")
```

pom.xml
```diff
@@ -79,6 +79,7 @@
         <module>tests</module>
         <module>udf-compiler</module>
         <module>udf-examples</module>
+        <module>ml</module>
```
jlowe, line 82: The build instructions imply this isn't building by default, but it looks like it is.  Should this be built only with a specific profile?

jlowe (commented):
I don't understand why this is being proposed for this repository.

- Shares no dependencies with the RAPIDS Accelerator
- Does not ship with the RAPIDS Accelerator jar
- Does not support all the same Spark versions
- Does not use RMM, cudf, or the GPU detection mechanism used by the RAPIDS Accelerator
- Has native build requirements
- Overrides/ignores most parent pom configuration

All of this points to this really being a separate project with a separate artifact, and that to me indicates a separate repository makes the most sense here.

pom.xml
wjxiz1992: @jlowe  Thank you Jason for the review and suggestions, I've kept all those comments and will try to fix/answer them in another new repo. According offline discussion, we will request a new repo and put those work there. 
Thanks again!

PR #3409, opened 2021-09-08
Use devel cuda image for premerge CI

wjxiz1992:
To be able to build the upcoming `ml` module which includes native code:
- use devel cuda image
- set JAVA_HOME to JDK folder
- install cmake and ninja


pxLi:
build

pxLi (approved):
LGTM. support https://github.com/NVIDIA/spark-rapids/pull/3407

PR #3411, opened 2021-09-08
Make new build default and combine into dist package

tgravescs:
contributes to https://github.com/NVIDIA/spark-rapids/issues/3223

This pr is based off of https://github.com/NVIDIA/spark-rapids/pull/3381 and includes:

1) makes the new build targets the defaults. it builds a single version of spark by default and the dist jar only includes that single version unless other profiles specified.
2) removes many of the old ones
3) Combines the spark version packages into a single dist package - this requires installing all other versions into .m2. This adds some profiles to combine snapshots (-Pallwithsnapshots), snapshots with db (-PallwithsnapshotsDB), no snapshots (-Pallnosnapshots) and no snapshots with DB (-Pallnosnapshotswithdb).
4)  Supports building databricks 301db and 311db with PR 3381 changes
5) tries to update the jenkins build scripts
6) adds a scripts build/buildall that just installs all the apache and CDH versions and create a dist package with -Pallwithsnapshots.   This is very basic and we can make it better with more options later.
7) checked jar contents against older jars and it seems like we have everything for dependencies and properties files and license, notice, etc. 
8) removed shims/aggregator module since no longer needed
9) Added an aggregator module that is responsible for shading each of the spark versions. This combines the sql-plugin, shuffle, udf, etc and then shades the dependencies. Those aggregated modules are what end up being combined into the file uber jar in the dist directory where they are put into a spark specific directory.

The jar is much larger right now until we can dedup the code. it was at 115M

Default build with no targest builds spark301. to build another version you specify -Dbuildver=spark3XX.

I tested databricks builds and one of the apache spark version integration tests.  all versions ran unit tests.
I can't test on cloudera right now because the override shim in pr 3381 isn't working and it needs these changes to test so I think we can fix this after they are merged.

There is more to do but I think the release jar is working for everything I've tested so would be good to switch to it and get a base.

I have not tested UCX or the udf examples.

In progress of righting a nightly build to see if that script works.


tgravescs:
build

tgravescs:
I was trying to test all the profiles and tests to make sure this breaks as alittle as possible. I'm guessing I've missed something but would be good to find out sooner rather then later.

build/buildall
```diff
@@ -0,0 +1,29 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -ex
+
+# Install all the versions we support
+mvn -U -Dbuildver=301 install -Drat.skip=true -DskipTests
```
gerashegalov, line 21: We either need to interpolate scala-maven-plugin outputDirectory per buildver or call `clean` in every subsequent maven  to be sure we don't mix class files from different builds. Having interpolated output directories will also enable to call these maven commands in parallel

gerashegalov: This should be a for loop over buildver values

tgravescs:
build

build/buildall
tgravescs: liek I mentioned in comments there is a lot of things I want to do to improve this script, I'll add clean here but otherwise can do other stuff later.

tgravescs:
build

pom.xml
```diff
@@ -178,7 +110,9 @@
                                 <phase>generate-sources</phase>
                                 <configuration>
                                     <sources>
-                                        <source>${project.basedir}/src/main/spark30+all/scala</source>
+                                        <source>${project.basedir}/src/main/spark301+311-all/scala</source>
+                                        <source>${project.basedir}/src/main/spark301+nondb/scala</source>
```
tgravescs, line 114: oh I didn't document nondb, @gerashegalov thoughts on naming ?

gerashegalov: sounds intuitive enough, +1

dist/com.nvidia.spark.rapids.SparkShimServiceProvider.sparkNonSnapshot
```diff
@@ -0,0 +1,6 @@
+com.nvidia.spark.rapids.shims.spark301.SparkShimServiceProvider
```
tgravescs, line 1: Note hardcoded these for now, hopefully can make this smarter and generate on the fly later.

dist/shimmed-classes.txt
```diff
@@ -0,0 +1,272 @@
+com/nvidia/spark/rapids/AvoidAdaptiveTransitionToRow*
```
jlowe, line 1: Is there some kind of document that explains how this works?  Last we talked offline, I don't recall there being a need to explicitly list classes like this.

At a bare minimum there needs to be something in the contributing or dev docs that explains what people need to do if they are required to update these.  I'm hoping this is all temporary and we don't need to manually maintain these lists.

jlowe (commented):
I haven't gone through this in any detail, but I immediately noticed a lack of any documentation explaining the new method for building and what developers need to do (since it looks like there's a lot of explicit source listings now?)

dist/shimmed-classes.txt
gerashegalov: the list was generated and probably tweaked manually a couple of times in the prototype  branch. 

We only use unshimmed-*.txt files in this PR. So we don't even need to include this file AFAICT, and when we come to dedupe part of the prototype we hopefully won't need it hardcoded but just generated. 

So in  this PR almost everything is shimmed and it should be easy to document the principle for unshimmed, i.e., because somehow publicly exposed either via documentation, Spark configuration before Plugin can do anything or just common dependencies.

tgravescs: sorry just pushed some contributing docs

tgravescs: this fiel actually isn't used right now.  I pulled it in from the original dev branch and just left thinking it might be used later, but I can remove from this.

CONTRIBUTING.md
```diff
@@ -39,14 +39,34 @@ mvn verify
 ```
 
 After a successful build the RAPIDS Accelerator jar will be in the `dist/target/` directory.
+This will build the plugin for a single version of Spark.  By default this is Apache Spark
+3.0.1. To get an uber jar with more then 1 version you have to `mvn install` each version
+and then use one of the defined profiles in the dist module.
```
tgravescs, line 44: need to add more options here

tgravescs:
Note for nightly tests we may want to switch that a bit to first just build/install all the versions then build the uber jar and then use that uber jar to run the tests for all the versions.

tgravescs:
with the latest change to python files included in base the cudf_udf tests work on databricks

tgravescs:
build

CONTRIBUTING.md
```diff
@@ -39,14 +39,46 @@ mvn verify
...
+
+By default the distribution jar only includes code for a single version of Spark. If you want
+to create a jar with multiple versions we currently have 4 options.
 
-By default the build will only include shims for released versions of Spark. To include shims
-for snapshot versions of Spark still under development, use the `snapshot-shims` Maven profile
-(e.g.: add `-Psnapshot-shims` to the Maven command-line). Note that when a snapshot Spark version
-later becomes an official release, the snapshot shim for that version may no longer build due to
-missing snapshot artifacts for that Spark version.
+1. Build for all Apache Spark versions and CDH with no SNAPSHOT versions of Spark, only released. Use `-Pallnosnapshots`.
```
tgravescs, line 63: suggestions on naming conventions here?  I realize its a bit weird with allno or all nowithDB, etc..

tgravescs:
Note updated the description. there is a new module added here:

Added an aggregator module that is responsible for shading each of the spark versions. Those aggregated modules are what end up being combined into the file uber jar in the dist directory

build/buildall
```diff
@@ -0,0 +1,29 @@
...
+set -ex
+
+# Install all the versions we support
+mvn -U -Dbuildver=301 clean install -Drat.skip=true -DskipTests
+mvn -U -Dbuildver=302 clean install -Drat.skip=true -DskipTests
+mvn -U -Dbuildver=303 clean install -Drat.skip=true -DskipTests
+mvn -U -Dbuildver=304 clean install -Drat.skip=true -DskipTests
+mvn -U -Dbuildver=311 clean install -Drat.skip=true -DskipTests
+mvn -U -Dbuildver=312 clean install -Drat.skip=true -DskipTests
+mvn -U -Dbuildver=313 clean install -Drat.skip=true -DskipTests
```
gerashegalov, line 27: are we missing 320 in here?

tgravescs: left it out intentionally since audit stuff and tests not all passing.  I can add it in if we want.

note I left it out of nightly too mostly since I wasn't sure if tests work there.

gerashegalov: since  incompatibilities  in the non-patch releases is the main motivation for this whole epic, it's important that when we test  320 (even if just locally) we test it with a combined jar including at least one 3.0.x and a at least one 3.1.y

tgravescs: added it

build/buildall
```diff
@@ -0,0 +1,30 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -ex
+
+# Install all the versions we support
+mvn -U -Dbuildver=301 clean install -Drat.skip=true -DskipTests
```
gerashegalov, line 21: on my machine scaladoc/javadoc adds 40+ seconds per maven invocation.  Can we add `-Dskip`?

In dist we seem to depend only on classified rapids-4-spark-aggregator. So installing only up to this point should reduce the build time significantly as well if we add  -pl aggregator -am

CONTRIBUTING.md
revans2: It cannot be all if it does not include databricks.  I would replace all with default, standard, or just drop it all together.

`-PnoSnapshots`
`-Psnapshots`
`-PnoSnaphsotsWithDatabricks`
`-PsnapshotsWithDatabricks`

Also we should have a way to distinguish words in the profiles. snake_case is good, but I don't know what convention (if any) maven has or that we have used in our poms.

build/buildall
jlowe: `-Dskip` will skip the scaladoc, but don't we need to specify `-Dmaven.javadoc.skip` to also skip the Javadoc?

CONTRIBUTING.md
```diff
@@ -39,14 +39,46 @@ mvn verify
...
+3.0.1. To build against other versions of Spark you use the `-Dbuildver=XXX` command line option
+to maven. For instance to build Spark 3.1.1 you would use:
 
-### Building Shims for Spark Snapshot Versions
+```shell script
+mvn -Dbuildver=311 verify
+```
+You can find all available build versions in the top level pom.xml file. Currently
+they include 301, 302, 303, 304, 311, 312, 313, 320 and 311cdh. If you are building
+Databricks suggest you use the `jenkins/databricks/build.sh` script and modify it for
```
jlowe, line 51:
```suggestion
for Databricks then you should use the `jenkins/databricks/build.sh` script and modify it for
```

CONTRIBUTING.md
```diff
@@ -39,14 +39,46 @@ mvn verify
...
+the version you want.
+
+To get an uber jar with more then 1 version you have to `mvn install` each version
+and then use one of the defined profiles in the dist module. See the next section
+for more details.
+
+### Building a distribution for multiple versions of Spark
```
jlowe, line 58: Nit: Capitalization is not consistent with other headings.
```suggestion
### Building a Distribution for Multiple Versions of Spark
```

CONTRIBUTING.md
```diff
@@ -39,14 +39,46 @@ mvn verify
...
+2. Build for all Apache Spark versions and CDH including SNAPSHOT versions of Spark we have supported for. Use `-Pallwithsnapshots`.
+3. Build for all Apache Spark versions, CDH and Databricks with no SNAPSHOT versions of Spark, only released. Use `-PallnosnapshotswithDB`.
+4. Build for all Apache Spark versions, CDH and Databricks including SNAPSHOT versions of Spark we have supported for. Use `-PallwithsnapshotsDB`
+
+You must first build and install each of the versions of Spark and then build one final time using the profile for the option you want.
+
+For instance if I want all Apache Spark versions and CDH with no SNAPSHOT versions I would run something like this:
+
+```shell script
+mvn -U -Dbuildver=301 clean install -Drat.skip=true -DskipTests
```
jlowe, line 73: Where does the rat check run in this series of build commands?  Should we not be skipping it on the first build?

CONTRIBUTING.md
```diff
@@ -39,14 +39,46 @@ mvn verify
...
+mvn -U -Dbuildver=302 clean install -Drat.skip=true -DskipTests
```
jlowe, line 74: Is the `-U` necessary?  Seems like this is only needed when building for snapshots, and that doesn't seem to apply for these commands.

CONTRIBUTING.md
```diff
@@ -39,14 +39,46 @@ mvn verify
...
+mvn -U -Dbuildver=303 clean install -Drat.skip=true -DskipTests
+mvn -U -Dbuildver=311 clean install -Drat.skip=true -DskipTests
+mvn -U -Dbuildver=312 clean install -Drat.skip=true -DskipTests
+mvn -U -Dbuildver=311cdh clean install -Drat.skip=true -DskipTests
+mvn -U -pl dist -Pallnosnapshots package -DskipTests
+```
+There is also a build script `build/buildall` to build everything with snapshots and this will have more options to build later.
```
jlowe, line 81: Shouldn't we lead this section with the pointer to the script and go into nitty-gritty details after?  We want to point people to the thing we want them to do first.

I guess `buildall` doesn't support building without snapshots and that's why?  Seems like we'd want to fix that pronto.

build/buildall
tgravescs: yes I'll include all suggestions

gerashegalov: you are right, for the javadoc plugin config element is skip but not the property https://maven.apache.org/plugins/maven-javadoc-plugin/jar-mojo.html#skip . 

Technically we could unify this by adding `<skp>${skip}</skip> ` , or the other way around by adding a more mnemonic flag for scaladoc skip https://davidb.github.io/scala-maven-plugin/doc-jar-mojo.html#skip

aggregator/pom.xml
```diff
@@ -0,0 +1,313 @@
...
+                    <scope>provided</scope>
+                </dependency>
+                <dependency>
+                    <!-- required for conf generation script -->
+                    <groupId>org.apache.spark</groupId>
+                    <artifactId>spark-hive_${scala.binary.version}</artifactId>
+                    <scope>provided</scope>
+                </dependency>
+            </dependencies>
+            <build>
```
jlowe, line 78: Does this section need to be in the profile?  Seems to be the same across all profiles in this pom.

dist/README.md
```diff
@@ -0,0 +1,40 @@
+---
+layout: page
+title: Testing
+nav_order: 1
+parent: Developer Overview
+---
+# RAPIDS Accelerator for Apache Spark Distribution Packaging
+
+Currently the distribution package is for making an uber jar with the Spark versions you need combined into a single jar.
```
jlowe, line 9:
```suggestion
The distribution module creates a jar with support for the Spark versions you need combined into a single jar.
```

dist/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+By default, it will only include the single version of Spark. This runs the profile `individual`.
```
jlowe, line 11:
```suggestion
By default, it will only include support for a single version of Spark. This runs the profile `individual`.
```

dist/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+If you want to create a jar with multiple versions we currently have 4 options.
```
jlowe, line 13:
```suggestion
If you want to create a jar with support for multiple versions of Spark we currently have 4 options.
```

dist/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+1. Build for all Apache Spark versions and CDH with no SNAPSHOT versions of Spark, only released. Use `-Pallnosnapshots`.
```
jlowe, line 15: There's a lot of overlap here with what's in the contributing guide.  Seems like one of these should point to the other so we don't have to remember to update both locations when something changes.

dist/README.md
```diff
@@ -0,0 +1,40 @@
...
+2. Build for all Apache Spark versions and CDH including SNAPSHOT versions of Spark we have supported for. Use `-Pallwithsnapshots`.
+3. Build for all Apache Spark versions, CDH and Databricks with no SNAPSHOT versions of Spark, only released. Use `-PallnosnapshotswithDB`.
+4. Build for all Apache Spark versions, CDH and Databricks including SNAPSHOT versions of Spark we have supported for. Use `-PallwithsnapshotsDB`
+
+You have to first build and install the Spark versions you want combined into your .m2 and then run using appropriate profile above to combine them.
+
+Note that when you use these profiles there are currently some hardcoded service provider files that get put into place. 1 files for each of the
```
jlowe, line 22:
```suggestion
Note that when you use these profiles there are currently some hardcoded service provider files that get put into place, one file for each of the
```

dist/README.md
```diff
@@ -0,0 +1,40 @@
...
+above profiles. Please note that you will need to update these if adding or removing supported Spark versions.
```
jlowe, line 23:
```suggestion
above profiles. Please note that you will need to update these if adding or removing support for a Spark version.
```

dist/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Files are: `com.nvidia.spark.rapids.SparkShimServiceProvider.sparkNonSnapshot`, `com.nvidia.spark.rapids.SparkShimServiceProvider.sparkSnapshot`, `com.nvidia.spark.rapids.SparkShimServiceProvider.sparkNonSnapshotDB`, and `com.nvidia.spark.rapids.SparkShimServiceProvider.sparkSnapshotDB`.
+
+The new Uber jar is structured like:
```
jlowe, line 27:
```suggestion
The new uber jar is structured like:
```

dist/README.md
```diff
@@ -0,0 +1,40 @@
...
+3. META-INF base files are from 3.0.1  - maven, LICENSE, NOTICE, etc
+4. shaded dependencies for Spark 3.0.1 in case the base common classes needed them.
+5. Spark specific directory for each version of Spark supported in the jar. ie spark301/, spark302/, spark311/, etc.
+
+If you have to change the contents of the uber jar the following files control what goes into the base jar as classes that are not shaded.
+
+1. `unshimmed-classes.txt` - this has classes that should go into the base jar with their normal package name (not shaded). This includes user visible classes (ie com/nvidia/spark/SQLPlugin). Uses Spark 3.0.1 built jar for any of the uber jars.
+2. `unshimmed-base-extras.txt` - these are other files applied to the base version of Spark that stay in the base of the jar and not put into Spark version specific directories. Note we choose Spark 3.0.1 as the base version to use for base and unshimmed classes.
+3. `unshimmed-extras.txt` - These are files that are put into the base of the jar and not into the Spark specific directory from all of the other Spark version jars.
```
jlowe, line 39: Curious, why do we need two distinctions here and what would qualify a file to go in one place or the other?  Wondering why we can't just put these in the jar normally (i.e.: just have unshimmed-extras.txt).

aggregator/pom.xml
```diff
@@ -0,0 +1,313 @@
...
+                    <groupId>org.apache.spark</groupId>
+                    <artifactId>spark-sql_${scala.binary.version}</artifactId>
+                    <scope>provided</scope>
+                </dependency>
+                <dependency>
+                    <!-- required for conf generation script -->
+                    <groupId>org.apache.spark</groupId>
+                    <artifactId>spark-hive_${scala.binary.version}</artifactId>
+                    <scope>provided</scope>
+                </dependency>
```
jlowe, line 76: This pom doesn't generate the conf, which makes me wonder if we need to specify spark any of this `provided` stuff at all, and instead just pull in the plugin artifacts and what needs to be shaded.  Wondering if we can get out of the special-case-for-CDH stuff that appears here and in the dist when it only seems to be needed to run against Spark which this doesn't do.

dist/unshimmed-extras.txt
```diff
@@ -0,0 +1,2 @@
+META-INF/services/**
+spark-*-info.properties
```
jlowe, line 2: Also appears in unshimmed-base-extras.txt?

CONTRIBUTING.md
tgravescs: yeah we can rearrange and yes want to expand buildall very quickly after this

aggregator/pom.xml
tgravescs: yes I removed it from here but didn't cleanup the deps

dist/README.md
tgravescs: I don't think we need the classes vs extras distinction, copy from prototype, I'll cleanup.  The base vs other versions we may need because for the base we just choose one version of spark to pull the common classes (SQLPlugin, etc)

tgravescs:
build

CONTRIBUTING.md
```diff
@@ -39,14 +39,46 @@ mvn verify
...
 
-### Building Shims for Spark Snapshot Versions
+```shell script
+mvn -Dbuildver=311 verify
+```
+You can find all available build versions in the top level pom.xml file. If you are building
+for Databricks then you should use the `jenkins/databricks/build.sh` script and modify it for
+the version you want.
+
+To get an uber jar with more then 1 version you have to `mvn install` each version
```
firestarman, line 53: ?
```suggestion
To get an uber jar with more than 1 version you have to `mvn install` each version
```

CONTRIBUTING.md
```diff
@@ -39,14 +39,46 @@ mvn verify
...
-later becomes an official release, the snapshot shim for that version may no longer build due to
-missing snapshot artifacts for that Spark version.
+You must first build and install each of the versions of Spark and then build one final time using the profile for the option you want.
+
+There is a build script `build/buildall` to build everything with snapshots and this will have more options to build later.
+
+You can also install some manually and build a combined jar. For instance to build non-snapshot versions:
+
+```shell script
+mvn -Dbuildver=301 clean install -DskipTests
```
firestarman, line 74: Is `-Drat.skip=true` not needed for building 301 ?

pxLi (commented):
Do we plan to release w/ `-PnoSnaphsotsWithDatabricks`? If yes, can we include updates for
https://github.com/NVIDIA/spark-rapids/blob/branch-21.10/jenkins/deploy.sh and
https://github.com/NVIDIA/spark-rapids/blob/branch-21.10/jenkins/databricks/deploy.sh
? thanks~

tgravescs:
failing with shuffle issues due to some thing made common that shouldn't be, @gerashegalov  was working on in 3381

tgravescs:
yes should be with PnoSnaphsotsWithDatabricks. I was going to followup with deploy changes as I also need to probably do something with /dependency-reduced-pom.xml. If I have time I'll just do it with this.

CONTRIBUTING.md
tgravescs: we want it to run with one of the builds

tgravescs:
Upmerged to latest 3381 but need to test on databricks because that didn't handle it so don't build/review until I do that.

tgravescs:
build

tgravescs:
build

tgravescs:
I think I've addressed everyone comments so please take another look.

tgravescs:
build

build/buildall
```diff
@@ -0,0 +1,29 @@
...
+# Install all the versions we support
+mvn -U -Dbuildver=302 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -pl aggregator -am
+mvn -U -Dbuildver=303 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -pl aggregator -am
+mvn -U -Dbuildver=304 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -pl aggregator -am
+mvn -U -Dbuildver=311 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -pl aggregator -am
+mvn -U -Dbuildver=312 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -pl aggregator -am
+mvn -U -Dbuildver=313 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -pl aggregator -am
+mvn -U -Dbuildver=311cdh clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -pl aggregator -am
+mvn -U -Dbuildver=320 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -pl aggregator -am
+mvn -U -Dbuildver=301 clean install -Drat.skip=true -DskipTests -Psnapshots
```
jlowe, line 29: It looks like we're never running the rat checks on any build?  Seems like this last one would be a good one to use since it's not trying to teleport to the aggregator module.

dist/README.md
```diff
@@ -0,0 +1,29 @@
...
+layout: page
+title: Testing
+nav_order: 1
+parent: Developer Overview
+---
+# RAPIDS Accelerator for Apache Spark Distribution Packaging
+
+The distribution module creates a jar with support for the Spark versions you need combined into a single jar.
+
+See the [CONTRIBUTING.md](../CONTRIBUTING.md) doc for details on building and profiles available to build an Uber jar.
```
jlowe, line 11: Nit: We're not consistent on the capitalization of "uber jar" in this doc.  Capitalizing it makes me think it relates to a particular company, so personally I'd prefer the lower-case form.

tgravescs: I'll lower-case them

tgravescs:
build

jlowe (dismissed):

tgravescs:
working on upmerging

tgravescs:
build

tgravescs:
for some reason a tool test was failing which makes no sense with these changes, so I'm running against base branch to see if broken there, either way I'll file an issue to followup.

tgravescs:
ah tests failing due to gitlab down

tgravescs:
tools test fails in base too - issue https://github.com/NVIDIA/spark-rapids/issues/3435

tgravescs:
build

pxLi:
> ah tests failing due to gitlab down

looks like the recent test fail was due to timeout running UCX test not related to the gitlab down

```
10487 deselected / 36 selected
[2021-09-09T23:08:10.356Z] 
[2021-09-09T23:08:11.286Z] ../../src/main/python/hash_aggregate_test.py::test_hash_grpby_sum[{'spark.rapids.sql.variableFloatAgg.enabled': 'true', 'spark.rapids.sql.hasNans': 'false', 'spark.rapids.sql.castStringToFloat.enabled': 'true'}-[('a', Long), ('b', Integer), ('c', Long)]][IGNORE_ORDER, INCOMPAT, APPROXIMATE_FLOAT] 21/09/09 23:09:19 WARN RapidsShuffleInternalManager: Rapids Shuffle Plugin enabled. Transport enabled (remote fetches will use com.nvidia.spark.rapids.shuffle.ucx.UCXShuffleTransport. To disable the RAPIDS Shuffle Manager set `spark.rapids.shuffle.enabled` to false
[2021-09-09T23:08:13.805Z] 21/09/09 23:09:22 ERROR TaskSchedulerImpl: Lost executor 0 on 10.233.110.41: Remote RPC client disassociated. Likely due to containers exceeding thresholds, or network issues. Check driver logs for WARN messages.
[2021-09-09T23:08:13.805Z] 21/09/09 23:09:22 ERROR TaskSchedulerImpl: Lost executor 1 on 10.233.110.41: Remote RPC client disassociated. Likely due to containers exceeding thresholds, or network issues. Check driver logs for WARN messages.
[2021-09-09T23:08:23.744Z] 21/09/09 23:09:31 ERROR TaskSchedulerImpl: Lost executor 2 on 10.233.110.41: Remote RPC client disassociated. Likely due to containers exceeding thresholds, or network issues. Check driver logs for WARN messages.
```

dist/unshimmed-base.txt
```diff
@@ -0,0 +1,26 @@
...
+com/nvidia/spark/rapids/ShimLoader*
+com/nvidia/spark/rapids/SparkShimServiceProvider*
+com/nvidia/spark/rapids/SparkShims*
+com/nvidia/spark/rapids/ShimVersion*
+com/nvidia/spark/rapids/SparkShimVersion*
+com/nvidia/spark/rapids/RapidsExecutorHeartbeatMsg*
+com/nvidia/spark/rapids/RapidsExecutorStartupMsg*
+com/nvidia/spark/rapids/RapidsExecutorUpdateMsg*
+org/apache/spark/sql/rapids/VisibleShuffleManager*
+org/apache/spark/sql/rapids/RapidsShuffleInternalManagerBase*
```
gerashegalov, line 15: I will rework this in a follow up after fixing the regression from prototype @tgravescs found. 
The general idea is that only the classes on the class load path from Spark are "unshimmed".  RapidsShuffleInternalManager* is wrapped into a Proxy that is loading it via ShimLoader from the plugin. So I need to make sure just like in the prototype it's not referenced directly in bytecode

docs/configs.md
```diff
@@ -138,259 +138,3 @@ incompatibilities.
 
 Name | SQL Function(s) | Description | Default Value | Notes
 -----|-----------------|-------------|---------------|------
-<a name="sql.expression.Abs"></a>spark.rapids.sql.expression.Abs|`abs`|Absolute value|true|None|
```
tgravescs, line 141: oop these should be here

jenkins/spark-nightly-build.sh
```diff
@@ -21,17 +21,18 @@ set -ex
...
-mvn -U -B -Pspark302tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
-mvn -U -B -Pspark303tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
-mvn -U -B -Pspark304tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
-mvn -U -B -Pspark311tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
-mvn -U -B -Pspark312tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
-mvn -U -B -Pspark313tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
-# Disabled until Spark 3.2 source incompatibility fixed, see https://github.com/NVIDIA/spark-rapids/issues/2052
-#mvn -U -B -Pspark320tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
+# Install all the versions we support
+mvn -U -B -Dbuildver=302 install $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
```
gerashegalov, line 25: for bullet-proof correctness, we need `clean install` until the compiler output paths are interpolated

tgravescs: updated

tgravescs:
build

dist/pom.xml
```diff
@@ -17,332 +17,1519 @@
...
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.rat</groupId>
+                <artifactId>apache-rat-plugin</artifactId>
+                <configuration>
+                    <excludes>
+                        <exclude>com.nvidia.spark.rapids.SparkShimServiceProvider.*</exclude>
```
gerashegalov, line 1526: the shade plugin outputs  `dependency-reduced-pom.xml` under the pom's basedir by default since we were not using `dependencyReducedPomLocation` config. As a consequence many developers who have used the old build in their local repo prior to this PR will have a stale file under dist that is not cleaned up by maven

I suggest to add it to the exclude list in this PR to minimize reported build failures.

In a follow up we can add dependencyReducedPomLocation to the  shade plugin configs

tgravescs:
there are other shuffle related classes I need to handle to make those non-shimmed, working through them.

tgravescs:
the shuffle stuff got very ugly, started requiring a lot of things to be unshimmed which then eventually broke using it without rapids shuffle, need to investigate a better way to handle shuffle

dist/pom.xml
```diff
@@ -17,332 +17,1519 @@
...
+                            </execution>
+                            <execution>
+                                <phase>package</phase>
+                                <goals>
+                                    <goal>run</goal>
+                                </goals>
+                                <id>create-parallel-world</id>
+                                <configuration>
+                                    <target>
+
```
gerashegalov, line 459: TODO explore ant-contrib for loops and try to have a single execution configured where the missing input for unzip is skipped

gerashegalov:
shuffle should be fixed if you merge https://github.com/tgravescs/spark-rapids/pull/3 into your branch

tgravescs:
build

aggregator/pom.xml
```diff
@@ -0,0 +1,212 @@
...
+                         The shade plugin won't generate it when using the classifier and shadedArtifactAttached=true.
+                    -->
+                    <execution>
+                        <id>main</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                        <artifactSet>
```
revans2, line 76: nit: The indentation under configuration appears to be off, only filters appears to be indented correctly.

aggregator/pom.xml
```diff
@@ -0,0 +1,212 @@
...
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>classifierversion</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                    <artifactSet>
```
revans2, line 140: nit: Indention appears to be off here.

aggregator/pom.xml
```diff
@@ -0,0 +1,212 @@
...
+                    <!-- Unfortunately have to have 2 executions here to get dependency reduced pom.
+                         The shade plugin won't generate it when using the classifier and shadedArtifactAttached=true.
+                    -->
+                    <execution>
+                        <id>main</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
```
revans2, line 75: Have you tried to put the common configuration under the plugin, and only the differences in the execution itself?  That should hopefully make this much more readable, with less duplication.

build/buildall
```diff
@@ -0,0 +1,29 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -ex
+
+# Install all the versions we support
+mvn -U -Dbuildver=302 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -pl aggregator -am
```
revans2, line 21: I'm not sure if this was discussed before, but if we don't want rat, or tests or javadocs/etc, why do we have them on by default in the maven build at all?  I would much rather see a way to pass parameters to the shell script on to the maven build so I can decide what I want and what I don't instead.

build/buildall
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -ex
+
```
revans2, line 19: Should we do something to make sure that the directory is the correct one? This assumes that you are in the root directory calling `build/buildall`.

integration_tests/run_pyspark_from_build.sh
```diff
@@ -29,11 +29,13 @@ else
     if [ -d "$LOCAL_JAR_PATH" ]; then
         CUDF_JARS=$(echo "$LOCAL_JAR_PATH"/cudf-*.jar)
         PLUGIN_JARS=$(echo "$LOCAL_JAR_PATH"/rapids-4-spark_*.jar)
+        # TODO - this should really only pick up the specific version being tested
```
revans2, line 32: Is there a follow on issue for this?

jenkins/spark-nightly-build.sh
```diff
@@ -21,17 +21,18 @@ set -ex
...
-mvn -U -B -Pspark302tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
-mvn -U -B -Pspark303tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
-mvn -U -B -Pspark304tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
-mvn -U -B -Pspark311tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
-mvn -U -B -Pspark312tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
-mvn -U -B -Pspark313tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
-# Disabled until Spark 3.2 source incompatibility fixed, see https://github.com/NVIDIA/spark-rapids/issues/2052
-#mvn -U -B -Pspark320tests,snapshot-shims test $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
+# Install all the versions we support
+mvn -U -B -Dbuildver=302 clean install $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR -Dcuda.version=$CUDA_CLASSIFIER
```
revans2, line 25: If we made `buildall` configurable then we could reuse it here instead of copying everything

build/buildall
tgravescs: yeah the build script is very basic right now just to get started, lots of improvements to it need to be done

integration_tests/run_pyspark_from_build.sh
tgravescs: its subitem in https://github.com/NVIDIA/spark-rapids/issues/3223

tgravescs:
@revans2 thanks for the review, if you are ok, I'd like to do nits and things in followup since ci passed here and would like to get it as base?

revans2 (approved):
All of the issues I pointed out before can be done as follow on work.

aggregator/pom.xml
tgravescs: will fix in https://github.com/NVIDIA/spark-rapids/issues/3440

PR #3414, opened 2021-09-08
Add in support for transform_keys

### LINKED ISSUE #3413: [FEA] Add in support for transform_keys
It would be great to get support for transform_keys, similar to transform, and transform_values support that has already been added in.

revans2:
This closes  #3413

This also depends on https://github.com/rapidsai/cudf/pull/9198


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMapUtils.scala
```diff
@@ -0,0 +1,161 @@
...
+      }
+    }
+    val sameLengthCol = withResource(dedupeCountCol) { dedupeCountCol =>
+      withResource(mapView.countElements()) { origCount =>
+        origCount.equalToNullAware(dedupeCountCol)
+      }
+    }
+    withResource(sameLengthCol) { sameLengthCol =>
+      withResource(sameLengthCol.all()) { allSame =>
+        if (!allSame.getBoolean) {
```
jlowe, line 155: Annoyingly, `all()` can produce an invalid scalar in some cases.
```suggestion
        if (allSame.isValid && !allSame.getBoolean) {
```

revans2: That is why I did `equalToNullAware` instead of `equalTo`. It cannot produce a null output, but I think you are right. If there are no rows it still produces null.

revans2:
build

jlowe (approved):

Issue #3194, opened 2021-08-11
[FEA] Simplify the aggregation logic


revans2:
This is pure tech debt and is a follow on from https://github.com/NVIDIA/spark-rapids/pull/3178#discussion_r686816773.  The aggregation code is horribly complicated and is doing lots of special cases that the open source Spark code is not doing. I am not 100% sure why we have done all of this special case work. A lot of it feels like `setupReferences` is part of the issue and we need to do a deeper dive into it.

As a high level explanation of how this is supposed to work.

Each aggregation comes with a mode. Each mode tells the aggregation what to do as a part of that stage. Originally the code assumed that there would only ever be one mode for all of the aggregations. I thought we had ripped that all out and each aggregation does the right thing.

To successfully do an aggregation there are a few steps used.

1. Initial projection to get input data setup properly.
2. Initial aggregation to produce partial result(s)
3. Merge aggregation to combine partial results (This requires that the input schema and the output schema be the same)
4. Final projection to take the combined partial results and produce a final result.

In general the steps take the pattern 1, 2, 3*, 4. Which means 1, 2 and 4 are required and step 3 can be done as often as needed because the input and output schemas are the same.

Step 4 requires that all of the data for a given group by key is on the same task and has been merged into a single output row. There are several different ways to do this, which is why we end up with several aggregation modes.

   * Partial mode means that we do Step 1 and Step 2. Then we can do Step 3 as many times as needed depending on how we are doing memory management, and how many batches are needed.
   * PartialMerge mode means we can do Step 3 at least once and possibly more times depending on how we are doing memory management and how many batches are needed.
  *  Final mode means that we do the same steps as with PartialMerge but do Step 4 when we are done doing the partial merges.
  *  Complete mode is something only Databricks does, but it essentially means we do Step 1, Step 2, Step 3* (depending on memory management requirements), and Step 4 all at once.

The majority of the complexity appears to be in setupReferences where it tries to match up the data coming into the AggregateExec with what each aggregation wants at a given stage. 

I think @abellina said it well

>  I think the main ask is to not ... assum{e}... that a hash aggregate exec has a certain shape. If this function could decide per aggregate expression mode what the right binding should be, it should be more robust to new aggregate exec setups that mix and match modes (if we encounter new ones). ...

abellina:
Related issue: https://github.com/NVIDIA/spark-rapids/issues/3442

abellina:
[...]

PR #3417, opened 2021-09-09
Adds pre/post steps for merge and update aggregate

abellina:
This PR is prequel/continuation of https://github.com/NVIDIA/spark-rapids/pull/3373.

The work here adds part of the work by @ttnghia, and I added it to make sense of the extra processing added to the aggregate function expressions, and to be able to test it all. 

- The refactor adds a pre/post step to updates and merges. An example of a "pre merge" step is casting, or creating a struct, as is needed by `MERGE_M2`. The "pre update" case is not overloaded, and so it's the attribute reference as is (a pass-through projection).  The "post update" step can be used to cast (as is done in `GpuM2`) in the update, and then later in the merge, the "post merge" where a struct is decomposed, and fields casted, as expected by Spark.

- These steps allow 1 set of casting to be removed from the grouped aggregates in `aggregates.scala`. I did not mess with reduction aggregates in this PR, I can do that next. It was not required for the `stddev_pop` work.

- An untested (other than some quick examples in a shell) impl of `stddev_pop` is adapted from https://github.com/NVIDIA/spark-rapids/pull/3373 to demonstrate how the buffers are put together to produce the final result (`sqrt(M2/n)`). 

The code here really needs testing, as such there is no `GpuOverrides` node added in this PR. In other words, the code is there, but it is not being actively used. The two new projections for pre/post steps are getting executed by existing aggs.

I tested the diffs with the integration tests locally, and in databricks 8.2. I have not run in databricks 7.3 yet, but I wanted to get this up to get some :eyes:. Note that on databricks 8.2 I am noticing other issues with the tests (as did @revans2), especially when we run with the parallel setting. Tests were failing due to some unrelated bugs, so I'll re-run tests and comment here tomorrow, and we'll need some follow ups for that.


abellina:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -241,7 +244,7 @@ class GpuHashAggregateIterator(
       boundInputReferences: Seq[GpuExpression],
       boundFinalProjections: Option[Seq[GpuExpression]],
       boundResultReferences: Seq[Expression],
-      aggModeCudfAggregates: Seq[(AggregateMode, Seq[CudfAggregate])])
+      aggModeCudfAggregates: Seq[(GpuAggregateExpression, AggregateMode, Seq[CudfAggregate])])
```
revans2, line 247: For me any tuple above 2 entries that is not produced and consumed in essentially the same few lines of code needs to be replaced with a case class for readability.

abellina: Done in: https://github.com/NVIDIA/spark-rapids/pull/3417/commits/d2ff0e17fcee45ec4120b5484854f14b60229984

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -121,13 +121,15 @@ object AggregateUtils {
   }
 
   /**
-   * Compute the aggregation modes and aggregate expressions for all aggregation expressions
+   * Bind cuDF aggregate expressions depending on the aggregate expression mode
    * @param aggExpressions the aggregate expressions
    * @param aggBufferAttributes attributes to be bound to the aggregate expressions
```
revans2, line 126: nit: Could we clarify this a bit more? It is hard to tell when aggBufferAttributes is used and when mergeBufferAttributes is used. This is very minor because the code is small enough I can see when it is used easily.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -145,14 +147,18 @@ object AggregateUtils {
...
       }
-      (expr.mode, cudfAggregates)
+      BoundCudfAggregate(expr, cudfAggregates)
     }
   }
 }
 
+case class BoundCudfAggregate(
+    aggExpression: GpuAggregateExpression,
+    boundCudfAggregate: Seq[CudfAggregate])
```
revans2, line 160: Could we add some docs describing the requirements/guarantees associated with each of these? Like is CudfAggregate just the aggregate with GpuBoundReference as the input?  Why is this a sequence? Will there be one column of output for each aggregate?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -427,7 +433,10 @@ class GpuHashAggregateIterator(
       batches: mutable.ArrayBuffer[LazySpillableColumnarBatch]): LazySpillableColumnarBatch = {
     withResource(batches) { _ =>
       withResource(concatenateBatches(batches)) { concatVectors =>
-        withResource(computeAggregate(concatVectors, merge = true)) { mergedBatch =>
+        val concatBatch = new ColumnarBatch(
+          concatVectors.toArray,
+          concatVectors.head.getRowCount.toInt)
```
revans2, line 438: Is it possible to have a batch with no rows in it here? I don't think it is, but every time we assume something like this I want to be sure.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -567,10 +576,10 @@ class GpuHashAggregateIterator(
...
   /** Perform the initial projection on the input batch and extract the result columns */
-  private def processIncomingBatch(batch: ColumnarBatch): Seq[GpuColumnVector] = {
+  private def processIncomingBatch(batch: ColumnarBatch): ColumnarBatch = {
     val aggTime = metrics.computeAggTime
     withResource(new NvtxWithMetrics("prep agg batch", NvtxColor.CYAN, aggTime)) { _ =>
-      boundExpressions.boundInputReferences.safeMap { ref =>
+      val cols = boundExpressions.boundInputReferences.safeMap { ref =>
         val childCv = GpuExpressionsUtils.columnarEvalToColumn(ref, batch)
         if (childCv.dataType == ref.dataType) {
           childCv
```
revans2, line 585: Can we have a follow on issue to remove the automated casting that is just below this and replace it with explicit casting in each aggregation?  That would allow us to replace this code with just a `GpuProjectExec.project`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -752,21 +765,26 @@ class GpuHashAggregateIterator(
...
    * @param isSorted true indicates the data is already sorted by the grouping keys
    * @return aggregated batch
    */
   private def computeAggregate(
-      toAggregateCvs: Seq[GpuColumnVector],
+      toAggregateBatch: ColumnarBatch,
       merge: Boolean,
       isSorted: Boolean = false): ColumnarBatch  = {
-    val aggModeCudfAggregates = boundExpressions.aggModeCudfAggregates
+    val groupingAttributes = groupingExpressions.map(_.asInstanceOf[NamedExpression].toAttribute)
```
revans2, line 782: Can we just change the type of groupingExpressions to match Spark?  `groupingExpressions: Seq[NamedExpression]` that way we don't have to cast here, but we cast when we make the HashAggregate so it is very clear what the requirements are?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -70,13 +70,28 @@ trait GpuAggregateFunction extends GpuExpression
...
   // update: first half of the aggregation (count = count)
   val updateExpressions: Seq[Expression]
 
+  // expression to use to modify pre and post a cuDF update aggregate
+  // preUpdate: modify an incoming batch to match the shape/type cuDF expects
+  // postUpdate and postUpdateAttr: take the output of a cuDF update aggregate and return
+  //   what spark expects
+  lazy val preUpdate: Seq[Expression] = aggBufferAttributes
+  lazy val postUpdate: Seq[Expression] = aggBufferAttributes
+  lazy val postUpdateAttr: Seq[AttributeReference] = aggBufferAttributes
```
revans2, line 84: I think it would be cleaner if we make `preUdate` and `postUpdate` be `NamedExpression`s. Then it tells us exactly what the attributes are that are coming out of it, and we don't need a separate postUpdateAttr.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
abellina: done.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
abellina: I am not sure I am able to do this. I tried doing this locally, but I end up using things that are not `NamedExpression`, for instance `GpuCast`. Is there way to obtain a `NamedExpression` in those scenarios?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
abellina: done.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
revans2: You wrap it with a GpuAlias. That can give a name to anything, but you have to be sure that the expression ids match, because that is what binding uses to find the same expression, not the name

abellina: Ok, I'll give this another try with GpuAlias.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -995,7 +1019,7 @@ abstract class GpuBaseAggregateMeta[INPUT <: SparkPlan](
   override def convertToGpu(): GpuExec = {
     GpuHashAggregateExec(
       requiredChildDistributionExpressions.map(_.map(_.convertToGpu())),
-      groupingExpressions.map(_.convertToGpu()),
+      groupingExpressions.map(_.convertToGpu()).asInstanceOf[Seq[NamedExpression]],
```
revans2, line 1022: Because of type erasure (how java does generics) the sequence could contain things that are not NamedExpressions and the cast would happily pass. I would prefer to have us do something more like
```
groupingExpressions.map(_.convertToGpu().asInstanceOf[NamedExpression])
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
abellina: I don't think we could (i.e a Filter adds a coalesce, and after a shuffle we coalesce also), but I did change this impl a bit.

abellina: If the input to the agg is a join, I do not know what happens. If the join could output empty batches.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
abellina: https://github.com/NVIDIA/spark-rapids/issues/3442

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
abellina: done.

abellina:
I see tests passing in 7.3 and 8.2 for databricks and locally. I found a leak, and addressed it with: https://github.com/NVIDIA/spark-rapids/pull/3417/commits/a626e65ecb6e4e613ee1e41cce283fd613465887.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
abellina: @revans2 and I discussed on the side and we are moving towards a positional way of dealing with most of the aggregates. The main issue is it's not clear from the code what an expression should bind to. Currently we bind to attributes that mean different things, depending on the stage. 

A clearer approach is to bind when it's perfectly clear what the bind is for and in other cases assume order of the slots for the aggregation (and change the interface to take indices instead) when it is appropriate. I am working on this change, but to keep this PR smaller, I'd like to merge it as is, and continue separately with this (https://github.com/NVIDIA/spark-rapids/issues/3194)

abellina:
build

revans2 (approved):

ttnghia:
It seems that we have enough material to merge this PR. @abellina please remove the standard deviation stuffs (`CudfM2`, `GpuM2`, `GpuStddevPop`) as they will be reworked and reviewed in a separately PR.

abellina:
> It seems that we have enough material to merge this PR. @abellina before merging this please remove the standard deviation stuffs (CudfM2, CudfMergeM2, GpuM2, GpuStddevPop) as they will be reworked and reviewed in a separately PR.

I started doing this here https://github.com/NVIDIA/spark-rapids/pull/3417/commits/d4807c100f8c62c28d8c19c7897d588551917651, but re-reviewing my code I had comments around this because some of the changes are only for the M2 aggregates. Do you want me to remove comments as well, or rework them to not refer to M2?

At this point, you could let this in with the M2 reference implementation and change what you need to change, or just take the branch and do something on your own.

revans2 (approved):

abellina:
build

ttnghia:
> > It seems that we have enough material to merge this PR. @abellina before merging this please remove the standard deviation stuffs (CudfM2, CudfMergeM2, GpuM2, GpuStddevPop) as they will be reworked and reviewed in a separately PR.
> 
> I started doing this here [d4807c1](https://github.com/NVIDIA/spark-rapids/commit/d4807c100f8c62c28d8c19c7897d588551917651), but re-reviewing my code I had comments around this because some of the changes are only for the M2 aggregates. Do you want me to remove comments as well, or rework them to not refer to M2?
> 
> At this point, you could let this in with the M2 reference implementation and change what you need to change, or just take the branch and do something on your own.
> 
> Or merge with the comments that point to future aggregates, if you are going to put your patch up soon.

Thanks, I just merged the code you removed and will continue working on them.

Issue #3043, opened 2021-07-27
[FEA] Qualification tool: Add support to filter specific configuration values


nartal1:
**Is your feature request related to a problem? Please describe.**
As a user, I want to select only logs for applications that have specified certain configuration parameters (to given values, at all, etc) at launch

**Additional context**
- configuration parameters set in the initial environment update event is a reasonable proxy for "at launch"
- one example use case is selecting only jobs that have / do not have the plugin installed

PR #3418, opened 2021-09-09
Qualification tool: Add filtering based on configuration parameters

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3043.
Filtering for configuration parameters is done for specific values of configuration or keys of configs if values are not provided. 
Added tests for all cases and updated Readme.
If multiple configs are given where some have specific values and some have just the keys, then specific values takes precedence and filtering is done based on that.


nartal1:
build

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -102,21 +103,59 @@ class AppFilterImpl(
...
 
+    val configFiltered = if (configurationNames.nonEmpty) {
+      val userNameLogicFiltered = if (appArgs.any()) {
+        apps
+      } else {
+        userNameFiltered
+      }
+
+      if (configurationNames.toString.contains(":")) { // check for specific configuration values
+        val configFilteredResult = userNameLogicFiltered.filter { x =>
```
tgravescs, line 119: give x a better name

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -102,21 +103,59 @@ class AppFilterImpl(
...
     }
 
+    val configFiltered = if (configurationNames.nonEmpty) {
+      val userNameLogicFiltered = if (appArgs.any()) {
+        apps
+      } else {
+        userNameFiltered
+      }
+
+      if (configurationNames.toString.contains(":")) { // check for specific configuration values
```
tgravescs, line 118: so if any configs have values?  What if you mix and some have values and some don't?

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -102,21 +103,59 @@ class AppFilterImpl(
...
+        val configFilteredResult = userNameLogicFiltered.filter { x =>
+          x.appInfo.configInfo.exists { config =>
+            val individualConfigs = configurationNames.map(str => str.split(":"))
+            val validConfigs = individualConfigs.filter(vConfig => vConfig.size == 2)
+            if (validConfigs.nonEmpty) {
+              val validConfigsMap = validConfigs.map(a => a(0) -> a(1)).toMap
+              val allConfigs = config.configName // all configs from eventlog
+              validConfigsMap.filter(configArgs =>
```
tgravescs, line 126: any multi-line  filter/map etc should use {} rather then ()

tgravescs: this might be more efficient to get the keyset and do intersection on that first and then compare values

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -102,21 +103,59 @@ class AppFilterImpl(
...
+            } else {
+              false
+            }
+          }
+        }
+        configFilteredResult
+      } else { // include all eventlogs whose keys match
+        val configFilteredResult = userNameLogicFiltered.filter { x =>
+          x.appInfo.configInfo.exists { y =>
+            val configArgs = configurationNames
```
tgravescs, line 138: this isn't needed

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -102,21 +103,59 @@ class AppFilterImpl(
...
+            val allConfigsKeys = y.configName.keys.toList
+            configArgs.filter(x => allConfigsKeys.exists(_ == x)).nonEmpty
```
tgravescs, line 140: use intersect

tools/src/main/scala/org/apache/spark/sql/rapids/tool/FilterAppInfo.scala
```diff
@@ -19,13 +19,15 @@ package org.apache.spark.sql.rapids.tool
...
 
-import org.apache.spark.scheduler.{SparkListenerApplicationStart, SparkListenerEvent}
+import org.apache.spark.scheduler.{SparkListenerApplicationStart, SparkListenerEnvironmentUpdate, SparkListenerEvent}
 
 case class ApplicationStartInfo(
     appName: String,
     startTime: Long,
     userName: String)
 
+case class EnvironmentInfo(configName: Map[String, String])
```
tgravescs, line 29: rename configName to sparkProperties

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -640,6 +640,40 @@ class AppFilterSuite extends FunSuite {
...
+
+  test("Test disjunction match appName and config") {
+    testConjunctionAndDisjunction(appsNameConjunctionAndDisjunctionToTest,
+      filterAppName("nds") ++ configs("spark.driver.port:43492"), 6, "any")
+  }
+
+  test("Test conjunction match fileName and appName with non existent configs") {
+    testConjunctionAndDisjunction(appsNameConjunctionAndDisjunctionToTest,
+      matchFileName("app-nds") ++ configs("spark.driver.hosta")
+          ++ configs("spark.driver.porta") ++ filterAppName("Nds"), 0, "all")
```
tgravescs, line 668: should have test mixing key=value and just key

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -660,6 +694,10 @@ class AppFilterSuite extends FunSuite {
     Array("--user-name", name)
   }
 
+  def configs(configNames: String): Array[String] = {
```
tgravescs, line 697: rename filterConfigs

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
```diff
@@ -673,7 +711,8 @@ class AppFilterSuite extends FunSuite {
           // scalastyle:off line.size.limit
           val supText =
             s"""{"Event":"SparkListenerLogStart","Spark Version":"3.1.1"}
-               |{"Event":"SparkListenerApplicationStart","App Name":"${app.appName}", "App ID":"local-16261043003${app.uniqueId}","Timestamp":${app.appTime}, "User":"${app.userName}"}""".stripMargin
+               |{"Event":"SparkListenerApplicationStart","App Name":"${app.appName}", "App ID":"local-16261043003${app.uniqueId}","Timestamp":${app.appTime}, "User":"${app.userName}"}
+               |{"Event":"SparkListenerEnvironmentUpdate","JVM Information":{"Java Home":"/usr/lib/jvm/java-8-openjdk-amd64/jre"},"Spark Properties":{"spark.driver.host":"10.10.19.19","spark.eventLog.enabled":"true","spark.driver.port":"4349${app.uniqueId}"},"Hadoop Properties":{"hadoop.service.shutdown.timeout":"30s"},"System Properties":{"java.io.tmpdir":"/tmp"},"Classpath Entries":{"/home/user1/runspace/spark311/spark-3.1.1-bin-hadoop3.2/jars/hive-exec-2.3.7-core.jar":"System Classpath"}}""".stripMargin
```
tgravescs, line 715: test other configs in here other then just port.

docs/additional-functionality/qualification-profiling-tools.md
```diff
@@ -219,6 +219,19 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
...
+  -c, --configuration-names <arg>   Filter applications based on certain configurations that were
+                                    set during launch of the application. It can filtered based
+                                    on specific values or any values for a config. Multiple configs
+                                    can be provided where the filtering is done if any of the
+                                    config is present in the eventlog. Specific configurations
+                                    takes precedence over at all configurations.
+                                    filter on specific configuration:
+                                    --configuration-names=spark.eventLog.enabled:true
+                                    filter all eventlogs which has config:
+                                    --configuration-names:spark.driver.port
```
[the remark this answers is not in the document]
nartal1: I had it `=` initially but changed it `:` for better readability. 
 If we are providing keys and values, it would be something like:
` --configuration-names=spark.eventLog.enabled=true` which may be confusing.  Will change it back to `=`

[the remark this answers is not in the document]
tgravescs: that isn't what I meant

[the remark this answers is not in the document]
tgravescs: --configuration-names:spark.driver.port that should be  --configuration-names=spark.driver.port right?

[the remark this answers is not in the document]
nartal1: Sorry, got confused.  Will change it here.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
nartal1: If it was mix of key:value pairs and only keys, I was ignoring the keys configs. Had documented the behavior.
But I guess I should be considering both. i.e we may want to filter some with key:value pairs and some with only keys in the same command.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/FilterAppInfo.scala
```diff
@@ -39,15 +41,27 @@ class FilterAppInfo(
...
+    logDebug("Processing event: " + event.getClass)
+    val sparkProperties = event.environmentDetails("Spark Properties").toMap
+    configInfo = Some(EnvironmentInfo(sparkProperties))
   }
 
-  var appInfo: Option[ApplicationStartInfo] = None
+  var appStartInfo: Option[ApplicationStartInfo] = None
+  var configInfo: Option[EnvironmentInfo] = None
+  // This acts as a counter and it returns true when both the Listener events are processed.
+  var eventsToProcess: Int = Some(appStartInfo).size + Some(configInfo).size
```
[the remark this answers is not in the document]
nartal1: done

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -88,41 +89,109 @@ class AppFilterImpl(
...
+    val configFilteredForDisjunction = if (sparkProperties.nonEmpty && appArgs.any()) {
+      configFiltered
+    } else {
+      Seq.empty[AppFilterReturnParameters]
+    }
+
+    val appTimeFilteredForDisjunction = if (appArgs.startAppTime.isSupplied && appArgs.any()) {
+      appTimeFiltered
+    } else {
+      Seq.empty[AppFilterReturnParameters]
```
nartal1, line 187: There was a bug here - for disjunction(any) with only one filter parameter,  if the filter did not match to any eventlogs, then all the event logs were selected. Reason being `appNameFiltered` would be set to `apps` and then `userNameFiltered` would be to `appNameFiltered`. So in the code below finalLogicFiltered would be set to `apps` which is wrong.  So now I am making sure finalLogicFiltered gets the correct value. Added a test to veridy this.

docs/additional-functionality/qualification-profiling-tools.md
```diff
@@ -219,6 +219,18 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
                                     only logs for which all conditions are satisfied.
                                     Example: <Filter1> <Filter2> <Filter3> --all -> result
                                     is <Filter1> AND <Filter2> AND <Filter3>. Default is all=true.
+  -c, --conf-names <arg>            Filter applications based on certain Spark properties that were
```
tgravescs, line 222: sorry just thinking of this now, I"m wondering if we shouldn't rename this to filter-conf because --conf-names could be read as configs for the tool and we don't really need "names" plural since its one at a time.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -75,6 +75,7 @@ class AppFilterImpl(
     val filterAppName = appArgs.applicationName.getOrElse("")
     val filterCriteria = appArgs.filterCriteria.getOrElse("")
     val userName = appArgs.userName.getOrElse("")
+    val sparkProperties = appArgs.confNames.getOrElse(List.empty[String])
```
tgravescs, line 78: add filter to the name to differentiate it vs real sparkProperties

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -88,41 +89,109 @@ class AppFilterImpl(
...
+      } else {
+        userNameFiltered
+      }
+
+      // check if key:value pair is provided in the confNames filter. If those are provided
+      // along with just the key configs, then we have to make sure we consider both variations
+      // i.e key:pair and key configs to filter the event logs.
+      if (sparkProperties.toString.contains(":")) {
+        val configFilteredResult = userNameLogicFiltered.filter { appFilterReturnParameters =>
+          val individualConfigs = sparkProperties.map(str => str.split(":"))
```
tgravescs, line 115: simialr comment as before here, move this out of the filter and exists check so we only have to do it once.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -88,41 +89,109 @@ class AppFilterImpl(
...
+
+          appFilterReturnParameters.appInfo.sparkProperties.exists { sparkProp =>
+            val keyValueConfigs = individualConfigs.filter(kvConfig => kvConfig.size == 2)
```
tgravescs, line 118: these should move out of filter/exists check so we only do once

tools/src/main/scala/org/apache/spark/sql/rapids/tool/FilterAppInfo.scala
```diff
@@ -39,15 +41,27 @@ class FilterAppInfo(
...
       event.sparkUser
     )
-    appInfo = Some(thisAppInfo)
+    appStartInfo = Some(thisAppInfo)
+  }
+
+  def doSparkListenerEnvironmentUpdate(event: SparkListenerEnvironmentUpdate): Unit = {
+    logDebug("Processing event: " + event.getClass)
+    val sparkProperties = event.environmentDetails("Spark Properties").toMap
+    configInfo = Some(EnvironmentInfo(sparkProperties))
```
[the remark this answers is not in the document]
tgravescs: still question?

docs/additional-functionality/qualification-profiling-tools.md
nartal1: Should we rename it to spark-property ? so that it would be easy for user to mention the spark properties for which the event logs needs to be filtered.

tgravescs: yes spark-property is fine

tools/src/main/scala/org/apache/spark/sql/rapids/tool/FilterAppInfo.scala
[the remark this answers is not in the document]
nartal1: since we have changed it to spark-property now I didn't specify that we don't support hadoopProperties/system Properties in the doc. Please let me know if that should be explicitly mentioned.

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
nartal1: added tests with different combinations for port and host.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala

docs/additional-functionality/qualification-profiling-tools.md
```diff
@@ -277,6 +277,18 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
...
                                     days.
+      --spark-property <arg>        Filter applications based on certain Spark properties that were
+                                    set during launch of the application. It can filter based
+                                    on key:value pair or just based on keys. Multiple configs
+                                    can be provided where the filtering is done if any of the
+                                    config is present in the eventlog.
+                                    filter on specific configuration(key:value):
+                                    --conf-names=spark.eventLog.enabled:true
+                                    filter all eventlogs which has config(key):
+                                    --conf-names=spark.driver.port
```
tgravescs, line 288: update conf-names here to spark-property

tools/src/main/scala/com/nvidia/spark/rapids/tool/qualification/QualificationArgs.scala
```diff
@@ -111,6 +111,16 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
...
         "be very long. Default is false.",
       default = Some(false))
+  val sparkProperty: ScallopOption[List[String]] =
+    opt[List[String]](required = false,
+      descr = "Filter applications based on certain Spark properties that were set during " +
+          "launch of the application. It can filter based on key:value pair or just based on " +
+          "keys. Multiple configs can be provided where the filtering is done if any of the" +
+          "config is present in the eventlog. " +
+          "filter on specific configuration: --conf-names=spark.eventLog.enabled:true" +
+          "filter all eventlogs which has config:  --conf-names=spark.driver.port" +
```
tgravescs, line 121: same here spark-property

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -88,41 +89,108 @@ class AppFilterImpl(
...
     }
 
+    val configFiltered = if (filterSparkProperties.nonEmpty) {
+      val userNameLogicFiltered = if (appArgs.any()) {
+        apps
+      } else {
+        userNameFiltered
+      }
+
+      // check if key:value pair is provided in the confNames filter. If those are provided
```
tgravescs, line 110: update confNames

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -88,41 +89,108 @@ class AppFilterImpl(
...
+              val allConfigs = sparkProp.configName // all configs from eventlog
+              val allConfigKeys = sparkProp.configName.keys.toList // for keys only confs
+              //Intersection of configs provided in the filter args with event log configs.
+              val commonConfigs = validConfigsMap.keySet.intersect(allConfigs.keySet).map(
+                k => (k, allConfigs(k))).toMap
+
+              validConfigsMap.filter { configArgs =>
+                commonConfigs.exists(eventLogConf =>
+                  configArgs._1 == eventLogConf._1 && configArgs._2 == eventLogConf._2)
+              }.nonEmpty || filterSparkProperties.intersect(allConfigKeys).nonEmpty
```
tgravescs, line 131: this should use keysOnlyConfigs

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -88,41 +89,106 @@ class AppFilterImpl(
...
+          appFilterReturnParameters.appInfo.sparkProperties.exists { sparkProp =>
+            if (keyValueConfigs.nonEmpty || keysOnlyConfigs.nonEmpty) {
+              val allConfigs = sparkProp.configName // all configs from eventlog
+              //Intersection of configs provided in the filter args with event log configs.
+              val commonConfigs = validConfigsMap.keySet.intersect(allConfigs.keySet).map(
+                k => (k, allConfigs(k))).toMap
+
+              commonConfigs.keys.filter { key =>
+                commonConfigs(key) == validConfigsMap(key)
+              }.nonEmpty || filterSparkProperties.intersect(keysOnlyConfigs).nonEmpty
```
tgravescs, line 129: You changed the wrong variables here.  If tests are passing we have a problem. 
isn't this intersecting the same thing, both of these comes from filterSparkProperties.  you should be intersecting with allConfigs.  My previous comment was to change filterSparkProperties to keysOnlyConfigs

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/AppFilterSuite.scala
tgravescs: my point here was to add in other variations of configs that might not always be there, things like the sql plugin config, shuffle manager config, what about confs that take a comma separated list.  Test other patterns to make sure we match properly

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
```diff
@@ -88,41 +89,106 @@ class AppFilterImpl(
...
+            } else {
+              false
+            }
+          }
+        }
+        configFilteredResult
+      } else { // include all eventlogs whose keys match
```
tgravescs, line 136: this whole else statement really seems extra, the above code should handle it with the key value configs just being empty

docs/additional-functionality/qualification-profiling-tools.md
```diff
@@ -277,6 +277,18 @@ Usage: java -cp rapids-4-spark-tools_2.12-<version>.jar:$SPARK_HOME/jars/*
                                     min(minute),h(hours),d(days),w(weeks),m(months).
                                     If a period is not specified it defaults to
                                     days.
+      --spark-property <arg>        Filter applications based on certain Spark properties that were
+                                    set during launch of the application. It can filter based
+                                    on key:value pair or just based on keys. Multiple configs
+                                    can be provided where the filtering is done if any of the
+                                    config is present in the eventlog.
+                                    filter on specific configuration(key:value):
```
tgravescs, line 285: does the : work with spark.master = spark://5.6.7.8:7077  ?

What about regex like spark.redaction.regex or spark.eventlog.dir (file:///tmp/spark-events)

This is why I want more tests, go through configs at http://spark.apache.org/docs/latest/configuration.html and look for various formats

nartal1: Thanks for the providing the link.  Went through the list to determine different formats and added tests. Please let me know if you would like to see more tests.  For spark.redaction.regex it seems we can filter only based on keys as the values are redacted.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/AppFilterImpl.scala
nartal1: You are right. I think I had this when I was not doing partial(key) matching when key:value pair was given.

tgravescs (approved):

tgravescs:
build

Issue #2777, opened 2021-06-22
[FEA] Audit 3.2.0 - [SPARK-35152][SQL] ANSI mode: IntegralDivide throws exception on overflow


andygrove:
**Is your feature request related to a problem? Please describe.**

IntegralDivide should throw an exception on overflow in ANSI mode.

There is only one case that can cause that:
```
Long.MinValue div -1
```

**Describe the solution you'd like**
Throw exception on onverflow

**Describe alternatives you've considered**
None

**Additional context**
See https://github.com/apache/spark/commit/43ad939a7e

PR #3420, opened 2021-09-09
`IntegralDivide` throws an exception on overflow in ANSI mode

firestarman:
Let `GpuIntegralDivide` throw an exception on overflow in ANSI mode, to align with what has been introduced by apache/spark@43ad939.

This fixes https://github.com/NVIDIA/spark-rapids/issues/2777


firestarman:
build

firestarman:
This is likey to conflict with https://github.com/NVIDIA/spark-rapids/pull/3411. 
The shim part needs to be updated if 3411 gets merged first.

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -535,3 +535,35 @@ def test_div_by_zero_ansi(expr):
 @pytest.mark.parametrize('expr', ['1/0', 'a/0', 'a/b'])
 def test_div_by_zero_nonansi(expr):
     _test_div_by_zero(ansi_mode='nonAnsi', expr=expr)
+
+
+def _test_div_overflow(ansi_mode, expr):
+    ansi_conf = {'spark.sql.ansi.enabled': ansi_mode == 'ansi'}
+    data_gen = lambda spark: two_col_df(
```
revans2, line 542: If you know what the data is that you want, you don't have to use data_gen to do it. Just put the exact data in that you want.

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -535,3 +535,35 @@ def test_div_by_zero_ansi(expr):
...
+    else:
+        assert_gpu_and_cpu_are_equal_collect(div_overflow_func, ansi_conf)
+
+div_overflow_exprs = [
+    'CAST(-9223372036854775808L as LONG) DIV -1',
+    'a DIV CAST(-1 AS INT)',
+    'a DIV b']
+
+@pytest.mark.parametrize('expr', div_overflow_exprs)
+@pytest.mark.xfail(condition=is_before_spark_320(), reason='https://github.com/apache/spark/pull/32260')
```
revans2, line 562: Can we instead have two tests? One that is for 320 and above which verifies that ansi mode is working and another for before 320 that verifies it is not.  If an xfail passes we don't fail the test run.

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -535,3 +535,35 @@ def test_div_by_zero_ansi(expr):
...
+def test_div_overflow_ansi(expr):
+    _test_div_overflow(ansi_mode='ansi', expr=expr)
+
+
+@pytest.mark.parametrize('expr', div_overflow_exprs)
+def test_div_overflow_nonansi(expr):
```
revans2, line 568: Why not just have a second boolean parameter for ANSI vs not-ANSI instead of a separate test with the same implementation?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -282,6 +282,46 @@ object GpuDivModLike {
       case t => throw new IllegalArgumentException(s"Unexpected type: $t")
     }
   }
+
+  def isDivOverflow(left: GpuColumnVector, right: GpuColumnVector): Boolean = {
```
revans2, line 286: nit: A comment here might be nice explaining the exact condition that this can happen in.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -282,6 +282,46 @@ object GpuDivModLike {
...
+          }
+        }
+      case _ => false
+    }
+  }
+
+  def isDivOverflow(left: GpuColumnVector, right: GpuScalar): Boolean = {
+    left.dataType() match {
+      case LongType =>
+        (right.getValue == -1) && {
```
revans2, line 309: Need null check for right.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -282,6 +282,46 @@ object GpuDivModLike {
...
+          withResource(Scalar.fromLong(Long.MinValue)) { minLong =>
+            left.getBase.contains(minLong)
+          }
+        }
+      case _ => false
+    }
+  }
+
+  def isDivOverflow(left: GpuScalar, right: GpuColumnVector): Boolean = {
+    (left.getValue == Long.MinValue) && {
```
revans2, line 319: Need null check for left

firestarman:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
firestarman: Good catch. Added it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
firestarman: Good catch. Added it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
firestarman: Thanks for reivew. Added.

integration_tests/src/main/python/arithmetic_ops_test.py
firestarman: Thanks for review. Updated the tests.

integration_tests/src/main/python/arithmetic_ops_test.py
firestarman: Thanks for review. Updated the tests.

integration_tests/src/main/python/arithmetic_ops_test.py
firestarman: Got it. Updated.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
```diff
@@ -282,6 +282,54 @@ object GpuDivModLike {
...
+   */
+  def isDivOverflow(left: GpuColumnVector, right: GpuColumnVector): Boolean = {
+    left.dataType() match {
+      case LongType =>
+        withResource(Scalar.fromLong(Long.MinValue)) { minLong =>
+          withResource(left.getBase.equalTo(minLong)) { eqToMinLong =>
+            withResource(Scalar.fromInt(-1)) { minusOne =>
+              withResource(right.getBase.equalTo(minusOne)) { eqToMinusOne =>
+                withResource(eqToMinLong.and(eqToMinusOne)) { overFlowVector =>
+                  withResource(overFlowVector.any()) { isOverFlow =>
```
revans2, line 302: @jlowe what is the behavior of any around nulls? I thought that there were some issues?

revans2 (commented):
Looks good, I just want to be sure that any is going to do the right thing when there are nulls.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala
jlowe: `any()` like `all()` will skip null values, but annoyingly it will return an invalid scalar if the column has no other valid values.
```
scala> val c = ai.rapids.cudf.ColumnVector.fromBoxedBooleans()
c: ai.rapids.cudf.ColumnVector = ColumnVector{rows=0, type=BOOL8, nullCount=Optional[0], offHeap=(ID: 53 7fa0a9fafed0)}

scala> c.any().isValid
res8: Boolean = false

scala> val c = ai.rapids.cudf.ColumnVector.fromBoxedBooleans(null, null, null)
c: ai.rapids.cudf.ColumnVector = ColumnVector{rows=3, type=BOOL8, nullCount=Optional[3], offHeap=(ID: 60 7fa0a8494670)}

scala> c.any().isValid
res9: Boolean = false
```

I filed https://github.com/rapidsai/cudf/issues/8757 a while back to track this unexpected behavior from cudf.

revans2 (approved):

firestarman:
build

firestarman:
build

tgravescs:
Please try to remember to check databricks whenever adding something to SparkShims

PR #3421, opened 2021-09-09
Revert to CUDA runtime image for build

jlowe:
Reverts #3409 which was added to support #3407, and we're not planning on adding an ml module with native build dependencies to this repo, at least in the short term.

Without native build requirements, the repository should build with just the CUDA runtime and without cmake or ninja.  Reverting the change in #3409 ensures that is regularly tested as part of CI.


jlowe:
build

abellina (approved):

tgravescs (approved):

revans2 (approved):

PR #3423, opened 2021-09-09
Fix hash_aggregate tests that leaked configs

revans2:
There were some tests in hash_aggregate_test.py that leaked ANSI mode configs.  This only caused issues in a few rare cases, but I went through all of the other tests and added in a new API to be sure we always do this right. To hopefully make everything consistent so if someone copy/pastes a test they get the right thing. It also makes it simpler to look for suspicious code because there is only one place in the code that should be calling update on a config now.


revans2:
build

abellina (approved):

PR #3425, opened 2021-09-09
Throw an error by default if CREATE TABLE AS SELECT overwrites data

### LINKED ISSUE #2758: GpuCreateTableAsSelectCommand should not allow a non-empty location
[SPARK-28551](https://issues.apache.org/jira/browse/SPARK-28551) fixed a bug in Spark 3.2.0 where `CreateTableAsSelectCommand` could silently overwrite a directory with other subdirectories within it when the directory is specified as the location.

This change, implemented in https://github.com/apache/spark/pull/32411 and followup https://github.com/apache/spark/pull/32618, needs to be pulled into the RAPIDS Accelerator's `GpuCreateTableAsSelectCommand`.

jlowe:
Fixes #2758 by incorporating the changes from [SPARK-28551](https://issues.apache.org/jira/browse/SPARK-28551) into `GpuCreateDataSourceAsSelectCommand`.

### DIFF
--- a/integration_tests/src/main/python/parquet_write_test.py
+++ b/integration_tests/src/main/python/parquet_write_test.py
@@ -323,26 +323,24 @@ def test_write_map_nullable(spark_tmp_path):
 
 @pytest.mark.allow_non_gpu("DataWritingCommandExec", "HiveTableScanExec")
 @pytest.mark.parametrize('allow_non_empty', [True, False])
-def test_non_empty_ctas(spark_tmp_path, allow_non_empty):
+def test_non_empty_ctas(spark_tmp_path, spark_tmp_table_factory, allow_non_empty):
     data_path = spark_tmp_path + "/CTAS"
     conf = {
         "spark.sql.hive.convertCTAS": "true",
         "spark.sql.legacy.allowNonEmptyLocationInCTAS": str(allow_non_empty)
     }
     def test_it(spark):
+        src_name = spark_tmp_table_factory.get()
+        spark.sql("CREATE TABLE {}(id string) LOCATION '{}/src1'".format(src_name, data_path))
+        spark.sql("INSERT INTO TABLE {} SELECT 'A'".format(src_name))
+        ctas1_name = spark_tmp_table_factory.get()
+        spark.sql("CREATE TABLE {}(id string) LOCATION '{}/ctas/ctas1'".format(ctas1_name, data_path))
+        spark.sql("INSERT INTO TABLE {} SELECT 'A'".format(ctas1_name))
         try:
-            spark.sql("CREATE TABLE src1(id string) LOCATION '{}/src1'".format(data_path))
-            spark.sql("INSERT INTO TABLE src1 SELECT 'A'")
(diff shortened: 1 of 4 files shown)


jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDataWritingCommandExec.scala
```diff
@@ -54,6 +58,37 @@ trait GpuDataWritingCommand extends DataWritingCommand with ShimUnaryCommand {
   def requireSingleBatch: Boolean
 }
 
+object GpuDataWritingCommand {
+  private val allowNonEmptyLocationInCTASKey = "spark.sql.legacy.allowNonEmptyLocationInCTAS"
+
+  private def getAllowNonEmptyLocationInCTAS: Boolean = {
+    // config only exists in Spark 3.2+, so looking it up manually for now.
```
revans2, line 65: So I am a bit conflicted here because we are not consistent with how we handle new configs. Could we please come up with some rules/guidelines that we can use to decide when to add a config to the shim layer, and when to just hard code the config.

The pros for doing it this way is that there is no shim layer changes needed. The cons are that if for some reason the user has the config set, possibly because they are switching back and forth between versions of Spark, the behavior is not going to be consistent with the version of Spark.

For me I see both sides of it and I don't care which we use, but I want us to be consistent in how we use do it.

jlowe: The reason I didn't shim this one is because it's related to data loss and there's a workaround emitted in the error message if it comes up in practice.  I'm a bit surprised the Spark fix wasn't backported to earlier versions as as result.  I'm happy to make this a shim thing if we feel that's the better approach.

revans2: My goal is mainly that we have a consistent set of rules. I am fine with us doing different things in different situations, but I do want it to be documented and consistent. Can we have a follow on issue to look at figuring out these rules?

integration_tests/src/main/python/parquet_write_test.py
```diff
@@ -319,3 +320,29 @@ def generate_map_with_empty_validity(spark, path):
             generate_map_with_empty_validity,
             lambda spark, path: spark.read.parquet(path),
             data_path)
+
+@pytest.mark.allow_non_gpu("DataWritingCommandExec", "HiveTableScanExec")
+@pytest.mark.parametrize('allow_non_empty', [True, False])
+def test_non_empty_ctas(spark_tmp_path, allow_non_empty):
```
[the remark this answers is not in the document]
jlowe: Yep, that's a ton cleaner.  Thanks, I had forgotten about that test framework feature!

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDataWritingCommandExec.scala
jlowe: Filed #3427

revans2 (approved):

jlowe:
build

PR #3428, opened 2021-09-09
Remove commented-out semanticEquals overrides

### LINKED ISSUE #3375: [BUG]  SPARK-35742 Replace semanticEquals with canonicalize
**Describe the bug**
In 3.2 we can no longer override semanticEquals, and should instead have an appropriate `canonicalize` override 

https://github.com/NVIDIA/spark-rapids/blob/d2051d14ad17b568fd289059a2a64e1e73b1c2c2/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalarSubquery.scala#L44

**Expected behavior**
Implement a robust canonicalize for our execs

 local

**Additional context**
#3323

jlowe:
Fixes #3375 

Removes commented-out `semanticEquals` overrides.  As far as I could tell, the logic being removed should match the behavior of the default `semanticEquals` since it should canonicalize the plan as GPU case classes, therefore we shouldn't inadvertently match against a CPU plan.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
@@ -1451,22 +1451,6 @@ case class GpuCast(
   override def withTimeZone(timeZoneId: String): TimeZoneAwareExpression =
     copy(timeZoneId = Option(timeZoneId))
 
-  /**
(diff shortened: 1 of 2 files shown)


revans2 (approved):

andygrove (approved):

jlowe:
build

gerashegalov (approved):

PR #3433, opened 2021-09-10
Batch scalastyle checks across all modules upfront

### LINKED ISSUE #3388: [BUG] maven scalastyle checks don't appear to work for alterneate source directories
**Describe the bug**
I have noticed a few bugs in the scala style that the IDE picks up on in the 20.10

https://github.com/NVIDIA/spark-rapids/blob/f774e1a231416aab7cacc86f06010ef797fc19c7/sql-plugin/src/main/spark320/scala/com/nvidia/spark/rapids/shims/v2/Spark32XShims.scala#L23
`There should be no empty line separating imports in the same group.`

and

https://github.com/NVIDIA/spark-rapids/blob/f774e1a231416aab7cacc86f06010ef797fc19c7/sql-plugin/src/main/spark320/scala/org/apache/spark/rapids/shims/v2/GpuShuffleBlockResolver.scala#L17
`Package name 'org.apache.spark.sql.rapids.shims.v2' does not correspond to the file path 'org.apache.spark.rapids.shims.v2' `

are some examples of this.

It looks like the scala style plugin is ignoring additional source code directories.

gerashegalov:
- replaces scalastyle-maven-plugin with an ant task, uses compact wildcard paths so we don't need to hardcode evolving source code layout
- separates build info generation into pluginManagement to be inherited by submodules while limiting scalastyle task to the parent pom only
- checks all modules in one go upfront so the developers don't have to wait for failures incurred later after lengthy scala builds
- fixes codestyle violations accummulated while the check was incomplete
    
Fixes #3388
    

jlowe (dismissed):
Small nit but otherwise lgtm.  Is this going to conflict with #3411?

gerashegalov:
build

jlowe (approved):

gerashegalov:
looking into 311db failure

gerashegalov:
build

tgravescs (approved):

PR #3439, opened 2021-09-10
Stop parsing special dates for Spark 3.2+

### LINKED ISSUE #3382: [BUG] Support years with up to 7 digits when casting from String to Date in Spark 3.2
**Describe the bug**

We have test failures in `CastOpSuite` now that Spark 3.2 supports years with up to 7 digits when casting from string to date.

```
- Cast from string to date using random inputs *** FAILED ***
  Mismatch casting string [\n90713] to DateType. CPU: 713-01-01; GPU: null (CastOpSuite.scala:234)
- Cast from string to date using random inputs with valid year prefix *** FAILED ***
  Mismatch casting string [2021838] to DateType. CPU: 838-01-01; GPU: null (CastOpSuite.scala:234)
```

Run `CastOpSuite` with Spark 3.2

**Expected behavior**
Tests should pass.

N/A

**Additional context**
Spark commit: https://github.com/apache/spark/commit/c9813f74e98ea42d86cdb86c0eb6f989809a4237

See issue https://github.com/NVIDIA/spark-rapids/issues/3406 for supporting other changes in SPARK-35780.

### LINKED ISSUE #3383: [BUG] ParseDateTime should not support special dates with Spark 3.2
**Describe the bug**

ParseDateTimeSuite:

```
- WITH DECIMALS: to_date default pattern *** FAILED ***
  Mismatch between CPU and GPU for the following rows:
  
  [#0] CPU: [epoch,null]
  [#0] GPU: [epoch,1970-01-01]
  
  
  
  [#1] CPU: [now,null]
  [#1] GPU: [now,2021-09-03]
  
  
  
  [#2] CPU: [today,null]
  [#2] GPU: [today,2021-09-03]
  
  
  
  [#3] CPU: [yesterday,null]
  [#3] GPU: [yesterday,2021-09-02]
  
  
  
  [#4] CPU: [tomorrow,null]
  [#4] GPU: [tomorrow,2021-09-04] (ParseDateTimeSuite.scala:410)
```

Run ParseDateTimeSuite

**Expected behavior**
Tests should  pass

N/A

**Additional context**
Spark PR: https://github.com/apache/spark/commit/a59063d5446a78d713c19a0d43f86c9f72ffd77d

sperlingxx:

Fixes #3383

Current PR also fixes cases on casting string to date/timestamp  in  `CastOpSuite`, which introduce special dates. However,  these cases won't be entirely fixed until we fix #3382 through supporting the full range date/timestamp on GPU as what  [SPARK-35780](https://github.com/apache/spark/pull/32959) does.


sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -1093,8 +1077,35 @@ object GpuCast extends Arm {
               convertFixedLenTimestampOrNull(sanitizedInput, 4, "%Y"))))
 
       // handle special dates like "epoch", "now", etc.
-      val finalResult = specialDates.foldLeft(converted)((prev, specialDate) =>
-        specialTimestampOr(sanitizedInput, specialDate._1, specialDate._2, prev))
+      val finalResult = withResource(daysEqual(sanitizedInput, DateUtils.EPOCH)) { isEpoch =>
```
firestarman, line 1080: It would be  a good to have this boilerplate code into a method.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -531,12 +531,31 @@ object GpuToTimestamp extends Arm {
     FIX_SINGLE_DIGIT_SECOND
   )
 
-  def daysScalarSeconds(name: String): Scalar = {
-    Scalar.timestampFromLong(DType.TIMESTAMP_SECONDS, DateUtils.specialDatesSeconds(name))
+  def daysScalarDays(name: String): Scalar = ShimLoader.getSparkVersion match {
+    // In Spark 3.2, special datetime values such as `epoch`, `today`, `yesterday`, `tomorrow`,
+    // and `now` are supported in typed literals only
+    case version if version >= "3.2" =>
```
firestarman, line 537: It would be better to add shims for this.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -938,8 +916,35 @@ object GpuCast extends Arm {
           convertFixedLenDateOrNull(sanitizedInput, 4, "%Y")))
 
       // handle special dates like "epoch", "now", etc.
-      specialDates.foldLeft(converted)((prev, specialDate) =>
-        specialDateOr(sanitizedInput, specialDate._1, specialDate._2, prev))
+      withResource(daysEqual(sanitizedInput, DateUtils.EPOCH)) { isEpoch =>
```
revans2, line 919: This change feels like a step backwards.  Before with the folding the maximum extra memory needed on the GPU was a boolean column, and the new output column.  With this change we now have 5 boolean columns and 5 temporary output columns.

revans2 (changes requested):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
andygrove: If we do keep the version check here (which I am ok with personally because it reduces complexity in the shim layers) then we should use more robust logic for the version check. We have some code in the test suite that could be moved elsewhere.

https://github.com/NVIDIA/spark-rapids/blob/branch-21.10/tests/src/test/scala/com/nvidia/spark/rapids/SparkQueryCompareTestSuite.scala#L1848-L1849

tests/src/test/scala/com/nvidia/spark/rapids/ParseDateTimeSuite.scala
```diff
@@ -228,11 +228,8 @@ class ParseDateTimeSuite extends SparkQueryCompareTestSuite with BeforeAndAfterE
           .repartition(2)
           .withColumn("c1", unix_timestamp(col("c0"), "yyyy-MM-dd HH:mm:ss"))
     }
-    val startTimeSeconds = System.currentTimeMillis()/1000L
     val cpuNowSeconds = withCpuSparkSession(now).collect().head.toSeq(1).asInstanceOf[Long]
     val gpuNowSeconds = withGpuSparkSession(now).collect().head.toSeq(1).asInstanceOf[Long]
-    assert(cpuNowSeconds >= startTimeSeconds)
```
andygrove, line 234: What is the reason for removing these assertions? Are they no longer valid?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
gerashegalov: +1 to keep the version check.   I would suggest to move @andygrove's  implementation to [ShimVerion ](https://github.com/NVIDIA/spark-rapids/blob/9742ea7790b8f56cac2dba80c1900e72ce76e5a1/sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala#L64)  to make shimversions comparable

revans2: The reason to use a shim over a version check is mostly databricks and other vendors that pull back "fixes" not directly related to version numbers. I doubt it will happen in this case, but similar things have happened in the past.

gerashegalov: I think it's something we will find out via the test being broken on databricks. If we go with the ShimVersion comparison implementation, the fix is either an additional condition on databricks shim version compare or it's more involved at which point we will have add a new shim method.

tests/src/test/scala/com/nvidia/spark/rapids/ParseDateTimeSuite.scala
sperlingxx: Yes, under spark 3.2+, the result will be zero instead of current time, since `NOW` is not longer being parsed in 3.2.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
sperlingxx: Yes, it was not optimal.  I reworked this part, cleaning up temporary GPU resources as early as possible.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
sperlingxx: I replaced the version dispatch with a shim method.

sperlingxx:
build

pxLi:
the CI has been assigned to non-reserved instances which failed PVC driver

pxLi:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -1093,8 +1060,19 @@ object GpuCast extends Arm {
               convertFixedLenTimestampOrNull(sanitizedInput, 4, "%Y"))))
 
       // handle special dates like "epoch", "now", etc.
-      val finalResult = specialDates.foldLeft(converted)((prev, specialDate) =>
-        specialTimestampOr(sanitizedInput, specialDate._1, specialDate._2, prev))
+      // `converted` will be closed in replaceSpecialDates. We wrap it with closeOnExcept in case
+      // of exception before replaceSpecialDates.
+      val finalResult = closeOnExcept(converted) { timeStampVector =>
+        val specialDates = Seq(DateUtils.EPOCH, DateUtils.NOW, DateUtils.TODAY,
```
andygrove, line 1066: It seems expensive to replace the special dates with null with Spark 3.2 rather than just skip handling them at all. They will already be ignored by `is_timestamp`.

I will pull this PR locally today and experiment with it and come back with more detailed suggestions.

andygrove:
I looked at how I would approach this and I don't think there is a need to make any shim changes here. Spark 3.2 doesn't support the special dates so it seems confusing to add related logic there. There are also no changes in Spark APIs that would cause us to need shims.

I think this issue can be resolved more simply with some version checks in a couple of places:

In `DateUtils`, where we return maps of special dates to literal values we can return empty maps with Spark 3.2:

```scala
def specialDatesDays: Map[String, Int] = if (spark320orLater) {
  Map.empty
} else {
  val today = currentDate()
  Map(
    EPOCH -> 0,
    NOW -> today,
    TODAY -> today,
    YESTERDAY -> (today - 1),
    TOMORROW -> (today + 1)
  )
}
```
(and the same for `specialDatesSeconds` and `specialDatesMicros`).

Then in `org.apache.spark.sql.rapids.GpuToTimestamp#parseStringAsTimestamp` we can do:

```scala
if (spark320orLater) {
  withResource(isTimestamp(lhs.getBase, sparkFormat, strfFormat)) { isTimestamp =>
    withResource(asTimestamp(lhs.getBase, strfFormat)) { converted =>
      withResource(Scalar.fromNull(dtype)) { nullValue =>
        isTimestamp.ifElse(converted, nullValue)
      }
    }
  }
} else {
 // original complex logic that handles special dates
}
```

sperlingxx:
> I looked at how I would approach this and I don't think there is a need to make any shim changes here. Spark 3.2 doesn't support the special dates so it seems confusing to add related logic there. There are also no changes in Spark APIs that would cause us to need shims.
> 
> I think this issue can be resolved more simply with some version checks in a couple of places:
> 
> In `DateUtils`, where we return maps of special dates to literal values we can return empty maps with Spark 3.2:
> 
> ```scala
> def specialDatesDays: Map[String, Int] = if (spark320orLater) {
>   Map.empty
> } else {
>   val today = currentDate()
>   Map(
>     EPOCH -> 0,
>     NOW -> today,
>     TODAY -> today,
>     YESTERDAY -> (today - 1),
>     TOMORROW -> (today + 1)
>   )
> }
> ```
> 
> (and the same for `specialDatesSeconds` and `specialDatesMicros`).
> 
> Then in `org.apache.spark.sql.rapids.GpuToTimestamp#parseStringAsTimestamp` we can do:
> 
> ```scala
> if (spark320orLater) {
>   withResource(isTimestamp(lhs.getBase, sparkFormat, strfFormat)) { isTimestamp =>
>     withResource(asTimestamp(lhs.getBase, strfFormat)) { converted =>
>       withResource(Scalar.fromNull(dtype)) { nullValue =>
>         isTimestamp.ifElse(converted, nullValue)
>       }
>     }
>   }
> } else {
>  // original complex logic that handles special dates
> }
> ```

I fully agree on this approach, and I've made corresponding changes. Thanks for such a detailed explanation!

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DateUtils.scala
```diff
@@ -92,6 +104,17 @@ object DateUtils {
     )
   }
 
+  def fetchSpecialDates(unit: DType): Map[String, Scalar] = unit match {
+    case DType.TIMESTAMP_DAYS =>
+      DateUtils.specialDatesDays.map { case (k, v) => k -> Scalar.timestampDaysFromInt(v) }
```
revans2, line 109: nit: There is a period of time from when this function is called and scalars are populated into the map and when those scalars are setup to be closed.  I personally would prefer to see this return a `Map[String, () -> Scalar]` so that the scalar is generated as it is used, and then closed as soon as it is done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DateUtils.scala
```diff
@@ -57,7 +59,13 @@ object DateUtils {
   val YESTERDAY = "yesterday"
   val TOMORROW = "tomorrow"
 
-  def specialDatesDays: Map[String, Int] = {
+  private lazy val isSpark320OrLater: Boolean = {
+    ShimLoader.getSparkShims.getSparkShimVersion.toString >= "3.2"
```
revans2, line 63: My main concern with doing it this was is that we are starting to be inventive about how we handle different situations now that we have the shim layer.  I am fine with this here, but at some point we are going to need to think a lot about the pros and cons of each approach so it is more maintainable.  Before the ShimLayer was a single class so when we had to update code for a version or add a new version there was only one relatively small location we had to look at and think about.  Now it is spread all through the shim jars like before, in the sql plugin in the shim v2 directories (still isolated, but a lot bigger and harder to follow in some cases), and with this it starts to be anywhere in the code.  This is what makes me somewhat concerned.  In order to save a few lines of code we are making it much harder to know what the differences are from one version to another.

To ameliorate that could we have a single VersionUtil or something similar would be a central point for dealing with this type of version checks so we can then trace back everywhere that it is used?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -531,20 +533,30 @@ object GpuToTimestamp extends Arm {
...
+  /**
+   * Replace special date strings such as "now" with timestampDays. This method does not
+   * close the `stringVector` and `specialValues`.
+   */
+  def replaceSpecialDates(
+      stringVector: ColumnVector,
+      chronoVector: ColumnVector,
+      specialNames: Seq[String],
+      specialValues: Seq[Scalar]): ColumnVector = {
+    specialValues.zip(specialNames).foldLeft(chronoVector) { case (buffer, (scalar, name)) =>
```
revans2, line 551: If we make the Map value a function that produces the scalar, then we won't need to pull the map apart just so we can zip it back together again.Just call `.toSeq` on it, so we could pass it in directly here.

tests/src/test/scala/com/nvidia/spark/rapids/ParseDateTimeSuite.scala
revans2: Then have the test check the version. and have it check that now is replaced with a 0 instead.  Or just have the test skip entirely if it is 3.2+

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DateUtils.scala
andygrove: I agree that we need to use some well-thought-out version comparison methods for this kind of situation. Comparing based on strings also isn't safe. The above conditional would fail with spark version 3.10.0 for example (unlikely situation but you never know).

Another option here is to move the special dates map in to the shim layer, with 320 returning empty maps.

sperlingxx: Perhaps we can migrate version utils in spark-test to form a general propose VersionUtil?

sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
sperlingxx: Good idea! I made the change.

tgravescs:
I pulled this PR and we still have CastOpSuite test failures, was this intended to handle all of those? its referenced in description so I was assuming so

revans2:
> It will be a temporary solution, since we need naive support to fix #3382 completely.

We have to fall back to the CPU for this case unless the user opts into letting us do the cast anyways.  Disabling the tests is not a solution.

sperlingxx:
> > It will be a temporary solution, since we need naive support to fix #3382 completely.
> 
> We have to fall back to the CPU for this case unless the user opts into letting us do the cast anyways. Disabling the tests is not a solution.

Oh, I thought we already had corresponding options to disable casting from string to date/timestamp by default. But I found that we have only the config `RapidsConf.ENABLE_CAST_STRING_TO_TIMESTAMP` for timestamp. IIUC,  we also need to add `RapidsConf.ENABLE_CAST_STRING_TO_DATE` ?

revans2:
We should add in `RapidsConf.ENABLE_CAST_STRING_TO_DATE` and only use it for 3.2.0+. We also need to make it clear in the docs for it that it is only for 3.2.0+ because of 7 digit year support.  We should also update the docs for `RapidsConf.ENABLE_CAST_STRING_TO_TIMESTAMP` to include information about the 7 digit year support in 3.2.0+

revans2 (approved):
I am not thrilled with VersionUtils moving to the plugin as I explained [here](https://github.com/NVIDIA/spark-rapids/pull/3519#issuecomment-921888540).  But just to get something shipped I am going to approve this and we can revisit the topic later on.

revans2:
That said we do need to fix the 7 digit year parsing as a separate issue.

revans2:
Just FYI I filed #3530 to handle the 7 digit year fallback issue.  I will take a crack at it because hopefully I can get it done today.

PR #3445, opened 2021-09-10
Update nightly build to pull Databricks jars

tgravescs:
The databricks jars are now pushed to URM so we can update the nightly script to pull them.

I had to update the maven dependency copy plugin with useBaseVersion=true so that it keeps the SNAPSHOT name in version when download vs pulling the timestamp version.

Tested the command manually by adding urm repo


tgravescs:
build

jlowe (approved):

Issue #3440, opened 2021-09-10
[BUG] Followup comments from PR3411


tgravescs:
**Describe the bug**
there were a few nits in https://github.com/NVIDIA/spark-rapids/pull/3411 that need to be followed up on.

1) the shade plugin outputs dependency-reduced-pom.xml under the pom's basedir by default since we were not using dependencyReducedPomLocation config. As a consequence many developers who have used the old build in their local repo prior to this PR will have a stale file under dist that is not cleaned up by maven
I suggest to add it to the exclude list in this PR to minimize reported build failures.
In a follow up we can add dependencyReducedPomLocation to the shade plugin configs

2) see other comments from Bobby.

tgravescs:
dependency-reduced-pom.xml  was handled under https://github.com/NVIDIA/spark-rapids/commit/906af85e36047a54bee7147dcc2152d76ecfcdf9

PR #3448, opened 2021-09-10
Update location of dependency reduced pom

tgravescs:
Put the dependency reduced pom under the target directory from review comments.

fixes https://github.com/NVIDIA/spark-rapids/issues/3440


tgravescs:
build

revans2 (approved):

tgravescs:
that is very weird, rat is now finding files in target directory, investigating

revans2:
> that is very weird, rat is now finding files in target directory, investigating

From what I have seen rat automatically excludes target directories if there is a pom.xml in the parent directory that is a part of the current build. If it is excluded as a part of the build then rat sees the target directory as fair game.

tgravescs:
will revisit later if needed

PR #3453, opened 2021-09-10
Fix spark-tests script for classifier

tgravescs:
update to download spark301 version of integration test jar and remove the classifier from pytest.tar.gz. even though we build pytest.tar.gz with the classifier of spark301, when you do an install or deploy it installs without spark301.  So fix this script for now, we have another task to have integration tests use the correct version of spark jar as follow on.

I used test jenkins build and it downloaded everything properly and started to run the tests, I didn't wait for it to finish.


jlowe (approved):

tgravescs:
again here premerge doesn't test this script so I'm going to merge early and kick integration tests again.

PR #3454, opened 2021-09-10
Catch UDF compiler exceptions and fallback to CPU

### LINKED ISSUE #3404: [FEA] UDF compiler crashes should not prevent query from running
#3369 is an example of a crash in the UDF compiler that prevented the query from running.  Since we can fallback to the CPU for UDFs, it seems unnecessary to fail the entire query if a particular UDF causes a crash in the UDF compiler.

The plugin should catch exceptions thrown by the UDF compiler during planning and have the UDF fallback to the CPU if the UDF compiler throws an exception processing the UDF.

jlowe:
Fixes #3404 

Updates the UDF compiler to catch other types of exceptions besides the typical `SparkException` used to indicate a translation failure and fallback to the CPU in case of any exception.  Exceptions other than `SparkException` are logged as a warning.

### DIFF
--- a/udf-compiler/src/main/scala/com/nvidia/spark/udf/GpuScalaUDF.scala
+++ b/udf-compiler/src/main/scala/com/nvidia/spark/udf/GpuScalaUDF.scala
@@ -16,6 +16,8 @@
 
 package com.nvidia.spark.udf
 
+import scala.util.control.NonFatal
+
 import com.nvidia.spark.rapids.shims.v2.ShimExpression
 
 import org.apache.spark.SparkException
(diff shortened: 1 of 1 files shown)


jlowe:
build

udf-compiler/src/main/scala/com/nvidia/spark/udf/GpuScalaUDF.scala
```diff
@@ -50,11 +50,16 @@ case class GpuScalaUDFLogical(udf: ScalaUDF) extends ShimExpression with Logging
...
     } catch {
       case e: SparkException =>
-        logDebug("UDF compilation failure: " + e)
+        val udfName = udf.udfName.getOrElse("<unknown>")
+        logDebug(s"UDF $udfName compilation failure: $e")
         if (isTestEnabled) {
           throw e
         }
         udf
+      case e: Exception =>
```
gerashegalov, line 59: prefer the idiom `case NonFatal(e) =>`

jlowe:
build

gerashegalov (approved):

pxLi:
build

pxLi:
build

Issue #3415, opened 2021-09-08
[BUG] Fix regressions in WindowFunctionSuite with Spark 3.2.0


andygrove:
**Describe the bug**

Spark 3.2.0 has a new DayTimeIntervalType which we do not support. See apache/spark@cd649e7

This is causing some test failures in WindowFunctionSuite

```
WITH DECIMALS: [Window] [RANGE] [ ASC] [- [ ] 2 DAYS, 3 DAYS] *** FAILED ***
WITH DECIMALS: [Window] [RANGE] [DESC] [- [ ] 2 DAYS, 3 DAYS] *** FAILED ***
WITH DECIMALS: [Window] [RANGE] [ ASC] [- [ ] 2 DAYS, CURRENT ROW] *** FAILED ***
WITH DECIMALS: [Window] [RANGE] [DESC] [- [ ] 2 DAYS, CURRENT ROW] *** FAILED ***
WITH DECIMALS: [Window] [RANGE] [ ASC] [CURRENT ROW, 3 DAYS] *** FAILED ***
WITH DECIMALS: [Window] [RANGE] [DESC] [CURRENT ROW, 3 DAYS] *** FAILED ***
IGNORE ORDER, WITH DECIMALS: [Window] [MIXED WINDOW SPECS] *** FAILED ***
```

**Steps/Code to reproduce bug**
```
mvn -Dbuildver=320 verify
```

**Expected behavior**
Tests should pass

**Environment details (please complete the following information)**
N/A

**Additional context**
N/A

wbo4958:
I can repro the errors, and the explain log is like below,

``` console
*Exec <ProjectExec> will run on GPU
  !Exec <WindowExec> cannot run on GPU because not all expressions can be replaced
    @Expression <Alias> count(dollars#5) windowspecdefinition(uid#0L, _w1#21 ASC NULLS FIRST, specifiedwindowframe(RangeFrame, INTERVAL '-3' DAY, INTERVAL '2' DAY)) AS count_dollars#19L could run on GPU
      !Expression <WindowExpression> count(dollars#5) windowspecdefinition(uid#0L, _w1#21 ASC NULLS FIRST, specifiedwindowframe(RangeFrame, INTERVAL '-3' DAY, INTERVAL '2' DAY)) cannot run on GPU because the type of boundary is not supported in a window range function, found INTERVAL '2' DAY; the type of boundary is not supported in a window range function, found INTERVAL '-3' DAY
[...]

PR #3460, opened 2021-09-13
Add DayTimeIntervalType/YearMonthIntervalType support

wbo4958:
This PR tries to add DayTimeIntervalType and YearMonthIntervalType for TypeSig. To fix part of https://github.com/NVIDIA/spark-rapids/issues/3415

This PR introduces a new TypeSig320 which extends TypeSig, and overrides some methods which has data type checks.

The expressions which supports DayTimeIntervalType and YearMonthIntervalType need to be moved to Spark 320 shims to add  TypeSig.DAYTIME and TypeSig.YEARMONTH into the checks.

@revans2, Could you give me some comments on this solution?


sql-plugin/src/main/320/scala/com/nvidia/spark/rapids/shims/v2/TypeSig320.scala
```diff
@@ -0,0 +1,75 @@
...
+
+package com.nvidia.spark.rapids.shims.v2
+
+import ai.rapids.cudf.DType
+import com.nvidia.spark.rapids.{TypeEnum, TypeSig}
+
+import org.apache.spark.sql.types.{DataType, DayTimeIntervalType, YearMonthIntervalType}
+
+/** TypeSig for Spark 3.2.0+ which adds DayTimeIntervalType and YearMonthIntervalType support */
+final class TypeSig320(
```
revans2, line 25: I see two possibilities on how to do this

1. We do something like this were we have a TypeSigBase in the common code and then an actual TypeSig implementation class in shim layers. The problem with this is `TypeSig.all` and the other Spark helper values will not reflect "all" of the types. To do that we would need to move all or part of object TypeSig into the shim layer, or we would have to have the overrides for each of the things that touch `TypeSig.all` etc go into the shim layer.
2. We turn it around and have a static TypeSigUtil object in the shim layer that TypeSig calls into to do checks with new types.

For me I am  leaning towards the second approach because it is consistent between the static and non-static code.

wbo4958: Thx @revans2, I just had a new commit for the second approach. Could you help to take a look at it? Thx

wbo4958:
build

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/Spark30XShims.scala
```diff
@@ -93,4 +93,6 @@ trait Spark30XShims extends SparkShims {
   }
 
   override def shouldFailDivOverflow(): Boolean = false
+
+  override def getTypeSigUtil(): TypeSigUtil = TypeSigUtilUntil320
```
revans2, line 97: We don't need anything in the Shim layer to make this happen. You can just have a single class named TypeSigUtil. Then you have different versions in the appropriate directories.

sql-plugin/src/main/320/scala/com/nvidia/spark/rapids/shims/v2/TypeSigUtilFrom320.scala
```diff
@@ -0,0 +1,78 @@
...
+    dataType match {
+      case _: DayTimeIntervalType =>
+        if (check.contains(TypeEnum.DAYTIME)) Seq.empty else notSupportedReason
+      case _: YearMonthIntervalType =>
+        if (check.contains(TypeEnum.YEARMONTH)) Seq.empty else notSupportedReason
+      case _ => notSupportedReason
+    }
+  }
+
+  override def getCastChecksAndSigs(
```
revans2, line 55: I think I would prefer it if we found a way to put `daytimeChecks`, `sparkDaytimeSig` `yearmonthChecks` and `sparkYearmonthSig` inside of the `CastChecks` class.  This is just for consistency with where the rest of the checks are, and how they are overridden.

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/Spark30XShims.scala
wbo4958: Yeah, You're correct. Thx. Done

sql-plugin/src/main/320/scala/com/nvidia/spark/rapids/shims/v2/TypeSigUtilFrom320.scala
wbo4958: Done. I just map the dataType to TypeEnum and invoke the existing function which accepts TypeEnum can resolve this issue.

wbo4958:
build

wbo4958:
build

revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -115,6 +157,8 @@ object TypeEnum extends Enumeration {
   val MAP: Value = Value
   val STRUCT: Value = Value
   val UDT: Value = Value
+  val DAYTIME: Value = Value
+  val YEARMONTH: Value = Value
 }
```
res-life, line 162: Can we util ShimLoader.getSparkVersion to construct static version to set(TypeEnum) map?
Not know if feasible, if ok, just not need the spark 320 code.

revans2: Adding more enums values dynamically is not something that scala supports as far as I can tell.

PR #3464, opened 2021-09-13
Add test for empty relation propigation

### LINKED ISSUE #2763: A deeper look into SPARK-35585
https://github.com/apache/spark/commit/daf9d198dc changed some of the optimization rules in AQE.

I don't think that they will impact us.  They deal with propagating an  empty relation through project/filter commands.  But I would like an AQE expert to take a look at it.

revans2:
This fixes #2763 

This does not impact us directly so we are good. I am putting up the test as a PR, but I am fine if we just close this and #2763 so as to not add too many tests to the build.


revans2:
build

jlowe (approved):

andygrove (approved):

PR #3467, opened 2021-09-13
Update docs to better describe support for floating point aggregation and NaNs

jlowe:
Adds partial support notes for aggregation expressions that have issues with NaNs and require `spark.rapids.sql.hasNans` to be false before floats and doubles can be supported.


jlowe:
build

revans2 (approved):

pxLi:
build

PR #3473, opened 2021-09-14
Create non-shim specific version of ParquetCachedBatchSerializer

tgravescs:
This adds upon https://github.com/NVIDIA/spark-rapids/pull/3390 (credit to @razajafri as well) to create a common user facing generic class for the ParquetCachedBatchSerializer that then loads the shim specific version of it as necessary.

fixes https://github.com/NVIDIA/spark-rapids/issues/3314

User just has to set:
--conf spark.sql.cache.serializer=com.nvidia.spark.ParquetCachedBatchSerializer

And that should handle loading the shim. Right now there is only one version underneath the 311+-all directory.  

This includes the spark 3.1.1 version of  com.nvidia.spark.ParquetCachedBatchSerializer into the base of the jar to be loaded after that the shim version load will be the one under the spark specific directory. spark312/, spark311, etc..

I did not add the function to ShimLoader (like newDriverPlugin) because the code is in the spark311+-all directory since the CachedBatchSerializer isn't available until spark 3.1.0 

Fixed the databricks 8.2 build and ran tests.

Ran tests on spark 3.1.1, 3.1.2, 3.2.0 and databricks 8.2

I was not able to validate the test script updates.


tgravescs:
build

sql-plugin/src/main/311+-all/scala/com/nvidia/spark/ParquetCachedBatchSerializer.scala
```diff
@@ -0,0 +1,195 @@
...
+   * the host
+   *
+   * @param input              the cached batches that should be converted.
+   * @param cacheAttributes    the attributes of the data in the batch.
+   * @param selectedAttributes the fields that should be loaded from the data and the order they
+   *                           should appear in the output batch.
+   * @param conf               the configuration for the job.
+   * @return an RDD of the input cached batches transformed into the ColumnarBatch format.
+   */
+  override def gpuConvertCachedBatchToColumnarBatch(
```
tgravescs, line 187: this is the only function not in the Spark CachedBatchSerializer

tgravescs:
build

tgravescs:
build

sql-plugin/src/main/311+-all/scala/com/nvidia/spark/ParquetCachedBatchSerializer.scala
```diff
@@ -0,0 +1,197 @@
...
+  // Note that since the config to set the serializer wasn't added until
+  // Spark 3.1.0 (https://issues.apache.org/jira/browse/SPARK-32274) this shouldn't
+  // ever throw.
+  if (sparkVersion < minSupportedVer) {
+    throw new IllegalArgumentException("ParquetCachedBaatchSerializer only supported for Spark " +
+      s"versions > 3.1.1, version found was: $sparkVersion")
+  }
+
+  private lazy val realSerializer: GpuCachedBatchSerializer = {
+    ShimLoader.newInstanceOf("com.nvidia.spark.rapids.shims.v2.ParquetCachedBatchSerializer")
```
revans2, line 55: Can we please move the reflection to the Shim loader? For me it makes it simpler to have all of the reflection in a single place so we know exactly how much is being used and what it is used for.

tgravescs: I mentioned this in the description, we can't because the class only existing in spark 3.1.1 and we have to pull a common version of Shimloader which is 3.0.1

tgravescs: By class I mean CachedBatchSErializer which comes from spark

tgravescs: or we could create a Shimloader version for 311+

revans2: That stinks, but I understand why.

revans2 (approved):

PR #3477, opened 2021-09-14
Update supported docs to clarify casting floating point to string

jlowe:
Cast documentation states casting from a floating point number to a string "just works," but this is actually disabled by default and must be enabled by a config.  Updated the documentation to better reflect the level of support.


jlowe:
build

revans2 (approved):

jlowe:
build

gerashegalov (approved):

PR #3479, opened 2021-09-14
Test with Spark 3.2.1-SNAPSHOT

andygrove:
Change spark320 version to 3.2.1-SNAPSHOT so we can test Spark 3.2 support in CI. This is necessary as a temporary measure until 3.2.0 is released because Spark no longer publishes 3.2.0-SNAPSHOT releases and we depend on a fix that went in since they stopped publishing them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -632,7 +632,7 @@
         <spark311cdh.version>3.1.1.3.1.7270.0-253</spark311cdh.version>
         <spark312.version>3.1.2</spark312.version>
         <spark313.version>3.1.3-SNAPSHOT</spark313.version>
-        <spark320.version>3.2.0-SNAPSHOT</spark320.version>
+        <spark320.version>3.2.1-SNAPSHOT</spark320.version>
         <mockito.version>3.6.0</mockito.version>
         <scala.plugin.version>4.3.0</scala.plugin.version>
         <maven.jar.plugin.version>3.2.0</maven.jar.plugin.version>
--- a/shims/spark320/src/main/scala/com/nvidia/spark/rapids/shims/spark320/SparkShimServiceProvider.scala
+++ b/shims/spark320/src/main/scala/com/nvidia/spark/rapids/shims/spark320/SparkShimServiceProvider.scala
@@ -20,8 +20,9 @@ import com.nvidia.spark.rapids.{SparkShims, SparkShimVersion}
 
 object SparkShimServiceProvider {
   // temporarily allow 3.2.1 while 3.2.0 release candidates are being produced
-  val VERSION320 = SparkShimVersion(3, 2, 1)
-  val VERSIONNAMES: Seq[String] = Seq(VERSION320)
+  val VERSION320 = SparkShimVersion(3, 2, 0)
+  val VERSION321 = SparkShimVersion(3, 2, 1)
+  val VERSIONNAMES: Seq[String] = Seq(VERSION320, VERSION321)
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -632,7 +632,7 @@
         <spark311cdh.version>3.1.1.3.1.7270.0-253</spark311cdh.version>
         <spark312.version>3.1.2</spark312.version>
         <spark313.version>3.1.3-SNAPSHOT</spark313.version>
-        <spark320.version>3.2.0-SNAPSHOT</spark320.version>
+        <spark320.version>3.2.1-SNAPSHOT</spark320.version>
```
jlowe, line 635: The Spark320 shim asserts that it sees version 3.2.0.  Don't we need to (temporarily) update the shim to also expect 3.2.1-SNAPSHOT?

tgravescs: yes

tgravescs (commented):
You need this too:

```
+++ b/shims/spark320/src/main/scala/com/nvidia/spark/rapids/shims/spark320/SparkShimServiceProvider.scala
@@ -20,7 +20,9 @@ import com.nvidia.spark.rapids.{SparkShims, SparkShimVersion}

 object SparkShimServiceProvider {
   val VERSION320 = SparkShimVersion(3, 2, 0)
-  val VERSIONNAMES: Seq[String] = Seq(VERSION320)
+  // temporarily allow 3.2.1 while 3.2.0 release candidates are being produced
+  val VERSION321 = SparkShimVersion(3, 2, 1)
+  val VERSIONNAMES: Seq[String] = Seq(VERSION320, VERSION321)
     .flatMap(v => Seq(s"$v", s"$v-SNAPSHOT"))
 }
```

shims/spark320/src/main/scala/com/nvidia/spark/rapids/shims/spark320/SparkShimServiceProvider.scala
```diff
@@ -19,7 +19,8 @@ package com.nvidia.spark.rapids.shims.spark320
 import com.nvidia.spark.rapids.{SparkShims, SparkShimVersion}
 
 object SparkShimServiceProvider {
-  val VERSION320 = SparkShimVersion(3, 2, 0)
+  // temporarily allow 3.2.1 while 3.2.0 release candidates are being produced
+  val VERSION320 = SparkShimVersion(3, 2, 1)
   val VERSIONNAMES: Seq[String] = Seq(VERSION320)
```
tgravescs, line 24: referencing version320 which was removed. Also I think we should leave 320 in there in case someone is using older snapshot.

abellina:
I tested with this diff and I was able to test shuffle with the 3.2.0 shim against spark 3.2.1.

shims/spark320/src/main/scala/com/nvidia/spark/rapids/shims/spark320/SparkShimServiceProvider.scala
tgravescs: never mind on the name, realize VERSION320 just points to 3.2.1.  I do think we should leave 3.2.0 as well though

andygrove: @tgravescs I have updated this now

tgravescs (approved):

tgravescs:
build

Issue #3478, opened 2021-09-14
[BUG] CastOpSuite and ParseDateTimeSuite failures spark 302 and others


tgravescs:
**Describe the bug**
running unit tests against spark 3.0.2 and the CastOpSuite has some failures:

- Cast from string to date using random inputs *** FAILED ***
  Mismatch casting string [12\rZ] to DateType. CPU: null; GPU: 0012-01-01 (CastOpSuite.scala:247)
- Cast from string to date using random inputs with valid year prefix *** FAILED ***
  Mismatch casting string [2021-9:] to DateType. CPU: null; GPU: 2021-09-01 (CastOpSuite.scala:247)

- Test all supported casts with in-range values *** FAILED ***
  Cast from StringType to DateType failed; ansi=false org.scalatest.exceptions.TestFaile

tgravescs:
there are also failures in ParseDateTimeSuite:

- WITH DECIMALS: to_date yyyy-MM-dd *** FAILED ***
  Mismatch between CPU and GPU for the following rows:

  [- WITH DECIMALS: to_timestamp yyyy-MM-dd *** FAILED ***

tgravescs:
looks like this fails on spark 3.0.1 as well

abellina:
Just ran into this with `mvn verify`

tgravescs:
something very weird because rolled back to yesterday and they still fail there

andygrove:
perhaps the random data generation changes somehow and we are hitting test cases we haven't seen before? I'll take a look

andygrove:
some of these failing tests do not use random data generation, so seems like something more build related somehow

tgravescs:
its a change in cudf that causes this, I tried an older version and it works there.

Perhaps https://github.com/rapidsai/cudf/pull/9173  this

revans2:
It looks like https://github.com/rapidsai/cudf/pull/9173 when into cudf and broke things. I am going to try and update the casing code to deal with this. It gives us some things that we want, and some things that we don't want all at once.

revans2:
If someone else wants to disable those tests in the short term to unblock others I am fine with that.

tgravescs:
working on disabling tests temporarily

revans2:
Moved to @andygrove because he has been working on a fix and is further along than I was

andygrove:
[...]

PR #3485, opened 2021-09-14
Fix regressions in cast from string to date and timestamp

andygrove:

Partially addresses https://github.com/NVIDIA/spark-rapids/issues/3478

cuDF recently started supporting variable-length date/timestamp components in `is_timestamp` and when parsing strings as timestamps. This caused regressions in the plugin.

This PR fixes the regressions and has the following changes:

- Removes code that modified strings to replace single digit month and day with two digit versions
- Removes methods based on assumptions around fixed-lengh parsing in cuDF
- Updates regex patterns to accept one or two-digit month, day, hour, minute, second components
- Enables a test for casting string to timestamp since the only existing tests were ignored due to other documented issues

We also now support single digit hour, minute and second in timestamps. We still require 6 digits for fractional second.


andygrove:
build

tests/src/test/scala/com/nvidia/spark/rapids/CastOpSuite.scala
```diff
@@ -960,30 +959,6 @@ class CastOpSuite extends GpuExpressionTestSuite {
     }
   }
 
-  test("CAST string to date - sanitize step") {
```
andygrove, line 963: This test is no longer needed since we no longer have the sanitization step

revans2 (approved):

revans2:
build

revans2:
build

revans2:
The last build failure looked like a networking issue (could not fetch from github...)

Issue #3468, opened 2021-09-13
[BUG] GpuKryoRegistrator ClassNotFoundException  


gerashegalov:
**Describe the bug**
Recent build changes accidentally move GpuKryoRegistrator in per-shim `/spark3xx` jar locations. It must remain unshimmed per our doc. There is no test using GpuKryoRegistrator to catch this regression

**Steps/Code to reproduce bug**
Start  a Spark application with `--conf spark.kryo.registrator=com.nvidia.spark.rapids.GpuKryoRegistrator`
reported by @rongou 

**Expected behavior**
No CNF 

**Environment details (please complete the following information)**
 local

**Additional context**
https://github.com/NVIDIA/spark-rapids/blob/960d52d1e3848b269d1171129a1cbca070dd9904/dist/README.md

PR #3489, opened 2021-09-15
Fix 311 databricks shim for AnsiCastOpSuite failures

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/3488

This fixes the databricks 3.1.1 build.  The only tests I saw there were unit tests which we don't run on databricks so I'm running the integration tests now.

Signed-off-by: Thomas Graves <tgraves@nvidia.com>

PR #3491, opened 2021-09-15
Shim GpuKryoRegistrator

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/3468

Add the GpuKryoRegistrator class to common unshimmed code and add the actual registration into the shim.  This made more sense to me to add into the shim module since its not create a new instance but just references the class by the spark version shim and I expect this code to be commonized when we commonize the base shim code.  If people have strong objections we can rework it but we haven't really put anything into the ShimLoader class like this yet.

I built all versions, tested on all apache versions. We need overrides to test cdh and need to figure out how to test on Databricks.  

### DIFF
--- a/dist/unshimmed-base.txt
+++ b/dist/unshimmed-base.txt
@@ -7,6 +7,7 @@ com/nvidia/spark/RapidsUDF*
 com/nvidia/spark/SQLPlugin*
 com/nvidia/spark/rapids/ExecutionPlanCaptureCallback*
 com/nvidia/spark/rapids/PlanUtils*
+com/nvidia/spark/rapids/GpuKryoRegistrator*
 com/nvidia/spark/rapids/RapidsExecutorHeartbeatMsg*
 com/nvidia/spark/rapids/RapidsExecutorStartupMsg*
 com/nvidia/spark/rapids/RapidsExecutorUpdateMsg*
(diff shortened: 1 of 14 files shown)


tgravescs:
this will need to be upmerged to https://github.com/NVIDIA/spark-rapids/pull/3489

tgravescs:
build

gerashegalov:
If it's necessary I am ok with it 

in general, we should  start biasing much more to not adding  to Shims if it's avoidable imo   
 
Did we check if this fix changes anything about running in REPL? https://github.com/NVIDIA/spark-rapids/issues/3482#issuecomment-919643406

GpuKryoRegistrator may be safe to leave in the standard location unless it transitively references some incompatible Scala classes. I think mostly cudf java which is fine and Spark which is fine.  GpuKryoRegistrator bytecode is identical across shims:

```shell
$ diff -s ./spark311cdh/com/nvidia/spark/rapids/GpuKryoRegistrator.class ./spark302/com/nvidia/spark/rapids/GpuKryoRegistrator.class
Files ./spark311cdh/com/nvidia/spark/rapids/GpuKryoRegistrator.class and ./spark302/com/nvidia/spark/rapids/GpuKryoRegistrator.class are identical

$ diff -s ./spark311cdh/com/nvidia/spark/rapids/GpuKryoRegistrator.class ./spark320/com/nvidia/spark/rapids/GpuKryoRegistrator.class
Files ./spark311cdh/com/nvidia/spark/rapids/GpuKryoRegistrator.class and ./spark320/com/nvidia/spark/rapids/GpuKryoRegistrator.class are identical
```

tgravescs:
yes I tested via spark-shell.

revans2 (approved):

Issue #3450, opened 2021-09-10
[FEA] test the UCX shuffle with the new build changes


abellina:
https://github.com/NVIDIA/spark-rapids/issues/3223 introduces a whole new way to build/package our plugin. The RAPIDS Shuffle Manage needs to be tested so we make sure it works with all versions of  Spark with all the diffs performed.

We also want to add here that as of the writing of this issue, the RAPIDS Shuffle Manager is lazily initialized in the driver. A test in the shell shows that it gets initialized when it is needed to register some shuffle, and we output the following at the worst time:

```
2021-09-09 04:04:08,563 WARN  RapidsShuffleInternalManager:69 - Rapids Shuffle Plugin enabled. Transport enabled (remote fetches will use com.nvidia.spark.rapids.shuffle.ucx.UCXShuffleTransport. To disable the RAPIDS Shuffle Manager set `spark.rapids.shuffle.enabled` to false
```

I'll add a change to make this happen earlier during DriverPlugin initialization also

abellina:
So far so good. I've tested with: 3.0.1, 3.0.2, 3.0.3, 3.0.4, 3.1.1, 3.1.2, 3.1.3, 3.2.0, and databricks 7.3 and 8.2. The last one that is remaining is CDH. 

I tested by running a shell and running a simple query, and verifying that the shuffle manager loads. One small issue I had seen before is handled by https://github.com/NVIDIA/spark-rapids/pull/3493, but otherwise this seems to be working fine.

abellina:
I am able to run in CDH as well. I don't see any issues there loading the shuffle manager, and using it to fetch shuffle blocks.

PR #3493, opened 2021-09-15
Initialze RAPIDS Shuffle Manager at driver/executor startup

abellina:
Related to https://github.com/NVIDIA/spark-rapids/issues/3450. 

Given changes added to support Spark 3.2.0, a proxy system was put in place wrapping the RAPIDS Shuffle Manager. The proxy class has a `self` property that it uses to refer to the actual shuffle manager, the one that instantiates the transport. This all worked where the manager gets instantiated (for real) when the first shuffle is registered, or when `self` is used with early startup in the driver, but it was not convenient since it didn't print a message meant for a user to be able to turn off the shuffle, and also the executor didn't allocate memory until the first shuffle was registered, which it could have been doing all along.

This PR makes the initialization explicit, and hooks it in the driver/executor after we decide that the RAPIDS Shuffle Manager is configured.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuShuffleEnv.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuShuffleEnv.scala
@@ -84,6 +84,19 @@ object GpuShuffleEnv extends Logging {
     SparkEnv.get.blockManager.externalShuffleServiceEnabled
   }
 
+  //
+  // The actual instantiation of the RAPIDS Shuffle Manager is lazy, and
+  // this forces the initialization when we know we are ready in the driver and executor.
+  //
+  def initShuffleManager(): Unit = {
+    SparkEnv.get.shuffleManager match {
+      case visibleMgr: VisibleShuffleManager=>
+        visibleMgr.initialize
+      case _ =>
+        throw new IllegalStateException(s"Cannot initialize the RAPIDS Shuffle Manager")
+    }
+  }
+
   def isRapidsShuffleAvailable: Boolean = {
     // the driver has `mgr` defined when this is checked
     val sparkEnv = SparkEnv.get
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuShuffleEnv.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuShuffleEnv.scala
@@ -70,10 +70,14 @@ object GpuShuffleEnv extends Logging {
 
   def shutdown() = {
     // check for nulls in tests
-    val shuffleManager = Option(SparkEnv.get)
+    Option(SparkEnv.get)
       .map(_.shuffleManager)
       .collect { case sm: VisibleShuffleManager => sm }
       .foreach(_.stop())
+
+    // when we shut down, make sure we clear `env`, as the convention is that
+    // `GpuShuffleEnv.init` will be called to re-establish it
+    env = null
   }
 
   //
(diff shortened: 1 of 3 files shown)


abellina:
build

tgravescs (approved):

tgravescs:
10:42:58  stderr: fatal: unable to access 'https://github.com/NVIDIA/spark-rapids.git/': The requested URL returned error: 429

tgravescs:
build

jlowe (approved):
Can you elaborate a bit more in the PR description what this is doing?  Helps the software archeologists do a bit less digging later. 😄

abellina:
> Can you elaborate a bit more in the PR description what this is doing? Helps the software archeologists do a bit less digging later.

Sorry, I thought (in my brain) I had written more, but I actually hadn't. Thanks for pointing it out.

gerashegalov (dismissed):

abellina:
There's a relevant test failure:
```
11:02:09  - Avoid transition to GPU for trivial projection after CPU SMJ *** FAILED ***
11:02:09    java.lang.IllegalStateException: Cannot initialize the RAPIDS Shuffle Manager
11:02:09    at org.apache.spark.sql.rapids.GpuShuffleEnv$.initShuffleManager(GpuShuffleEnv.scala:96)
11:02:09    at com.nvidia.spark.rapids.RapidsDriverPlugin.init(Plugin.scala:156)
11:02:09    at org.apache.spark.internal.plugin.DriverPluginContainer.$anonfun$driverPlugins$1(PluginContainer.scala:50)
11:02:09    at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:244)
11:02:09    at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
11:02:09    at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
11:02:09    at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
11:02:09    at scala.collection.TraversableLike.flatMap(TraversableLike.scala:244)
11:02:09    at scala.collection.TraversableLike.flatMap$(TraversableLike.scala:241)
11:02:09    at scala.collection.AbstractTraversable.flatMap(Traversable.scala:108)
```

Looking into it.

gerashegalov:
Try to avoid using GpuShuffleEnv, it's been quite difficult to work with in the prototype.

abellina:
Yes @gerashegalov, that's right. The lazy vals in `GpuShuffleEnv` are the root of this, where test X configures the shuffle manager, but test Y doesn't want that enabled.

Actually this is a latent bug... I added an exception here, which is exposing the issue. Looking into a fix.

abellina:
@gerashegalov I restored a step that was removed with the shim rework for 3.2, where `shutdown` in `GpuShuffleEnv` would null out `env`. Please take another look when you get a chance.

abellina:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuShuffleEnv.scala
```diff
@@ -84,6 +88,19 @@ object GpuShuffleEnv extends Logging {
     SparkEnv.get.blockManager.externalShuffleServiceEnabled
   }
 
+  //
+  // The actual instantiation of the RAPIDS Shuffle Manager is lazy, and
+  // this forces the initialization when we know we are ready in the driver and executor.
+  //
+  def initShuffleManager(): Unit = {
+    SparkEnv.get.shuffleManager match {
+      case visibleMgr: VisibleShuffleManager=>
```
jlowe, line 97:
```suggestion
      case visibleMgr: VisibleShuffleManager =>
```

jlowe (approved):
Small whitespace nit but otherwise lgtm.

abellina:
build

jlowe (approved):

PR #3500, opened 2021-09-15
Fix GpuSum type to match resultType

abellina:
This fixes an issue that @andygrove had found in Spark 3.2.0. The explicit casting was removed as part of https://github.com/NVIDIA/spark-rapids/pull/3417, and I am adding it here in the `preUpdate` step. Some of this could be folded into the input projection, as done with `GpuAverage`, but for now this works and should unblock those testing with 3.2.0.

The binding is odd and we are cleaning up as part of https://github.com/NVIDIA/spark-rapids/issues/3194


andygrove (approved):
I confirmed that merging this to branch-21.10 fixes the regressions when run with Spark 3.2

abellina:
build

jlowe (approved):

jlowe:
build

pxLi:
build

pxLi:
build

pxLi:
build

abellina:
Thanks @pxLi !

PR #3501, opened 2021-09-15
Add limitations for Databricks doc

viadea:
Add limitations for databricks.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -267,7 +267,7 @@ In the 0.2 release, AQE is supported but all exchanges will default to the CPU. 
 release, running on Spark 3.0.1 and higher any operation that is supported on GPU will now stay on 
 the GPU when AQE is enabled. 
 
-AQE is not supported on databricks with the plugin.
+AQE is not supported on Databricks with the plugin.
 
 #### Why does my query show as not on the GPU when Adaptive Query Execution is enabled?
 
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -259,11 +259,15 @@ efficient to stay on the CPU instead of going back and forth.
 
 Yes, DPP still works.  It might not be as efficient as it could be, and we are working to improve it.
 
+DPP is not supported on databricks with the plugin.
+
 ### Is Adaptive Query Execution (AQE) Supported?
 
 In the 0.2 release, AQE is supported but all exchanges will default to the CPU.  As of the 0.3 
 release, running on Spark 3.0.1 and higher any operation that is supported on GPU will now stay on 
 the GPU when AQE is enabled. 
+
+AQE is not supported on databricks with the plugin.
 
 #### Why does my query show as not on the GPU when Adaptive Query Execution is enabled?
 
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -259,7 +259,7 @@ efficient to stay on the CPU instead of going back and forth.
 
 Yes, DPP still works.  It might not be as efficient as it could be, and we are working to improve it.
 
-DPP is not supported on databricks with the plugin.
+DPP is not supported on Databricks with the plugin.
 
 ### Is Adaptive Query Execution (AQE) Supported?
 
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -260,6 +260,7 @@ efficient to stay on the CPU instead of going back and forth.
 Yes, DPP still works.  It might not be as efficient as it could be, and we are working to improve it.
 
 DPP is not supported on Databricks with the plugin.
+Queries on Databricks will not fail but it can not benefit from DPP.
 
 ### Is Adaptive Query Execution (AQE) Supported?
 
@@ -267,7 +268,8 @@ In the 0.2 release, AQE is supported but all exchanges will default to the CPU. 
 release, running on Spark 3.0.1 and higher any operation that is supported on GPU will now stay on 
 the GPU when AQE is enabled. 
 
-AQE is not supported on Databricks with the plugin.
+AQE is not supported on Databricks with the plugin. 
+If AQE is enabled on Databricks, queries may fail with `StackOverflowError` error.
 
 #### Why does my query show as not on the GPU when Adaptive Query Execution is enabled?
 
--- a/docs/get-started/getting-started-databricks.md
+++ b/docs/get-started/getting-started-databricks.md
@@ -157,30 +157,32 @@ option before starting the cluster to capture the logs.
 
 ## Limitations
 
-1. Adaptive query execution(AQE) and Delta optimization write do not work. 
-[issue-1059](https://github.com/NVIDIA/spark-rapids/issues/1059)
+1. Adaptive query execution(AQE) and Delta optimization write do not work.  These should be disabled
+when using the plugin. See [issue-1059](https://github.com/NVIDIA/spark-rapids/issues/1059) for more
+detail. 
 
-```bash
-spark.databricks.delta.optimizeWrite.enabled false
-spark.sql.adaptive.enabled false
-```
+    ```bash 
+    spark.databricks.delta.optimizeWrite.enabled false
+    spark.sql.adaptive.enabled false
+    ```
 
-2. Dynamic partition pruning(DPP) does not work.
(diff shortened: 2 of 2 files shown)


docs/get-started/getting-started-databricks.md
```diff
@@ -155,3 +155,32 @@ storage location using Databricks [cluster log
 delivery](https://docs.databricks.com/clusters/configure.html#cluster-log-delivery-1).  Enable this
 option before starting the cluster to capture the logs.
 
+## Limitations
```
tgravescs, line 158: I think it would be good to add something at the very beginning of the getting started on data bricks doc rather than here.  We also have other text in here that mentioned AQE .  Perhaps also put in FAQ.  I think @jlowe  had some other ideas.

docs/get-started/getting-started-databricks.md
```diff
@@ -155,3 +155,32 @@ storage location using Databricks [cluster log
...
+Databricks overrides the `spark.executor.resource.gpu.amount=1` even though 
+users set in Spark Configuration tab to `spark.executor.resource.gpu.amount=N` 
+where N is the number of GPUs per node. 
+This will result in failed executors when starting the cluster.
+
+5. Databricks makes changes to the runtime without notification.
+
+[issue-3098](https://github.com/NVIDIA/spark-rapids/issues/3098) is one example 
+and we have to fix it in our plugin once those types of issues are detected.
+
```
sameerz, line 186: Lists require paragraphs to be indented for the list to continue.  Made some suggestions on the text. 

```suggestion
1. Adaptive query execution(AQE) and Delta optimization write do not work.  These should be disabled
when using the plugin. See [issue-1059](https://github.com/NVIDIA/spark-rapids/issues/1059) for more
detail. 

    ```bash 
    spark.databricks.delta.optimizeWrite.enabled false
    spark.sql.adaptive.enabled false
    ```

2. Dynamic partition pruning(DPP) does not work.  This results in poor performance for queries which
   would normally benefit from DPP.  See
   [issue-3143](https://github.com/NVIDIA/spark-rapids/issues/3143) for more detail.

3. When selecting GPU nodes, Databricks requires the driver node to be a GPU node.  Outside of
   Databricks the plugin can operate with the driver as a CPU node and workers as GPU nodes.

4. Cannot spin off multiple executors on a multi-GPU node. 

	Even though it is possible to set `spark.executor.resource.gpu.amount=N` (where N is the number
    of GPUs per node) in the in Spark Configuration tab, Databricks overrides this to
    `spark.executor.resource.gpu.amount=1`.  This will result in failed executors when starting the
    cluster.

5. Databricks makes changes to the runtime without notification.

    Databricks makes changes to existing runtimes, applying patches, without notification.
	[Issue-3098](https://github.com/NVIDIA/spark-rapids/issues/3098) is one example of this.  We run
	regular integration tests on the Databricks environment to catch these issues and fix them once
	detected.
```

docs/get-started/getting-started-databricks.md
jlowe: Agree this shouldn't be hidden at the end of the document.  I also think the [FAQ entry for DPP](https://github.com/NVIDIA/spark-rapids/blob/branch-21.10/docs/FAQ.md#is-dynamic-partition-pruning-dpp-supported) and the [FAQ entry for AQE](https://github.com/NVIDIA/spark-rapids/blob/branch-21.10/docs/FAQ.md#is-adaptive-query-execution-aqe-supported) should be updated.

docs/get-started/getting-started-databricks.md
```diff
@@ -155,3 +155,32 @@ storage location using Databricks [cluster log
 delivery](https://docs.databricks.com/clusters/configure.html#cluster-log-delivery-1).  Enable this
 option before starting the cluster to capture the logs.
 
+## Limitations
+
+1. Adaptive query execution(AQE) and Delta optimization write do not work.
```
jlowe, line 160: I think it's important to emphasize here that queries may still see significant speedups even with AQE disabled.  Some users may see this and think, "well I run with AQE on, so this is a non-starter" which is not necessarily the case.

viadea: @jlowe Just added one line for this. How about now?

docs/get-started/getting-started-databricks.md
viadea: @jlowe  @tgravescs  I just added lines to mention DPP/AQE is not supported on databricks in FAQ. How is now?

docs/FAQ.md
```diff
@@ -259,12 +259,16 @@ efficient to stay on the CPU instead of going back and forth.
 
 Yes, DPP still works.  It might not be as efficient as it could be, and we are working to improve it.
 
+DPP is not supported on databricks with the plugin.
```
sameerz, line 262:
```suggestion
DPP is not supported on Databricks with the plugin.
```

docs/FAQ.md
```diff
@@ -259,12 +259,16 @@ efficient to stay on the CPU instead of going back and forth.
...
+
 ### Is Adaptive Query Execution (AQE) Supported?
 
 In the 0.2 release, AQE is supported but all exchanges will default to the CPU.  As of the 0.3 
 release, running on Spark 3.0.1 and higher any operation that is supported on GPU will now stay on 
 the GPU when AQE is enabled. 
 
+AQE is not supported on databricks with the plugin.
```
sameerz, line 270:
```suggestion
AQE is not supported on Databricks with the plugin.
```

sameerz (changes requested):

docs/FAQ.md
```diff
@@ -259,12 +259,16 @@ efficient to stay on the CPU instead of going back and forth.
 
 Yes, DPP still works.  It might not be as efficient as it could be, and we are working to improve it.
 
+DPP is not supported on Databricks with the plugin.
```
jlowe, line 262: There needs to be more detail here.  Will a query crash if DPP is enabled or will it run to completion and correctly but with less performance?

docs/FAQ.md
```diff
@@ -259,12 +259,16 @@ efficient to stay on the CPU instead of going back and forth.
...
+
 ### Is Adaptive Query Execution (AQE) Supported?
 
 In the 0.2 release, AQE is supported but all exchanges will default to the CPU.  As of the 0.3 
 release, running on Spark 3.0.1 and higher any operation that is supported on GPU will now stay on 
 the GPU when AQE is enabled. 
 
+AQE is not supported on Databricks with the plugin.
```
jlowe, line 270: Similarly need a bit more detail on what "not supported" means.

docs/get-started/getting-started-databricks.md
jlowe: Thanks for the FAQ entries, but this section is still hidden at the end of the doc which was the original issue raised by this comment.

docs/get-started/getting-started-databricks.md
```diff
@@ -155,3 +155,36 @@ storage location using Databricks [cluster log
 delivery](https://docs.databricks.com/clusters/configure.html#cluster-log-delivery-1).  Enable this
 option before starting the cluster to capture the logs.
 
+## Limitations
+
+1. Adaptive query execution(AQE) and Delta optimization write do not work. These should be disabled
+when using the plugin. 
+Though queries may still see significant speedups even with AQE disabled.
```
jlowe, line 162:
```suggestion
when using the plugin. Queries may still see significant speedups even with AQE disabled.
```

jlowe (approved):

docs/FAQ.md
viadea: @jlowe I added the line saying DPP will not fail query but queries can not benefit.

docs/FAQ.md
viadea: @jlowe I added the line saying AQE may fail queries on Databricks.

docs/get-started/getting-started-databricks.md
viadea: @jlowe  @sameerz  I put limitation section to top. How do you think now?

sameerz (approved):

jlowe (approved):

sameerz:
build

PR #3503, opened 2021-09-15
Have average support nulls for 3.2.0

revans2:
This fixes some of the integration tests, but there are more to work on too.


revans2:
build

jlowe (approved):
Wondering if it would make sense to shim just the `TypeSig` for average so we don't need to copy the entire expr override code block in each shim.

revans2:
> Wondering if it would make sense to shim just the `TypeSig` for average so we don't need to copy the entire expr override code block in each shim.

I thought about this, but I decided to go this route because we might add in support for adding intervals at some point too.  Also if this is the only one that follows this pattern, then it is an odd one out, and I would rather keep it the same as the others until we come up with an appropriate set of patterns for how not to have all of these duplicated so much.

PR #3519, opened 2021-09-16
Shim version utils

andygrove:
This is just a small cleanup of shim version checking, with these changes:

- Move `cmpSparkVersion` from test suite to `ShimVersion` class so that we can use it in the plugin
- Update `ParquetCachedBatchSerializer` to use this mechanism instead of a version string comparison
- Remove all uses of `assumeSpark301orLater` now that 3.0.1 is the minimum supported version

### DIFF
--- a/sql-plugin/src/main/311+-all/scala/com/nvidia/spark/ParquetCachedBatchSerializer.scala
+++ b/sql-plugin/src/main/311+-all/scala/com/nvidia/spark/ParquetCachedBatchSerializer.scala
@@ -41,14 +41,14 @@ trait GpuCachedBatchSerializer extends CachedBatchSerializer {
  */
 class ParquetCachedBatchSerializer extends GpuCachedBatchSerializer {
 
-  val minSupportedVer = "3.1.1"
-  val sparkVersion = ShimLoader.getSparkVersion
+  val minSupportedVer = (3, 1, 1)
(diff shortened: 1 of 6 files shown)


revans2:
So in talking to @jlowe we have some issues with Databricks 8.3 it reports that it is compatible with Spark 3.1.1 but it is actually based mostly off of Spark 3.2.0. What is more Databricks 7.?? also reports Spark 3.1.1. I don't know if 8.3 has pulled in anything that is a user facing incompatibility.  He is still trying to figure out how to handle all of this. As such we need to be very careful with where we use version numbers for checking and where we do not.  I personally would rather hold off on this type of a change until we really understand what is happening.

andygrove:
Ok, I will close this for now.

gerashegalov:
I think the API should be 

```scala
sealed abstract class ShimVersion extends PartialOrdering[ShimVersion] {
  def tryCompare(that: ShimVersion)
}
```

DatabricksShimVersion should include the db runtime version similar to CDH but better 

```scala
case class DatabricksShimVersion(sparkMajor: Int, sparkFeature: Int, sparkPatch: Int, runtimeMajor: Int, runtimeMinor) extends ShimVersion
```

Vendor-version class X generally should only be comparable to X unless the vendor adds only orthogonal functionality to upstream Spark version in which case they can also be compared SparkShimVersion.

PR #3521, opened 2021-09-16
Qualification tool: Report nested complex types in Potential Problems and improve write csv identification.

nartal1:
This partially addresses https://github.com/NVIDIA/spark-rapids/issues/3329 . Out of 3 tasks in the issue, this PR addresses 2.
1. Report nested complex types in potential problems. Updated existing tests.
2. Removed the hardcode of csv write format. Instead we are checking if the format != TypeSig.none. Running mvn verify did not update supportedDataSource.csv. Hence it does the same functionality without hardcoding. And later if we support csv write then we don't have to update the code here.
3. Write format score is not reported as we might need further discussion on scoring in general.


nartal1:
build

tools/src/main/scala/org/apache/spark/sql/rapids/tool/qualification/QualAppInfo.scala
```diff
@@ -136,6 +136,20 @@ class QualAppInfo(
     probNotDataset.values.flatten.toSet.mkString(":")
   }
 
+  private def getAllPotentialProblems(getPotentialProb: String, nestedComplex: String): String = {
```
tgravescs, line 139: rename getPotentialProb to something useful, that is just a string to its not getting anything

tools/src/main/scala/org/apache/spark/sql/rapids/tool/qualification/QualAppInfo.scala
```diff
@@ -136,6 +136,20 @@ class QualAppInfo(
...
+    val nestedComplexT = if (nestedComplex.nonEmpty) "NESTED COMPLEX TYPE" else ""
```
tgravescs, line 140: so we don't print the actual type here then?  Please add documentation to this function and explain.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/qualification/QualAppInfo.scala
```diff
@@ -217,6 +230,7 @@ class QualAppInfo(
       }.mkString(";")
       val writeFormat = writeFormatNotSupported(writeDataFormat)
       val (allComplexTypes, nestedComplexTypes) = reportComplexTypes
+      val problems = getAllPotentialProblems(getPotentialProblems, nestedComplexTypes)
```
tgravescs, line 233: it seems odd to have 2 getPotentialProblems functions, if they are specific rename the functions.  I would have expected this just to go into getPotentialProblems and I kind of would expect from the name findPotentialIssues would find these.  I'm fine if that doesn't make sense but we need to document better. Also findPotentialIssues is in the base which would let us use in profiling tool easily, this is not. It doesn't need to be at this point but just want code to be easy to follow.

tgravescs (commented):
we should also update the docs to explain this and perhaps say if you see this go look at what type it is in other field.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/qualification/QualAppInfo.scala
nartal1: Updated the function name. Please let me know if the new name is okay.

tgravescs (approved):

tgravescs:
build

PR #3524, opened 2021-09-17
Update hashing to normalize -0.0 on 3.2+

revans2:
This fixes issues with hash normalization related to SPARK-35207 (Thanks @jlowe for finding the root cause of the issue).

This adds in some new APIs that should make this type of transformation simpler for other code, but because this is late in the 21.10 release I didn't want to update any existing code to use it yet.


revans2:
build

tgravescs (approved):

PR #3531, opened 2021-09-17
Fall back to the CPU for date/time parsing we cannot support yet

### LINKED ISSUE #3382: [BUG] Support years with up to 7 digits when casting from String to Date in Spark 3.2

revans2:
This fixes #3382


revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -81,11 +82,16 @@ class CastExprMeta[INPUT <: CastBase](
...
             "while CPU returns \"+Infinity\" and \"-Infinity\" respectively. To enable this " +
             s"operation on the GPU, set ${RapidsConf.ENABLE_CAST_STRING_TO_FLOAT} to true.")
-      case (_: StringType, _: TimestampType) if !conf.isCastStringToTimestampEnabled =>
-        willNotWorkOnGpu("the GPU only supports a subset of formats " +
-            "when casting strings to timestamps. Refer to the CAST documentation " +
-            "for more details. To enable this operation on the GPU, set" +
-            s" ${RapidsConf.ENABLE_CAST_STRING_TO_TIMESTAMP} to true.")
+      case (_: StringType, _: TimestampType) =>
+        if (!conf.isCastStringToTimestampEnabled) {
+          willNotWorkOnGpu("the GPU only supports a subset of formats " +
```
jlowe, line 87: When we add a new config that controls whether or not a type is allowed on the GPU for a certain operation, we also need to update the supported ops documentation with at least a ps note explaining the situation and pointing to the config.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -630,6 +630,15 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(false)
 
+  val ENABLE_EXTENDED_YEAR_PARSING = conf("spark.rapids.sql.parseExtendedYears.enabled")
+      .doc("SPARK-35780 went into Spark 3.2.0+ and extended parsing of years in dates and " +
```
jlowe, line 634: Nit: Is the Spark JIRA too much detail for users here?  Wondering if this should be simplified to something like:
```
Spark 3.2+ parses the full range of possible year values in dates and timestamps beyond the typical 4-digit year. When this is enabled it is assumed all year values are not using the extended syntax and thus can be processed on the GPU which does not yet support extended range year values.
```

Also wondering if the configuration name is a bit misleading.  It's not so much enabling parsing of extended years as it is the user confirming there are no extended year values in the data.  In that sense it seems similar to the hasNans config.

revans2: Do you have abetter name?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
revans2: We have not been consistent about this anywhere. I understand your desire to have supported ops be complete, but we had this discussion before about what is the point of that documentation and how complete do we need to be. Have we decided that it should be exhaustive? If so then we should file a follow on issue to add in all of the notes that are missing. If not then we need to decide what exactly it is that we do want.

jlowe: I bring it up because QA has been filing bugs based on what they're seeing in the supported ops doc, so they're treating it as exhaustive.  If it's not supposed to be exhaustive then the doc probably needs to be more clear on that.  I'll file a followup issue to discuss what to do regarding these one-off configs and that doc.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
jlowe: `spark.rapids.sql.noExtendedYearValues` sounds like a better name to me, since it's focusing on the user validating there are no extended year values that need to be dealt with rather than implying there's extended year parsing that is disabled by default.

revans2:
build

jlowe (approved):

PR #3533, opened 2021-09-17
Define a SPARK_SHIM_VER to pick up specific rapids-4-spark-integration-tests jars

abellina:
This covers a nice to have for the new build system changes, where integration_tests/target/ could have several versions of the integration_tests jars, like so:

```
integration_tests/target/rapids-4-spark-integration-tests_2.12-21.10.0-SNAPSHOT-spark301.jar
integration_tests/target/rapids-4-spark-integration-tests_2.12-21.10.0-SNAPSHOT-spark301-jar-with-dependencies.jar
integration_tests/target/rapids-4-spark-integration-tests_2.12-21.10.0-SNAPSHOT-spark301tests.jar
integration_tests/target/rapids-4-spark-integration-tests_2.12-21.10.0-SNAPSHOT-spark311.jar
integration_tests/target/rapids-4-spark-integration-tests_2.12-21.10.0-SNAPSHOT-spark311-jar-with-dependencies.jar
integration_tests/target/rapids-4-spark-integration-tests_2.12-21.10.0-SNAPSHOT-spark311tests.jar
```

The change is so we pick a set of jars that are specific to the version of spark given by `VERSION_STRING`. Note this can also be overwritten by setting `SPARK_SHIM_VER`.


jenkins/databricks/test.sh
```diff
@@ -20,6 +20,7 @@ set -ex
 LOCAL_JAR_PATH=${LOCAL_JAR_PATH:-''}
 SPARK_CONF=${SPARK_CONF:-''}
 BASE_SPARK_VER=${BASE_SPARK_VER:-'3.0.1'}
+[[ -z $SPARK_SHIM_VER ]] && { SPARK_SHIM_VER=spark${BASE_SPARK_VER//.}db }
```
tgravescs, line 23: only downside to this is for dev use we didn't specify and it would still run correct tests just skip cache test.  may just need to update our internal docs

abellina: why would it skip cache tests?

abellina: I made this 3.1.1 to match build.sh. It won't skip the cache test by default.

tgravescs (approved):

abellina:
build

PR #3507, opened 2021-09-16
TimeAdd supports DayTimeIntervalType

wbo4958:
This PR is to fix https://github.com/NVIDIA/spark-rapids/issues/3502 by adding DayTimeIntervalType support for TimeAdd.

Tested date_time_test.py against spark 3.2.0, and pass

BTW,  This RP didn't add DayTimeIntervalType support for TimeSub since it has been deleted from Spark 3.1

PR #3534, opened 2021-09-17
Fix package name and sql string issue for GpuTimeAdd

wbo4958:
This is an following PR for https://github.com/NVIDIA/spark-rapids/pull/3507#discussion_r711378272. It will fix the sql/toString of GpuTimeAdd since TimeAdd in spark 3.2+ can support not only `Time +` but also `Time -`. And thx @tgravescs for pointing the package naming issue


wbo4958:
build

wbo4958:
build

tgravescs (approved):

Issue #3472, opened 2021-09-13
[BUG] GpuAdd and GpuMultiply do not include failOnError


revans2:
**Describe the bug**
Add and Multiply added in failOnError fields.

This is similar to https://github.com/NVIDIA/spark-rapids/issues/2078 but for overflow. We either need to fall back to the GPU in these cases or implement the feature.  Note that this impacts Canonicalization too.

firestarman:
> ... We either need to fall back to the GPU in ... 

fall back to the CPU ?

revans2:
Yes that was a typo. We should fall back to the CPU unless we have a good way to detect overflow and we want to add it in.

firestarman:
We can leverage the same solution with Java to detect overflow for “Add” op. 
[addExact(int x, int y)](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Math.java#L874) and [addExact(long x, long y)](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Math.java#L894) are using the same algorithm to check if overflow occurs, which is likely to apply to Byte and Short.

Unfortunately it is much more complitcated for "Multiply" op, so we can fall back to CPU now.

Any idea ?

revans2:
Yes we could detect a sign change like add exact is doing.  This will not work for the AST version of add so we still will need to fall back in that case. We also will have to be careful with Decimal vs non-Decimal add.

revans2:
I just saw that UnaryMinus also has a fail on error added in...

revans2:
multiplyExact is more complicated, but it should work here too.  We support everything that is a part of it.  I would file a follow on issue though, for us to implement our own kernel to be able to do the checking ourselves.

firestarman:
> I just saw that UnaryMinus also has a fail on error added in...

So do almost all the expressions in file [arithmetic.scala](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala), e.g. `Subtract`, `Abs`...

firestarman:
> Yes we could detect a sign change like add exact is doing. This will not work for the AST version of add so we still will need to fall back in that case. We also will have to be careful with Decimal vs non-Decimal add.

Yes seems we need to always fall back to CPU when ansi is enabled. Since there is no reliable way to know whether the AST version will be used. 
[...]

PR #3537, opened 2021-09-18
GpuAdd supports ANSI mode.

firestarman:
And let Multiply fall back to CPU when ANSI mode is enabled.

This closes https://github.com/NVIDIA/spark-rapids/issues/3472

### DIFF
--- a/integration_tests/src/main/python/arithmetic_ops_test.py
+++ b/integration_tests/src/main/python/arithmetic_ops_test.py
@@ -14,10 +14,11 @@
 
 import pytest
 
-from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_and_cpu_error
+from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_and_cpu_error, assert_gpu_fallback_collect
 from data_gen import *
-from marks import incompat, approximate_float
+from marks import incompat, approximate_float, allow_non_gpu
 from pyspark.sql.types import *
+from pyspark.sql.types import IntegralType
 from spark_session import with_cpu_session, with_gpu_session, with_spark_session, is_before_spark_311, is_before_spark_320
 import pyspark.sql.functions as f
 from pyspark.sql.utils import IllegalArgumentException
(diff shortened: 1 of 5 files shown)


firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1658,8 +1658,17 @@ object GpuOverrides extends Logging {
         ("lhs", TypeSig.gpuNumeric, TypeSig.numericAndInterval),
         ("rhs", TypeSig.gpuNumeric, TypeSig.numericAndInterval)),
       (a, conf, p, r) => new BinaryAstExprMeta[Add](a, conf, p, r) {
+        val ansiEnabled = SQLConf.get.ansiEnabled
+
+        override def tagSelfForAst(): Unit = {
+          super.tagSelfForAst()
+          if (ansiEnabled) {
+            willNotWorkInAst("AST Addition does not support ansi mode.")
```
jlowe, line 1666:
```suggestion
            willNotWorkInAst("AST Addition does not support ANSI mode.")
```

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -62,6 +63,26 @@ def test_multiplication(data_gen):
                 f.col('a') * f.col('b')),
             conf=allow_negative_scale_of_decimal_conf)
 
+# No overflow gens here because we just focus on verifying the fallback to CPU when
+# enabling ansi mode. But overflows will fail the tests because CPU runs raise
```
jlowe, line 67:
```suggestion
# enabling ANSI mode. But overflows will fail the tests because CPU runs raise
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1709,6 +1718,10 @@ object GpuOverrides extends Logging {
               }
             case _ => // NOOP
           }
+
+          if (SQLConf.get.ansiEnabled) {
+            willNotWorkOnGpu("GPU Multiplication does not support ansi mode")
```
jlowe, line 1723:
```suggestion
            willNotWorkOnGpu("GPU Multiplication does not support ANSI mode")
```

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -62,6 +63,26 @@ def test_multiplication(data_gen):
...
+# exceptions.
+_no_overflow_multiply_gens = [
+    ByteGen(min_val = 1, max_val = 10, special_cases=[]),
+    ShortGen(min_val = 1, max_val = 100, special_cases=[]),
+    IntegerGen(min_val = 1, max_val = 1000, special_cases=[]),
+    LongGen(min_val = 1, max_val = 3000, special_cases=[]),
+    float_gen, double_gen,
```
revans2, line 74: nit: I think we can have a follow on issue to re-enable float, double, and decimal processing in ANSI mode.  Each of them handle overflows already in different ways that should be okay.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1380,7 +1380,7 @@ object GpuOverrides extends Logging {
       ExprChecks.mathUnary,
       (a, conf, p, r) => new UnaryExprMeta[Log1p](a, conf, p, r) {
         override def convertToGpu(child: Expression): GpuExpression =
-          GpuLog(GpuAdd(child, GpuLiteral(1d, DataTypes.DoubleType)))
+          GpuLog(GpuAdd(child, GpuLiteral(1d, DataTypes.DoubleType), SQLConf.get.ansiEnabled))
```
revans2, line 1383: nit: This is a little misleading, because there is no fail on error overflow processing on doubles, and doubles are the only thing supported by Log1p

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -1658,8 +1658,17 @@ object GpuOverrides extends Logging {
         ("lhs", TypeSig.gpuNumeric, TypeSig.numericAndInterval),
         ("rhs", TypeSig.gpuNumeric, TypeSig.numericAndInterval)),
       (a, conf, p, r) => new BinaryAstExprMeta[Add](a, conf, p, r) {
+        val ansiEnabled = SQLConf.get.ansiEnabled
+
+        override def tagSelfForAst(): Unit = {
+          super.tagSelfForAst()
+          if (ansiEnabled) {
```
revans2, line 1665: nit: It would be good to check the data type, because double and float don't have special overflow processing and should be able to work in AST mode.

revans2 (approved):
Just a few nits that can be handled by follow on issues.

jlowe (commented):
Small capitalization nits but otherwise lgtm.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
revans2: I'll do this in a follow on issue as @firestarman is now out on vacation.

PR #3556, opened 2021-09-20
Fix databricks test.sh script for passing spark shim version

tgravescs:
the test.sh script fails on Databricks so fix that syntax error and then export SPARK_SHIM_VER so it gets passed along to all the run scripts properly.

Signed-off-by: Thomas Graves <tgraves@nvidia.com>

PR #3504, opened 2021-09-15
Fix reserialization of broadcasted tables

jlowe:
Fixes #3266.

Re-serializing a broadcast caused issues because it purposely deserializes into a different state on the executor than it does on the driver, so serializing on the executor due to memory pressure after being deserialized could cause an NPE.

This updates the broadcast to always deserialize only to memory but then throw away the host buffers once the GPU batch is requested at which point it creates it.

PR #3559, opened 2021-09-20
Commonize shim module join and shuffle files

tgravescs:
Many of the join and shuffle related files in the shim module were just there to build against the specific version of spark.  Moves those files into the sql-plugin.  Most of those had differences between 3.0.x and 3.1.x and then many times databricks had slight variations, so consolidated the shims module files into 3 or 4 in the sql-plugin directory.

Also note more to commonize later, doing it in pieces to hopefully keep easier to review.  

I ran tests on databricks and some Apache but not all.


tgravescs:
build

tgravescs:
databricks build failed because of https://github.com/NVIDIA/spark-rapids/pull/3556

tgravescs:
build

gerashegalov (approved):
LGTM

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBroadcastNestedLoopJoinExec.scala
```diff
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
-package com.nvidia.spark.rapids.shims.spark311
```
jlowe, line 17: Now that `GpuBroadcastNestedLoopJoinExec` is no longer shimmed in any way, should this be moved into the same file as `GpuBroadcastNestedLoopJoinExecBase`?  Is there still a reason for the latter class?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
```diff
@@ -356,7 +356,7 @@ abstract class GpuBroadcastExchangeExecBase(
         }
       }
     }
-    GpuBroadcastExchangeExec.executionContext.submit[Broadcast[Any]](task)
```
jlowe, line 359: Nit: Would like to see this file named `GpuBroadcastExchangeExecBase` or something more accurate to what it contains.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuShuffleExchangeExec.scala
```diff
@@ -230,7 +230,7 @@ abstract class GpuShuffleExchangeExecBase(
     }
 }
 
-object GpuShuffleExchangeExec {
+object GpuShuffleExchangeExecBase {
```
jlowe, line 233: Nit: Update filename to GpuShuffleExchangeExecBase.scala.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBroadcastNestedLoopJoinExec.scala
tgravescs: there is definitely more things to move around and commonize, I was trying to keep it simple here and easy to review by just moving the files out of shim modules without a lot of code changes that could add bugs and make it harder to review.   We can then follow up with more changes.

tgravescs:
build

tgravescs:
very odd > 07:22:08  [ERROR] /home/ubuntu/spark-rapids/tests/src/test/scala/com/nvidia/spark/rapids/SerializationSuite.scala:20: object lang3 is not a member of package org.apache.commons

builds fine locally and didn't change anything that would cause this

tgravescs:
that failed on databricks and I bet is from https://github.com/NVIDIA/spark-rapids/pull/3504

tgravescs:
build

jlowe:
> rapids/tests/src/test/scala/com/nvidia/spark/rapids/SerializationSuite.scala:20: object lang3 is not a member of package org.apache.commons

Looks like the test was implicitly picking up commons lang3 from Spark and should have explicitly specified the dependency.  I'll post a PR to fix.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBroadcastNestedLoopJoinExec.scala
jlowe: I'm fine with followups as long as we're tracking somewhere the things we want to do so they don't get dropped.

pom.xml
```diff
@@ -263,7 +267,9 @@
                                         <source>${project.basedir}/src/main/301until320-all/scala</source>
                                         <source>${project.basedir}/src/main/311until320-all/scala</source>
                                         <source>${project.basedir}/src/main/311+-all/scala</source>
+                                        <source>${project.basedir}/src/main/311+-nondb/scala</source>
                                         <source>${project.basedir}/src/main/311until320-apache/scala</source>
+                                        <source>${project.basedir}/src/main/301+-nondb/scala</source>
```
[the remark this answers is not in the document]
tgravescs: also note I have older ide which doesn't have that option.  I'll look at upgrading later, but I'm not sure I agree with its sorting going by the other ones, for instance:

```
<source>${project.basedir}/src/main/301until310-nondb/scala</source>
<source>${project.basedir}/src/main/301until320-nondb/scala</source>
<source>${project.basedir}/src/main/301until320-all/scala</source>
 <source>${project.basedir}/src/main/301until310-all/scala</source>
 <source>${project.basedir}/src/main/301+-nondb/scala</source>
```

Why is it putting nondb before all, doesn't seem like by alphabetical? As long as we are consistent I'm fine with it but don't want to require certain version of IDE either though.  Or perhaps those weren't sorted?

jlowe (approved):

PR #3566, opened 2021-09-21
Enable String Array Max and Min

revans2:
This depends on https://github.com/rapidsai/cudf/pull/9263 which is targeted to 21.10.  I am putting this up mostly so we don't forget to put this in either to 21.10 or to 21.12.

### DIFF
--- a/docs/supported_ops.md
+++ b/docs/supported_ops.md
@@ -1980,7 +1980,7 @@ are limited.
 <td> </td>
 <td> </td>
 <td> </td>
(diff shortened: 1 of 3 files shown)


jlowe (approved):

revans2:
The dependency was merged about 20 mins ago. Marking this as ready for review. Will try to kick off the internal builds so we can get this tested ASAP.

revans2:
build

PR #3570, opened 2021-09-21
ANSI Support for Abs, UnaryMinus, and Subtract

revans2:
I will do divide, IntegralDivide, Remainder, and Pmod as a separate PR because there are some parts that are done, but other parts that are not and each is spread out among different Spark versions, which makes this fun to try and shim correctly.


revans2:
build

revans2:
Never mind divide is already done, it uses the shim layer differently that I did so it took a bit to find the code for it.

jlowe (approved):

revans2:
Apparently in older versions of Spark `1.0f` is parsed as trying to access the `0f` element of the struct `1`. I'll see if I can fix this in the tests...

PR #3574, opened 2021-09-21
Add in ANSI date time fallback

revans2:
It is hard for us to match exactly what spark is doing for date/time parsing fallback in ANSI mode. I will file a follow on issue to see if we can actually support it, but for now we just fall back to the CPU.


revans2:
build

revans2:
Apparently in older versions of Spark `1.0f` is parsed as trying to access the `0f` element of the struct `1`. I'll see if I can fix this in the tests...

PR #3577, opened 2021-09-21
Commonize more shim module files [databricks]

tgravescs:
This commonizes the majority of the rest of the shim module files excluding shuffle and the main Spark3XXShim file.

I tried to keep this easy to review and mostly moved files, a few functions in SparkBaseShims had to move to the respective Spark3XX file to make it common.


tgravescs:
build

pom.xml
```diff
@@ -141,6 +142,7 @@
                                         <source>${project.basedir}/src/main/301until310-nondb/scala</source>
                                         <source>${project.basedir}/src/main/301until320-nondb/scala</source>
                                         <source>${project.basedir}/src/main/301until320-all/scala</source>
+                                        <source>${project.basedir}/src/main/301until310-all/scala</source>
```
revans2, line 145: nit: There are getting to be enough directories now, that it would be good to have a consistent ordering to them, so it is simpler to tell if something is missing for example.

revans2 (approved):

pom.xml
tgravescs: ah, I missed a few I tried moved some of them but didn't get them all. I need to update my IntelliJ as well as Gera mentioned it has a sort lines in new version, but we should decide on order

revans2 (approved):

tgravescs:
wait to build until https://github.com/NVIDIA/spark-rapids/issues/3580 fixed

pom.xml
tgravescs: sorted them

sql-plugin/src/main/301until310-all/scala/com/nvidia/spark/rapids/shims/v2/OffsetWindowFunctionMeta.scala
```diff
@@ -14,62 +14,47 @@
  * limitations under the License.
  */
 
-package com.nvidia.spark.rapids.shims.spark313
+package com.nvidia.spark.rapids.shims.v2
```
tgravescs, line 17: this whole file is a very odd diff, it did not get moved from 313 ithis got copied from GpuWindowExpression.scala

tgravescs:
build

gerashegalov (approved):
LGTM

PR #3595, opened 2021-09-22
Update the overflow check algorithm for Subtract

firestarman:
The new algorithm is copied from [the openJDK version](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Math.java#L916), which looks simpler and probably gets better performance.


firestarman:
build

revans2 (approved):
Thanks for cleaning this up

PR #3597, opened 2021-09-22
ANSI check for aggregates

abellina:
Adds fallback for aggregates if ANSI mode is enabled. Specifically: `Sum`, `Count`, and `Average` all fallback in various ways. `Sum` and `Count` check the types that they are producing (`resultType`), where `Average` unconditionally falls back, because it uses `Long` internally.  The other aggregates do not fallback. 

I am finding that adding `DecimalGen` to the mix for the testing is probably not possible. Spark is either optimizing it out (`UnscaledValue`), or we don't support the precision. But I could have missed something.

This does not add the `CentralMomentAgg` variety to the checks, since they are not merged yet, but those definitely need to fallback.


abellina:
build

abellina:
Note I didn't include window agg tests, but I can certainly add some. I did verify that a simple window op using a sum, would fallback given the ANSI flag.

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -1088,3 +1073,81 @@ def do_it(spark):
...
+
+@ignore_order
+@allow_non_gpu('HashAggregateExec', 'Alias', 'AggregateExpression', 'Cast',
+  'HashPartitioning', 'ShuffleExchangeExec', 'Average')
+@pytest.mark.parametrize('data_gen', [_grpkey_strings_with_extra_nulls], ids=idfn)
+@pytest.mark.parametrize('conf', get_params(_confs, params_markers_for_confs), ids=idfn)
+@pytest.mark.parametrize('ansi_enabled', ['true', 'false'])
+def test_hash_grpby_avg_nulls_ansi(data_gen, conf, ansi_enabled):
+    local_conf = copy_and_update(conf, {'spark.sql.ansi.enabled': ansi_enabled})
+    assert_gpu_and_cpu_are_equal_collect(
```
revans2, line 1086: I m confused why want to combine this into a single test. Doing it this way makes it so we cannot verify that we did fallback for Average and it also means that we explicitly allow things to not be on the GPU when we know that they should be in non-ansi mode.

abellina: Sorry, I moved that test down together with the other ANSI stuff, I didn't change it. Will split it, totally agree.

abellina: Done here: https://github.com/NVIDIA/spark-rapids/pull/3597/commits/4ddbb6bdfbbe92a3fee286a022c321ebe33d162a

revans2 (approved):

revans2:
build

Issue #3601, opened 2021-09-22
[BUG] Latest 21.10 snapshot jars failing with java.lang.ClassNotFoundException: com.nvidia.spark.rapids.ColumnarRdd with XGBoost


NvTimLiu:
**Describe the bug**
The latest snapshot jars for v21.10 are failing with the following error:

```
21/09/22 09:13:04 INFO SparkContext: SparkContext already stopped.
Exception in thread "main" java.lang.ClassNotFoundException: com.nvidia.spark.rapids.ColumnarRdd
        at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
        at java.lang.Class.forName0(Native Method)
        at java.lang.Class.forName(Class.java:264)
        at ml.dmlc.xgboost4j.scala.spark.rapids.GpuUtils$.toColumnarRdd(GpuUtils.scala:37)
        at ml.dmlc.xgboost4j.scala.spark.rapids.GpuXGBoost$.trainOnGpuInternal(GpuXGBoost.scala:240)
        at ml.dmlc.xgboost4j.scala.spark.rapids.GpuXGBoost$.trainDistributedOnGpu(GpuXGBoost.scala:186)
        at ml.dmlc.xgboost4j.scala.spark.rapids.GpuXGBoost$.trainOnGpu(GpuXGBoost.scala:91)
        at ml.dmlc.xgboost4j.scala.spark.rapids.GpuXGBoost$.fitOnGpu(GpuXGBoost.scala:52)
        at ml.dmlc.xgboost4j.scala.spark.XGBoostClassifier.fit(XGBoostClassifier.scala:170)
        at com.nvidia.spark.examples.mortgage.GPUMain$.$anonfun$main$4(GPUMain.scala:72)
        at com.nvidia.spark.examples.utility.Benchmark.time(Benchmark.scala:29)
        at com.nvidia.spark.examples.mortgage.GPUMain$.main(GPUMain.scala:72)
        at com.nvidia.spark.examples.mortgage.GPUMain.main(GPUMain.scala)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #3607, opened 2021-09-22
com.nvidia.spark.rapids.ColumnarRdd not exposed to user for XGBoost

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/3601

we were not exposing ColumnarRDD as an unshimmed class.  This fixes that. I tested on xgboost with the python notebook and verified it works now.

This is a bit different since its an object and not a real class we instantiate, so if we have better ways to do this let me know, or we can followup as I'm sure we will have more things that need things like this.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarRdd.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarRdd.scala
@@ -39,11 +39,18 @@ import org.apache.spark.sql.rapids.execution.InternalColumnarRddConverter
  * when they no longer need it.
  */
 object ColumnarRdd {
+
+  lazy val internalConverter: Class[_] = ShimLoader.newColumnarRDDClass()
+
   def apply(df: DataFrame): RDD[Table] = {
-    InternalColumnarRddConverter(df)
+    internalConverter
+      .getDeclaredMethod("apply", classOf[DataFrame])
+      .invoke(null, df).asInstanceOf[RDD[Table]]
   }
 
   def convert(df: DataFrame): RDD[Table] = {
-    InternalColumnarRddConverter.convert(df)
+    internalConverter
+      .getDeclaredMethod("convert", classOf[DataFrame])
+      .invoke(null, df).asInstanceOf[RDD[Table]]
   }
 }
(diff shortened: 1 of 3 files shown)


tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarRdd.scala
```diff
@@ -39,11 +38,16 @@ import org.apache.spark.sql.rapids.execution.InternalColumnarRddConverter
...
+
   def apply(df: DataFrame): RDD[Table] = {
-    InternalColumnarRddConverter(df)
+    convert(df)
   }
 
   def convert(df: DataFrame): RDD[Table] = {
-    InternalColumnarRddConverter.convert(df)
+    internalConverter
+      .getDeclaredMethod("convert", classOf[DataFrame])
```
jlowe, line 50: Shouldn't we be caching the method rather than looking it up each time?

revans2 (dismissed):

tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarRdd.scala
tgravescs: yes updated

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
```diff
@@ -289,6 +289,12 @@ object ShimLoader extends Logging {
     shimProviderClass = classname
   }
 
+  def newClassOf(className: String): Class[_] = {
```
gerashegalov, line 292: nit: consider `loadClass` as the name

nit be made resusable for newInstanceOf

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
```diff
@@ -343,4 +349,8 @@ object ShimLoader extends Logging {
   def newInternalExclusiveModeGpuDiscoveryPlugin(): ResourceDiscoveryPlugin = {
     newInstanceOf("com.nvidia.spark.rapids.InternalExclusiveModeGpuDiscoveryPlugin")
   }
+
+  def newColumnarRDDClass(): Class[_] = {
```
gerashegalov, line 353: nit: I'd name newXyz for methds returning objects and loadXyz for classes

tgravescs:
build

gerashegalov (approved):
LGTM

jlowe (approved):

Issue #3116, opened 2021-08-03
[FEA] Support Maps, Lists, and Structs as non-key columns on joins


revans2:
**Is your feature request related to a problem? Please describe.**
Joins should generally work out of the box for all non key columns that cudf can gather (which should be everything that we can also support).

We have run into a few issues when testing this.
https://github.com/rapidsai/cudf/issues/8938
https://github.com/rapidsai/cudf/issues/8929

The first is a blocker. The second is a corner case that would be good to fix, but is not that likely to show up outside of a specially crafted test.

**Describe the solution you'd like**
Enable Maps, Lists and Structs as non-key columns in BroadcastExchangeExec (only broadcast join uses this right now), BroadcastJoin, ShuffledHashJoin, SortMergeJoin, BroadcastNestedLoopJoin, and CartesianProduct.  This should include adding tests for them too, which mostly should be extending existing tests.

PR #3612, opened 2021-09-22
Add support for nested types as non-key columns on joins 

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3116. 
Tested various joins(BHJ, SMJ, Shuffle Hash Join, Broadcast Nested loop join and Catersian Product) on maps, structs and arrays.
We already had support to most of them. This PR adds support for missing types and adds missing tests.

### DIFF
--- a/integration_tests/src/main/python/join_test.py
+++ b/integration_tests/src/main/python/join_test.py
@@ -182,6 +182,16 @@ def test_hash_join_map(data_gen, join_type):
         return left.join(right, left.key == right.r_key, join_type)
     assert_gpu_and_cpu_are_equal_collect(do_join, conf=_hash_join_conf)
 
+@validate_execs_in_gpu_plan('GpuShuffledHashJoinExec')
+@ignore_order(local=True)
+@pytest.mark.parametrize('data_gen', [all_basic_struct_gen], ids=idfn)
+@pytest.mark.parametrize('join_type', all_join_types, ids=idfn)
+def test_hash_join_struct(data_gen, join_type):
+    def do_join(spark):
+        left, right = create_nested_df(spark, short_gen, data_gen, 50, 500)
+        return left.join(right, left.key == right.r_key, join_type)
+    assert_gpu_and_cpu_are_equal_collect(do_join, conf=_hash_join_conf)
+
 # local sort because of https://github.com/NVIDIA/spark-rapids/issues/84
 # After 3.1.0 is the min spark version we can drop this
 @ignore_order(local=True)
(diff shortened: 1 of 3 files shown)


nartal1:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3366,8 +3366,9 @@ object GpuOverrides extends Logging {
       (join, conf, p, r) => new GpuBroadcastNestedLoopJoinMeta(join, conf, p, r)),
     exec[CartesianProductExec](
       "Implementation of join using brute force",
-      ExecChecks(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64 +
-          TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64),
+      ExecChecks((TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64 +
+          TypeSig.ARRAY + TypeSig.MAP + TypeSig.STRUCT)
+          .nested(TypeSig.commonCudfTypes + TypeSig.NULL + TypeSig.DECIMAL_64),
```
jlowe, line 3371: Why not support nested arrays and structs?  It's just a gather, and we support gather on arbitrarily nested structures.

integration_tests/src/main/python/join_test.py
```diff
@@ -182,6 +182,16 @@ def do_join(spark):
         return left.join(right, left.key == right.r_key, join_type)
     assert_gpu_and_cpu_are_equal_collect(do_join, conf=_hash_join_conf)
 
+@validate_execs_in_gpu_plan('GpuShuffledHashJoinExec')
+@ignore_order(local=True)
+@pytest.mark.parametrize('data_gen', [all_basic_struct_gen], ids=idfn)
+@pytest.mark.parametrize('join_type', all_join_types, ids=idfn)
+def test_hash_join_struct(data_gen, join_type):
```
jlowe, line 189: Shouldn't we simply combine `test_hash_join_array`, `test_hash_join_map`, and `test_hash_join_struct` into a single `test_hash_join` that is parameterized with all the various types to test?  That matches the approach we use almost everywhere else in the tests and reduces the amount of boilerplate for this type of test.

integration_tests/src/main/python/join_test.py
```diff
@@ -255,6 +265,19 @@ def do_join(spark):
     conf = copy_and_update(allow_negative_scale_of_decimal_conf, {'spark.rapids.sql.batchSizeBytes': batch_size})
     assert_gpu_and_cpu_are_equal_collect(do_join, conf=conf)
 
+# local sort because of https://github.com/NVIDIA/spark-rapids/issues/84
+# After 3.1.0 is the min spark version we can drop this
+@ignore_order(local=True)
+@pytest.mark.order(1) # at the head of xdist worker queue if pytest-order is installed
+@pytest.mark.parametrize('data_gen', single_level_array_gens + map_string_string_gen + [all_basic_struct_gen], ids=idfn)
+@pytest.mark.parametrize('batch_size', ['100', '1g'], ids=idfn) # set the batch size so we can test multiple stream batches
+def test_cartesian_join_nestedtype_nonkey(data_gen, batch_size):
```
jlowe, line 274: Cartesian joins don't have a key, it's just a nested for loop to generate every row combination between the two tables.  This test should be combined with the previous test to test against all types, nested or not.

integration_tests/src/main/python/join_test.py
```diff
@@ -314,6 +337,18 @@ def do_join(spark):
     conf = copy_and_update(allow_negative_scale_of_decimal_conf, {'spark.rapids.sql.batchSizeBytes': batch_size})
     assert_gpu_and_cpu_are_equal_collect(do_join, conf=conf)
 
+# local sort because of https://github.com/NVIDIA/spark-rapids/issues/84
+# After 3.1.0 is the min spark version we can drop this
+@ignore_order(local=True)
+@pytest.mark.parametrize('data_gen', single_level_array_gens + map_string_string_gen + [all_basic_struct_gen], ids=idfn)
+@pytest.mark.parametrize('batch_size', ['100', '1g'], ids=idfn) # set the batch size so we can test multiple stream batches
+def test_broadcast_nested_loop_join_nestedtype_nonkey(data_gen, batch_size):
```
jlowe, line 345: Similar comment as the previous.  There's no "key" here,  it's a cross join.

integration_tests/src/main/python/join_test.py
nartal1: Thanks for the review. You are right, combining nested types test reduces the boilerplate code. I have combined other tests as well(SMJ, BHJ). Please take another look.

integration_tests/src/main/python/join_test.py
nartal1: combined with previous test

integration_tests/src/main/python/join_test.py
nartal1: done.

jlowe (approved):

nartal1:
build

PR #3615, opened 2021-09-22
Ignore order for the test_no_fallback_when_ansi_enabled

### LINKED ISSUE #3611: [BUG] test_no_fallback_when_ansi_enabled failed in databricks
@razajafri found this in one of his PRs, the results from the CPU and GPU do not match for `test_no_fallback_when_ansi_enabled`, which I added here #3597:

First few rows from the CPU: 

```
 [Row(a=None, first(b)=118, last(b)=507, min(b)=118, max(b)=507), Row(a=1, first(b)=507, last(b)=507, min(b)=507, max(b)=507),  Row(a=2, first(b)=848, last(b)=848, min(b)=848, max(b)=848),
```

First few rows from the GPU:
```
[Row(a=1, first(b)=507, last(b)=507, min(b)=507, max(b)=507), Row(a=2, first(b)=848, last(b)=848, min(b)=848, max(b)=848)
```

Last row on the GPU:
```
Row(a=None, first(b)=118, last(b)=507, min(b)=118, max(b)=507
```

I am not entirely sure how this is happening, after a `coalesce(1)` and `orderBy(every column)`.

abellina:

Closes #3611 

This was a dumb mistake that @razajafri found in https://github.com/NVIDIA/spark-rapids/pull/3330. As @jlowe nicely put it, yeah what I was trying to do is not deterministic given the hash tables involved (partitioning and in the group by), since I didn't have a sort _after_ the operation.

### DIFF
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -1158,7 +1158,7 @@ def test_count_fallback_when_ansi_enabled(data_gen):
     assert_gpu_fallback_collect(do_it, 'Count',
         conf={'spark.sql.ansi.enabled': 'true'})
 
-@ignore_order
+@ignore_order(local=True)
 @pytest.mark.parametrize('data_gen', _no_overflow_ansi_gens, ids=idfn)
 def test_no_fallback_when_ansi_enabled(data_gen):
     def do_it(spark):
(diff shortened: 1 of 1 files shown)


abellina:
build

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -1158,7 +1158,7 @@ def do_it(spark):
     assert_gpu_fallback_collect(do_it, 'Count',
         conf={'spark.sql.ansi.enabled': 'true'})
 
-
+@ignore_order
```
jlowe, line 1161:
```suggestion
@ignore_order(local=True)
```

abellina:
build

abellina:
No longer needed, since this PR had the same fix: https://github.com/NVIDIA/spark-rapids/pull/3330

tgravescs (approved):

razajafri:
Will merging this in before #3627 have the desired effect? Because the revert will revert this change as part of my PR #3330

abellina:
Well your PR diff should be a no-op given this change (I think), but good to check.

Issue #3520, opened 2021-09-16
[BUG] Scaladoc warnings emitted during build


jlowe:
The following Scaladoc warnings appear in recent builds:
```
/spark-rapids/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuListUtils.scala:27: warning: Could not find any member to link for "scala.IllegalArgumentException".
  /**
  ^
```
and
```
/spark-rapids/shims/spark301/src/main/scala/org/apache/spark/sql/rapids/execution/python/shims/spark301/GpuArrowEvalPythonExec.scala:484: warning: Could not find any member to link for "GpuPythonUDF".
/**
^
```

firestarman:
I am wondering whether we could supress the warnings by add the option `-no-link-warnings` to scala plugin, if it was hard to fix this.

jlowe:
These warnings occur because the classes aren't in the same module and Scaladoc often has difficulty linking across modules.  Simplest thing to do here is change the references to use backticks for monospacing rather than trying to link, as was done for other occurrences like this.

firestarman:
And the warning log on GpuPythonUDF is no longer found in recent builds.

PR #3621, opened 2021-09-23
Update the doc of GpuListUtils to eliminate a linkage warning.

firestarman:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3520.

And the warning log on `GpuPythonUDF` is no longer found in recent builds.

Scaladoc will always try to create a link for the first word after the `@throws`, so here chooses to change it to a common comment instead of a tag.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuListUtils.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuListUtils.scala
@@ -30,7 +30,7 @@ object GpuListUtils extends Arm {
    * both listCol and newDataCol alive longer than the returned ColumnView.
    * @param listCol the list column to use as a template
    * @param newDataCol the new data column.
-   * @return a new ColumnView. But throws an scala.IllegalArgumentException if data column does
+   * @return a new ColumnView. Throws IllegalArgumentException if data column does
    *                           not match the original data column in size.
    */
   def replaceListDataColumnAsView(
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuListUtils.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuListUtils.scala
@@ -30,9 +30,8 @@ object GpuListUtils extends Arm {
    * both listCol and newDataCol alive longer than the returned ColumnView.
    * @param listCol the list column to use as a template
    * @param newDataCol the new data column.
-   * @return a new ColumnView.
-   * @throws scala.IllegalArgumentException if data column does not match the original data column
-   *                                        in size.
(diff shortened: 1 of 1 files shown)


firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuListUtils.scala
```diff
@@ -30,9 +30,8 @@ object GpuListUtils extends Arm {
    * both listCol and newDataCol alive longer than the returned ColumnView.
    * @param listCol the list column to use as a template
    * @param newDataCol the new data column.
-   * @return a new ColumnView.
-   * @throws scala.IllegalArgumentException if data column does not match the original data column
-   *                                        in size.
+   * @return a new ColumnView. But throws an scala.IllegalArgumentException if data column does
```
jlowe, line 33: The only reason `scala.IllegalArgumentException` was originally used was to try to appease scaladoc.  We should just use IllegalArgumentException.
```suggestion
   * @return a new ColumnView. Throws IllegalArgumentException if data column does
```

jlowe (commented):
It would be interesting to study how Apache Spark is setting up scaladoc, as they have `@throws IllegalArgumentException` directives in their scaladoc.  It looks like they too get linkage errors for this when building under Maven, but the instructions point to using `sbt unidoc` to generate their Scaladoc which may work around the linkage problems we're seeing here.

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuListUtils.scala
firestarman: Updated

firestarman:
> It would be interesting to study how Apache Spark is setting up scaladoc, as they have `@throws IllegalArgumentException` directives in their scaladoc. It looks like they too get linkage errors for this when building under Maven, but the instructions point to using `sbt unidoc` to generate their Scaladoc which may work around the linkage problems we're seeing here.

Yes, so it is not easy to clean all the warnings. You can find more such warnings in recent nightly builds besides this one. 
Here is [another fix](https://github.com/firestarman/spark-rapids/commit/2cf3ac4bb9d28e13e14417698ca8a2cbce8f7191) for this issue, to which I prefer. It should fix all the linkage warnings for the scala lang classes/objects.
What's your idea ?

jlowe:
> Here is another fix for this issue, to which I prefer.

That approach seems fine if it fixes the issue.  Annoying that we have to do that extent of hand-holding for scaladoc, but I like that we don't have to avoid the `throws` directive for exceptions that are external to the module.

firestarman:
This is replaced by https://github.com/NVIDIA/spark-rapids/pull/3673.

PR #3626, opened 2021-09-23
Add in support for casting decimal to other number types

### LINKED ISSUE #3610: [BUG] support casts of decimal to numeric
This is something that @ttnghia found while working on tests for https://github.com/NVIDIA/spark-rapids/pull/3373, it looks like we cannot cast from decimal to other types (like Double) in his case:

```
scala> df.printSchema
root
 |-- value: decimal(5,1) (nullable = true)


scala> spark.sql("select cast(value as float) from foo").explain
21/09/22 20:56:49 WARN GpuOverrides: 
!Exec <ProjectExec> cannot run on GPU because not all expressions can be replaced
  @Expression <Alias> cast(value#0 as float) AS value#4 could run on GPU
    !Expression <Cast> cast(value#0 as float) cannot run on GPU because Cast from DecimalType(5,1) to FloatType is not supported
      @Expression <AttributeReference> value#0 could run on GPU
  *Exec <FileSourceScanExec> will run on GPU

== Physical Plan ==
*(1) Project [cast(value#0 as float) AS value#4]
+- GpuColumnarToRow false
   +- GpuFileGpuScan parquet [value#0] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex[file:/tmp/abellina], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<value:decimal(5,1)>
```

I am not entirely sure what types we want to enable, and we also need to add tests as suggested by @revans2

revans2:
This fixes #3610. I mistakenly told @ttnghia that we supported this already, but I was wrong so I am adding it in.  Not sure if this good for 21.10 or not because it was filed as a bug, but it is kind of new work also.


revans2:
build

jlowe (commented):
I see changes only to the 320 and 311until320-nondb shims.  Seems like we're missing some of the shims?

revans2:
> I see changes only to the 320 and 311until320-nondb shims. Seems like we're missing some of the shims?

looks like I need to upmerge. I grepped before, but there are new ones that I missed

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -355,20 +358,44 @@ object GpuCast extends Arm {
...
+      case (dt: DecimalType, LongType) if ansiMode =>
+        // This is a work around for https://github.com/rapidsai/cudf/issues/9282
+        val min = BigDecimal(Long.MinValue)
+            .setScale(dt.scale, BigDecimal.RoundingMode.DOWN).bigDecimal
+        val max = BigDecimal(Long.MaxValue)
+            .setScale(dt.scale, BigDecimal.RoundingMode.DOWN).bigDecimal
+        assertValuesInRange(input, Scalar.fromDecimal(min), Scalar.fromDecimal(max))
+        if (dt.precision <= DType.DECIMAL32_MAX_PRECISION && dt.scale < 0) {
+          // This is a work around for https://github.com/rapidsai/cudf/issues/9281
+          withResource(input.castTo(DType.create(DType.DTypeEnum.DECIMAL64, -dt.scale))) { tmp =>
```
abellina, line 371: Mostly curiosity, the issue linked doesn't mention negative scales, so why do we need to negate the scale here again?

revans2: cudf scale = - spark scale

abellina: thanks!

revans2:
build

revans2:
@jlowe I have addressed your review comments, but I am seeing some build errors on cdh and 313.  Not 100% sure what is happening but @tgravescs is looking into it because it appears to be on the base branch now.

sql-plugin/src/main/311until320-nondb/scala/com/nvidia/spark/rapids/shims/v2/SparkBaseShims.scala
```diff
@@ -165,7 +165,7 @@ abstract class SparkBaseShims extends Spark31XShims {
 
         // stringChecks are the same
         // binaryChecks are the same
-        override val decimalChecks: TypeSig = DECIMAL_64 + STRING
+        override val decimalChecks: TypeSig = gpuNumeric + STRING
```
jlowe, line 168: It looks like the shims are overriding this to the same behavior as the base method.  Would it be a better change to stop overriding the base version in the shims that don't need to change the behavior, or was this intentionally overridden redundantly?

revans2: Some of this was done because Spark itself changed what could be cast (dropped some things that just returned nulls). So at a minimum we had to override `sparkDecimalSig` but I thought it was cleaner to keep it all together and included `decimalChecks` with it.

jlowe (approved):

PR #3636, opened 2021-09-23
Change nightly build to not deploy dist for each classifier version [skip ci]

tgravescs:
This unfortunately adds a few more minutes per spark version but this doesn't deploy the dist package containing only the single version of spark, it will deploy all the other jars for each spark classifier version, which we want and it deploys the final dist package with all versions.

I'll try to look into a more time efficient way to do this but want to get it fixed so not pushing that dist jar.

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -29,6 +29,11 @@
     <description>Creates the distribution package of the RAPIDS plugin for Apache Spark</description>
     <version>21.10.0-SNAPSHOT</version>
 
+    <!--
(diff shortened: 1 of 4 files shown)


tgravescs:
build

tgravescs:
oops, I'm going to kill the build for this since this script not tested by premerge

jlowe (approved):

abellina (approved):

tgravescs:
build

PR #3643, opened 2021-09-24
Properly override Shims for int96Rebase [databricks]

### LINKED ISSUE #1135: [FEA] Spark3.1.0 test spark.sql.legacy.parquet.int96RebaseModeInRead
Spark 3.1.0 added some configs for int96 rebase spark.sql.legacy.parquet.int96RebaseModeInRead and spark.sql.legacy.parquet.int96RebaseModeInWrite.  Make sure we support and test accordingly

### LINKED ISSUE #3642: [BUG] Shims improperly overridden
**Describe the bug**
As part of #3330, the way I over-rode shims were a mix of the new and the old approach which caused confusion and also test errors for Spark-3.2.x

Build the plugin and test `test_ts_write_fails_datetime_exception` against Spark 3.2.x

**Expected behavior**
The above test should pass

razajafri:

* Added methods for returning the Rebase conf for different versions of Spark in their proper Shims
* No further tests were added

closes #3642 
closes #1135

### DIFF
--- a/shims/spark312/src/main/scala/com/nvidia/spark/rapids/shims/spark312/Spark312Shims.scala
+++ b/shims/spark312/src/main/scala/com/nvidia/spark/rapids/shims/spark312/Spark312Shims.scala
@@ -34,22 +34,6 @@ class Spark312Shims extends SparkBaseShims {
 
   override def hasCastFloatTimestampUpcast: Boolean = true
 
-  override def int96ParquetRebaseRead(conf: SQLConf): String = {
-    conf.getConf(SQLConf.LEGACY_PARQUET_INT96_REBASE_MODE_IN_READ)
-  }
-
-  override def int96ParquetRebaseWrite(conf: SQLConf): String = {
-    conf.getConf(SQLConf.LEGACY_PARQUET_INT96_REBASE_MODE_IN_WRITE)
-  }
-
-  override def int96ParquetRebaseReadKey: String = {
-    SQLConf.LEGACY_PARQUET_INT96_REBASE_MODE_IN_READ.key
-  }
-
-  override def int96ParquetRebaseWriteKey: String = {
-    SQLConf.LEGACY_PARQUET_INT96_REBASE_MODE_IN_WRITE.key
-  }
-
   override def getParquetFilters(
       schema: MessageType,
       pushDownDate: Boolean,
--- a/sql-plugin/src/main/301until320-nondb/scala/com/nvidia/spark/rapids/shims/v2/Spark30XShims.scala
+++ b/sql-plugin/src/main/301until320-nondb/scala/com/nvidia/spark/rapids/shims/v2/Spark30XShims.scala
@@ -50,6 +50,15 @@ trait Spark30XShims extends SparkShims {
     conf.getConf(SQLConf.LEGACY_PARQUET_REBASE_MODE_IN_READ)
   override def parquetRebaseWrite(conf: SQLConf): String =
     conf.getConf(SQLConf.LEGACY_PARQUET_REBASE_MODE_IN_WRITE)
+  override def int96ParquetRebaseRead(conf: SQLConf): String =
+    parquetRebaseRead(conf)
+  override def int96ParquetRebaseWrite(conf: SQLConf): String =
+    parquetRebaseWrite(conf)
+  override def int96ParquetRebaseReadKey: String =
+    parquetRebaseReadKey
+  override def int96ParquetRebaseWriteKey: String =
+    parquetRebaseWriteKey
+  override def hasSeparateINT96RebaseConf: Boolean = false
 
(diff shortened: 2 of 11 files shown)


razajafri:
build

sql-plugin/src/main/301until320-nondb/scala/com/nvidia/spark/rapids/shims/v2/Spark30XShims.scala
```diff
@@ -50,6 +50,15 @@ trait Spark30XShims extends SparkShims {
     conf.getConf(SQLConf.LEGACY_PARQUET_REBASE_MODE_IN_READ)
   override def parquetRebaseWrite(conf: SQLConf): String =
     conf.getConf(SQLConf.LEGACY_PARQUET_REBASE_MODE_IN_WRITE)
+  override def int96ParquetRebaseRead(conf: SQLConf): String =
```
tgravescs, line 53: this is in 301until320 but we have 31x shim below, should this be in 301until310?

razajafri: Yes, I think that makes more sense. The reason why this still works is because Spark31XShims will override this behavior. Depending on how expedient of a resolution is required as this is failing in nightly builds, should we handle this refactor as a follow-on?

tgravescs: its fine to follow on

shims/spark312/src/main/scala/com/nvidia/spark/rapids/shims/spark312/Spark312Shims.scala
```diff
@@ -34,22 +34,6 @@ class Spark312Shims extends SparkBaseShims {
 
   override def hasCastFloatTimestampUpcast: Boolean = true
 
-  override def int96ParquetRebaseRead(conf: SQLConf): String = {
```
tgravescs, line 37: these are still defined in 311 and 313 Shims.scala, I assume they can be removed?

sql-plugin/src/main/301until320-nondb/scala/com/nvidia/spark/rapids/shims/v2/Spark30XShims.scala
tgravescs: sorry changed my mind because there are other issues here I think, lets get this cleaned up

sql-plugin/src/main/311db/scala/com/nvidia/spark/rapids/shims/v2/Spark30XShims.scala
```diff
@@ -50,6 +50,15 @@ trait Spark30XShims extends SparkShims {
     conf.getConf(SQLConf.LEGACY_PARQUET_REBASE_MODE_IN_READ)
   override def parquetRebaseWrite(conf: SQLConf): String =
     conf.getConf(SQLConf.LEGACY_PARQUET_REBASE_MODE_IN_WRITE)
+  override def int96ParquetRebaseRead(conf: SQLConf): String =
```
tgravescs, line 53: these are still in shims/spark311db/src/main/scala/com/nvidia/spark/rapids/shims/spark311db/Spark311dbShims.scala

sql-plugin/src/main/301until320-nondb/scala/com/nvidia/spark/rapids/shims/v2/Spark30XShims.scala
razajafri: Looking at this more, as per current structure, I think this is correct because 301until310-nondb/SparkBaseShims extends 301until320-nondb/Spark30XShims. 

Won't it be better to do this is the other way where 301until320-nondb/SparkBaseShims extends 301until310-nondb/Spark30XShims? I don't know if we want to do this change as part of this PR though

tgravescs: ok, I get why 30XShims might be here since everything bases on it.  We might want to come up with better name and really there is SparkBaseShims that we can likely combine, so this file is fine, but the 31XShims is under 301until320-nondb and would expect it to be only 310+

razajafri: 31XShims is already in 311until320-nondb folder as part of this PR. Unless I am misreading your comment and you mean something else

tgravescs (approved):

tgravescs:
build

razajafri:
Failed due to OOM

razajafri:
build

gerashegalov (approved):

PR #3645, opened 2021-09-24
Verify unshimmed classes are bitwise-identical

### LINKED ISSUE #3582: [FEA] Verify unshimmed-base classes are identical across shims
Build should verify that the class files in the standard layout  that don't have shim classifier as part of their package name (from unshimmed-base.txt) are indeed bitwise-identical 

Our build currently assumes that it just works to use spark301 bits. Using the same mechanics as binary-dedupe.sh the build should guarantee that and fail otherwise

Defer issues to run time issues in tests.

**Additional context**
shim rework #3223

gerashegalov:
Closes #3582
- make sure that count distinct md5sum for classes moved to the top of the jar / is 1
- delete duplicates from parallel worlds after that 


gerashegalov:
build

dist/scripts/binary-dedupe.sh
```diff
@@ -75,3 +75,43 @@ mv "$SPARK3XX_COMMON_DIR" $PARALLEL_WORLDS_DIR/
...
+#
+#  So if the number of distinct class files per class in the unshimmed location is < 2, the jar
+#  is content is as expected
+#
+#  If we find an unshimmed class file occurring > 1  we fail the build and the code must be refactored
+#  until bitwise-identity of each unshimmed class is restored.
+
+# Determine the list of unshimmed class files
+UNSHIMMED_LIST_TXT=unshimmed-result.txt
+find . -name '*.class' -not -path './'$PARALLEL_WORLDS_DIR/'spark*' | \
```
tgravescs, line 103: I'm confused by this, how is this looking at unshimmed classes, this looks like its looking at all classes in the shims?  How is this not failing on the ProxyShuffle class that we were grabbing from 3.0?  I must be missing something basic here

gerashegalov: On L103 we look for all classes that are *not* in /spark*  `-not -path './'$PARALLEL_WORLDS_DIR/'spark*'` 
then for all those unshimmed classes  (including `org/apache/spark/sql/rapids/ProxyRapidsShuffleInternalManagerBase.class`) 

then using find rooted off `target` (probably should tighten to `target/parallel-world` ) we find generally all copies of every unshimmed class.
```
find . -path "./*/org/apache/spark/sql/rapids/ProxyRapidsShuffleInternalManagerBase.class"
```
we take md5sum of every copy and "count distinct" the checksums. If our assumption about using 301 for all shims is right, the count is 1, if > 1 the assumption was wrong and we fail the build.

tgravescs: my mistake, I missed the -not in here

tgravescs (dismissed):

gerashegalov:
build

gerashegalov:
build

tgravescs (approved):

PR #3646, opened 2021-09-24
Debug utility method to dump a table or columnar batch to Parquet

### LINKED ISSUE #3115: Debug utility method to dump a table or columnar batch to Parquet
Many times we need to capture some data in a `ColumnarBatch` or cudf `Table` in order to help provide input data for a libcudf repro case.  In the old days of simple schemas, it was easy to just call `Table.writeParquet(java.io.File)` to do this, but now with nested type schemas it's much more complicated to create a file to capture the data for a table or columnar batch.

A static utility method that can take a Table or ColumnarBatch along with a filename prefix and produce a Parquet file containing the contents of the table.  Column names should be automatically generated based on the schema of the data -- names are not important for this use-case, it's all about recreating the cudf table from the Parquet file later.  The utility method should return the absolute path to the created Parquet file so the caller can log it in a custom message.

The utility method should handle file collisions from other threads/executors and retry to generate a unique filename.

res-life:
This fixes #3115 


res-life:
@jlowe  @firestarman It's a draft PR, help to review.
Only implemented dump columnar batch, not implemented dump cudf Table.

revans2 (commented):
I am a little conflicted here. On one hand I would like to be able to write out a table, not just a ColumnarBatch (because I think it is more likely that we are debugging some intermediate computation that will be in a table form and not a ColumnarBatch form.  But on the other hand I really like that you were able to reuse so much existing code so there is less of a maintenance issue.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DumpUtils.scala
```diff
@@ -0,0 +1,120 @@
...
+  def dumpToParquetFile(columnarBatch: ColumnarBatch, filePrefix: String): Option[String] = {
+    try {
+      if(columnarBatch.numCols() == 0) {
+        logWarning("dump to parquet failed, has no column, file prefix is " + filePrefix)
+        None
+      } else {
+        Some(dumpToParquetFileImpl(columnarBatch, filePrefix))
+      }
+    } catch {
+      case e: Exception =>
```
jlowe, line 52: Why is this suppressing exceptions?  Do we think it's common that we wouldn't mind if the data fails to write?  I'd rather have the caller suppress exceptions when desired if that's not the common case.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DumpUtils.scala
```diff
@@ -0,0 +1,120 @@
...
+    val jobId = new JobID(createJobString, 0)
+    val taskId = new TaskID(jobId, TaskType.MAP, 0)
+    val taskAttemptId = new TaskAttemptID(taskId, 0)
+    val taskAttemptContext: TaskAttemptContext = {
+      // Set up the configuration object
+      val hadoopConf = new Configuration()
+      ParquetWriteSupport.setSchema(schema, hadoopConf)
+      hadoopConf.setBoolean(SQLConf.PARQUET_WRITE_LEGACY_FORMAT.key, false)
+      hadoopConf.set(SQLConf.PARQUET_OUTPUT_TIMESTAMP_TYPE.key,
+        ParquetOutputTimestampType.INT96.toString)
```
jlowe, line 94: Timestamps should be written with `TIMESTAMP_MICROSECONDS` which means there's no conversions necessary from the internal format.  INT96 requires conversions and we only use it because Spark's defaults to it.  This is a debug tool, thus we want ideally no conversions applied to the data.

res-life:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DumpUtils.scala
res-life: done, will throw if exception occurs

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DumpUtils.scala
res-life: done, dump now avoid any conversion.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DumpUtils.scala
```diff
@@ -0,0 +1,120 @@
+/*
+ * Copyright (c) 2020-2021, NVIDIA CORPORATION.
```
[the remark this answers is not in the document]
res-life: done

res-life:
@revans2  added dump code for CUDF table, please help to review
@jlowe please help to review
@firestarman please help to review

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarOutputWriter.scala
```diff
@@ -180,3 +166,27 @@ abstract class ColumnarOutputWriter(path: String, context: TaskAttemptContext,
...
 }
+
+object ColumnarOutputWriter {
+  // write buffers to outputStream via tempBuffer and close buffers
+  def writeBufferedData(buffers: mutable.Queue[(HostMemoryBuffer, Long)],
+      tempBuffer: Array[Byte], outputStream: OutputStream): Unit = {
+    val toProcess = buffers.dequeueAll(_ => true)
+    try {
+      toProcess.foreach(ops => {
+        val buffer = ops._1
```
revans2, line 177: nit:

I know this is copy and paste so you can keep it as is if you want, but...

```
toProcess.foreach { case (buffer, len) =>
  ...
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarOutputWriter.scala
```diff
@@ -180,3 +166,27 @@ abstract class ColumnarOutputWriter(path: String, context: TaskAttemptContext,
...
+        while (len > 0) {
+          val toCopy = math.min(tempBuffer.length, len).toInt
+          buffer.getBytes(tempBuffer, 0, offset, toCopy)
+          outputStream.write(tempBuffer, 0, toCopy)
+          len = len - toCopy
+          offset = offset + toCopy
+        }
+      })
+    } finally {
+      toProcess.map(_._1).safeClose()
```
revans2, line 189: nit: I know this is copy and paste so you can keep it as is if you want but ...

```
toProcess.values().safeClose()
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DumpUtils.scala
```diff
@@ -0,0 +1,285 @@
...
+    } else {
+      Some(dumpToParquetFileImp(table, filePrefix))
+    }
+  }
+
+  private def dumpToParquetFileImp(table: Table, filePrefix: String): String = {
+    val path = genPath(filePrefix)
+
+    // write data, Note not close the table
+    val dumper = new GpuTableToParquetDumper(path, table)
```
revans2, line 72: nit:

```
withResource(new GpuTableToParquetDumper(path, table)) { dumper =>
    dumper.writeTable(table)
    path
}
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DumpUtils.scala
```diff
@@ -0,0 +1,285 @@
...
+
+  private def dumpToParquetFileImpl(columnarBatch: ColumnarBatch, filePrefix: String): String = {
+    // generate path, retry if file exists
+    val path = genPath(filePrefix)
+
+    // construct schema according field types
+    val schema = genSchema(columnarBatch)
+
+    // write data, Note not close the column
+    val dumper = new GpuColumnBatchToParquetDumper(path, schema)
```
revans2, line 91: nit:

```
withResource(new GpuTableToParquetDumper(path, schema)) { dumper =>
    dumper.writeTable(columnarBatch)
    path
}
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DumpUtils.scala
```diff
@@ -0,0 +1,285 @@
...
+    ColumnarOutputWriter.writeBufferedData(buffers, tempBuffer, outputStream)
+  }
+
+  def writeBatch(batch: ColumnarBatch): Long = {
+    val startTimestamp = System.nanoTime
+    withResource(GpuColumnVector.from(batch)) { table =>
+      tableWriter.write(table)
+    }
+
+    GpuSemaphore.releaseIfNecessary(TaskContext.get)
```
revans2, line 179: This should not be here. We are debugging things. It should have no impact on the semaphore

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DumpUtils.scala
```diff
@@ -0,0 +1,285 @@
...
+    val gpuTime = System.nanoTime - startTimestamp
```
revans2, line 180: nit: I don't see why we need to time how long it takes to write the data out.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DumpUtils.scala
```diff
@@ -0,0 +1,285 @@
...
+    writeBufferedData()
+    gpuTime
+  }
+
+  def writeTable(table: Table): Long = {
+    val startTimestamp = System.nanoTime
+
+    val columns = new Array[ColumnVector](table.getNumberOfColumns)
```
revans2, line 188: Why are we copying the table? There is no need to close it.  Why not just.

```
def writeTable(table: Table): Unit = {
  tableWriter.write(table)
  writeBufferedData()
}
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DumpUtils.scala
```diff
@@ -0,0 +1,285 @@
...
+
+  override
+  def handleBuffer(buffer: HostMemoryBuffer, len: Long): Unit =
+    buffers += Tuple2(buffer, len)
+
+  def writeBufferedData(): Unit = {
+    ColumnarOutputWriter.writeBufferedData(buffers, tempBuffer, outputStream)
+  }
+
+  def writeBatch(batch: ColumnarBatch): Long = {
```
revans2, line 173: This really should just be the following to reduce any kind of duplication in the code.

```
withResource(GpuColumnVector.from(batch)) { table =>
    writeTable(table)
}
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DumpUtils.scala
```diff
@@ -0,0 +1,285 @@
...
+  /**
+   * Debug utility to dump table to parquet file. <br>
+   * It's running on GPU. Parquet column names are generated from table column type info. <br>
+   *
+   * @param table      the table to be dumped, should be GPU columnar batch
+   * @param filePrefix parquet file prefix, e.g. /tmp/my-debug-prefix-
+   * @return parquet file path if dump is successful, e.g. /tmp/my-debug-prefix-123.parquet
+   */
+  def dumpToParquetFile(table: Table, filePrefix: String): Option[String] = {
+    logWarning("my-debug begin dump")
```
revans2, line 59: nit: can we remove the debug logging messages?

revans2 (commented):
This feels really overly complicated.  Would it be better to just normalize everything to a prefix, schema, and table? Then if it is a ColumnarBatch we have the schema, and can translate it to a table without much effort. And for the table we have a helper method, similar to `parquetWriterOptionsFromTable` that will generate a schema from the table?

res-life:
build

res-life:
@revans2  addressed all your comments, please review again.

revans2 (approved):

Issue #2354, opened 2021-05-06
[FEA] Support a sort merge join as a fallback on the GPU


revans2:
This work should happen after the basic out of core join is implemented that just allows for chunking of the output of an exploding join #20.

Hash joins are always going to be faster than sort merge joins. Because there is no need to sort the data. But sort merge joins are really the only option in some cases when we want to do a join for data that is large on both sides of the join. Also a FullOuter join cannot work unless all of the data for a given key is a part of the current join, and a sort merge join can handle that, where as a hash join cannot unless all of the data for both sides is in memory.  

The idea is that we start off trying to do a regular hash join. We read in the build side table until it hits a size limit (probably the target batch size). If the build table is too large then we switch over to doing a sort merge join.

In the case of a FullOuter join we also need to batch the stream side (because there really is no stream side).  It would be similar that we would read in the stream side table until we got all of the data or we hit a size limit. If we go over the size limit, then we switch over to a sort merge join.

To switch to a sort merge join we will create a sorting iterator for each side (build side and stream side).  We will also need to insert into it any data that we have already read/cached.  After that we get a batch from each side, and look at the last row join keys in each batch. We then do a lower bound on both batches on both sets of keys to see how much we can keep to do the join (really it should be the lowest bound for the lowest keys, but instead of trying to compare keys on the CPU we can just let the GPU do it and compare the offsets). We then split the batches based off of the lowest point for either set of keys, save the data that is not ready (everything above that cutoff) so it can be combined with the next batch we read and do the join on the data that is ready (everything below the cutoff).

In some cases we might end up with an empty table on one side or the other. That is OK. We just do the join and go on. If we end up with an empty table on both sides that means we have a key that is large enough it is taking up an entire batch. Just skip the join and continue to read the next batch, and hopefully we have enough memory/rows to make it work.

If we start to run into a lot of cases where we have keys that do not fit in a single batch, we might be able to add in some other optimizations. For all of the joins except FullOuter if we have a full set of keys on the build side we can do the join without waiting for all of the keys on the stream side to be there. But then we have to also make sure that we do not throw out that data on the build side until all of the corresponding keys have been seen on the stream side.

For InnerLike joins we can also do the equivalent of a cross join for a given key. do a join for each batch on the left hand side pared with each batch on the right hand side.

But both of these optimizations are complicated and hopefully unnecessary.

revans2:
[...]

PR #3657, opened 2021-09-24
Revert "Use cudf to compute exact hash join output row sizes (#3288)"

### LINKED ISSUE #3640: [BUG] q82 regression after #3288
It looks like the changes in this PR: https://github.com/NVIDIA/spark-rapids/pull/3288 are causing a regression in performance for q82 (to several minutes). 

I noticed a lot of time spent in the following code and @jlowe suggested to revert. After doing so, we are able to get back to ~20 seconds.

```
ai.rapids.cudf.HashJoin.create(Native Method)
ai.rapids.cudf.HashJoin.<init>(HashJoin.java:87)
org.apache.spark.sql.rapids.execution.HashJoinIterator.$anonfun$maybeBuildHashTable$3(GpuHashJoin.scala:358)
org.apache.spark.sql.rapids.execution.HashJoinIterator$$Lambda$2420/1713476990.apply(Unknown Source)
```

jlowe:
This reverts commit 25bad3d95ab3c464eb04df46e8a9300f81e75de3.

Fixes #3640.  When we switched to building the hash table explicitly, we lost the ability to be dynamic with which table is used as the build-side table for an inner join.  It's definitely something we can do ourselves, but it will be tricky to do properly given how the join code assumes the table designated as the build-side will be used for a hash and the stream side is the only one that is splittable.

Since we're in the process of finishing up 21.10, I think it's prudent to revert #3288 and tackle this in a future release. 
 #2354 tracks solving the real root issue which is requiring everything on an arbitrarily-chosen build side of an inner join to be pulled in at once.  Ideally this should try to examine both sides to choose the best build side and have a fallback if we cannot pull in either table completely to build the final hash table.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
@@ -49,12 +49,6 @@ abstract class AbstractGpuJoinIterator(
   protected def hasNextStreamBatch: Boolean
 
   /**
-   * Called when a result batch is about to be produced. Any custom resources that were allocated
-   * to produce a result should be freed or made spillable here.
-   */
-  protected def freeIntermediateResources(): Unit = {}
-
-  /**
    * Called to setup the next join gatherer instance when the previous instance is done or
    * there is no previous instance.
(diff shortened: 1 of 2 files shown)


revans2 (approved):

jlowe:
build

PR #3663, opened 2021-09-24
Add back the tests-spark310+

tgravescs:
At least some of these seem to be real unit tests because they access internal functions so pull the aggregator jar here instead of the final dist jar. We may want to consider moving these into a sql-plugin test directory but for now make them run again on all non databricks 311+ builds.


tgravescs:
build

jlowe (approved):

PR #3666, opened 2021-09-24
Recompute Parquet block metadata when estimating footer from multiple file input

### LINKED ISSUE #3516: [BUG] MultiFileParquetPartitionReader can fail while trying to write the footer
**Describe the bug**
I was trying to test out #3509 and was not able to reproduce it locally. Instead on 3.2 I ran into some issues with reading parquet data.  Reading works fine when using PERFILE, but when I set it to AUTO or COALESCING it fails with errors in parquet.

It turns out that when we want to get the size of the footer we write out the footer, but newer versions of parquet actually check if our footer looks correct, no overlap in the blocks.  So this fails because we are using the blocks unmodified from different input files.

On Spark 3.2.1-SNAPSHOT (3.2.0) run locally with 1 executor and maxPartitionBytes of 1g.  Then setup TPCDS for Scale Factor 200 on parquet and run the query.

```
spark.time(spark.sql("select count(*) from store_sales, date_dim d1 where ss_sold_date_sk = d1.d_date_sk and d1.d_year between 1999 AND 1999 + 2").show)
```

It fails 100% of the time for me with errors like 

```
Caused by: java.lang.IllegalStateException: Invalid block starting position:2106143682
  at org.apache.parquet.Preconditions.checkState(Preconditions.java:93)
  at org.apache.parquet.format.converter.ParquetMetadataConverter.toParquetMetadata(ParquetMetadataConverter.java:216)
  at org.apache.parquet.format.converter.ParquetMetadataConverter.toParquetMetadata(ParquetMetadataConverter.java:197)
  at com.nvidia.spark.rapids.ParquetPartitionReaderBase.writeFooter(GpuParquetScan.scala:517)
  at com.nvidia.spark.rapids.ParquetPartitionReaderBase.writeFooter$(GpuParquetScan.scala:486)
```

**Expected behavior**
It will pass

I am not 100% sure how to fix this. I could go muck around with the blocks passed in, but the starting offset comes from the first column's metadata, which means I would have to mess with each column's metadata too.

jlowe:
Fixes #3516 

Computes realistic Parquet block metadata when writing a footer to estimate its final output size.  Refactored the existing code that updates the block metadata as it copies blocks to the computed Parquet file so the code can be reused without copying blocks.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
@@ -57,12 +57,11 @@ import org.apache.spark.sql.execution.datasources.parquet.ParquetReadSupport
 import org.apache.spark.sql.execution.datasources.v2.FilePartitionReaderFactory
 import org.apache.spark.sql.execution.datasources.v2.parquet.ParquetScan
 import org.apache.spark.sql.internal.SQLConf
-import org.apache.spark.sql.internal.SQLConf.LegacyBehaviorPolicy
 import org.apache.spark.sql.rapids.execution.TrampolineUtil
 import org.apache.spark.sql.sources.Filter
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.vectorized.ColumnarBatch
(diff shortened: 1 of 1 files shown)


jlowe:
build

revans2 (approved):

PR #3680, opened 2021-09-27
cudaMalloc UCX bounce buffers when async allocator is used

### LINKED ISSUE #3603: [FEA] Allocate UCX bounce buffers outside of RMM if ASYNC allocator is enabled
Depends on: https://github.com/rapidsai/cudf/issues/9270

Given the option to enable the ASYNC allocator (https://github.com/NVIDIA/spark-rapids/pull/3447), UCX in this mode will currently fail when attempting to use `nv_peer_mem`. The reason is that memory allocated with `cudaMallocAsync` does not support GPUDirect RDMA, and the pool we go to allocate memory from would be using this allocator exclusively. 

When https://github.com/rapidsai/cudf/issues/9270 goes in, we should be able to reserve the bounce buffer amount from the RMM pool, and allocate directly, bypassing RMM. This would allow us to use `nv_peer_mem` and UCX as we do today with other RMM allocators.

rongou:
Fixes #3603

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala
@@ -173,7 +173,8 @@ object GpuDeviceManager extends Logging {
 
     val minAllocation = truncateToAlignment((conf.rmmAllocMinFraction * info.total).toLong)
     val maxAllocation = truncateToAlignment((conf.rmmAllocMaxFraction * info.total).toLong)
-    val reserveAmount = if (conf.shuffleManagerEnabled && conf.rmmPool.equalsIgnoreCase("ASYNC")) {
+    val reserveAmount = if (GpuShuffleEnv.shouldUseRapidsShuffle(conf)
+        && conf.rmmPool.equalsIgnoreCase("ASYNC")) {
       // When using the async allocator, UCX calls `cudaMalloc` directly to allocate the
       // bounce buffers.
       conf.rmmAllocReserve + conf.shuffleUcxBounceBuffersSize * 2
@@ -223,6 +224,11 @@ object GpuDeviceManager extends Logging {
     if (!Rmm.isInitialized) {
       val conf = rapidsConf.getOrElse(new RapidsConf(SparkEnv.get.conf))
       val info = Cuda.memGetInfo()
+
+      // We need to reserve more memory when RAPIDS shuffle is enabled and we are using the CUDA
+      // async allocator, so initialize the shuffle environment first.
+      GpuShuffleEnv.init(conf)
+
       val (initialAllocation, maxAllocation) = computeRmmInitSizes(conf, info)
       var init = RmmAllocationMode.CUDA_DEFAULT
       val features = ArrayBuffer[String]()
@@ -298,7 +304,6 @@ object GpuDeviceManager extends Logging {
         Rmm.initialize(
           init, logConf, initialAllocation, maxAllocation, allocationAlignment, alignmentThreshold)
         RapidsBufferCatalog.init(conf)
-        GpuShuffleEnv.init(conf)
       } catch {
         case e: Exception => logError("Could not initialize RMM", e)
       }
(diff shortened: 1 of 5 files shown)


rongou:
build

abellina (changes requested):
I think we should change this too: https://github.com/NVIDIA/spark-rapids/blob/branch-21.10/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala#L176

Would we not allocate some memory outside of the limiting tracker and potentially cause an OOM?

rongou:
> I think we should change this too: https://github.com/NVIDIA/spark-rapids/blob/branch-21.10/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala#L176
> 
> Would we not allocate some memory outside of the limiting tracker and potentially cause an OOM?

Done.

abellina:
So actually the patch you have is going to do this by default (https://github.com/NVIDIA/spark-rapids/pull/3680/commits/482931a691bf710aa8e5b5913da07696fedc72da), reason being that the shuffle enabled flag is true by default. It is used to disable the shuffle manager at runtime, and so it should not be relied on for this. 

We could check `GpuShuffleEnv.isRapidsShuffleConfigured`. If this is true, it seems to me the user is trying to use the UCX shuffle, so we should be good to assume we are reserving this memory. Currently this method is part of an instance of `GpuShuffleEnv`, we could refactor it to the `object`, but need to be careful with the places that use this.

For now we could just document that a reserve should be give, then have a follow on PR to tackle the above. We should revert  https://github.com/NVIDIA/spark-rapids/pull/3680/commits/482931a691bf710aa8e5b5913da07696fedc72da, however. We could change the order of initialization to something like:

1. If UCX is configured lets load it (GpuShuffleEnv.init). This should by default reach out to the new direct allocation method.
2. Initialize Rmm. "free" should be less given memory allocated in (1), or we'll need check if the shuffle is configured.

rongou:
Ok, moved the init up so we can check if we should use rapids shuffle.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala
```diff
@@ -223,6 +224,11 @@ object GpuDeviceManager extends Logging {
     if (!Rmm.isInitialized) {
       val conf = rapidsConf.getOrElse(new RapidsConf(SparkEnv.get.conf))
       val info = Cuda.memGetInfo()
+
+      // We need to reserve more memory when RAPIDS shuffle is enabled and we are using the CUDA
+      // async allocator, so initialize the shuffle environment first.
+      GpuShuffleEnv.init(conf)
```
abellina, line 230: but this is happening for all allocators right? I think this is probably where you want this for ASYNC, but not for the others?

rongou: We need to always initialize it regardless of the allocator used, right?

rongou:
build

rongou:
@abellina please take another look.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala
```diff
@@ -217,6 +224,11 @@ object GpuDeviceManager extends Logging {
     if (!Rmm.isInitialized) {
       val conf = rapidsConf.getOrElse(new RapidsConf(SparkEnv.get.conf))
       val info = Cuda.memGetInfo()
+
+      // We need to reserve more memory when RAPIDS shuffle is enabled and we are using the CUDA
```
abellina, line 228: So the `GpuShuffleEnv.init(conf)`  doesn't end up allocating the bounce buffers. We are accounting for the reserve up above, so I think the init call should just return to what it was before.

What do you think?

rongou: We are calling `GpuShuffleEnv.shouldUseRapidsShuffle(conf)` before initializing RMM, which requires `GpuShuffleEnv` to be initialized.

abellina: That's absolutely right.

abellina (approved):

PR #3681, opened 2021-09-27
Clean up and document metrics [databricks]

### LINKED ISSUE #2527: [FEA] clean up timing metrics and make them consistent

### LINKED ISSUE #3243: Remove total time metrics from most GPU nodes in the plan
Many exec nodes in the plan have a "total time" metric that records the entire time spent in that node, including time spent waiting for the input iterator(s) `hasNext` and `next` calls.  I'm not sure this is very useful in practice, and many users think this metric represents time spent solely operating on the node, when often it is more measuring the time spent waiting for inputs.  Almost all the exec nodes have a separate metric that records the time spent locally operating on the GPU which in practice is much more useful.

Given we've seen cases where too many metrics have been generated while executing the plan, we should consider removing this one across the board, or at least only keeping it for a select few exec nodes where it makes sense.

revans2:
This fixes #3243 
This fixes #2527

This removes total time and replaces it with `op time`, among other metrics. I am open to modifying/adding in other metrics like in #2527.

### DIFF
--- a/build/buildall
+++ b/build/buildall
@@ -39,7 +39,7 @@ SPARK_SHIM_VERSIONS="
 BUILD_PARALLELISM=1
 
 time (
-echo -n "$SPARK_SHIM_VERSIONS" | xargs -I% -P $BUILD_PARALLELISM -n 1 \
+echo -n "$SPARK_SHIM_VERSIONS" | xargs -t -I% -P $BUILD_PARALLELISM -n 1 \
   bash -c "
   mvn -U clean install \
     -Dbuildver=% \
--- a/docs/tuning-guide.md
+++ b/docs/tuning-guide.md
@@ -134,12 +134,11 @@ Default value: `1`
 
 The RAPIDS Accelerator can further limit the number of tasks that are actively sharing the GPU.
 It does this using a semaphore. When metrics or documentation refers to the GPU semaphore it
-is referring to this.
-This restriction is useful for avoiding GPU out of memory errors while still allowing full
-concurrency for the
-portions of the job that are not executing on the GPU. Care is taken to try and avoid doing
-I/O or other CPU operations while the GPU semaphore is held. But in the case of a join two batches
-are required for processing and it is not always possible to avoid this case.
+is referring to this. This restriction is useful for avoiding GPU out of memory errors while 
+still allowing full concurrency for the portions of the job that are not executing on the GPU. 
+Care is taken to try and avoid doing I/O or other CPU operations while the GPU semaphore is held. 
+But in the case of a join two batches are required for processing, and it is not always possible 
+to avoid this case.
 
 Some queries benefit significantly from
 setting this to a value between `2` and `4`, with `2` typically providing the most benefit, and
--- a/docs/tuning-guide.md
+++ b/docs/tuning-guide.md
@@ -133,8 +133,15 @@ Configuration key: [`spark.rapids.sql.concurrentGpuTasks`](configs.md#sql.concur
 Default value: `1`
 
 The RAPIDS Accelerator can further limit the number of tasks that are actively sharing the GPU.
-This is useful for avoiding GPU out of memory errors while still allowing full concurrency for the
-portions of the job that are not executing on the GPU.  Some queries benefit significantly from
+It does this using a semaphore. When metrics or documentation refers to the GPU semaphore it
+is referring to this.
+This restriction is useful for avoiding GPU out of memory errors while still allowing full
+concurrency for the
+portions of the job that are not executing on the GPU. Care is taken to try and avoid doing
+I/O or other CPU operations while the GPU semaphore is held. But in the case of a join two batches
+are required for processing and it is not always possible to avoid this case.
+
+Some queries benefit significantly from
 setting this to a value between `2` and `4`, with `2` typically providing the most benefit, and
 higher numbers giving diminishing returns, but a lot of it depends on the size of the GPU you have.
 An 80 GiB A100 will be able to run a lot more in parallel without seeing degradation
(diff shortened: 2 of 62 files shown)


revans2:
build

revans2:
build

build/buildall
```diff
@@ -39,7 +39,7 @@ SPARK_SHIM_VERSIONS="
 BUILD_PARALLELISM=1
 
 time (
-echo -n "$SPARK_SHIM_VERSIONS" | xargs -I% -P $BUILD_PARALLELISM -n 1 \
+echo -n "$SPARK_SHIM_VERSIONS" | xargs -t -I% -P $BUILD_PARALLELISM -n 1 \
```
abellina, line 42: nit, debug?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -56,32 +61,38 @@ abstract class AbstractGpuJoinIterator(
...
-      } else if (hasNextStreamBatch) {
-        // Need to refill the gatherer
-        gathererStore.foreach(_.close())
-        gathererStore = None
-        gathererStore = setupNextGatherer(startNanoTime)
-        nextCb = nextCbFromGatherer()
       } else {
-        mayContinue = false
+        // The only part we want to skip in the timing is hasNextStreamBatch
+        val tmp = timedHasNextStreamBatch
```
abellina, line 78: nit, would be nice to decompose the return of `timedHasNextStreamBatch` here.

build/buildall
revans2: From just looking at the build log it is very hard to tell what version of Spark is being built against. This gives a very clear clue.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
```diff
@@ -371,28 +371,44 @@ case class GpuFilterExec(condition: Expression, child: SparkPlan)
 /**
  * Physical plan for range (generating a range of 64 bit numbers).
  */
-case class GpuRangeExec(range: org.apache.spark.sql.catalyst.plans.logical.Range,
+case class GpuRangeExec(
```
abellina, line 374: this one seems different than other changes in this PR. Just curious on why the change.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCartesianProductExec.scala
```diff
@@ -71,7 +70,8 @@ class GpuSerializableBatch(batch: ColumnarBatch)
   }
 
   private def readObject(in: ObjectInputStream): Unit = {
-    GpuSemaphore.acquireIfNecessary(TaskContext.get())
+    // TODO I don't know of a way to make this work...
```
abellina, line 73: needs follow up issue?

build/buildall
abellina: ack, sounds good

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -36,8 +35,8 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
 abstract class AbstractGpuJoinIterator(
     gatherNvtxName: String,
     targetSize: Long,
-    joinTime: GpuMetric,
-    totalTime: GpuMetric) extends Iterator[ColumnarBatch] with Arm with AutoCloseable {
+    opTime: GpuMetric,
```
abellina, line 38: the scaladoc needs updating in this constructor

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -144,14 +155,13 @@ abstract class SplittableJoinIterator(
     builtBatch: LazySpillableColumnarBatch,
     targetSize: Long,
     spillCallback: SpillCallback,
-    joinTime: GpuMetric,
-    streamTime: GpuMetric,
-    totalTime: GpuMetric)
+    opTime: GpuMetric,
+    joinTime: GpuMetric)
```
abellina, line 159: scaladoc changed needed on this one too

abellina (commented):
Looks good to me, just nits.

docs/tuning-guide.md
```diff
@@ -133,8 +133,15 @@ Configuration key: [`spark.rapids.sql.concurrentGpuTasks`](configs.md#sql.concur
 Default value: `1`
 
 The RAPIDS Accelerator can further limit the number of tasks that are actively sharing the GPU.
-This is useful for avoiding GPU out of memory errors while still allowing full concurrency for the
-portions of the job that are not executing on the GPU.  Some queries benefit significantly from
+It does this using a semaphore. When metrics or documentation refers to the GPU semaphore it
+is referring to this.
+This restriction is useful for avoiding GPU out of memory errors while still allowing full
+concurrency for the
```
jlowe, line 139: Nit: the wrapping here looks odd if one is examining the .md file directly via a terminal rather than via github's rendering.

jlowe (commented):
Overall looks good to me, just a small formatting nit in addition to the other comments.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
revans2: I inlined `logical.Range` into it. It is not needed, but made the code a little more robust in my opinion. I can remove it if we want to.

docs/tuning-guide.md
revans2: Feel free to ignore my rant. I will make the change. Really? One of the reasons for mark down is that it hides this when rendered. I have a similar issue just a few lines below. That one is there mostly because I want the diff to be small for reviewing. If we care enough to do it here why not do it everywhere and forget about the diff?

revans2:
build

docs/tuning-guide.md
jlowe: One of the main advantages of markdown over other doc formats is that is is pretty human-readable in a terminal without requiring rendering in a web-page or other tool, so I'd like to keep it looking nice for those that are looking at things offline.  It is just a nit, and I may be in the minority feeling this way about our markdown formatting.

jlowe (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
abellina: no that's fine just curious

abellina (approved):

Issue #3152, opened 2021-08-05
[FEA] support nested structs in collect_set


revans2:
**Is your feature request related to a problem? Please describe.**
I filed the cudf issue https://github.com/rapidsai/cudf/issues/8972 to get cudf support for nested structs in collect_set and merge_set. When that is done we need to update the tests and the type checks for collect_set and verify that it works.

ttnghia:
I'm not sure if this is duplicate issue: https://github.com/NVIDIA/spark-rapids/issues/3514

ttnghia:
FYI: The cudf issue is addressed. Now we can start plugin work.

PR #3700, opened 2021-09-29
CollectSet supports structs

wbo4958:
This is to fix https://github.com/NVIDIA/spark-rapids/issues/3514 and https://github.com/NVIDIA/spark-rapids/issues/3152. Currently, with this PR, the CollectSet on structs can run on GPU. Since CPU and GPU can't ensure the order of collect-set, we need to add sort_array on collect-set.  But for now, sort-array didn't support on nested type, I just xfail the newly added collect-set test on structs.

### DIFF
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -422,6 +422,12 @@ _gen_data_for_collect_list_op = _gen_data_for_collect_op + [[
     ('a', RepeatSeqGen(LongGen(), length=20)),
     ('b', value_gen)] for value_gen in _repeat_agg_column_for_collect_list_op]
 
+_repeat_agg_column_for_collect_set_op = [ RepeatSeqGen(all_basic_struct_gen, length=15) ]
+
+_gen_data_for_collect_set_op = [[
+    ('a', RepeatSeqGen(LongGen(), length=20)),
+    ('b', value_gen)] for value_gen in _repeat_agg_column_for_collect_set_op]
+
 # to avoid ordering issues with collect_list we do it all in a single task
 @ignore_order(local=True)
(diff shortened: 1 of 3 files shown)


wbo4958:
build

razajafri:
We should test this on DB as well?

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -444,6 +450,18 @@ def test_hash_groupby_collect_set(data_gen):
             .groupby('a')
             .agg(f.sort_array(f.collect_set('b')), f.count('b')))
 
+@approximate_float
+@ignore_order(local=True)
+@incompat
+@pytest.mark.parametrize('data_gen', _gen_data_for_collect_set_op, ids=idfn)
+@pytest.mark.xfail(reason="the result order from collect-set can not be ensured for CPU and GPU."
```
revans2, line 457: Is there a follow on issue for this in particular?  I personally would prefer a skip over an xfail, or is there a way that we can do a simple small test where there will only ever be a single value in the set so we can at least test that case.

wbo4958:
build

integration_tests/src/main/python/hash_aggregate_test.py
wbo4958: Done, the new commit has added a new test case with unique group by key for this.

wbo4958:
build

wbo4958:
> We should test this on DB as well?

Done

sameerz:
Please retarget to branch-21.12.

integration_tests/src/main/python/hash_aggregate_test.py
wbo4958: @revans2 Could you help to review it again? Thx

revans2 (approved):

PR #3691, opened 2021-09-28
Fix issues with AQE and DPP enabled on Spark 3.2 [databricks]

jlowe:
Fixes #3653.

This fixes a number of issues with `GpuBroadcastToCpuExec`.  In Spark 3.2+, there's an assert to verify that the child of an `AdaptiveSparkPlanExec` is a `BroadcastQueryStageExec` if it tries to compute the broadcast.  The transition plans would place `GpuBroadcastToCpuExec` immediately under `AdaptiveSparkPlanExec` which triggered the assert.

Unfortunately creating a new `BroadcastQueryStageExec` instance requires shimming, since the signature of `BroadcastQueryStageExec` has changed across Spark versions.  Newer versions require a canonicalized plan parameter, so I needed to add a shim method to create a new instance given an old one and a new child to place underneath.

Besides fixing that assert, there was also a mishandling of the broadcast data in `GpuBroadcastToCpuExec`.  The code assumed that `ColumnarBatch.rowIterator` would return a unique row per iteration, but it returns _the same row object instance_ for each iteration and mutates that object's internal state to produce the proper row when the row's data is requested.  The code was producing an array of `InternalRow` without manifesting the data from the row, which caused each entry to be the same `ColumnarBatchRow` instance, pointing at the last row of data.  In addition the `UnsafeRow` instances were being reused in a similar manner.  The fix is to not force the internal rows to manifest in an array until they are being converted to an unsafe row, and when converting we need to make a copy of each unsafe row.

One last issue, which ended up being the majority of the change in this PR, is that `BroadcastQueryStageExec` expects a `BroadcastExchangeLike` child, so to fix the original assert, `GpuBroadcastToCpuExec` needs to be a `BroadcastExchangeLike`.  Unfortunately `BroadcastExchangeLike` was forcing shims due to Databricks changing the signature of that class.  Rather than shim yet another class that needs `BroadcastExchangeLike`, I ended up creating a shim v2 version of `BroadcastExchangeLike` called `ShimBroadcastExchangeLike` that encapsulates the differences between Apache Spark and Databricks with this class.  This not only allows us to avoid shimming `GpuBroadcastToCpuExec` but also allows us to unshim `GpuBroadcastExchangeExec` which was only being shimmed due to this trait signature difference.

PR #3719, opened 2021-09-30
Advertise CPU sort order and partitioning expressions to Catalyst [databricks]

### LINKED ISSUE #3713: [BUG] AQE shuffle coalesce optimization is broken with Spark 3.2
[SPARK-36315](https://issues.apache.org/jira/browse/SPARK-36315) added explicit `Partitioning` case class checks for updating the output distribution for coalesced shuffle reads.  Unfortunately since the RAPIDS Accelerator provides GPU versions of these `Partitioning` case classes, the code fails to recognize the partitioning and classifies it as `UnknownPartitioning`.  This often causes validation of the shuffle-coalesced optimized plan to fail since it appears the shuffle does not provide the required partitioning.

This ends up manifesting as stages run on the GPU that do not have a reduced partition count due to AQE shuffle coalescing that the corresponding CPU run does.

jlowe:
Fixes #3713 

Changes the GPU nodes in the Catalyst plan to advertise CPU `SortOrder` and `Partitioning` classes and child expressions.  This helps the Spark code better recognize that portions of the plan are still compatible from a sorting and partitioning perspective even after the GPU has updated the plan.  Spark has a number of places in the code where it is matching against specific Spark CPU partition classes or checking if CPU sort/partition expressions match,  Returning the CPU expressions  from the GPU nodes' `outputPartitioning`, `outputOrdering`, `requiredChildDistribution` and `requiredChildOrdering`methods helps us pass these checks in Apache Spark code.

This is accomplished by passing _both_ the GPU and CPU expressions to exec nodes that need to override their output ordering or partitioning.  It uses the GPU expressions during computation but "advertises" the CPU expressions from the Catalyst methods to query the node's intentions.


jlowe:
This depends on #3691.  Marking as a draft until the dependency is merged.

jlowe:
build

jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -396,13 +397,17 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
       // Insert a sort after the last hash-based op before the query result if there are no
       // intermediate nodes that have a specified sort order. This helps with the size of
       // Parquet and Orc files
+      // FIXME: This is passing a GPU sort order as a CPU sort order since there is not an easy
```
abellina, line 400: is this something you plan to tackle here? or a subsequent issue. would we need to `columnarEval` and turn it into a cpu literal? I assume that you can't have a column vector here (just thinking out loud)

abellina: On the same note here, should we require/assert that that the order is not a complex expression?

abellina (commented):
Overall this makes sense, I had one comment on whether we wanted to fix or add more explicit asserts if the gpu order was complex (required an eval). I am not sure what the outcome of the failure is here => failed query, or less performant.

tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
```diff
@@ -187,10 +187,11 @@ case class TargetSize(override val targetSizeBytes: Long) extends CoalesceSizeGo
...
  * that it is required that a node in the Spark plan that requires this should also require
  * an input ordering that satisfies this ordering as well.
- * @param order the keys that should be used for batching.
+ * @param gpuOrder the GPU keys that should be used for batching.
+ * @param cpuOrder the CPU keys that should be used for batching.
  */
-case class BatchedByKey(order: Seq[SortOrder]) extends CoalesceGoal {
-  override def children: Seq[Expression] = order
+case class BatchedByKey(gpuOrder: Seq[SortOrder], cpuOrder: Seq[SortOrder]) extends CoalesceGoal {
+  override def children: Seq[Expression] = gpuOrder ++ cpuOrder
```
tgravescs, line 194: this looks odd, why is it both gpu and cpu order?  perhaps a comment

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
jlowe: This isn't a runtime need-to-execute-this problem, rather a plans-match-up problem.  Note that before this PR, we're reporting potentially a GPU SortOrder expression and it isn't going to fail.  The worst that should happen is that the planner doesn't realize that some CPU nodes later in the plan that require a certain sort order already lines up with what this node is doing because CPU and GPU expressions won't semantically compare unless they're trivial attribute references.  If it doesn't compare then it will insert a redundant sort to "fix" the required ordering.

However I just noticed that this code is explicitly checking for the case where no node later in the plan specify a sort order.  Therefore this should never be a problem in practice, since nothing should care about the advertised sort order expression.  I'll update the comment to say we're cheating by advertising GPU sort expressions but explain why that's probably OK.

shims/spark320/src/main/scala/com/nvidia/spark/rapids/shims/spark320/Spark320Shims.scala
```diff
@@ -788,11 +790,12 @@ class Spark320Shims extends Spark32XShims {
   }
 
   override def getGpuShuffleExchangeExec(
-      outputPartitioning: Partitioning,
+      gpuOutputPartitioning: Partitioning,
```
revans2, line 793: Can we have a follow on issue to separate out the GpuPartitioning from the Partitioning. Or explicitly mark it as GpuPartitioning in the method signature?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
```diff
@@ -113,7 +115,7 @@ case class GpuSortExec(
   private [this] lazy val targetSize = RapidsConf.GPU_BATCH_SIZE_BYTES.get(conf)
 
   override def doExecuteColumnar(): RDD[ColumnarBatch] = {
-    val sorter = new GpuSorter(sortOrder, output)
+    val sorter = new GpuSorter(gpuSortOrder, output)
```
revans2, line 118: Could we have a follow on issue to look at what to do with GpuSorter if we know the CPU sort order too? I did some less than ideal things with materializing columns when the we had to do out of core sort for a GPU sort order that required computation. I am not sure if we can remove that or not, but it would be good to investigate.

revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
jlowe: I need to look into this further.  I originally left it as just the GPU SortOrder, but if I do then it leads to an `IndexOutOfBoundsException` during WindowFunctionSuite.  It's as if something sees the case class arguments, expects two children from the `children` result, but only finds one.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuSortExec.scala
jlowe: Filed #3731

shims/spark320/src/main/scala/com/nvidia/spark/rapids/shims/spark320/Spark320Shims.scala
jlowe: Good catch, I meant to update the signature for `GpuPartitioning` and forgot to go back to update.  I'll at least do that in this PR and file a followup to separate them further to avoid accidentally passing a `GpuPartitioning` as a `Partitioning`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
jlowe: So this ended up being messy because it's an expression that has two sequences in it now.  I ended up modeling it after [WindowSpecDefinition](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/windowExpressions.scala#L44-L56) in Spark where they needed to have two separate sequences as arguments to an expression.  In order to override `withNewChildren` without shimming the class, I needed to extend our base shim tree node classes to allow a common override across Spark versions.  cc: @gerashegalov who will probably be interested in checking what I did there.

jlowe:
build

shims/spark320/src/main/scala/com/nvidia/spark/rapids/shims/spark320/Spark320Shims.scala
jlowe: Filed #3733 as the followup.

jlowe:
build

jlowe:
Ran into a couple of issues with passing the CPU expressions as part of the regular arguments to the case classes:
- These arguments would appear in the plan, being mostly redundant with the GPU form of the expressions
- The `assertOnGpu` code would "find" these expressions and assert that they're not on the GPU

By passing them as a separate parameter list to the case classes, they no longer appear in the `expressions` list for these classes.  That avoids the CPU expressions appearing in the plan explain output and being seen by generic tree traversal code.

jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
jlowe: Now that this is passing the CPU expressions as an separate parameter list, we don't need to override `withNewChildren` and thus don't need the changes to the shims and can return just the GPU expressions as children as it did before.

tgravescs:
db 8.2 failures:
`   pyspark.sql.utils.IllegalArgumentException: The expression knownfloatingpointnormalized(normalizenanandzero(a#133227)) AS a#133227 is not columnar class org.apache.spark.sql.catalyst.expressions.Alias

`
```
18:34:16  FAILED ../../src/main/python/join_test.py::test_right_broadcast_nested_loop_join_condition_missing[Left-Float][IGNORE_ORDER({'local': True}), INCOMPAT]
18:34:16  FAILED ../../src/main/python/join_test.py::test_right_broadcast_nested_loop_join_condition_missing[Left-Double][IGNORE_ORDER({'local': True}), INCOMPAT]
18:34:16  FAILED ../../src/main/python/join_test.py::test_right_broadcast_nested_loop_join_condition_missing[LeftSemi-Float][IGNORE_ORDER({'local': True}), INCOMPAT]
18:34:16  FAILED ../../src/main/python/join_test.py::test_right_broadcast_nested_loop_join_condition_missing[LeftSemi-Double][IGNORE_ORDER({'local': True}), INCOMPAT]
18:34:16  FAILED ../../src/main/python/join_test.py::test_right_broadcast_nested_loop_join_condition_missing[LeftAnti-Float][IGNORE_ORDER({'local': True}), INCOMPAT]
18:34:16  FAILED ../../src/main/python/join_test.py::test_right_broadcast_nested_loop_join_condition_missing[LeftAnti-Double][IGNORE_ORDER({'local': True}), INCOMPAT]
18:34:16  FAILED ../../src/main/python/join_test.py::test_left_broadcast_nested_loop_join_condition_missing[Right-Float][IGNORE_ORDER({'local': True}), INCOMPAT]
18:34:16  FAILED ../../src/main/python/join_test.py::test_left_broadcast_nested_loop_join_condition_missing[Right-Double][IGNORE_ORDER({'local': True}), INCOMPAT]

```

jlowe:
build

revans2 (approved):

jlowe:
build

revans2 (approved):

PR #3726, opened 2021-09-30
Replace Class.forName with ShimLoader.loadClass

### LINKED ISSUE #3710: [BUG] Usage of Class.forName without specifying a classloader
**Describe the bug**
There are instances of `Class.forName` usage throughout the plugin, some of which don't specify the classloader.
https://github.com/NVIDIA/spark-rapids/blob/d2051d14ad17b568fd289059a2a64e1e73b1c2c2/sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala#L43

grep source code

**Expected behavior**
We should route all these invocations through the ShimLoader. Even if we decide not to do this, we should consistently use the Thread context classloader. 

local

**Additional context**
#3658

gerashegalov:
Most calls use Bootstrap classloader and may create copies for classes loaded via Shim and context classloaders. Fixes #3710 

    
### DIFF
--- a/api_validation/src/main/scala/com/nvidia/spark/rapids/api/ApiValidation.scala
+++ b/api_validation/src/main/scala/com/nvidia/spark/rapids/api/ApiValidation.scala
@@ -101,7 +101,7 @@ object ApiValidation extends Logging {
         }
 
         // TODO: Add error handling if Type is not present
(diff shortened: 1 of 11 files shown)


gerashegalov:
build

andygrove (approved):

tgravescs (approved):

Issue #3720, opened 2021-09-30
[BUG] Qualification tool warnings


tgravescs:
**Describe the bug**
```
[WARNING] [Warn] /home/tgraves/workspace/spark-rapids/tools/src/main/scala/org/apache/spark/sql/rapids/tool/FilterAppInfo.scala:62: comparing values of types Unit and Int using `==' will always yield false
[WARNING] [Warn] /home/tgraves/workspace/spark-rapids/tools/src/main/scala/org/apache/spark/sql/rapids/tool/FilterAppInfo.scala:65: comparing values of types Unit and Int using `==' will always yield false

```
these look like the code is functionally wrong.. I think worst case here is it processes more of event log then it needs to be we should fix.

PR #3728, opened 2021-10-01
Qualification tool: Fix bug to process correct listeners

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3720 . 
Earlier it was sending false and hence processing more listeners than required.
I ran mvn install with this change and the warnings are gone. Also now the loop exits once both the listeners are processed.


tgravescs:
lets put this in 21.10, please retarget PR, seems small enough fix

tgravescs (approved):

PR #3729, opened 2021-10-01
Update to ucx 1.11.2 for 21.10

### LINKED ISSUE #3651: [DOC] Point users to UCX 1.11.2
UCX 1.11.2 is fixing two bugs we have seen:

- When running on a machine that has a non-RDMA NIC (i.e. an ethernet-only device used as a the primary NIC for the executor), and a secondary RDMA NIC, we see failures since UCX would skip a step to register memory against the non-RDMA NIC and fail when attempting to use it later after realizing it exists: https://github.com/openucx/ucx/pull/7436.

- Seen in the DGX-2, an error where when we have so many GPUs (16) and nics (8) and the various transports that UCX supports, that an internal datastructure in UCX could not hold all of this: https://github.com/openucx/ucx/pull/7398

UCX 1.11.2 is in RC state, with the first RC out: https://github.com/openucx/ucx/releases/tag/v1.11.2-rc1. This issue is to change the docs to point to UCX 1.11.2, at least to the RC.

abellina:

UCX 1.11.2 was just released and it addresses the issues outlined in #3651. I have been running with it and not finding issues. This updates the docs and fetches it for pre_merge. I am updating 1 more job manually tomorrow, so I'll leave this as draft until I run that.

Closes #3651


abellina:
I am not seeing ucx specific issues in the builds/benchmarks we have executed with 1.11.2

jenkins/Dockerfile-blossom.ubuntu
```diff
@@ -25,7 +25,7 @@
 
 ARG CUDA_VER=11.0
 ARG UBUNTU_VER=18.04
-ARG UCX_VER=v1.11.0
+ARG UCX_VER=v1.11.2
```
jlowe, line 28: Should this be updated to not require the leading `v` to be consistent with the other Docker examples?

jlowe (dismissed):
Looks OK to me with a small suggestion for consistency.

jenkins/Dockerfile-blossom.ubuntu
abellina: Sure, I can do that.

abellina:
build

jlowe (approved):

tgravescs (approved):

PR #3746, opened 2021-10-04
Avoid using AST on inner joins and avoid coalesce after post-join filter

### LINKED ISSUE #3736: [BUG] Crossjoin performance degraded a lot on Spark 3.2rc3 + RAPIDS 21.10 snapshot
**Describe the bug**
Crossjoin performance degraded a lot on Spark 3.2rc3 + RAPIDS 21.10 snapshot.
@nvliyuan found that on Spark2a cluster, the performance difference is as below:

Spark 3.1: 13s (event log: app-20210908040027-3651)
Spark 3.2rc3: 324s (event log: app-20211001082934-0655)

As per the query plan, there are some non-GPU plan in the Spark 3.2rc3 event log.

Run microbenchmark cross-join query which is a 1 million row self-join:
```
spark.read.parquet("/data/tmp/customer1m").repartition(200).createOrReplaceTempView("costomer_df_1_million")
query = '''
select count(*) from costomer_df_1_million c1 inner join costomer_df_1_million c2 on c1.c_customer_sk>c2.c_customer_sk
'''
```

**Expected behavior**
It should have similar performance as before.

 - Environment location: [Standalone, YARN, Kubernetes, Cloud(specify cloud provider)]
 - Spark configuration settings related to the issue

rapids-4-spark_2.12-21.10.0-20210929.183153-116.jar
cudf-21.10.0-20210930.121728-64.jar
Spark2a 8 nodes standalone cluster.

**Additional context**

jlowe:
Fixes #3736.

@abellina and I looked into the nested loop join performance degradation and there were primarily two issues:
1. Computing the AST during the join was particularly expensive, resulting in approximately 4X performance improvement for the microbenchmark nested loop join if we revert back to generating a full cross join output and then post-filtering the results.
2. The `GpuFilterExec` extracted from the join in #3105 was causing a subsequent batch coalesce to occur that wasn't there before.  Traces showed `contiguous_split` was quite expensive on the post-join output for this query.  Updating the post-join filter to not coalesce batches resulted in another approximate 4X speedup from the previous change.

This updates nested loop joins to avoid using the AST if a post-filter can be used and also updates the post-filter injection code for joins to specify no coalesce after goal on the filter node.

### DIFF
--- a/sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
+++ b/sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
@@ -83,7 +83,9 @@ class GpuBroadcastHashJoinMeta(
       left, right)
     // The GPU does not yet support conditional joins, so conditions are implemented
     // as a filter after the join when possible.
-    condition.map(c => GpuFilterExec(c.convertToGpu(), joinExec)).getOrElse(joinExec)
+    condition.map {
+      c => GpuFilterExec(c.convertToGpu(), joinExec, coalesceAfter = false)
+    }.getOrElse(joinExec)
   }
 }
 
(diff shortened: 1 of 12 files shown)


jlowe:
build

tgravescs (approved):

abellina (approved):

jlowe:
build

jlowe:
Marking this as a draft until NDS performance runs are complete to verify there are no regressions.

jlowe:
NDS saw a perf regression in q72, the coalesce after the post-join filter turned out to be crucial in that case.  Refined this to only avoid the coalesce on a nested loop join due to the microbenchmark results, filed #3749 and #3750 to track fixing this in the general case.

jlowe:
build

abellina (approved):

jlowe:
NDS and cross join microbenchmark results on latest patch look good, marking this ready to merge.

revans2 (approved):

PR #3756, opened 2021-10-06
Simplify shim classloader logic

gerashegalov:
- uprev spark320 breaking DiskManager changes
- use the Serializer instance to find mutable classloader
- make the the update logic oblivious to the executor/driver side
- add missing binary dedupes

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -1586,6 +1586,10 @@
                                                 dest="${project.build.directory}/parallel-world/${spark.version.classifier}"
                                         >
                                         </unzip>
+                                        <exec executable="${project.basedir}/scripts/binary-dedupe.sh"
+                                              dir="${project.build.directory}"
+                                              failonerror="true"
+                                        />
                                         <jar
                                                 destfile="${project.build.directory}/rapids-4-spark_${scala.binary.version}-${project.version}.jar"
                                                 basedir="${project.build.directory}/parallel-world"
@@ -1684,6 +1688,10 @@
                                                 dest="${project.build.directory}/parallel-world/${spark.version.classifier}"
                                         >
                                         </unzip>
+                                        <exec executable="${project.basedir}/scripts/binary-dedupe.sh"
+                                              dir="${project.build.directory}"
+                                              failonerror="true"
+                                        />
                                         <jar
                                                 destfile="${project.build.directory}/rapids-4-spark_${scala.binary.version}-${project.version}.jar"
                                                 basedir="${project.build.directory}/parallel-world"
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
@@ -18,16 +18,18 @@ package com.nvidia.spark.rapids
 
 import java.net.URL
 
+import org.apache.commons.lang3.reflect.MethodUtils
+import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 
-import org.apache.spark.{SPARK_BUILD_USER, SPARK_VERSION, SparkConf, SparkEnv}
+import org.apache.spark.{SPARK_BRANCH, SPARK_BUILD_DATE, SPARK_BUILD_USER, SPARK_REPO_URL, SPARK_REVISION, SPARK_VERSION, SparkConf, SparkEnv}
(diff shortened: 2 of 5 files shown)


dist/pom.xml
```diff
@@ -1684,6 +1688,10 @@
                                                 dest="${project.build.directory}/parallel-world/${spark.version.classifier}"
                                         >
                                         </unzip>
+                                        <exec executable="${project.basedir}/scripts/binary-dedupe.sh"
```
tgravescs, line 1691: why are we deduping on individual, it only contains a single spark version anyway?

tgravescs (commented):
can you please update the description with what the issues were that this is fixing.  I'm guessing the build env doesn't have the newer spark jars yet to pick up that change.

tgravescs:
note the 3.2 deploy is running now so should be under an hour

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
```diff
@@ -119,41 +125,99 @@ object ShimLoader extends Logging {
...
+        // fast path
+        logInfo(s"findURLClassLoader found a URLClassLoader $urlCl")
+        Option(urlCl)
+      case replCl if replCl.getClass.getName == "org.apache.spark.repl.ExecutorClassLoader" =>
+        // https://issues.apache.org/jira/browse/SPARK-18646
+        val parentLoader = MethodUtils.invokeMethod(replCl, true, "parentLoader")
+          .asInstanceOf[ClassLoader]
+        logInfo(s"findURLClassLoader found $replCl, trying parentLoader=$parentLoader")
+        findURLClassLoader(parentLoader)
+      case urlAddable: ClassLoader if null != MethodUtils.getMatchingMethod(
```
abellina, line 184: should: `case urlCl: java.net.URLClassLoader` not match the class loader you are matching in the case at this line (184)?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
```diff
@@ -119,41 +125,99 @@ object ShimLoader extends Logging {
...
-          val addURL = parentLoader.getClass.getDeclaredMethod("addURL", classOf[java.net.URL])
-          addURL.invoke(parentLoader, shimURL)
-          addURL.invoke(parentLoader, shimCommonURL)
-          pluginClassLoader = contextClassLoader
-        }
+    serializerClassloader().foreach { urlAddable =>
+      logInfo(s"Updating spark classloader $urlAddable with the URLs: " +
+        urlsForSparkClassLoader.mkString(", "))
+      urlsForSparkClassLoader.foreach { url =>
+        MethodUtils.invokeMethod(urlAddable, true,"addURL", url)
```
abellina, line 207: minor nit,
```suggestion
        MethodUtils.invokeMethod(urlAddable, true, "addURL", url)
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
abellina: And if both are instances of `URLClassLoader` should the method return `Option[URLClassLoader]`?

abellina (commented):
LGTM, just had a minor nit that should be skipped if no other changes are made to this PR. The description of the PR is not clear on what this is solving (are we just simplifying or are we fixing an issue.. other than the `DiskBlockManager` breaking change).

tgravescs:
build

dist/pom.xml
gerashegalov: top level classes are duplicated
```find  dist/target/parallel-world/ -name \*.class | xargs sha1sum | sort -k1,1 | awk '{print $2 " " $1}' | uniq -f 1 -d
dist/target/parallel-world/org/openucx/jucx/ucp/UcpListener.class 055d141a3d00e9503946e93f24e09ef46dc5bf98
dist/target/parallel-world/org/openucx/jucx/ucp/UcpEndpointErrorHandler.class 065e86ee52ca0a287d0e23d03727f4d1e6219930
dist/target/parallel-world/org/openucx/jucx/ucp/UcpListenerConnectionHandler.class 0a512cc7021c09173797dceb501fe38acfe95ddd
...
```

tgravescs (approved):
Description still needs updating.

PR #3762, opened 2021-10-06
Call the right method to convert table from row major <=> col major [databricks]

### LINKED ISSUE #3761: [FEA] Call the appropriate method when transitioning from Row<=>Col
When converting table from row major to col major and back we should call the appropriate kernel. 

Currently, we only call a kernel optimized for fixed width data only and its limited by the number of columns (<100). Otherwise it falls back to the unsafe projection. 

As part of rapidsai/cudf#8444 we will have a new kernel which should be called for cases when number of columns is > 100 and later for data types other than fixed-width types.

razajafri:
This is a draft PR that depends on rapidsai/cudf#8444

* Added logic to branch the code based on the number of columns selected while convert from Columnar to Row and back
* Added test to test the new method which will handle columns > 100 

closes #3761


razajafri:
build

razajafri:
The cudf nightly failed today which should be fixed by https://github.com/rapidsai/cudf/pull/10015. Will kick off the build tomorrow if the nightly is successful

sql-plugin/src/main/java/com/nvidia/spark/rapids/UnsafeRowToColumnarBatchIterator.java
```diff
@@ -163,7 +166,16 @@ public ColumnarBatch next() {
     }
     try (NvtxRange ignored = buildRange;
          ColumnVector cv = devColumn;
-         Table tab = Table.convertFromRows(cv, rapidsTypes)) {
+         // We are branching over the output.length to know which kernel to call.
+         // If output.length < 100 we call the fixed-width optimized version, otherwise the
```
revans2, line 170: It is really odd to have all of these comments explaining the limitations of the fixed width optimized here, when the check for it is in a entirely different place. Could we do one of the following

1. Get rid of `isOptimizedForFixedWidth` and instead use `rapidsTypes.length` to decide which to call. If this is a problem for testing, then we can have an test only constructor that exposes an override for the check.
2. move the comments to GpuRowToColumnarExec.scala where isOptimizedForFixedWidth is calculated.

I personally prefer option 1, but it is more code.

razajafri: I have made the requested changes to `UnsafeRowToColumnarBatchIterator`.  And a similar change was made to the `GpuColumnarToRowExec`. PTAL

revans2 (approved):

razajafri:
build

PR #3763, opened 2021-10-07
Force parallel world in Shim caller's classloader

### LINKED ISSUE #3704: [BUG] ClassCastException when testing with Spark 3.2.1-SNAPSHOT in k8s environment
**Describe the bug**

Benchmark queries are failing in SJC4 k8s cluster:

```
java.lang.ClassCastException: cannot assign instance of scala.collection.immutable.List$SerializationProxy to field org.apache.spark.rdd.RDD.dependencies_ of type scala.collection.Seq in instance of org.apache.spark.rdd.MapPartitionsRDD
        at java.io.ObjectStreamClass$FieldReflector.setObjFieldValues(ObjectStreamClass.java:2301)
        at java.io.ObjectStreamClass.setObjFieldValues(ObjectStreamClass.java:1431)
        at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2411)
        at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2329)
        at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2187)
        at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1667)
        at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2405)
        at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2329)
        at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2187)
        at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1667)
        at java.io.ObjectInputStream.readObject(ObjectInputStream.java:503)
        at java.io.ObjectInputStream.readObject(ObjectInputStream.java:461)
        at scala.collection.immutable.List$SerializationProxy.readObject(List.scala:527)
```

See benchmark runs in sjc4 cluster.

**Expected behavior**

Should work.

**Additional context**
N/A

gerashegalov:
This PR codes up the approach that was tested on Databricks as an alternative while working on #3756 
 - making this option default because it's mostly equivalent to the old flat jar classloader 
 - making it optional because we still debug how we miss the non-default classloader in #3704, and it's not the right behavior for addJar with userClassPathFirst. However, I think we should generally stop documenting --jars as the plugin deploy option
    
Fixes #3704
    
### DIFF
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -135,8 +135,8 @@ else
 
     export PYSP_TEST_spark_driver_extraClassPath="${ALL_JARS// /:}"
     export PYSP_TEST_spark_executor_extraClassPath="${ALL_JARS// /:}"
-    export PYSP_TEST_spark_driver_extraJavaOptions="-ea -Duser.timezone=UTC $COVERAGE_SUBMIT_FLAGS -Dsun.misc.URLClassPath.debug=true"
-    export PYSP_TEST_spark_executor_extraJavaOptions='-ea -Duser.timezone=UTC -Dsun.misc.URLClassPath.debug=true'
+    export PYSP_TEST_spark_driver_extraJavaOptions="-ea -Duser.timezone=UTC $COVERAGE_SUBMIT_FLAGS"
+    export PYSP_TEST_spark_executor_extraJavaOptions='-ea -Duser.timezone=UTC'
     export PYSP_TEST_spark_ui_showConsoleProgress='false'
     export PYSP_TEST_spark_sql_session_timeZone='UTC'
     export PYSP_TEST_spark_sql_shuffle_partitions='12'


gerashegalov:
> Why would we do this? That is generally how you are supposed to distribute jars on spark, what is your alternative? this needs more explanation. for instance, on yarn you use --jars to get things into distributed cache to get sent to the nodes, without this you expect people to install it on every node?

Good point. My main point that we should decide which of multiple ways to deploy plugin jars we recommend and support, and make sure that it works with all the features we provide such as the shuffle manager in all Spark deploy modes.

gerashegalov:
While talking through the list of pros and cons with @tgravescs we concluded that my concern `userClassPathFirst` is moot with the following reasoning:
pre-21.10 the whole jar was visible to the ShimLoader's classloader, and we are not changing what classloader was chosen by Spark to load the shim.  Thus by making the right sections of the jar accessible to the same class loader makes it only equivalent to the pre-21.10 setup.  

So maybe having a reflection call into a non-public API is the only known downside at this moment.

PR #3779, opened 2021-10-08
Enable some union of structs tests that were marked xfail

razajafri:
This PR just removes the mark to xfail some tests


jlowe:
build

PR #3790, opened 2021-10-11
Fix overflow when casting integral to neg scale decimal

### LINKED ISSUE #3788: [BUG] DECIMAL32 overflow when casting Integer to Decimal
**Describe the bug**
When casting int values who are close to INT32 bounds into Decimal32 with negative scale, it will lead to the INT32 overflow.
The reason of overflow is that HALF_UP rounding enlarges the original value. For instance,  `2147483647`(max INT32) will be `2147483650` after rounding, which exceeds the up bound of INT32. 

```scala
import org.apache.spark.sql.SparkSession
import org.apache.spark.sql.types.DataTypes

SparkSession.getActiveSession.get.sqlContext.setConf("spark.sql.legacy.allowNegativeScaleOfDecimal", "true")
val df = List(Int.MinValue, Int.MaxValue).toDF("cInt")
df.select(df.col("cInt").cast(DataTypes.createDecimalType(9, -1))).collect()
```

revans2:
This fixes #3788

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
@@ -1108,13 +1108,15 @@ object GpuCast extends Arm {
     }
   }
 
-  private def getPrecisionForIntegralInput(input: ColumnView): Int = {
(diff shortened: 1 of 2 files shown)


revans2:
build

PR #3791, opened 2021-10-11
Add audit script to get list of commits from Apache Spark master branch

nartal1:
This PR is to add script to get the list of commits to be audited for Spark-3.3(master) branch. 
We already have similar script to get list of commits for Spark-3.2(branch-3.2). The scripts are separated so that it would be easier to setup Jenkins job to get the commits for different branches.

### DIFF
--- a/scripts/audit-spark.sh
+++ b/scripts/audit-spark.sh
@@ -32,14 +32,16 @@ ABSOLUTE_PATH=$(cd $(dirname $0) && pwd)
 lastcommit=""
 basebranch="master"
 tag="branch-3.2"
+commonancestor=""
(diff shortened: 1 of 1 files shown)


nartal1:
> Most of the file is a copy of the file scripts/audit-spark-3.2.sh. Can we modularize it?

I had kept it separate as it would be easier to setup jenkins job for separate Spark versions(as we have different commit ids) and also it would be easier to remove the scripts/jenkins job once we stop supporting a particular version. Please let me know if it's okay.

nartal1:
I will update the current script so that it reflects commits from different Spark version.

Issue #3654, opened 2021-09-24
Premerge test with combined jar


tgravescs:
**Is your feature request related to a problem? Please describe.**
Currently the premerge just uses the individual profile to run the tests, it would be nice to use the combined jar -Psnapshots to make sure no issues with multiple jars combined.  The problem is we run jacoco and it doesn't seem to like the combined jar.

```
10:12:54  java.lang.IllegalStateException: Can't add different class with same name: com/nvidia/spark/rapids/AvoidAdaptiveTransitionToRow
10:12:54  	at org.jacoco.core.analysis.CoverageBuilder.visitCoverage(CoverageBuilder.java:106)
10:12:54  	at org.jacoco.core.analysis.Analyzer$1.visitEnd(Analyzer.java:99)
10:12:54  	at org.objectweb.asm.ClassVisitor.visitEnd(ClassVisitor.java:377)
10:12:54  	at org.jacoco.core.internal.flow.ClassProbesAdapter.visitEnd(ClassProbesAdapter.java:100)
10:12:54  	at org.objectweb.asm.ClassReader.accept(ClassReader.java:725)
10:12:54  	at org.objectweb.asm.ClassReader.accept(ClassReader.java:401)
10:12:54  	at org.jacoco.core.analysis.Analyzer.analyzeClass(Analyzer.java:116)
10:12:54  	at org.jacoco.core.analysis.Analyzer.analyzeClass(Analyzer.java:132)
10:12:54  Caused: java.io.IOException: Error while analyzing /var/jenkins/jobs/rapids_premerge-g
```

Investigate solutions or perhaps run jacoco differently.

revans2:
@sameerz One option to fix this is to just remove all of the JaCoCo code coverage. No one really looks at the numbers and enough has changed with the build, 3.2 and splitting the tests into separate builds, the numbers no longer really show that much of a full picture of what is happening.  Should we spend more time to make this useful? Or do we just remove it all together?

sameerz:
Short term we can disable JaCoCo.  Mid term we should re-investigate how we measure code coverage.

PR #3801, opened 2021-10-12
Update premerge to use the combined snapshots jar 

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/3654

Here we need to disable jacoco because it doesn't understand the combined jar. See the issue for my details, we filed a separate issue to look at longer term solution for that.

This enables premerge to use the snapshots combined jar so we are testing with that jar during premerge.  I tested this in separate premerge build in order to test Jenkinsfile changes.


tgravescs:
build

PR #3808, opened 2021-10-13
Rewording qualification and profiling tools doc files[skip ci]

nvliyuan:
Rewording qualification and profiling tools doc files.


pxLi:
build

Issue #3535, opened 2021-09-17
[FEA] Qualification tool can detect RDD APIs in SQL plan


viadea:
**Is your feature request related to a problem? Please describe.**
I wish Qualification tool can detect RDD APIs in SQL plan.

**Describe the solution you'd like**
User may start with a RDD and does transformation using RDD APIs like `map`, `mapPartitions`, etc.
And in the end, if the user translate the result RDD to a dataframe, the Spark UI still report total time as SQL duration.

For example:
```
val rdd=spark.range(1000000).rdd.distinct().map(_*10).filter(e => e%2000000==0)
rdd.toDF.collect()
```

Inside the SQL plan, there could be RDD API related key words such as "MapPartitionsRDD".
Qualification tool can detect that , and report it problematic.

**Describe alternatives you've considered**

**Additional context**
Add any other context, code examples, or references to existing implementations about the feature request here.

PR #3819, opened 2021-10-14
Qualification tool: Detect RDD Api's in SQL plan

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3535.

Whenever RDD is converted to Dataset/DataFrame, there is `SerializeFromObject` in the SQL plan. In this PR, the logic is same as that for Dataset i.e if we see RDD to Dataset/DataFrame conversion, then we don't consider the time taken by that operation while calculating the score.


nartal1:
build

tgravescs:
build

Issue #3837, opened 2021-10-16
[BUG] Spark-rapids v21.10.0 release candidate jars failed on the OSS validation check.


NvTimLiu:
**Describe the bug**
Spark-rapids v21.10.0 release candidate jars failed on the OSS validation check as below.

![image](https://user-images.githubusercontent.com/50287591/137586727-3188642d-0484-43f0-80a0-82394db039ff.png)

Need to upload 'dist/pom.xml' as the spark-rapids project's pom file, because OSS takes the unspecified/auto-generated pom as invalid.
Need to deploy a jar without any classifier tag as the main jar of integration-test, because OSS requires the main jar without classifier.

**Steps/Code to reproduce bug**
Click `close` on the spark-rapids v21.10.0 release candidate jars onto OSS staging web page:  https://oss.sonatype.org/

PR #3838, opened 2021-10-16
Update deploy script [skip ci]

NvTimLiu:
Need to upload 'dist/pom.xml' as the spark-rapids project's pom file, because OSS takes the unspecified/auto-generated pom as invalid.

Need to deploy a jar without any classifier tag as the main jar of integration-test, because OSS requires the main jar without classifier.


NvTimLiu:
To fix issue https://github.com/NVIDIA/spark-rapids/issues/3837

PR #3848, opened 2021-10-18
Fix binary-dedupe failures and improve its performance on macOS

### LINKED ISSUE #3823: [BUG]binary-dedup.sh script fails on mac
**Describe the bug**
looks like xargs on mac doesn't support --arg-file option:
```

     [exec] + xargs --arg-file=/Users/tgraves/Documents/workspace/spark-rapids/dist/target/spark3xx-common.txt -P 6 -n 1 -I% bash -c 'remove_duplicates "$@"' _ %
     [exec] xargs: illegal option -- -
     [exec] usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements]] [-J replstr]
     [exec]              [-L number] [-n number [-x]] [-P maxprocs] [-s size]
     [exec]              [utility [argument ...]]
```

line`: xargs --arg-file="$SPARK3XX_COMMON_TXT" -P 6 -n 1 -I% bash -c 'remove_duplicates "$@"' _ %`

instead of using the --arg-file option I think we can just cat the the file and | it into xargs

Note there is a line at the end that uses it too:

`xargs --arg-file="$UNSHIMMED_LIST_TXT" -P 6 -n 100 -I% \
`
But that didn't fail so we probably have a bug where it should fail.

gerashegalov:
This PR fixes #3823. 

The minimum fix for this issue is to use `shasum` on macOS vs `sha1sum` on Linux. Once the script is fixed functionally it exhibits unacceptable performance on macOS. The single-threaded performance was also bad on Linux but it was mitigated by using parallel xargs. The slowness may  be attributable to much higher cost of forking a child process on macOS. This PR reworks the script such that it is much faster without requiring parallelism, and the run time is similar for both Linux and macOS.

1. eliminate instances of forked process per class file by using bash builtins 
2. compute sha for all files just once
3. build large lists for a few large `rsync` and `rm` calls

On my mbp the run time goes down from at least a few minutes (did not wait for completion ) to 25 seconds
On my Linux desktop the script runtime decreases from over a minute to 3-5 seconds depending on the profile.
 
### DIFF
--- a/build/buildall
+++ b/build/buildall
@@ -27,8 +27,8 @@ case "$1" in
   echo "   --profile=DIST_PROFILE   profile for the dist module, default: noSnapshots,"
   echo "                            also supported: snapshots, minimumFeatureVersionMix"
   echo "   --profile=BUILDVER       builds a single Spark shim, e.g. --profile=320"
-  echo "   --module=MODULE          after finishing parallel builds, resume from dist and build up"
-  echo "                            up to MODULE, e.g., integration_tests"
+  echo "   --module=MODULE          after finishing parallel builds, resume from dist and build"
+  echo "                            up to and including module MODULE, e.g., integration_tests"
   exit 0
   ;;
 


gerashegalov:
build

gerashegalov:
build

Issue #3659, opened 2021-09-24
[FEA] External tool integration with Qualification tool


tgravescs:
**Is your feature request related to a problem? Please describe.**
External tools/api's may want to call functions similar to qualification tool, investigate options there.

- [x] Plugin explain output on CPU dataframe
- [x] Qualification tool API called from running application

tgravescs:
prs merged

PR #3850, opened 2021-10-18
Add explain Plugin API for CPU plan

tgravescs:
part of https://github.com/NVIDIA/spark-rapids/issues/3659

I specifically did not put in user docs anywhere, we need to decide what is best place or if we want to leave this undocumented until we have tried it on a few customers.

This adds an API that will run the CPU plan through the plugin api's to see what it can convert and what it can't and just outputs a string with the explain output.    Since we are looking at the CPU plan it runs other things as compared to when the rapids spark plugin would normally run in the query execution, so we strip those back out. ie cpu plan that we look at here will have CollapseCodegenStages and subqueries and ReuseExchangeAndSubquery applied, whereas normally our plugin runs before those get applied to the plan.  See more details in `prepareExplainOnly`

It currently still requires the rapids 4 spark jar and cudf jars to be present in the Spark application , but doesn't need gpu or cuda.

The main api exposed:  _com.nvidia.spark.rapids.ExplainPlan.explainPotentialGpuPlan(df)_

The changes in GPUOverrides are mostly moving code from the class into the object and then adding the functions needed to process the CPU plan without doing the convert.

Perhaps in the future we can try to make it more standalone tool but we need to get the GPUoverrides meta stuff where we decide if we can convert or not out by itself.   

We may also want to add more API's there to perhaps get some sort of summary report rather then dumping out all the explain text.

I ran this api through the NDS queries and compared to actual GPU explain and it did pretty good when comparing the 2. This is how I found I missed the subquery expressions.  So now we follow those and explain on those for cpu side.

I manually tested this on Databricks, only difference is you have to add libraries through UI so that it is available in python notebook.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ExplainPlan.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ExplainPlan.scala
@@ -15,7 +15,15 @@
  */
 package com.nvidia.spark.rapids
 
+import scala.util.control.NonFatal
+
 import org.apache.spark.sql.DataFrame
+
+// Base trait visible publicly outside of parallel world packaging.
+// It can't be named the same as ExplainPlan object to allow calling from PySpark.
+trait ExplainPlanBase {
+  def explainPotentialGpuPlan(df: DataFrame, explain: String = "ALL"): String
+}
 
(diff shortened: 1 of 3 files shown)


tgravescs:
build

PR #3853, opened 2021-10-19
Fix typo in nightly deploy project list

### LINKED ISSUE #3852: spark-nightly-build deploys all modules due to typo in `-pl`
`-pl -dist` instead of `-pl dist` in https://github.com/NVIDIA/spark-rapids/blob/branch-21.10/jenkins/spark-nightly-build.sh#L28 

***Context***
https://github.com/NVIDIA/spark-rapids/pull/3523#issuecomment-943759901

gerashegalov:
Use the correct module name in deploy `-pl`. Fixes #3852 


tgravescs:
>> Still deploying everything but dist is probably not what we want. I think what we mean here:
>> install individual buidlvers up to the aggregator module install -pl aggregator -am
>> then just deploy -pl dist -PsnapshotsWithDatabricks

No, we want it to do everything except for the dist. The reason being is we want it to deploy the integration tests jar and files that are used later to do integration testing on the "released" jars for various CSPs, spark versions, deployments, etc.  So we want the final dist combined jar but then we need the spark specific versions of the integration tests.

PR #3862, opened 2021-10-19
Prevent approx_percentile aggregate from being split between CPU and GPU

### LINKED ISSUE #3834: [BUG] Approx_percentile deserialize error when calling "show" rather than "collect"
**Describe the bug**
When running a `show` against an approx_percentile aggregate query, the query fails with a serialization error due to the partial aggregate running on GPU and the final aggregate running on CPU. The intermediate aggregate buffer is not in the expected format for CPU so deserialization fails.

The final stage falls back to the CPU due to a cast to string that is trigger by `show` and is not supported on GPU.

```
!Expression <Cast> cast(approx_percentile(v#476, [0.1,0.2], 10000, 0, 0)#479 as string) cannot run on GPU because Cast from ArrayType(DoubleType,false) to StringType is not supported
```

Example of the error stack trace:

```
Caused by: java.lang.NegativeArraySizeException
at org.apache.spark.sql.catalyst.expressions.aggregate.ApproximatePercentile$PercentileDigestSerializer.deserialize(ApproximatePercentile.scala:319)
```

I can reproduce with this integration test (added to `hash_aggregate_test.py`).

```python
@pytest.mark.parametrize('aqe_enabled', ['false', 'true'], ids=idfn)
@ignore_order(local=True)
@allow_non_gpu('TakeOrderedAndProjectExec', 'Alias', 'Cast', 'ObjectHashAggregateExec', 'AggregateExpression',
    'ApproximatePercentile', 'Literal', 'ShuffleExchangeExec', 'HashPartitioning', 'CollectLimitExec')
def test_hash_groupby_approx_percentile_double_show(aqe_enabled):
    conf = copy_and_update(_approx_percentile_conf, {
        'spark.sql.adaptive.enabled': aqe_enabled,
        'spark.rapids.sql.explain': 'ALL'
    })

    def foo(spark):
        df = gen_df(spark, [('k', StringGen(nullable=False)),
                                         ('v', DoubleGen())], length=100)
        df.createOrReplaceTempView("t")
        df2 = spark.sql("SELECT k, approx_percentile(v, array(0.1, 0.2)) from t group by k")

        # this "show" triggers the exception
        df2.show()

        return df2

    run_with_cpu_and_gpu(foo, 'COLLECT', conf)
```

**Expected behavior**
We should recognize that the final aggregate cannot run on GPU and force the partial aggregate to fall back to CPU as well.

N/A

**Additional context**
N/A

andygrove:

Closes #3834 

We have a mechanism for tagging the underlying SparkPlan when operators are not supported on GPU because we rely on this information when planning query stages when AQE is enabled. The reason for this is that we are only planning a subset of the plan without context about the parent of the query stage being planned. However, we were only respecting these tags for exchange operators and were ignoring them for aggregates. The main change in this PR is to check these tags on all operators.

This change caused some regressions that also had to be addressed in this PR, such as:

- The `CoalesceShufflePartitions` optimization performs a `transformUp` and may replace `ShuffleQueryStageExec` with `CustomShuffleReaderExec`, causing Spark to copy tags from `ShuffleQueryStageExec` to `CustomShuffleReaderExec`, including the "no need to replace ShuffleQueryStageExec" tag, so this tag needed to be ignored.
- `ObjectHashAggregateExec` and `SortAggregateExec` both had type signatures that declared that `BinaryType` was not supported, which is not always the case. We do support `BinaryType` for aggregate buffers and there is special handling in the aggregate code for both the case when we are able to convert between CPU and GPU for these buffers and also for the case we are not. This PR adds `BinaryType` to the type signatures to prevent them from being tagged early on as unsupported on GPU, which was causing regressions in the AQE case.
- `SortExec` was incorrectly declaring that `BinaryType` is not supported so the type checks are updated and new tests added to demonstrate that we fall back to CPU if a sort expression is binary, but we allow non-sort columns to be binary.

There are some other smaller changes in the PR:

- `gpuSupportedTag` was moved to a new `object RapidsMeta` so that it could be referenced from classes outside of the `RapidsMeta` hierarchy.
- When checking existing tags, we only call `RapidsMeta.willNotWorkOnGpu` for reasons that have not already been added to the `cannotBeReplacedReasons` set.


revans2 (approved):

andygrove:
build

andygrove:
build failed with:

```
Error: 1-01T18:35:45.150Z] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:3.0.0:run (create-parallel-world) on project rapids-4-spark_2.12: An Ant BuildException has occured: The following error occurred while executing this line:
Error: 1-01T18:35:45.150Z] [ERROR] /home/jenkins/agent/workspace/jenkins-rapids_premerge-github-3157/dist/maven-antrun/build-parallel-worlds.xml:98: exec returned: 255
Error: 1-01T18:35:45.150Z] [ERROR] around Ant part ...<ant antfile="/home/jenkins/agent/workspace/jenkins-rapids_premerge-github-3157/dist/maven-antrun/build-parallel-worlds.xml" target="build-parallel-worlds" />... @ 9:163 in /home/jenkins/agent/workspace/jenkins-rapids_premerge-github-3157/dist/target/antrun/build-main.xml
```

andygrove:
build

PR #3869, opened 2021-10-20
Add Std dev samp for windowing [databricks]

### LINKED ISSUE #3814: [FEA] stddev stddev_samp and std should be supported over a window
It will be good to support std, stddev and stddev_samp over a window

razajafri:
This PR adds windowing support for stddev. Its built on top of @revans2 changes for adding support for `GpuReplaceWindowFunction` 

* Added changes to AggregateFunction.scala for the actual Expression changes. Had to add a `CudfWindowStddev` class to avoid stack overflow when replacing `StddevSamp` 
* Added tests in `window_function_test.py`. 

depends on rapidsai/cudf#9527

closes #3814 


abellina:
> There are known issues in this implementation. 

Should we have it disabled by default for now? We should link the cuDF issue if there is one.

> The NaNs and Nulls are not handled properly. The correct way to do this is to run two aggregations and then filter out the result.

Is this something you are planning to handle in the plugin? Or is that going to be handled under the hood by cuDF

razajafri:
build

razajafri:
@revans2 can take another look?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -1204,10 +1204,29 @@ case class GpuStddevPop(child: Expression, nullOnDivideByZero: Boolean)
...
+    RollingAggregation.standardDeviation().onColumn(inputs.head._2)
+  }
+}
+
 case class GpuStddevSamp(child: Expression, nullOnDivideByZero: Boolean)
-  extends GpuM2(child, nullOnDivideByZero) {
+    extends GpuM2(child, nullOnDivideByZero) with GpuReplaceWindowFunction {
 
-  override lazy val evaluateExpression: Expression = {
+  override val evaluateExpression: Expression = {
```
revans2, line 1229: nit: Can we add the lazy back in?

revans2 (dismissed):
Just one minor nit

razajafri:
build

revans2 (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3090,15 +3090,22 @@ object GpuOverrides extends Logging {
...
-          GpuStddevSamp(childExprs.head, !legacyStatisticalAggregate)
-        }
-      }),
+      ExprChecksImpl(
+        ExprChecks.groupByOnly(
+          TypeSig.DOUBLE, TypeSig.DOUBLE,
+          Seq(ParamCheck("input", TypeSig.DOUBLE,
+            TypeSig.DOUBLE))).asInstanceOf[ExprChecksImpl].contexts
+            ++
+            ExprChecks.windowOnly(
```
ttnghia, line 3099: Should this be just `aggNotWindow`?

razajafri: I didn't know about that ExprCheck. I think that's what we need, @revans2 do you see anything wrong with using `aggNotReduction` here?

revans2: aggNotReduction should do the same thing, just be cleaner.

razajafri:
build

razajafri:
@ttnghia @revans2 can you please take another look

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -1204,8 +1204,27 @@ case class GpuStddevPop(child: Expression, nullOnDivideByZero: Boolean)
   override def prettyName: String = "stddev_pop"
 }
 
+case class WindowStddevSamp(
```
ttnghia, line 1207: Are we going to support `WindowStddevPop` too?

razajafri: Not in this PR

ttnghia: Why not? The `std` family can be supported all at once with very little code added. Otherwise, we may have new requests from customers to go back, read the code, add more code + tests etc... in a significantly much more time.

IMO adding all `stddev_pop/samp` and `var_pop/samp` at once is the most optimal way.

razajafri: It might be very easy to do but that doesn't mean that we have to bundle them together. This PR is specific to a client's request and is scheduled for this release.

revans2 (approved):

abellina (approved):

PR #3821, opened 2021-10-14
Support ORC write Map column

res-life:
This fixes #3784

Signed-off-by: Chong Gao <res_life@163.com>

PR #3874, opened 2021-10-21
[test only, do not review] Orc write map

res-life:


res-life:
Original was merged https://github.com/NVIDIA/spark-rapids/pull/3821

Issue #3866, opened 2021-10-20
[BUG] RLike integration tests failing on Azure Databricks 7.3


andygrove:
**Describe the bug**

## string_test.py::test_rlike

```
07:14:27  21/10/20 12:22:16 ERROR GpuOverrideUtil: Encountered an exception applying GPU overrides java.lang.IllegalArgumentException: Part of the plan is not columnar class org.apache.spark.sql.execution.ProjectExec
07:14:27  Project [a#1930890 RLIKE a{2} AS a RLIKE a{2}#1930892, a#1930890 RLIKE a{1,3} AS a RLIKE a{1,3}#1930893, a#1930890 RLIKE a{1,} AS a RLIKE a{1,}#1930894, a#1930890 RLIKE a[bc]d AS a RLIKE a[bc]d#1930895]
07:14:27  +- Scan ExistingRDD[a#1930890]
```

## string_test.py::test_rlike_escape

```
07:14:27  21/10/20 12:22:16 ERROR GpuOverrideUtil: Encountered an exception applying GPU overrides java.lang.IllegalArgumentException: Part of the plan is not columnar class org.apache.spark.sql.execution.ProjectExec
07:14:27  Project [a#1930904 RLIKE a[\-] AS a RLIKE a[\-]#1930906]
07:14:27  +- Scan ExistingRDD[a#1930904]
```

**Steps/Code to reproduce bug**
Run CI

**Expected behavior**
Tests should pass

**Environment details (please complete the following information)**
TBD

**Additional context**
Internal link to CI logs: https://blossom.nvidia.com/sw-gpu-spark-jenkins/job/rapids_it-AzureDatabricks-7.3/170/consoleFull

PR #3947, opened 2021-10-28
Enable rlike tests on databricks [databricks]

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/3866

I was unable to reproduce the issue when testing on Databricks 7.3 on Azure so am re-enabling these tests.


sameerz:
Need to keep an eye on these tests to see if they fail again.

PR #3960, opened 2021-10-28
Use a fixed RMM pool size

rongou:
As discussed in https://github.com/rapidsai/cudf/pull/9505#discussion_r736551016, setting RMM pool with a fixed size, only using the min and max sizes as boundary checks.


jlowe (dismissed):
> in some cases we may want to limit the max size for other things (e.g. python).

That's fine.  If we discover later that the max fraction is confusing users since the pool no longer grows or it doesn't seem to be useful in practice we can remove it then.

rongou:
build

rongou:
build

rongou:
build

tests/src/test/scala/com/nvidia/spark/rapids/GpuDeviceManagerSuite.scala
```diff
@@ -48,12 +47,14 @@ class GpuDeviceManagerSuite extends FunSuite with Arm with BeforeAndAfter {
         .set(RapidsConf.RMM_ALLOC_MAX_FRACTION.key, maxPoolFraction.toString)
         .set(RapidsConf.RMM_ALLOC_RESERVE.key, "0")
     TestUtils.withGpuSparkSession(conf) { _ =>
+      val freeGpuSize = Cuda.memGetInfo().free
       val poolSize = (freeGpuSize * poolFraction).toLong
+      val allocSize = Math.max(poolSize - 1024 * 1024, 0)
```
jlowe, line 52: It's nice that this protects against allocating a negative number, but do we really expect the test to fail if we try to allocate zero bytes? Maybe this should try to allocate a percentage (e.g.: ~75%) of the calculated pool size rather than a hardcoded offset, asserting the allocSize > 0.

rongou: Done.

jlowe (approved):

rongou:
build

PR #3982, opened 2021-11-01
Fix the issue of parquet reading with case insensitive schema

wbo4958:
## How to reproduce

1.  generate a parquet file

``` scala
import org.apache.spark.sql.Row

val data = Seq(
    Row("1"),
    Row("2"),
    Row("3")
)

val schema = (new StructType().add("ID",StringType))

val df = spark.createDataFrame(spark.sparkContext.parallelize(data),schema)
```

2. create hive table

``` mysql
CREATE EXTERNAL TABLE `testorder`(
  `ID` string
)
ROW FORMAT SERDE                                                                                                                                          
  'org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe'                                                                                           
STORED AS INPUTFORMAT                                                                                                                                     
  'org.apache.hadoop.hive.ql.io.parquet.MapredParquetInputFormat'                                                                                         
OUTPUTFORMAT                                                                                                                                              
  'org.apache.hadoop.hive.ql.io.parquet.MapredParquetOutputFormat'                                                                                        
LOCATION                                                                                                                                                  
  '/tmp/testparquet';
```

3. reading the hive table by GPU

``` scala
spark.sql("REFRESH TABLE testorder")
spark.sql("select * from testorder limit 1 ").show
```

## Exception

``` console
Exception in thread "main" org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 0.0 failed 1 times, most recent failure: Lost task 0.0 in stage 0.0 (TID 0) (spark-bobby executor driver): java.lang.ArrayIndexOutOfBoundsException: 0
	at ai.rapids.cudf.Table.<init>(Table.java:94)
	at ai.rapids.cudf.Table.readParquet(Table.java:862)
	at com.nvidia.spark.rapids.ParquetPartitionReader.$anonfun$readToTable$1(GpuParquetScanBase.scala:1485)
	at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
	at com.nvidia.spark.rapids.Arm.withResource$(Arm.scala:26)
	at com.nvidia.spark.rapids.FilePartitionReaderBase.withResource(GpuMultiFileReader.scala:236)
	at com.nvidia.spark.rapids.ParquetPartitionReader.readToTable(GpuParquetScanBase.scala:1484)
	at com.nvidia.spark.rapids.ParquetPartitionReader.$anonfun$readBatch$1(GpuParquetScanBase.scala:1445)
	at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
	at com.nvidia.spark.rapids.Arm.withResource$(Arm.scala:26)
	at com.nvidia.spark.rapids.FilePartitionReaderBase.withResource(GpuMultiFileReader.scala:236)
	at com.nvidia.spark.rapids.ParquetPartitionReader.readBatch(GpuParquetScanBase.scala:1433)
	at com.nvidia.spark.rapids.ParquetPartitionReader.next(GpuParquetScanBase.scala:1418)
	at com.nvidia.spark.rapids.PartitionReaderWithBytesRead.next(GpuDataSourceRDD.scala:94)
	at com.nvidia.spark.rapids.ColumnarPartitionReaderWithPartitionValues.next(ColumnarPartitionReaderWithPartitionValues.scala:36)
	at com.nvidia.spark.rapids.PartitionReaderIterator.hasNext(PartitionReaderIterator.scala:37)
	at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.hasNext(FileScanRDD.scala:93)
```

## Root cause

``` console
== Physical Plan ==
GpuColumnarToRow false
+- GpuFileGpuScan parquet default.testorder[id#0] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex[file:/home/bobwang/tmp/hive/testparquet], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<id:string>
```

The spark reading data schema is "id", while the schema of parquet file is "ID". 

The original code specifies the spark reading data column name, but it does not exist in the schema of parquet file. So This PR takes the case sensitive into consideration when specify the column names.

### DIFF
--- a/integration_tests/src/main/python/parquet_test.py
+++ b/integration_tests/src/main/python/parquet_test.py
@@ -477,14 +477,20 @@ _nested_pruning_schemas = [
         ([["a", StructGen([["c_1", StringGen()], ["c_2", LongGen()], ["c_3", ShortGen()]])]],
             [["a", StructGen([["c_3", ShortGen()], ["c_2", LongGen()], ["c_1", StringGen()]])]]),
         ([["ar", ArrayGen(StructGen([["str_1", StringGen()],["str_2", StringGen()]]))]],
-            [["ar", ArrayGen(StructGen([["str_2", StringGen()]]))]])
+            [["ar", ArrayGen(StructGen([["str_2", StringGen()]]))]]),
+        ([["struct", StructGen([["c_1", StringGen()], ["case_insensitive", LongGen()], ["c_3", ShortGen()]])]],
+            [["STRUCT", StructGen([["case_INSENsitive", LongGen()]])]]),
+        ([["struct", StructGen([["c_1", StringGen()], ["case_insensitive", LongGen()], ["c_3", ShortGen()]])]],
+            [["struct", StructGen([["CASE_INSENSITIVE", LongGen()]])]]),
+        ([["struct", StructGen([["c_1", StringGen()], ["case_insensitive", LongGen()], ["c_3", ShortGen()]])]],
+            [["stRUct", StructGen([["CASE_INSENSITIVE", LongGen()]])]]),
         ]
 
 @pytest.mark.parametrize('data_gen,read_schema', _nested_pruning_schemas, ids=idfn)
 @pytest.mark.parametrize('reader_confs', reader_opt_confs)
 @pytest.mark.parametrize('v1_enabled_list', ["", "parquet"])
 @pytest.mark.parametrize('nested_enabled', ["true", "false"])
-def test_nested_pruning(spark_tmp_path, data_gen, read_schema, reader_confs, v1_enabled_list, nested_enabled):
+def test_nested_pruning_and_case_insensitive(spark_tmp_path, data_gen, read_schema, reader_confs, v1_enabled_list, nested_enabled):
     data_path = spark_tmp_path + '/PARQUET_DATA'
     with_cpu_session(
             lambda spark : gen_df(spark, data_gen).write.parquet(data_path),
--- a/integration_tests/src/main/python/parquet_test.py
+++ b/integration_tests/src/main/python/parquet_test.py
@@ -675,3 +675,32 @@ def test_reading_from_unaligned_pages_basic_filters_with_nulls(spark_tmp_path, r
         assert_gpu_and_cpu_are_equal_collect(
                 lambda spark : spark.read.parquet(data_path).filter(filter_str),
                 all_confs)
+
+
+@pytest.mark.parametrize('v1_enabled_list', ["", "parquet"])
+@pytest.mark.parametrize('reader_confs', reader_opt_confs, ids=idfn)
+def test_user_defined_schema_with_case_insensitive(spark_tmp_path, reader_confs, v1_enabled_list):
+    all_confs = copy_and_update(reader_confs, {'spark.sql.sources.useV1SourceList': v1_enabled_list})
+
+    data_gens = [string_gen, int_gen, long_gen]
+    gen_list = [('c_' + str(i), gen) for i, gen in enumerate(data_gens)]
+    gen = StructGen([['struct', StructGen(gen_list, nullable=False)]], nullable=False)
+    data_path = spark_tmp_path + '/PARQUET_DATA'
+    with_cpu_session(
+            lambda spark : gen_df(spark, gen).write.parquet(data_path))
+
+    schema = StructType([StructField('struct', StructType([StructField('c_1', IntegerType())]))])
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: spark.read.schema(schema).parquet(data_path), all_confs)
+
+    schema = StructType([StructField('STRUCT', StructType([StructField('c_1', IntegerType())]))])
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: spark.read.schema(schema).parquet(data_path), all_confs)
+
+    schema = StructType([StructField('struct', StructType([StructField('C_1', IntegerType())]))])
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: spark.read.schema(schema).parquet(data_path), all_confs)
+
+    schema = StructType([StructField('stRUCt', StructType([StructField('C_1', IntegerType())]))])
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: spark.read.schema(schema).parquet(data_path), all_confs)
--- a/integration_tests/src/main/python/parquet_test.py
+++ b/integration_tests/src/main/python/parquet_test.py
@@ -681,32 +681,3 @@ def test_reading_from_unaligned_pages_basic_filters_with_nulls(spark_tmp_path, r
         assert_gpu_and_cpu_are_equal_collect(
                 lambda spark : spark.read.parquet(data_path).filter(filter_str),
                 all_confs)
-
-
-@pytest.mark.parametrize('v1_enabled_list', ["", "parquet"])
-@pytest.mark.parametrize('reader_confs', reader_opt_confs, ids=idfn)
-def test_user_defined_schema_with_case_insensitive(spark_tmp_path, reader_confs, v1_enabled_list):
-    all_confs = copy_and_update(reader_confs, {'spark.sql.sources.useV1SourceList': v1_enabled_list})
-
-    data_gens = [string_gen, int_gen, long_gen]
-    gen_list = [('c_' + str(i), gen) for i, gen in enumerate(data_gens)]
-    gen = StructGen([['struct', StructGen(gen_list, nullable=False)]], nullable=False)
-    data_path = spark_tmp_path + '/PARQUET_DATA'
-    with_cpu_session(
-            lambda spark : gen_df(spark, gen).write.parquet(data_path))
-
-    schema = StructType([StructField('struct', StructType([StructField('c_1', IntegerType())]))])
-    assert_gpu_and_cpu_are_equal_collect(
-        lambda spark: spark.read.schema(schema).parquet(data_path), all_confs)
-
-    schema = StructType([StructField('STRUCT', StructType([StructField('c_1', IntegerType())]))])
-    assert_gpu_and_cpu_are_equal_collect(
-        lambda spark: spark.read.schema(schema).parquet(data_path), all_confs)
-
-    schema = StructType([StructField('struct', StructType([StructField('C_1', IntegerType())]))])
-    assert_gpu_and_cpu_are_equal_collect(
-        lambda spark: spark.read.schema(schema).parquet(data_path), all_confs)
-
-    schema = StructType([StructField('stRUCt', StructType([StructField('C_1', IntegerType())]))])
-    assert_gpu_and_cpu_are_equal_collect(
-        lambda spark: spark.read.schema(schema).parquet(data_path), all_confs)
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
@@ -889,7 +889,9 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
 
     if (!isCaseSensitive) {
       val m = CaseInsensitiveMap(fields.zip(fields).toMap)
-      readDataSchema.fieldNames.map(m(_))
+      // The schemas may be different among parquet files, so some column names may not be existing
+      // in the CaseInsensitiveMap. In that case, we just use the field name of readDataSchema
+      readDataSchema.fieldNames.map { name => m.get(name).getOrElse(name) }
     } else {
       readDataSchema.fieldNames.toSeq
     }
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
@@ -874,9 +874,9 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
 
   /**
    * Take case-sensitive into consideration when getting the data reading column names
-   * before sending parquet-format buffer to cudf.
+   * before sending parquet-formatted buffer to cudf.
    *
-   * @param readDataSchema spark schema to read
+   * @param readDataSchema Spark schema to read
    * @param fileSchema  the schema of the dumped parquet-format buffer
    * @param isCaseSensitive if it is case sensitive
    * @return a sequence of column names following the order of readDataSchema
@@ -885,9 +885,9 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
       readDataSchema: StructType,
       fileSchema: MessageType,
       isCaseSensitive: Boolean): Seq[String] = {
-    val fields = fileSchema.getFields.asScala.map(_.getName).toSet
 
     if (!isCaseSensitive) {
+      val fields = fileSchema.asGroupType().getFields.asScala.map(_.getName).toSet
       val m = CaseInsensitiveMap(fields.zip(fields).toMap)
       // The schemas may be different among parquet files, so some column names may not be existing
       // in the CaseInsensitiveMap. In that case, we just use the field name of readDataSchema
@@ -1072,12 +1072,12 @@ class MultiFileParquetPartitionReader(
     // Dump parquet data into a file
     dumpDataToFile(dataBuffer, dataSize, splits, Option(debugDumpPrefix), Some("parquet"))
 
-    val includeColumn = toCudfColumnNames(readDataSchema, clippedSchema,
+    val includeColumns = toCudfColumnNames(readDataSchema, clippedSchema,
       isSchemaCaseSensitive)
     val parseOpts = ParquetOptions.builder()
       .withTimeUnit(DType.TIMESTAMP_MICROSECONDS)
       .enableStrictDecimalType(true)
-      .includeColumn(includeColumn: _*).build()
+      .includeColumn(includeColumns: _*).build()
 
     // About to start using the GPU
     GpuSemaphore.acquireIfNecessary(TaskContext.get(), metrics(SEMAPHORE_WAIT_TIME))
@@ -1331,12 +1331,12 @@ class MultiFileCloudParquetPartitionReader(
       // Dump parquet data into a file
       dumpDataToFile(hostBuffer, dataSize, files, Option(debugDumpPrefix), Some("parquet"))
 
-      val includeColumn = toCudfColumnNames(readDataSchema, clippedSchema,
+      val includeColumns = toCudfColumnNames(readDataSchema, clippedSchema,
         isSchemaCaseSensitive)
       val parseOpts = ParquetOptions.builder()
         .withTimeUnit(DType.TIMESTAMP_MICROSECONDS)
         .enableStrictDecimalType(true)
-        .includeColumn(includeColumn: _*).build()
+        .includeColumn(includeColumns: _*).build()
 
       // about to start using the GPU
       GpuSemaphore.acquireIfNecessary(TaskContext.get(), metrics(SEMAPHORE_WAIT_TIME))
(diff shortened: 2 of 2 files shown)


wbo4958:
build

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
```diff
@@ -871,6 +872,28 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
     currentChunk
   }
 
+  /**
+   * Take case-sensitive into consideration when getting the data reading column names
+   * before sending parquet-format buffer to cudf.
+   *
+   * @param readDataSchema spark schema to read
```
jlowe, line 879:
```suggestion
   * @param readDataSchema Spark schema to read
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
```diff
@@ -871,6 +872,28 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
     currentChunk
   }
 
+  /**
+   * Take case-sensitive into consideration when getting the data reading column names
+   * before sending parquet-format buffer to cudf.
```
jlowe, line 877:
```suggestion
   * before sending Parquet-formatted buffer to cudf.
```

Also applies to other instance of this below.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
```diff
@@ -871,6 +872,28 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
...
+   * @param isCaseSensitive if it is case sensitive
+   * @return a sequence of column names following the order of readDataSchema
+   */
+  protected def toCudfColumnNames(
+      readDataSchema: StructType,
+      fileSchema: MessageType,
+      isCaseSensitive: Boolean): Seq[String] = {
+    val fields = fileSchema.getFields.asScala.map(_.getName).toSet
+
+    if (!isCaseSensitive) {
```
jlowe, line 890: Move `fields` into the only block that needs it.
```suggestion
    if (!isCaseSensitive) {
      val fields = fileSchema.getFields.asScala.map(_.getName).toSet
```

tgravescs:
Are there any configs related to this? can you describe the behavior of Spark in this case?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
```diff
@@ -1047,10 +1070,12 @@ class MultiFileParquetPartitionReader(
     // Dump parquet data into a file
     dumpDataToFile(dataBuffer, dataSize, splits, Option(debugDumpPrefix), Some("parquet"))
 
+    val includeColumn = toCudfColumnNames(readDataSchema, clippedSchema,
```
jlowe, line 1073: Nit: This should probably be a pluralized name (e.g. `includeColumns`) since it's a collection of columns.  Applies to other places where this code was copied.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
```diff
@@ -871,6 +872,28 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
...
+   */
+  protected def toCudfColumnNames(
+      readDataSchema: StructType,
+      fileSchema: MessageType,
+      isCaseSensitive: Boolean): Seq[String] = {
+    val fields = fileSchema.getFields.asScala.map(_.getName).toSet
+
+    if (!isCaseSensitive) {
+      val m = CaseInsensitiveMap(fields.zip(fields).toMap)
+      readDataSchema.fieldNames.map(m(_))
```
jlowe, line 892: Test cases are failing for missing columns, this will need to gracefully handle the case where lookup fails.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
```diff
@@ -871,6 +872,28 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
...
+   * @param readDataSchema spark schema to read
+   * @param fileSchema  the schema of the dumped parquet-format buffer
+   * @param isCaseSensitive if it is case sensitive
+   * @return a sequence of column names following the order of readDataSchema
+   */
+  protected def toCudfColumnNames(
+      readDataSchema: StructType,
+      fileSchema: MessageType,
+      isCaseSensitive: Boolean): Seq[String] = {
+    val fields = fileSchema.getFields.asScala.map(_.getName).toSet
```
jlowe, line 888: Is this approach sufficient to handle nested struct field names, or will this only handle top-level column names?

sameerz:
Will we add a test case for this fix?

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
wbo4958: Done. Thx

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
wbo4958: Done. Thx

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
wbo4958: Dibe

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
wbo4958: Done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
wbo4958: Yeah, since `ParquetReadSupport.clipParquetSchema` has helped us to clip out the unnecessary columns

wbo4958:
> Are there any configs related to this? can you describe the behavior of Spark in this case?

Before reading parquet file, Spark always set the "id" data reader schema while the schema of parquet file is "ID". Since spark have set to `spark.sql.caseSensitive` to false by default, which means the parquet reader will treat **"id"/"ID"** to be same, and read without any issue.

If I set `spark.sql.caseSensitive` to true, then Spark reads all "null" value.

wbo4958:
build

wbo4958:
> Will we add a test case for this fix?

It's hard to add test since it will involve mysql/hive.

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
```diff
@@ -1047,10 +1072,12 @@ class MultiFileParquetPartitionReader(
     // Dump parquet data into a file
     dumpDataToFile(dataBuffer, dataSize, splits, Option(debugDumpPrefix), Some("parquet"))
 
+    val includeColumns = toCudfColumnNames(readDataSchema, clippedSchema,
```
tgravescs, line 1075: do we have the same issue with the orc reader?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
```diff
@@ -871,6 +872,30 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
     currentChunk
   }
 
+  /**
+   * Take case-sensitive into consideration when getting the data reading column names
+   * before sending parquet-formatted buffer to cudf.
+   *
+   * @param readDataSchema Spark schema to read
+   * @param fileSchema  the schema of the dumped parquet-formatted buffer
```
tgravescs, line 880: nit extra space after fileSchema

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
```diff
@@ -871,6 +872,30 @@ trait ParquetPartitionReaderBase extends Logging with Arm with ScanWithMetrics
...
+      readDataSchema: StructType,
+      fileSchema: MessageType,
+      isCaseSensitive: Boolean): Seq[String] = {
+
+    if (!isCaseSensitive) {
+      val fields = fileSchema.asGroupType().getFields.asScala.map(_.getName).toSet
+      val m = CaseInsensitiveMap(fields.zip(fields).toMap)
+      // The schemas may be different among parquet files, so some column names may not be existing
+      // in the CaseInsensitiveMap. In that case, we just use the field name of readDataSchema
+      readDataSchema.fieldNames.map { name => m.get(name).getOrElse(name) }
```
tgravescs, line 894: might be nice to add comment saying when case insensitive readDataSchema is lower case

tgravescs:
Perhaps we need tests that only run in certain envs, like setup to run on the YARN cluster

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
jlowe: This comment was not addressed.  Is there a commit that has not been pushed to Github yet?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
jlowe: I'm not sure why clipping has anything to do with my question, as this isn't a problem of failing to remove columns and accidentally loading the wrong column but rather mapping Spark's requested names to the column names in the file.

anfeng:
@wbo4958 is it hard to enable Hive Metastore in our integration tests? https://spark.apache.org/docs/latest/sql-data-sources-hive-tables.html stated "Users who do not have an existing Hive deployment can still enable Hive support. When not configured by the hive-site.xml, the context automatically creates metastore_db in the current directory and creates a directory configured by spark.sql.warehouse.dir, which defaults to the directory spark-warehouse in the current directory that the Spark application is started."

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
wbo4958: It's strange, I just checked the source file and it should be changed to parquet-formatted

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
wbo4958: Sorry for my wrong answer. Yeah, we only handle the top-level columns.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
wbo4958: Done

wbo4958: Done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
wbo4958: Done

wbo4958:
build

wbo4958:
> @wbo4958 is it hard to enable Hive Metastore in our integration tests? https://spark.apache.org/docs/latest/sql-data-sources-hive-tables.html stated "Users who do not have an existing Hive deployment can still enable Hive support. When not configured by the hive-site.xml, the context automatically creates metastore_db in the current directory and creates a directory configured by spark.sql.warehouse.dir, which defaults to the directory spark-warehouse in the current directory that the Spark application is started."

Hi @anfeng, I just tried the example provided in https://spark.apache.org/docs/latest/sql-data-sources-hive-tables.html. It could work without any issue, but it can't simulate the issue this PR is trying to fix. The schema name of readData is normal, while the issue we're facing is that the schema from hive metastore is always lower-case.

For the test, seems we can setup a jenkins pipeline to setup mysql/hive/spark env and run an example.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
jlowe: Is there a followup needed for this, or does the case-insensitivity only apply to top-level columns?

tgravescs:
this is why I mentioned the YARN cluster, we already run IT tests there and it should have hive already setup.  So could we have a test that checks if its setup and runs?  It would be nice to somehow make sure it runs on that setup though too so that is something on the cluster gets messed up the test could never get run.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
jlowe: That's why I asked if there's a commit that hasn't been pushed to Github yet.  According to https://github.com/NVIDIA/spark-rapids/pull/3982/files#diff-8fa3aa74be556181661b80e4735b00e8d453311293c30ccc6d92353df4636af9R877 it's not updated yet, and I also pulled the source locally to verify it's not some weird browser cached state problem with Github's rendering:
```
$ git fetch https://github.com/wbo4958/spark-rapids.git hive-issue
From https://github.com/wbo4958/spark-rapids
 * branch                hive-issue -> FETCH_HEAD
$ git co FETCH_HEAD
HEAD is now at 16c081d8c resolve comments
$ grep parquet-formatted sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala 
   * before sending parquet-formatted buffer to cudf.
   * @param fileSchema the schema of the dumped parquet-formatted buffer
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
wbo4958: Seems we don't need a follow up. Since ParquetReadSupport.clipParquetSchema will help us to filter the necessary columns according to the flag of case sensitive, which means all columns are needed in the clipped schema. So in case-insensitive scenario, matching the top-level columns should be ok.

wbo4958:
> this is why I mentioned the YARN cluster, we already run IT tests there and it should have hive already setup. So could we have a test that checks if its setup and runs? It would be nice to somehow make sure it runs on that setup though too so that is something on the cluster gets messed up the test could never get run.

I just asked some guys about the Jenkins pipeline, and seems we don't have the hive env yet. Looks like we need to set it up from 0 to 1.

pxLi:
> > this is why I mentioned the YARN cluster, we already run IT tests there and it should have hive already setup. So could we have a test that checks if its setup and runs? It would be nice to somehow make sure it runs on that setup though too so that is something on the cluster gets messed up the test could never get run.
> 
> I just asked some guys about the Jenkins pipeline, and seems we don't have the hive env yet

I think we said no mysql setup, but for hive we have a yarn cluster sitting there for years

wbo4958:
> > > this is why I mentioned the YARN cluster, we already run IT tests there and it should have hive already setup. So could we have a test that checks if its setup and runs? It would be nice to somehow make sure it runs on that setup though too so that is something on the cluster gets messed up the test could never get run.
> > 
> > 
> > I just asked some guys about the Jenkins pipeline, and seems we don't have the hive env yet
> 
> I think we said no mysql setup, but for hive we have a yarn cluster sitting there for years

In that case, we can set up mysql and add test for it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
jlowe: > Since ParquetReadSupport.clipParquetSchema will help us to filter the necessary columns according to the flag of case sensitive

We're filtering the columns based on case-insensitivity, but are we requesting them in that way as well?  Specifically I'm thinking of the nested pruning case, where we ask the reader to load individual fields of a struct column without loading the entire struct column.

For example, consider a struct column named x containing three fields: a, b, c.  Spark asks to load column X.A.  We will clip to just include x.a, but when we ask libcudf to load the column, will we also fix the column name to be x.a or will we ask for x.A?  The latter will probably fail since libcudf is case-sensitive in its column lookup.  We probably need a test case for nested pruning case insensitivity handling.

wbo4958: Hi Jason, Yeah, we're only caring the top-level columns. For your example, we only pass the top-level "x" to libcudf.

wbo4958: I just tried to specify the user defined schema which can repro this issue without involving any hive things. And the new commit has added the integration test.

wbo4958:
build

wbo4958:
Hi @jlowe @tgravescs, I just repro this issue the PR is trying to resolve without involving hive things. and added the integration test. Thx for reviewing. 

BTW, since CUDF jar has changed a RMM API and I just rebased this PR to the HEAD. So this PR has a force-push. Sorry for causing any trouble.

jlowe:
> BTW, since CUDF jar has changed a RMM API and I just rebased this PR to the HEAD. So this PR has a force-push.

Why was it rebased?  Updating to the latest changes doesn't require a rebase.  Instead you can merge `branch-21.12` into this PR's branch.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
jlowe: > Yeah, we're only caring the top-level columns.

For nested pruning that's not true, correct?  That's a case where we're not loading an entire top-level struct column but only individual fields from it.

integration_tests/src/main/python/parquet_test.py
```diff
@@ -675,3 +675,32 @@ def test_reading_from_unaligned_pages_basic_filters_with_nulls(spark_tmp_path, r
...
+
+    schema = StructType([StructField('STRUCT', StructType([StructField('c_1', IntegerType())]))])
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: spark.read.schema(schema).parquet(data_path), all_confs)
+
+    schema = StructType([StructField('struct', StructType([StructField('C_1', IntegerType())]))])
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: spark.read.schema(schema).parquet(data_path), all_confs)
+
+    schema = StructType([StructField('stRUCt', StructType([StructField('C_1', IntegerType())]))])
```
jlowe, line 704: Would like to see a nested pruning case.  See `test_nested_pruning` for an example, we just need to make the requested subset schema not match the exact case found in the file.

integration_tests/src/main/python/parquet_test.py
```diff
@@ -675,3 +675,32 @@ def test_reading_from_unaligned_pages_basic_filters_with_nulls(spark_tmp_path, r
...
+    data_gens = [string_gen, int_gen, long_gen]
+    gen_list = [('c_' + str(i), gen) for i, gen in enumerate(data_gens)]
+    gen = StructGen([['struct', StructGen(gen_list, nullable=False)]], nullable=False)
+    data_path = spark_tmp_path + '/PARQUET_DATA'
+    with_cpu_session(
+            lambda spark : gen_df(spark, gen).write.parquet(data_path))
+
+    schema = StructType([StructField('struct', StructType([StructField('c_1', IntegerType())]))])
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: spark.read.schema(schema).parquet(data_path), all_confs)
```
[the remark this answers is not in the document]
wbo4958: Yes, make sense, remove the whole test and move it into test_nested_pruning

integration_tests/src/main/python/parquet_test.py
wbo4958: Thx, I just removed this test case and moved it into test_nested_pruning

wbo4958:
build

wbo4958:
build

wbo4958:
build

jlowe (approved):

PR #3984, opened 2021-11-01
Check for data write command before inserting hash sort optimization

### LINKED ISSUE #3730: [BUG] Hash optimized sort insertion should only insert on plans writing output
When enabling a sort to optimize the storage result of a hash-based operation via `spark.rapids.sql.hashOptimizeSort.enabled=true`, the logic should examine the plan looking for root nodes that are storing data to disk (e.g.: `DataWritingCommand` and anything similar).  When AQE is enabled we may be optimizing only a portion of the plan.  Therefore we should not insert an order optimizing sort unless we have high confidence there are no further stages in the plan, otherwise we end up inserting an extra sort whose results could easily be made redundant by further operations in the query in a subsequent AQE planning stage (e.g.: another hash-based operation, another sort, etc.).

res-life:
This fixes #3730 


res-life:
Does not handle AQE scenario, because of not have high confidence.
Logic is: 
  // find
  // GpuDataWritingCommandExec(GpuHashJoin)
  // GpuDataWritingCommandExec(GpuHashAggregateExec)
  // GpuDataWritingCommandExec(GpuProject(GpuHashJoin))
  // GpuDataWritingCommandExec(GpuProject(GpuHashAggregateExec))
  // from top to bottom and insert sort to optimize the file size.

res-life:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -407,31 +407,50 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
     }
   }
 
+  // find
+  // GpuDataWritingCommandExec(GpuHashJoin)
+  // GpuDataWritingCommandExec(GpuHashAggregateExec)
+  // GpuDataWritingCommandExec(GpuProject(GpuHashJoin))
+  // GpuDataWritingCommandExec(GpuProject(GpuHashAggregateExec))
```
jlowe, line 414: I don't understand why this is looking for very specific patterns leading to a write.  It seems likely to break because we forgot some pattern or some new pattern shows up later.  Note the original comment says:
```
// Insert a sort after the last hash-based op before the query result if there are no
// intermediate nodes that have a specified sort order.
```
meaning we don't care about any nodes between the hashing node and the writing node if they don't change the sort order.  The original code did not require that the hashing node appeared immediately before the write, but this effectively does (with some hardcoded exceptions for a project).

res-life:
build

res-life:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
res-life: Yes, you are right, updated the code to follow the original logic.
Please review again.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -417,15 +422,25 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
...
+          } else {
+            plan
+          }
+        case hashAgg: GpuHashAggregateExec =>
+          if (hasWriteParent) {
+            val gpuSortOrder = getOptimizedSortOrder(plan)
+            GpuSortExec(gpuSortOrder, false, hashAgg, SortEachBatch)(gpuSortOrder)
+          } else {
+            plan
+          }
```
jlowe, line 440: Since we treat the hash join and hash aggregate the same, we can simplify this a bit:
```suggestion
        case _: GpuHashJoin | _: GpuHashAggregateExec if hasWriteParent =>
            val gpuSortOrder = getOptimizedSortOrder(plan)
            GpuSortExec(gpuSortOrder, false, plan, SortEachBatch)(gpuSortOrder)
        case _: GpuHashJoin | _: GpuHashAggregateExec => plan
```

tests/src/test/scala/com/nvidia/spark/rapids/HashSortOptimizeSuite.scala
```diff
@@ -121,12 +123,73 @@ class HashSortOptimizeSuite extends FunSuite {
...
+      df.collect()
+
+      // write should be on GPU
+      val sortExec = TestUtils.findOperator(df.queryExecution.executedPlan,
+        _.isInstanceOf[GpuSortExec])
+      assert(sortExec.isDefined)
+    }, conf)
+  }
+
+  test("should not insert sort because of without GpuDataWritingCommandExec") {
```
jlowe, line 153:
```suggestion
  test("should not insert sort because missing GpuDataWritingCommandExec") {
```

tests/src/test/scala/com/nvidia/spark/rapids/HashSortOptimizeSuite.scala
```diff
@@ -17,14 +17,13 @@
 package com.nvidia.spark.rapids
```
jlowe, line 17: Need to add 2021 copyrights for this file.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -407,7 +407,12 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
     }
   }
 
-  private def insertHashOptimizeSorts(plan: SparkPlan): SparkPlan = {
+  // Find  GpuDataWritingCommandExec(......(GpuHashJoin)),
+  //       GpuDataWritingCommandExec(......(GpuHashAggregateExec))
+  // from top to bottom and insert sort to optimize the file size.
+  // such as: GpuDataWritingCommandExec(GpuProject(GpuHashAggregateExec))
+  private def insertHashOptimizeSorts(plan: SparkPlan,
+                                      hasWriteParent:Boolean = false): SparkPlan = {
```
andygrove, line 415: nit: indent doesn't match style guide https://github.com/databricks/scala-style-guide#indent

```suggestion
                                      hasWriteParent:Boolean = false): SparkPlan = {
      hasWriteParent:Boolean = false): SparkPlan = {
```

res-life:
build

res-life:
build

res-life:
updated all the comments, please review.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -407,7 +406,12 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
     }
   }
 
-  private def insertHashOptimizeSorts(plan: SparkPlan): SparkPlan = {
+  // Find  GpuDataWritingCommandExec(......(GpuHashJoin)),
+  //       GpuDataWritingCommandExec(......(GpuHashAggregateExec))
+  // from top to bottom and insert sort to optimize the file size.
+  // such as: GpuDataWritingCommandExec(GpuProject(GpuHashAggregateExec))
```
jlowe, line 412: This comment is no longer accurate, suggest something like:
```suggestion
  // If a GPU hash-based operation, such as GpuHashJoin or GpuHashAggregateExec,
  // is followed eventually by a data writing command without an intermediate node
  // changing the sort order, insert a sort to optimize the output file size.
```

jlowe (approved):
Minor nit on an outdated comment but otherwise lgtm.

res-life:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
res-life: done

jlowe (approved):

PR #3986, opened 2021-11-01
Avoid fetching batch likely to exceed target size when coalescing batches

### LINKED ISSUE #3750: Improve batch coalesce logic
Currently the logic for coalescing a batch will pull in a new batch as long as the current total size of all held batches is below the target size regardless of how close we are to the target size or the size of the batches that have been seen.  This can cause a batch to be manifested in memory that cannot be added to the set of batches to coalesce for the next output batch, forcing it to be made spillable and adding undesired memory pressure.  The logic tries very hard to reach as close as it can to the target size, causing extra batches to be held in memory as a result.

The logic for coalescing should be a bit smarter, avoiding pulling in a new batch if it appears likely that the new batch will cause the next output batch to exceed the target size.  One idea is to track the maximum size of any batch seen so far and only fetch the next batch if there is enough room to accommodate another batch of that size for the next coalesced output batch.

jlowe:
Closes #3750.

Updates the batch coalesce logic to track the largest number of rows and size seen in a single input batch to guesstimate whether the next batch is likely to fit in the current set of batches being coalesced within the specified target size.  Avoiding pulling in a batch that is unlikely to fit should help reduce the number of times we need to call `contiguous_split` to make the batch splittable and reduce the memory pressure from caching the manifested batch that we cannot immediately process.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
@@ -214,6 +214,8 @@ abstract class AbstractGpuCoalesceIterator(
   private val iter = new CollectTimeIterator(s"$opName: collect", batches, streamTime)
 
   private var batchInitialized: Boolean = false
+  private var largestInputBatchRows: Long = 0L
+  private var largestInputBatchSize: Long = 0L
 
   /**
    * Return true if there is something saved on deck for later processing.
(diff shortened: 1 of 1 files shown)


jlowe:
Marking this a draft while this is run through some performance tests.

jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
```diff
@@ -336,8 +338,22 @@ abstract class AbstractGpuCoalesceIterator(
...
-      // there is a hard limit of 2^31 rows
-      while (numRows < Int.MaxValue && !hasOnDeck && iter.hasNext) {
+      def shouldFetchNextBatch: Boolean =
+        if (numRows < Int.MaxValue && !hasOnDeck && iter.hasNext) {
+          goal match {
+            case g: TargetSize =>
+              // avoid fetching if next batch is likely to exceed the specified row or size limits
+              // based on the stats of previous batches
+              val rowLimit = if (batchRowLimit > 0) batchRowLimit else Int.MaxValue
+              numRows + largestInputBatchSize <= rowLimit &&
```
revans2, line 348: It looks like you switches size and rows here. The check has `largestInputBatchSize` paired with row checks, and `largetInputBatchRows` paired with bytes checks.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
```diff
@@ -346,6 +362,8 @@ abstract class AbstractGpuCoalesceIterator(
           if (nextRows > 0) {
             numInputRows += nextRows
             val nextBytes = getBatchDataSize(cb)
+            largestInputBatchSize = largestInputBatchSize.max(nextRows)
```
revans2, line 365: Here it is switched too. So it looks like you are doing the right thing, but the names really confuse me.

jlowe: Yeah that is super confusing.  I could see accidentally switching one, but not sure how I managed to switch both. 🤦

jlowe:
build

revans2 (dismissed):

jlowe:
Odd, premerge CI failed with no sanitized log.

jlowe:
build

jlowe:
Profiling on NDS queries shows this is as likely to hurt as to help, so closing this for now.

PR #3989, opened 2021-11-01
Upmerge Decimal128 to latest spark-rapids branch

revans2:
The main reason this is a PR is because there were some merge conflicts around aggregation.


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -642,81 +643,77 @@ case class GpuSum(child: Expression,
       with GpuAggregateWindowFunction
```
revans2, line 643: @abellina could you take a look at this file. Specifically around GpuSum and GpuAverage?  This is the only place that actually had any real merge conflicts.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -268,7 +339,7 @@ class CudfSum(ref: Expression) extends CudfAggregate(ref) {
   //
   @transient val rapidsSumType: DType = GpuColumnVector.getNonNestedRapidsType(dataType)
 
-  override val updateReductionAggregateInternal: cudf.ColumnVector => cudf.Scalar = {
+  override val reductionAggregate: cudf.ColumnVector => cudf.Scalar = {
     dataType match {
       case _: DecimalType =>
         (col: cudf.ColumnVector) => withResource(col.castTo(rapidsSumType)) { tmp =>
```
abellina, line 345: I would have expected the `GpuAggregateFunction` invoking `CudfSum` to cast in `inputProjection` and `preMerge`. Do we need to to cast inside `CudfSum`?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -642,81 +643,77 @@ case class GpuSum(child: Expression,
...
-
-  override lazy val aggBufferAttributes: Seq[AttributeReference] = resultType match {
-    case _: DecimalType => cudfSum :: isEmpty :: Nil
-    case _ => cudfSum :: Nil
-  }
-
-  override lazy val initialValues: Seq[GpuLiteral] = resultType match {
-    case _: DecimalType => Seq(zeroDec, GpuLiteral(true, BooleanType))
-    case _ => Seq(GpuLiteral(null, resultType))
+      GpuCheckOverflowAfterSum(sum, isEmpty, dt, !failOnErrorOverride)
```
abellina, line 715: a side note, `GpuCheckOverflowAfterSum` has the logic reversed on when it should keep nulls or when it should throw in its impl.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -642,81 +643,77 @@ case class GpuSum(child: Expression,
...
 
-  override lazy val updateExpressions: Seq[Expression] = resultType match {
-    case _: DecimalType => Seq(new CudfSum(cudfSum), new CudfMin(isEmpty))
-    case _ => Seq(new CudfSum(cudfSum))
+  override lazy val updateAggregates: Seq[CudfAggregate] = resultType match {
+    case _: DecimalType => Seq(new CudfSum(resultType), new CudfMin(BooleanType))
+    case _ => Seq(new CudfSum(resultType))
   }
 
+  private lazy val postUpdateSum = updateAggregates.head.attr
```
abellina, line 670: Minor but I think it can be cleaner (than referring to the members of `updateAggregates` positionally) 

Instead of referencing the `attr` here, I'd propose we define:
```
lazy val updateSum = new CudfSum(resultType)
lazy val updateIsEmpty = new CudfMin(BooleanType)
```

Then `updateAggregates` refers to these as such:
```
  override lazy val updateAggregates: Seq[CudfAggregate] = resultType match {
    case _: DecimalType => Seq(updateSum, updateIsEmpty)
    case _ => Seq(updateSum)
  }
```

Allowing us to pick the `attr` member in `postUpdate` like so:

```
  override lazy val postUpdate: Seq[Expression] = resultType match {
    case dt: DecimalType =>
      Seq(GpuCheckOverflow(updateSum.attr, dt, !failOnErrorOverride), updateIsEmpty.attr)
    case _ => updateSum.attr
  }
```

abellina: This pattern can be applied in other places where the attributes are picked using head or using an index in other `GpuAggregateFunction` impls.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
revans2: Sorry I should have had a better check. I didn't want to spend an extra min of test time just to check overflow for decimals on SUM, but I probably should have done it.

revans2:
@abellina I have addressed your comments. Please take another look

abellina (approved):
LGTM

Issue #3764, opened 2021-10-07
[FEA] In qualification tool output, remove some "unsupported" words


viadea:
**Is your feature request related to a problem? Please describe.**
In qualification tool output, remove some "unsupported" words.
I would propose we change:
Unsupported Read File Formats and Types =>
Problematic Read File Formats and Types

Unsupported Nested Complex Types => 
Nested Complex Types

viadea:
@tgravescs Sure i am fine with leaving read formats unchanged.Thanks for the explanation.

How about only changing:
Unsupported Nested Complex Types =>
Nested Complex Types

Since we are adding more and more data nested data type support recently in 21.10.
Unless there is a similar way to have a support matrix in the qualification tool similar as above read file formats, then we should be good.

PR #3993, opened 2021-11-02
Qualification tool: Remove "unsupported" word for nested complex types

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3764 .

This is a minor PR to remove "Unsupported" word for nested complex types.  Updated the output files to match the new format.


nartal1:
build

tgravescs (dismissed):
please watch build to see if we need to update qualification IT tests

docs/spark-qualification-tool.md
```diff
@@ -380,7 +380,7 @@ Here is a brief description of each of column that is in the CSV:
 15. Unsupported Write Data Format: Reports the data format which we currently don’t support, i.e.
     if the result is written in JSON or CSV format.
 16. Complex Types: Looks at the Read Schema and reports if there are any complex types(array, struct or maps) in the schema.
-17. Unsupported Nested Complex Types: Nested complex types are complex types which
+17. Nested Complex Types: Nested complex types are complex types which
```
tgravescs, line 383: actually if we aren't saying these are unsupported do we need more text here stating why we are reporting these separately?  Should we say something like not all nested types supported, please see the docs for the release you are using?
@viadea ideas here to make sure its clear why this field is here?

viadea: I think we do claim that there is 
```
Here the Problematic Duration is zero but please keep in mind that we are only able to detect certain issues. This currently includes some UDFs, some decimal operations and nested complex types. 
```
So I think this should be fine.

nartal1: @tgravescs Please let me know if you think more context is required here in the docs. Or the one which @viadea has posted would be sufficient.

tgravescs: ok its fine.

tgravescs (approved):

Issue #3942, opened 2021-10-28
[FEA] Support arrays with nulls in the input of `GpuRowBasedUDF` for Spark 3.1.1+


firestarman:
This is a follow on issue for #3897.

When the input of `GpuRowBasedUDF` contains arrays with nulls, the UDF will run into the below error for Spark 3.1.1+ during our Unit tests where the Java assertion is enabled.
```
  org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 27.0 failed 1 times, most recent failure: Lost task 0.0 in stage 27.0 (TID 27) (liangcail-ubuntu18 executor driver): java.lang.AssertionError:  value at 6 is null
	at ai.rapids.cudf.HostColumnVectorCore.assertsForGet(HostColumnVectorCore.java:228)
	at ai.rapids.cudf.HostColumnVectorCore.getInt(HostColumnVectorCore.java:254)
	at com.nvidia.spark.rapids.RapidsHostColumnVectorCore.getInt(RapidsHostColumnVectorCore.java:107)
	at org.apache.spark.sql.vectorized.ColumnarArray.getInt(ColumnarArray.java:128)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificSafeProjection.MapObjects_0$(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificSafeProjection.apply(Unknown Source)
	at org.apache.spark.sql.catalyst.encoders.ExpressionEncoder$Deserializer.apply(ExpressionEncoder.scala:182)
	at org.apache.spark.sql.rapids.GpuRowBasedScalaUDF.$anonfun$scalaConverter$2(GpuScalaUDF.scala:123)
	at org.apache.spark.sql.rapids.GpuRowBasedScalaUDF.$anonfun$childAccessors$2(GpuScalaUDF.scala:156)
```

So in the PR #3897, we let it fall back to CPU for this case for now.
However, we need to eventually support this case for Spark 3.1.1+, since we already do for Spark3.0.x.

[Update] Everything works well as the CPU under production envs where the Java assertion is disabled.

firestarman:
The null error mentioned in description is probably a bug of code generation in Spark .

The below code is generated from the 3 lines [1095-1097](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala#L1095) for this test case here.
[...]

PR #3995, opened 2021-11-02
Use Catalyst converter for array with nulls

firestarman:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3942

### DIFF
--- a/sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRowBasedScalaUDF.scala
+++ b/sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRowBasedScalaUDF.scala
@@ -22,7 +22,7 @@ import org.apache.spark.sql.catalyst.{CatalystTypeConverters, InternalRow}
 import org.apache.spark.sql.catalyst.encoders.ExpressionEncoder
 import org.apache.spark.sql.catalyst.expressions.{Expression, GenericInternalRow, ScalaUDF}
 import org.apache.spark.sql.rapids.{GpuRowBasedScalaUDFBase, ScalaUDFMetaBase}
-import org.apache.spark.sql.types.DataType
+import org.apache.spark.sql.types.{ArrayType, DataType}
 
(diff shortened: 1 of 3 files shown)


firestarman:
build

firestarman:
> Do we need this change if #3997 is merged?

Yes, but not all of the code.  The change in file `GpuRowBasedScalaUDF` will be optional.

sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRowBasedScalaUDF.scala
```diff
@@ -55,10 +55,15 @@ case class GpuRowBasedScalaUDF(
...
+    val arrayWithNull = dataType match {
+      case ArrayType(_, containsNull) => containsNull
+      case _ => false
+    }
     val useEncoder =
       !(inputEncoders.isEmpty || // for untyped Scala UDF and Java UDF
         inputEncoders(i).isEmpty || // for types aren't supported by encoder, e.g. Any
-        inputPrimitives(i)) // for primitive types
+        inputPrimitives(i) || // for primitive types
+        arrayWithNull) // https://github.com/NVIDIA/spark-rapids/issues/3942#issuecomment-957146443
```
jlowe, line 66: I thought the Catalyst converter doesn't handle some UDFs that would expect a structured class as input and therefore the Catalyst converters are not appropriate?  This change seems like it could break some UDFs since an input encoder exists and we're not using it when Apache Spark would.

firestarman: Yes, you are right.
This change could break some UDFs with the array element types not supported by Catalyst converters.

So, seems we need to wait for the Spark fix.  I will close this PR. Thanks.

firestarman:
This change could break some UDFs with the array element types not supported by Catalyst converters.

So, seems we need to wait for the Spark fix. Close this PR.

Issue #3996, opened 2021-11-02
[BUG] Exception happened when getting a null row


wbo4958:
This issue is for https://github.com/NVIDIA/spark-rapids/issues/3942

## Exception

``` console
  Cause: java.lang.AssertionError: value at 6 is null
  at ai.rapids.cudf.HostColumnVectorCore.assertsForGet(HostColumnVectorCore.java:228)
  at ai.rapids.cudf.HostColumnVectorCore.getInt(HostColumnVectorCore.java:254)
  at com.nvidia.spark.rapids.RapidsHostColumnVectorCore.getInt(RapidsHostColumnVectorCore.java:126)
  at org.apache.spark.sql.vectorized.ColumnarArray.getInt(ColumnarArray.java:128)
  at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificSafeProjection.MapObjects_0$(generated.java:57)
  at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificSafeProjection.apply(generated.java:28)
  at org.apache.spark.sql.catalyst.encoders.ExpressionEncoder$Deserializer.apply(ExpressionEncoder.scala:181)
  at com.nvidia.spark.rapids.shims.v2.GpuRowBasedScalaUDF.$anonfun$scalaConverter$2(GpuRowBasedScalaUDF.scala:70)
  at org.apache.spark.sql.rapids.GpuRowBasedScalaUDFBase.$anonfun$childAccessors$2(GpuScalaUDF.scala:141)
```

## Root Cause,

Spark may have some bug on the CodeGen for MapObjects, please see https://github.com/apache/spark/blob/branch-3.2/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala#L1095-L1097

The generated code is like below 
``` java
  private scala.collection.mutable.WrappedArray MapObjects_0(InternalRow i) {
    boolean isNull_1 = i.isNullAt(0);
    ArrayData value_1 = isNull_1 ?
        null : (i.getArray(0));
    scala.collection.mutable.WrappedArray value_0 = null;

    if (!isNull_1) {

      int dataLength_0 = value_1.numElements();

      scala.collection.mutable.Builder collectionBuilder_0 =
          scala.collection.mutable.WrappedArray$.MODULE$.newBuilder();
      collectionBuilder_0.sizeHint(dataLength_0);


      int loopIndex_0 = 0;

[...]

PR #3997, opened 2021-11-02
Return the default value when getting the null row

wbo4958:
This is to fix https://github.com/NVIDIA/spark-rapids/issues/3996

### DIFF
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
@@ -89,36 +89,57 @@ public class RapidsHostColumnVectorCore extends ColumnVector {
 
   @Override
   public final boolean getBoolean(int rowId) {
+    if (isNullAt(rowId)) {
+      return false;
+    }
     return cudfCv.getBoolean(rowId);
   }
 
   @Override
   public final byte getByte(int rowId) {
+    if (isNullAt(rowId)) {
+      return 0;
+    }
     return cudfCv.getByte(rowId);
   }
 
   @Override
   public final short getShort(int rowId) {
+    if (isNullAt(rowId)) {
+      return 0;
+    }
     return cudfCv.getShort(rowId);
   }
 
   @Override
   public final int getInt(int rowId) {
+    if (isNullAt(rowId)) {
+      return 0;
+    }
     return cudfCv.getInt(rowId);
   }
 
   @Override
   public final long getLong(int rowId) {
+    if (isNullAt(rowId)) {
+      return 0L;
+    }
     return cudfCv.getLong(rowId);
   }
 
   @Override
   public final float getFloat(int rowId) {
+    if (isNullAt(rowId)) {
+      return 0.0f;
+    }
     return cudfCv.getFloat(rowId);
   }
 
   @Override
   public final double getDouble(int rowId) {
+    if (isNullAt(rowId)) {
+      return 0.0f;
+    }
     return cudfCv.getDouble(rowId);
   }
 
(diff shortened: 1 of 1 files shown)


sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
```diff
@@ -89,36 +89,57 @@ public final boolean isNullAt(int rowId) {
...
+    if (isNullAt(rowId)) {
+      return 0.0f;
+    }
     return cudfCv.getFloat(rowId);
   }
 
   @Override
   public final double getDouble(int rowId) {
+    if (isNullAt(rowId)) {
+      return 0.0f;
```
firestarman, line 141:
```suggestion
      return 0.0;
```

sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
```diff
@@ -165,16 +186,19 @@ public final Decimal getDecimal(int rowId, int precision, int scale) {
...
-      return Decimal.createUnsafe(cudfCv.getLong(rowId), precision, scale);
+      return Decimal.createUnsafe(getLong(rowId), precision, scale);
     }
 
   }
 
   @Override
   public final UTF8String getUTF8String(int rowId) {
+    if (isNullAt(rowId)) {
+      return UTF8String.fromBytes(new byte[]{0});
```
firestarman, line 200: ?
```suggestion
      return UTF8String.fromBytes("".getBytes("UTF8"));
```

wbo4958:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/RapidsHostColumnVectorCore.java
jlowe: Curious if this could simply return `null`, since the calling code shouldn't be examining the value if the entry is marked as null.

revans2:
Do we know if this is a key of a value in the map?  Maps are not allowed to have null keys, so if we are showing up with a null in a key we have a bug and we should not be adding it a lot of performance overhead everywhere to work around it.

revans2 (changes requested):
Until I know if this is a key of a value that is causing issues I don't want this to go in.

jlowe:
> Until I know if this is a key of a value that is causing issues I don't want this to go in.

This is caused by Spark's code generation blindly assuming it's OK to get a value, even if it's a null entry, as long as one doesn't examine it.  Details are in #3996 where Spark generated code like this:
```
        value_MapObject_lambda_variable_1 = (int) (value_1.getInt(loopIndex_0));
        isNull_MapObject_lambda_variable_1 = value_1.isNullAt(loopIndex_0);

        boolean isNull_2 = true;
        int value_2 = -1;

        if (!isNull_MapObject_lambda_variable_1) {


          isNull_2 = false; // resultCode could change nullability.

          value_2 = value_MapObject_lambda_variable_1 + 1;


        }
        if (isNull_2) {
          collectionBuilder_0.$plus$eq(null);
        } else {
          collectionBuilder_0.$plus$eq(value_2);
        }
```

Note that it calls `getInt` before calling `isNullAt` for the same row, and then only later checks for a valid value before wielding the value returned by `getInt`.

@wbo4958 can you elaborate a bit more in the PR description so it's clearer in the future if someone goes back to look on this PR why this change was made?

revans2 (changes requested):
Okay I understand better what is happening it is not for a MapType it is for MapObjects. But now we have two problems.  

The first one is that this patch is going to have a real and noticeable impact to performance for all of the other use cases that don't currently have any problems.  Putting in a conditional check in the data path for each item that is accessed is going to be really slow. Especially if it is just to skip assertions that will be off in production.

The second problem is that now we will not have any way to detect if we have bugs in our code, besides data corruption.  There are multiple times when this has saved me in that we had nulls in a place where Spark didn't expect them and we were able to see it very quickly.  

Because this is a bug in Spark, I want to be 100% sure that we work towards fixing the issue in Spark.  But I also want to see if there is a way we can take the performance hit only for this one use case, and not for all of the other ones. Perhaps we can wrap the columns in something that will do the null checks. Even better if we can find a way to replace it only if assertions are enabled, because in the other cases it will just return the uninitialized data.

wbo4958:
Hi @revans2 @jlowe @firestarman, Thx for reviewing. Yeah, since the assertion in java will not be enabled in production version, then there should not be a problem. The only concern is in unit tests/integration tests, since the assertion will be enabled by default. Just like Bobby Evans said, we may need to do a wrapper for the null check, @firestarman, Could you follow this. I will close this PR

For the long term, we should fix this in SPARK

firestarman:
I have a fix here https://github.com/NVIDIA/spark-rapids/pull/3995 for my case.

We can file an issue to Spark for this probably incorrect calling order in code gen.

PR #4007, opened 2021-11-02
Implement duplicate key handling for GpuCreateMap

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/3250

Depends on https://github.com/rapidsai/cudf/pull/9553

```
scala> val df = Seq(1,2,3).toDF("c0")
df: org.apache.spark.sql.DataFrame = [c0: int]

scala> df.repartition(2).createOrReplaceTempView("foo")

scala> spark.sql("SELECT map(c0, 1, 1 + c0 - 1, 2) FROM foo").collect

21/11/02 20:46:05 ERROR Executor: Exception in task 0.0 in stage 2.0 (TID 3)
java.lang.RuntimeException: Duplicate map key was found, please check the input data. If you want to remove the duplicated keys, you can set spark.sql.mapKeyDedupPolicy to LAST_WIN so that the key inserted at last takes precedence.
	at com.nvidia.spark.rapids.GpuMapUtils$.duplicateMapKeyFoundError(GpuMapUtils.scala:134)
	at org.apache.spark.sql.rapids.GpuCreateMap.$anonfun$columnarEval$10(complexTypeCreator.scala:118)
	
scala> spark.conf.set("spark.sql.mapKeyDedupPolicy", "LAST_WIN")

scala> spark.sql("SELECT map(c0, 1, 1 + c0 - 1, 2) FROM foo").collect

res3: Array[org.apache.spark.sql.Row] = Array([Map(1 -> 2)], [Map(2 -> 2)], [Map(3 -> 2)])
```


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -106,7 +106,23 @@ case class GpuCreateMap(children: Seq[Expression], useStringTypeWhenEmpty: Boole
...
+              GpuColumnVector.from(deduped.incRefCount(), dataType)
+            } else {
+              withResource(deduped.getChildColumnView(0)) { a =>
+                withResource(listOfStruct.getChildColumnView(0)) { b =>
+                  if (a.getRowCount != b.getRowCount) {
+                    throw GpuMapUtils.duplicateMapKeyFoundError
+                  }
+                }
+              }
+              GpuColumnVector.from(deduped.incRefCount(), dataType)
```
ttnghia, line 122: So are we going to have the same output expr for both if/else cases?

andygrove: Yes. The `MAP_KEY_DEDUP_POLICY` config is either `LAST_WIN` or `EXCEPTION`. For ` LAST_WIN` we return the deduped data. For `EXCEPTION` we throw an exception if there are duplicates but if there are no duplicates we return the original map, which is the same as the deduped map, so we can return either.

andygrove:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -106,7 +106,23 @@ case class GpuCreateMap(children: Seq[Expression], useStringTypeWhenEmpty: Boole
       val structs = Range(0, columns.length, 2)
         .safeMap(i => ColumnVector.makeStruct(columns(i), columns(i + 1)))
       withResource(structs) { _ =>
-        GpuColumnVector.from(ColumnVector.makeList(numRows, DType.STRUCT, structs: _*), dataType)
+        withResource(ColumnVector.makeList(numRows, DType.STRUCT, structs: _*)) { listOfStruct =>
+          withResource(listOfStruct.dropListDuplicatesWithKeysValues()) { deduped =>
+            if (SQLConf.get.getConf(SQLConf.MAP_KEY_DEDUP_POLICY) ==
+                SQLConf.MapKeyDedupPolicy.EXCEPTION.toString) {
```
jlowe, line 112: Nit: Rather than perform a conf lookup for each columnar batch, this should either be cached as a lazy val of this class or looked up once on the driver and passed in as a class parameter.

jlowe (dismissed):

jlowe:
build

andygrove:
build

integration_tests/src/main/python/map_test.py
```diff
@@ -96,27 +96,20 @@ def test_map_expr_multiple_pairs():
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : gen_df(spark, data_gen).selectExpr(
                 'map("key1", b, "key2", a) as m1',
-                'map(a, b, b, a) as m2'),
-                conf={'spark.rapids.sql.createMap.enabled':True})
+                'map(a, b, b, a) as m2'))
```
revans2, line 100: nit: Can we add a very small test with duplicate guaranteed keys so we can verify that both are throwing exceptions properly?

revans2 (approved):

revans2 (approved):

revans2:
build

PR #4010, opened 2021-11-02
Enable MapType for ParquetCachedBatchSerializer

### LINKED ISSUE #4009: [FEA] It will be nice for ParquetCachedBatchSerializer to support MapType
Currently the Maps aren't cached using PCBS. We can do it without much effort as cudf ParquetWriter supports Structs and Lists

razajafri:
This PR adds MapType support for PCBS. 

* Added test to generate a df containing map
* Merged the array test (which had static data) with the map test 
* Enabled the MapType for ExprCheck and added support for it in PCBS class.

fixes #4009

### DIFF
--- a/shims/spark311cdh/src/main/scala/com/nvidia/spark/rapids/shims/spark311cdh/Spark311CDHShims.scala
+++ b/shims/spark311cdh/src/main/scala/com/nvidia/spark/rapids/shims/spark311cdh/Spark311CDHShims.scala
@@ -41,7 +41,7 @@ class Spark311CDHShims extends SparkBaseShims {
       GpuOverrides.exec[InMemoryTableScanExec](
         "Implementation of InMemoryTableScanExec to use GPU accelerated Caching",
         ExecChecks((TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 + TypeSig.STRUCT
-            + TypeSig.ARRAY).nested().withPsNote(TypeEnum.DECIMAL,
+            + TypeSig.ARRAY + TypeSig.MAP).nested().withPsNote(TypeEnum.DECIMAL,
               "Negative scales aren't supported at the moment even with " +
                   "spark.sql.legacy.allowNegativeScaleOfDecimal set to true. " +
                   "This is because Parquet doesn't support negative scale for decimal values"),
--- a/integration_tests/src/main/python/cache_test.py
+++ b/integration_tests/src/main/python/cache_test.py
@@ -54,7 +54,7 @@ double_special_cases = [
 all_gen = [StringGen(), ByteGen(), ShortGen(), IntegerGen(), LongGen(),
            pytest.param(FloatGen(special_cases=[FLOAT_MIN, FLOAT_MAX, 0.0, 1.0, -1.0]), marks=[incompat]),
            pytest.param(DoubleGen(special_cases=double_special_cases), marks=[incompat]),
-           BooleanGen(), DateGen(), TimestampGen()] + decimal_gens
+           BooleanGen(), DateGen(), TimestampGen()] + decimal_gens + all_basic_map_gens
 
 @pytest.mark.parametrize('data_gen', all_gen, ids=idfn)
 @pytest.mark.parametrize('join_type', ['Left', 'Right', 'Inner', 'LeftSemi', 'LeftAnti'], ids=idfn)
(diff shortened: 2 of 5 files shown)


razajafri:
build

shims/spark311cdh/src/main/scala/com/nvidia/spark/rapids/shims/spark311cdh/Spark311CDHShims.scala
```diff
@@ -41,7 +41,7 @@ class Spark311CDHShims extends SparkBaseShims {
       GpuOverrides.exec[InMemoryTableScanExec](
         "Implementation of InMemoryTableScanExec to use GPU accelerated Caching",
         ExecChecks((TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 + TypeSig.STRUCT
-            + TypeSig.ARRAY).nested().withPsNote(TypeEnum.DECIMAL,
+            + TypeSig.ARRAY + TypeSig.MAP).nested().withPsNote(TypeEnum.DECIMAL,
```
jlowe, line 44: Curious why there's an override for `InMemoryTableScanExec` in the CDH shim yet it's also overridden in the v2-style shims in sql-plugin.  CDH build should be using 311until320-nondb and therefore should already be picking up a mapping for this override.

jlowe (commented):
Missing corresponding changes in sql-plugin/src/main/320+/scala/com/nvidia/spark/rapids/shims/v2/Spark32XShims.scala?

shims/spark311cdh/src/main/scala/com/nvidia/spark/rapids/shims/spark311cdh/Spark311CDHShims.scala
razajafri: I think that makes sense, we shouldn't have the override for `InMemoryTableScanExec`. @gerashegalov are we reading this correctly? Spark311CDHShims extends SparkBaseShims which already has the `InMemoryTableScanExec` override.

razajafri:
> Missing corresponding changes in sql-plugin/src/main/320+/scala/com/nvidia/spark/rapids/shims/v2/Spark32XShims.scala?

Good catch. My replace script didn't make the change because of the `+` sign in the path.

shims/spark311cdh/src/main/scala/com/nvidia/spark/rapids/shims/spark311cdh/Spark311CDHShims.scala
gerashegalov: override only if necessary for CDH. there is significant code  duplication for PCBS.  Feel also free to introduce a finer-grained rule overrides so you don't have to deal with all execs at the same time https://github.com/NVIDIA/spark-rapids/commit/1441f308821ae7abf0d94589bd3c51c28e9cdac3#diff-8926565c5f6bc26203bed459ac37ea0fba15f9f5488aa2412aeac26b31f80ea6R3601

razajafri: Does the CI test the CDH shim?

jlowe: There currently isn't nightly testing of the CDH shim on our part (I believe Cloudera does this internally).  The CDH shim is built as part of the premerge.

The override in the CDH shim is almost the same source code as in the 311until320-nondb codepath except for the following diffs:
- CDH shim has a type checks and psnote about negative decimal scales which is missing in the other (oversight?)
- CDH shim has code to check serializer class if it's not the right instance and throws an illegal state exception

All of those diffs seem like they should be made consistent one way or the other and don't appear to be specific to CDH.  Therefore I don't see why the override in the CDH shim is needed.

razajafri: Yes, the differences are an oversight. Let me see what I can about the code duplication.

jlowe (approved):

jlowe:
build

razajafri:
build

jlowe (approved):

Issue #4023, opened 2021-11-03
[FEA] Support running on MIG on Spark on YARN


tgravescs:
Allow the plugin to run using MIG on Spark on YARN.

This will likely depend on changes to YARN to support properly, but filing this for changes in plugin and documentation changes.

tgravescs:
Note Spark support for remapping yarn.io/gpu to any resource type: https://issues.apache.org/jira/browse/SPARK-37208

tgravescs:
YARN 3.3.0+ device plugin to support MIG gpus: https://github.com/NVIDIA/spark-rapids-examples/pull/48

tgravescs:
YARN 3.1.2 thru 3.3.0 pr: https://github.com/NVIDIA/spark-rapids-examples/pull/51

PR #4024, opened 2021-11-03
Allow using a custom Spark Resource Name for a GPU

tgravescs:
This covers the case if someone uses a custom resource name with Spark that is actually a GPU it lets the plugin discover that resource.

For instance instead of using the standard gpu scheduling config: spark.executor.resource.gpu.amount=1, instead they make a custom GPU resource named nvidia/gpu and specify that like spark.executor.resource.nvidia/gpu.amount=1.

Our device plugin specifically was expecting the spark resource to be called "gpu". This adds a config to allow it to be changed.

Specifically adding this to later be used with MIG support for Spark on YARN. relates to https://github.com/NVIDIA/spark-rapids/issues/4023
More documentation in the MIG context will come later.


tgravescs:
build

jlowe (approved):

sql-plugin/src/test/scala/com/nvidia/spark/rapids/GpuDeviceManagerSuite.scala
```diff
@@ -0,0 +1,64 @@
...
+
+class GpuDeviceManagerSuite extends FunSuite {
+
+  test("Test Spark gpu resource") {
+    val sparkConf = new SparkConf()
+    val conf = new RapidsConf(sparkConf)
+    val gpu = new ResourceInformation("gpu", Array("3"))
+    val resources = Map("gpu" -> gpu)
+    val gpuAddr = GpuDeviceManager.getGPUAddrFromResources(resources, conf)
+    assert(gpuAddr.nonEmpty)
```
gerashegalov, line 32: nit: could use https://www.scalatest.org/user_guide/using_OptionValues

gerashegalov (approved):

Issue #3932, opened 2021-10-27
[BUG] Aggregator jars are in the test classpath


jlowe:
Originally the tests and integration tests were separated out from the sql-plugin project to ensure we were testing with the distribution jar and the shading that it was performing.  That way we were testing with the same jar that users would use at runtime.

After the new v2 shims were introduced in 21.10, the aggregator jars now appear in the test and integration tests classpath which defeats the purpose of only having the distribution jar in the classpath.  Effectively our test environment no longer closely reflects the environment that will be seen in real use cases.

The test code is referencing classes in the parallel-world setup which cannot be directly accessed via the dist jar, as those classes requires the application classloader which is unavailable at compile time.  We need to find a way to compile the test code against these classes but run them solely against the distribution jar.

Fixing this will involve the completion of multiple tasks:

- [ ] Update CONTRIBUTING.md to describe where unit tests should be placed vs. integration tests
- [ ] Migrate unit tests under the top-level tests module to sql-plugin/src/tests
- [ ] Migrate integration-style tests under the top-level tests module to integration-tests/
- [ ] Remove the top-level tests/ module after all tests have been migrated to sql-plugin/tests or integration-tests/
- [ ] Update integration-tests jar dependencies to not require aggregator jars, relates to #3934

res-life:
@jlowe  For this issue, the integration tests module should depend on dist.jar, but classes in dist.jar are in sub-directories like common3xx, spark301, spark3xx. I write a simple project to depend on dist artifact, actually can't import the classes.  The dist is not a valid jar, the package and the path of a class should be the same, but there are extra root directories (common3xx, spark301, spark3xx...). After all, we use the idea IDE to develop. Even though we find a way to run the tests, the idea IDE can't recognize the classes in the dist jar. 

Runtime is: java -classpath <spark-home>/jars/* : spark-plugin.jar
The jars in spark have high priority. It's better to use the same jars in the plugin as spark does. 

The root cause we separated the tests from sql-plugin project originly is the jars conflict.
If we carefully tackle the dependencies, the integration python tests and the Scala tests will use some jars.  And I think it's easy to use the jars as Spark does, just delete the dependencies that Spark shipped.

I suggest the following solution:
  sql-plugin depends on spark-sql, remove the orc, flatbuffer, protobuffer, because spark-sql already shipped. Add shims for different conflicting jars.
  Move the tests in the tests module to integration-tests or sql-plugin; Remove tests module.
[...]

PR #4030, opened 2021-11-04
Support code coverage report with single version jar [skip ci]

zhanga5:
and re-enable code coverage report from premerge job

Test passed on test PR https://github.com/pxLi/spark-rapids/pull/226

### DIFF
--- a/jenkins/spark-premerge-build.sh
+++ b/jenkins/spark-premerge-build.sh
@@ -55,6 +55,16 @@ mvn_verify() {
     mvn -B $MVN_URM_MIRROR '-Psnapshots,pre-merge' clean verify -Dpytest.TEST_TAGS="premerge_ci_1" \
         -Dpytest.TEST_TYPE="pre-commit" -Dpytest.TEST_PARALLEL=4 -Dcuda.version=$CUDA_CLASSIFIER
 
+    # The jacoco coverage should have been collected, but because of how the shade plugin
+    # works and jacoco we need to clean some things up so jacoco will only report for the
+    # things we care about
+    mkdir -p target/jacoco_classes/
+    FILE=$(ls dist/target/rapids-4-spark_2.12-*.jar | grep -v test | xargs readlink -f)
+    pushd target/jacoco_classes/
+    jar xf $FILE com org rapids spark3xx-common "spark${JACOCO_SPARK_VER:-301}/"
+    rm -rf com/nvidia/shaded/ org/openucx/
+    popd
+
     # Triggering here until we change the jenkins file
     rapids_shuffle_smoke_test
 }
--- a/jenkins/spark-premerge-build.sh
+++ b/jenkins/spark-premerge-build.sh
@@ -58,10 +58,16 @@ mvn_verify() {
     # The jacoco coverage should have been collected, but because of how the shade plugin
     # works and jacoco we need to clean some things up so jacoco will only report for the
     # things we care about
+    SPK_VER=${JACOCO_SPARK_VER:-"301"}
     mkdir -p target/jacoco_classes/
     FILE=$(ls dist/target/rapids-4-spark_2.12-*.jar | grep -v test | xargs readlink -f)
+    UDF_JAR=$(ls ./udf-compiler/target/spark${SPK_VER}/rapids-4-spark-udf_2.12-*-spark${SPK_VER}.jar | grep -v test | xargs readlink -f)
     pushd target/jacoco_classes/
     jar xf $FILE com org rapids spark3xx-common "spark${JACOCO_SPARK_VER:-301}/"
+    # extract the .class files in udf jar and replace the existing ones in spark3xx-ommon and spark$SPK_VER
+    # because the class files in udf jar will be modified in aggregator's shade phase
+    jar xf "$UDF_JAR" com/nvidia/spark/udf
+    rm -rf com/nvidia/shaded/ org/openucx/ spark3xx-common/com/nvidia/spark/udf/ spark${SPK_VER}/com/nvidia/spark/udf/
     rm -rf com/nvidia/shaded/ org/openucx/
     popd
 
--- a/jenkins/spark-premerge-build.sh
+++ b/jenkins/spark-premerge-build.sh
@@ -68,7 +68,6 @@ mvn_verify() {
     # because the class files in udf jar will be modified in aggregator's shade phase
     jar xf "$UDF_JAR" com/nvidia/spark/udf
     rm -rf com/nvidia/shaded/ org/openucx/ spark3xx-common/com/nvidia/spark/udf/ spark${SPK_VER}/com/nvidia/spark/udf/
-    rm -rf com/nvidia/shaded/ org/openucx/
     popd
 
     # Triggering here until we change the jenkins file
--- a/build/coverage-report
+++ b/build/coverage-report
@@ -23,7 +23,9 @@ TMP_CLASS=${TEMP_CLASS_LOC:-"./target/jacoco_classes/"}
 HTML_LOC=${HTML_LOCATION:="./target/jacoco-report/"}
 XML_LOC=${XML_LOCATION:="${HTML_LOC}"}
 DIST_JAR=${RAPIDS_DIST_JAR:-$(ls ./dist/target/rapids-4-spark_2.12-*.jar | grep -v test | xargs readlink -f)}
-SOURCE_DIRS=${SOURCE_DIRS:-"./sql-plugin/src/main/scala/:./sql-plugin/src/main/java/:./shuffle-plugin/src/main/scala/"}
+SPK_VER=${JACOCO_SPARK_VER:-"301"}
+UDF_JAR=${RAPIDS_UDF_JAR:-$(ls ./udf-compiler/target/spark${SPK_VER}/rapids-4-spark-udf_2.12-*-SNAPSHOT-spark${SPK_VER}.jar | grep -v test | xargs readlink -f)}
+SOURCE_DIRS=${SOURCE_DIRS:-"./sql-plugin/src/main/scala/:./sql-plugin/src/main/java/:./shuffle-plugin/src/main/scala/:./udf-compiler/src/main/scala/"}
 
 SOURCE_WITH_ARGS="--sourcefiles "$(echo $SOURCE_DIRS | sed -e 's/:/ --sourcefiles /g')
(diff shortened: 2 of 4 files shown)


jenkins/spark-premerge-build.sh
```diff
@@ -55,6 +55,16 @@ mvn_verify() {
     mvn -B $MVN_URM_MIRROR '-Psnapshots,pre-merge' clean verify -Dpytest.TEST_TAGS="premerge_ci_1" \
         -Dpytest.TEST_TYPE="pre-commit" -Dpytest.TEST_PARALLEL=4 -Dcuda.version=$CUDA_CLASSIFIER
 
+    # The jacoco coverage should have been collected, but because of how the shade plugin
+    # works and jacoco we need to clean some things up so jacoco will only report for the
+    # things we care about
+    mkdir -p target/jacoco_classes/
+    FILE=$(ls dist/target/rapids-4-spark_2.12-*.jar | grep -v test | xargs readlink -f)
+    pushd target/jacoco_classes/
+    jar xf $FILE com org rapids spark3xx-common "spark${JACOCO_SPARK_VER:-301}/"
```
tgravescs, line 64: I don't know enough about jacoco, but with the layout being different and code being under spark3xx-common, jacoco is still able to understand that?

zhanga5: right, there're some warnings like following
```
[WARN] Execution data for class org/apache/spark/sql/rapids/GpuBitwiseNot does not match.
[WARN] Execution data for class com/nvidia/spark/rapids/shims/v2/ShimDataSourceRDD does not match.
[WARN] Execution data for class org/apache/spark/sql/rapids/GpuStringRepeat does not match.
...
```
Main report was still be generated properly for `com.nvidia.spark.rapids`, `org.apache.spark.sql.rapids.execution`, `com.nvidia.spark.rapids.shuffle` etc. Please refer to sample coverage report http://nv/2lb (need vpn) for more details. Any suggestions to fix above warnings are welcome

revans2: "Execution data for class ..." typically means that the checksum for the .class that the data was collected for didn't match the .class provided when generating the report.  It would be nice for us to dig in a bit and understand if we have an actual mismatch, or if something else is happening, like the path is getting included in that checksum somehow.

zhanga5: more testing found only `udf` related class reported mismatch. All others can be resolved with clean environment build (i.e, `spark-premerge-build.sh mvn_verify`). Is there any special handling for `udf`?

New coverage report is updated from http://nv/2lb (need vpn)

```
...
[INFO] Loading execution data file /home/alexzhang/dev/github.com/nvidia/spark-rapids/./udf-compiler/target/spark301/jacoco.exec.
[INFO] Loading execution data file /home/alexzhang/dev/github.com/nvidia/spark-rapids/./sql-plugin/target/spark301/jacoco.exec.
[INFO] Loading execution data file /home/alexzhang/dev/github.com/nvidia/spark-rapids/./integration_tests/target/jacoco.exec.
[INFO] Loading execution data file /home/alexzhang/dev/github.com/nvidia/spark-rapids/./tests/target/spark301/jacoco.exec.
[INFO] Loading execution data file /home/alexzhang/dev/github.com/nvidia/spark-rapids/./shims/target/spark301/jacoco.exec.
[INFO] Loading execution data file /home/alexzhang/dev/github.com/nvidia/spark-rapids/./shims/spark301/target/jacoco.exec.
[WARN] Some classes do not match with execution data.
[WARN] For report generation the same class files must be used as at runtime.
[WARN] Execution data for class com/nvidia/spark/udf/Instruction does not match.
[WARN] Execution data for class com/nvidia/spark/udf/Instruction$ does not match.
[WARN] Execution data for class com/nvidia/spark/udf/GpuScalaUDFLogical does not match.
[WARN] Execution data for class com/nvidia/spark/udf/State$ does not match.
[WARN] Execution data for class com/nvidia/spark/udf/CFG$ does not match.
[WARN] Execution data for class com/nvidia/spark/udf/Repr$LocalDateTime$ does not match.
[WARN] Execution data for class com/nvidia/spark/udf/Repr$DateTimeFormatter does not match.
[WARN] Execution data for class com/nvidia/spark/udf/Repr$StringBuilder$ does not match.
[WARN] Execution data for class com/nvidia/spark/udf/Repr$DateTimeFormatter$ does not match.
[WARN] Execution data for class com/nvidia/spark/udf/CatalystExpressionBuilder$ does not match.
[WARN] Execution data for class com/nvidia/spark/udf/CatalystExpressionBuilder does not match.
[WARN] Execution data for class com/nvidia/spark/udf/BB does not match.
[WARN] Execution data for class com/nvidia/spark/udf/CFG does not match.
[WARN] Execution data for class com/nvidia/spark/udf/Repr$ArrayBuffer does not match.
[WARN] Execution data for class com/nvidia/spark/udf/Repr$ClassTag does not match.
[WARN] Execution data for class com/nvidia/spark/udf/Repr$StringBuilder does not match.
[WARN] Execution data for class com/nvidia/spark/udf/Repr$ArrayBuffer$ does not match.
[WARN] Execution data for class com/nvidia/spark/udf/LambdaReflection does not match.
[WARN] Execution data for class com/nvidia/spark/udf/State does not match.
[WARN] Execution data for class com/nvidia/spark/udf/LogicalPlanRules$$anonfun$replacePartialFunc$1 does not match.
[WARN] Execution data for class com/nvidia/spark/udf/Repr$LocalDateTime does not match.
[WARN] Execution data for class com/nvidia/spark/udf/LambdaReflection$ does not match.
[INFO] Analyzing 3964 classes.
...
```

tgravescs: do we have any reports from before it was shut off to compare to?  Unfortunately the blossom-jenkins premerge jobs have expired.

zhanga5: these classes can be found from either `spark3xx-common` or `spark301` (but no any collision). I believe it's the issue about checksum mismatch like @revans2 mentioned earlier. Perhaps, udf class was built multiple times and test against one of them but different from dist package?

> do we have any reports from before it was shut off to compare to? Unfortunately the blossom-jenkins premerge jobs have expired.

there's no detail report like previous link provided but there's historical result stored on Codecov. We can see ` Plugin.scala` and ` LogicalPlanRules.scala` had some coverage figures: https://codecov.io/gh/nvidia/spark-rapids/tree/branch-21.12/udf-compiler/src/main/scala/com/nvidia/spark/udf

tgravescs: I'm not sure I follow, if they are in spark3xx-common then the checksum should have been the same across all versions of the class.  If they are in spark301 then it would be only for Apache Spark 301 and shouldn't be used by others.   Like Bobby mentioned its likely the path or something else. so the question I have is where is that path going wrong and why is it different then the other classes

GaryShen2008: I found the .class files in rapids-4-spark-udf-spark301_2.12.jar are not as same as the ones in rapids-4-spark-aggregator-spark301.jar. When I copied the class files from the udf jar, Jacoco doesn't warn again. I haven't figured out what caused it.

zhanga5: anyway, after applied the workaround @GaryShen2008 mentioned at above coverage report can be generated without any warnings. PR is updated and passed testing from test PR https://github.com/pxLi/spark-rapids/pull/226

GaryShen2008: After some investigation, the maven shade plugin modified the class files in rapids-4-spark-udf-spark301_2.12.jar when it created the uber jar of aggregator. We copied the original class files from udf jar to target/jacoco_classes to align with the jacoco.exec.

tgravescs: I'm assuming by modified you mean it relocated some classes.
I guess I'm not sure why it modified UDFs classes and not others, I didn't think there was anything specific to UDFs in the shade plugin configuration.  For instance we shade orc so would have expected classes calling those to be different then the sql-plugin classes.  I'm just wondering if there is something else not setup properly.
But I guess if this works, its ok. would be good if @revans2 takes a look.

GaryShen2008: Here's the comparison of BB.class between aggregator.jar and udf.jar.
Screen Shot 2021-11-30 at 9 05 54 AM

revans2: Now I understand. The tests for the UDF compiler plugin are a run in that maven package, per-shadeing.  But we hand Jacoco the files post-shading. I think the simplest way to fix this would be to move all of the tests for the UDF compiler into the tests maven package.  Then they would run post-shade. Or we can just disable the code coverage for the UDF compiler in the short term.

GaryShen2008: The workaround in this patch is that we changed the build script to copy the class files from the udf jar to jacoco's classes folder. So that the class and the jacoco.exec can be according.

zhanga5:
build

jenkins/spark-premerge-build.sh
revans2: That is fine with me, but I would like to have a follow on issue to move the tests.

revans2 (approved):

tgravescs (approved):

jenkins/spark-premerge-build.sh
GaryShen2008: > 

Hi @revans2 , in https://github.com/NVIDIA/spark-rapids/issues/3932, we plan to remove the submodule tests, as my understanding. Which way should we use?

revans2: @jlowe #3932 is supposed to migrate tests under the sql-plugin sub-project. That is going to mess with code coverage because one will be shaded and the other will not be. So now we need to decide if we need to collect code coverage for unit tests or not. I am inclined to say that we skip it for now unless someone else has a better solution on how to fix this.

jlowe: Agree, let's not worry about coverage for the unshaded code in unit tests for now.  Almost all of our tests are integration tests, and we can tackle coverage for unit tests in a followup.

revans2: @GaryShen2008 then what we have now is fine, and if/when someone starts to complain about the issues described above we can turn off code coverage for the unit tests.

PR #4035, opened 2021-11-04
Hash aggregate fix empty resultExpressions [databricks]

### LINKED ISSUE #4017: [BUG] Why is the hash aggregate not handling empty result expressions
The hash aggregate is currently disabled for cases where there are no `resultExpressions`. This is something that was missed and has been in the hash aggregate code for a long time. 

https://github.com/NVIDIA/spark-rapids/blob/branch-21.12/sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala#L872

Prior issues reference the mortgage test as a potential repro case. I can imagine that without much effort this can be fixed and we can remove this special case.

abellina:
Closes #4017

This fixes a case where the hash aggregate was not handling the case where `resultExpressions` was empty. We can see this case for the `count` action pretty consistently, where spark may perform an agg and not want an aggregation column, or any output, other than the number of rows.

I cleaned up a projection that we used to do for the reduction aggregate with no agg function case, where we added a column of zeros. This column gets removed right after and it is a waste of time.

### DIFF
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -14,7 +14,8 @@
 
 import pytest
 
-from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_and_cpu_are_equal_sql,\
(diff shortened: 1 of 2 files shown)


abellina:
build

revans2 (approved):

PR #4036, opened 2021-11-04
Use libcudf to compute hash join output size

### LINKED ISSUE #2440: [FEA] Use CUDF API for getting join output size
This is a follow on issue to #2433 and depends on rapidsai/cudf#8237.

#2433 is enough to get NDS query 72 to succeed, but it is not nearly as fast as it should/could be. This is to use the CUDF API when it is available instead of getting an OOM error. We should evaluate if we want to remove the pre-check/spilt that happens for joins and just go off of the CUDF API or not.

jlowe:
Fixes #2440.

Removes the hash join code to estimate the join size and replaces it with the cudf output join size APIs. This also removes the OOM catch-and-retry logic since it theoretically should no longer be necessary since we are no longer producing an estimate but instead an exact amount.

This starts with the same approach that was taken in #3288 but refines it in the following ways for inner joins:
- If the build side is larger than the stream side batch then it switches to using the stream side batch as the hash table to compute the join for that batch
- If the stream side batch is empty then it avoids building a hash table at all since the join result is known to be empty


jlowe:
Leaving this in draft until an NDS scale performance test can be run.

jlowe:
build

jlowe:
build

jlowe:
This ended up being slower on some queries because it always builds a hash table, whereas the estimation can be much cheaper when the build-side table is large.  Closing this for now, we can revisit #2440 when the hash joins have conditional expression support via AST, as at that point the estimation error will likely be very significant.

PR #4038, opened 2021-11-05
Change the `catalystConverter` to be a Scala `val`.

firestarman:
Change the `catalystConverter` to be a Scala `val` to avoid evaluating a converter per row. 

This repeating evaluation for the converter is unnecessary, leading to really bad performance. 

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalaUDF.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalaUDF.scala
@@ -120,7 +120,7 @@ abstract class GpuRowBasedScalaUDFBase(
    * converter for typed ScalaUDF only, since its the only case where we know the type tag
    * of the return data type of udf function.
    */
-  private[this] val catalystConverter: Any => Any = outputEncoder.map { enc =>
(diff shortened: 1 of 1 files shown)


firestarman:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuScalaUDF.scala
```diff
@@ -145,7 +145,7 @@ abstract class GpuRowBasedScalaUDFBase(
   /**
    * Spark Scala UDF supports at most 22 parameters.
    */
-  private lazy val wrappedFunc: InternalRow => Any = {
+  private[this] val wrappedFunc: InternalRow => Any = {
```
jlowe, line 148: Normally we would want this to be `lazy` since we don't need the driver to evaluate it and it doesn't need to be serialized and sent over to the executors.  Curious what the performance impact is if this remains `lazy`?

firestarman: No, it does not. I have changed it back to be `lazy`. Keeping `lazy` is fine with me.

I just tried to follow how Spark does for this function, so made this diff here.

firestarman:
build

jlowe (approved):

PR #4043, opened 2021-11-05
Promote cudf as dist direct dependency, mark aggregator provided

### LINKED ISSUE #3935: Revisit dependencies of distribution jar
#3930 removes all dependencies from the distribution jar artifact, but ideally we still want to advertise that this jar needs cudf, spark-sql, spark-hive jars, and any other aggregator jar dependencies that have not been pulled into the dist jar to be provided in the dist jar runtime environment.

gerashegalov:
Closes #3935 

- Manually emulate dependency promotion of cudf dependency by the shade plugin 
- Mark aggregator as provided
- Stop overriding default-jar execution, just bind it to none and introduce a dedicated execution for jar of parallel world directories.
- Add dependency-reduced-pom* to the clean phase.
- Undo buggy default-install execution that was not using dependency-reduced-pom


gerashegalov:
build

tgravescs:
so is this actually the right thing to do?  I thought our old dependency reduced pom didn't have the spark dependencies as provided in it, but it would be good to double check that.   So this changes the scope to be compile?   would be nice to have more description because we aren't removing the dependency from the pom we are changing the dependency scope.

tgravescs:
actually, I think I was wrong, must have been looking at the wrong pom, or dependency reduced from aggregator.

https://repo1.maven.org/maven2/com/nvidia/rapids-4-spark_2.12/21.08.0/rapids-4-spark_2.12-21.08.0.pom

Shows them as provided.

jlowe:
I don't think this is what we want.  If you look at the generated `dependency-reduced-pom.xml`, the dependencies are still effectively gone.  This jar clearly cannot stand on its own and needs other jars at runtime to work, and I think the dependencies in the pom should reflect that.

The dist pom is pulling in the aggregator pom with the understanding that we'll get all of the aggregator dependencies, transitively.  That's exactly what we want during the compile/build phase, but when it comes to advertising the dependencies in the deployed artifact, we need to remove the aggregator jar from the dependency tree underneath us, but also move all of the aggregator jar dependencies up to become dist jar dependencies.  The dist jar contains the aggregator jar contents but not its transitive dependencies, so the dist jar dependencies should look similar to the aggregator jar dependencies.

In the end I would expect the dependency reduced pom to list the Apache Spark jars along with the cudf jar as provided dependencies.

gerashegalov:
> In the end I would expect the dependency reduced pom to list the Apache Spark jars along with the cudf jar as provided dependencies.

@jlowe please update #3935  to reflect that you mean cudf  and other aggregator's immediate child dependencies need to be promoted to the dist jar  as provided. The issue currently talks only about spark-hive and spark-sql. 

Look like we want to just pick any aggregator (say the first one) and transplant its dependencies

jlowe:
> Look like we want to just pick any aggregator (say the first one) and transplant its dependencies

Argh, I don't think we can just grab the aggregator dependencies, since some of those dependencies have been pulled into the dist jar as part of the custom shading (sql-plugin, shuffle-plugin, etc.).  We'd have to filter out the aggregator dependencies we know have been pulled in to the dist jar.

As for listing whatever Spark jars are specified by the first aggregator, is that what we want?  It would be nice to show the matrix of support -- we could put the `<profile>` options in there showing how the dependencies switch by profile, but not sure that really helps for dependency analysis or resolution.  Listing the oldest Spark version for the jars may be the best option if we list the Spark dependency at all.  In that case it will get "interesting" if future Spark versions change their jar composition such that we need more jars or differently named jars.  Anyone else have ideas on how to best present this?

gerashegalov:
I also thought of using profiles

gerashegalov:
I investigated the option `promoteTransitiveDependencies` . 

The only way I was able to make it work is by switching `cudf` (note it is not advertised in our 21.08- poms at all) and `spark` to `compile`-scope dependencies so they don't get filtered out by the shade plugin. Then to make sure that those artifacts don't get included into the uber jar, we can add `aI.rapids:*` and `org.apache.spark:*`.  In this case `dependency-reduced-pom.xml` has them  in the dependency list, however with the  `compile` dependency scope. In summary these are too many changes for the result we can easily hardcode for the moment.

As for advertising the right version of the spark dependencies, I think we can skip adding profiles to this pom.  We publish the parent pom. `-Dbuildver` should work

gerashegalov:
build

jlowe (dismissed):
Looks OK to me, but would be good to check with @NvTimLiu and @pxLi that the CI scripts will be OK with the plugin jar attempting to pull cudf and Spark dependencies when it did not declare them before.

aggregator/pom.xml
```diff
@@ -210,6 +210,31 @@
...
+                <version>3.1.0</version>
+                <executions>
+                    <execution>
+                        <id>clean-reduced-dependency-poms</id>
+                        <phase>clean</phase>
+                        <goals>
+                            <goal>clean</goal>
+                        </goals>
+                        <configuration>
+                            <skip>${skipDrpClean}</skip>
```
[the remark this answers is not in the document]
gerashegalov: I found it useful to wipe out everything but the uniquified  dependency-reduced-poms for diffing.

gerashegalov:
I reproduced the CI failure locally:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:3.0.0-M1:install-file (install-with-dependency-reduced-pom) on project rapids-4-spark_2.12: Execution install-with-dependency-reduced-pom of goal org.apache.maven.plugins:maven-install-plugin:3.0.0-M1:install-file failed: Cannot add two different pieces of metadata for: project com.nvidia:rapids-4-spark-udf_2.12 -> [Help 1]
```
working on the fix.

gerashegalov:
build

README.md
```diff
@@ -60,6 +60,29 @@ The profiling tool generates information which can be used for debugging and pro
...
 and [spark profiling tool documentation](docs/spark-profiling-tool.md)
-for more details on how to use the tools.
\ No newline at end of file
+for more details on how to use the tools.
+
+## Dependency for External Projects
+
+If you need to develop some functionality on top of RAPIDS Accelerator For Apache Spark (we currently
+limit support to GPU-accelerated UDFs) we recommend you declare our distribution artifact
+as a `provided` dependency.
```
jlowe, line 71: Curious why this is necessary?

gerashegalov: > Curious why this is necessary?
marking our artifact as provided is not  necessary, this is why I chose the word "recommend". It's mostly to prevent our jar and its dependency tree from being included into a customer's assembly jar if the customer chooses to use something like the shade plugin to build an uber jar

> Should we be marking our dist artifact's dependencies as provided instead
to me this is like not [having those dependencies at all](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope), and we utilize this maven feature  for the aggregator in this PR. If we declare say cudf  as provided in the dist pom, when the customer project directly depends on dist (even with the compile scope). cudf  would become its transitive dependency were it not declared provided. Maven drops provided transitive dependencies from the dependency tree.

jlowe: > It's mostly to prevent our jar and its dependency tree from being included into a customer's assembly jar if the customer chooses to use something like the shade plugin to build an uber jar

Thanks, that makes sense.  I just wanted to check there wasn't something else going on there.

jlowe: Given the Spark dependency is so complicated to communicate, does it make sense to leave that one off and only list the cudf dependency?  Seems silly to pull in Spark jars as part of pulling in the dist jar as a dependency if the Spark version has not bearing on what the user will use in practice.  I can see the pro of listing it, as then depending just on the dist jar will get you _some_ kind of test environment that should work for your supplied RAPIDS UDF.  The drawback is I could see users needing to go out of their way to exclude the spark dependencies from the dist jar so it doesn't conflict with the Spark version they actually care about when testing.

gerashegalov: > Given the Spark dependency is so complicated to communicate, does it make sense to leave that one off and only list the cudf dependency?  
I don't have a strong opinion on this

> Seems silly to pull in Spark jars as part of pulling in the dist jar as a dependency if the Spark version has not bearing on what the user will use in practice. 
just a reminder that the user can control this with -Dbuildver even though it's not user-friendly as we discussed with @tgravescs offline 

> I can see the pro of listing it, as then depending just on the dist jar will get you _some_ kind of test environment that should work for your supplied RAPIDS UDF.  The drawback is I could see users needing to go out of their way to exclude the spark dependencies from the dist jar so it doesn't conflict with the Spark version they actually care about when testing.

Looking at our udf-examples , and we don't require Spark to compile them necessarily (e.g. Scala UDF). Thus, we should indeed  drop it and leave it up to the customer declare it.

jlowe (approved):

gerashegalov:
build

pxLi (approved):
LGTM.

we will keep an eye on this to see if any CI setup need to be updated

Issue #3797, opened 2021-10-12
[FEA] Make RLike support consistent with Apache Spark


andygrove:
**Is your feature request related to a problem? Please describe.**
PR https://github.com/NVIDIA/spark-rapids/pull/3796 added an initial RLike implementation but there are a number of differences with the CPU implementation as noted in the documentation:

```
## RLike

The GPU implementation of RLike has a number of known issues where behavior is not consistent with Apache Spark and
this expression is disabled by default. It can be enabled setting `spark.rapids.sql.expression.RLike=true`.

A summary of known issues is shown below but this is not intended to be a comprehensive list. We recommend that you
do your own testing to verify whether the GPU implementation of `RLike` is suitable for your use case.

### Multi-line handling

The GPU implementation of RLike supports `^` and `$` to represent the start and end of lines within a string but
Spark uses `^` and `$` to refer to the start and end of the entire string (equivalent to `\A` and `\Z`).

| Pattern | Input  | Spark on CPU | Spark on GPU |
|---------|--------|--------------|--------------|
| `^A`    | `A\nB` | Match        | Match        |
| `A$`    | `A\nB` | No Match     | Match        |
| `^B`    | `A\nB` | No Match     | Match        |
| `B$`    | `A\nB` | Match        | Match        |

### Null character in input

The GPU implementation of RLike will not match anything after a null character within a string.

| Pattern   | Input     | Spark on CPU | Spark on GPU |
|-----------|-----------|--------------|--------------|
| `A`       | `\u0000A` | Match        | No Match     |

### Qualifiers with nothing to repeat

Spark supports qualifiers in cases where there is nothing to repeat. For example, Spark supports `a*+` and this
will match all inputs. The GPU implementation of RLike does not support this syntax and will throw an exception with
the message `nothing to repeat at position 0`.

### Stricter escaping requirements

The GPU implementation of RLike has stricter requirements around escaping special characters in some cases.

| Pattern   | Input  | Spark on CPU | Spark on GPU |
|-----------|--------|--------------|--------------|
| `a[-+]`   | `a-`   | Match        | No Match     |
| `a[\-\+]` | `a-`   | Match        | Match        |

### Emprty Groups

The GPU implementation of RLike does not support empty groups correctly.

| Pattern   | Input  | Spark on CPU | Spark on GPU |
|-----------|--------|--------------|--------------|
| `z()?`    | `a`    | No Match     | Match        |
[...]

PR #4044, opened 2021-11-05
RLike: Fall back to CPU for regex that would produce incorrect results

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/3797

This PR introduces a lightweight regular expression parser that allows us to inspect patterns to determine if they can be supported on GPU or not so that we can fall back to CPU in those cases. In most cases, this is necessary to handle edge cases that would cause cuDF to throw an `invalid regex: nothing to match` exception. Examples include:

- Possessive quantifiers: `a*+`
- Empty groups: `a()?`
- Choice where one side is potentially empty: `^|a` or `a*|b`

There are other cases where Java has support for advanced regex features that are not available in cuDF:

- Complex character class usage such as `[a-d[m-p]]`

There is also the beginning of a transpiler so that we can alter the pattern before passing it to cuDF. So far there is only one trivial example of this and that is escaping `-` if it appears within a character class to represent the character `-` rather than being used to specify a character range, as in `[abc-]`.

This is a large PR with a lot of new functionality and I have been leaning heavily on the fuzzing approach to find differences between CPU and GPU. The fuzz tests are included as part of the new unit test suite.


andygrove:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -767,6 +771,595 @@ class GpuRLikeMeta(
...
+ *
+ * Java regex: https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html
+ * cuDF regex: https://docs.rapids.ai/api/libcudf/stable/md_regex.html
+ *
+ * The following blog posts provide some background on Pratt Parsers and parsing regex.
+ *
+ * - https://journal.stuffwithstuff.com/2011/03/19/pratt-parsers-expression-parsing-made-easy/
+ * - https://matt.might.net/articles/parsing-regex-with-recursive-descent/
+ */
+class RegexParser(pattern: String) {
```
jlowe, line 793: This, along with the support classes, is enough code to warrant moving it to a separate file.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -767,6 +771,595 @@ class GpuRLikeMeta(
...
+    var base = parseBase()
+    while (!eof() && (peek().exists(ch => ch == '*' || ch == '+' || ch == '?')
+        || isValidQuantifierAhead())) {
+
+      if (peek().contains('{')) {
+        consumeExpected('{')
+        base = RegexRepetition(base, parseQuantifierOrLiteralBrace().asInstanceOf[RegexQuantifier])
+      } else {
+        base = RegexRepetition(base, SimpleQuantifier(consume()))
+      }
```
jlowe, line 854: Nit: Suggest factoring out the common code:
```suggestion
      val quantifier = if (peek().contains('{')) {
        consumeExpected('{')
        parseQuantifierOrLiteralBrace().asInstanceOf[RegexQuantifier]\
      } else {
        SimpleQuantifier(consume())
      }
      base = RegexRepetition(base, quantifier)
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -767,6 +771,595 @@ class GpuRLikeMeta(
...
+ *
+ * The following blog posts provide some background on Pratt Parsers and parsing regex.
+ *
+ * - https://journal.stuffwithstuff.com/2011/03/19/pratt-parsers-expression-parsing-made-easy/
+ * - https://matt.might.net/articles/parsing-regex-with-recursive-descent/
+ */
+class RegexParser(pattern: String) {
+
+  /** index of current position within the string being parsed */
+  private var i = 0
```
jlowe, line 796: Nit: Suggest using a slightly more descriptive name for this like `pos`, `charIndex`, `charIdx`, etc. Typically a variable name like `i` is used in a small loop within a method, and it's confusing to see it referenced in a method that doesn't also define it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -767,6 +771,595 @@ class GpuRLikeMeta(
...
+        "Invalid hex digit", Some(i))
+    }
+  }
+
+  private def parseOctalDigit = {
+
+    if (i + 2 < pattern.length
+        && pattern.charAt(i).isDigit
+        && pattern.charAt(i+1).isDigit
+        && pattern.charAt(i+2).isDigit) {
```
jlowe, line 1048: What if the digit is `8` or `9`?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -767,6 +771,595 @@ class GpuRLikeMeta(
...
+      val hex = pattern.substring(i, i + 2)
+      i += 3
+      RegexOctalChar(hex)
```
jlowe, line 1051: It's confusing to see hex being passed to an octal method.

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionParserSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  test("group") {
+      assert(parse("(a)(b)") ===
+        RegexSequence(ListBuffer(
+          RegexGroup(RegexSequence(ListBuffer(RegexChar('a')))),
+          RegexGroup(RegexSequence(ListBuffer(RegexChar('b')))))))
+  }
+
+  test("character class") {
```
jlowe, line 59: Are there more tests planned, such as testing hex digit parsing, octal parsing, etc.?

andygrove: I've added tests for hex and octal. Most of the testing is currently happening via the transpiler suite but I think it would be good to have more pure parser tests so I will keep plugging away at this and expand the tests here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -0,0 +1,641 @@
...
+    // parse the source regular expression
+    val regex = new RegexParser(pattern).parse()
+    // validate that the regex is supported by cuDF
+    validate(regex)
+    // write out to regex string, performing minor transformations
+    // such as adding additional escaping
+    regex.toRegexString
+  }
+
+  private def validate(regex: RegexAST): Unit = {
```
revans2, line 407: nit: Do we want to possibly look at some rules that can re-write some of these? like if we see "()" as the regular expression can we replace it with ".*"?  I honestly don't know if that even would work, because I don't remember what java does in this case. This should probably be follow on work if we do want to look into this, because I don't want to hold this up from going in.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -0,0 +1,641 @@
...
+ */
+class CudfRegexTranspiler {
+
+  val nothingToRepeat = "nothing to repeat"
+
+  def transpile(pattern: String): String = {
+    // parse the source regular expression
+    val regex = new RegexParser(pattern).parse()
+    // validate that the regex is supported by cuDF
+    validate(regex)
```
revans2, line 401: nit: Do we want to try and validate the size/complexity of the regular expression? I don't know exactly what CUDF does to figure out if it needs a small, medium, large, or crazy big stack/memory, but it looks like we could do something similar, and fall back to the CPU if it is too large.  The main reason for this is because we just dropped the default for spark.rapids.memory.gpu.reserve from 1GiB to 256 MiB. The reason we set it at 1GiB was because of hard coded regular expressions that we used. If we are going to fully support arbitrary regular expressions it would be nice to try and tie these two together in some way so we fall back to the CPU if there is not enough reserved memory, or we let users opt into larger regular expressions, but in the instructions we tell them that they need to increase the reserved memory accordingly.

Again this would probably be better as follow on work.

revans2 (commented):
This is really great.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
andygrove: I do think we should do this but I wanted to take this one step at a time and start off with simply falling back to CPU and then follow up with optimizations so that this PR doesn't become overwhelming to review. Ideally, I think we should follow up with one PR per specific optimization, so we can make sure that each one has comprehensive tests.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
andygrove: That sounds like a great idea. I will file a follow-on issue for this.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
revans2: That is fine with me. Being customer driven on what we pull in sounds good. After all most of these are corner cases, hey should be rare, and if we do add a modification step we need a lot of testing to really be sure it is doing the right thing.

revans2 (dismissed):

jlowe (dismissed):

revans2 (approved):

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
andygrove: Filed as https://github.com/NVIDIA/spark-rapids/issues/4061

andygrove:
build failed with

```
Unable to find image '***' locally
docker: Error response from daemon: unauthorized: access to the requested resource is not authorized.
See 'docker run --help'.
exit status 125
Error: Process completed with exit code 255.
```

andygrove:
build

PR #4046, opened 2021-11-05
lower GPU memory reserve to 256MB

### LINKED ISSUE #4045: [BUG] q93 failed in this week's NDS runs
We ran q93 without any setting changes and got an OOM exception, where we haven't in the past:

```
Executor task launch worker for task 123.0 in stage 36.0 (TID 12818) 21/11/04 21:13:46:119 INFO DeviceMemoryEventHandler: Device allocation of 345571976 bytes failed, device store has 0 bytes. Total RMM allocated is 9319438848 bytes.
```

This looks like a fragmented pool, as the relatively small ~350MB allocation should fit if we have 9GB worth of memory in RMM, and our GPU has 40GB of memory total.

That said, in the past this query worked without issues, since we had a maximum config in the RMM pool where it could go and allocate the extra memory required to satisfy the 350MB needed. Since the pool limit changes (i.e. https://github.com/NVIDIA/spark-rapids/pull/4019 and https://github.com/rapidsai/cudf/pull/9583) we now have a single value for the size of the pool, and that is calculated at initialization using `free` instead of `total` GPU memory, and it does not grow. The initial value is further reduced by the "reserve" config, set to 1GB by default.

Since the failure we've run a few scenarios by lowering the `spark.rapids.memory.gpu.reserve` setting to `128MB` and have gotten it to pass (the default is 1GB). But in the past we have had issues with kernel that need ~1GB just for the kernel launch, so it seems that a 128MB default may be too restrictive.

Filing this to track as this is something we would like to clarify at least in docs/configs for the 21.12 release.

rongou:
The current reserve is set to 1GB, this may be too high causing some queries to run out of memory. Lowering it to a more reasonable value. If a user runs out of memory launching large kernels, they can still tweak this parameter.

Fixes #4045 


jlowe:
The reason this was set to a large value originally is because a customer query was casting strings to timestamps, and that's not a rare thing for a query to do.  Back then, the string-to-timestamp casting code involved quite a few complicated regex replace with back references, and that kernel had a large thread stack space requirement.  So large that the query would not run at all unless the reserve was set to 1GB.

I totally get the desire to lower the amount of reserve memory (it's like getting a bigger GPU for free! :smile:), but IMO there's a significant associated risk.  The default reserve setting could end up being too low to the point where it would crash on some relatively common query operations.  That leads to a very poor first impression, as users would not be told when it crashes that they could try tuning this reserve config or, counterintuitively, lower the amount of memory the plugin memory pool is using to fix the out of memory issue.  (This is insufficient memory for the driver to complete a kernel launch rather than insufficient memory for a plugin allocation request.)

The timestamp casting code has changed quite a bit since the reserve was tuned, and there may have been libcudf improvements in the memory utilization of the regex backref kernel.  We might be able to get away with a significantly lower setting.  However I think we should test the proposed setting given that casting strings to timestamps is probably not a rare occurrence in many ETL pipelines.

rongou:
For `q93`, we tried 128MB and 64MB, and both succeeded with 10 runs. Running tests locally on my desktop, it fails with 64MB but passes with 128MB. So I guess we are pretty close to the boundary.

For reference, when I originally worked on the arena allocator, I had to reserve some memory too when the max is not specified, and some testing also led to 64MB: https://github.com/rapidsai/rmm/blob/branch-21.12/include/rmm/mr/device/detail/arena.hpp#L295

I think the reason it worked before was more or less a coincidence, as setting the max pool size to total memory minus a 1GB reserve was not very sensical, but just happened to work.

I'm hoping the changes I'm making to the arena allocator would make it more resistant to memory fragmentation, so maybe we can raise this reserve later.

rongou:
build

rongou:
build

jlowe (approved):

abellina (approved):

PR #4051, opened 2021-11-08
Upgrade ORC version from 1.5.8 to 1.5.10

### LINKED ISSUE #3227: [FEA] Upgrade the ORC version
Our plugin is using ORC `1.5.8` currently, however Spark 3.0 and 3.1 are using ORC `1.5.10` and `1.5.12` separately.
It will be better to upgrade it to at least` 1.5.10`.

res-life:
This fixes #3227


res-life:
build

res-life:
1. Upgrades ORC to 1.5.12
2. The "generate-build-info" execution in integration pom.xml is duplicated with root pom, so delete it.

integration_tests/pom.xml
```diff
@@ -211,32 +211,6 @@
                     <classifier>${spark.version.classifier}</classifier>
                 </configuration>
             </plugin>
-            <plugin>
```
revans2, line 214: @pxLi and @tgravescs git blame indicates that the two of you touched this. It looks like it was added in on purpose so we get version info in both jars. Is that correct?

tgravescs:
so can we have a description on this as to why we are upgrading and why did we chose 1.5.12 when Spark 3.0.1 support 1.5.10?  It probably doesn't matter anymore since we are shading it, but if it doesn't matter why aren't we upgrading to newer?

integration_tests/pom.xml
tgravescs: Yes this is to make sure all jars have version properties which we then use later to make sure all the jars we pull are built from same version of source code, please put it back.

res-life: Reverted.

res-life:
> so can we have a description on this as to why we are upgrading and why did we chose 1.5.12 when Spark 3.0.1 support 1.5.10? It probably doesn't matter anymore since we are shading it, but if it doesn't matter why aren't we upgrading to newer?

The default "buildver" version is 3.0.1, so let's use 1.5.10. The purpose of upgrading ORC I can see is to keep consistent when the "buildver" is 3.0.1.

res-life:
build

res-life:
@jlowe  @revans2  help to review.

jlowe (approved):

tgravescs (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScanBase.scala
```diff
@@ -1224,7 +1224,7 @@ private case class GpuOrcFileFilterHandler(
       val searchArg = readerOpts.getSearchArgument
       if (searchArg != null && orcReader.getRowIndexStride != 0) {
         val sa = new SargApplier(searchArg, orcReader.getRowIndexStride, evolution,
-          orcReader.getWriterVersion, useUTCTimestamp)
+          orcReader.getWriterVersion, useUTCTimestamp, false, false)
```
jlowe, line 1227: Actually on second thought, we shouldn't be passing `false, false` here.  We should be following [what ORC 1.5.10 is doing here](https://github.com/apache/orc/blob/rel/release-1.5.10/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java#L219-L220) which is looking up the parameters from the reader and options.

jlowe (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScanBase.scala
res-life: Updated and added test case.

res-life:
build

jlowe (approved):

PR #4052, opened 2021-11-08
Support null safe host column vector

firestarman:
This is trying to fix https://github.com/NVIDIA/spark-rapids/issues/3942.

Since this issue will not occur and everything will work well as CPU under production envs where the assertion is disabled, it will be better to WAR the assertion error in our unit tests than just falling back to CPU.  

### DIFF
--- a/sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRowBasedScalaUDF.scala
+++ b/sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRowBasedScalaUDF.scala
@@ -22,7 +22,7 @@ import org.apache.spark.sql.catalyst.{CatalystTypeConverters, InternalRow}
 import org.apache.spark.sql.catalyst.encoders.ExpressionEncoder
 import org.apache.spark.sql.catalyst.expressions.{Expression, GenericInternalRow, ScalaUDF}
 import org.apache.spark.sql.rapids.{GpuRowBasedScalaUDFBase, ScalaUDFMetaBase}
-import org.apache.spark.sql.types.DataType
+import org.apache.spark.sql.types.{ArrayType, DataType}
 
 /** Run a row-based UDF in a GPU operation */
 case class GpuRowBasedScalaUDF(
@@ -72,6 +72,17 @@ case class GpuRowBasedScalaUDF(
       (converter, true)
     } else { // use CatalystTypeConverters
       (CatalystTypeConverters.createToScalaConverter(dataType), false)
+    }
+  }
+
+  /**
+   *  Need nulls check when there are array types with nulls in the input.
+   *  This is for `https://github.com/NVIDIA/spark-rapids/issues/3942`.
+   */
+  override val checkNull: Boolean = children.exists { child =>
+    child.dataType match {
+      case ArrayType(_, containsNull) => containsNull
+      case _ => false
     }
   }
 }
--- a/sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRowBasedScalaUDF.scala
+++ b/sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRowBasedScalaUDF.scala
@@ -22,7 +22,7 @@ import org.apache.spark.sql.catalyst.{CatalystTypeConverters, InternalRow}
 import org.apache.spark.sql.catalyst.encoders.ExpressionEncoder
 import org.apache.spark.sql.catalyst.expressions.{Expression, GenericInternalRow, ScalaUDF}
 import org.apache.spark.sql.rapids.{GpuRowBasedScalaUDFBase, ScalaUDFMetaBase}
-import org.apache.spark.sql.types.{ArrayType, DataType}
+import org.apache.spark.sql.types.{ArrayType, DataType, MapType, StructType}
 
 /** Run a row-based UDF in a GPU operation */
 case class GpuRowBasedScalaUDF(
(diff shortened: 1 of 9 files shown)


firestarman:
Made it be a draft for early review.

firestarman:
Seems this does not work, will check more.

sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRowBasedScalaUDF.scala
```diff
@@ -74,6 +74,17 @@ case class GpuRowBasedScalaUDF(
...
     }
   }
+
+  /**
+   *  Need nulls check when there are array types with nulls in the input.
+   *  This is for `https://github.com/NVIDIA/spark-rapids/issues/3942`.
+   */
+  override val checkNull: Boolean = children.exists { child =>
+    child.dataType match {
+      case ArrayType(_, containsNull) => containsNull
```
andygrove, line 84: This looks like it is only checking for a top-level array? Do we need to recurse into other types to check for nested arrays?

firestarman: Nice catch. will update it.

firestarman:
Reopen this since I have made it work now.

firestarman:
build

firestarman:
Still a draft PR for early review.

sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRowBasedScalaUDF.scala
firestarman: Updated

firestarman:
local verification passed, ready for reivew.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuUserDefinedFunction.scala
```diff
@@ -89,11 +100,21 @@ trait GpuRowBasedUserDefinedFunction extends GpuExpression
...
 
   private[this] lazy val inputTypesString = children.map(_.dataType.catalogString).mkString(", ")
   private[this] lazy val outputType = dataType.catalogString
+  /**
+   * FIXME This should be called per JVM (aka process), not sure the `lazy` can help here.
+   * This is for https://github.com/NVIDIA/spark-rapids/issues/3942.
+   * And more details can be found from
+   *     https://github.com/NVIDIA/spark-rapids/pull/3997#issuecomment-957650846
+   */
+  private[this] lazy val nullSafe: Boolean = GpuUserDefinedFunction.isAssertionEnabled && checkNull
```
revans2, line 117: This is actually called per instance of GpuUserDefinedFunction. But that is not a big deal, unless we see it impacting performance. If you really are concerned about it put change it to 

```
private[this] lazy val nullSafe: Boolean = checkNull && GpuUserDefinedFunction.isAssertionEnabled
```

That way if checkNull is false we never even try to check the assertions enabled function.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuUserDefinedFunction.scala
```diff
@@ -75,6 +75,17 @@ object GpuUserDefinedFunction {
   // UDFs can support all types except UDT which does not have a clear columnar representation.
   val udfTypeSig: TypeSig = (TypeSig.commonCudfTypes + TypeSig.DECIMAL_64 + TypeSig.NULL +
       TypeSig.BINARY + TypeSig.CALENDAR + TypeSig.ARRAY + TypeSig.MAP + TypeSig.STRUCT).nested()
+
+  /**
+   * Detect whether the assertion in a JVM where the method is called is enabled.
+   * (Yes, this is an ugly solution, but I fail to figure out a better way.)
+   */
+  def isAssertionEnabled: Boolean = try {
+    assert(false)
```
revans2, line 84: I am not 100% sure how scala asserts translate to java asserts.  Also java asserts can be controlled per package, so this would not detect them if someone just enabled asserts for `ai.rapids.cudf`. I would almost rather just assume assertions are enabled unless we have a better way to detect them.

revans2 (commented):
Overall it looks good. I am not super happy with the assertion detection logic and perhaps we can do without it.  Unless you see a really big performance difference between the two.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuUserDefinedFunction.scala
jlowe: Scala asserts are not translated to Java asserts.  IIUC Scala asserts always fire unless explicitly disabled during compilation.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuUserDefinedFunction.scala
firestarman: Thanks for the suggestion, and updated this according to my latest investigaion.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuUserDefinedFunction.scala
firestarman: Really appreciate all of this. Good to know about Java and Scala assertion. Here we should concern only the Java assertion as you know.

And I just found the API `public boolean desiredAssertionStatus()` in `java.lang.Class` matches our requirements here.

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuUserDefinedFunction.scala
```diff
@@ -75,6 +75,17 @@ object GpuUserDefinedFunction {
...
+  /** Detect whether the Java assertion is going to be enabled for the `clsName`. */
+  def isClassAssertionEnabled(clsName: String): Boolean = try {
+    Class.forName(clsName).desiredAssertionStatus()
+  } catch {
+    case _: ClassNotFoundException => false
+  }
+
+  /** (This will be initialized once per process) */
+  lazy val hostColumnAssertionEnabled: Boolean =
+    isClassAssertionEnabled("ai.rapids.cudf.HostColumnVectorCore")
```
revans2, line 88: `HostColumnVectorCore` is public. Why not just do the following instead?

`classOf[HostColumnVectorCore].desiredAssertionStatus()`

firestarman: Good suggestion, updated

firestarman:
build

firestarman:
@revans2  Hi Bobby, Could you help review it again ? Just a little change in the latest commit to address your last comment.

revans2 (approved):

PR #4070, opened 2021-11-10
Support building from sub directory

### LINKED ISSUE #3461: [BUG] Cannot build project from a sub-directory
**Describe the bug**
I wanted to just run a single test without building anything/etc. The IDE has issues without having to blow away the existing IDE setup for the plugin and start over, so I tried to do this with maven. The typical way to do this in maven is to go to the sub-directory of a project and build/test from there.  This does not work.

```
$ cd tests
$ env -u SPARK_HOME -u SPARK_CONF_DIR mvn clean install -Dpytest.TEST_TAGS="" -Dpytest.TEST_PARALLEL=9 -Dcuda.version=cuda11 -Dbuildver=320 -DwildcardSuites=com.nvidia.spark.rapids.AdaptiveQueryExecSuite
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.408 s
[INFO] Finished at: 2021-09-13T05:54:57-05:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:1.8:run (generate-build-info) on project rapids-4-spark-tests_2.12: An Ant BuildException has occured: exec returned: 127
[ERROR] around Ant part ...<exec output=".../tests/target/extra-resources/rapids4spark-version-info.properties" failonerror="true" executable="bash">... @ 6:161 in .../tests/target/antrun/build-main.xml
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```

I looked in CONTRIBUTING.md and there is nothing in there about this.

There are alternative ways to make this work, hint add `-pl tests`. I would prefer to have us fix this so it actually works, but I am fine with documenting the work around.  And even more so I would love it if we had a good way to make the IDE actually switch back and forth easily because `wildcardSuites` is not perfect.

res-life:
Use directory-maven-plugin to get spark-rapids root path instead of user.dir which is a relative path.
Fixes #3461 

### DIFF
--- a/integration_tests/pom.xml
+++ b/integration_tests/pom.xml
@@ -214,7 +214,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-antrun-plugin</artifactId>
-                <version>1.8</version>
                 <executions>
                     <execution>
                         <id>generate-build-info</id>
@@ -225,7 +224,7 @@
                                 <mkdir dir="${project.build.directory}/extra-resources"/>
                                 <mkdir dir="${project.build.directory}/tmp"/>
                                 <exec executable="bash" failonerror="true" output="${project.build.directory}/extra-resources/rapids4spark-version-info.properties">
-                                    <arg value="${user.dir}/build/build-info"/>
+                                    <arg value="${spark.rapids.source.basedir}/build/build-info"/>
                                     <arg value="${project.version}"/>
                                     <arg value="${cudf.version}"/>
                                 </exec>
--- a/shims/spark301/pom.xml
+++ b/shims/spark301/pom.xml
@@ -48,7 +48,7 @@
                             <target>
                                 <mkdir dir="${project.build.directory}/extra-resources"/>
                                 <exec executable="bash" failonerror="true" output="${project.build.directory}/extra-resources/spark-${spark301.version}-info.properties">
-                                    <arg value="${user.dir}/build/dependency-info.sh"/>
+                                    <arg value="${spark.rapids.source.basedir}/build/dependency-info.sh"/>
                                     <arg value="${cudf.version}"/>
                                     <arg value="${cuda.version}"/>
                                     <arg value="${spark301.version}"/>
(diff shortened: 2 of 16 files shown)


res-life:
build

integration_tests/pom.xml
```diff
@@ -214,7 +214,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-antrun-plugin</artifactId>
-                <version>1.8</version>
```
res-life, line 217: Root pom is 3.0.0, but here it's 1.8, so delete this line.

res-life:
Use directory-maven-plugin to get spark-rapids root path instead of user.dir which is a relative path.
"user.dir" is mvn execution path, when excute mvn command under sub project "user.dir" refers to wrong path.
The expected path is: <spark-rapids-root-code-path>/build/build-info

andygrove (approved):
I tried this out locally and the `build-info` was generated to the correct location. This change is more deterministic than using `user.dir` so this looks good to me. I'm not an expert on the build system so would be good to get another review.

revans2 (approved):
I agree this looks good to me, but I really would like to hear from @gerashegalov about this because he has been doing most of the work on the maven build system lately.

gerashegalov (approved):
LGTM

shims/spark301/pom.xml
```diff
@@ -48,7 +48,7 @@
                             <target>
                                 <mkdir dir="${project.build.directory}/extra-resources"/>
                                 <exec executable="bash" failonerror="true" output="${project.build.directory}/extra-resources/spark-${spark301.version}-info.properties">
```
gerashegalov, line 50: In a follow-on we need to generalize this invocation to use spark.version and move to the shim/pom.xml and delete from child shims.

PR #4071, opened 2021-11-10
Treat scalac warnings as errors

gerashegalov:
- Upgrade to Scala 2.12.15
- Add `-Xfatal-warnings` to scalac params
- Add `nowarn` annotations to existing warnings

Closes #3473


gerashegalov:
build

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -94,7 +94,7 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
     // see https://github.com/rapidsai/cudf/issues/9620
     val pattern = "2$"
     // this matches "2" but not "2\n" on the GPU
-    assertCpuGpuContainsMatches(Seq(pattern), Seq("2", "2\n", "2\r", "\2\r\n"))
+    assertCpuGpuContainsMatches(Seq(pattern), Seq("2", "2\n", "2\r", "\u0002\r\n"))
```
jlowe, line 97: Given the comment, I think the original was a typo and should be:
```suggestion
    assertCpuGpuContainsMatches(Seq(pattern), Seq("2", "2\n", "2\r", "2\r\n"))
```
cc: @andygrove

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -176,6 +176,8 @@ class RegexParser(pattern: String) {
                 case Some(end) =>
                   skip()
                   characterClass.appendRange(start, end)
+                case _ =>
+                  // TODO @andygrove
```
jlowe, line 180: This should throw an exception since that's what it did before.  If it's not supposed to throw then that can be fixed in a followup specific to the regex parser.

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
andygrove: Yes, this is what I had intended.

gerashegalov:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -176,6 +176,8 @@ class RegexParser(pattern: String) {
                 case Some(end) =>
                   skip()
                   characterClass.appendRange(start, end)
+                case _ =>
+                  // TODO https://github.com/NVIDIA/spark-rapids/issues/4075
```
jlowe, line 180: The code used to throw a `MatchError` but now will silently suppress it.  If we're waiting for #4075 to fix this the right way then this should throw `IllegalStateException`, `MatchError`, `AssertionException` or some kind of stand-in exception so we don't silently eat this case with a PR that shouldn't be changing the code semantics.  cc: @andygrove for visibility in case this should simply throw `RegexUnsupportedException` (not sure what the message arg should be in that case).

andygrove: The code is currently matching `Some(']')` or `Some(_)` and is missing `None` which could happen with a malformed regular expression. I would suggest adding this:

```scala
case None =>
  throw new RegexUnsupportedException("unexpected EOF while parsing character range", Some(pos))
```

gerashegalov: I am aware of the semantics, and incorporated the change. The thought was that it won't take long for the follow-on

jlowe: Understood.  I've seen too many followup PRs that don't get addressed as timely a we think they will, and a PR with this title shouldn't be the result of a `git bisect` to figure out what in the world changed to cause some bizarre behavior. :smile:

jlowe (approved):

jlowe:
build

andygrove (dismissed):

gerashegalov:
build

jlowe (dismissed):

gerashegalov:
@andygrove can you take a look at:
```
[ERROR] [Error] /home/gshegalov/gits/NVIDIA/spark-rapids/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala:412: match may not be exhaustive.
It would fail on the following inputs: QuantifierFixedLength(_), QuantifierVariableLength(_, _), SimpleQuantifier(_)
[ERROR] one error found
```

andygrove:
@gerashegalov The pre-merge build is failing due to a non-exhaustive match in branch-21.12. Here is the PR to fix this: https://github.com/NVIDIA/spark-rapids/pull/4083

andygrove:
build

gerashegalov:
build

gerashegalov:
build

jlowe (approved):

PR #4072, opened 2021-11-10
Fix overflow checking on optimized decimal sum

revans2:
I found this yesterday while trying to check a few things with overflow on the decimal 128 branch. It turns out the underflow, and wrapped overflow were not being checked properly. I have fixed it, but the test would take a very long time to run and I am not sure we want to put into into our regular integration tests

```
spark.range(174467442481L).selectExpr("SUM(CAST(99999999 as Decimal(8, 0)))")
```

The changes for `GpuUnscaledValue` are there mostly to get the test to work. It turns out that UnscaledValue is applied after constant folding so if you have a Decimal constant value with precision 8 or lower that you are doing a SUM on  it will not run on the GPU.  It is a bit of a corner case. The changes for `GpuUnscaledValue` were already applied to the decimal128 support branch.


revans2:
build

jlowe (approved):

kuhushukla (approved):

PR #4076, opened 2021-11-10
Sets the GPU device id in the UCX early start thread

abellina:

This sets the GPU device in the thread that initializes UCX. The UCX team found that not setting the device ID here could affect the topology calculations when trying to find the "closest NIC" in a multi-NIC machine.

I need to run NDS with UCX fully, but I did run a few queries with this and I am not finding issues.


abellina:
build

tgravescs (approved):
changes look fine to me

jlowe (approved):

abellina:
I was able to run with NDS. Overall most queries where at 1x or above except for those that have shown variability and we are looking into them. I am not concerned with this change at this rate.

PR #4078, opened 2021-11-10
Allow built batch to spill and release gpu before processing stream side input in SpillableJoinIterator.

jbrennan333:

While analyzing tpc-ds query 23b, I noticed that there was a significant slow-down due to gpu contention during a period in the query where all of the executors were doing shuffled join streams.  With concurrent threads set to 4, you could see that four executors held the gpu during the full shuffled join stream.  I experimented with increasing the number of concurrent threads, and saw significant improvements in the query.

I shared these findings with @abellina and @jlowe, and they pointed out that there is no need for the gpu to be held during the cpu-only (streaming input) side of the shuffled join stream.  Which lead to this change, which improves q23b by > 30% in my testing.

The change is to modify the SplittableJoinIterator to mark the built batch as spillable and release the gpu as soon as we start iterating the stream inputs.  If the built batch is spilled, it can be rebuilt very quickly (according to @jlowe).


jbrennan333:
build

abellina:
Can I ask for a run of NDS with this patch? It would be great to see what other queries your patch speeds up, but also may give us a quick check to see if nothing else slows down badly.

jbrennan333:
@abellina I will run a full set of queries with/without this change.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -179,6 +179,13 @@ abstract class SplittableJoinIterator(
   protected def createGatherer(cb: ColumnarBatch, numJoinRows: Option[Long]): Option[JoinGatherer]
 
   override def hasNextStreamBatch: Boolean = {
+    // make built batch spillable
+    if (isInitialJoin) {
```
jlowe, line 183: This is not be specific to the initial join, as this problem can occur on every stream iterator query.  This is really  all about `stream.hasNext`.  Since we typically manifest our batches (and perform shuffle pulls) during `hasNext`, that's when we want to release the semaphore.  Therefore I would expect it to look more like this:
```
if (pendingSplits.nonEmpty) {
  // there's at least one cached stream-side batch ready to go
  true
} else {
  // Querying the stream side may involve network or other non-GPU computation,
  // so ensure our memory buffers are spillable and release the GPU to allow other
  // threads to use the GPU in the interim. The stream iterator will re-grab the GPU
  // semaphore when it's ready to produce a batch.
  builtBatch.allowSpilling()
  GpuSemaphore.releaseIfNecessary(TaskContext.get())
  stream.hasNext || isInitialJoin
}
```

jbrennan333: Thanks @jlowe! I will make the suggested change.

jbrennan333:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
jbrennan333: I put up this change, but I'm not sure I fully understand why just doing it on the initialJoin isn't sufficient?   It doesn't look like builtBatch changes after we construct this iterator.

jbrennan333:
With this version of the change, I ran into OOM issues in q72, and q95 also ran really long.  Need to discuss offline.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
jlowe: This isn't so much about builtBatch as it is releasing the semaphore.  We need to call `allowSpilling` before the first iteration (or the caller of the iterator needs to do it before passing the lazy spillable to the iterator), so that's why we're doing that here.  After the first iteration, `allowSpilling` should be getting called in `nextCbFromGatherer` as it produces the next output batch.  `allowSpilling` is idempotent, so shouldn't matter that we're doing it here even after the first iteration.

However releasing the semaphore seems key to getting the performance win.  IIUC the performance problem was holding the lock while trying to fetch the next batch from the stream iterator and that fetch may involve network shuffle fetch or other non-GPU activity.  To turn the question around a bit: why would we only want to release the GPU semaphore when fetching the first stream batch instead of consistently doing it before fetching every stream batch?

jbrennan333: My understanding (which may be flawed), is that the issue is with a join where the first half needed the GPU and produced builtBatch, but the second half does not need the GPU.  The case we were seeing was that the GPU semaphore was being held the entire time.   It's not clear to me why we would need to release the semaphore more than once at the beginning of processing the stream side?

jlowe: In general, pulling batches from an input iterator may involve performing non-GPU activities.  This is not a problem for "simple" exec nodes where they immediately process an input batch to produce an output batch.  They're not caching any data locally and don't have code to acquire or release the semaphore.  However if an exec node is more complicated, where it can cache data on the GPU in-between fetching from input iterators, then we need to consider carefully what happens with the semaphore.

Fetching from the build side is not so much a problem because it fetches exactly one batch (due to a GpuCoalesceExec being placed before it in the plan on the build side which guarantees all data will be coalesced into a single batch).  GpuCoalesceExec in that case is responsible for releasing the semaphore in-between fetches as it caches the data, and if you look at that code you'll see that logic.

So that's the build side.  However after pulling from the build side we now have a build-side batch sitting in GPU memory and go to fetch from the stream side.  Fetching from the stream side may involve (and often does in the case of join) fetching from a shuffle iterator.  We don't want to hold the GPU semaphore when fetching from the stream side and it involves doing network, RowToColumnExec (i.e.: executing part of the query on the CPU), or other non-GPU activity.  However we can't just release the semaphore anytime we feel like it, because we have this build-side batch sitting in GPU memory and it's not yet spillable.  That's why we call `allowSpilling` on it before releasing the semaphore, because that allows the data we want to cache to be released on-demand if needed.

> It's not clear to me why we would need to release the semaphore more than once at the beginning of processing the stream side?

That's because the stream side, unlike the build side, is not guaranteed to arrive in a single batch.  It could be many batches and therefore many iterations.  The same concerns we have on the first batch can happen on every batch.  If we only release the semaphore on the first iteration, we only get the benefit on the first stream-side batch.  That's great if there's only one batch, but if there's more than one batch then we will continue to hold the semaphore while potentially processing network traffic or CPU-produced batches on the stream side for subsequent batches.

jbrennan333: Thanks for the explanations @jlowe.  I was looking at q23b, which appears to effectively only have one batch per task, so I wasn't seeing the bigger picture.

jbrennan333:
A bit less than half of the queries are faster, about 15 of them 10% or more.  But about the same number are 10% or more slower, and more than half are slower.  I need to look into this more to see if there is some reasonable heuristic we can use to control when to use this optimization.   Note that q72 in the list below is a place-holder showing no change, because I was unable to run it to completion with this change.

Query times sorted by speedup with more speedup first.

Query | Original | Current Patch | Speedup
-- | -- | -- | --
q23b | 48630 | 34297 | 1.417908272
q19 | 7488 | 5587 | 1.340254161
q23a | 36317 | 28809 | 1.260613003
q24a | 43532 | 34990 | 1.244126893
q93 | 26874 | 22082 | 1.217009329
q24b | 44156 | 36354 | 1.214611872
q33 | 6601 | 5614 | 1.175810474
q14b | 28080 | 23920 | 1.173913043
q13 | 6876 | 6020 | 1.142192691
q61 | 8006 | 7024 | 1.139806378
q50 | 20398 | 17946 | 1.136632119
q41 | 3954 | 3485 | 1.134576758
q64 | 40146 | 35668 | 1.125546709
q10 | 7627 | 6948 | 1.097725964
q78 | 21165 | 19532 | 1.08360639
q39a | 7724 | 7172 | 1.076965979
q14a | 29441 | 27360 | 1.076059942
q21 | 4685 | 4360 | 1.074541284
q29 | 8908 | 8324 | 1.070158578
q75 | 17131 | 16110 | 1.063376785
q44 | 10109 | 9519 | 1.061981301
q8 | 6848 | 6483 | 1.056301095
q20 | 6061 | 5765 | 1.051344319
q83 | 6439 | 6128 | 1.050750653
q68 | 6214 | 5951 | 1.044194253
q80 | 11053 | 10636 | 1.039206469
q71 | 7127 | 6928 | 1.028724018
q38 | 12408 | 12070 | 1.028003314
q26 | 5530 | 5398 | 1.024453501
q94 | 25441 | 24836 | 1.0243598
q40 | 6789 | 6641 | 1.0222858
q39b | 7142 | 6989 | 1.021891544
q15 | 5627 | 5516 | 1.020123278
q87 | 12664 | 12497 | 1.013363207
q17 | 7950 | 7847 | 1.013126035
q1 | 6441 | 6364 | 1.012099309
q98 | 7258 | 7172 | 1.011991076
q76 | 8926 | 8824 | 1.011559383
q9 | 27066 | 26818 | 1.00924752
q60 | 6544 | 6487 | 1.008786804
q31 | 10648 | 10621 | 1.002542134
q11 | 14091 | 14077 | 1.00099453
q18 | 11467 | 11483 | 0.9986066359
q69 | 6516 | 6528 | 0.9981617647
q82 | 7786 | 7819 | 0.9957795114
ss_max | 6693 | 6727 | 0.994945741
q25 | 6765 | 6815 | 0.9926632428
q88 | 21348 | 21545 | 0.9908563472
q22 | 5894 | 5951 | 0.9904217779
q99 | 6329 | 6393 | 0.9899890505
q16 | 41523 | 41947 | 0.9898920066
q28 | 17781 | 17998 | 0.9879431048
q12 | 5711 | 5781 | 0.9878913683
q42 | 4057 | 4111 | 0.9868645099
q30 | 7957 | 8085 | 0.9841682127
ss_maxb | 4547 | 4671 | 0.973453222
q3 | 4683 | 4851 | 0.9653679654
q37 | 6234 | 6458 | 0.9653143388
q53 | 5937 | 6154 | 0.9647383815
q90 | 5394 | 5594 | 0.9642474079
q77 | 15279 | 15866 | 0.9630026472
q57 | 9148 | 9512 | 0.9617325484
q4 | 24184 | 25211 | 0.9592638134
q91 | 5907 | 6161 | 0.9587729265
q35 | 8398 | 8768 | 0.9578010949
q55 | 4445 | 4660 | 0.9538626609
q81 | 7736 | 8152 | 0.948969578
q5 | 9037 | 9533 | 0.9479702087
q2 | 8716 | 9235 | 0.943800758
q65 | 11636 | 12348 | 0.9423388403
q63 | 5489 | 5831 | 0.9413479678
q46 | 6310 | 6707 | 0.9408081109
q72 | 242459 | 258346 | 0.9385049507
q6 | 6766 | 7210 | 0.9384188627
q79 | 6274 | 6688 | 0.9380980861
q43 | 5183 | 5535 | 0.9364046974
q96 | 6165 | 6597 | 0.9345156889
q45 | 6892 | 7383 | 0.9334958689
q86 | 5334 | 5715 | 0.9333333333
q92 | 4717 | 5069 | 0.9305582955
q52 | 3934 | 4260 | 0.9234741784
q34 | 6880 | 7464 | 0.9217577706
q58 | 9985 | 10853 | 0.9200221137
q59 | 8000 | 8703 | 0.9192232563
q27 | 5555 | 6048 | 0.9184854497
q51 | 7815 | 8514 | 0.9178999295
q67 | 45944 | 50219 | 0.9148728569
q48 | 5717 | 6257 | 0.9136966597
q74 | 12135 | 13299 | 0.9124746222
q49 | 9796 | 10832 | 0.9043574594
q89 | 5796 | 6413 | 0.9037891782
q70 | 8353 | 9249 | 0.9031246621
q47 | 8640 | 9626 | 0.8975690837
q56 | 5653 | 6312 | 0.8955956907
q54 | 7979 | 8919 | 0.8946070187
q85 | 6926 | 7813 | 0.8864712658
q62 | 5022 | 5680 | 0.8841549296
q84 | 4120 | 4693 | 0.8779032602
q97 | 6933 | 7926 | 0.8747161241
q32 | 4524 | 5403 | 0.8373126041
q7 | 5010 | 5997 | 0.8354177089
q73 | 6040 | 7297 | 0.8277374263
q36 | 5986 | 7236 | 0.8272526258
q66 | 6655 | 10191 | 0.6530271808
q95 | 40584 | 274730 | 0.1477232192
TOTAL | 1456794 | 1672541 | 0.8710064507

jlowe:
> I need to look into this more to see if there is some reasonable heuristic we can use to control when to use this optimization.

Converting this PR to a draft for the investigation.  Or should we close this and handle any heuristic as a separate PR?

jbrennan333:
I think draft makes sense for now.  Thanks.

jbrennan333:
I did an experiment where I used GPU Free memory as a heuristic to decide when to apply this optimization, to see if it improves the results.  It did not.  I tried setting a threshold of 16GB and even as high as 32GB (only apply optimization if free memory is above the threshold), and there were still more queries that were slower than faster, and some queries (like q72) still ran out of memory.

These were the numbers with the threshold set to 16GB (8 x batch size) - note that q72 is removed from these results because it failed to complete due to OOM:
<google-sheets-html-origin><style type="text/css"></style>

Query | Original | 8x Batch Size | Original / 8x Batch Size
-- | -- | -- | --
q23b | 49237 | 34003 | 1.448019292
q24b | 45299 | 35507 | 1.275776607
q23a | 37175 | 29150 | 1.275300172
q24a | 43021 | 35541 | 1.210461158
q93 | 26763 | 22265 | 1.202021109
q33 | 6527 | 5530 | 1.180289331
q64 | 41910 | 35887 | 1.167832363
q68 | 6481 | 5644 | 1.148299079
q50 | 20558 | 18168 | 1.131549978
q52 | 4538 | 4011 | 1.131388681
q39b | 8023 | 7096 | 1.130636979
q92 | 5355 | 4748 | 1.127843302
q78 | 21655 | 19517 | 1.109545524
q43 | 6140 | 5585 | 1.099373321
q14a | 30165 | 27465 | 1.098306936
q84 | 4465 | 4070 | 1.097051597
q91 | 6156 | 5643 | 1.090909091
q30 | 8402 | 7985 | 1.052222918
q14b | 27375 | 26036 | 1.051428791
q87 | 14788 | 14155 | 1.044719181
q26 | 5328 | 5149 | 1.034764032
ss_maxb | 4672 | 4547 | 1.027490653
q85 | 7130 | 6940 | 1.027377522
ss_max | 6590 | 6423 | 1.026000311
q31 | 10614 | 10348 | 1.02570545
q94 | 25780 | 25146 | 1.025212757
q45 | 7251 | 7142 | 1.015261831
q96 | 6887 | 6786 | 1.014883584
q76 | 9189 | 9082 | 1.011781546
q28 | 29733 | 29437 | 1.010055372
q95 | 40532 | 40152 | 1.009464037
q6 | 6443 | 6387 | 1.00876781
q22 | 6166 | 6133 | 1.005380727
q20 | 6284 | 6257 | 1.004315167
q38 | 13908 | 13863 | 1.003246051
q44 | 10065 | 10046 | 1.0018913
q79 | 6969 | 6961 | 1.00114926
q58 | 9894 | 9905 | 0.9988894498
q9 | 32602 | 32740 | 0.9957849725
q82 | 8626 | 8665 | 0.9954991344
q17 | 7293 | 7351 | 0.992109917
q5 | 9522 | 9600 | 0.991875
q19 | 5773 | 5838 | 0.98886605
q81 | 7584 | 7708 | 0.9839128179
q12 | 5560 | 5651 | 0.9838966555
q88 | 24487 | 24897 | 0.9835321525
q62 | 5763 | 5862 | 0.983111566
q39a | 7208 | 7344 | 0.9814814815
q8 | 6464 | 6587 | 0.9813268559
q35 | 8310 | 8482 | 0.9797217637
q54 | 8088 | 8266 | 0.9784660053
q80 | 10078 | 10325 | 0.9760774818
q75 | 15893 | 16283 | 0.9760486397
q60 | 6106 | 6258 | 0.9757110898
q13 | 6892 | 7068 | 0.9750990379
q29 | 8316 | 8564 | 0.9710415694
q4 | 23092 | 23804 | 0.9700890607
q34 | 6812 | 7051 | 0.9661040987
q66 | 7033 | 7299 | 0.9635566516
q99 | 6690 | 6950 | 0.9625899281
q10 | 7436 | 7727 | 0.9623398473
q40 | 6015 | 6253 | 0.9619382696
q47 | 8911 | 9274 | 0.9608583136
q25 | 6510 | 6804 | 0.9567901235
q18 | 11861 | 12405 | 0.956146715
q11 | 14408 | 15071 | 0.9560082277
q49 | 9148 | 9601 | 0.9528174149
q27 | 5481 | 5754 | 0.9525547445
q37 | 6227 | 6544 | 0.9515586797
q16 | 44237 | 46492 | 0.9514970317
q51 | 8432 | 8872 | 0.950405771
q98 | 7571 | 7967 | 0.9502949667
q89 | 6053 | 6399 | 0.9459290514
q74 | 12164 | 12947 | 0.9395226693
q69 | 6062 | 6496 | 0.9331896552
q21 | 4006 | 4299 | 0.931844615
q36 | 6054 | 6500 | 0.9313846154
q86 | 5603 | 6028 | 0.9294956868
q41 | 3291 | 3546 | 0.9280879865
q55 | 4124 | 4467 | 0.9232146855
q83 | 6095 | 6625 | 0.92
q1 | 6361 | 6927 | 0.9182907464
q53 | 5752 | 6266 | 0.9179699968
q32 | 4666 | 5092 | 0.9163393559
q57 | 8611 | 9413 | 0.9147986827
q2 | 8911 | 9749 | 0.9140424659
q97 | 7645 | 8366 | 0.9138178341
q65 | 10898 | 12045 | 0.904773765
q73 | 6571 | 7267 | 0.9042245769
q56 | 5556 | 6169 | 0.9006321932
q63 | 5821 | 6466 | 0.9002474482
q46 | 6440 | 7178 | 0.8971858456
q77 | 14614 | 16374 | 0.8925125198
q3 | 4934 | 5573 | 0.8853400323
q61 | 7014 | 8079 | 0.8681767545
q59 | 7896 | 9121 | 0.865694551
q42 | 3588 | 4145 | 0.8656212304
q90 | 5199 | 6036 | 0.861332008
q70 | 7939 | 9248 | 0.8584558824
q48 | 6200 | 7313 | 0.8478052783
q7 | 5237 | 6209 | 0.843453052
q71 | 6325 | 7567 | 0.8358662614
q15 | 5554 | 6735 | 0.8246473645
q67 | 40780 | 59269 | 0.6880494019
TOTAL | 1237861 | 1221981 | 1.012995292

jbrennan333:
I am going to close this PR.  We can always open a new PR if we decide to try other approaches.

Issue #4069, opened 2021-11-10
[BUG] parquet_test.py pytests FAILED on Databricks-9.1-ML-spark-3.1.2


NvTimLiu:
**Describe the bug**

```
2021-11-10T06:53:08.173Z] 
[2021-11-10T06:53:08.173Z] =================================== FAILURES ===================================
[2021-11-10T06:53:08.173Z]  test_nested_pruning_and_case_insensitive[true--reader_confs0-[['struct', Struct(['c_1', String],['case_insensitive', Long],['c_3', Short])]]-[['STRUCT', Struct(['case_INSENsitive', Long])]]] _[0m
[2021-11-10T06:53:08.173Z] [gw0] linux -- Python 3.8.12 /databricks/conda/envs/cudf-udf/bin/python
[2021-11-10T06:53:08.173Z] 
[2021-11-10T06:53:08.173Z] spark_tmp_path = '/tmp/pyspark_tests//754491/'
[2021-11-10T06:53:08.173Z] data_gen = [['struct', Struct(['c_1', String],['case_insensitive', Long],['c_3', Short])]]
[2021-11-10T06:53:08.173Z] read_schema = [['STRUCT', Struct(['case_INSENsitive', Long])]]
[2021-11-10T06:53:08.173Z] reader_confs = {'spark.rapids.sql.format.parquet.reader.type': 'PERFILE'}
[2021-11-10T06:53:08.173Z] v1_enabled_list = '', nested_enabled = 'true'
[2021-11-10T06:53:08.173Z] 
[2021-11-10T06:53:08.173Z]     @pytest.mark.parametrize('data_gen,read_schema', _nested_pruning_schemas, ids=idfn)
[2021-11-10T06:53:08.173Z]     @pytest.mark.parametrize('reader_confs', reader_opt_confs)
[2021-11-10T06:53:08.173Z]     @pytest.mark.parametrize('v1_enabled_list', ["", "parquet"])
[2021-11-10T06:53:08.173Z]     @pytest.mark.parametrize('nested_enabled', ["true", "false"])
[2021-11-10T06:53:08.173Z]     def test_nested_pruning_and_case_insensitive(spark_tmp_path, data_gen, read_schema, reader_confs, v1_enabled_list, nested_enabled):
[2021-11-10T06:53:08.173Z]         data_path = spark_tmp_path + '/PARQUET_DATA'
[2021-11-10T06:53:08.173Z]         with_cpu_session(
[...]

PR #4080, opened 2021-11-11
Add case insensitive when clipping parquet blocks [databricks]

wbo4958:
Databricks 9.1 runtime has changed the API clipParquetSchema which returns
the readSchema-same-name schema when case insensitive, which will cause
clipBlocks return in-correct results since clipBlocks only takes care of
case sensitive matching.

To fix https://github.com/NVIDIA/spark-rapids/issues/4069

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
@@ -378,7 +378,8 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
     val clippedSchema = GpuParquetPartitionReaderFactoryBase.filterClippedSchema(clippedSchemaTmp,
       fileSchema, isCaseSensitive)
     val columnPaths = clippedSchema.getPaths.asScala.map(x => ColumnPath.get(x: _*))
-    val clipped = ParquetPartitionReader.clipBlocks(columnPaths, blocks.asScala)
+    val clipped = ParquetPartitionReader.clipBlocks(columnPaths, blocks.asScala, isCaseSensitive)
+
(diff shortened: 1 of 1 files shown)


wbo4958:
build

NvTimLiu:
Verifed PASS on 9.1 Databricks, + V

revans2 (approved):

tgravescs (approved):

jlowe (approved):

jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
```diff
@@ -1545,14 +1546,25 @@ object ParquetPartitionReader {
...
    * @return the updated block metadata with undesired column chunks removed
    */
   private[spark] def clipBlocks(columnPaths: Seq[ColumnPath],
-      blocks: Seq[BlockMetaData]): Seq[BlockMetaData] = {
-    val pathSet = columnPaths.toSet
+      blocks: Seq[BlockMetaData], isCaseSensitive: Boolean): Seq[BlockMetaData] = {
+    val pathSet = if (isCaseSensitive) {
+      columnPaths.map(cp => cp.toDotString).toSet
+    } else {
+      columnPaths.map(cp => cp.toDotString.toLowerCase(Locale.ROOT)).toSet
```
gerashegalov, line 1557: using toUpper/Lower even if you have the right Locale may lead to interesting issues https://news.ycombinator.com/item?id=8892157 . It's safer to delegate this mess to java.lang.String *ignoreCase family

wbo4958: Thx @gerashegalov, since there are already some places to use toLowerCase in the rapids code. we can create a following issue for that.

PR #4081, opened 2021-11-11
Convert String to DecimalType without casting to FloatType [databricks]

### LINKED ISSUE #2019: [BUG] The result of casting string to decimal is not identical to Apache Spark
**Describe the bug**
Currently string to decimal type on the GPU might produce results which slightly differed from the correct results when the string 
represents any number exceeding the max precision that  the operation CAST_STRING_TO_FLOAT can keep. For instance, the GPU returns 99999999999999987 for input string "99999999999999999".

The cause of divergence is that we can not directly cast strings containing scientific notation to decimal via cuDF API. So, we have to cast strings to floats firstly. Then, cast floats to decimals. The first step may lead to precision loss.

Here are samples to produce inconsistent results:
* cast("99999999999999999" to Decimal(18, 0))
    * 99999999999999999(CPU) vs 99999999999999987(GPU)
* cast("999999999999999" to Decimal(18, -2))
    * 999999999999999.00(CPU) vs 999999999999999.10(GPU)

**Expected behavior**
The GPU runtime can produce exact same results as Apache Spark. 

**Additional context**
Current issue is related to issue #1625 and PR #1999.

razajafri:
* Added a `castStringToDecimal` method
* Some refactoring was needed to not sanitize the string `Infinity` 

This doesn't need to go in branch-21.12 but branch-22.02 hasn't been cut yet which is why I have targeted it to branch-21.12

Fixes #2019 
depends on /rapidsai/cudf#9658


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -777,17 +768,65 @@ object GpuCast extends Arm {
...
+      input: ColumnView,
+      ansiEnabled: Boolean,
+      dt: DecimalType): ColumnVector = {
+    // 1. Sanitize strings to make sure all are floating points
+    // 2. Identify all fixed point values
+    // 3. Cast String to newDt (newDt = dt.scale + 1). Promote precision if needed. This step is
+    //    required so we can round up if needed in the final step
+    // 4. Now cast to newDt to dt (Decimal to Decimal)
+    def getInterimDecimalPromoteIfNeeded(dt: DecimalType): DecimalType = {
+      if (dt.scale + 1 > dt.precision) {
```
revans2, line 781: If we are adding 1 to the scale we always have to add one to the precision. Or we are reducing the amount of digits above the decimal point that we can support.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -777,17 +768,65 @@ object GpuCast extends Arm {
...
+        // promote if possible or throw
+        if (dt.precision == Decimal.MAX_LONG_DIGITS) {
+          //We don't support Decimal 128
```
revans2, line 784: This comment does not match the checks being done. MAX_LONG_DIGITS, is the maximum precision that a Decimal64 can support. So something with a precision of MAX_LONG_DIGITS is still allowed as a Decimal64.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -777,17 +768,65 @@ object GpuCast extends Arm {
...
+      dt: DecimalType): ColumnVector = {
+    // 1. Sanitize strings to make sure all are floating points
+    // 2. Identify all fixed point values
+    // 3. Cast String to newDt (newDt = dt.scale + 1). Promote precision if needed. This step is
+    //    required so we can round up if needed in the final step
+    // 4. Now cast to newDt to dt (Decimal to Decimal)
+    def getInterimDecimalPromoteIfNeeded(dt: DecimalType): DecimalType = {
+      if (dt.scale + 1 > dt.precision) {
+        // promote if possible or throw
+        if (dt.precision == Decimal.MAX_LONG_DIGITS) {
```
revans2, line 783: nit: we will support decimal128 soon so it would be good to have a >= check instead to be on the safe side.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -777,17 +768,65 @@ object GpuCast extends Arm {
...
+          //We don't support Decimal 128
+          throw new IllegalArgumentException("One or more values exceed the maximum supported " +
```
revans2, line 785: We need a check to fall back to the CPU in addition to throwing an exception in this case.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -777,17 +768,65 @@ object GpuCast extends Arm {
...
+              "Decimal precision while conversion")
+        }
+        DecimalType(dt.precision + 1, dt.scale + 1)
+      }
+      DecimalType(dt.precision, dt.scale + 1)
+    }
+
+    withResource(input.strip()) { trimmed =>
```
revans2, line 793: `sanitizeStringToFloat` already strips the input.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -777,17 +768,65 @@ object GpuCast extends Arm {
...
+      withResource(GpuCast.sanitizeStringToFloat(trimmed, ansiEnabled, false)) { sanitized =>
```
revans2, line 794: `sanitizeStringToFloat` feels like a lot of overkill in this situation.  Do we really need that full regexp? Will `isFixedPoint` cover us? If not fully we can at least make the regexp a lot simpler, which should make it faster to process.  Also technically I don't think we support the [fFdD] for decimal types?  We should test that though.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -777,17 +768,65 @@ object GpuCast extends Arm {
...
+    // 4. Now cast to newDt to dt (Decimal to Decimal)
+    def getInterimDecimalPromoteIfNeeded(dt: DecimalType): DecimalType = {
+      if (dt.scale + 1 > dt.precision) {
+        // promote if possible or throw
+        if (dt.precision == Decimal.MAX_LONG_DIGITS) {
+          //We don't support Decimal 128
+          throw new IllegalArgumentException("One or more values exceed the maximum supported " +
+              "Decimal precision while conversion")
+        }
+        DecimalType(dt.precision + 1, dt.scale + 1)
```
revans2, line 788: This does not actually set anything.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -777,17 +768,65 @@ object GpuCast extends Arm {
...
+      }
+      DecimalType(dt.precision, dt.scale + 1)
+    }
+
+    withResource(input.strip()) { trimmed =>
+      withResource(GpuCast.sanitizeStringToFloat(trimmed, ansiEnabled, false)) { sanitized =>
+        val interimSparkDt = getInterimDecimalPromoteIfNeeded(dt)
+        val interimDt = DecimalUtil.createCudfDecimal(interimSparkDt)
+        withResource(Scalar.fromNull(interimDt)) { nulls =>
```
revans2, line 797: nit: can we put the nulls closer to where it is used? There are cases where we will never need to create it, so why bother?

razajafri:
@revans2 PTAL

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -777,17 +817,59 @@ object GpuCast extends Arm {
...
+    def getInterimDecimalPromoteIfNeeded(dt: DecimalType): DecimalType = {
+      if (dt.precision + 1 > Decimal.MAX_LONG_DIGITS) {
+        //We don't support Decimal 128
+        throw new IllegalArgumentException("One or more values exceed the maximum supported " +
+            "Decimal precision while conversion")
+      }
+      DecimalType(dt.precision + 1, dt.scale + 1)
+    }
+
+    withResource(GpuCast.sanitizeStringToDecimal(input, ansiEnabled)) { sanitized =>
```
andygrove, line 838: It looks like `sanitized` could be released earlier here, something like:

```scala
val interimSparkDt = getInterimDecimalPromoteIfNeeded(dt)
val interimDt = DecimalUtil.createCudfDecimal(interimSparkDt)

val isFixedPoints = withResource(GpuCast.sanitizeStringToDecimal(input, ansiEnabled)) { 
    sanitized =>
  sanitized.isFixedPoint(interimDt)
}

withResource(isFixedPoints) { _ =>
  if (ansiEnabled) {
    ...
  }
  ...
}
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -162,7 +154,60 @@ object GpuCast extends Arm {
...
+            "([0-9]+)|" +           // digits, OR
+            "([0-9]*\\.[0-9]+)|" +  // decimal with optional leading and mandatory trailing, OR
+            "([0-9]+\\.[0-9]*)" +   // decimal with mandatory leading and optional trailing
+          ")" +
+          "([eE][+\\-]?[0-9]+)?" +  // exponent
+        ")" +
+      ")" +
+      "$"                           // end of line
+
+    withResource(input.strip()) { stripped =>
```
andygrove, line 178: I think `stripped` can be released earlier here:

```scala
val withoutWhitespace = withResource(input.strip()) { stripped =>
  withResource(GpuScalar.from(null, DataTypes.StringType)) { nullString =>
    // filter out strings containing breaking whitespace
    withResource(ColumnVector.fromStrings("\r", "\n")) {
      verticalWhitespace =>
        withResource(stripped.contains(verticalWhitespace)) {
          _.ifElse(nullString, stripped)
        }
    }
  }
}

withResource(withoutWhitespace) { _ =>
  ...
}
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/FloatUtils.scala
```diff
@@ -18,6 +18,8 @@ package com.nvidia.spark.rapids
 
 import ai.rapids.cudf.{ColumnVector, ColumnView, DType, Scalar}
 
+import org.apache.spark.sql.types.DecimalType
```
revans2, line 21: I don't think this is used any more and some code was just merged in to check for unused imports and report an error.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -79,6 +79,9 @@ class CastExprMeta[INPUT <: CastBase](
             "converting floating point data types to strings and this can produce results that " +
             "differ from the default behavior in Spark.  To enable this operation on the GPU, set" +
             s" ${RapidsConf.ENABLE_CAST_FLOAT_TO_STRING} to true.")
+      case (_: StringType, dt: DecimalType) if (dt.precision + 1 > Decimal.MAX_LONG_DIGITS) =>
+        willNotWorkOnGpu(s"Converting to $dt will result in a 128-bit temporary type that is " +
```
revans2, line 83: nit: I would explain something like.

"Because of rounding requirements we cannot support the requested precision for $dt"

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -162,7 +154,60 @@ object GpuCast extends Arm {
 
   val INVALID_FLOAT_CAST_MSG: String = "At least one value is either null or is an invalid number"
 
-  def sanitizeStringToFloat(input: ColumnVector, ansiEnabled: Boolean): ColumnVector = {
+  def sanitizeStringToDecimal(
+      input: ColumnView,
+      ansiEnabled: Boolean): ColumnVector = {
+
+    // This regex gets applied to filter out known edge cases that would result in incorrect
```
revans2, line 161: What are the known edge cases? would be very nice to know what we have to include this as it is expensive. Looking at the code it appears that is_fixed_point cuts off early if it sees something that it does not expect, so it might be nice to have a follow on issue to actually fix that, either in CUDF or in Spark specific code.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -162,7 +154,60 @@ object GpuCast extends Arm {
...
+          ")" +
+          "([eE][+\\-]?[0-9]+)?" +  // exponent
+        ")" +
+      ")" +
+      "$"                           // end of line
+
+    withResource(input.strip()) { stripped =>
+      withResource(GpuScalar.from(null, DataTypes.StringType)) { nullString =>
+        // filter out strings containing breaking whitespace
+        val withoutWhitespace = withResource(ColumnVector.fromStrings("\r", "\n")) {
```
revans2, line 181: So in ANSI mode is this not an error?  Does the regular expression not match this, because it sure looks like the regexp would error out on anything that has any white space in it at all?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -162,7 +154,60 @@ object GpuCast extends Arm {
...
+        }
+        // filter out any strings that are not valid fixed-point numbers according
+        // to the regex pattern
+        withResource(withoutWhitespace) { _ =>
+          withResource(withoutWhitespace.matchesRe(VALID_DEC_REGEX)) { isFixedPoint =>
+            if (ansiEnabled) {
+              withResource(isFixedPoint.all()) { allMatch =>
+                // Check that all non-null values are valid floats.
+                if (allMatch.isValid && !allMatch.getBoolean) {
+                  throw new NumberFormatException(GpuCast.INVALID_FLOAT_CAST_MSG)
```
revans2, line 195: nit: could we rename INVALID_FLOAT_CAST_MSG? as it is not just for floats, and it is generic enough to not be.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
razajafri: This is a very good point. ANSI doesn't like spaces, and throws an ansi exception. I will file an issue for Floats as well.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
razajafri: Where is it cutting off early? 

Are you saying if I pass c = ["", "1.2", "3", ""] and if the boolean vector is initialized to true
d = c.is_fixed_point() = [false, true, true,  true]

basically everything after the first value in `d` is bogus?

revans2: Sorry I was wrong. Reading through the code it looked like the check ignored anything after it saw something it didn't expect, but that is not true.

It looks like "1.5ABC" will result in a false being returned. Which if that is true, then I don't think we need the regular expression check at all any more.  That is what triggered this? Why do we need the regexp. What "edge cases" does it cover that are not covered by the existing type check code?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
razajafri: This is actually an unnecessary check as `\r` is being checked as a string which would be caught by the regex check.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
razajafri: You are right we don't need the regex check anymore as the cudf is reporting everything we need. This is relevant in case of a float because it needs to convert the "infinity" => "inf"

razajafri:
build

razajafri:
build

razajafri:
build

revans2 (approved):

PR #4082, opened 2021-11-11
Add abfs and abfss to the cloud scheme

wbo4958:


wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -740,10 +740,11 @@ object RapidsConf {
 
   val CLOUD_SCHEMES = conf("spark.rapids.cloudSchemes")
     .doc("Comma separated list of additional URI schemes that are to be considered cloud based " +
-      "filesystems. Schemes already included: dbfs, s3, s3a, s3n, wasbs, gs. Cloud based stores " +
-      "generally would be total separate from the executors and likely have a higher I/O read " +
-      "cost. Many times the cloud filesystems also get better throughput when you have multiple " +
-      "readers in parallel. This is used with spark.rapids.sql.format.parquet.reader.type")
+      "filesystems. Schemes already included: dbfs, s3, s3a, s3n, wasbs, gs, abfs, abfss. Cloud " +
```
jlowe, line 743: Nit: It would be nice to not have to keep these manually in-sync, and also would be nice to list them alphabetically.  For example, moving the default cloud scheme list from `GpuMultiFileReader` to here with something like:
```scala
/** List of schemes that are always considered cloud storage schemes */
val DEFAULT_CLOUD_SCHEMES = Seq("abfs", "abfss", "dbfs", "gs", "s3", "s3a", "s3n", "wasbs")

val CLOUD_SCHEMES = conf("spark.rapids.cloudSchemes")
[...]
    s"filesystems. Schemes already included: ${DEFAULT_CLOUD_SCHEMES.mkString(", ")}. Cloud " +
```

Then GpuMultiFileReader can create its cloud schemes HashSet from the list in RapidsConf.

jlowe (dismissed):
Looks OK to me.  I'd like to see the nit addressed, but it can be a followup.

tgravescs:
did we run any perf tests on this just to make sure this performed better?

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
wbo4958: Thx, Done

wbo4958:
> did we run any perf tests on this just to make sure this performed better?

I  just did the perf test for COALESCING and cloud reading for total 1.3G and 5000 orc files residing in Azure.

||total time|
|---|---|
|Cloud reading|506.653s|
|Coalescing reading|696.223s|

tgravescs (approved):

jlowe (approved):

jlowe:
build

PR #4087, opened 2021-11-11
Regex parser improvements and bug fixes

andygrove:

This PR fixes some issues I found during more extensive testing with random inputs.

- Character classes can now contain `]` as the first character
- Choices within groups are now being parsed correctly
- Improved checks for base terms started with quantifier symbols

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
@@ -44,18 +44,18 @@ class RegexParser(pattern: String) {
   private var pos = 0
 
   def parse(): RegexAST = {
-    val ast = parseInternal()
+    val ast = parseUntil(() => eof())
     if (!eof()) {
       throw new RegexUnsupportedException("failed to parse full regex")
     }
     ast
   }
 
-  private def parseInternal(): RegexAST = {
-    val term = parseTerm(() => peek().contains('|'))
+  private def parseUntil(until: () => Boolean): RegexAST = {
+    val term = parseTerm(() => until() || peek().contains('|'))
     if (!eof() && peek().contains('|')) {
       consumeExpected('|')
-      RegexChoice(term, parseInternal())
+      RegexChoice(term, parseUntil(until))
     } else {
       term
     }
@@ -64,7 +64,7 @@ class RegexParser(pattern: String) {
   private def parseTerm(until: () => Boolean): RegexAST = {
     val sequence = RegexSequence(new ListBuffer())
     while (!eof() && !until()) {
-      parseFactor() match {
+      parseFactor(until) match {
         case RegexSequence(parts) =>
           sequence.parts ++= parts
         case other =>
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionParserSuite.scala
```diff
@@ -105,51 +136,50 @@ class RegularExpressionParserSuite extends FunSuite {
...
+                RegexGroup(capture = true, RegexSequence(ListBuffer(
                 RegexRepetition(
-                  RegexCharacterClass(negated = false,
-                    ListBuffer(RegexCharacterRange('0', '9'))),SimpleQuantifier('*')),
-                RegexEscaped('.'),
+                    RegexCharacterClass(negated = false, ListBuffer(RegexCharacterRange('0', '9'))),
+                    SimpleQuantifier('+')), RegexEscaped('.'),
                 RegexRepetition(RegexCharacterClass(negated = false,
-                  ListBuffer(RegexCharacterRange('0', '9'))),SimpleQuantifier('+'))))),
-                RegexChar('|'),
```
andygrove, line 128: Previously we were detecting `|` as a character and not as a choice delimiter

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -44,18 +44,18 @@ class RegexParser(pattern: String) {
...
       throw new RegexUnsupportedException("failed to parse full regex")
     }
     ast
   }
 
-  private def parseInternal(): RegexAST = {
-    val term = parseTerm(() => peek().contains('|'))
+  private def parseUntil(until: () => Boolean): RegexAST = {
+    val term = parseTerm(() => until() || peek().contains('|'))
     if (!eof() && peek().contains('|')) {
```
jlowe, line 56: Nit: Not needed as part of this PR, but I think `peek()` will return `None` at EOF, so this can be simplified to:
```suggestion
    if (peek().contains('|')) {
```
Applies to a few other places where `eof` and `peek` are combined.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -89,9 +89,10 @@ class RegexParser(pattern: String) {
     }
   }
 
-  private def parseFactor(): RegexAST = {
+  private def parseFactor(until: () => Boolean): RegexAST = {
     var base = parseBase()
-    while (!eof() && (peek().exists(ch => ch == '*' || ch == '+' || ch == '?')
+    while (!eof() && !until()
+        && (peek().exists(ch => ch == '*' || ch == '+' || ch == '?')
```
jlowe, line 95: Per previous nit, should be able to simplify this to something like:
```suggestion
    while (!until() && (peek().exists(ch => ch == '*' || ch == '+' || ch == '?')
```

jlowe (approved):

andygrove:
build

PR #4093, opened 2021-11-12
Add tests and docs for regexp() and regexp_like()

andygrove:

Spark 3.2.0 adds `regexp` and `regexp_like` SQL functions which are synonyms for the `RLike` expression which we already support.

This PR simply adds tests and docs for this.


andygrove:
build

integration_tests/src/main/python/string_test.py
```diff
@@ -469,7 +469,29 @@ def test_like_complex_escape():
                 'a like "\\%SystemDrive\\%\\\\\\\\Users%"',
                 'a like "_oo"'),
             conf={'spark.sql.parser.escapedStringLiterals': 'true'})
+
+@pytest.mark.skipif(not is_before_spark_320() or is_databricks91_or_later(), reason='regexp is synonym for RLike starting in Spark 3.2.0')
```
jlowe, line 473: This seems to be written backwards and will only run on Spark < 3.2.0 or Databricks 9.1.  Unless Databricks supports these in 9.1, I think this needs to be:
```suggestion
@pytest.mark.skipif(is_before_spark_320(), reason='regexp is synonym for RLike starting in Spark 3.2.0')
```

integration_tests/src/main/python/string_test.py
```diff
@@ -469,7 +469,29 @@ def test_like_complex_escape():
...
+    gen = mk_str_gen('[abcd]{1,3}')
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark: unary_op_df(spark, gen).selectExpr(
+                'regexp(a, "a{2}")',
+                'regexp(a, "a{1,3}")',
+                'regexp(a, "a{1,}")',
+                'regexp(a, "a[bc]d")'),
+            conf={'spark.rapids.sql.expression.RLike': 'true'})
  
+@pytest.mark.skipif(not is_before_spark_320() or is_databricks91_or_later(), reason='regexp_like is synonym for RLike starting in Spark 3.2.0')
```
jlowe, line 484: Same "is this backwards?" comment here.

andygrove:
build

jlowe (dismissed):
Need to resolve the conflict but otherwise lgtm.

andygrove:
build

andygrove:
build

andygrove:
build

jlowe (approved):

PR #4116, opened 2021-11-16
Add in tests to check for overflow in unbounded window [databricks]

### LINKED ISSUE #4084: [BUG] window on unbounded preceeding and unbounded following can produce incorrect results.
**Describe the bug**
I tried to do a very basic window operation. The simplest I could think of.

```
spark.time(spark.range(5L).selectExpr("SUM(1) OVER()").show(truncate=false))
```

This produces nulls for all values on the GPU, but on the CPU it is all 5s. I thought this might be related to ordering, but it is not.

```
spark.time(spark.range(5L).selectExpr("SUM(1) OVER(ORDER BY id ROWS BETWEEN unbounded preceding and unbounded following)").collect())
```

does the same thing.

It gets even odder with COUNT, that returns values, that are wrong, but the cast to string is all empty strings...

revans2:
This fixes #4084 

The actual fix went into CUDF, but this adds in a test to show that it is working.

### DIFF
--- a/integration_tests/src/main/python/window_function_test.py
+++ b/integration_tests/src/main/python/window_function_test.py
@@ -940,3 +940,11 @@ def test_window_rows_stddev(preceding, following):
 
     assert_gpu_and_cpu_are_equal_collect(do_it)
 
+@ignore_order
+def test_unbounded_to_unbounded_window():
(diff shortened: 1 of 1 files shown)


revans2:
build

jlowe (approved):

PR #4122, opened 2021-11-16
Row(CPU)-based UDF supports metric for the run time.

firestarman:
This fixes https://github.com/NVIDIA/spark-rapids/issues/4092


firestarman:
Filed this for early review.

firestarman:
Verified locally and it works well.

firestarman:
rebased to 22.02

firestarman:
build

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala
```diff
@@ -260,7 +262,7 @@ trait GpuExec extends SparkPlan with Arm {
   lazy val allMetrics: Map[String, GpuMetric] = Map(
     NUM_OUTPUT_ROWS -> createMetric(outputRowsLevel, DESCRIPTION_NUM_OUTPUT_ROWS),
     NUM_OUTPUT_BATCHES -> createMetric(outputBatchesLevel, DESCRIPTION_NUM_OUTPUT_BATCHES)) ++
-      additionalMetrics
+      udfRunTimeMetrics ++ additionalMetrics
```
jlowe, line 265: Do we need to worry about any execs that have overidden `allMetrics`?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala
```diff
@@ -278,6 +280,26 @@ trait GpuExec extends SparkPlan with Arm {
     SEMAPHORE_WAIT_TIME -> createNanoTimingMetric(DEBUG_LEVEL, DESCRIPTION_SEMAPHORE_WAIT_TIME)
   )
 
+  // FIXME Will it be better to call this in child execs where UDFs may exist, which requires
+  // listing all the execs that could have UDFs, instead of always iterating the expressions
```
jlowe, line 284: UDFs can appear just about anywhere `Expression` appears, so I don't think tracking down every specific exec is going to work well in practice.

sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRowBasedScalaUDF.scala
```diff
@@ -33,7 +33,8 @@ case class GpuRowBasedScalaUDF(
     outputEncoder: Option[ExpressionEncoder[_]],
     udfName: Option[String],
     nullable: Boolean,
-    udfDeterministic: Boolean)
+    udfDeterministic: Boolean,
+    var runTime: GpuMetric = NoopMetric)
```
jlowe, line 37: I'm a bit nervous about this approach.  Plan nodes get copied around and compared all the time, and the use of `var` in a case class is sort of an anti-pattern.  Also metrics are typically lazily created all over the code, yet we're forcing them to be evaluated up-front here.

Plans are supposed to be immutable, so typically when we want to change something in the plan we end up re-writing the plan with updated nodes.  I'm wondering if that's what we should be doing here, i.e.: rewriting the plan with an updated metric parameter which is not variable and is either a metric as above or possibly a function argument that provides an optional metric which would allow the metric creation to remain lazy.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala
firestarman: Got it. Thanks a lot.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExec.scala
firestarman: I think it should be the developer's job to take care of these if he want to override this `allMetrics`.

sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRowBasedScalaUDF.scala
firestarman: Oh, yes, I think you are right. I need more consideration for this.

firestarman:
Close this PR since it is not easy to support this.

Issue #3970, opened 2021-10-29
[BUG] mismatching timezone settings on executor and driver can cause ORC read data corruption


tgravescs:
**Describe the bug**
A user reported orc_test failing when running in their environment. The diffs were in the hours of the DateTime types.

One of them was:
FAILED src/main/python/orc_test.py::test_basic_read[{'spark.rapids.sql.format.orc.reader.type': 'PERFILE'}-native--read_orc_df-timestamp-date-test.orc] 
with lots of others as well....

After some debugging it turns out they only set:
--conf spark.driver.extraJavaOptions=-Duser.timezone=UTC 

and were missing:
--conf spark.executor.extraJavaOptions=-Duser.timezone=UTC \
--conf spark.sql.session.timeZone=UTC

The Host timezone was set to: Time zone: America/New_York (EDT, -0400)

So here I think the planning on driver passed because it was UTC but executors weren't UTC so data returned wasn't the same as CPU generated.

Perhaps we can add more validation on executor side to make sure timezone UTC, if nothing else throw so it fails rather then corrupting data.

Note I haven't tried to reproduce this yet.  The user set all the timezone settings properly and the test started to pass.

nartal1:
I was able to reproduce this bug on YARN cluster by passing `--conf spark.driver.extraJavaOptions=-Duser.timezone=America/New_York` to the tests. 

After further looking into the results, the mismatch in GPU and CPU results is that in CPU the timestamp is read in the timezone provided i.e America/New_York(EDT) time but in the GPU results the timestamps are read in "UTC" .  

I thought it would have failed by giving unsupported data type error(as spark-rapids supports on UTC) but that is not the case. GPU is reading in UTC.

PR #4129, opened 2021-11-17
Check executor timezone is same as driver timezone when running on GPU

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3970 

In this PR we are validating the executor timezone. If the driver timezone is supported to run on GPU, then we check if the executor timezone is also same. If executor timezone is different then we throw exception with the message to set executor timezone to match driver  time zone. 
Since it is reproducible only in cluster setup, I haven't added any tests. Verified this PR by running orc_test.py in YARN cluster.

Thanks @tgravescs for helping in debugging this issue.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -53,6 +54,7 @@ case class ColumnarOverrideRules() extends ColumnarRule with Logging {
 object RapidsPluginUtils extends Logging {
   val CUDF_PROPS_FILENAME = "cudf-java-version-info.properties"
   val PLUGIN_PROPS_FILENAME = "rapids4spark-version-info.properties"
+  val DRIVER_TIMEZONE_KEY = "spark.rapids.driver.user.timezone"
```
jlowe, line 57: Nit: This key should be added to RapidsConf as an internal config with the corresponding lazy val accessor to document the usage in the namespace of the configs.  It can be made an optional config, and the executor code fetching it can throw if it fetches the conf key and gets `None`.  That eliminates the need for a new `get` method that throws.

nartal1: Thanks @jlowe for reviewing the PR. The reason I had kept it here was that I didn't want to expose this config to the user. If we added in RapidsConf as an internal config, then do we risk the config to be set by user which may not be the correct timezone. Please let me know if my understanding is wrong.

jlowe: Leaving the config completely undocumented does not remove the risk of what you're worried about.  The user could still end up setting that config, and if we fail to override it properly at the point where it matters then it will result in the same situation.

By documenting it as an internal config that will be programmatically set, we're educating the maintainers about this use of the config namespace.  I'm not worried about the user specifying this config, as the plugin should properly smash any previous value with the value on the driver side.  You can verify this by trying to set it yourself and see if it gets picked up or not.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -179,6 +183,16 @@ class RapidsExecutorPlugin extends ExecutorPlugin with Logging {
       // by setting this config spark.rapids.cudfVersionOverride=true
       checkCudfVersion(conf)
 
+      // Validate executor time zone to be "UTC" if the driver time zone is "UTC".
+      val driverTimezone = conf.get(RapidsPluginUtils.DRIVER_TIMEZONE_KEY)
+      if (driverTimezone == GpuOverrides.UTC_TIMEZONE_ID.toString) {
```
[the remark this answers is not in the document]
nartal1: Yes, you are right.  I have added the check in TypeChecks. In the future if we support different time zones this check still needs to be updated, right? 
I see that we are doing similar checks if the datatype is Timestamp i.e  `ZoneId.systemDefault().normalized() == GpuOverrides.UTC_TIMEZONE_ID` . So I am assuming these checks will also be updated if additional timezones are supported.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -179,6 +183,16 @@ class RapidsExecutorPlugin extends ExecutorPlugin with Logging {
...
+        val executorTimezone = ZoneId.systemDefault().normalized()
+        if (executorTimezone != GpuOverrides.UTC_TIMEZONE_ID) {
+          throw new RuntimeException("Driver timezone is UTC. Set executor timezone " +
+              "to UTC as well. Add --conf spark.executor.extraJavaOptions=-Duser.timezone=UTC")
```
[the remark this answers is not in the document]
nartal1: done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -1656,6 +1656,17 @@ object ExprChecks {
...
 }
 
+object DriverTimeZone {
+
+  /**
+   * Check if the time zone of the driver is supported by plugin.
+   * @param driverZoneId time zone of the driver
+   */
+  def isSupportedByPlugin(driverZoneId: String): Boolean = {
+    driverZoneId == GpuOverrides.UTC_TIMEZONE_ID.toString
```
jlowe, line 1666: I'd like to see this be part of `TypeChecks`, called something like `areTimestampsSupported`, and take a `ZoneId` which is normalized before checking against `UTC_TIMEZONE_ID`.  That way the code is more robust in handling timezones that are aliased.  As written, it will fail to compare them equal.

Regarding the comment of needing to remember to update the timestamp checking code, the best way to handle that is have the timezone checking code call this new method.  That way there's only one place that needs to be updated.

I see there's a number of places that are reaching into GpuOverrides to access this timezone ID and do their own checking, so those places should be updated to call this new TypeChecks method.  That can be done as a followup if desired.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -440,6 +440,13 @@ object RapidsConf {
     .bytesConf(ByteUnit.BYTE)
     .createWithDefault(Integer.MAX_VALUE)
 
+  val DRIVER_TIMEZONE = conf("spark.rapids.driver.user.timezone")
+    .doc("This is to set driver timezone. Note that if the user provides this config " +
```
jlowe, line 444: As written, this implies it could be used to set the timezone on the driver.  It would be clearer to state that the config is used to inform the executor plugin about the driver's timezone and is not intended to be set by the user.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
nartal1: Added `areTimestampsSupported`  method in TypeChecks and updated the code in other places which were doing their own checks to use this method. PTAL.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBatchScanExec.scala
```diff
@@ -266,7 +266,7 @@ object GpuCSVScan {
         meta.willNotWorkOnGpu("GpuCSVScan does not support parsing timestamp types. To " +
           s"enable it please set ${RapidsConf.ENABLE_CSV_TIMESTAMPS} to true.")
       }
-      if (parsedOptions.zoneId.normalized() != GpuOverrides.UTC_TIMEZONE_ID) {
+      if (!TypeChecks.areTimestampsSupported(parsedOptions.zoneId.normalized())) {
```
jlowe, line 269: `areTimestampsSupported` needs to do the normalization internally to properly compare arbitrary timezone IDs passed to it.  It should not be the caller's responsibility to normalize first, or we risk someone forgetting to normalize.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1444,6 +1451,14 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
     entry.get(conf)
   }
 
+  def get(key: String): String = {
```
jlowe, line 1454: This method should not be here.

jlowe (approved):

jlowe:
build

nartal1:
CI failure due to this error:
```
Error: :35:22.567Z] [ERROR] [Error] /home/jenkins/agent/workspace/jenkins-rapids_premerge-github-3317/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuOrcFileFormat.scala:132: not found: value OrcFileFormat
Error: 1-19T00:35:22.569Z] [ERROR] one error found
```

I will upmerge and restart CI once https://github.com/NVIDIA/spark-rapids/pull/4139 gets merged which has the fix for above error.

nartal1:
build

nartal1:
build

PR #4147, opened 2021-11-18
Add nSpect allow list to skip false positive issues [skip ci]

zhanga5:

More details please refer to http://nv/2mB (need vpn)


.nspect-vuln-allowlist.toml
```diff
@@ -0,0 +1,39 @@
+version = "1.0.0"
```
pxLi, line 1: we need a license header here, can you add and check locally to pass the rat check?

PR #4153, opened 2021-11-18
Regexp: Only transpile once per expression rather than once per batch

andygrove:
The main motivation for this PR was to implement a performance optimization so that regular expressions get transpiled just once per expression (twice if you count the tagging) rather than being transpiled for each invocation (per batch).

While refactoring, I found some inconsistencies and unreachable code paths related to null patterns and empty patterns which are also fixed in this PR along with additional tests.

### DIFF
--- a/sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
+++ b/sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
@@ -31,8 +31,6 @@ class GpuRegExpReplaceMeta(
 
   override def tagExprForGpu(): Unit = {
     expr.regexp match {
-      case Literal(null, _) =>
-        willNotWorkOnGpu(s"null pattern is not supported on GPU")
       case Literal(s: UTF8String, DataTypes.StringType) =>
         val pattern = s.toString
         if (pattern.isEmpty) {
--- a/sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
+++ b/sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
@@ -32,16 +32,11 @@ class GpuRegExpReplaceMeta(
   override def tagExprForGpu(): Unit = {
     expr.regexp match {
       case Literal(s: UTF8String, DataTypes.StringType) =>
-        val pattern = s.toString
-        if (pattern.isEmpty) {
-          willNotWorkOnGpu(s"empty pattern is not supported on GPU")
-        }
-
         if (GpuOverrides.isSupportedStringReplacePattern(expr.regexp)) {
           // use GpuStringReplace
         } else {
           try {
-            new CudfRegexTranspiler(replace = true).transpile(pattern)
+            new CudfRegexTranspiler(replace = true).transpile(s.toString)
           } catch {
             case e: RegexUnsupportedException =>
               willNotWorkOnGpu(e.getMessage)
--- a/sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
+++ b/sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
@@ -15,8 +15,6 @@
  */
 package com.nvidia.spark.rapids.shims.v2
 
-import java.sql.SQLException
-
 import com.nvidia.spark.rapids.{CudfRegexTranspiler, DataFromReplacementRule, GpuExpression, GpuOverrides, RapidsConf, RapidsMeta, RegexUnsupportedException, TernaryExprMeta}
 
 import org.apache.spark.sql.catalyst.expressions.{Expression, Literal, RegExpReplace}
@@ -60,7 +58,7 @@ class GpuRegExpReplaceMeta(
       GpuStringReplace(lhs, regexp, rep)
     } else {
       GpuRegExpReplace(lhs, regexp, rep, pattern.getOrElse(
-        throw new SQLException("Expression has not been tagged with cuDF regex pattern")))
+        throw new IllegalStateException("Expression has not been tagged with cuDF regex pattern")))
     }
   }
 }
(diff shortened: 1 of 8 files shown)


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -874,21 +878,7 @@ case class GpuRegExpReplace(
       searchExpr: GpuScalar,
       replaceExpr: GpuScalar): ColumnVector = {
 
-    searchExpr.getValue match {
-      case null =>
-        // Return original string if search string is null
-        strExpr.getBase.asStrings()
```
andygrove, line 880: This code path was unreachable

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
```diff
@@ -31,19 +31,12 @@ class GpuRegExpReplaceMeta(
 
   override def tagExprForGpu(): Unit = {
     expr.regexp match {
-      case Literal(null, _) =>
-        willNotWorkOnGpu(s"null pattern is not supported on GPU")
```
jlowe, line 35: I don't see this case handled anymore?  If the null's type isn't UTF8String then it will incorrectly state that a non-literal pattern is not supported (but it is a literal).  If the null's type is UTF8String then it looks like we'll NPE when we try to convert it to a string below.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -772,6 +772,20 @@ class GpuRLikeMeta(
 case class GpuRLike(left: Expression, right: Expression)
   extends GpuBinaryExpression with ImplicitCastInputTypes with NullIntolerant  {
 
+  private val cudfPattern = right.asInstanceOf[GpuLiteral].value match {
```
jlowe, line 775: Should this be lazy so we don't try to evaluate it every time this node is copied in the plan or the node is otherwise wielded in a way where we're not trying to execute it (e.g.: on the driver)?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -835,6 +825,20 @@ case class GpuRegExpReplace(
   override def second: Expression = searchExpr
   override def third: Expression = replaceExpr
 
+  private val cudfPattern = searchExpr.asInstanceOf[GpuLiteral].value match {
```
jlowe, line 828: same lazy question here

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -44,6 +44,9 @@ class RegexParser(pattern: String) {
   private var pos = 0
 
   def parse(): RegexAST = {
+    if (pattern == null) {
```
jlowe, line 47: Since the code seems to always pass the result of a toString on the parameter, does this happen in practice?

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
```diff
@@ -31,19 +31,12 @@ class GpuRegExpReplaceMeta(
...
-        if (pattern.isEmpty) {
-          willNotWorkOnGpu(s"empty pattern is not supported on GPU")
-        }
-
         if (GpuOverrides.isSupportedStringReplacePattern(expr.regexp)) {
           // use GpuStringReplace
         } else {
           try {
-            new CudfRegexTranspiler(replace = true).transpile(pattern)
+            new CudfRegexTranspiler(replace = true).transpile(s.toString)
```
jlowe, line 39: Seems like a waste to throw away the transpile and make us do it again later.  Curious why we don't cache the transpiled expression in the tagging and then pass it to the GPU exec node as a parameter when we convert to the GPU.

andygrove: Thanks .. I have re-implemented so that we only transpile once and cache the result in the meta classes.

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
andygrove: It shouldn't be possible to have a null literal in this context since spark would translate it to a `cast(null as string)` but I will make this more defensive and update the error messages.

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
```diff
@@ -62,7 +59,8 @@ class GpuRegExpReplaceMeta(
     if (GpuOverrides.isSupportedStringReplacePattern(expr.regexp)) {
       GpuStringReplace(lhs, regexp, rep)
     } else {
-      GpuRegExpReplace(lhs, regexp, rep)
+      GpuRegExpReplace(lhs, regexp, rep, pattern.getOrElse(
+        throw new SQLException("Expression has not been tagged with cuDF regex pattern")))
```
jlowe, line 63: Curious why this is using SQLException?  We've traditionally used IllegalStateException since this is more of a logic error in the meta/override code if it occurs.

jlowe (approved):

jlowe:
build

revans2 (approved):

andygrove:
I'm not sure why the bulid failed. The last entry in the log was:

```
2021-11-29T18:10:22.4493599Z [2021-11-29T18:09:46.702Z] [gw3]^[[36m [ 81%] ^[[0m^[[33mSKIPPED^[[0m ../../src/main/python/repart_test.py::test_union_struct_missing_children[(Struct(not_null)(['child0', Struct(['child0', Struct(['child0', Struct(['child0', Struct(['child0', Decimal(7,2)])])])])]), Struct(not_null)(['child1', Integer]))]
2021-11-29T18:10:22.4494720Z script returned exit code 120
2021-11-29T18:10:22.4535049Z Cleaning up orphan processes
```

jlowe:
> I'm not sure why the bulid failed.

This was likely caused by the recent CI storage space exhaustion.  Should be fixed now.

Issue #3894, opened 2021-10-21
[BUG] test_pandas_scalar_udf and test_pandas_map_udf failed in UCX standalone CI run


abellina:
Errors seen in the logs

```
AssertionError: GPU and CPU int values are different at [0, 'nested_size(a)']
```
and 
```
AssertionError: CPU and GPU list have different lengths at [] CPU: 48 GPU: 36
```

```
FAILED ../../src/main/python/udf_test.py::test_pandas_scalar_udf_nested_type[Array(Integer)]
FAILED ../../src/main/python/udf_test.py::test_pandas_scalar_udf_nested_type[Array(Array(String))]
FAILED ../../src/main/python/udf_test.py::test_pandas_map_udf_nested_type[Array(String)][IGNORE_ORDER({'local': True})]
FAILED ../../src/main/python/udf_test.py::test_pandas_map_udf_nested_type[Struct(['child0', Byte],['child1', Short],['child2', Integer],['child3', Long],['child4', Float],['child5', Double],['child6', String],['child7', Boolean],['child8', Date],['child9', Timestamp])][IGNORE_ORDER({'local': True})]
```

Related: https://github.com/NVIDIA/spark-rapids/issues/3029

abellina:
Update on this, I've been looking at reproducing and the good news is that it's fairly easy to reproduce. If I cause failures in the executors (in my case I set a GPU to be disabled, so we failed at startup), I can get tests within the udf_test suite to fail. It's not just these two, but there are others that also fail, and those others are the ones I am most worried about.

Essentially the way to fix this is to set `num_slices`, since the cluster shape is changing as we go from CPU to GPU runs. Some of the tests that fail are checking the length of the column in rows, which is bound to be different. But there are a couple of tests `test_pandas_math_udf` and `test_iterator_math_udf` that fail in odd ways and I am investigating further.

abellina:
For `test_pandas_math_udf` and `test_iterator_math_udf` the tests themselves are rather simple (just an addition) but I am seeing some really odd behavior when we have a case with task failures, like in my scenario with the disabled GPU (1 out of 4 gpus is in mode `PROHIBITED`)

This is a row which is the result of select("a","b", my_udf(a - 3, b)):

GPU: `Row(a=1558966102, b=2126661637, add((a - 3), b)=-609339560)`
CPU: `Row(a=1558966102, b=2126661637, add((a - 3), b)=3685627736)`

[...]

PR #4154, opened 2021-11-18
Use num_slices to guarantee partition shape in the pandas udf tests

abellina:

Closes https://github.com/NVIDIA/spark-rapids/issues/3894

This uses the `num_slices` argument when generating the dataframe such that our tests are resilient to cluster shape changes. This is particular to the pandas/python tests, given that the shape of the partition can mean that math can overflow, due to the nature of the conversions in python.


abellina:
build

revans2 (approved):

abellina:
build

jlowe (approved):

Issue #4091, opened 2021-11-12
[FEA] regexp_replace: Improve support for ^ and $


andygrove:
**Is your feature request related to a problem? Please describe.**
Once https://github.com/NVIDIA/spark-rapids/pull/4063 is merged, we support `regexp_replace` on GPU but we fall back to CPU whenever the pattern contains the BOL/EOL anchors `^` and `$`.

**Describe the solution you'd like**
We should be able to support these anchors in some cases.

**Describe alternatives you've considered**
None

**Additional context**
Depends on https://github.com/rapidsai/cudf/issues/9845

PR #4155, opened 2021-11-18
Add partial support for line begin and end anchors in regexp_replace

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/4091


integration_tests/src/main/python/string_test.py
```diff
@@ -335,6 +335,9 @@ def test_re_replace():
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark: unary_op_df(spark, gen).selectExpr(
                 'REGEXP_REPLACE(a, "TEST", "PROD")',
+                'REGEXP_REPLACE(a, "^TEST", "PROD")',
+                'REGEXP_REPLACE(a, "^TEST$", "PROD")',
+                'REGEXP_REPLACE(a, "TEST$", "PROD")',
```
jlowe, line 340: Do we need any "trickier" tests that include escaped characters or has this been covered by separate fuzz testing?

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -36,7 +36,10 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
       "a*+",
       "\t+|a",
       "(\t+|a)Dc$1",
-      "(?d)"
+      "(?d)",
+      "$|$[^\n]2]}|B",
+      "a^|b",
+      "w$|b"
```
jlowe, line 42: Should a test case like `^`, `$`, or `^$` be added here?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -501,9 +496,16 @@ class CudfRegexTranspiler(replace: Boolean) {
           // falling back to CPU
           throw new RegexUnsupportedException(nothingToRepeat)
         }
-        if (replace && parts.length == 1 && (isRegexChar(parts.head, '^')
-            || isRegexChar(parts.head, '$'))) {
-          throw new RegexUnsupportedException("regexp_replace on GPU does not support ^ or $")
+        if (parts.forall {
+          case RegexChar(ch) => ch == '^' || ch == '$'
```
jlowe, line 500: Do we need to worry about any degenerate cases of other directives (e.g.: empty group, character class, etc.) that would essentially be equivalent to just having these characters in the pattern?

andygrove:
Converting this to WIP since there are regressions caused by other regexp PRs that were recently merged

andygrove:
build

integration_tests/src/main/python/string_test.py
andygrove: I have been mostly relying on fuzz testing to find edge cases that we cannot currently support (and the goal is to fall back to CPU for those cases). I have added some more tests here with escaped `^` and `$` though,

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
andygrove: This test is specifically for cases where a pattern would cause a compilation error in cuDF. The patterns `^`, `$`, and `^$` compile on cuDF but produce different results to Spark.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
andygrove: For empty groups, we would fall back to CPU, but I did find one case that is not detected by this rule (`(^)($)`) so will spend some more time on this.

andygrove:
build

andygrove:
build

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -94,42 +97,38 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
...
-      assertUnsupported(pattern, "cuDF does not support null characters in regular expressions"))
-  }
-
-  test("nothing to repeat") {
-    val patterns = Seq("$*", "^+")
-    patterns.foreach(pattern =>
-      assertUnsupported(pattern, "nothing to repeat"))
+      assertUnsupported(pattern, replace = false,
+        "cuDF does not support null characters in regular expressions"))
   }
```
andygrove, line 132: This test is no longer valid and the patterns tested here have moved to a different test

andygrove:
Sign-off check is failing due to:

```
  Step 1/5 : FROM python:alpine
  Head "https://registry-1.docker.io/v2/library/python/manifests/alpine": received unexpected HTTP status: 503 Service Unavailable
  ```

andygrove:
build

jlowe (approved):
CI is failing due to an unrelated issue, see #4423.

sameerz:
build

PR #4159, opened 2021-11-19
Optimize sample perf

### LINKED ISSUE #4096: [BUG] GpuSampleExec is very slow, and missing opTime in some cases
**Describe the bug**
I was trying to process some data quickly that was more then could fit in memory, and I didn't need an exact answer, so I decided to use sample to do it.

I found that to process 21,474,836,470 rows at a 1% sample rate took 3.9 mins to do it. And that was only after I added in the opTime metric myself. It was almost faster to just spill the data, then it was to use this.

Switching the for loop to

```
                    var i = 0
                    while (i < numRows) {
                      i = i + 1
```

Took off 2 mins of the time, but it is still very slow. We should do some profiling and try to figure out exactly what is happening. How we can speed it up.

We might also want to think about a config that makes it so we use the GPU sampling code instead.

res-life:
This fixes #4096 

Samples data by GPU JNI to improve the performance. 
Added "spark.rapids.sql.fast.sample" to switch.


res-life:
```
The bottlenecks are XORShiftRandom.sample and ArrayBuffer append.
	It's hard to optimize the XORShiftRandom.sample. XORShiftRandom is serialized algorithm(current value depends on privious value), can't run parallelly.
	For ArrayBuffer append, it's also hard to avoid, we should collect the sampled row numbers.
Optimize
	Add pure GPU solution, will be turned on by swithing the config "spark.rapids.sql.fast.sample". 
	This solution is inconsistent with CPU exec although it's faster.
   This solution is 3x faster than the previous.
	Test Deail
		Data set: sample 1% from 70,000,000 rows
		Result:
			1. CPU samples and GPU gathers(Previous solution): avg 275 ms
			2. GPU sample JNI:  used time: avg 81ms
Others
	Add the metrics for all the scenarios
	Fixed the bug: When ColumnBach has no coulmns, throws an exception.
  ```

Note: depends on Cudf issue: https://github.com/rapidsai/cudf/pull/9728

**Comment by res-life** [comment]
build

**Comment by res-life** [comment]
Build failed because of CUDF PR is not merged.

**Comment by revans2** [comment]
Marking as draft because the CUDF JNI dependency is not in yet.

File: sql-plugin/src/main/scala/com/nvidia/spark/rapids/GatherUtils.scala [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+      c.setNumRows(rows.length)
+      c
+    } else {
+      val intBytes = DType.INT32.getSizeInBytes()
+      val totalBytes = rows.length * intBytes
+      withResource(HostMemoryBuffer.allocate(totalBytes)) { hostBuffer =>
+        // copy row indexes to host buffer
+        var idx = 0
+        while (idx < rows.length) {
+          hostBuffer.setInt(idx * intBytes, rows(idx))
```
revans2, line 43: This is mostly for my own curiosity.  You could have also done this with

```scala
withResource(cudf.ColumnVector.fromInts(rows: _*)) { gatherCv =>
  withResource(GpuColumnVector.from(cb)) { table =>
    // GPU gather
    withResource(table.gather(gatherCv)) { gatheredTable =>
      GpuColumnVector.from(gatheredTable, colTypes)
    }
  }
}
```

Did you try this? If so was your current way more performant? I realize it might involve an extra memory copy because the `ArrayBuffer` needs to be transformed into an `Array`, so it can be passed to `fromInts`. So if you didn't I don't think it is worth spending too much time on it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
```diff
@@ -406,56 +411,81 @@ case class GpuSampleExec(lowerBound: Double, upperBound: Double, withReplacement
...
     val rdd = child.executeColumnar()
-    if (withReplacement) {
-      new GpuPartitionwiseSampledRDD(
-        rdd,
-        new GpuPoissonSampler(upperBound - lowerBound, useGapSamplingIfPossible = false,
-          numOutputRows, numOutputBatches, opTime),
-        preservesPartitioning = true,
-        seed)
-    } else {
+    val useGpuToSample = new RapidsConf(conf).isFastSampleEnabled
```
revans2, line 416: nit: I think it would be cleaner if we had a `GpuFastSampleExec` and a `GpuSampleExec`, then we can select which one to use when we replace it.  Part of this is because creating a RapidsConf is not cheap.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
```diff
@@ -406,56 +411,81 @@ case class GpuSampleExec(lowerBound: Double, upperBound: Double, withReplacement
...
+                if (numSampleRows == 0L) {
+                  GpuColumnVector.emptyBatchFromTypes(colTypes)
+                } else if(cb.numCols() == 0) {
+                    // for count agg, num of cols is 0
+                    val c = GpuColumnVector.emptyBatchFromTypes(colTypes)
+                    c.setNumRows(numSampleRows.toInt)
+                    c
+                } else {
+                  withResource(GpuColumnVector.from(cb)) { table =>
+                    withResource(table.sample(numSampleRows, withReplacement, seed)) { sampled =>
```
revans2, line 437: nit: Because we use the same seed every time then I think all of the batches will be sampled the same way. I am not sure mathematically how that all works out.  Could we at least set the seed to `seed + index` so each task does it slightly differently?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
res-life: Yes, done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
res-life: Updated, use the way like CPU to set the seed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GatherUtils.scala
res-life: Updated, code is more neat and more effecient.

res-life:
```
The test result is as follow:
data set
	parquet file only has one int column, one partition, has 2,100,000,000 rows.
config is
	spark.executor.cores=1
	spark.rapids.sql.concurrentGpuTasks=1
	// limit to one thead
sql is:
	spark.read.parquet("/home/chong/data/parquet").sample(0.01, 0).agg(f.sum("salary")).show()
result
	CPU used time: 4.6s
	GPU: 5.1s  		   // 	spark.conf.set("spark.rapids.sql.fast.sample", True)
	CPU GPU hybrid: 11.41s  // 	spark.conf.set("spark.rapids.sql.fast.sample", False)
conclusion
	GPU is a little slow than CPU, maybe we need to file a CUDF issue to improve the performance of the sample JNI
```

res-life:
build

res-life:
@revans2  Please see the above test result; Please help to review.

revans2:
I get very different results, and I think your issue is because of Parquet reading.

I ran `spark.time(spark.range(Int.MaxValue).sample(0.01, 0).count)`

First single task/thread. This is the median of 3 runs, and measured in ms.

|CPU|GPU|FAST GPU|
|-|-|-|
|10633|11108|83|

It is also clear from the op time metrics on the UI for the runs.

When I run with 12 tasks/threads I get


|CPU|GPU|FAST GPU|
|-|-|-|
|1236|2952|100|

The regular GPU being slower is because I have my GPU semaphore set to 4 in parallel. So it cannot fully utilize the CPU when running.

If I change the query to include replacement `spark.time(spark.range(Int.MaxValue).sample(0.01, 0).count)` I get.

|CPU|GPU|FAST GPU|
|-|-|-|
|11135|22556|116|

So at this point the only thing that is lacking here is some documentation to let people know that there is a crazy fast version that is not the same as the Spark version.

res-life:
The following optimization makes counting a data frame very fast.

```
GpuFastSampleExec
  } else if (cb.numCols() == 0) {
	  // for count agg, num of cols is 0
	  val c = GpuColumnVector.emptyBatchFromTypes(colTypes)
	  c.setNumRows(numSampleRows.toInt)  // can skip GPU sample.
	  c
  } else {
```

revans2:
Yup I got ahead of myself and did things too quickly. You are 100% correct. Could you file a follow on issue to then look at what we can do to possibly speed up sampling on the GPU?

revans2 (approved):

revans2:
To be clear the issue appears to be related to doing it without replacement.  With replacement is really fast. This indicates that `thrust::shuffle_copy` is the reason for the slowness.

```
spark.time(spark.range(Int.MaxValue * 12L).sample(true, 0.01, 0).selectExpr("SUM(id)", "COUNT(id)").show())
+-------------------+---------+
|            sum(id)|count(id)|
+-------------------+---------+
|3320410258800588221|257697960|
+-------------------+---------+

Time taken: 670 ms
```

res-life:
build

res-life:
build

res-life:
The issue filed https://github.com/rapidsai/cudf/issues/9834

res-life:
Updated the doc, please review again @revans2

revans2 (approved):

PR #4163, opened 2021-11-19
Enable config to check for casting decimals to strings

### LINKED ISSUE #4162: [BUG] isCastDecimalToStringEnabled is never called
Turns out that when we added in this config it was never actually called.  This is probably okay because decimal types have been disabled by default. But in 21.12 they will be turned on so we should get this fixed.

revans2:
This fixes #4162

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
@@ -72,9 +72,9 @@ final class CastExprMeta[INPUT <: CastBase](
         willNotWorkOnGpu(buildTagMessage(RapidsConf.ENABLE_CAST_FLOAT_TO_INTEGRAL_TYPES))
       case (dt: DecimalType, _: StringType) =>
         if (!conf.isCastDecimalToStringEnabled) {
-          willNotWorkOnGpu("The GPU does NOT produce exact same string as Spark produces, " +
-              s"set ${RapidsConf.ENABLE_CAST_DECIMAL_TO_STRING} to true for semantically " +
-              s"equivalent decimal strings produced on the GPU.")
+          willNotWorkOnGpu("the GPU does not produce the exact same string as Spark produces, " +
+              s"set ${RapidsConf.ENABLE_CAST_DECIMAL_TO_STRING} to true if semantically " +
+              s"equivalent decimal strings are sufficient for your application.")
         }
         if (dt.precision > DType.DECIMAL64_MAX_PRECISION) {
           willNotWorkOnGpu(s"decimal to string with a " +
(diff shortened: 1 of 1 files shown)


revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -70,9 +70,16 @@ final class CastExprMeta[INPUT <: CastBase](
       case (FloatType | DoubleType, ByteType | ShortType | IntegerType | LongType) if
           doFloatToIntCheck && !conf.isCastFloatToIntegralTypesEnabled =>
         willNotWorkOnGpu(buildTagMessage(RapidsConf.ENABLE_CAST_FLOAT_TO_INTEGRAL_TYPES))
-      case (dt: DecimalType, _: StringType) if dt.precision > DType.DECIMAL64_MAX_PRECISION =>
-        willNotWorkOnGpu(s"decimal to string with a " +
-            s"precision > ${DType.DECIMAL64_MAX_PRECISION} is not supported yet")
+      case (dt: DecimalType, _: StringType) =>
+        if (!conf.isCastDecimalToStringEnabled) {
+          willNotWorkOnGpu("The GPU does NOT produce exact same string as Spark produces, " +
```
jlowe, line 75:
```suggestion
          willNotWorkOnGpu("the GPU does NOT produce the exact same string as Spark produces, " +
```

Nit: the capitalization of NOT seems a bit excessive given we have plenty of other "because" reasons that don't also capitalize.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -70,9 +70,16 @@ final class CastExprMeta[INPUT <: CastBase](
...
+              s"set ${RapidsConf.ENABLE_CAST_DECIMAL_TO_STRING} to true for semantically " +
+              s"equivalent decimal strings produced on the GPU.")
```
jlowe, line 77:
```suggestion
              s"set ${RapidsConf.ENABLE_CAST_DECIMAL_TO_STRING} to true if semantically " +
              s"equivalent decimal strings are sufficient for your application.")
```

jlowe (approved):

jlowe:
build

gerashegalov (approved):

razajafri (approved):

razajafri:
Shouldn't this require a cudf change to ColumnViewJni.cpp?

razajafri:
> Shouldn't this require a cudf change to ColumnViewJni.cpp?

nm, I thought this was a decimal 128 change

PR #4171, opened 2021-11-19
Update regular expression tests related to `$` and strings ending with `\n`

andygrove:
Update tests and documentation now that https://github.com/rapidsai/cudf/issues/9620 is resolved.

The cuDF fix means that `$` now correctly matches strings that end with `\n` but unfortunately, we still have an issue with strings that end with `\r` so the documentation is updated to refer to a new issue for that.

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
@@ -133,14 +133,13 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
   }
 
   ignore("known issue - multiline difference between CPU and GPU") {
-    // see https://github.com/rapidsai/cudf/issues/9620
+    // see https://github.com/NVIDIA/spark-rapids/issues/4170
     val pattern = "2$"
-    // this matches "2" but not "2\n" on the GPU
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -133,14 +133,13 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
   }
 
   ignore("known issue - multiline difference between CPU and GPU") {
-    // see https://github.com/rapidsai/cudf/issues/9620
+    // see https://github.com/NVIDIA/spark-rapids/issues/4170
     val pattern = "2$"
-    // this matches "2" but not "2\n" on the GPU
+    // this matches "2" and "2\n" but not "2\r" on the GPU
     assertCpuGpuMatchesRegexpFind(Seq(pattern), Seq("2", "2\n", "2\r", "2\r\n"))
```
jlowe, line 139: Do we want to split out the cases that work from the cases that do not so we don't over-ignore here?  For example, are we regularly testing something that ends in `\r\n` anywhere?

andygrove:
I am closing this because the cuDF change is being reverted. I have an alternate approach to fixing this via the transpiler and will open up another PR,

Issue #3879, opened 2021-10-21
[FEA] Decimal 128 Support: Casting


revans2:
We need to finish the rest of the casting support for Decimal 128.

- [X] Casting of decimal values in nested types has not been updated/tested for DECIMAL_128_FULL. PR https://github.com/NVIDIA/spark-rapids/pull/4181
- [ ] Casting to String from a decimal 128 value is not supported yet. PR https://github.com/NVIDIA/spark-rapids/pull/4172
- [ ] Casting from a String is not supported yet. PR https://github.com/NVIDIA/spark-rapids/pull/4172

Many of these are likely to require some changes to CUDF to be able to fully support them.  Casting a string to a Decimal128 might end up being exceptionally difficult as the decimal 64 casting goes through a double first, which can lose some precision compared to a decimal64, but really has problems compared to a decimal 128.

kuhushukla:
The first one for nested cast is fixed via #4181

PR #4172, opened 2021-11-19
Support String to Decimal 128 [databricks]

razajafri:
* Change the check to appropriately check Decimal 128 instead of Decimal 64
* Modified tests 

addresses only part of #3879


revans2 (dismissed):
This looks good. I would like to see us also file a follow on issue for us to look at how we could support rounding directly and support casting with a precision of 38.

razajafri:
build

razajafri:
There seems to be a regression because of a performance improvement in how Strings were casted to Decimal by [this](https://issues.apache.org/jira/browse/SPARK-32706) PR. This is causing the tests to fail against Spark 3.1.1+

I have filed an [issue](https://issues.apache.org/jira/browse/SPARK-37451) for it

razajafri:
build

razajafri:
I accidentally merged branch-21.12. I will have to rectify and force push

integration_tests/src/main/python/cast_test.py
```diff
@@ -167,3 +166,16 @@ def test_cast_long_to_decimal_overflow():
             f.col('a').cast(DecimalType(18, -1))),
         conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': True})
 
+
+@pytest.mark.xfail(not is_before_spark_311(), reason="Only in Spark 3.1.1+ do we have incompatibility with negative decimal scale", raises=IllegalArgumentException)
```
revans2, line 170: If an xfail passes it does not fail it is an XPass that everyone ignores. I would prefer for us to write tests that are much more specific instead.  Is it returning the wrong value? is not on the GPU. What is supposed to actually be happening in these cases?

integration_tests/src/main/python/cast_test.py
```diff
@@ -167,3 +166,16 @@ def test_cast_long_to_decimal_overflow():
...
+def test_cast_string_to_decimal_fail(spark_tmp_path):
+    if (is_before_spark_311()):
+        with_gpu_session(
+            lambda spark: unary_op_df(spark, StringGen("[0-9]{9}")).select(
+                f.col('a').cast(DecimalType(8, -3))), conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': True})
+    else:
+        data_path = spark_tmp_path + '/PARQUET_DATA'
+        with_gpu_session(
+            lambda spark: unary_op_df(spark, StringGen("[0-9]{9}")).select(
```
revans2, line 179: But this does not fail, so why is the test called `_fail`?  It just writes a file out to parquet??? why? Just skip the test if it is before spark 311.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -605,15 +605,13 @@ object RapidsConf {
...
-  val ENABLE_CAST_STRING_TO_DECIMAL = conf("spark.rapids.sql.castStringToDecimal.enabled")
-    .doc("When set to true, enables casting from strings to decimal type on the GPU. Currently " +
-      "string to decimal type on the GPU might produce results which slightly differed from the " +
-      "correct results when the string represents any number exceeding the max precision that " +
-      "CAST_STRING_TO_FLOAT can keep. For instance, the GPU returns 99999999999999987 given " +
-      "input string \"99999999999999999\". The cause of divergence is that we can not cast " +
-      "strings containing scientific notation to decimal directly. So, we have to cast strings " +
-      "to floats firstly. Then, cast floats to decimals. The first step may lead to precision " +
-      "loss.")
+  val ENABLE_LIMITED_NEGATIVE_DECIMAL_SCALE =
```
revans2, line 608: If this is a bug in Spark, then I would rather that we just document that we do not have the same bug, instead of adding a new config.  Also this config is really confusing because it does not mention cast anywhere in it. So users have no idea how we are partially compatible. If we feel we should keep the config, then it needs to be rewritten to express exactly how we are different.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
```diff
@@ -100,6 +100,7 @@ trait SparkShims {
   def int96ParquetRebaseWrite(conf: SQLConf): String
   def int96ParquetRebaseReadKey: String
   def int96ParquetRebaseWriteKey: String
+  def isNegativeDecimalScaleSupportEnabled(conf: RapidsConf): Boolean
```
revans2, line 103: This is too generic of a name. It does not indicate at all in the name, or in the comments that this is just specific to casting.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -115,6 +115,12 @@ final class CastExprMeta[INPUT <: CastBase](
         YearParseUtil.tagParseStringAsDate(conf, this)
       case (_: StringType, _: DateType) =>
         YearParseUtil.tagParseStringAsDate(conf, this)
+      case (_: StringType, _:DecimalType) =>
+        if (!ShimLoader.getSparkShims.isNegativeDecimalScaleSupportEnabled(conf)) {
+          willNotWorkOnGpu("Rapids supports negative decimal scale compatible with Spark " +
```
revans2, line 120: Again this does not mention anything about casting.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
razajafri: We can rename the config to whatever makes the most sense. Do we need the config though? I think we do because if I, as a user, am accelerating a query that is casting a String to Decimal and has a negative scale, the results will not match the CPU spark. How else do we make sure the user knows about this? Is documentation enough? If we decide it is enough then I will remove the config.

revans2: We are trying to be bug for bug compatible with Spark. But it is for negative scale decimals. They are off by default and cannot be written to/read from ORC or Parquet. Also you are the one that found the bug because no one has upgraded to that version yet, or not one is using negative scale decimals. All of that together makes me okay with just documenting that we do it differently.  I would also be fine with just falling back to the CPU 100% of the time for negative scale decimals on these versions of Spark and documenting why. To me the config is just too much overhead to worry about. But you are the one writing this patch, so I will defer to you on that. But if you do have a config it has to be very clear about it being only cor casting negative scale decimal values to strings. In the name of the config. In the documentation for the config. and in the naming of the code/shim layer for all of that too.

integration_tests/src/main/python/cast_test.py
razajafri: Basically what I am trying to achieve is that if the Spark version is >= 311 then this test should throw an IllegalArgument exception for this to xfail.

integration_tests/src/main/python/cast_test.py
razajafri: I have renamed it. It will throw an IllegalArgument exception which will cause the test to xfail. I don't want to skip it, I want to make sure it's falling back to the CPU. Is there a better way to do this?

razajafri:
build

razajafri:
build

integration_tests/src/main/python/cast_test.py
jlowe: >  if the Spark version is >= 311 then this test should throw an IllegalArgument exception for this to xfail.

Do we expect this to ever change (i.e.: is there a bug in the plugin or in Spark that we expect to get fixed in the future such that this test would pass)?  xfail should be reserved for cases where we eventually expect them to pass.  We should not be xfailing tests that we never expect to pass, as that just creates noise in the test summary output and wastes test bandwidth.  Instead we should be skipping such configurations for the test since we know the test premise is invalid for that setup.

integration_tests/src/main/python/cast_test.py
revans2: There are specific APIs for checking if something fell back to the CPU.  If that is what you want you should use that API.

razajafri:
build

integration_tests/src/main/python/cast_test.py
```diff
@@ -167,3 +166,8 @@ def test_cast_long_to_decimal_overflow():
             f.col('a').cast(DecimalType(18, -1))),
         conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': True})
 
+@pytest.mark.skip(not is_before_spark_311(), reason="Only in Spark 3.1.1+ do we have incompatibility with negative decimal scale")
+def test_cast_string_to_decimal(spark_tmp_path):
+        assert_gpu_and_cpu_are_equal_collect(
+            lambda spark: unary_op_df(spark, StringGen("[0-9]{9}")).select(
+                f.col('a').cast(DecimalType(8, -3))), conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': True})
```
jlowe, line 173:
```suggestion
    assert_gpu_and_cpu_are_equal_collect(
        lambda spark: unary_op_df(spark, StringGen("[0-9]{9}")).select(
            f.col('a').cast(DecimalType(8, -3))), conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': True})
```

integration_tests/src/main/python/cast_test.py
```diff
@@ -14,13 +14,12 @@
...
-from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_and_cpu_are_equal_sql, assert_gpu_and_cpu_error, assert_gpu_fallback_collect
+from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_and_cpu_are_equal_sql, assert_gpu_and_cpu_error, assert_gpu_fallback_collect, assert_py4j_exception
 from data_gen import *
-from functools import reduce
-from spark_session import is_before_spark_311, is_before_spark_320
+from spark_session import is_before_spark_311, is_before_spark_320, with_gpu_session
 from marks import allow_non_gpu, approximate_float
 from pyspark.sql.types import *
-from pyspark.sql.functions import array_contains, col, first, isnan, lit, element_at
+from pyspark.sql.utils import IllegalArgumentException
```
jlowe, line 22: Unused import?

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/Spark30XdbShimsBase.scala
```diff
@@ -18,7 +18,7 @@ package com.nvidia.spark.rapids.shims.v2
 
 import scala.collection.mutable.ListBuffer
 
-import com.nvidia.spark.rapids.{ExecChecks, ExecRule, SparkPlanMeta, SparkShims, TypeSig}
+import com.nvidia.spark.rapids.{ExecChecks, ExecRule, RapidsConf, SparkPlanMeta, SparkShims, TypeSig}
```
jlowe, line 21: Unused import

integration_tests/src/main/python/cast_test.py
```diff
@@ -167,3 +166,8 @@ def test_cast_long_to_decimal_overflow():
             f.col('a').cast(DecimalType(18, -1))),
         conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': True})
 
+@pytest.mark.skip(not is_before_spark_311(), reason="Only in Spark 3.1.1+ do we have incompatibility with negative decimal scale")
```
jlowe, line 169: Is this an incompatibility or simply that negative scale decimals aren't supported in Spark 3.1.1+?

integration_tests/src/main/python/cast_test.py
```diff
@@ -167,3 +166,8 @@ def test_cast_long_to_decimal_overflow():
...
+def test_cast_string_to_decimal(spark_tmp_path):
```
jlowe, line 170: Should this be named `test_cast_string_to_negative_scale_decimal` or something similar to note that this isn't just a generic decimal cast test?  Otherwise one would wonder why this isn't being tested across all Spark versions.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -115,6 +115,11 @@ final class CastExprMeta[INPUT <: CastBase](
         YearParseUtil.tagParseStringAsDate(conf, this)
       case (_: StringType, _: DateType) =>
         YearParseUtil.tagParseStringAsDate(conf, this)
+      case (_: StringType, dt:DecimalType) =>
+        if (dt.scale < 0 && !ShimLoader.getSparkShims.isNegativeDecimalScaleSupportEnabled) {
+          willNotWorkOnGpu("Rapids doesn't support negative decimal scale for this version of " +
```
jlowe, line 120:
```suggestion
          willNotWorkOnGpu("RAPIDS doesn't support negative decimal scale for this version of " +
```

jlowe, line 120: Now I'm confused.  The test was changed to skip implying it is not a bug, but this text implies it's a problem/limitation in RAPIDS.  Do we expect this to be fixed by RAPIDS in the future?

sql-plugin/src/main/301until310-nondb/scala/com/nvidia/spark/rapids/shims/v2/Spark30XShims.scala
```diff
@@ -361,6 +361,8 @@ abstract class Spark30XShims extends Spark301util320Shims with Logging {
     adaptivePlan.initialPlan
   }
 
+  override def isNegativeDecimalScaleSupportEnabled: Boolean = true
```
revans2, line 364: This is still too generic. It needs to mention something about casting strings to neg scale decimal

integration_tests/src/main/python/cast_test.py
revans2, line 169: Can we mention the Spark issue here?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
revans2: "RAPIDS doesn't support casting string to decimal for negative scale decimal in this version of Spark because of SPARK-....

razajafri: This isn't a problem with RAPIDS but with Spark. The error message is simply saying that we are incompatible with this particular version of Spark. I have added the Spark bug number to the description for more context

integration_tests/src/main/python/cast_test.py
razajafri: Done

integration_tests/src/main/python/cast_test.py
razajafri: There was a bug introduced in Spark 3.1.1 that has lead to us being incompatible with Spark. I have filed an issue [here](https://issues.apache.org/jira/browse/SPARK-37451). Let me know if this still isn't clear by this message

razajafri:
build

razajafri:
build

razajafri:
build

razajafri:
build

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -911,7 +916,7 @@ object GpuCast extends Arm {
     //    needed. This step is required so we can round up if needed in the final step
     // 4. Now cast newDt to dt (Decimal to Decimal)
     def getInterimDecimalPromoteIfNeeded(dt: DecimalType): DecimalType = {
-      if (dt.precision + 1 > Decimal.MAX_LONG_DIGITS) {
+      if (dt.precision + 1 > DecimalType.MAX_PRECISION) {
         //We don't support Decimal 128
```
revans2, line 920: We do now...

revans2 (dismissed):
Just one minor nit about a comment.

revans2 (dismissed):

razajafri:
build

razajafri:
build

razajafri:
build

revans2 (approved):

PR #4177, opened 2021-11-22
Decimal128: added missing decimal128 signature on Spark 32X

sperlingxx:
Signed-off-by: sperlingxx <lovedreamf@gmail.com>

Fixes #4175

PR #4181, opened 2021-11-22
Fix element_at for 3.2.0 and array/struct cast

revans2:
This will undo some of the changes #4177. It does not break any of the fixes, it just makes things consistent with the rest of the code.  This fixes casting of decimals when they are in an array or a struct for all version of Spark, and fixes element_at for 3.2.0


revans2:
build

jlowe (approved):

revans2:
I just ran through all of the integration tests for 3.2.0 so we should have all of the issues with this figured out.

PR #4195, opened 2021-11-23
[Draft] Trying to support HiveTableScanExec

wbo4958:
This PR is trying to support HiveTableScanExec running on GPU.

HiveTableScanExec will be planned if there is HiveTableRelation, pls refer to [the code]( https://github.com/apache/spark/blob/master/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveStrategies.scala#L254-L270)

But if the Hive table is backend by **Parquet** or **Orc** file format, then the  [HiveMetastoreCatalog](https://github.com/apache/spark/blob/master/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala#L133-L157) will convert HiveTableRelation to HadoopFsRelation by default (there are some confs to control if need to convert).  Finally, HadoopFsRelation will be planed into FileSourceScanExec which will be replaced by GpuFileSourceScanExec.

So if the Hive table is backend by "CSV", then we will get HiveTableScanExec which will not run on GPU. Since rapids plugin has supported to read CSV by GPU, it will be better if we can support HiveTableScanExec backed by CSV running on GPU.

For now, This PR tries to convert HiveTableRelation to HadoopFsRelation and then replace HiveTableScanExec with GpuFileSourceScanExec.

For now, I can't get the correct result with this PR since some exceptions happened

``` console
java.lang.NullPointerException
	at ai.rapids.cudf.HostColumnVectorCore.getStartListOffset(HostColumnVectorCore.java:275)
	at ai.rapids.cudf.HostColumnVectorCore.getUTF8(HostColumnVectorCore.java:365)
	at com.nvidia.spark.rapids.RapidsHostColumnVectorCore.getUTF8String(RapidsHostColumnVectorCore.java:183)
	at org.apache.spark.sql.vectorized.ColumnarBatchRow.getUTF8String(ColumnarBatch.java:220)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.writeFields_0_0$(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.apply(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.apply(Unknown Source)
	at scala.collection.Iterator$$anon$10.next(Iterator.scala:459)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$getByteArrayRdd$1(SparkPlan.scala:346)
	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2(RDD.scala:898)
	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2$adapted(RDD.scala:898)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:373)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:337)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:90)
	at org.apache.spark.scheduler.Task.run(Task.scala:131)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:497)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1439)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:500)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
```

## Do we really need to support HiveTableScanExec?
Spark decide to convert parquet or orc backend hive table according to the Hive SerDe. But for CSV, there is no explicit keyword to match, since there are many SerDe for Csv file.

So I don't know if I need to continue to do this. @revans2 @jlowe Could you provide some advice on this?

### DIFF
--- a/shims/spark312/pom.xml
+++ b/shims/spark312/pom.xml
@@ -81,5 +81,11 @@
             <version>${spark312.version}</version>
 	    <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-hive_2.12</artifactId>
+            <version>${spark312.version}</version>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/shims/spark312/src/main/scala/com/nvidia/spark/rapids/shims/spark312/Spark312Shims.scala
+++ b/shims/spark312/src/main/scala/com/nvidia/spark/rapids/shims/spark312/Spark312Shims.scala
@@ -20,7 +20,9 @@ import com.nvidia.spark.rapids._
 import com.nvidia.spark.rapids.shims.v2._
 import org.apache.parquet.schema.MessageType
 
+import org.apache.spark.sql.execution.SparkPlan
 import org.apache.spark.sql.execution.datasources.parquet.ParquetFilters
+import org.apache.spark.sql.hive.execution.HiveOverrides
 import org.apache.spark.sql.internal.SQLConf
 
 class Spark312Shims extends SparkBaseShims with Spark30Xuntil33XShims {
@@ -41,4 +43,7 @@ class Spark312Shims extends SparkBaseShims with Spark30Xuntil33XShims {
     new ParquetFilters(schema, pushDownDate, pushDownTimestamp, pushDownDecimal, pushDownStartWith,
       pushDownInFilterThreshold, caseSensitive)
   }
+
+  override def getExecs: Map[Class[_ <: SparkPlan], ExecRule[_ <: SparkPlan]] =
+    HiveOverrides.hiveExec ++ super.getExecs
 }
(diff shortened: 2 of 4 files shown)


jlowe:
> Spark decide to convert parquet or orc backend hive table according to the Hive SerDe. But for CSV, there is no explicit keyword to match, since there are many SerDe for Csv file.

We will have to do something similar.  We won't be able to cover ever possible scenario since the serde is pluggable, but we could try to detect common setups and bypass the serde when we recognize what it does and know that what the GPU will do is equivalent.  This is what we're doing today when we bypass the Parquet and ORC FileFormat Java classes, we recognize that the read is using that format, examine the configs to verify what we're going to do will be compatible, then replace the code path with a GPU operation that does not use much, if any, of the Java input format classes.

So the next step here is figuring out what are the cases we want to support and what classes and configs we need to examine to determine whether it is safe to replace the operation with a GPU load instead.

wbo4958:
Close this PR for now.

PR #4222, opened 2021-11-26
Add arguments support to 'databricks/run-tests.py' [databricks]

zhanga5:
in order to run other scripts with arguments, e.g, save event logs for rapids tools testing.
And fixed flake8 reported issues for related .py files.

Signed-off-by: Alex Zhang alex4zhang@gmail.com

Integration test passed on Databricks: http://nv/2nf (need vpn)

### DIFF
--- a/jenkins/databricks/params.py
+++ b/jenkins/databricks/params.py
@@ -11,9 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""Parse input parameters."""
 
-import sys
 import getopt
+import sys
 
 workspace = 'https://dbc-9ff9942e-a9c4.cloud.databricks.com'
 token = ''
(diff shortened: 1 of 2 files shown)


GaryShen2008 (approved):

GaryShen2008:
build

pxLi:
build

jenkins/databricks/params.py
```diff
@@ -11,37 +11,42 @@
...
 
 try:
     opts, args = getopt.getopt(sys.argv[1:], 'hw:t:c:p:l:d:z:m:v:b:j:f:i:',
-                               ['workspace=', 'token=', 'clusterid=', 'private=', 'localscript=', 'dest=', 'sparktgz=', 'basesparkpomversion=', 'buildprofiles=', 'jarpath', 'sparkconf', 'sparkinstallver='])
+                               ['workspace=', 'token=', 'clusterid=', 'private=', 'localscript=', 'dest=', 'sparktgz=',
+                                'basesparkpomversion=', 'buildprofiles=', 'jarpath', 'sparkconf', 'sparkinstallver='])
 except getopt.GetoptError:
-    print(
-        'run-tests.py -s <workspace> -t <token> -c <clusterid> -p <privatekeyfile> -l <localscript> -d <scriptdestinatino> -z <sparktgz> -v <basesparkpomversion> -b <buildprofiles> -j <jarpath> -f <sparkconf> -i <sparkinstallver>')
+    print('run-tests.py -s <workspace> -t <token> -c <clusterid> -p <privatekeyfile> -l <localscript> '
```
jlowe, line 40: This usage string should be defined once and reused to avoid getting them out of sync.  For example, this usage string doesn't mention `-n <skipstartingclusters>` yet the string used for usage help below does.  This string lists `-j <jarpath>` but that option is missing from the usage string below.

jenkins/databricks/params.py
```diff
@@ -68,11 +73,15 @@
     elif opt in ('-i', '--sparkinstallver'):
         base_spark_version_to_install_databricks_jars = arg
 
+if args:
+    script_args = args
```
jlowe, line 77: The conditional should not be necessary.  If there are no arguments then `args` is set to an empty list.  We can use `args` directly, and `script_args` is unnecessary.

jenkins/databricks/params.py
zhanga5: agree. This PR tries to keep minimal change scope except flake8 errors/warnings and new arguments. I would prefer config file rather than so many parameters on command line but it's out of the scope. Anyway, will address your comments. Thanks

jenkins/databricks/params.py
zhanga5: OK, will remove it

jlowe (approved):

zhanga5:
build

Issue #3374, opened 2021-09-02
[BUG] minor deprecation warnings in a 3.2 shim build


gerashegalov:
**Describe the bug**
```
[WARNING] [Warn] GpuMultiFileReader.scala:87: method getAllStatistics in class FileSystem is deprecated: see corresponding Javadoc for more information.

[WARNING] [Warn] sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala:604: method getDecimalMetadata in class PrimitiveType is deprecated: see corresponding Javadoc for more information.

[WARNING] [Warn] sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/rapids/GpuPartitioningUtils.scala:121: method isDirectory in class FileSystem is deprecated: see corresponding Javadoc for more information.
```

**Steps/Code to reproduce bug**
```
$ mvn clean package -Dbuildver=320 -Dskip -DskipTests -pl integration_tests -am 
```

**Expected behavior**
build free of deprecation warnings

**Environment details (please complete the following information)**
local

**Additional context**
#3323

jlowe:
These new warnings with 3.2 imply that we're not compiling against the Parquet jars that we're explicitly shading and including in the distribution jar.  Seems like there's an issue with the pom where we should be excluding the Parquet version Spark wants so we can use the version we bundle explicitly.

res-life:
@gerashegalov Cannot find the warnings, please double check.

gerashegalov:
After #4071 we no longer allow new warnings to be introduced without explicitly silencing them when necessary as part of the PR under review. Existing warnings are silenced, so this issue is now about looking into reducing the usage of  `@scala.annotation.nowarn`

Issue #4219, opened 2021-11-26
[FEA]Upgrade the hadoop-common hadoop-client  version


res-life:
Our plugin is using hadoop-common:2.7.4 hadoop-client 2.7.4 currently, however Spark 3.0 and 3.2 are using hadoop-common-3.2.0 and hadoop-client-api:3.3.1 separately.
It will be better to upgrade it to hadoop-common-3.2.0 hadoop-client-3.2.0
For version 3.0.1 the dependency path: 
   spark-hive 3.0.1 -> spark-core 3.0.1 -> hadoop-client 2.7.4 -> hadoop-common 2.7.4

jlowe:
The plugin uses whichever hadoop-common and hadoop-client versions Spark uses.  We do not want to depend on a specific Hadoop version, otherwise we would need to pull it into the plugin jar and shade it to avoid conflicting with the version provided by Spark.

Therefore I'm not exactly sure what this issue is requesting.  Outside of the tools module, we do not directly depend on Hadoop jars and therefore cannot upgrade them.  I do not think we should start pulling in explicit Hadoop jar versions per the discussion above.  If the desire to upgrade is related to fixing an issue then I think I'd rather see us try to resolve the issue in the shims rather than pull in Hadoop jars and shading them.

res-life:
OK, got it. It's not related to an issue.

jlowe:
Closing this per the discussion above.

PR #4223, opened 2021-11-26
Fixes a minor deprecation warning

res-life:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3374

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/rapids/GpuPartitioningUtils.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/rapids/GpuPartitioningUtils.scala
@@ -109,9 +109,6 @@ object GpuPartitioningUtils {
    *
    * mainly copied from PartitioningAwareFileIndex.basePaths
    */
-  @scala.annotation.nowarn(
-    "msg=method isDirectory in class FileSystem is deprecated"
-  )
   private def getBasePaths(
     hadoopConf: Configuration,
     basePathOption: Option[Path],
@@ -121,7 +118,7 @@ object GpuPartitioningUtils {
     basePathOption match {
       case Some(userDefinedBasePath) =>
         val fs = userDefinedBasePath.getFileSystem(hadoopConf)
-        if (!fs.isDirectory(userDefinedBasePath)) {
+        if (!fs.getFileStatus(userDefinedBasePath).isDirectory) {
           throw new IllegalArgumentException(s"Option '$BASE_PATH_PARAM' must be a directory")
         }
         val qualifiedBasePath = fs.makeQualified(userDefinedBasePath)


res-life:
**1. getAllStatistics deprecated**
```
[WARNING] [Warn] GpuMultiFileReader.scala:87: 
method getAllStatistics in class FileSystem is deprecated: 
see corresponding Javadoc for more information.
```
This is because of getAllStatistics is deprecated in hadoop-client-api-3.3.1.jar which is depended by spark 3.2.0.
Recommends to use 

see: 
```
https://hadoop.apache.org/docs/r2.8.2/api/org/apache/hadoop/fs/FileSystem.html
```
But getGlobalStorageStatistics does not provide the capacity to fetch thread local statistics like previous version:
```
FileSystem.getAllStatistics.asScala.map(_.getThreadStatistics.getBytesRead).sum
```
So, just leave it as it is. BTW, one issue is filed: https://github.com/NVIDIA/spark-rapids/issues/4219
**2. getDecimalMetadata deprecated**
```
[WARNING] [Warn] sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala:604: 
method getDecimalMetadata in class PrimitiveType is deprecated: 
see corresponding Javadoc for more information.
```
spark 3.0.1 uses parquet-column-1.10.1.jar and spark 3.2.0 uses parquet-column-1.12.1.jar.
parquet-column 1.12.1 deprecated the getDecimalMetadata method and recommends to use "getLogicalTypeAnnotation()".
parquet-column-1.10.1 does not have getLogicalTypeAnnotation.
So if want to remove the warning we must write shim codes separately for the 2 versions.
It's a little heavy, let's keep it.

**3.  isDirectory deprecated**
```
sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/rapids/GpuPartitioningUtils.scala:121:
 method isDirectory in class FileSystem is deprecated: 
see corresponding Javadoc for more information.
```
Fixed.

res-life:
build

GaryShen2008 (approved):

sql-plugin/src/main/scala/org/apache/spark/sql/execution/datasources/rapids/GpuPartitioningUtils.scala
```diff
@@ -121,7 +118,7 @@ object GpuPartitioningUtils {
     basePathOption match {
       case Some(userDefinedBasePath) =>
         val fs = userDefinedBasePath.getFileSystem(hadoopConf)
-        if (!fs.isDirectory(userDefinedBasePath)) {
+        if (!fs.getFileStatus(userDefinedBasePath).isDirectory) {
```
firestarman, line 121: Is it possible to return a `null` from `getFileStatus` ?

res-life: Throws FileNotFoundException, its javadoc:
    FileNotFoundException – when the path does not exist;

Issue #3904, opened 2021-10-25
Update hive UDFs to support the feature of `CPU based UDF`


firestarman:
This is part of https://github.com/NVIDIA/spark-rapids/issues/3855.

And https://github.com/NVIDIA/spark-rapids/pull/3897 adds this kind of support in ScalaUDF.

Issue #1398, opened 2020-12-15
Improve shading scope


jlowe:
**Is your feature request related to a problem? Please describe.**
Currently the build system is shading things like ORC and Hive across the entire SQL plugin, and this triggers some issues with areas of the plugin that need to access Hive classes when interacting with spark-hive Catalyst classes (e.g.: #1393).

**Describe the solution you'd like**
I think it would be cleaner if the portions of the plugin were in a separate Maven submodule that was shaded, then we can avoid shading over the entire SQL plugin code.  Shading would be localized to the code that requires it.

**Describe alternatives you've considered**
If we could get out of the business of shading entirely that would be awesome, but I don't see how we can support multiple versions of Apache Spark that could contain conflicting versions of Parquet/ORC if we try to use those dependencies directly.

res-life:
**Issue**
The hive class packages are relocated in rapids UDF, but these hive classes are not included in the shaded jar;
Hive jars are provided dependencies and shade plugin only includes compiled dependencies. 
For example
```
In HiveSimpleUDF.scala
lazy val function = funcWrapper.createFunction[UDF]()
```
The type of  "function" is org.apache.hadoop.hive.ql.exec.UDF,  If not excluded in aggregator.xml, this line is relocated to
``` 
lazy val function: com.nvidia.shaded.spark.hadoop.hive.exec.UDF = funcWrapper.createFunction[UDF]()
```
The package of UDF in the spark runtime hive-exec jar is still org.apache.hadoop.hive.ql.exec.UDF.
This will cause ClassNotFound.  So some of exclutions are added in aggregator.xml.
The ORC dpends on hive-storage-api and hive-storage-api also has package org.apache.hadoop.hive.
The ORC codes are relocated and shade, but also relocated hive classes that UDF referer to.
So the rapids code involving ORC should be shaded and the rapids code involving UDF shoudl not be shaded.
**Option 1:  Split ORC reader/writer related code to a new submodule.**
Make ORC reader/writer related code shaded.  Make other code of sql-plubin not shaded.
Actually, the creation of new module is not a regular refector.  All the related classes should also be moved.
For example, when moving GpuOrcScanBase, the arm.scala must be moved, and then RapidsBuffer, then RapidsDiskBlockManager, then ShimDiskBlockManager. This will make the code mess.
**Option 2:  Use include instead of exclude in aggregator.xml.**
[...]

PR #4224, opened 2021-11-26
Support row-based Hive UDFs [databricks]

firestarman:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3904


firestarman:
Mark as draft, because needing to try an external app next.

firestarman:
build

firestarman:
build

firestarman:
build

firestarman:
build

firestarman:
build

aggregator/pom.xml
```diff
@@ -100,9 +100,27 @@
                             <pattern>org.apache.hadoop.hive.</pattern>
                             <shadedPattern>${rapids.shade.package}.hadoop.hive.</shadedPattern>
                             <excludes>
+                                <!-- FIXME
+                                    Seems we need to exclude all the public classes being used
+                                    directly and indirectly by plugin to support row-based Hive
+                                    UDF, otherwise it will get a ClassNotFoundException as below:
+                                        E Caused by: java.lang.ClassNotFoundException: com.nvidia.shaded.spark.hadoop.hive.serde2.objectinspector.ObjectInspector
+                                        E     at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
+                                    Or, just simply remove this relocation.
```
jlowe, line 109: We need to shade hive as part of shading ORC (we need to shade it and all of its dependencies, or we risk the hive jar either missing in the Spark distribution at runtime or conflicting with the version that the ORC version we're shading needs).

I think we only need to limit the hive relocations to the hive-storage-api classes.  Unfortunately there isn't a nice package prefix that would select only those classes.  We may need to split up the input/output classes that need ORC (and thus the relocations via shading) from the UDF classes that just need the aggregation part of the shading (i.e.: no relocations).  Separating out the Hive-related UDF classes and/or the ORC input/output classes into a separate Maven project would give us a way to control the shading on a more granular basis.

firestarman: Thanks Jason. Good to know.
Is this change here OK for now ?

firestarman:
build

aggregator/pom.xml
jlowe: I think what's here is quite fragile, but I guess it's OK for now.  However we should prioritize #1398 to fix this in a better way.

sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/GpuHiveOverrides.scala
```diff
@@ -45,59 +45,86 @@ object GpuHiveOverrides {
...
           override def tagExprForGpu(): Unit = {
-            a.function match {
-              case _: RapidsUDF =>
-              case _ =>
-                willNotWorkOnGpu(s"Hive UDF ${a.name} implemented by " +
-                    s"${a.funcWrapper.functionClassName} does not provide a GPU implementation")
+            if (opRapidsFunc.isEmpty && !conf.isCpuBasedUDFEnabled) {
+              willNotWorkOnGpu(s"neither Hive SimpleUDF ${a.name} implemented by " +
+                s"${a.funcWrapper.functionClassName} does not provide a GPU implementation," +
+                s" nor the conf `${RapidsConf.ENABLE_CPU_BASED_UDF.key}` is enabled")
```
jlowe, line 64: I think the `neither` here ends up being a double-negative and confusing.  I suggest dropping it and simply stating what's not provided, e.g.:
```suggestion
              willNotWorkOnGpu(s"Hive SimpleUDF ${a.name} implemented by " +
                s"${a.funcWrapper.functionClassName} does not provide a GPU implementation " +
                s"and CPU-based UDFs are not enabled by `${RapidsConf.ENABLE_CPU_BASED_UDF.key}`")
```

sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/GpuHiveOverrides.scala
```diff
@@ -45,59 +45,86 @@ object GpuHiveOverrides {
...
           override def tagExprForGpu(): Unit = {
-            a.function match {
-              case _: RapidsUDF =>
-              case _ =>
-                willNotWorkOnGpu(s"Hive GenericUDF ${a.name} implemented by " +
-                    s"${a.funcWrapper.functionClassName} does not provide a GPU implementation")
+            if (opRapidsFunc.isEmpty && !conf.isCpuBasedUDFEnabled) {
+              willNotWorkOnGpu(s"neither Hive GenericUDF ${a.name} implemented by " +
+                s"${a.funcWrapper.functionClassName} does not provide a GPU implementation," +
+                s" nor the conf `${RapidsConf.ENABLE_CPU_BASED_UDF.key}` is enabled")
```
jlowe, line 105: Similar double-negative with neither:
```suggestion
              willNotWorkOnGpu(s"Hive GenericUDF ${a.name} implemented by " +
                s"${a.funcWrapper.functionClassName} does not provide a GPU implementation " +
                s"and CPU-based UDFs are not enabled by `${RapidsConf.ENABLE_CPU_BASED_UDF.key}`")
```

jlowe (commented):
Minor suggestion on willNotWork wording but otherwise lgtm.

sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/GpuHiveOverrides.scala
firestarman: Nice catch

sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/GpuHiveOverrides.scala
firestarman: Nice catch

firestarman:
build

jlowe (approved):

firestarman:
build

firestarman:
build

PR #4228, opened 2021-11-29
Fix unconditional nested loop joins on empty tables

### LINKED ISSUE #4188: [BUG] data corruption in GpuBroadcastNestedLoopJoin with empty relations edge case
I am testing a patch for empty relations for the broadcast* joins and ran into an edge case that is a bug in 21.12 where we return nothing when the CPU returns results, or the other way around (we return a result where the CPU returns nothing), this is specific to unconditional joins.

This test can reproduce the issue:

```
@ignore_order(local=True)
@pytest.mark.parametrize('join_type', ['Left', 'Inner', 'LeftSemi', 'LeftAnti', 'Cross'], ids=idfn)
@pytest.mark.parametrize('batch_size', ['100', '1g'], ids=idfn)
def test_right_broadcast_nested_loop_join_without_condition_empty(join_type, batch_size):
    def do_join(spark):
        left, right = create_df(spark, long_gen, 50, 0)
        return left.join(broadcast(right), how=join_type)
    conf = copy_and_update(allow_negative_scale_of_decimal_conf, {'spark.rapids.sql.batchSizeBytes': batch_size})
    assert_gpu_and_cpu_are_equal_collect(do_join, conf=conf)
```

This will fail for BuildRight: Left, LeftSemi and LeftAnti. I think I have a good idea about where to fix for the LeftSemi case anyway. We currently just return the [left relation](https://github.com/NVIDIA/spark-rapids/blob/branch-21.12/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExecBase.scala#L496), whereas spark has other logic here:

```
  private def leftExistenceJoin(
      relation: Broadcast[Array[InternalRow]],
      exists: Boolean): RDD[InternalRow] = {
    assert(buildSide == BuildRight)
    streamed.execute().mapPartitionsInternal { streamedIter =>
      val buildRows = relation.value
      val joinedRow = new JoinedRow

      if (condition.isDefined) {
        streamedIter.filter(l =>
          buildRows.exists(r => boundCondition(joinedRow(l, r))) == exists
        )
      } else if (buildRows.nonEmpty == exists) {
        streamedIter
      } else {
        Iterator.empty
      }
    }
  }
```

I think this also covers LeftAnti, but I haven't tested it.

jlowe:
Fixes #4188.  Unconditional nested loop joins can be treated like a cross join as long as both tables are non-empty, but there needs to be special handling when one of the tables is empty based on the join type.

### DIFF
--- a/integration_tests/src/main/python/join_test.py
+++ b/integration_tests/src/main/python/join_test.py
@@ -98,6 +98,36 @@ def create_ridealong_df(spark, key_data_gen, data_gen, left_length, right_length
             .withColumnRenamed("b", "r_b")
     return left, right
 
+@ignore_order(local=True)
+@pytest.mark.parametrize('join_type', ['Left', 'Inner', 'LeftSemi', 'LeftAnti'], ids=idfn)
+@pytest.mark.parametrize('batch_size', ['100', '1g'], ids=idfn)
+def test_right_broadcast_nested_loop_join_without_condition_empty(join_type, batch_size):
+    def do_join(spark):
+        left, right = create_df(spark, long_gen, 50, 0)
+        return left.join(broadcast(right), how=join_type)
+    conf = copy_and_update(allow_negative_scale_of_decimal_conf, {'spark.rapids.sql.batchSizeBytes': batch_size})
+    assert_gpu_and_cpu_are_equal_collect(do_join, conf=conf)
+
+@ignore_order(local=True)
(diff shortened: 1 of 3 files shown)


jlowe:
build

revans2 (approved):

abellina (approved):

jlowe:
CI script exited suddenly without an obvious failure, trying again.

jlowe:
rerun tests

jlowe:
build

jlowe:
build

PR #4241, opened 2021-11-30
fix bug in parsing regex character classes that start with `^` and contain an unescaped `]`

andygrove:

This bug was exposed by one of the existing fuzzer tests in another issue I was working on.


jlowe (approved):

jlowe:
build

andygrove:
build failed with

```
 [2021-11-30T23:59:20.320Z] - Cast from string to float using hand-picked values *** FAILED ***
[2021-11-30T23:59:20.320Z]   Mismatch casting string [+nAn] to FloatType. CPU: null; GPU: NaN (CastOpSuite.scala:258)
```

andygrove:
build

Issue #4243, opened 2021-12-01
[BUG] test_regexp_replace_null_pattern_fallback[ALLOW_NON_GPU(ProjectExec,RegExpReplace)] failed in databricks


pxLi:
**Describe the bug**
related to https://github.com/NVIDIA/spark-rapids/pull/4153

```
[2021-12-01T00:25:50.253Z] __________________ test_regexp_replace_null_pattern_fallback ___________________

[2021-12-01T00:25:50.253Z] [gw0] linux -- Python 3.8.12 /databricks/conda/envs/cudf-udf/bin/python

[2021-12-01T00:25:50.253Z] 

[2021-12-01T00:25:50.253Z]     @allow_non_gpu('ProjectExec', 'RegExpReplace')

[2021-12-01T00:25:50.253Z]     def test_regexp_replace_null_pattern_fallback():

[2021-12-01T00:25:50.253Z]         gen = mk_str_gen('[abcd]{0,3}')

[2021-12-01T00:25:50.253Z]         # Spark translates `NULL` to `CAST(NULL as STRING)` and we only support

[2021-12-01T00:25:50.253Z]         # literal expressions for the regex pattern

[2021-12-01T00:25:50.253Z] >       assert_gpu_fallback_collect(

[2021-12-01T00:25:50.253Z]                 lambda spark: unary_op_df(spark, gen).selectExpr(

[2021-12-01T00:25:50.253Z]                     'regexp_replace(a, NULL, "A")'),

[2021-12-01T00:25:50.254Z]                 'RegExpReplace',

[2021-12-01T00:25:50.254Z]                 conf={'spark.rapids.sql.expression.RegExpReplace': 'true'})

[2021-12-01T00:25:50.254Z] 

[2021-12-01T00:25:50.254Z] ../../src/main/python/string_test.py:491: 

[2021-12-01T00:25:50.254Z] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[2021-12-01T00:25:50.254Z] ../../src/main/python/asserts.py:377: in assert_gpu_fallback_collect

[2021-12-01T00:25:50.254Z]     jvm.com.nvidia.spark.rapids.ExecutionPlanCaptureCallback.assertDidFallBack(gpu_df._jdf, cpu_fallback_class_name)

[...]

PR #4251, opened 2021-12-01
Skip test_regexp_replace_null_pattern_fallback on Databricks [databricks]

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/4243

### DIFF
--- a/integration_tests/src/main/python/string_test.py
+++ b/integration_tests/src/main/python/string_test.py
@@ -220,7 +220,7 @@ def test_concat_ws_sql_col_sep():
                 'concat_ws(c, b, a, cast(null as string)) from concat_ws_table')
 
 
-@pytest.mark.xfail(condition=is_databricks_runtime(),
+@pytest.mark.skipif(is_databricks_runtime(),
     reason='Databricks optimizes out concat_ws call in this case')
 @allow_non_gpu('ProjectExec', 'Alias', 'ConcatWs')
 def test_concat_ws_sql_col_sep_only_sep_specified():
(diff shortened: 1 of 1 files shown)


tgravescs:
if we are disabling this is it tracked by another issue to turn it back on?

tgravescs:
oh never mind, after looking at code looks like you are saying databricks optimizes that out, perhaps update description to say the same

tgravescs (dismissed):

tgravescs:
build

integration_tests/src/main/python/string_test.py
```diff
@@ -483,11 +483,14 @@ def test_regexp_replace():
                 'regexp_replace(a, "a|b|c", "A")'),
             conf={'spark.rapids.sql.expression.RegExpReplace': 'true'})
 
+@pytest.mark.xfail(condition=is_databricks_runtime(),
```
jlowe, line 486: Do we want to xfail this or simply skip it?  The point of xfailing is to track tests that we eventually expect to pass.  I think skip is more appropriate for this one.

andygrove: Skipping it makes sense to me. Should I also update `test_concat_ws_sql_col_sep_only_sep_specified` in the same file to skip instead of xfail? (that's where I copied the pattern from).

jlowe: Yes, any test that we are never expecting to pass because it's an invalid premise for testing in a particular environment should be skipped rather than xfailed.  We should reserve xfail for tests we think should pass at some point, as they essentially serve as TODO's in the test output.

jlowe:
build

Issue #4229, opened 2021-11-29
[BUG] regexp_replace `[^a]` has different behavior between CPU and GPU for multiline strings


andygrove:
**Describe the bug**

With the pattern `[^a]` (meaning match any character except for 'a'), cuDF does not match newline characters but Java/Spark does, resulting in different behavior as demonstrated by these code snippets.

## Spark/Java behavior

```scala
scala> import java.util.regex._
scala> val p = Pattern.compile("[^a]")
scala> p.matcher("a\nb\nc").replaceAll("1")
res0: String = a1111

```

## cuDF behavior

```python
>>> import cudf
>>> cudf.Series(['a\nb\nc']).str.replace('[^a]', '1', regex=True)
0    a\n1\n1
```

**Steps/Code to reproduce bug**

Add the following integration test to string_test.py

```python
def test_regexp_replace_multiline():
    gen = mk_str_gen('[abcd]{0,3}\n[abcd]{0,3}')
    assert_gpu_and_cpu_are_equal_collect(
            lambda spark: unary_op_df(spark, gen).selectExpr(
                'regexp_replace(a, "([^a])|([^b])", "1")'),
            conf={'spark.rapids.sql.expression.RegExpReplace': 'true'})
```

**Expected behavior**
We either need to document the difference in the compatibility guide or we need to find a way to make the behavior consistent.

**Environment details (please complete the following information)**
N/A

**Additional context**
None

PR #4255, opened 2021-12-01
Make regexp pattern `[^a]` consistent with Spark for multiline strings

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/4229

The following documentation from this PR explains the transpiler change that makes us consistent with CPU for patterns such as `[^a]`.

```
// There are differences between cuDF and Java handling of newlines
// for negative character matches. The expression `[^a]` will match
// `\r` and `\n` in Java but not in cuDF, so we replace `[^a]` with
// `(?:[\r\n]|[^a])`. We also have to take into account whether any
// newline characters are included in the character range.
//
// Examples:
//
// `[^a]`     => `(?:[\r\n]|[^a])`
// `[^a\r]`   => `(?:[\n]|[^a])`
// `[^a\n]`   => `(?:[\r]|[^a])`
// `[^a\r\n]` => `[^a]`
```

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
@@ -169,11 +169,9 @@ class RegexParser(pattern: String) {
               throw new RegexUnsupportedException(
                 s"unexpected EOF while parsing escaped character", Some(pos))
             case Some(ch) =>
-              ch match {
-                case '\\' | '^' | '-' | ']' | '+' =>
-                  // escaped metacharacter within character class
-                  characterClass.appendEscaped(consumeExpected(ch))
-              }
+              // typically an escaped metacharacter ('\\' | '^' | '-' | ']' | '+')
+              // within the character class, but could be any escaped character
+              characterClass.appendEscaped(consumeExpected(ch))
           }
         case '\u0000' =>
           throw new RegexUnsupportedException(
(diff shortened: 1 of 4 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -494,7 +494,40 @@ class CudfRegexTranspiler(replace: Boolean) {
         }
         val components: Seq[RegexCharacterClassComponent] = characters
           .map(x => rewrite(x).asInstanceOf[RegexCharacterClassComponent])
-        RegexCharacterClass(negated, ListBuffer(components: _*))
+
+        if (negated) {
+          // There are differences between cuDF and Java handling of newlines
+          // for negative character matches. The expression `[^a]` will match
+          // `\r` and `\n` in Java but not in cuDF, so we replace `[^a]` with
+          // `(?:[\r\n]|[^a])`. We also have to take into account whether any
```
jlowe, line 502: Do we need to worry about "a" being (possibly escaped) carriage-return or newline characters in the example for `[^a]`?

andygrove: Nice catch... I added some more tests and they did expose a bug with the input pattern `[^\r\n]`.

andygrove:
build

jlowe (approved):

jlowe:
build

PR #4256, opened 2021-12-01
Allow returning an EmptyHashedRelation when a broadcast result is empty [databricks]

### LINKED ISSUE #4134: [FEA] Allow `EliminateJoinToEmptyRelation` in `GpuBroadcastExchangeExec`
AQE has an optimization rule: `EliminateJoinToEmptyRelation` that we are seeing in NDS q16. The query on the GPU takes roughly 40s to run in one of our test environments, but it takes the CPU 14s to run. This is because most of the query goes away and is replaced by a `LocalScan <empty>` given `EliminateJoinToEmptyRelation` on a broadcast exchange that produces 0 rows.

AQE performs this optimization for inner join, left semi, or a specific anti join (`ExtraSingleColumnNullAwareAntiJoin`). It looks at the children of the join and actively resolves the `broadcast.relationFuture` member. If it returns `EmptyHashedRelation` it goes ahead and eliminates this join.

The `relationFuture` is defined in the plugin, but we do not emit an `EmptyHashedRelation` when we have 0 rows in the broadcast. If I make this change, AQE eliminates most of q16 and now it executes in ~7.5 seconds (or roughly 2x faster than the CPU).

What I do not know yet is what else can break with this change. I did run all of NDS and none of the queries failed, but that's just one example. I think we need to understand better if `EmptyHashedRelation` is OK to have around in a GPU plan.

abellina:

Closes #4134.

The PR allows the broadcast exchange to produce an `EmptyHashedRelation` or an empty array in the case of the identity broadcast, in order for AQE's `EliminateJoinToEmptyRelation` rule to be able to optimize the plan. This changes q16 massively in the way we run things, but it lets us match what the CPU does.

In terms of performance, I ran this at 3TB and Q16 is now ~12 seconds which is slightly faster than the CPU (this is 3.5x faster than what we had before on the GPU). I don't see regressions with other queries.

This PR enables `isFoldableNonLitAllowed` for `UnaryExprMeta` so that expressions like `cast(null as bigint)` can be handled. These casts show up given an empty projection due to the AQE rule to remove the join. That said, `ConstantFolding` does not re-execute as part of AQE, so they are left in the plan. Both tests I have added will generate plans with these for AQE.


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastHelper.scala
```diff
@@ -0,0 +1,88 @@
...
+                        broadcastPlan: SparkPlan): ColumnarBatch = {
+    val broadcastRelationValue = broadcastRelation.value
+    broadcastRelationValue match {
+      case broadcastBatch: SerializeConcatHostBuffersDeserializeBatch =>
+        val builtBatch = broadcastBatch.batch
+        GpuColumnVector.incRefCounts(builtBatch)
+        builtBatch
+      case EmptyHashedRelation =>
+        GpuColumnVector.emptyBatch(broadcastPlan.schema)
+      case identity: Array[Any] if identity.length == 0 =>
```
jlowe, line 51: I'm confused why we need this logic.  We support broadcast nested loop joins today yet don't have code that is explicitly checking for identity vs. hash broadcasts.  Everything is a SerializeConcatHostBufferDeserializeBatch which can also handle empty tables and row-count-only tables.  Do we not have control over the driver-side code that is setting up the broadcast or something?

abellina: Because of this: https://github.com/NVIDIA/spark-rapids/pull/4256/files#diff-7ca82be5615724f3ba4376b82be1c8a34e3fb5df34a3dedfa9a21eb3e6e54b97R321. So we are making it not return a `SerializedConcatHostBufferDeserializeBatch` on purpose to allow AQE's optimization to work, but that means we need to handle the special cases later.

abellina:
Ok `EliminateJoinToEmptyRelation` changed in Spark 3.2 to `EliminateUnnecessaryJoin` and now it's looking at `getRowCount` from the `Statistics` object in the ShuffleExchangeExec it seems. This caught me off guard, I need to spend time on Spark 3.2 to understand if this is going to cause problems or how they are handling it now.

revans2 (dismissed):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
```diff
@@ -303,28 +303,36 @@ abstract class GpuBroadcastExchangeExecBase(
...
-            val numRows = batch.numRows
-            checkRowLimit(numRows)
-            numOutputBatches += 1
-            numOutputRows += numRows
+            var dataSize = 0L
+            val broadcastResult =
+              withResource(new NvtxWithMetrics("broadcast collect", NvtxColor.GREEN,
+                collectTime)) { _ =>
+                val childRdd = child.executeColumnar()
+                val data = childRdd.map(cb => try {
```
jlowe, line 311: Nit: since we're in the neighborhood, seems like a good candidate for `withResource`.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
```diff
@@ -303,28 +303,36 @@ abstract class GpuBroadcastExchangeExecBase(
...
+                  new SerializeBatchDeserializeHostBuffer(cb)
+                } finally {
+                  cb.close()
+                })
+                val d = data.collect()
+                if (d.length == 0) {
```
jlowe, line 317:
```suggestion
                if (d.isEmpty) {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
```diff
@@ -303,28 +303,36 @@ abstract class GpuBroadcastExchangeExecBase(
...
+                  // This call for `HashedRelationBroadcastMode` produces
+                  // `EmptyHashedRelation` allowing the AQE rule `EliminateJoinToEmptyRelation` to
+                  // optimize out our parent join given that this is a empty broadcast result.
+                  mode.transform(Iterator.empty, None)
```
jlowe, line 321: Curious, why pass `None` here when we know the row count?  Seems like the single-parameter `mode.transform` would be the typical path to use here?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
```diff
@@ -303,28 +303,36 @@ abstract class GpuBroadcastExchangeExecBase(
...
+                }
+              }
 
             withResource(new NvtxWithMetrics("broadcast build", NvtxColor.DARK_GREEN,
                 buildTime)) { _ =>
               // we only support hashjoin so this is a noop
               // val relation = mode.transform(input, Some(numRows))
-              val dataSize = batch.dataSize
-
               gpuLongMetric("dataSize") += dataSize
```
jlowe, line 336: Isn't dataSize zero at this point?  It was initialized to zero but I don't see it updated anywhere.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastHelper.scala
```diff
@@ -0,0 +1,88 @@
...
+   * @param broadcastRelation - the broadcast as produced by a broadcast exchange
+   * @return number of rows for a batch received, or 0 if it's an empty relation
+   */
+  def getBroadcastBatchNumRows(broadcastRelation: Broadcast[Any]): Int = {
+    val broadcastRelationValue = broadcastRelation.value
+    broadcastRelationValue match {
+      case broadcastBatch: SerializeConcatHostBuffersDeserializeBatch =>
+        broadcastBatch.batch.numRows()
+      case EmptyHashedRelation => 0
+      case identity: Array[Any] if identity.length == 0 => 0
```
jlowe, line 82:
```suggestion
      case identity: Array[Any] if identity.isEmpty => 0
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastHelper.scala
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * If a broadcast result is unexpected we throw, but at this moment other
+   * cases are not known, so this is a defensive measure.
+   *
+   * @param broadcastRelation - the broadcast as produced by a broadcast exchange
+   * @return number of rows for a batch received, or 0 if it's an empty relation
+   */
+  def getBroadcastBatchNumRows(broadcastRelation: Broadcast[Any]): Int = {
+    val broadcastRelationValue = broadcastRelation.value
+    broadcastRelationValue match {
```
jlowe, line 78:
```suggestion
    broadcastRelation.value match {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExecBase.scala
```diff
@@ -435,23 +435,26 @@ abstract class GpuBroadcastNestedLoopJoinExecBase(
...
+      broadcastRelation: Broadcast[Any],
       buildTime: GpuMetric,
       buildDataSize: GpuMetric): ColumnarBatch = {
     withResource(new NvtxWithMetrics("build join table", NvtxColor.GREEN, buildTime)) { _ =>
-      val ret = broadcastRelation.value.batch
-      buildDataSize += GpuColumnVector.getTotalDeviceMemoryUsed(ret)
-      GpuColumnVector.incRefCounts(ret)
+      withResource(GpuBroadcastHelper.getBroadcastBatch(
+          broadcastRelation, broadcast)) { builtBatch =>
+        GpuColumnVector.incRefCounts(builtBatch)
```
jlowe, line 444: getBroadcastBatch is already incrementing the ref counts, so seems like we could avoid some unnecessary inc/decs here.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
abellina: So I tried doing this, but I got a serialization error. It's a little cryptic on what is going on. I tried both `withResource` with a batch specifically (i.e. `withResource(cb)` or with the placeholder `withResource(_)` thinking the placeholder was doing something odd type-wise):

```
E                   Serialization stack:
E                   	- object not serializable (class: org.apache.spark.sql.rapids.execution.GpuBroadcastExchangeExecBase$$anon$1, value: org.apache.spark.sql.rapids.execution.GpuBroadcastExchangeExecBase$$anon$1@7b0c20b2)
E                   	- element of array (index: 0)
E                   	- array (class [Ljava.lang.Object;, size 1)
E                   	- field (class: java.lang.invoke.SerializedLambda, name: capturedArgs, type: class [Ljava.lang.Object;)
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
abellina: You are right, just fixed.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastHelper.scala
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * The broadcast relation may or may not contain any data, so we special case
+   * the empty relation case (hash or identity depending on the type of join).
+   *
+   * If a broadcast result is unexpected we throw, but at this moment other
+   * cases are not known, so this is a defensive measure.
+   *
+   * @param broadcastRelation - the broadcast as produced by a broadcast exchange
+   * @param broadcastPlan - the SparkPlan to use to obtain the schema for the broadcast
+   *                      batch
```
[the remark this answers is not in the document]
abellina: This should be fixed now.

abellina:
build

integration_tests/src/main/python/join_test.py
```diff
@@ -101,31 +101,40 @@ def create_ridealong_df(spark, key_data_gen, data_gen, left_length, right_length
 @ignore_order(local=True)
 @pytest.mark.parametrize('join_type', ['Left', 'Inner', 'LeftSemi', 'LeftAnti'], ids=idfn)
 @pytest.mark.parametrize('batch_size', ['100', '1g'], ids=idfn)
-def test_right_broadcast_nested_loop_join_without_condition_empty(join_type, batch_size):
+@pytest.mark.parametrize('enable_aqe', [False, True], ids=idfn)
+def test_right_broadcast_nested_loop_join_without_condition_empty(join_type, batch_size, enable_aqe):
```
revans2, line 105: nit: perhaps we can have a new test instead of reusing the existing tests? Specifically I don't see a lot of value in testing a join with a batch size of 100 and aqe on vs off.  The batch size is there to verify that the algorithm can handle lots of batches and that algorithm is not going to change internally based off of AQE.  Also the batch size of 100 is really slow to run.  If fact if you want a `test_right_broadcast_nested_loop_join_without_condition_empty_small_batch` test and remove the batch_size parameter from this one I would be fine with that. Because it would isolate the slowness to one specific test that is simpler to skip/separate.

This comment goes for all of the tests that you updated.

sql-plugin/src/main/311+-nondb/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
```diff
@@ -149,16 +149,16 @@ case class GpuBroadcastHashJoinExec(
...
 
     val rdd = streamedPlan.executeColumnar()
     rdd.mapPartitions { it =>
       val stIt = new CollectTimeIterator("broadcast join stream", it, streamTime)
-      val builtBatch = broadcastRelation.value.batch
-      GpuColumnVector.extractBases(builtBatch).foreach(_.noWarnLeakExpected())
-      doJoin(builtBatch, stIt, targetSize, spillCallback,
-        numOutputRows, joinOutputRows, numOutputBatches, opTime, joinTime)
+      withResource(
+          GpuBroadcastHelper.getBroadcastBatch(broadcastRelation, buildPlan.schema)) { builtBatch =>
```
revans2, line 158: Can you please cache the schema outside of the mapPartitions? I don't want to have to serialize/deserialize the entire `buildPlan` if all we need from it is the schema. This applies to all of the shims GpuBroadcastHashJoinExec instances.

abellina:
Note that: https://github.com/NVIDIA/spark-rapids/pull/4256/commits/f62167cd0818c3d7b59938dbc30912d855906ef8 adds `isFoldableNonLitAllowed` to the `UnaryExprMeta`. What @revans2 explained was (if I understand correctly) that an AQE rule that removes a join is likely running after constant folding, and therefore we are left with some unary expressions that didn't get folded.

That said I am not entirely sure yet of the order of things, so I am not 100% there yet.

abellina:
> That said I am not entirely sure yet of the order of things, so I am not 100% there yet.

I've been reading more about this and I think it makes sense now. Yes `ConstantFolding` isn't in the path after AQE removes the join because the `ConstantFolding` optimization happens for the basic logical plan optimizer (which happens early on), and not the `AQEOptimizer`. As far as I understand, once the plan is wrapped in an `AdaptiveSparkPlanExec` via `InsertAdaptiveSparkPlan` (from `org.apache.spark.sql.executions.QueryExecution.preparations`), the optimizer that includes `ConstantFolding` is not executed again, instead the optimizers used are the AQE optimizer (which only worries about Propagate Empty Relations,Dynamic Join Selection), and the canonicalizer  (CleanExpressions). @andygrove does this make sense?

Given the optimization, we are looking at the `logicalLink` of the plan wrapped in the adaptive exec, and then producing a new projection, which in  my case included a `cast(null as bigint)` (aka something that should have gotten folded to a literal). Nowadays, support for this should be available on the GPU, but we don't have test cases for it, so it is disabled by default.

abellina:
build

abellina:
build

abellina:
build

abellina:
This is broken in databricks:
```
  override def broadcastModeTransform(mode: BroadcastMode, rows: Array[InternalRow]): Any =
    mode.transform(rows, TaskContext.get.taskMemoryManager())
```
I was expecting to be able to transform the `BroadcastMode`, as spark does (https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/exchange/BroadcastExchangeExec.scala#L148), but it seems we need the task context.

I am looking into it further.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastExchangeExec.scala
```diff
@@ -303,28 +303,37 @@ abstract class GpuBroadcastExchangeExecBase(
...
+                val childRdd = child.executeColumnar()
+                val data = childRdd.map(cb => try {
+                  new SerializeBatchDeserializeHostBuffer(cb)
+                } finally {
+                  cb.close()
+                })
+
+                val d = data.collect()
+                val emptyRelation: Option[Any] = if (d.isEmpty) {
+
```
jlowe, line 319:
```suggestion
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastHelper.scala
```diff
@@ -0,0 +1,78 @@
...
+   *
+   * @param broadcastRelation - the broadcast as produced by a broadcast exchange
+   * @return number of rows for a batch received, or 0 if it's an empty relation
+   */
+  def getBroadcastBatchNumRows(broadcastRelation: Broadcast[Any]): Int = {
+    val broadcastRelationValue = broadcastRelation.value
+    broadcastRelationValue match {
+      case broadcastBatch: SerializeConcatHostBuffersDeserializeBatch =>
+        broadcastBatch.batch.numRows()
+      case _ if ShimLoader.getSparkShims.isEmptyRelation(broadcastRelationValue) => 0
```
jlowe, line 72: Nit: `broadcastRelationValue` is a name as long as the expression it is capturing yet seems unnecessary.
```suggestion
    broadcastRelation.value match {
      case broadcastBatch: SerializeConcatHostBuffersDeserializeBatch =>
        broadcastBatch.batch.numRows()
      case v if ShimLoader.getSparkShims.isEmptyRelation(v) => 0
```

abellina:
build

abellina:
Thanks @jlowe, I believe I have addressed the comments

abellina:
@revans2 this is ready for another look when you get a chance.

revans2 (approved):

andygrove:
> I've been reading more about this and I think it makes sense now. Yes `ConstantFolding` isn't in the path after AQE removes the join because the `ConstantFolding` optimization happens for the basic logical plan optimizer (which happens early on), and not the `AQEOptimizer`. As far as I understand, once the plan is wrapped in an `AdaptiveSparkPlanExec` via `InsertAdaptiveSparkPlan` (from `org.apache.spark.sql.executions.QueryExecution.preparations`), the optimizer that includes `ConstantFolding` is not executed again, instead the optimizers used are the AQE optimizer (which only worries about Propagate Empty Relations,Dynamic Join Selection), and the canonicalizer (CleanExpressions). @andygrove does this make sense?

Yes, that is correct. I ran some tests to confirm this.

integration_tests/src/main/python/join_test.py
```diff
@@ -128,6 +129,33 @@ def do_join(spark):
...
+                    reason="Disabled for databricks because of lack of AQE support, and "
+                           "differences in BroadcastMode.transform")
+@pytest.mark.parametrize('join_type', ['Left', 'Right', 'Inner', 'LeftSemi', 'LeftAnti'], ids=idfn)
+def test_empty_broadcast_hash_join(join_type):
+    def do_join(spark):
+        left, right = create_df(spark, long_gen, 50, 0)
+        return left.join(right.hint("broadcast"), left.a == right.r_a, join_type)
+    conf = copy_and_update(allow_negative_scale_of_decimal_conf,
+            {'spark.sql.adaptive.enabled': 'true'})
+    assert_cpu_and_gpu_are_equal_collect_with_capture(do_join, conf = conf)
```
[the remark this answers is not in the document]
abellina: nice catch, I had some asserts here but I removed those since they seemed too brittle, I'll go back to the non-capturing one if that's OK

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastHelper.scala
```diff
@@ -0,0 +1,76 @@
...
+   * cases are not known, so this is a defensive measure.
+   *
+   * @param broadcastRelation - the broadcast as produced by a broadcast exchange
+   * @return number of rows for a batch received, or 0 if it's an empty relation
+   */
+  def getBroadcastBatchNumRows(broadcastRelation: Broadcast[Any]): Int = {
+    broadcastRelation.value match {
+      case broadcastBatch: SerializeConcatHostBuffersDeserializeBatch =>
+        broadcastBatch.batch.numRows()
+      case _ if ShimLoader.getSparkShims.isEmptyRelation(broadcastRelation.value) => 0
```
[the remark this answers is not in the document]
abellina: thanks, this should be taken care of now.

abellina:
build

jlowe (approved):

PR #4235, opened 2021-11-30
Commonize v2 shim [databricks]

res-life:
Rename SparkBaseShims to Spark30XShims, delete unused 312db sources directory in sql-plugin.
Fixes #3660 

Signed-off-by: Chong Gao <res_life@163.com>

PR #4257, opened 2021-12-01
Implement getShuffleRDD and fixup mismatched output types on shuffle reuse [databricks]

### LINKED ISSUE #4216: [BUG] AQE Crashing Spark RAPIDS when using filter() and union()
Hey,

the [**following code**](https://pastebin.com/t5VQBkgF) crashes Spark RAPIDS.
The output of the job looks, including the error messages looks like [**this**](https://pastebin.com/ey1qFqky).
I feel that this is a bug due to it working fine with RAPIDS disabled.

Example input file that reproduces of the bug can be found [**here**](https://filebin.net/0ak30mmq7b6vr2yt).

The following are used:
```
- Standalone single node
- Spark 3.2.0
- RAPIDS v21.10.0
- NVIDIA Tesla V100 SXM2 32 GB
```

The configurations used to launch the job were as follows:
```
--conf spark.executor.extraClassPath=${SPARK_CUDF_JAR}:${SPARK_RAPIDS_PLUGIN_JAR}
--conf spark.driver.extraClassPath=${SPARK_CUDF_JAR}:${SPARK_RAPIDS_PLUGIN_JAR}        
--conf spark.rapids.sql.concurrentGpuTasks=3        
--driver-memory 24G     
--conf spark.executor.memory=250G        
--conf spark.executor.cores=5        
--conf spark.task.cpus=1        
--conf spark.executor.resource.gpu.amount=1        
--conf spark.task.resource.gpu.amount=1        
--conf spark.rapids.memory.pinnedPool.size=4G        
--conf spark.locality.wait=0s        
--conf spark.sql.files.maxPartitionBytes=512m        
--conf spark.plugins=com.nvidia.spark.SQLPlugin        
--conf spark.sql.shuffle.partitions=32        
```
**Important notes regarding crashes**
The crash can be avoided, resulting in a succesful job, by making either one of the following changes to the code/configuration:

- By disabling AQE by adding:
```
--conf spark.sql.adaptive.enabled=false
```
to the configuration

- Removing the ``filter`` (Although this does not give the expected outcome regards to the output) operation from line 16 [**of the code**](https://pastebin.com/t5VQBkgF):
```
val cDF = bDF.select(col("mid"))
```

- By adding a ``cache`` operation to the line 16 [**of the code**](https://pastebin.com/t5VQBkgF):
```
val cDF = bDF.filter(col("left") === 0).select(col("mid")).cache()
```
There may be more things that result into a succesful job, these are just some things that I personally found that impacted whether the job results into a success or a failure.
The code also works as expected with RAPIDS disabled, resulting in a succesful job without making any of the changes mentioned.

Feel free to ask me more details incase I missed something.
Thanks!

jlowe:
Fixes #4216.

This implements the getShuffleRDD interface for GpuShuffleExchangeExec and fixes an issue with ReusedExchangeExec caching CPU aggregation buffer types before they're updated with GPU aggregation buffer types.  The plan is searched for instances where the ReusedShuffleExec output types do not match the underlying GPU shuffle output types, fixing the output types of ReusedShuffleExec to match the GPU shuffle.

### DIFF
--- a/sql-plugin/src/main/301+-nondb/scala/com/nvidia/spark/rapids/shims/v2/AQEUtils.scala
+++ b/sql-plugin/src/main/301+-nondb/scala/com/nvidia/spark/rapids/shims/v2/AQEUtils.scala
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.nvidia.spark.rapids.shims.v2
+
+import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.execution.adaptive.{QueryStageExec, ShuffleQueryStageExec}
+
+/** Utility methods for manipulating Catalyst classes involved in Adaptive Query Execution */
+object AQEUtils {
+  /** Return a new QueryStageExec reuse instance with updated output attributes */
+  def newReuseInstance(sqse: ShuffleQueryStageExec, newOutput: Seq[Attribute]): QueryStageExec = {
+    sqse.newReuseInstance(sqse.id, newOutput)
+  }
+}
--- a/integration_tests/src/main/python/repart_test.py
+++ b/integration_tests/src/main/python/repart_test.py
@@ -267,3 +267,14 @@ def test_hash_repartition_exact(gen, num_parts):
                     .withColumn('hashed', f.hash(*part_on))\
                     .selectExpr('*', 'pmod(hashed, {})'.format(num_parts)),
             conf = allow_negative_scale_of_decimal_conf)
+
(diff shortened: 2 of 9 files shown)


jlowe:
build

revans2 (approved):

abellina (approved):

jlowe:
Databricks 9.1 build failed due to access restrictions on `newReuseInstance`.  Added a shim for this and my best guess at a workaround for it.

jlowe:
build

sql-plugin/src/main/301+-nondb/scala/com/nvidia/spark/rapids/shims/v2/AQEUtils.scala
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package com.nvidia.spark.rapids.shims.v2
+
+import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.execution.adaptive.{QueryStageExec, ShuffleQueryStageExec}
+
+/** Utility methods for manipulating Catalyst classes involved in Adaptive Query Execution */
+object AQEUtils {
```
gerashegalov, line 23: same as the db version? hopefully this will be combined after the commonizing PR #4235

gerashegalov (approved):
LGTM

sql-plugin/src/main/301+-nondb/scala/com/nvidia/spark/rapids/shims/v2/AQEUtils.scala
jlowe: I don't think #4235 is going to address this.  This change is only needed by spark312db, and there's no source directory for "everything except spark312db."  I didn't bother to create one for this since the code is so small.

revans2 (approved):

Issue #4246, opened 2021-12-01
[BUG] Regression in CastOpSuite due to cuDF change in parsing NaN


andygrove:
**Describe the bug**

cuDF commit https://github.com/rapidsai/cudf/commit/69d576543b5414372f36d02a189a7217d3bb8006 appears to have caused a regression in CastOpSuite because cuDF now supports `+` and `-` preceding `Nan` and `Inf`. There are other changes that impact the plugin as well, such as supporting mixed case (e.g. `nAn`) and also supporting the word `Infinity` in full.

```
- Cast from string to float using hand-picked values *** FAILED ***
  Mismatch casting string [+nAn] to FloatType. CPU: null; GPU: NaN (CastOpSuite.scala:258)
```

**Steps/Code to reproduce bug**
Run test using cuDF that includes https://github.com/rapidsai/cudf/commit/69d576543b5414372f36d02a189a7217d3bb8006

**Expected behavior**
Test should pass

**Environment details (please complete the following information)**
N/A

**Additional context**
N/A

PR #4259, opened 2021-12-02
fix regression in cast from string to float that caused signed NaN to be considered valid

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/4246

Changes in this PR:

- Remove transformations that are no longer necessary now that cuDF is more flexible when parsing `NaN` and `Inf` (case insensitive, supports `+` and `-`,  supports both `Inf` and `Infinity`)
- Fixes a bug in the `VALID_FLOAT_REGEX` to reject signed NaN values (`+NaN`, `-NaN`) which are now supported by cuDF but not by Spark
- Reinstate unit test


andygrove:
build

jlowe (approved):

Issue #4204, opened 2021-11-24
casting double to string does not match Spark


HaoYang670:
**Describe the bug**
I tried to cast `5.0e-10` to string. On Spark 3.2, I got `"5.0E-10"`; on spark-rapids I got `"5.0e-10"`

**Steps/Code to reproduce bug**
Spark result:
```
scala> val schema = StructType(Array(StructField("a", DoubleType)))
schema: org.apache.spark.sql.types.StructType = StructType(StructField(a,DoubleType,true))

scala> val df0 = spark.createDataFrame(spark.sparkContext.parallelize(data0), schema)
<console>:29: error: not found: value data0
       val df0 = spark.createDataFrame(spark.sparkContext.parallelize(data0), schema)
                                                                      ^

scala> val data0 = Seq(Row(5e-10))
data0: Seq[org.apache.spark.sql.Row] = List([5.0E-10])

scala> val df0 = spark.createDataFrame(spark.sparkContext.parallelize(data0), schema)
df0: org.apache.spark.sql.DataFrame = [a: double]

scala> df0.show
+-------+
|      a|
+-------+
|5.0E-10|
+-------+
```
spark-rapids result:
```
scala> val schema = StructType(Array(StructField("a", DoubleType)))
schema: org.apache.spark.sql.types.StructType = StructType(StructField(a,DoubleType,true))

scala> val data0 = Seq(Row(5e-10))
data0: Seq[org.apache.spark.sql.Row] = List([5.0E-10])

scala> val df0 = spark.createDataFrame(spark.sparkContext.parallelize(data0), schema)
df0: org.apache.spark.sql.DataFrame = [a: double]

scala> df0.sqlContext.setConf("spark.rapids.sql.castFloatToString.enabled", "true")

scala> df0.show
21/11/24 16:29:40 WARN GpuOverrides: 
!Exec <CollectLimitExec> cannot run on GPU because the Exec CollectLimitExec has been disabled, and is disabled by default because Collect Limit replacement can be slower on the GPU, if huge number of rows in a batch it could help by limiting the number of rows transferred from GPU to CPU. Set spark.rapids.sql.exec.CollectLimitExec to true if you wish to enable it
  @Partitioning <SinglePartition$> could run on GPU
  *Exec <ProjectExec> will run on GPU
    *Expression <Alias> cast(a#1 as string) AS a#4 will run on GPU
      *Expression <Cast> cast(a#1 as string) will run on GPU
[...]

PR #4270, opened 2021-12-02
Document exponent differences when casting floating point to string [skip ci]

jlowe:
Relates to #4204.  Updates the compatibility documentation to describe the differences in the exponent when producing a string from floating point.


revans2 (approved):

jlowe:
build

PR #4272, opened 2021-12-02
Full support for SUM overflow detection on decimal [databricks]

### LINKED ISSUE #3944: [FEA] Full support for sum with overflow on Decimal 128
I have a bit of a crazy idea.  

Background: Sum in Spark also lets us do average. So this can end up applying a little more generally.  One of the main reasons we cannot full support sum on all decimal128 values is because Spark will do the sum as a BigDecimal value. The at different points in the process it will check if hat BigDecimal is too large (overflow checking). This means that no matter how many values are in an aggregation Spark is guaranteed to be able to detect the overflow and do the right thing*. Except Spark has this optimization where if the output size is small enough to fit in a Long then it will do the aggregation as a long instead.  This speeds things up a lot and everyone is happy. However this means that there is a range of values where Spark can no longer detect overflow.  It would take just over 97-million max or min values being summed to overflow. So generally we have taken the approach that if we can detect overflow for 100-million values then we can run it on the GPU without issues. That results in us not being able to detect overflow on anything with a precision higher than 28, so we fall back to the CPU, max precision of decimal-128 is 38.

So in the worst case we would need to be able to do a sum with a precision of 49 to detect the overflow on a Decimal value with a precision of 38.  But we should be able to do that if we split the sum into two separate parts, each of which can handle doing the sum on 100-million entries without overflowing.

If we want to do a sum on a value with a precision of 29. We can split this into two separate sums.  One with a precision of 1 (higher order data) and another with a precision of 28 (lower order data). Then we add the results back together at the end (checking for overflow on each sub-part first). Similarly if we wanted to do a sum on a value with a precision of 38. We would split it into a sum with a precision of 10 and another with a precision of 28. Then add them back together at the end.

To get the higher order data we cudf cast the input to a (scale - 28) from the Spark definiton of scale, we should also be smart about it and figure out if we can drop the precision too.  We do a CUDF cast instead of a Spark cast because Spark rounds on casting, and we just want to truncate the result.  To get the lower order data we keep the same scale/precision as before, but we subtract the higher order data from the lower order data.  This might involve casting things back to the same precision/scale but it should work.

One we have done the sum on the separate parts we need to check for overflow. I don't 100% know the math for this, but I think if we cudf cast the lower order sum to the same scale as the higher order sum, and add the higher order and casted lower order together, we can check the higher order bytes to know if the output would fit. Then we can cast the higher order data to the original scale and add it to the lower order data and do a final check for overflow.  The final overflow results would be an overflow for the higher order check, or for the lower order check.

This is likely to use a lot more memory and slow down processing. But it will mean we only have to worry about multiply and divide on the GPU for decimal 128 after this.

revans2:
This fixes #3944 

I ran some performance tests and I can see no real measurable difference when doing doing the extra overflow checking on the merge stage vs not on the merge stage.  As such I have turned it on so we can have true overflow detection on decimal SUMs no matter what.  This also made the docs a lot simpler to write.

I would like to do some follow on work and split apart AggregateFunctions.scala. It is way too large, but I didn't think that doing it here would be ideal.


revans2:
build

docs/compatibility.md
```diff
@@ -109,63 +109,31 @@ a few operations that we cannot support to the same degree as Spark can on the C
...
+[SPARK-32018](https://issues.apache.org/jira/browse/SPARK-32018) for more detailed information.
+Some of these fixes we were able to back port, but some of them require Spark 3.1.0 or above to
+fully be able to detect overflow in all cases. As such on versions of Spark older than 3.1.0 for
+large decimal values there is the possibility of data corruption in some corner cases. 
+This is true for both the CPU and GPU implementations, but there are fewer of these cases for the 
+GPU. If this concerns you, you should upgrade to Spark 3.1.0 or above. 
+
+When Apache Spark does a sum aggregation on decimal values it will store the result in a value
+with a precision that is the input precision + 10, but with a maximum precision of 38.
+For an input precision of 9 and above, Spark will do the aggregations as a java `BigDecimal`
```
jlowe, line 123:
```suggestion
For an input precision of 9 and above, Spark will do the aggregations as a Java `BigDecimal`
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuWindowExec.scala
```diff
@@ -322,11 +322,13 @@ object GpuWindowExec extends Arm {
...
           case GpuWindowExpression(
-          GpuAggregateExpression(rep: GpuReplaceWindowFunction, _, _, _, _), spec) =>
+          GpuAggregateExpression(rep: GpuReplaceWindowFunction, _, _, _, _), spec)
+            if rep.shouldReplaceWindow(spec) =>
             // We don't actually care about the GpuAggregateExpression because it is ignored
             // by our GPU window operations anyways.
             rep.windowReplacement(spec)
-          case GpuWindowExpression(rep: GpuReplaceWindowFunction, spec) =>
+          case GpuWindowExpression(rep: GpuReplaceWindowFunction, spec)
+            if rep.shouldReplaceWindow(spec)=>
```
jlowe, line 331:
```suggestion
            if rep.shouldReplaceWindow(spec) =>
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
```diff
@@ -381,6 +381,7 @@ case class GpuSampleExec(lowerBound: Double, upperBound: Double, withReplacement
 
   override lazy val additionalMetrics: Map[String, GpuMetric] = Map(
     OP_TIME -> createNanoTimingMetric(MODERATE_LEVEL, DESCRIPTION_OP_TIME))
+  // TODO CPU vs GPU OP TIME in Debug mode???
```
jlowe, line 384: Something missing here or is this a followup issue?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -575,6 +563,95 @@ case class GpuMax(child: Expression) extends GpuAggregateFunction
...
+ *    value would have wrapped around. The reason this works is because we have a hard limit on the
+ *    maximum number of values in a single batch being processed. `Int.MaxValue`, or about 2.2
+ *    billion values. So we use a precision on the higher values that is large enough to handle
+ *    2.2 billion values and still detect overflow. This equates to a precision of about 10 more
+ *    than is needed to hold the higher digits. This effectively gives us unlimited overflow
+ *    detection.
+ *
+ *
+ *
+ * but we do want to have some kind of guarantees that are large
```
jlowe, line 600: Not sure if this is a missing capitalization or if a line was accidentally deleted before this one.

jlowe (commented):
Minor nits but otherwise looks OK.  Would be good to have @abellina double-check the aggregation changes.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/basicPhysicalOperators.scala
revans2: Sorry no the follow on was already done. I just ran into some things with sample when I tried to use it for testing and ended up debugging too much on this branch.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
revans2: sorry about that. I didn't finish updating that comment when I was pulled off to other things and ended up forgetting to finish it.

revans2:
build

jlowe (dismissed):
Looks OK to me, but would be good to hear from @abellina on the agg changes before merging.  At first I was shocked to not find the `evaluateColumnar` that was in the now-deleted `processIncomingBatch`, but I assume this is covered by `AggHelper.preProcess`.

revans2:
> Looks OK to me, but would be good to hear from @abellina on the agg changes before merging. At first I was shocked to not find the `evaluateColumnar` that was in the now-deleted `processIncomingBatch`, but I assume this is covered by `AggHelper.preProcess`.

Yes. There was a bug were the output of `processIncommingBatch` was being bound to `aggBufferAttributes` in `AggHelper.preProcess`. This made it so I could not have an output of `inputProjection` that didn't have the same number of values as `aggBufferAttributes`. I fixed it by moving the processing of `inputProjection` into `AggHelper.preProcess`, and then binding it to the input schema for update, or binding to `aggBufferAttributes` only if it is a merge or is being force to be a merge.

abellina:
Yeap, I'll take a crack at this review and have some comments by tomorrow.

abellina (dismissed):
I've tried a few things with this patch like forcing the fallback (i.e. always merge) and decimals with spark 3.2.0. I don't see issues with it. I had a small nit, and it's not a blocker.

I spent time in the `boundInputReferences` change. I don't think the new code removes cases we covered before, and now adds the ability to have the differently shaped aggregation buffers that the overflow logic requires. The cast going away in `processingIncomingBatch` threw me off, but it should not be needed if our types are correct at bind time.

revans2:
build

revans2:
@abellina and @jlowe please take another look

jlowe (approved):

abellina (dismissed):

sameerz:
build

revans2:
build

revans2:
It looks like one of my new tests is crashing databricks.

```
[2021-12-04T15:19:07.772Z] [gw3] [ 69%] PASSED ../../src/main/python/hash_aggregate_test.py::test_hash_reduction_decimal_overflow_sum[31] 
[2021-12-04T15:19:13.634Z] [gw3] [ 69%] FAILED ../../src/main/python/hash_aggregate_test.py::test_hash_reduction_decimal_overflow_sum[30] 
```

```
[2021-12-04T15:31:55.481Z] E                   py4j.protocol.Py4JJavaError: An error occurred while calling o258684.collectToPython.
[2021-12-04T15:31:55.481Z] E                   : org.apache.spark.SparkException: Job aborted due to stage failure: Task 4 in stage 7638.0 failed 1 times, most recent failure: Lost task 4.0 in stage 7638.0 (TID 41051) (10.2.128.5 executor driver): java.lang.OutOfMemoryError: Could not allocate native memory: std::bad_alloc: out_of_memory: RMM failure at:/home/jenkins/agent/workspace/jenkins-cudf_nightly-dev-github-554-cuda11/cpp/build/_deps/rmm-src/include/rmm/mr/device/arena_memory_resource.hpp:157: Maximum pool size exceeded
[2021-12-04T15:31:55.481Z] E                    at ai.rapids.cudf.ColumnView.binaryOpVS(Native Method)
[2021-12-04T15:31:55.481Z] E                    at ai.rapids.cudf.ColumnView.binaryOp(ColumnView.java:1205)
[2021-12-04T15:31:55.481Z] E                    at ai.rapids.cudf.ColumnView.binaryOp(ColumnView.java:1195)
[2021-12-04T15:31:55.481Z] E                    at ai.rapids.cudf.BinaryOperable.div(BinaryOperable.java:171)
[2021-12-04T15:31:55.481Z] E                    at org.apache.spark.sql.rapids.GpuDecimalSumHighDigits.$anonfun$columnarEval$10(AggregateFunctions.scala:720)
[2021-12-04T15:31:55.481Z] E                    at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
[2021-12-04T15:31:55.481Z] E                    at com.nvidia.spark.rapids.Arm.withResource$(Arm.scala:26)
```

revans2:
So I did a little back of the envelope math on this. It gets a little complicated because of how some recent changes.  We are doing a reduction on 10^8 values.  If they are all 128-bit decimal values, then we end up with about 763 MB for all of the data in a single batch. But this should be split up among all of the tasks, which should be the number of CPU cores 8 in this case I think. That means a reduction should need about 95-96 MiB of memory to hold the input, and much less for the output. In the worst case we need to make a copy of the input so we can get the higher order digits out for extra checking.  Where we failed it was trying to calculate this value, and was in the middle of a divide.  That means we should have needed at most 200 MiB of memory to hold a single task's data.

This is running on a T4, but it is split up 4 ways because we are running the tests 4 at a time. The 16GiB is split into 5 parts, but it ends up being 1/5th of what is free and we take off 256 MiB of reserved space each time.  That results in about 1321 MiB of total memory for the worst case task. It should be more than enough.

That is unless databricks is some how optimizing things in this case and we end up doing a lot more data in a single task If we do all of the data in a single task, then we end up with 1600 MiB needed, which is more then what we have available in most of the tasks.  I'll see if I can reproduce this on my own.

revans2:
Okay I have a bit more information.

I did my math wrong above, and it was in one of the simple places, not the complex ones :(. I used a long instead of a 128-bit value. This means that we actually need about 1.5 GiB to hold all of the data in a single batch. But it is being divided up between 2 tasks for some reason. Not sure why, but thank you Databricks. Also the input columns have been factored out of the processing, so it is outputting batches that assume that the memory used would be a long.

If you combine all of this together, along with the other columns that we calculate as a part of SUM, we end up with an input batch with no columns and 50,000,000 rows.  This is turned into a column of 763 MiB for the SUM itself.  Also a boolean column of 48 MiB to keep track of what is and isn't empty. When we try to calculate the higher order bytes, it takes 763 MiB to produce the input column, and then when it tries to do the divide to produce the output column it fails to allocate enough memory. So it looks like I need to do a better job of properly sizing the input batches. I am not thrilled with this solution, but it is a rare problem, because someone is just doing a really dumb job of multiplying numbers, which should not really happen in real life.

revans2:
build

revans2:
build

jlowe (approved):

abellina (approved):

Issue #4275, opened 2021-12-03
[BUG] Row-based Hive UDF will fail if arguments contain a foldable expression.


firestarman:
Row-based Hive UDF will fail if there is a foldable expression in UDF arguments.

When creating an inspector from an expression, Spark will try to fold the foldable expression by calling its `eval` method, which is not supported by GPU expressions. 

Typical error stack
```shell
Caused by: java.lang.UnsupportedOperationException: Cannot evaluate expression: xxx
        at com.nvidia.spark.rapids.GpuExpression.eval(GpuExpressions.scala:118)
        at com.nvidia.spark.rapids.GpuExpression.eval$(GpuExpressions.scala:117)
        at com.nvidia.spark.rapids.GpuLeafExpression.eval(GpuExpressions.scala:166)
        at org.apache.spark.sql.hive.HiveInspectors.toInspector(HiveInspectors.scala:865)
        at org.apache.spark.sql.hive.HiveInspectors.toInspector$(HiveInspectors.scala:806)
        at org.apache.spark.sql.hive.rapids.GpuRowBasedHiveSimpleUDF.toInspector(rowBasedHiveUDFs.scala:70)
        at org.apache.spark.sql.hive.rapids.GpuRowBasedHiveSimpleUDF.$anonfun$wrappers$1(rowBasedHiveUDFs.scala:80)
        at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
        at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
        at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
```

PR #4282, opened 2021-12-03
Use data type to create an inspector for a foldable GPU expression.

firestarman:
Take care of the `GpuLiteral` and foldable GPU expressions specially when creating an inspector from an expression.

This fixes https://github.com/NVIDIA/spark-rapids/issues/4275

- [x] Add tests

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/rowBasedHiveUDFs.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/rowBasedHiveUDFs.scala
@@ -59,7 +59,7 @@ trait GpuRowBasedHiveUDFBase extends GpuRowBasedUserDefinedFunction with HiveIns
   protected def gpuToInspector(expr: Expression): ObjectInspector = expr match {
     case GpuLiteral(value, dataType) =>
       // Convert to CPU literal as possible as we can before going into `toInspector(Expression)`.
-      // Because inspectors for literals of primitive types are likely to get better performance.
+      // Because some optimization will be made for Literal expressions of primitive types.
       value match {
         case scalar: ai.rapids.cudf.Scalar =>
           if (scalar.getType.isNestedType) {
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/rowBasedHiveUDFs.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/rowBasedHiveUDFs.scala
@@ -55,11 +55,15 @@ trait GpuRowBasedHiveUDFBase extends GpuRowBasedUserDefinedFunction with HiveIns
 
   override def prettyName: String = name
 
-  /** Handle GpuLiteral and foldable GpuExpression, similar to `toInspector(Expression)` */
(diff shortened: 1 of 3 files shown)


firestarman:
Mark as draft because it is missing tests. I am working on it.

firestarman:
build

sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/rowBasedHiveUDFs.scala
```diff
@@ -55,6 +55,33 @@ trait GpuRowBasedHiveUDFBase extends GpuRowBasedUserDefinedFunction with HiveIns
 
   override def prettyName: String = name
 
+  /** Handle GpuLiteral and foldable GpuExpression, similar to `toInspector(Expression)` */
+  protected def gpuToInspector(expr: Expression): ObjectInspector = expr match {
+    case GpuLiteral(value, dataType) =>
+      // Convert to CPU literal as possible as we can before going into `toInspector(Expression)`.
+      // Because some optimization will be made for Literal expressions of primitive types.
```
jlowe, line 62: It would be nice to elaborate here what the issue is and why we need to do this.  "Some optimizations will be made" isn't clear why we need to worry and do anything special for literals.  If this code ends up causing problems in the future, it would be nice for those future maintainers to understand why this code was written in the first place.

sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/rowBasedHiveUDFs.scala
```diff
@@ -55,6 +55,33 @@ trait GpuRowBasedHiveUDFBase extends GpuRowBasedUserDefinedFunction with HiveIns
...
+        case scalar: ai.rapids.cudf.Scalar =>
+          if (scalar.getType.isNestedType) {
+            // Nested type, so create an inspector from the data type.
+            toInspector(dataType)
+          } else {
+            try {
+              toInspector(Literal.create(GpuScalar.extract(scalar), dataType))
+            } catch {
+              // Unsupported type for extraction, so use the data type way instead.
+              case _: Throwable => toInspector(dataType)
```
jlowe, line 73: Normally we don't call our own code and catch exceptions after the fact, rather we check what we know will work and avoid the exceptional path.  This ends up hiding bugs like NPEs and other unexpected exceptions, so I'd rather see us check types up-front or catch the specific exceptions that would be thrown. Or are there exceptions being thrown in `Literal.create` here as well?

firestarman: Nice catch. 
My mistake. Forgot to update this which was filled automatically by the IDE context tool.

sql-plugin/src/main/scala/org/apache/spark/sql/hive/rapids/rowBasedHiveUDFs.scala
firestarman: Updated.

firestarman:
build

firestarman:
build

jlowe (approved):

Issue #4002, opened 2021-11-02
[FEA] Implement regexp_extract on GPU


andygrove:
**Is your feature request related to a problem? Please describe.**
Once https://github.com/NVIDIA/spark-rapids/issues/4001 is implemented, we should implement regexp_extract.

**Describe the solution you'd like**
See above.

**Describe alternatives you've considered**
None

**Additional context**
None

Issue #4284, opened 2021-12-03
[FEA] Support idx = 0 in GpuRegExpExtract


andygrove:
**Is your feature request related to a problem? Please describe.**
The initial implementation of `GpuRegExpExtract` in https://github.com/NVIDIA/spark-rapids/issues/4002 will only support an idx that is greater than 0 (extract a single group). This issue is to implement support for idx of 0 which has different semantics.

**Describe the solution you'd like**
Match Spark behavior.

**Describe alternatives you've considered**
None

**Additional context**
None

andygrove:
I have a working implementation and will have a PR up soon. There are no dependencies on cuDF for this one.

andygrove:
This was implemented in https://github.com/NVIDIA/spark-rapids/pull/4285

PR #4239, opened 2021-11-30
Make EOL anchor behavior match CPU for strings ending with newline

andygrove:
Signed-off-by: Andy Grove <andygrove@nvidia.com>

Closes https://github.com/NVIDIA/spark-rapids/issues/4170

Update transpiler to replace `$` with `[\r]?[\n]?$` so that the GPU behavior matches the Java regular expression engine such that `abc$` will now match `abc`, `abc\n`, `abc\r`, and `abc\r\n`, but not `abc\n\r`.

PR #4285, opened 2021-12-03
Add support for regexp_extract on the GPU

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/4002 and https://github.com/NVIDIA/spark-rapids/issues/4284

This PR adds support for `regexp_extract`.


andygrove:
build

docs/compatibility.md
```diff
@@ -504,12 +506,12 @@ Here are some examples of regular expression patterns that are not supported on
 - Empty groups: `()`
 - Regular expressions containing null characters (unless the pattern is a simple literal string)
 - Beginning-of-line and end-of-line anchors (`^` and `$`) are not supported in some contexts, such as when combined 
-- with a choice (`^|a`) or when used anywhere in `regexp_replace` patterns.
+  with a choice (`^|a`) or when used anywhere in `regexp_replace` patterns.
 
 In addition to these cases that can be detected, there is also one known issue that can cause incorrect results:
 
-- `$` does not match the end of a string if the string ends with a line-terminator 
-  ([cuDF issue #9620](https://github.com/rapidsai/cudf/issues/9620))
```
andygrove, line 512: This issue was resolved in https://github.com/NVIDIA/spark-rapids/pull/4239 but the docs did not get updated. I can raise a separate PR for this change if necessary.

andygrove:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -852,23 +835,146 @@ case class GpuRegExpReplace(
...
+    // | NULL   | NULL  | NULL  |
+    // | 'a1a'  | '1'   | '1'   |
+    // | '1a1'  | ''    | NULL  |
+
+    withResource(GpuScalar.from("", DataTypes.StringType)) { emptyString =>
+      withResource(GpuScalar.from(null, DataTypes.StringType)) { nullString =>
+        withResource(str.getBase.extractRe(cudfRegexPattern)) { extract =>
+          withResource(str.getBase.matchesRe(cudfRegexPattern)) { matches =>
+            withResource(str.getBase.isNull) { isNull =>
+              withResource(extract.getColumn(i - 1)) { extractedGroup =>
```
andygrove, line 965: It would be nice if we could just ask cuDF to extract a single column and I have filed https://github.com/rapidsai/cudf/issues/9855 to request that capability

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -852,23 +835,146 @@ case class GpuRegExpReplace(
...
+        }
+      case _ =>
+        willNotWorkOnGpu(s"only non-null literal strings are supported on GPU")
+    }
+
+    val idx = expr.idx.asInstanceOf[Literal].value.asInstanceOf[Int]
+    if (idx == 0) {
+      // see https://github.com/NVIDIA/spark-rapids/issues/4284
+      willNotWorkOnGpu("idx 0 is not supported")
+    }
```
andygrove, line 906: TODO: I just realized that I need to add validation here to make sure idx is in range for the number of groups in the regexp pattern

docs/compatibility.md
```diff
@@ -448,6 +448,7 @@ The following Apache Spark regular expression functions and expressions are supp
 - `regexp`
 - `regexp_like`
 - `regexp_replace`
+- `regexp_extract`
```
jlowe, line 451: Nit: would be nice to keep this list lexicographically sorted.

integration_tests/src/main/python/string_test.py
```diff
@@ -538,6 +538,53 @@ def test_regexp_replace_character_set_negated():
...
+            lambda spark: unary_op_df(spark, gen).selectExpr(
+                'regexp_extract(a, "^([a-d]*)([^a-z]*)([a-d]*)$", 2)'),
+            conf={'spark.rapids.sql.expression.RegExpExtract': 'true'})
+
+def test_regexp_extract_idx_0():
+    gen = mk_str_gen('[abcd]{1,3}[0-9]{1,3}[abcd]{1,3}')
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark: unary_op_df(spark, gen).selectExpr(
+                'regexp_extract(a, "^([a-d]*)([0-9]*)([a-d]*)$", 0)',
+                'regexp_extract(a, "^([a-d]*)[0-9]*([a-d]*)$", 0)'),
```
jlowe, line 585: Would be nice to test the case where the regexp_extract group does _not_ match anything.  Looks like this will only test cases where a match is always found.  Maybe the repeat ranges above for at least one of the groups being generated should allow for the possibility of zero occurrences?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -852,23 +835,171 @@ case class GpuRegExpReplace(
...
+}
+
+case class GpuRegExpReplace(
+    srcExpr: Expression,
+    searchExpr: Expression,
+    replaceExpr: Expression,
+    cudfRegexPattern: String)
+  extends GpuRegExpTernaryBase with ImplicitCastInputTypes {
+
+  override def dataType: DataType = srcExpr.dataType
```
jlowe, line 858: Was surprised that this was not implemented in `GpuRegExpTernaryBase`.  When would we expect `dataType` for any regexp-like operator to _not_ be `StringType`?  Definitely no change necessary here and totally fine if this is just parroting the Spark source for consistency, but I'm curious if this really needs to be overridden here rather than specified in the new base class as always being `StringType`.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -852,23 +835,171 @@ case class GpuRegExpReplace(
...
+          numGroups = countGroups(new RegexParser(javaRegexpPattern).parse())
+        } catch {
+          case e: RegexUnsupportedException =>
+            willNotWorkOnGpu(e.getMessage)
+        }
+      case _ =>
+        willNotWorkOnGpu(s"only non-null literal strings are supported on GPU")
+    }
+
+    val idx = expr.idx.asInstanceOf[Literal].value.asInstanceOf[Int]
```
jlowe, line 915: Do we want to "future-proof" this and avoid putting on the GPU if `expr.idx` is not a `Literal` of `IntegerType`?  This is blindly assuming and will crash if Spark ever supports a non-literal index expression in the future.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -852,23 +835,171 @@ case class GpuRegExpReplace(
...
+    // | NULL   | NULL  | NULL  |
+    // | 'a1a'  | '1'   | '1'   |
+    // | '1a1'  | ''    | NULL  |
+
+    if (groupIndex == 0) {
+      withResource(GpuScalar.from("", DataTypes.StringType)) { emptyString =>
+        withResource(GpuScalar.from(null, DataTypes.StringType)) { nullString =>
+          withResource(str.getBase.matchesRe(cudfRegexPattern)) { matches =>
+            withResource(str.getBase.isNull) { isNull =>
+              withResource(matches.ifElse(str.getBase.incRefCount(), emptyString)) {
```
jlowe, line 977: Why the `str.getBase.incRefCount()`?  It seems that would trigger a refcount-induced leak.  Applies to same code below.

andygrove: This is one of the reasons I moved the PR back to draft. I did see a leak yesterday during testing but removing these `incrRefCount` calls results in a `closed too many times` error, so I need to figure out what is going on here.

jlowe: I think the issue has to do with this:
```
          withResource(str.getBase.extractRe(cudfRegexPattern)) { extract =>
[...]
                withResource(extract.getColumn(groupIndex - 1)) { extractedGroup =>
```
`extractRe` returns a `Table` which we definitely need to close, but `Table.getColumn` does not increment ref counts, so we should not be putting that in a `withResource` and trying to close it additionally.  Drop the `withResource` on the `getColumn` line and remove the extraneous `incRefCount` stuff and it should be better.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
andygrove: Fixed.

jlowe (dismissed):
Looks good to me, just a small comment on a test.

andygrove:
build

jlowe (approved):

PR #4291, opened 2021-12-04
Do not review

res-life:
This fixes https://github.com/NVIDIA/spark-rapids/issues/3932


res-life:
build

res-life:
build

res-life:
build

integration_tests/pom.xml
```diff
@@ -58,39 +58,67 @@
             <classifier>${cuda.version}</classifier>
             <scope>provided</scope>
         </dependency>
+
         <dependency>
             <groupId>com.nvidia</groupId>
-            <artifactId>rapids-4-spark_${scala.binary.version}</artifactId>
+            <artifactId>rapids-4-spark-sql_${scala.binary.version}</artifactId>
```
jlowe, line 64: The integration tests **must not** depend on the rapids-4-spark-sql artifact.  The whole point of having tests outside of the rapids-4-spark-sql project is to test against only the dist artifact jar (i.e.: what users will use), and pulling in this jar is like pulling in the aggregator jars -- it defeats the purpose of separating the tests into a different project and does not test against the artifact that will be shipped (i.e.: is not a true integration test).

jlowe (changes requested):

tgravescs:
can we please update the description of the approach taken?

res-life:
> 

It's a draft PR. I added some comments to the issue.

res-life:
@gerashegalov @jlowe @tgravescs
It's only a draft, but please review the solution:

Dynamically generate a jar for compile, then install it into the local maven repository, the script is in the dist module.

The Integration-tests module depends on the dynamical jar.
The Integration-tests does not depends on the sql-plugin jar.

integration_tests/pom.xml
```diff
@@ -59,38 +59,40 @@
...
         </dependency>
         <dependency>
-            <groupId>com.nvidia</groupId>
-            <artifactId>rapids-4-spark-udf-examples_${scala.binary.version}</artifactId>
-            <version>${project.version}</version>
-            <scope>test</scope>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-sql_${scala.binary.version}</artifactId>
+            <version>${spark.test.version}</version>
+            <scope>compile</scope>
```
jlowe, line 75: The integration tests are still depending on the spark-sql artifact and it will appear in the classpath when the tests are run.

jlowe (changes requested):
Why are we moving a bunch of unit tests into the integration tests module?  That is what is causing all the grief.  The original plan posted in #3932 was to move tests that require direct access to internal classes into the sql-plugin project as regular unit tests and the remaining tests that run at the Spark SQL/Dataframe API layer into integration tests.  Integration tests **must not** depend on the sql-plugin jar (or any other internal artifact).

integration_tests/pom.xml
res-life: I agree with **must not** depend on the **rapids-4-spark-sql** artifact, but we must depend on **spark-sql** to run the Scala test cases. When running the Python test cases the spark-sql artifact will appear in the classpath because of command "$SPARK_HOME/bin/pyspark". But for the Scala test cases is a different path, the class path is specified by "mvn test".
I'll move some unit tests to the rapids-4-spark-sql module, but the unit tests can also run in integration test.
I want to discuss if this solution is ok?

res-life:
build

res-life:
build

res-life:
The #3932 is invalid, so close it.

Issue #4238, opened 2021-11-30
[FEA] Add a Spark 3.X Explain only mode to the plugin


tgravescs:
**Is your feature request related to a problem? Please describe.**
For Spark versions 3.0+ add support to the plugin to run on the CPU in  explain only mode.  This would be similar to the explainPotentialGpuPlan api added but instead of user having to modify their code just have a config that says run the plugin in explain only mode and for each of the plans it just dumps out explain output to the logs but still runs everything on the CPU.

This could be used by customers who are evaluating using the plugin but don't have access to GPUs yet and want to see what parts of their query will run on the GPU and perhaps which features aren't yet supported.

PR #4322, opened 2021-12-07
Add an explain only mode to the plugin

tgravescs:
This allows users to run on the CPU and have the plugin evaluate the plan as if it would have run on GPU and output explain output the driver log file.  Note that this isn't perfect, specifically for AQE where the plan may change as its executed.

fixes https://github.com/NVIDIA/spark-rapids/issues/4238

In explain only mode we don't acquire GPU and don't enable spark rapids shuffle (fallback to Spark version if configured), but processes the plan through the GpuOverrides like it would if it were running on the GPU. In the end we return the CPU plan still so it runs on CPU and log the explain output.

This requires the cudf and rapids jar be present and plugin enabled with the  mode set to explainOnly. spark.rapids.sql.mode=explainOnly and  spark.plugins=com.nvidia.spark.SQLPlugin.  The alternate mode I called executeOnGPU just thinking if we happen to add other modes that might be best. Happy to change it if people have better ideas.

This PR updates the logging on startup to explain the mode, previously it just always printed how to turn the plugin off even if it was disabled.  Now it printed enabled, disabled or explain only mode and reference configs to change.

I tested hits on a bunch of NDS queries and compared output vs actually running on GPU. AQE isn't perfect but put some docs in there about it.  Added one basic integration test, I don't really have a good way to test the output since just goes to the logger.  Manually tested on Databricks.

We should make sure we have QA test for explain only mode as well.


tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -506,6 +506,17 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(true)
 
+  val SQL_EXPLAIN_ONLY_ENABLED = conf("spark.rapids.sql.explainOnly.enabled")
+    .doc("Enable (true) or disable (false) explain only mode for the plugin. This allows " +
+         "running queries on the CPU and the plugin will evaluate the queries as " +
+         "if it was going to run on the GPU. The the explanations of what would have run on " +
```
jlowe, line 512:
```suggestion
         "if it was going to run on the GPU. The explanations of what would have run on " +
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -506,6 +506,17 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(true)
 
+  val SQL_EXPLAIN_ONLY_ENABLED = conf("spark.rapids.sql.explainOnly.enabled")
```
jlowe, line 509: Since this essentially disables running on the GPU, would it make more sense to extend the existing `spark.rapids.sql.enabled` property to allow an 'explain' or `explainOnly` value that means it's only enabled for explanation only?  Otherwise you get into a situation where `explainOnly` is enabled but `spark.rapids.sql.enabled` is `false`, and then what would a user expect to happen?

jlowe (commented):
Quick high-level comments, have not reviewed in detail.  Also seems to be missing corresponding configs.md doc update from the RapidsConf change.

tgravescs:
I need to update the config docs

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
tgravescs: we certainly can, was thinking changing from boolean might be weird since to me "enabled"is generally a true or false setting.  Other possibilities would be change this to be spark.rapids.sql.mode where values are explain or convert (or something similar), then the enabled = false would truly disable it no matter the mode.
thoughts?

tgravescs: the other option I had considered was splitting it out more so you didn't specify the spark.plugins=com.nvidia.spark.SQLPlugin... if we are ever able to separate the code enough to not need cudf and such that might make more sense.

tgravescs: discussed offline and going to hold off a bit on deciding between the mode config and the separate plugin to investigate Spark 2.0 explain api.

tgravescs: the current 2.x implementation uses a completely separate jar and no configs are required, the user just calls an API. So from that perspective I'm not sure how this is implemented compared to that matters since no configuration needed.  

I'm not sure I have a strong opinion on either of the proposals. A few of the pros and cons:
Using just the config like spark.rapids.sql.mode will be less code changes, in some ways it might be easier to flip back and forth between the modes as well, but it won't be able to be changed run time because of the initialization differences. So in that aspect it could be confusing.
If we go with separate plugin it might be less likely to accidentally break if someone changes things in current code and doesn't think about this code.  Here I'm talking about things like accidentally initializing gpu or shuffle in this mode.
if we are able to separate out and have just a jar for tagging logic, that can be good and bad, less to download if just doing the explain but yet another jar user might need to deal with and if they switch to gpu nodes need to setup it up differently.

If anyone has other thoughts or opinions let me know. Otherwise I might go with the 'spark.rapids.sql.mode' config as its less change.

tgravescs:
I updated to have spark.rapids.sql.mode

tgravescs:
I need to regenerate the configs docs, will update shortly.

tgravescs:
build

tgravescs:
it seems our tests are relying on the behavior of device manager ot create the rapids buffer store (and initialize gpu and memory) even though sql plugin is disabled.  Looking at better way to handle this.

docs/get-started/getting-started-workload-qualification.md
```diff
@@ -54,15 +54,75 @@ Since the two tools are only analyzing Spark event logs they do not have the det
 captured from a running Spark job.  However it is very convenient because you can run the tools on
 existing logs and do not need a GPU cluster to run the tools.
 
-## 2. Function `explainPotentialGpuPlan` 
+## 2. Get the Explain Output
+
+This allows running queries on the CPU and the plugin will evaluate the queries as if it was
+going to run on the GPU and tell you what would and wouldn't have been run on the GPU.
+There are two ways to run this, one is running with the plugin set to explain only mode and
```
jlowe, line 61: Nit: Ideally should be using "RAPIDS Accelerator" rather than "plugin" in user docs.  Applies to other places in the PR.

docs/get-started/getting-started-workload-qualification.md
```diff
@@ -54,15 +54,75 @@ Since the two tools are only analyzing Spark event logs they do not have the det
...
+   For example:
+
+   ```bash
+   spark-shell --jars /PathTo/cudf-<version>.jar,/PathTo/rapids-4-spark_<version>.jar --conf spark.rapids.sql.mode=explainOnly --conf spark.plugins=com.nvidia.spark.SQLPlugin
+   ```
+2.  Enable optional RAPIDS Accelerator related parameters based on your setup.
+
+   Enabling optional parameters may allow more operations to run on the GPU but please understand
+   the meaning and risk of above parameters before enabling it. Please refer to [configs
+   doc](../configs.md) for details of RAPIDS Accelerator parameters.
```
jlowe, line 94:
```suggestion
   the meaning and risk of above parameters before enabling it. Please refer to the
   [configuration documentation](../configs.md) for details of RAPIDS Accelerator
   parameters.
```

docs/get-started/getting-started-workload-qualification.md
```diff
@@ -54,15 +54,75 @@ Since the two tools are only analyzing Spark event logs they do not have the det
...
+   spark.conf.set("spark.rapids.sql.udfCompiler.enabled",true)
+   ```
+
+3. Run your query and check the driver logs for the explain output.
+
+   Below are sample driver log messages starting with `!` which indicate the unsupported features in
+   this version:
+
+   ```
+   !NOT_FOUND <RowDataSourceScanExec> cannot run on GPU because no GPU enabled version of operator class org.apache.spark.sql.execution.RowDataSourceScanExec could be found
```
jlowe, line 114: This was changed in #4328 and the text here should be updated accordingly (and match the same instance that occurs later in this same file), e.g.:
```suggestion
   ! <RowDataSourceScanExec> cannot run on GPU because GPU does not currently support the operator class org.apache.spark.sql.execution.RowDataSourceScanExec
```
cannot run on GPU because GPU does not currently support the operator

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuShuffleEnv.scala
```diff
@@ -109,7 +109,7 @@ object GpuShuffleEnv extends Logging {
     }
```
jlowe, line 109: 2022 copyrights

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/RapidsShuffleInternalManagerBase.scala
```diff
@@ -254,6 +254,9 @@ abstract class RapidsShuffleInternalManagerBase(conf: SparkConf, val isDriver: B
     if (GpuShuffleEnv.isSparkAuthenticateEnabled) {
```
jlowe, line 254: 2022 copyrights

tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -505,6 +505,21 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(true)
 
+  val SQL_MODE = conf("spark.rapids.sql.mode")
+    .doc("Set the mode for the Rapids Accelerator. The supported modes are explainOnly and " +
+         "executeOnGPU. This config can not be changed at runtime, you must restart the " +
+         "application for it to take affect. The default mode is executeOnGPU, which means " +
+         "the Rapids Accelerator plugin convert the Spark operations and execute them on the " +
```
jlowe, line 512:
```suggestion
         "the RAPIDS Accelerator plugin convert the Spark operations and execute them on the " +
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -505,6 +505,21 @@ object RapidsConf {
...
+         "GPU when possible. The explainOnly mode allows running queries on the CPU and the " +
+         "Rapids Accelerator will evaluate the queries as if it was going to run on the GPU. " +
```
jlowe, line 514:
```suggestion
         "RAPIDS Accelerator will evaluate the queries as if it was going to run on the GPU. " +
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -505,6 +505,21 @@ object RapidsConf {
...
+         "The explanations of what would have run on the GPU and why are output in log " +
+         "messages. When using explainOnly mode, the default explain output is ALL, this can " +
+         "be changed by setting spark.rapids.sql.explain. Seeing that config for more details.")
```
jlowe, line 517:
```suggestion
         "be changed by setting spark.rapids.sql.explain. See that config for more details.")
```

tgravescs:
build

docs/configs.md
```diff
@@ -109,6 +109,7 @@ Name | Description | Default Value
...
+<a name="sql.mode"></a>spark.rapids.sql.mode|Set the mode for the Rapids Accelerator. The supported modes are explainOnly and executeOnGPU. This config can not be changed at runtime, you must restart the application for it to take affect. The default mode is executeOnGPU, which means the Rapids Accelerator plugin convert the Spark operations and execute them on the GPU when possible. The explainOnly mode allows running queries on the CPU and the Rapids Accelerator will evaluate the queries as if it was going to run on the GPU. The explanations of what would have run on the GPU and why are output in log messages. When using explainOnly mode, the default explain output is ALL, this can be changed by setting spark.rapids.sql.explain. Seeing that config for more details.|executeongpu
```
jlowe, line 112: Config docs need to be regenerated

tgravescs: thanks for pointing this out

tgravescs:
build

jlowe (approved):

Issue #4316, opened 2021-12-07
[BUG] Exception: Unable to find py4j, your SPARK_HOME may not be configured correctly intermittently


pxLi:
**Describe the bug**

Intermittently seeing 
```
[2021-12-07T06:58:00.874Z]     from spark_init_internal import get_spark_i_know_what_i_am_doing
[2021-12-07T06:58:00.874Z] ../../src/main/python/spark_init_internal.py:21: in <module>
[2021-12-07T06:58:00.874Z]     findspark.init()
[2021-12-07T06:58:00.874Z] /usr/local/lib/python3.8/dist-packages/findspark.py:145: in init
[2021-12-07T06:58:00.874Z]     raise Exception(
[2021-12-07T06:58:00.874Z] E   Exception: Unable to find py4j, your SPARK_HOME may not be configured correctly
```
in pre_merge and nightly CI

Hard to reproduce, looks like PYTHONPAYH somehow got changed during pytest
Open an issue to track

pxLi:
will keep monitoring this after #4349

pxLi:
closed. this did not appear again after https://github.com/NVIDIA/spark-rapids/pull/4349

pxLi:
seeing this again in one build of https://github.com/NVIDIA/spark-rapids/pull/6355
As this still seems ephemeral, I will spend some time try dig out the root cause

pxLi:
did some investigation, the path is correct and all the required files are there.

Seems the issue was due some NFS glitches, I will try figure out a way to remediate this later

gerashegalov:
I noticed in #6358 that we don't properly call findspark.init when pyspark is manually added to PYTHONPATH. Maybe it is a fix for py4j as well.

PR #4328, opened 2021-12-08
Change the explanation of why the operator will not work on GPU

### LINKED ISSUE #4088: [FEA] Change Explain to not print NOT_FOUND and to have some known CPU only tasks
We have had some feedback that in some cases with explain the NOT_FOUND is confusing. We should change this to print something about "this operator is not supported"

Along with that there are some known operators, like JDBC where we know we can never optimize it on the GPU and we should make the message more clear about how this is not a problem.

HaoYang670:
Signed-off-by: HaoYang670 13716567376yh@gmail.com
close #4088 
* Change explanation in log messages for unsupported operators. Instead of printing “NOT_FOUND”, we tell users that the GPU does not support the operator.

### DIFF
--- a/docs/get-started/getting-started-workload-qualification.md
+++ b/docs/get-started/getting-started-workload-qualification.md
@@ -124,7 +124,7 @@ the driver logs with `spark.rapids.sql.explain=all`.
    this version:
    
    ```
-   !Unsupported operator: <RowDataSourceScanExec> cannot run on GPU because no GPU enabled version of operator class org.apache.spark.sql.execution.RowDataSourceScanExec could be found
+   !Not supported by GPU: <RowDataSourceScanExec> cannot run on GPU because GPU does not support the operator class org.apache.spark.sql.execution.RowDataSourceScanExec
    ```
 
 This log can show you which operators (on what data type) can not run on GPU and the reason.
@@ -152,7 +152,7 @@ analysis.
 For example, the log lines starting with `!` is the so-called not-supported messages:
 ```
 !Exec <GenerateExec> cannot run on GPU because not all expressions can be replaced
-  !Unsupported operator:  <ReplicateRows> replicaterows(sum#99L, gender#76) cannot run on GPU because no GPU enabled version of expression class 
+  !Not supported by GPU:  <ReplicateRows> replicaterows(sum#99L, gender#76) cannot run on GPU because GPU does not support the operator ReplicateRows
 ```
 The indentation indicates the parent and child relationship for those expressions.
 If not all of the children expressions can run on GPU, the parent can not run on GPU either.
--- a/docs/get-started/getting-started-workload-qualification.md
+++ b/docs/get-started/getting-started-workload-qualification.md
@@ -124,7 +124,7 @@ the driver logs with `spark.rapids.sql.explain=all`.
    this version:
    
    ```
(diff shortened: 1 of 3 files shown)


HaoYang670:
build

docs/get-started/getting-started-workload-qualification.md
```diff
@@ -124,7 +124,7 @@ the driver logs with `spark.rapids.sql.explain=all`.
    this version:
    
    ```
-   !NOT_FOUND <RowDataSourceScanExec> cannot run on GPU because no GPU enabled version of operator class org.apache.spark.sql.execution.RowDataSourceScanExec could be found
+   !Unsupported operator: <RowDataSourceScanExec> cannot run on GPU because no GPU enabled version of operator class org.apache.spark.sql.execution.RowDataSourceScanExec could be found
```
jlowe, line 127: This still says "no GPU enabled version of operator class .... could be found".  The original problem that triggered #4088 was that the "class not found" message led the user to believe this was a classloader/classpath issue rather than a normal situation where we know this Catalyst operation is not supported.

HaoYang670:
build

HaoYang670:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -1307,7 +1307,7 @@ final class RuleNotFoundExprMeta[INPUT <: Expression](
   extends ExprMeta[INPUT](expr, conf, parent, new NoRuleDataFromReplacementRule) {
 
   override def tagExprForGpu(): Unit =
-    willNotWorkOnGpu(s"no GPU enabled version of expression ${expr.getClass} could be found")
+    willNotWorkOnGpu(s"GPU does not currently support the operator ${expr.getClass}")
```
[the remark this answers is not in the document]
jlowe: I think "currently" is appropriate, as without it the user may incorrectly believe the operation is impossible to support.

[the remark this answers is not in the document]
jlowe: Ah, sorry, I didn't see the context of the other changes.  I agree that we should be consistent here.  If we're not using "currently" in the other instances when we state something is unsupported then we shouldn't state it only here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsMeta.scala
```diff
@@ -46,7 +46,7 @@ trait DataFromReplacementRule {
  * A version of DataFromReplacementRule that is used when no replacement rule can be found.
  */
 final class NoRuleDataFromReplacementRule extends DataFromReplacementRule {
-  override val operationName: String = "NOT_FOUND"
+  override val operationName: String = "Not supported by GPU: "
```
jlowe, line 49: This is redundant, as it later explains on the same line that the operation is not supported.  We don't put "Not supported by GPU: " in front of other lines that will not end up on the GPU, so it's weird to call out this specially.  I think this should just be an empty string.

HaoYang670: Thank you for your advice. I will change it.

HaoYang670:
build

pxLi:
build

pxLi:
re-triggered due to weird https://github.com/NVIDIA/spark-rapids/issues/4316

jlowe (approved):

PR #4329, opened 2021-12-08
A small improvment in GPU `If` and `CaseWhen` for the all-true and all-false cases.

firestarman:
This PR is to support the lazy evaluation of the true or false expression in GPU `If` and `CaseWhen` only when the predictions evaluated from an input batch are all true or false.

This can improve the performance if the true or false expression is an expensive operation (e.g. row-based UDF),  and will be skipped according to the prediction result.

fixes https://github.com/NVIDIA/spark-rapids/issues/4350


firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -29,37 +29,51 @@ trait GpuConditionalExpression extends ComplexTypeMergingExpression with GpuExpr
...
       }
     }
   }
 
+  protected def isAllTrue(col: GpuColumnVector): Boolean = {
+    assert(BooleanType == col.dataType())
+    withResource(col.getBase.all()) { allTrue =>
+      // null is treated as false in Spark, but skipped by 'all()' method.
+      allTrue.isValid && allTrue.getBoolean && !col.hasNull
+    }
```
jlowe, line 60: Checking for any nulls is often very cheap, and given we need to do it anyway it should be done up-front to avoid the unnecessary `all()` kernel when there are nulls.  Also `isValid` forces a DtoH synchronization that I think is unnecessary here since we're already going to check if there are any nulls.
```suggestion
    if (col.getRowCount == 0) {
      return true
    }
    if (col.hasNull) {
      return false
    }
    withResource(col.getBase.all()) { allTrue =>
      // guaranteed there is at least one row and no nulls so result must be valid
      allTrue.getBoolean
    }
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -82,8 +96,30 @@ case class GpuIf(
     }
   }
 
-  override def columnarEval(batch: ColumnarBatch): Any = computeIfElse(batch, predicateExpr,
-    trueExpr, falseExpr.columnarEval(batch))
+  @transient
+  private[this] lazy val func = if (optAllTrueFalse) {
```
jlowe, line 100: Nit: `func` here is too generic, we can afford a better name for readability.  Actually `lazy val` can override a function def, so should be able to do something more direct like this:
```
override lazy val columnarEval(batch: ColumnarBatch): Any = if (...) {
```

Apples to the other places where this was done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -93,7 +129,8 @@ case class GpuIf(
 
 case class GpuCaseWhen(
     branches: Seq[(Expression, Expression)],
-    elseValue: Option[Expression] = None) extends GpuConditionalExpression with Serializable {
+    elseValue: Option[Expression] = None,
+    optAllTrueFalse: Boolean) extends GpuConditionalExpression with Serializable {
```
jlowe, line 133: Non-default arguments should not appear after default arguments.

jlowe (commented):
> This can improve the performance if the true or false expression is an expensive operation (e.g. row-based UDF), and will be skipped according to the prediction result.

This skip benefit is _very_ specific to the case where everything is skewing away from one side of the condition.  Have we done any performance measurements on this?  Personally I do not want yet another config that users will not notice or remember to enable/disable.  If this is really cheap, I'd rather just enable it all the time than have a config.  We should run this on the NDS benchmark to see if it is noticeable when enabled.

firestarman:
> This skip benefit is very specific to the case where everything is skewing away from one side of the condition.

We indeed have this kind of case. So made this change and @viadea verified it got better perf.

firestarman:
> We should run this on the NDS benchmark to see if it is noticeable when enabled.

Will work on this next.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: I tried but got an error.
```
[ERROR] [Error] /home/liangcail/work/projects/on_github/spark-rapids/sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala:222: not found: value columnarEval
[ERROR] one error found
```
Then, I changed to 
`override lazy val columnarEval: ColumnarBatch => Any = if (...) {`
, but still got another error
```
[ERROR] [Error] /home/liangcail/work/projects/on_github/spark-rapids/sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala:222: lazy value columnarEval overrides nothing.
Note: the super classes of class GpuCaseWhen contain the following, non final members named columnarEval:
```
So I just renamed the function.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: Good to know, updated.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: I used `lazy val` to implement an abstrct method before.

firestarman: I decided to remove the config, so this is out of date.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: I decided to remove the config, so this is out of date.

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -129,13 +159,39 @@ case class GpuCaseWhen(
...
+        isAllPredsFalse = isAllPredsFalse && isAllFalse(p)
+        i += 1
+      }
+
+      val elseRet = elseValue
+        .map(_.columnarEval(batch))
+        .getOrElse(GpuScalar(null, branches.last._2.dataType))
+      if (isAllPredsFalse) {
+        // No predication has a true, so return the else value.
+        elseRet
```
jlowe, line 189: We can't return a raw `GpuScalar` here, this isn't a reduction expression.  It needs to be a `GpuColumnVector` since the input being evaluated for the entire expression was a `ColumnarBatch`.  That probably explains the failing test case in the CI run.  Before this code change, that was automatically handled by `ColumnView.ifElse` which supports taking a scalar directly.

I think this needs to be something like:
```suggestion
        elseRet match {
          case s: GpuScalar =>
            withResource(s) { s =>
              GpuColumnVector.from(s, batch.numRows, s.dataType)
            }
          case c => c
        }
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -129,13 +159,39 @@ case class GpuCaseWhen(
...
+    val predictions = new Array[GpuColumnVector](size)
+    var isAllPredsFalse = true
+    var i = 0
+
+    withResource(predictions) { preds =>
+      while (i < size) {
+        // If any predication is the first all-true, then evaluate its true expression
+        // and return the result.
+        preds(i) = GpuExpressionsUtils.columnarEvalToColumn(branches(i)._1, batch)
+        val p = preds(i)
```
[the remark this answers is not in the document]
firestarman: Fine with me. Updated.
I just wanted to put it into the cache as soon as possible.

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: Great catch, updated. also updated this in `GpuIf`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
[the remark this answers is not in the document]
firestarman: I just wanted to put the result into the cache as soon as possible to avoid leaking, even the assignment `preds(i) = p` is almost impossible to fail. But I am find with the suggestion. Updated.

integration_tests/src/main/python/conditionals_test.py
```diff
@@ -40,11 +40,12 @@
 if_nested_gens = if_array_gens_sample + if_struct_gens_sample
 
 @pytest.mark.parametrize('data_gen', all_gens + if_nested_gens + decimal_128_gens_no_neg, ids=idfn)
-def test_if_else(data_gen):
+@pytest.mark.parametrize('row_num', [2048, 30]) # Small row number can get all-true or all-false batch easily.
```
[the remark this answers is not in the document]
firestarman: Good suggestion, updated.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -129,13 +159,39 @@ case class GpuCaseWhen(
...
+  private[this] lazy val trueExpressions = branches.map(_._2)
+
+  override def columnarEval(batch:ColumnarBatch): Any = {
+    val size = branches.size
+    val predictions = new Array[GpuColumnVector](size)
+    var isAllPredsFalse = true
+    var i = 0
+
+    withResource(predictions) { preds =>
+      while (i < size) {
```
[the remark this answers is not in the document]
firestarman: Updated

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -29,31 +29,50 @@ trait GpuConditionalExpression extends ComplexTypeMergingExpression with GpuExpr
...
+    withResource(col.getBase.all()) { allTrue =>
+      // Guaranteed there is at least one row and no nulls so result must be valid
+      allTrue.getBoolean
+    }
+  }
+
+  protected def isAllFalse(col: GpuColumnVector): Boolean = {
+    assert(BooleanType == col.dataType())
+    withResource(col.getBase.any()) { anyTrue =>
+      anyTrue.isValid && !anyTrue.getBoolean
```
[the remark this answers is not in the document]
firestarman: Great suggestion, updated.

firestarman:
build

firestarman:
build

firestarman:
build

integration_tests/src/main/python/data_gen.py
```diff
@@ -416,11 +416,16 @@ def gen_part_double():
...
 class BooleanGen(DataGen):
     """Generate Bools (True/False)"""
-    def __init__(self, nullable=True):
+    def __init__(self, nullable=True, const_value=None):
         super().__init__(BooleanType(), nullable=nullable)
+        self.const_value = const_value
 
     def start(self, rand):
-        self._start(rand, lambda : bool(rand.getrandbits(1)))
+        if self.const_value:
```
jlowe, line 424: This does not generate a constant value if `const_value` is `False`.  It will fall through to the random generator instead.

I don't think we need to modify the data generators for this.  There's already a `SetValuesGen` we should be able to use via something like this in the respective tests:
```python
  bool_gen = boolean_gen if const_bool is None else SetValuesGen(BooleanType(), [const_bool])
```

integration_tests/src/main/python/conditionals_test.py
```diff
@@ -67,11 +69,15 @@ def test_if_else_map(data_gen):
...
 @pytest.mark.order(1) # at the head of xdist worker queue if pytest-order is installed
 @pytest.mark.parametrize('data_gen', all_gens + all_nested_gens + decimal_128_gens, ids=idfn)
-def test_case_when(data_gen):
+@pytest.mark.parametrize('const_bool', [True, False, None])
+def test_case_when(data_gen, const_bool):
     num_cmps = 20
     s1 = gen_scalar(data_gen, force_no_nulls=not isinstance(data_gen, NullGen))
-    # we want lots of false
-    bool_gen = BooleanGen().with_special_case(False, weight=1000.0)
+    if const_bool:
```
jlowe, line 76: Same issue where `const_bool == False` will not go through this code path.  `if const_bool is not None` seems more appropriate here.

integration_tests/src/main/python/conditionals_test.py
```diff
@@ -40,11 +40,13 @@
 if_nested_gens = if_array_gens_sample + if_struct_gens_sample
 
 @pytest.mark.parametrize('data_gen', all_gens + if_nested_gens + decimal_128_gens_no_neg, ids=idfn)
-def test_if_else(data_gen):
+@pytest.mark.parametrize('const_bool', [True, False, None])
```
jlowe, line 43: It would be good to test not just with all true, all false, and all random values but also all-null values.

integration_tests/src/main/python/data_gen.py
firestarman: Nice catch. Updated

integration_tests/src/main/python/conditionals_test.py
firestarman: Nice catch, updated.

integration_tests/src/main/python/conditionals_test.py
firestarman: Added it.

firestarman:
build

firestarman:
build

pxLi:
some of the parallel java agents crashed in CI w/ this change, this is always reproducible in pre-merge CI but not locally.
still not quite sure what's the root cause

pxLi:
build

firestarman:
try to sync up to the latest commit.

firestarman:
build

firestarman:
build

firestarman:
build

firestarman:
build

firestarman:
build

firestarman:
build

firestarman:
build

jlowe (commented):
The new CaseWhen algorithm can take a lot more memory when there's a large number of cases to check, so I'm not sure it makes sense to unconditionally do this new algorithm.  CI showed that use-cases that ran perfectly well before this change can suddenly crash with the new algorithm.  As we move down the predicates and cache the result of their evaluation, we really should be making them spillable to avoid the OOM condition.  However even if we make them spillable, if we cache so many that we start spilling the whole point of doing this algorithm (i.e.: performance) is foiled.

We may want to be smarter about deciding when to perform the algorithm for CaseWhen (e.g.: only perform the new algorithm when the number of predicates to check is small).  For IfElse it seems clear this is generally a good thing, but I'm becoming doubtful that this is a net win for CaseWhen as it is written now.

integration_tests/src/main/python/conditionals_test.py
```diff
@@ -97,6 +99,34 @@ def test_case_when(data_gen):
                 f.when(f.lit(False), f.col('_c0'))),
             conf = allow_negative_scale_of_decimal_conf)
 
+# Seperate this test to avoid OOM in premerge build.
+@pytest.mark.parametrize('data_gen', all_gens + all_nested_gens + decimal_128_gens, ids=idfn)
+@pytest.mark.parametrize('pred_value', [True, False, None])
+def test_case_when_all_true_false(data_gen, pred_value):
+    # Smaller number to reduce the memory size for this big test. Bigger number may get
+    # OOM easily and be killed in premerge build.
```
jlowe, line 107: The fact that we had to write a separate test to avoid an OOM screams this is not something we want to do purely in search of a performance win.

firestarman:
> The new CaseWhen algorithm can take a lot more memory when there's a large number of cases to check, so I'm not sure it makes sense to unconditionally ...

Yes, need to take more care in CaseWhen to get a balance between the performance and memory.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -129,13 +160,36 @@ case class GpuCaseWhen(
...
-    val elseRet = elseValue
-      .map(_.columnarEval(batch))
-      .getOrElse(GpuScalar(null, branches.last._2.dataType))
-    branches.foldRight[Any](elseRet) { case ((predicateExpr, trueExpr), falseRet) =>
-      computeIfElse(batch, predicateExpr, trueExpr, falseRet)
+  @transient
+  private[this] lazy val trueExpressions = branches.map(_._2)
+
+  override def columnarEval(batch:ColumnarBatch): Any = {
+    val predictions = new Array[GpuColumnVector](branches.size)
```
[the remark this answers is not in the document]
firestarman: Thanks. Updated

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -29,31 +29,51 @@ trait GpuConditionalExpression extends ComplexTypeMergingExpression with GpuExpr
...
+    withResource(col.getBase.all()) { allTrue =>
+      // Guaranteed there is at least one row and no nulls so result must be valid
+      allTrue.getBoolean
+    }
+  }
+
+  protected def isAllFalse(col: GpuColumnVector): Boolean = {
+    assert(BooleanType == col.dataType())
+    withResource(col.getBase.any()) { anyTrue =>
+      // null values are considered false values in this context
```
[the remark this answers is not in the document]
firestarman: Good suggestion. Added

integration_tests/src/main/python/conditionals_test.py
firestarman: Updated.

firestarman:
build

integration_tests/src/main/python/conditionals_test.py
firestarman: Merged the two tests

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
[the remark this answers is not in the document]
firestarman: Thanks. Updated

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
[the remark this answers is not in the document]
firestarman: Good suggestion, updated

firestarman:
Move to draft since premerge still failed without caching the predicates.

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -29,31 +29,55 @@ trait GpuConditionalExpression extends ComplexTypeMergingExpression with GpuExpr
...
+
+  protected def isAllFalse(col: GpuColumnVector): Boolean = {
+    assert(BooleanType == col.dataType())
+    if (col.getRowCount == col.numNulls()) {
+      // all nulls, and null values are false values here
+      return true
+    }
+    withResource(col.getBase.any()) { anyTrue =>
+      // null values are considered false values in this context
+      !anyTrue.isValid || !anyTrue.getBoolean
```
jlowe, line 77: Now that the row count is being checked against the null count, we shouldn't need to check the validity of the `any` result (and thus avoid another GPU synchronization to fetch the value).  It can only be null if there are no rows or all rows are null, but that case is now covered by the row count == null count check.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -129,13 +164,34 @@ case class GpuCaseWhen(
...
-      computeIfElse(batch, predicateExpr, trueExpr, falseRet)
+    if (isAllPredsFalse) {
+      // No predicate has a true, so return the else value directly.
+      GpuExpressionsUtils.resolveColumnVector(elseRet, batch.numRows())
+    } else {
+      branches.foldRight[Any](elseRet) { case ((predExpr, trueExpr), falseRet) =>
+        // Here evaluates the predicate expressions again, sacrificing a little of perf
+        // for minimizing the memory size of computations. Because caching the predicates
+        // may get OOM if there are too many branches.
+        withResource(GpuExpressionsUtils.columnarEvalToColumn(predExpr, batch)) { p =>
```
jlowe, line 191: Do we really want to double-evaluate the expression?  What if the predicate expression is relatively expensive to evaluate (e.g.: is a regular expression match on a large string column or a UDF)?

Rather than waste work re-evaluating expressions that have unknown time and complexity, I'd rather see an optimization for the common cases where the case/when is really just an if/else or we budget how much memory we're willing to spend on caching the predicate expressions.  In other words, if we see a case/when that has a lot of predicate expressions, maybe we abandon trying to perform the optimization (or maybe just check the very first predicate for the all true case but otherwise fallback) to avoid excessive memory computation.

As it is now, I'm not convinced this is a net-win since we now are making this significantly more expensive in the case where we don't fall entirely in one predicate for each batch.

jlowe (commented):
Given the complexity of the CaseWhen optimization, we may want to split this PR and get the more straightforward GpuIf optimization in while we work through the complexity of CaseWhen.

firestarman:
build

firestarman:
build

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: This was for OOM test. Now updated.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
firestarman: Updated

firestarman:
I have confirmed the OOM issue in the premerge build has nothing to do with the code change in Scala. But it makes me to consider about the possible GPU OOM if caching too many predicate columns. So I introduce a config to limit the number of columns to be cached.

firestarman:
> Given the complexity of the CaseWhen optimization, we may want to split this PR and get the more straightforward GpuIf optimization in while we work through the complexity of CaseWhen.

Yes, however the all-true case is using `case-when`, so I would want `GpuCaseWhen` to get in more than 'GpuIf'.

jenkins/spark-premerge-build.sh
```diff
@@ -108,7 +108,7 @@ ci_2() {
     export TEST_TYPE="pre-commit"
     export TEST_PARALLEL=4
     # separate process to avoid OOM kill
-    TEST='conditionals_test or window_function_test' ./integration_tests/run_pyspark_from_build.sh
+    TEST_PARALLEL=2 TEST='conditionals_test or window_function_test' ./integration_tests/run_pyspark_from_build.sh
```
jlowe, line 111: Not sure why this change is part of this PR if GPU OOM during premerge tests are not related to the case when changes?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1325,6 +1325,13 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(value = false)
 
+  val MAX_CONDITION_BRANCH_NUMBER = conf("spark.rapids.sql.opt.condition.maxBranchNumber")
+    .doc("Maximum number of branches for GPU case-when to enable the lazy evaluation of true " +
+      "and else expressions if the predicates on a batch are all-true or all-false. Big number " +
+      "may get GPU OOM easily since the predicates are cached during the computation.")
+    .integerConf
+    .createWithDefault(2)
```
jlowe, line 1333: I'm not a fan of this config.  Users will not know what to set it to for their queries, nor will they realize this config is what needs to be tuned if the GPU runs out of memory due to it.

For small input batches we probably could cache far more than 2, but for very large batches we may not even want to do 2.  What we really want is a budget, in memory size not branch count or row sizes, that the CaseWhen expression is allowed to use.  Unfortunately we don't currently have a clear budget for this (and other expressions could similarly leverage this budget for its processing/optimizations).

cc: @revans2 since he may have further thoughts on this.

jlowe (commented):
> however the all-true case is using case-when

Is this use-case using a degenerate case-when (i.e.: just one branch and therefore like an `If`) or does it have three possible paths (i.e.: two predicate branches + else)?  I'm assuming the latter since you defaulted the config to 2.

jenkins/spark-premerge-build.sh
firestarman: This is to pass the premerge build. 
The OOM we are talking about is not on GPU, but the host memory on CPU.

What I just found is if we add enough new tests of case_when with the default parallel value (it is 4 according to the build log), the ci_2 process will run into OOM (Host memory reached the up limit of a pod).  When reudce it to 2, it can pass.

I doult if we missed to release some memory between tests, or there would be some issues in our premerge build. Since the host memory in the ci_2 process kept increasing during the tests running.

firestarman:
> Is this use-case using a degenerate case-when (i.e.: just one branch and therefore like an `If`) or does it have three possible paths (i.e.: two predicate branches + else)? I'm assuming the latter since you defaulted the config to 2.

It is not related to the use-case. 2 is just one more than that in GpuIf, and still tries to minimize the GPU memory being used.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
firestarman: Agree. 
Besides, different GPUs have different memory sizes, a value can work on A100 but may not work on T4.
For most cases, I think, users do not need to take care of this config, since 2 should be small enough to run without GPU OOM.

jlowe:
Based on offline discussions we're holding off on CaseWhen corner-case optimizations for now due to the memory impact concerns, and the GpuIf all-true / all-false optimization is already folded into #4358.

Issue #4334, opened 2021-12-09
[BUG] GpuColumnarToRowExec will always be tagged False for exportColumnarRdd after Spark311 


wjxiz1992:
**Describe the bug**
when using [ColumnarRdd](https://github.com/NVIDIA/spark-rapids/blob/branch-22.02/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarRdd.scala#L47) in Spark 3.1.2 (all after), the Shim layer will always use the default value(false) for `exportColumnRdd` , e.g. https://github.com/NVIDIA/spark-rapids/blob/branch-22.02/sql-plugin/src/main/311until320-nondb/scala/com/nvidia/spark/rapids/shims/v2/Spark31XShims.scala#L457-L459

**Steps/Code to reproduce bug**
Calling ColumnarRdd in any version after 311, will cause this problem (XGBoost, PCA train):
![image](https://user-images.githubusercontent.com/20476954/145357391-4d0219aa-0135-4d91-8d54-e3493f70158a.png)

**Expected behavior**
There should not be "columnar to row" conversion.

**Environment details (please complete the following information)**
Spark 3.1.2 Standalone

**Additional context**
Version before 311 is fine.

tgravescs:
this looks like its been broken since release 0.2, do we not have a test for this?  Do we have an idea of what exactly it causes?  If perf issue how much does it affect it.

wjxiz1992:
Not clear about the XGBoost case. but for PCA case. with 4.6G parquet data , the column is of ArrayType(DoubleType) and the array size if 2048. the perf will drop from 6 seconds to 8 minutes in this case.

without the fix in 3.1.2, training is 8 minutes 
![image](https://user-images.githubusercontent.com/20476954/145425596-aaa1e6ea-f775-41f1-b884-3d6c47061e10.png)

but in 3.0.1: 6 seconds
![image](https://user-images.githubusercontent.com/20476954/145425630-297f4c55-2302-4382-9421-e9a2b06e1126.png)

PR #4335, opened 2021-12-09
Fix the issue of exporting Column RDD [databricks]

wbo4958:
This PR is trying to fix https://github.com/NVIDIA/spark-rapids/issues/4334.

After 3.1.x (included), `ColumnarRDD(df)` can't extract RDD[Table] directly, instead it will involve columnar to row and row to column which causes perf bad. It turned out the exportColumnRdd is not passed to `GpuColumnarToRowExecParent`


wbo4958:
build

revans2 (dismissed):

jlowe (dismissed):
Ideally the test should be written as an integration test using the public `ColumnarRdd` API to replicate what user code would do, but I'm OK with this being a followup.

sameerz (changes requested):
Please retarget to 22.02, as we are in code freeze for 21.12 and this is not a data corruption, crash, or customer critical fix (at this time).

wbo4958:
> Ideally the test should be written as an integration test using the public `ColumnarRdd` API to replicate what user code would do, but I'm OK with this being a followup.

Hmm, Seems the user can't get `ColumnarRdd` API from PySpark.

wbo4958:
build

sameerz (approved):
Thanks for moving to 22.02.

jlowe (approved):

jlowe:
Premerge failed in what appears to be an unrelated Python env error.  Rekicking.

jlowe:
build

wbo4958:
build

PR #4338, opened 2021-12-09
Use scala Int.box instead of Integer constructors 

abellina:
This is a small fix for the test code where it was using `new Integer` instead of `Int.box`. The integer constructors were deprecated in JDK 11, so this is not a must have for us, but it is not a great user experience to have the plugin fail to build because of this.

@sameerz up to you if I should target this to 21.12. @jlowe @revans2  fyi.

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleTestHelper.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleTestHelper.scala
@@ -165,7 +165,7 @@ object RapidsShuffleTestHelper extends MockitoSugar with Arm {
   }
 
   def withMockContiguousTable[T](numRows: Long)(body: ContiguousTable => T): T = {
-    val rows: Seq[Integer] = (0 until numRows.toInt).map(new Integer(_))
+    val rows: Seq[Integer] = (0 until numRows.toInt).map(Int.box)
     withResource(ColumnVector.fromBoxedInts(rows:_*)) { cvBase =>
       cvBase.incRefCount()
       val gpuCv = GpuColumnVector.from(cvBase, IntegerType)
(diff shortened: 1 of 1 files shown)


mythrocks (dismissed):
Wow. TIL.

LGTM! lolkthxby.

jlowe (dismissed):

abellina:
build

sameerz:
> @sameerz up to you if I should target this to 21.12. @jlowe @revans2 fyi.

We do not support Java 8 in 21.12, so let's target this for the next release.

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleTestHelper.scala
```diff
@@ -165,7 +165,7 @@ object RapidsShuffleTestHelper extends MockitoSugar with Arm {
   }
 
   def withMockContiguousTable[T](numRows: Long)(body: ContiguousTable => T): T = {
-    val rows: Seq[Integer] = (0 until numRows.toInt).map(new Integer(_))
+    val rows: Seq[Integer] = (0 until numRows.toInt).map(Int.box)
```
gerashegalov, line 168: nit:
```suggestion
    val rows: Seq[Integer] = (0 until numRows).map(Int.box)
```

gerashegalov (approved):
LGTM, nit

abellina:
build

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleTestHelper.scala
gerashegalov: sorry, that was a bad comment, scalac won't do a silent lossy conversion for us. I'll undo it

abellina: @gerashegalov no problem, the constructor (I was trying not to use) also came back with the revert, but I changed it back to `Int.box`. Please take another look!

abellina:
build

gerashegalov (approved):

Issue #4339, opened 2021-12-10
The parameter `dataType` is not necessary in `resolveColumnVector` method.


firestarman:
Since the data type can be got from all the supported types, better to remove this parameter to simplify the calling.

PR #4340, opened 2021-12-10
Remove the unnecessary parameter `dataType` in `resolveColumnVector` method

firestarman:
This fixes https://github.com/NVIDIA/spark-rapids/issues/4339

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpandExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpandExec.scala
@@ -163,7 +163,7 @@ class GpuExpandIterator(
         val (cv, nullColumnReused) = expr.columnarEval(cb) match {
           case null => getOrCreateNullCV(sparkType)
           case other =>
-            (GpuExpressionsUtils.resolveColumnVector(other, cb.numRows, sparkType), false)
+            (GpuExpressionsUtils.resolveColumnVector(other, cb.numRows), false)
         }
         if (!nullColumnReused) {
           uniqueDeviceColumns += cv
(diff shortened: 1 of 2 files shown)


firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpandExec.scala
```diff
@@ -163,7 +163,7 @@ class GpuExpandIterator(
         val (cv, nullColumnReused) = expr.columnarEval(cb) match {
           case null => getOrCreateNullCV(sparkType)
           case other =>
-            (GpuExpressionsUtils.resolveColumnVector(other, cb.numRows, sparkType), false)
+            (GpuExpressionsUtils.resolveColumnVector(other, cb.numRows), false)
```
GaryShen2008, line 166: Does the return value "other" of expr.columnarEval() have the same type as the expr.dataType?

sperlingxx:
LGTM

sperlingxx (approved):
LGTM

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpandExec.scala
sperlingxx: As long as the expression itself is correctly programmed, sparkType is the result type of the "expr".

firestarman: Yes. You can find it here
https://github.com/NVIDIA/spark-rapids/blob/branch-22.02/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpandExec.scala#L162

PR #4343, opened 2021-12-10
Fix issue with binding to hash agg columns with computation [databricks]

revans2:
Fixes an issue where on older versions of Spark the hash aggregate included computation binding was wrong. This was introduced by #4272

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
@@ -423,8 +423,8 @@ class GpuHashAggregateIterator(
     }
 
     val shims = ShimLoader.getSparkShims
-    val ordering = groupingExpressions.map(shims.sortOrder(_, Ascending, NullsFirst))
     val groupingAttributes = groupingExpressions.map(_.toAttribute)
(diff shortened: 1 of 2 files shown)


revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/aggregate.scala
```diff
@@ -679,7 +686,7 @@ class GpuHashAggregateIterator(
 
     // a bound expression that is applied before the cuDF aggregate
     private val preStepBound = if (forceMerge) {
-      GpuBindReferences.bindGpuReferences(preStep, aggBufferAttributes)
+      GpuBindReferences.bindGpuReferences(preStep.toList, aggBufferAttributes.toList)
```
abellina, line 689: nit, do we need the `toList` here?

abellina (approved):
1 nit, otherwise LGTM

sameerz:
build

jlowe:
The CI failure was due to cudf missing a dynamic link dependency.  Should be fixed by https://github.com/rapidsai/cudf/pull/9890.

firestarman:
build

PR #4345, opened 2021-12-10
WIP: Leverage cudf conditional nested loop join to implement semi/anti hash join with condition

### LINKED ISSUE #4309: [FEA] Leverage cudf conditional nested loop join to implement semi/anti hash join with condition
Currently semi and anti hash joins with a condition fallback to the CPU because cudf does not support a hash-based semi or anti join with a condition, and unlike inner join we cannot simply apply the condition as a post-filter on the join.  However cudf does support semi/anti joins with a condition using `conditional_left_semi_join` and `conditional_left_anti_join`.  Despite the fact that a nested loop join is algorithmically worse than a hash join, it may be more performant in practice because it could avoid the overhead of CPU<->GPU transitions in the plan.

The RAPIDS Accelerator leverages the same code as used for broadcast nested loop joins to check if the semi/anti join condition can be supported by the cudf AST.  If it can, it replaces the CPU semi/anti join in the plan with a GPU nested loop semi/anti join using the AST condition.

Adding AST conditional support to hash-based joins which is being worked on in cudf separately, but that may take some time to realize and this may be a useful interim solution.

jbrennan333:

Closes #4309 

This is an experiment to see if replacing conditional hash joins that currently run on the cpu with conditional nested loop joins that run on the gpu will result in an overall performance improvement for tpc-ds queries, or at least to provide evidence that this improves some queries.   Note that CUDF team is working on adding support for conditional hash joins, which would be a better solution in the long run.

This initial commit only has the conversion for GpuBroadcastHashJoin -> GpuBroadcastNestedLoopJoin.  Support for the others (GpuShuffledHashJoin and GpuSortMergeJoin) will be added in later commits.


sql-plugin/src/main/311+-nondb/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
```diff
@@ -42,16 +43,60 @@ class GpuBroadcastHashJoinMeta(
...
+    }
+    def unSupportStructKeys(): Unit = if (keyDataTypes.exists(_.isInstanceOf[StructType])) {
+      this.willNotWorkOnGpu(s"$joinType joins currently do not support with struct keys")
+    }
+
+    JoinTypeChecks.tagForGpu(joinType, this)
+    joinType match {
+      case _: InnerLike =>
+      case RightOuter | LeftOuter | LeftSemi | LeftAnti =>
+        unSupportNonEqualNonAst()
```
abellina, line 78: I vote for the `unSupport*` methods to be just part of the case statement.

sql-plugin/src/main/301until310-nondb/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
```diff
@@ -68,22 +113,40 @@ class GpuBroadcastHashJoinMeta(
...
+
+    val condition = conditionMeta.map(_.convertToGpu())
+    val substituteBroadcastNestedLoopJoin: Boolean = join.joinType match {
+      case RightOuter | LeftOuter | LeftSemi | LeftAnti  =>
+        isAstCondition()
+      case _ => false
+    }
+
+    if (substituteBroadcastNestedLoopJoin) {
+      val joinExec = ShimLoader.getSparkShims.getGpuBroadcastNestedLoopJoinShim(
```
abellina, line 132: could just return here, without defining `joinExec`)

jbrennan333:
build

sql-plugin/src/main/311+-nondb/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
jbrennan333: For this commit, I was trying to minimize diffs with GpuHashJoin.tagJoin, in the hopes that once I implement the other two joins we can use the common method for all of them again.  So I will hold off on making this change for now until I can make it in the common method.

sql-plugin/src/main/301until310-nondb/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
jbrennan333: Good point.  I will change this.

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
```diff
@@ -42,16 +43,60 @@ class GpuBroadcastHashJoinMeta(
...
   override val namedChildExprs: Map[String, Seq[BaseExprMeta[_]]] =
-    JoinTypeChecks.equiJoinMeta(leftKeys, rightKeys, condition)
+    JoinTypeChecks.equiJoinMeta(leftKeys, rightKeys, conditionMeta)
 
-  override val childExprs: Seq[BaseExprMeta[_]] = leftKeys ++ rightKeys ++ condition
+  override val childExprs: Seq[BaseExprMeta[_]] = leftKeys ++ rightKeys ++ conditionMeta
 
   override def tagPlanForGpu(): Unit = {
-    GpuHashJoin.tagJoin(this, join.joinType, join.leftKeys, join.rightKeys, join.condition)
+    //GpuHashJoin.tagJoin(this, join.joinType, join.leftKeys, join.rightKeys, join.condition)
```
abellina, line 59: can this be removed?

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
```diff
@@ -42,16 +43,60 @@ class GpuBroadcastHashJoinMeta(
...
+        //  * https://github.com/NVIDIA/spark-rapids/issues/2126
+        //  * https://github.com/rapidsai/cudf/issues/7947
+        unSupportStructKeys()
+      case _ =>
+        this.willNotWorkOnGpu(s"$joinType currently is not supported")
+    }
+    // If there is an AST condition, we need to make sure the build side works for
+    // GpuBroadcastNestedLoopJoin
+    if (isAstCondition()) {
+      join.joinType match {
```
abellina, line 91: nit:
```suggestion
      joinType match {
```

there are other uses of `join.joinType` but you defined `joinType` so I'd make them all consistent.

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
```diff
@@ -69,22 +114,40 @@ class GpuBroadcastHashJoinMeta(
...
-      rightKeys.map(_.convertToGpu()),
-      join.joinType,
-      GpuJoinUtils.getGpuBuildSide(join.buildSide),
-      None,
-      left, right)
-    // The GPU does not yet support conditional joins, so conditions are implemented
-    // as a filter after the join when possible.
-    condition.map(c => GpuFilterExec(c.convertToGpu(), joinExec)).getOrElse(joinExec)
+
+    val condition = conditionMeta.map(_.convertToGpu())
```
abellina, line 125: should `convertToGpu` happen if and only if this is the correct type of join and `isAstCondition`? In other words, should it happen given `substituteBroadcastNestedLoopJoin` is true?

sql-plugin/src/main/311+-nondb/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
abellina: Thanks for the pointer @jbrennan333, agree with keeping the diffs minimal with that existing function.

abellina (changes requested):
Overall LGTM, just minor things.

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
abellina: same comment for each shim.

jbrennan333: It can.  I left it there because my intention is to have GpuHashJoin be common again after I implement the other two hash joins.    If we decide we want to do those in a separate PR, I will fix this.   Otherwise it will be fixed in later commit to this PR.

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
jbrennan333: Yeah this was another minor change to minimize diffs with GpuHashJoin.tagJoin.  I will make them consistent.

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
```diff
@@ -42,16 +43,60 @@ class GpuBroadcastHashJoinMeta(
...
+    // GpuBroadcastNestedLoopJoin
+    if (isAstCondition()) {
+      join.joinType match {
+        case LeftOuter | LeftSemi | LeftAnti if gpuBuildSide == GpuBuildLeft =>
+          willNotWorkOnGpu(s"build left not supported for conditional ${join.joinType}")
+        case RightOuter if gpuBuildSide == GpuBuildRight =>
+          willNotWorkOnGpu(s"build right not supported for conditional ${join.joinType}")
+        case _ =>
+      }
+    }
```
jlowe, line 98: Why is this conditional on having an AST condition?  I don't think we support build-side-matches-join-side for any type of join at the moment.

However since we are switching a hash-join to a nested loop join (which doesn't really have a build side), we could get tricky here and swap the build side as we swap the join implementation to allow it to be supported.

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
jbrennan333: I think it can be.  I will remove the local val and move it to where it's used.

tests/src/test/scala/com/nvidia/spark/rapids/BroadcastHashJoinSuite.scala
```diff
@@ -71,4 +71,31 @@ class BroadcastHashJoinSuite extends SparkQueryCompareTestSuite {
...
+      longsDf(spark).createOrReplaceTempView("u")
+
+      val plan1 = spark.sql(s"SELECT /*+ BROADCAST(t) */ * FROM t JOIN u ON t.longs <= u.longs")
+      val plan2 = spark.sql(s"SELECT /*+ BROADCAST(u) */ * FROM t JOIN u ON t.longs > u.longs")
+
+      // execute the plan so that the final adaptive plan is available when AQE is on
+      plan1.collect()
+
+      val finalPlan1 = findOperator(plan1.queryExecution.executedPlan,
+        _.isInstanceOf[GpuBroadcastNestedLoopJoinExecBase])
```
jlowe, line 87: This test should be integrated into the new integration tests using `assert_cpu_and_gpu_are_equal_collect_with_capture`

integration_tests/src/main/python/join_test.py
```diff
@@ -210,6 +210,27 @@ def do_join(spark):
...
 
+# Test GpuBroadcastShuffledHashJoin with condition is converted to GpuBroadcastNestedLoopJoin
+@validate_execs_in_gpu_plan('GpuBroadcastNestedLoopJoinExec')
+@ignore_order(local=True)
+@pytest.mark.parametrize('data_gen', ast_gen, ids=idfn)
+@pytest.mark.parametrize('join_type', ['Left', 'LeftSemi', 'LeftAnti'], ids=idfn)
+def test_right_broadcast_hash_join_ast_override(data_gen, join_type):
+    def do_join(spark):
+        left, right = create_ridealong_df(spark, short_gen, data_gen, 500, 50)
+        return left.join(broadcast(right), (left.b >= right.r_b), join_type)
```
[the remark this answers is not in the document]
jbrennan333: I was trying to test the conversion from GpuBroadcastHashJoin to GpuBroadcastNestedLoopJoin.  I originally had versions of these with no condition, and they were definitely using GpuBroadcastHashJoin, which I verified with: `@validate_execs_in_gpu_plan('GpuBroadcastHashJoinExec')`.  I'm not sure why I deleted those versions.  I will add them back in my next commit so you can see what I have.  They were modeled after `test_hash_join_ridealong`, and use the `_hash_join_conf`.

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
jbrennan333: I made this conditional because `GpuHashJoin.tagJoin` and `GpuBroadcastHashJoin.tagJoin` did not previously have this check (maybe I am not seeing it?).  Since it appeared to be required for `GpuBroadcastNestedLoopJoin`, I thought I should add it in the substitution case.  If I can make it unconditional, I will.  Or I can remove it and try doing the switcheroo in `convertToGpu`.

integration_tests/src/main/python/join_test.py
```diff
@@ -210,6 +213,48 @@ def do_join(spark):
         return left.join(right, left.key == right.r_key, join_type)
     assert_gpu_and_cpu_are_equal_collect(do_join, conf=_hash_join_conf)
 
+# Test GpuBroadcastHashJoin with right broadcast and no condition
+@validate_execs_in_gpu_plan('GpuBroadcastHashJoinExec')
+@ignore_order(local=True)
+@pytest.mark.parametrize('data_gen', basic_nested_gens + decimal_128_gens, ids=idfn)
+@pytest.mark.parametrize('join_type', ['Left', 'LeftSemi', 'LeftAnti'], ids=idfn)
+def test_right_broadcast_hash_join_ridealong(data_gen, join_type):
```
jlowe, line 221: Isn't this a duplicate of the existing `test_broadcast_join_right_table_ridealong` test?  I'm not sure why this PR would be adding tests for joins without any condition.

integration_tests/src/main/python/join_test.py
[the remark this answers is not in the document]
jlowe: The join condition will need to be on two key columns.  One will be testing for equality while the other needs to test for an inequality.  Therefore `create_ridealong_df` is not appropriate for this, since it only creates one key column that is joinable and another that is not (it's a nested type).

integration_tests/src/main/python/join_test.py
```diff
@@ -78,10 +78,13 @@
...
 # 3 times smaller than the other side.  So it is not likely to happen
 # unless we can give it some help.
+# To enable replacement of conditional hash joins with nested loop join,
+# set "spark.rapids.sql.replaceConditionalHashJoin.enabled" = "true"
 _hash_join_conf = {'spark.sql.autoBroadcastJoinThreshold': '160',
                    'spark.sql.join.preferSortMergeJoin': 'false',
                    'spark.sql.shuffle.partitions': '2',
-                   'spark.sql.legacy.allowNegativeScaleOfDecimal': 'true'
+                   'spark.sql.legacy.allowNegativeScaleOfDecimal': 'true',
+                   'spark.rapids.sql.replaceConditionalHashJoin.enabled': 'true'
```
abellina, line 87: I think this needs to be a test parameter (`@pytest.mark.parametrize`). We don't want to change the behavior of all tests necessarily, but test with it on and off in specific tests. Thoughts?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -562,6 +562,12 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(true)
 
+  val ENABLE_REPLACE_CONDITIONAL_HASHJOIN =
+    conf("spark.rapids.sql.replaceConditionalHashJoin.enabled")
```
abellina, line 566: indentation is a little off, but then again its off in other places in this file.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -562,6 +562,12 @@ object RapidsConf {
...
+    .doc("Allow replacing conditional hash joins with nested loop joins")
+    .booleanConf
```
abellina, line 568: I think we want to add `.internal()`

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuShuffledNestedLoopJoinBase.scala
```diff
@@ -0,0 +1,155 @@
...
+    case GpuBuildLeft => (left, right)
+    case GpuBuildRight => (right, left)
+  }
+
+  override val outputRowsLevel: MetricsLevel = ESSENTIAL_LEVEL
+  override val outputBatchesLevel: MetricsLevel = MODERATE_LEVEL
+  override lazy val additionalMetrics: Map[String, GpuMetric] = Map(
+    OP_TIME -> createNanoTimingMetric(MODERATE_LEVEL, DESCRIPTION_OP_TIME),
+    BUILD_DATA_SIZE -> createSizeMetric(ESSENTIAL_LEVEL, DESCRIPTION_BUILD_DATA_SIZE),
+    BUILD_TIME -> createNanoTimingMetric(ESSENTIAL_LEVEL, DESCRIPTION_BUILD_TIME),
```
jlowe, line 51: This is a bit odd for a nested loop join, as we're not building a hash table.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuShuffledNestedLoopJoinBase.scala
```diff
@@ -0,0 +1,155 @@
...
+
+    streamedPlan.executeColumnar().zipPartitions(buildPlan.executeColumnar()) {
+      (streamIter, buildIter) => {
+        val stIt = new CollectTimeIterator("shuffled nested loop join stream",
+          streamIter, streamTime)
+        val startTime = System.nanoTime()
+
+        withResource(ConcatAndConsumeAll.getSingleBatchWithVerification(buildIter,
+          localBuildOutput)) { builtBatch =>
+          GpuColumnVector.incRefCounts(builtBatch)
```
jlowe, line 124: This should not be here, it's getting double-closed due to the two `withResource` calls on `builtBatch`.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuShuffledNestedLoopJoinBase.scala
```diff
@@ -0,0 +1,155 @@
...
+          val delta = System.nanoTime() - startTime
+          buildTime += delta
+          buildDataSize += GpuColumnVector.getTotalDeviceMemoryUsed(builtBatch)
+
+          val spillableBuiltBatch = withResource(builtBatch) {
+            LazySpillableColumnarBatch(_, spillCallback, "built")
+          }
```
jlowe, line 131: Should not hold `builtBatch` open once we've built the spillable batch

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuBroadcastHashJoinExec.scala
```diff
@@ -42,16 +43,64 @@ class GpuBroadcastHashJoinMeta(
     join.leftKeys.map(GpuOverrides.wrapExpr(_, conf, Some(this)))
   val rightKeys: Seq[BaseExprMeta[_]] =
     join.rightKeys.map(GpuOverrides.wrapExpr(_, conf, Some(this)))
-  val condition: Option[BaseExprMeta[_]] =
+  val conditionMeta: Option[BaseExprMeta[_]] =
```
jlowe, line 46: In order to replace this with a nested loop join, we need to fold in the "equality" condition into the "inequality" condition that Spark already separated when choosing a hash join.  The nested loop join needs the entire condition which is _not_ covered by the `condition` parameter in the join node in Spark.  This needs a separate condition, like nestedLoopCondition, where we "tack-on" the equality condition to the expression.  For example if the hash join is on keys X and Y with a condition of W > Z then the `condition` will be simply `W > Z` but the nested loop join needs it to be `X == Y AND W > Z`.  There are some types that could cause the equality itself to preclude being an AST (e.g.: AST cannot compare strings even for equality today).

jbrennan333:
Running q72 with this patch on my desktop at a scale of 10GB was 61x slower than without.  So using conditional nested loop joins in place of the existing hash-join-with-post-filter makes q72 61x slower.   On my desktop the query time went from 29 seconds to 1789 seconds with the patch.

So this experiment successfully showed that replacing hash-join-with-filter with a conditional nested loop join is not a good idea for q72.  Fortunately, it looks like this cudf change: https://github.com/rapidsai/cudf/pull/9917 will make this unnecessary.

@jlowe, @abellina are there other situations where we think this substitution might be useful?

jlowe:
Given how slow this was for inner joins, it seems very unlikely this will be a good idea even for semi/anti joins that currently fallback to the CPU.  We'd have to see a significant performance improvement in libcudf nested loop join first, so I'm OK with closing this for now.

Issue #3106, opened 2021-07-30
[FEA] Use clientId to reject connection requests for peers with existing UCX endpoints


abellina:
This is a tracking issue for work that is going to go into UCX 1.12, so it won't be done anytime soon. That said, I'd like to use this to track our progress testing it.

The issue is that when we connect to peer executors there are two ways of doing that: connect to peer UcpListener, or a peer is connecting to our UcpListener. When we handle a connection from a peer we do not know anything about the remote peer (we just get a "connection request" object from UCX but it doesn't have any id). Because of this, we need to create UCX endpoints and handshake data, which can cause us to loose a race adding extra UCX endpoints. This is not a functional bug, but a resource waste we'd like to fix.

In UCX 1.12 executor A should be able to send the executorId with the connection request to a peer (B), and the request may be rejected if executor B already had initiated a request to executor A.

This is blocked by: https://github.com/openucx/ucx/pull/7136, and the JUCX jar + UCX native libraries for 1.12 being available.

PR #4357, opened 2021-12-14
Use the new ucx clientId apis

abellina:

Closes https://github.com/NVIDIA/spark-rapids/issues/3106.

This is a change to enable use of the new `clientId` in UCX 1.12.0. This also moves us to the new version of UCX. In a nutshell, this functionally prevents us from having redundant connections between peers. As the connection establishment is a race condition, since peers are connecting to each other when they receive a heartbeat or when they intend to fetch shuffle blocks, we now have enough information in the `UcpConnectionRequest` to reject it, if we already had initiated a connection to the peer that is trying to connect to us.

This is in draft. UCX 1.12 is not released yet so, I am posting to get some reviews from @petro-rudenko and to give a heads up this will go into branch-22.02 when UCX 1.12 is officially released.

### DIFF
--- a/docs/additional-functionality/rapids-shuffle.md
+++ b/docs/additional-functionality/rapids-shuffle.md
@@ -43,7 +43,7 @@ be installed on the host and inside Docker containers (if not baremetal). A host
 requirements, like the MLNX_OFED driver and `nv_peer_mem` kernel module.
 
 The minimum UCX requirement for the RAPIDS Shuffle Manager is
-[UCX 1.11.2](https://github.com/openucx/ucx/releases/tag/v1.11.2).
+[UCX 1.12.1](https://github.com/openucx/ucx/releases/tag/v1.12.1).
 
 #### Baremetal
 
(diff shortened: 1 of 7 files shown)


petro-rudenko (dismissed):

abellina:
A quick update on this PR.

- There is a race condition in the code in this PR. This is the case where there is a pending connection request from executor A to B, and also from B to A. The way the code currently works, both connections will be rejected, because both executors A and B had an outbound connection to the peer. One solution that I am looking at is to only reject if the executorId for my peer is less than (or greater than) my executor. 

- We are seeing a regression in UCX 1.12 where the progress thread is not going to sleep per the wakeup mechanism we've been using so far. This leads to inconsistent timings that hid the issue but an nsight trace shows the progress thread doing this unfortunately. This means we are going to need to wait to get this in until UCX 1.12.1 very likely.

abellina:
I have yet another race condition with this change unfortunately. Here's the scenario:

1a. Executor 1 starts an endpoint to executor 2
1b. Executor 1 sends a message using this endpoint.
1c. Executor 2 starts an endpoint to executor 1
2a. Executor 2 rejects the `ConnectionRequest` from executor 1 (1a) since it raced at creation time.
2b. Executor 1 sees an error, since the message sent in (1b) is rejected because the endpoint used ended up getting rejected.

So this should mean we need to queue messages until the executor's endpoint is "accepted", so we are guaranteed it is not going away, or may need a retry mechanism.

abellina:
Closing this for now, as it is not a high priority. We may come back to this in the future.

Issue #3849, opened 2021-10-18
[BUG] In ANSI mode we can fail in cases Spark would not due to conditionals


revans2:
**Describe the bug**
When Spark evaluates an `if` or a `case/when` function it will do lazy evaluation of the children.  So the else clause of if/else is only evaluated if the condition is false.  For the cudf version we evaluate both sides no matter what.  This is fine when an operation has no side effects. But in ANSI mode many operations can have big side effects (failing  query). This can also be true for UDFs and for `assert_true/raise_error` which we do not support yet in non-ANSI mode.

**Steps/Code to reproduce bug**
```scala
spark.conf.set("spark.sql.ansi.enabled", "true")
val df = Seq((true, 0L), (false, Long.MaxValue)).toDF("cond", "val")
df.repartition(1).selectExpr("IF(cond, val + 1, null) as my_div").show
+------+
|my_div|
+------+
|     1|
|  null|
+------+
```

The above query works just fine on the CPU because the if/else avoids executing the overflow condition. On the GPU it fails.

**Expected behavior**
It should behave the same way as the CPU

I think in the short term we can just document this, and possibly fall back to the CPU if ANSI is enabled and we encounter any conditional statements like this.  Longer term I think we need to come up with a plan on how to address this.

revans2:
Talking to @sameerz and @tgravescs I think the best way to try and fix this would be to have an optional boolean column passed with the batch when doing a project.  It would be a mask of which rows are actually being executed, and which ones are not.  That way in ANSI mode we could look at this mask and not throw exceptions if the row was not intended to be taken.  And in the case of #3855 we could just skip execution on that path and insert a null or some other default value instead as the result.

firestarman:
> This is related to #3855 because it could happen for them too.

Hi @revans2, I am working on this issue 3855, but don't fully get what you mean here. Could you share an example ?

PR #4358, opened 2021-12-14
Update GpuIf to support expressions with side effects

andygrove:

This partially addresses https://github.com/NVIDIA/spark-rapids/issues/3849.

There will be a follow-on PR to apply this to `GpuCaseWhen` since this is much more involved.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -1451,6 +1451,10 @@ case class GpuCast(
 
   import GpuCast._
 
+  // when ansi mode is enabled, some cast expressions can throw exceptions on invalid inputs
+  //TODO not all casts can throw exceptions so this needs to be more specific
+  override def hasSideEffects: Boolean = ansiMode
```
andygrove, line 1456: I will expand this to only return true for casts that can throw exceptions

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -55,6 +55,31 @@ trait GpuConditionalExpression extends ComplexTypeMergingExpression with GpuExpr
...
+
+  protected def isAllFalse(col: GpuColumnVector): Boolean = {
+    assert(BooleanType == col.dataType())
+    if (col.getRowCount == col.numNulls()) {
+      // all nulls, and null values are false values here
+      return true
+    }
+    withResource(col.getBase.any()) { anyTrue =>
+      // null values are considered false values in this context
+      !anyTrue.isValid || !anyTrue.getBoolean
```
jlowe, line 80: Per my comment on #4329, this should not need to perform the `isValid` check (and thus take another GPU synchronization hit).  The scalar can only be invalid if there are no rows or all rows are null, but both those cases will trigger the `getRowCount == numNulls` condition that is checked earlier.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -82,8 +109,142 @@ case class GpuIf(
...
+    val gpuFalseExpr = falseExpr.asInstanceOf[GpuExpression]
+    val colTypes = GpuColumnVector.extractTypes(batch)
+
+    withResource(GpuExpressionsUtils.columnarEvalToColumn(predicateExpr, batch)) { pred =>
+      if (isAllTrue(pred)) {
+        trueExpr.columnarEval(batch)
+      } else if (isAllFalse(pred)) {
+        falseExpr.columnarEval(batch)
+      } else if (gpuTrueExpr.hasSideEffects || gpuFalseExpr.hasSideEffects) {
+        withResource(GpuColumnVector.from(batch)) { tbl =>
```
jlowe, line 124: It would be nice to factor this out to a separate method so it's easier to read the `columnarEval` code at a higher level, and it also has the nice side-effect (ha!) of allowing us to give the algorithm a nice name and some comments before it explaining at a high level what's happening.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -82,8 +109,163 @@ case class GpuIf(
...
-    trueExpr, falseExpr.columnarEval(batch))
+  override def columnarEval(batch: ColumnarBatch): Any = {
+
+    val gpuTrueExpr = trueExpr.asInstanceOf[GpuExpression]
+    val gpuFalseExpr = falseExpr.asInstanceOf[GpuExpression]
+    val colTypes = GpuColumnVector.extractTypes(batch)
+
+    withResource(GpuExpressionsUtils.columnarEvalToColumn(predicateExpr, batch)) { pred =>
+      if (isAllTrue(pred)) {
+        trueExpr.columnarEval(batch)
```
firestarman, line 120: `GpuIf` is not an aggregation, so a `GpuColumnVector` is expected from the `columnarEval`. 
```suggestion
        GpuExpressionsUtils.columnarEvalToColumn(trueExpr, batch)
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -82,8 +109,163 @@ case class GpuIf(
...
+      } else if (isAllFalse(pred)) {
+        falseExpr.columnarEval(batch)
```
firestarman, line 122:
```suggestion
        GpuExpressionsUtils.columnarEvalToColumn(falseExpr, batch)
```

andygrove:
build

andygrove:
@firestarman Thanks for the suggestions. That resolved the test failures I was seeing.

andygrove:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -82,8 +108,165 @@ case class GpuIf(
...
+      batch: ColumnarBatch,
+      pred: GpuColumnVector,
+      gpuTrueExpr: GpuExpression,
+      gpuFalseExpr: GpuExpression,
+      colTypes: Array[DataType]): GpuColumnVector = {
+
+    def batchFromCv(t: GpuColumnVector, dt: DataType): ColumnarBatch = {
+      withResource(new Table(t.getBase)) { tbl =>
+        GpuColumnVector.from(tbl, Array(dt))
+      }
```
jlowe, line 171: We do not need to create a Table then throw it away to create a batch.  The ColumnarBatch can be created from the GpuColumnVector directly (e.g.: `new ColumnarBatch(Array(t), t.getRowCount.toInt)`).  However I don't see why we're creating columnar batches when what we really want are tables.  This creates a table (and throws it away) to create a columnar batch which we then pass to the new `gather` method which in turn wishes it had a table argument instead of a columnar batch.  Why not create a table and pass that to the new `gather` method directly?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -82,8 +108,165 @@ case class GpuIf(
...
+      GpuColumnVector.from(filteredData, colTypes)
+    }
+  }
+
+  private def boolToInt(cv: ColumnVector): ColumnVector = {
+    withResource(GpuScalar.from(1, DataTypes.IntegerType)) { one =>
+      withResource(GpuScalar.from(0, DataTypes.IntegerType)) { zero =>
+        cv.ifElse(one, zero)
+      }
+    }
```
jlowe, line 234: Can this just be `cv.castTo(DType.INT32)`?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -82,8 +108,165 @@ case class GpuIf(
...
+    // will perform an exclusive sum on these values and
+    // produce [0, 0, 0, 1, 1]. Combining this with the original
+    // predicate boolean array results in the two T values mapping to
+    // indices 0 and 1, respectively.
+
+    withResource(boolToInt(predicate)) { boolsAsInts =>
+      withResource(boolsAsInts.scan(
+        ScanAggregation.sum(),
+        ScanType.EXCLUSIVE,
+        NullPolicy.INCLUDE)) { prefixSumExclusive =>
```
jlowe, line 253: Is this going to generate an output null if input has a null occurring before it?  We need to interpret nulls as zeros for this, and I cannot remember how sum aggregations handle nulls.  I would expect it to return null if they are included which is not what we want here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -82,8 +108,165 @@ case class GpuIf(
...
+
+        // for the entries in the gather map that do not represent valid
+        // values to be gathered, we change the value to -MAX_INT which
+        // will be treated as null values in the gather algorithm
+        val gatherMap = withResource(GpuScalar.from(Int.MinValue, DataTypes.IntegerType)) {
```
jlowe, line 258: No need to do a type dispatch at runtime.  We know exactly what we need.
```suggestion
        val gatherMap = withResource(Scalar.fromInt(Int.MinValue)) {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
andygrove: Thanks. That certainly simplifies things.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -54,6 +55,31 @@ trait GpuConditionalExpression extends ComplexTypeMergingExpression with GpuExpr
     }
   }
 
+  protected def isAllTrue(col: GpuColumnVector): Boolean = {
```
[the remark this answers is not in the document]
andygrove: These are already in `GpuConditionalExpression`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
andygrove: I tried this and it caused a regression. I will need to look into this more to understand why.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuUserDefinedFunction.scala
```diff
@@ -41,6 +41,8 @@ trait GpuUserDefinedFunction extends GpuExpression
   /** True if the UDF is deterministic */
   val udfDeterministic: Boolean
 
+  override def hasSideEffects: Boolean = true
```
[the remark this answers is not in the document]
andygrove: I have now added this,

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -16,18 +16,19 @@
...
-import org.apache.spark.sql.types.{BooleanType, DataType}
+import org.apache.spark.sql.types.{BooleanType, DataType, DataTypes}
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
 trait GpuConditionalExpression extends ComplexTypeMergingExpression with GpuExpression
-    with ShimExpression {
+  with ShimExpression {
 
-  protected def computeIfElse(
+  def computeIfElse(
```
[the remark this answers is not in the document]
andygrove: For no good reason ... this is reverted.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
jlowe: Ah, I bet it's because `ifElse` will cause nulls to become 0 whereas casting will not.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
jlowe: I think I answered my own question based on the `boolToInt` comment.  By using `ifElse` in the implementation with `1` and `0`, it's guaranteed that the resulting column will contain no nulls so the null concern is moot.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -82,8 +108,151 @@ case class GpuIf(
...
+
+  private def boolToInt(cv: ColumnVector): ColumnVector = {
+    withResource(GpuScalar.from(1, DataTypes.IntegerType)) { one =>
+      withResource(GpuScalar.from(0, DataTypes.IntegerType)) { zero =>
+        cv.ifElse(one, zero)
+      }
+    }
+  }
+
+  private def gather(predicate: ColumnVector, t: GpuColumnVector): Table = {
```
jlowe, line 223: Nit: I think the code gets a bit cleaner if this returned a `ColumnVector` rather than a `Table`.  It's a bit weird to see a gather on a column return a table, and every caller ends up having to call `.getColumn(0)` on the resulting table since everyone wants to treat the result as just a column.  This method can `incRefCount` the column to allow closing the outer `Table` yet keep the underlying column alive to be returned.

jlowe (approved):
Minor nit, lgtm.

andygrove:
build

jlowe (approved):

PR #4361, opened 2021-12-14
Accommodate altered semantics of `cudf::lists::contains()`

mythrocks:
Depends on rapidsai/cudf/pull/9510.
Depends on rapidsai/cudf/pull/9901.

`spark-rapids` relies on [`libcudf`'s](https://github.com/rapidsai/cudf/) `cudf::lists::contains()` for its implementation of `array_contains()`.

rapidsai/cudf/pull/9510 changes the semantics of `lists::contains()`, with regard to rows containing nulls. Specifically, if a list row contains at least one null element, and is found not to contain the search key, `libcudf` will now return `false` instead of `null`.
SparkSQL expects to return `null` in those cases.

This commit accommodates the change in `libcudf's` semantics, to keep its own existing behaviour.


mythrocks:
build

mythrocks:
This PR is ready for review, although it's still in draft. It can't be merged until the dependencies in CUDF have been merged.

mythrocks:
(Modified for changed CUDF function name. `copyWithValidity` -> `copyWithBooleanColumnAsValidity`.)

mythrocks:
Re-tested after changes to dependencies in `libcudf`. Will rebuild after rapidsai/cudf/pull/9510 is merged.

integration_tests/src/main/python/array_test.py
```diff
@@ -99,10 +99,15 @@ def test_orderby_array_of_structs(data_gen):
 def test_array_contains(data_gen):
     arr_gen = ArrayGen(data_gen)
     lit = gen_scalar(data_gen, force_no_nulls=True)
-    assert_gpu_and_cpu_are_equal_collect(lambda spark: two_col_df(
-        spark, arr_gen, data_gen).select(array_contains(col('a'), lit.cast(data_gen.data_type)),
-                                         array_contains(col('a'), col('b')),
-                                         array_contains(col('a'), col('a')[5])), no_nans_conf)
+
+    def get_input(spark):
```
revans2, line 103: nit: I always feel much better about a fixes like this if the tests weren't changed. They look to be identical to before, but with a lot of formatting changes.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -223,17 +223,43 @@ case class GpuArrayContains(left: Expression, right: Expression)
...
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuScalar): ColumnVector = {
+    val contains = lhs.getBase.listContains(rhs.getBase)
+    withResource(contains) { containsCV =>
+      val containsNull = lhs.getBase.listContainsNulls()
+      withResource(containsNull) { containsNullCV =>
+        val notContainsNull = containsNullCV.not
+        withResource(notContainsNull) { notContainsNullCV =>
+          val validity = containsCV.or(notContainsNullCV)
+          withResource(validity) { validityCV =>
+            containsCV.copyWithBooleanColumnAsValidity(validityCV)
```
revans2, line 235: nit: we could save some memory by closing some of these sooner. Also once `listContains` returns everything after that is common. We should probably turn that into a separate method.

revans2 (commented):
Just a few minor nits

integration_tests/src/main/python/array_test.py
mythrocks: Thanks for reviewing, @revans2.
I'd be willing to undo this change, if we could fix the original formatting. I found the previous formatting very hard to parse.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
mythrocks: Oh, that's an interesting thought. I'll take a crack at it.

integration_tests/src/main/python/array_test.py
revans2: Like I said it is a nit. If you have strong feelings just keep it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
mythrocks: Sorry for the delay. Needed to get my head around `withResource()`.
Now, intermediate results are closed early. A new helper function has been added.

mythrocks:
rapidsai/cudf/pull/9510 was just committed. I've pulled this PR out of draft.

revans2 (approved):

revans2:
build

revans2:
Even though rapidsai/cudf/pull/9510 has not gone all the way through CI yet. I think we can still test/commit this. Because the extra processing should be a noop if listContains is behaving the old way too.

revans2:
oops I was wrong. It needs a new API that was missing before.

tgravescs:
build

mythrocks:
Merging this to head off impending CI breakage.

mythrocks:
Thank you for the review, @revans2. @tgravescs, thank you for confirming that the CI has passed with the new `libcudf` artifact.

Issue #3934, opened 2021-10-27
[BUG] Dependencies of published integration tests jar are missing


jlowe:
Attempting to download the integration tests jar as a dependency fails with missing dependencies:
```
$ mvn dependency:get -B -Ddest=snap -DgroupId=com.nvidia -DartifactId=rapids-4-spark-integration-tests_2.12 -Dversion=21.08.0 -Dpackaging=jar
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:2.8:get (default-cli) on project rapids-4-spark-parent: Couldn't download artifact: Missing:
[ERROR] ----------
[ERROR] 1) com.nvidia:rapids-4-spark-tests_2.12:test-jar:tests:21.08.0
[ERROR] 
[ERROR]   Try downloading the file manually from the project website.
[ERROR] 
[ERROR]   Then, install it using the command: 
[ERROR]       mvn install:install-file -DgroupId=com.nvidia -DartifactId=rapids-4-spark-tests_2.12 -Dversion=21.08.0 -Dclassifier=tests -Dpackaging=test-jar -Dfile=/path/to/file
[ERROR] 
[ERROR]   Alternatively, if you host your own repository you can deploy the file there: 
[ERROR]       mvn deploy:deploy-file -DgroupId=com.nvidia -DartifactId=rapids-4-spark-tests_2.12 -Dversion=21.08.0 -Dclassifier=tests -Dpackaging=test-jar -Dfile=/path/to/file -Durl=[url] -DrepositoryId=[id]
[ERROR] 
[ERROR]   Path to dependency: 
[ERROR]   	1) org.apache.maven.plugins:maven-downloader-plugin:jar:1.0
[ERROR]   	2) com.nvidia:rapids-4-spark-integration-tests_2.12:jar:21.08.0
[ERROR]   	3) com.nvidia:rapids-4-spark-tests_2.12:test-jar:tests:21.08.0
[ERROR] 
[ERROR] 2) com.nvidia:rapids-4-spark-udf-examples_2.12:jar:21.08.0
[ERROR] 
[ERROR]   Try downloading the file manually from the project website.
[ERROR] 
[ERROR]   Then, install it using the command: 
[ERROR]       mvn install:install-file -DgroupId=com.nvidia -DartifactId=rapids-4-spark-udf-examples_2.12 -Dversion=21.08.0 -Dpackaging=jar -Dfile=/path/to/file
[ERROR] 
[...]

PR #4365, opened 2021-12-15
Not publishing integration_tests jar to Maven Central [skip ci]

NvTimLiu:
Fix issue: https://github.com/NVIDIA/spark-rapids/issues/3934

Stop publishing the integration tests jar to Maven Central.

Remove the aggregator dependency in the integration_tests module pom.xml.

### DIFF
--- a/integration_tests/pom.xml
+++ b/integration_tests/pom.xml
@@ -64,6 +64,14 @@
             <version>${project.version}</version>
             <scope>provided</scope>
         </dependency>
+        <!-- use aggregator jar because accessing internal classes -->
+        <dependency>
+            <groupId>com.nvidia</groupId>
+            <artifactId>rapids-4-spark-aggregator_${scala.binary.version}</artifactId>
+            <version>${project.version}</version>
+            <classifier>${spark.version.classifier}</classifier>
+            <scope>test</scope>
+        </dependency>
         <dependency>
             <groupId>com.nvidia</groupId>
             <artifactId>rapids-4-spark-udf-examples_${scala.binary.version}</artifactId>
--- a/integration_tests/pom.xml
+++ b/integration_tests/pom.xml
@@ -64,14 +64,6 @@
             <version>${project.version}</version>
             <scope>provided</scope>
         </dependency>
-        <!-- use aggregator jar because accessing internal classes -->
-        <dependency>
-            <groupId>com.nvidia</groupId>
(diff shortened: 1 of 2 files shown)


NvTimLiu:
build

tgravescs:
I thought aggregator jar was needed here because we picked up internal classes.  This PR doesn't change that so that changed somewhere else?  I guess we will see if it builds

integration_tests/pom.xml
```diff
@@ -64,14 +64,6 @@
             <version>${project.version}</version>
             <scope>provided</scope>
         </dependency>
-        <!-- use aggregator jar because accessing internal classes -->
-        <dependency>
-            <groupId>com.nvidia</groupId>
-            <artifactId>rapids-4-spark-aggregator_${scala.binary.version}</artifactId>
```
jlowe, line 70: This alone does not remove the dependency.  I think it's a step in the right direction, but rapids-4-spark-udf-examples and rapids-4-spark-tests are still here and both pull in the aggregator jar as a dependency.  Thus this will transitively depend on the aggregator jars, so the issue is still not fixed.  We would need to sever the connection to these other two modules (or update them so they no longer need the aggregator jars).

jlowe:
> I guess we will see if it builds

It likely will because the aggregator jar will still be picked up transitively by other dependencies (e.g.: the tests module dependency) during the test build.

integration_tests/pom.xml
GaryShen2008: I think we don't need to change it in this PR.
This PR is just to stop publishing the integration jar in release building.

NvTimLiu: fixed

NvTimLiu:
Only updated the public release script, skip ci

NvTimLiu:
build

jlowe (approved):

tgravescs:
so what are our IT jobs going to do use to test the staging jars?   I thought part of the reason we are pushing the IT tests was so we could have jobs that tested them without having to do a build.

GaryShen2008:
> 

We can deploy the IT jar to internal maven repo to be used for testing.

tgravescs:
ok, do we need to update any jobs for that or are they not setup to pull from staging anyway?

GaryShen2008:
> ok, do we need to update any jobs for that or are they not setup to pull from staging anyway?

Yes, Tim will help to change the deploy script and the related test jobs.

tgravescs (approved):

PR #4367, opened 2021-12-15
update 21.12 branch for doc [skip ci]

viadea:
update 21.12branch for doc
Signed-off-by: Hao Zhu <hazhu@nvidia.com>


jlowe (dismissed):

docs/get-started/getting-started-on-prem.md
```diff
@@ -56,7 +56,7 @@ CUDA and will not run on other versions. The jars use a maven classifier to keep
 
 For example, here is a sample version of the jars and cudf with CUDA 11.0 support:
 - cudf-21.12.0-cuda11.jar
-- rapids-4-spark_2.12-21.12.0-SNAPSHOT.jar
+- rapids-4-spark_2.12-21.12.0.jar
```
jlowe, line 59: Other docs aren't removing SNAPSHOT but this one is

viadea: Hi Jason, this is used by the example I think we should change it right?

jlowe: The other occurrences where you removed `SNAPSHOT` are also examples, so what makes this example different than the others?  Note that when one builds on branch-21.12, they will get a SNAPSHOT jar and arguably we should use SNAPSHOT here.  Docs on the `main` branch would not, and there's separate commits, only to `main`, that remove `SNAPSHOT` from both docs and the pom versions.

viadea: Sure @jlowe  i just added the snapshot back.

jlowe (approved):

sameerz (approved):
In a future release we ought to consider updating the docs/demo/Databricks/generate-init-script-cuda11.ipynb to use a more recent toolkit and driver (at least driver 450.80.02 or higher).

Issue #3512, opened 2021-09-16
[FEA] Support org.apache.spark.sql.catalyst.expressions.Sequence


viadea:
**Is your feature request related to a problem? Please describe.**
I wish we can support org.apache.spark.sql.catalyst.expressions.Sequence which is for "Sequence" function.

**Describe the solution you'd like**
"Sequence" function can be supported.

**Describe alternatives you've considered**

**Additional context**

wbo4958:
@viadea Do you have any code for this feature?

wbo4958:
For now, Plugin has supported 

``` scala
spark.sql("select sequence(-1, 2, 1)")
```

and has not supported

``` scala
spark.sql("select sequence(start_col, end_col, step_col) from table").show()
```

I just filed an cudf dependency issue https://github.com/rapidsai/cudf/issues/9424

PR #4376, opened 2021-12-16
Add sequence support [databricks]

wbo4958:
This is to close https://github.com/NVIDIA/spark-rapids/issues/3512. and this PR depends on https://github.com/rapidsai/cudf/pull/9839 and https://github.com/NVIDIA/spark-rapids/pull/4376

For now, the PR only supports sequence on IntegerType.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -3293,6 +3293,16 @@ object GpuOverrides extends Logging {
       (a, conf, p, r) => new ExprMeta[CreateMap](a, conf, p, r) {
         override def convertToGpu(): GpuExpression = GpuCreateMap(childExprs.map(_.convertToGpu()))
       }
+    ),
+    expr[Sequence](
+      desc = "Sequence",
+      ExprChecks.projectOnly(
+        TypeSig.ARRAY.nested(TypeSig.INT), TypeSig.ARRAY.nested(TypeSig.INT + TypeSig.TIMESTAMP +
+          TypeSig.DATE),
+        Seq(ParamCheck("start", TypeSig.INT, TypeSig.INT),
+          ParamCheck("stop", TypeSig.INT, TypeSig.INT)),
+        Some(RepeatingParamCheck("step", TypeSig.INT, TypeSig.INT))),
+      (a, conf, p, r) => new GpuSequenceMeta(a, conf, p, r)
     )
   ).map(r => (r.getClassFor.asSubclass(classOf[Expression]), r)).toMap
 
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -3297,11 +3297,13 @@ object GpuOverrides extends Logging {
     expr[Sequence](
       desc = "Sequence",
       ExprChecks.projectOnly(
-        TypeSig.ARRAY.nested(TypeSig.INT), TypeSig.ARRAY.nested(TypeSig.INT + TypeSig.TIMESTAMP +
+        TypeSig.ARRAY.nested(TypeSig.integral), TypeSig.ARRAY.nested(TypeSig.integral +
+          TypeSig.TIMESTAMP + TypeSig.DATE),
+        Seq(ParamCheck("start", TypeSig.integral, TypeSig.integral + TypeSig.TIMESTAMP +
           TypeSig.DATE),
-        Seq(ParamCheck("start", TypeSig.INT, TypeSig.INT),
-          ParamCheck("stop", TypeSig.INT, TypeSig.INT)),
-        Some(RepeatingParamCheck("step", TypeSig.INT, TypeSig.INT))),
+          ParamCheck("stop", TypeSig.integral, TypeSig.integral + TypeSig.TIMESTAMP +
+            TypeSig.DATE)),
+        Some(RepeatingParamCheck("step", TypeSig.integral, TypeSig.integral + TypeSig.CALENDAR))),
       (a, conf, p, r) => new GpuSequenceMeta(a, conf, p, r)
     )
   ).map(r => (r.getClassFor.asSubclass(classOf[Expression]), r)).toMap
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, NVIDIA CORPORATION.
+ * Copyright (c) 2021-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -444,13 +444,12 @@ case class GpuSequence(start: Expression, stop: Expression, timeZoneId: Option[S
   override def withTimeZone(timeZoneId: String): TimeZoneAwareExpression =
     copy(timeZoneId = Some(timeZoneId))
 
-  /** Calculate the size between start and stop, both inclusive
-   * |stop - start| + 1 */
-
   /**
-   * Calculate the size between start and stop both inclusive and the step
-   * @param start
-   * @param stop
+   * Calculate the size and step (1 or -1) between start and stop both inclusive
+   * size = |stop - start| + 1
+   * step = 1 if stop >= start else -1
+   * @param start first values in the result sequences
+   * @param stop end values in the result sequences
    * @return (size, step)
    */
   private def calculateSizeAndStep(start: BinaryOperable, stop: BinaryOperable, dt: DataType):
@@ -502,7 +501,7 @@ case class GpuSequence(start: Expression, stop: Expression, timeZoneId: Option[S
 /** GpuSequence with step */
 case class GpuSequenceWithStep(start: Expression, stop: Expression, step: Expression,
     timeZoneId: Option[String] = None) extends GpuTernaryExpression with TimeZoneAwareExpression
-    with GpuSequenceTrait {
+  with GpuSequenceTrait {
 
   override def withTimeZone(timeZoneId: String): TimeZoneAwareExpression =
     copy(timeZoneId = Some(timeZoneId))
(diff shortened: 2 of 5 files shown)


revans2 (commented):
Overall looking really good. I assume all of this issues I am pointing out are just because this is still a work in progress.

wbo4958:
> Overall looking really good. I assume all of this issues I am pointing out are just because this is still a work in progress.

@revans2, Thx for the reviewing. Yeah, this PR is still WIP, but it can work for IntegerType. Will refine this and add more types for supporting. But I'd not like add TimeStamp and DateType for supporting in this PR, since the size calculation may be quite different which may cause this PR pretty big.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -392,4 +392,236 @@ case class GpuArrayMax(child: Expression) extends GpuBaseArrayAgg with ImplicitC
...
+  /**
+   * Calculate the size between start and stop both inclusive and the step
+   * @param start
+   * @param stop
+   * @return (size, step)
+   */
+  private def calculateSizeAndStep(start: BinaryOperable, stop: BinaryOperable):
+      Seq[ColumnVector] = {
+    withResource(stop.sub(start)) { difference =>
+      withResource(Scalar.fromInt(1)) { scalarOne =>
```
ttnghia, line 440: Should be `GpuScalar(1, dataType)` or similar, so we can support various types not just integer.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -392,4 +392,236 @@ case class GpuArrayMax(child: Expression) extends GpuBaseArrayAgg with ImplicitC
...
+              withResource(sizeWithNegative.greaterOrEqualTo(scalarZero)) { pred =>
+                withResource(pred.ifElse(sizeWithNegative, scalarZero)) { tmpSize =>
+                  // when start==stop, step==0, size will be 0. but we should change size to 1
+                  withResource(difference.equalTo(scalarZero)) { diffHasZero =>
+                    step match {
+                      case stepScalar: Scalar =>
+                        withResource(ColumnVector.fromScalar(stepScalar, rows)) { stepV =>
+                          withResource(stepV.equalTo(scalarZero)) { stepHasZero =>
+                            withResource(diffHasZero.and(stepHasZero)) { predWithZero =>
+                              predWithZero.ifElse(scalarOne, tmpSize)
```
ttnghia, line 523: Is there any way to get rid of such nested `withResource`? Otherwise, this looks so cluster.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -392,4 +392,267 @@ case class GpuArrayMax(child: Expression) extends GpuBaseArrayAgg with ImplicitC
...
+      dt: DataType): ColumnVector = {
+    // First, calculate sizeWithNegative=floor((stop-start)/step)+1.
+    //     if step = 0, the div operation in cudf will get MIN_VALUE, which is ok for this case,
+    //     since when size < 0, cudf will not generate sequence
+    // Second, calculate size = if(sizeWithNegative < 0) 0 else sizeWithNegative
+    // Third, if (start == stop && step == 0), let size = 1.
+    withResource(stop.sub(start)) { difference =>
+      withResource(difference.floorDiv(step)) { quotient =>
+        withResource(numberScalar(dt, 1)) { one =>
+          withResource(quotient.add(one)) { sizeWithNegative =>
```
jlowe, line 537: When `withResource` nests this deeply, that's usually an indication that we're holding onto one or more GPU results longer than necessary, adding undesired and avoidable memory pressure.  For example, we compute `quotient` here and only need it to compute `sizeWithNegative`, yet we hold onto the GPU memory for the `quotient` result until after the entire calculation completes.  The memory can be freed earlier with something like this:
```scala
  withResource(numberScalar(dt, 1)) { one =>
    val sizeWithNegative = withResource(difference.floorDiv(step)) { quotient =>
      quotient.add(one)
    }
    withResource(sizeWithNegative) { sizeWithNegative =>
      ....
```

ttnghia:
FYI: cudf PR has been merged.

wbo4958:
> 

Thx for the information

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -392,4 +392,236 @@ case class GpuArrayMax(child: Expression) extends GpuBaseArrayAgg with ImplicitC
...
+    expr: Sequence,
+    conf: RapidsConf,
+    parent: Option[RapidsMeta[_, _, _]],
+    rule: DataFromReplacementRule)
+  extends ExprMeta[Sequence](expr, conf, parent, rule) {
+
+  override def convertToGpu(): GpuExpression = {
+    if (expr.stepOpt.isDefined) {
+      val Seq(start, stop, step) = childExprs.map(_.convertToGpu())
+      GpuSequenceWithStep(start, stop, step, expr.timeZoneId)
```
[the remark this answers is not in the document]
wbo4958: Done.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -392,4 +392,236 @@ case class GpuArrayMax(child: Expression) extends GpuBaseArrayAgg with ImplicitC
...
+   * Calculate the size between start and stop both inclusive and the step
+   * @param start
+   * @param stop
+   * @return (size, step)
+   */
+  private def calculateSizeAndStep(start: BinaryOperable, stop: BinaryOperable):
+      Seq[ColumnVector] = {
+    withResource(stop.sub(start)) { difference =>
+      withResource(Scalar.fromInt(1)) { scalarOne =>
+        val step = withResource(Scalar.fromInt(-1)) { scalarNegativeOne =>
```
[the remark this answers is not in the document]
wbo4958: Thx. Done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -392,4 +392,236 @@ case class GpuArrayMax(child: Expression) extends GpuBaseArrayAgg with ImplicitC
...
+        ColumnVector.sequence(start.getBase, size, stepV)
+      }
+    }
+
+  override def doColumnar(
+      numRows: Int,
+      val0: GpuScalar,
+      val1: GpuScalar,
+      val2: GpuScalar): ColumnVector =
+    throw new IllegalStateException("This is not supported yet")
```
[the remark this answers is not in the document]
wbo4958: Hmm, Could you add the support when we come across this kind of odd corner cases? since I don't know how to simulate this corner case.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
wbo4958: Done. Thx

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
wbo4958: Hmm, refined this.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
wbo4958: Thx @jlowe. Changed this accordingly.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3293,6 +3293,16 @@ object GpuOverrides extends Logging {
...
+    ),
+    expr[Sequence](
+      desc = "Sequence",
+      ExprChecks.projectOnly(
+        TypeSig.ARRAY.nested(TypeSig.INT), TypeSig.ARRAY.nested(TypeSig.INT + TypeSig.TIMESTAMP +
+          TypeSig.DATE),
+        Seq(ParamCheck("start", TypeSig.INT, TypeSig.INT),
+          ParamCheck("stop", TypeSig.INT, TypeSig.INT)),
+        Some(RepeatingParamCheck("step", TypeSig.INT, TypeSig.INT))),
+      (a, conf, p, r) => new GpuSequenceMeta(a, conf, p, r)
```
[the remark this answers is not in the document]
wbo4958: Thx. For now, this PR is only supporting integral type. I will file the following issue for TimeStamp and Date types supporting for @firestarman who will also support it for different spark versions.

wbo4958:
build

wbo4958:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -392,4 +392,278 @@ case class GpuArrayMax(child: Expression) extends GpuBaseArrayAgg with ImplicitC
...
+      val Seq(start, stop, step) = childExprs.map(_.convertToGpu())
+      GpuSequenceWithStep(start, stop, step, expr.timeZoneId)
+    } else {
+      val Seq(start, stop) = childExprs.map(_.convertToGpu())
+      GpuSequence(start, stop, expr.timeZoneId)
+    }
+  }
+}
+
+trait GpuSequenceTrait {
```
revans2, line 421: nit: Why is this a trait and not an object?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -392,4 +392,278 @@ case class GpuArrayMax(child: Expression) extends GpuBaseArrayAgg with ImplicitC
...
+
+  override def first: Expression = start
+
+  override def second: Expression = stop
+
+  override def third: Expression = step
+
+  override def dataType: DataType = ArrayType(start.dataType, containsNull = false)
+
+  /* TODO do we need to check below conditions?
```
revans2, line 517: Is this still a TODO?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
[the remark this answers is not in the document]
revans2: Just turn `val0`/`start` into a `GpuColumnVector` and then call the other `doColumnar`.  That way you don't really need to test it that much.  The corner cases show up when a constant is injected into a plan after constant folding has finished, this happens some times but it is hard to reproduce.  It can also happen when some operations, like `GpuCoalesce` where Spark cannot determine that a constant value is always returned by the coalesce, but we can.

For example...
```
sequence(coalesce(100, some_column), 200, 10)
```

I think will cause this to happen. This is the same thing we need for the version of sequence without a step.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
wbo4958: Done.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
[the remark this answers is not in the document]
wbo4958: Thx. Done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
wbo4958: Done. Thx

wbo4958:
build

wbo4958:
build

pxLi:
build

revans2 (approved):

PR #4374, opened 2021-12-16
Fix the IT module depends on the tests module

res-life:
Fix the IT module depends on the tests module
This contributes #3934 

Signed-off-by: Chong Gao <res_life@163.com>

PR #4382, opened 2021-12-17
Remove unused scallop dependency from integration_tests

jlowe:
While reviewing #4374 I noticed the integration tests have a dependency on scallop that is no longer used.  This cleans up the unused dependency.


jlowe:
build

tgravescs (approved):

PR #4383, opened 2021-12-17
Fix correctness issue with CASE WHEN with expressions that have side-effects

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/3849

Detect when a `CASE WHEN` has expressions that have side-effects and only evaluate them for the relevant rows. This follows on from similar work in https://github.com/NVIDIA/spark-rapids/pull/4358 for `IF ELSE` expressions.

There are optimizations to stop evaluating branches if every row has already caused a previous branch predicate to evaluate to true. The scala tests hit these cases and I tested manually by adding debug logging to confirm that the optimizations took effect.


andygrove:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +306,132 @@ case class GpuCaseWhen(
...
+
+        // iterate over the WHEN THEN branches first
+        branches.foreach {
+          case (whenExpr, thenExpr) =>
+            // evaluate the WHEN predicate
+            withResource(GpuExpressionsUtils.columnarEvalToColumn(whenExpr, batch)) { whenBool =>
+              // we only want to evaluate where this WHEN is true and no previous WHEN has been true
+              val whenBoolAdjusted: GpuColumnVector = cumulativePred match {
+                case Some(prev) =>
+                  withResource(prev.getBase.not()) { notPreviouslyTrue =>
```
jlowe, line 349: There's an early-out opportunity where `notPreviouslyTrue` is all false.  In that case all further branches can be avoided since all inputs have been covered by previous branches.  It's tempting to test for that here, but we already evaluated `whenBool` by this point, and that would be wasted computation if we early-out.  It's also tempting to evaluate it after we update `cumulativePred` but that would be wasteful if this is the last branch to evaluate.  So a bit tricky to implement, but I could see it being very beneficial in many case when statements, especially when they are elaborate.

andygrove: I pushed an optimization for this, checking `cumulativePred` after we set it. Even if we are processing the last `WHEN` branch, we can still skip the `ELSE` branch which comes after the main loop.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +306,139 @@ case class GpuCaseWhen(
...
+                    currentValue = Some(currentValue match {
+                      case Some(v) =>
+                        withResource(v) { _ =>
+                          GpuColumnVector.from(whenBoolAdjusted.getBase.ifElse(gather, v.getBase),
+                            dataType)
+                        }
+                      case _ =>
+                        GpuColumnVector.from(gather.incRefCount(), dataType)
+                    })
+                  }
```
jlowe, line 378: Nit: I'm not a fan of putting a `match` directly in the parameter list of something -- it adds quite a bit of complexity to reading the line.  I think this could be a good candidate for refactoring, as there's a core piece here that is useful and easy to understand in isolation, something like the following (my names probably could be better)
```scala
private def updateCumulativeResult(
    currentValue: Option[GpuColumnVector],
    thenValues: GpuColumnVector,
    gatherMask: ColumnVector): GpuColumnVector = {
  withResource(gather(gatherMask, thenValues)) { gather =>
    currentValue = Some(currentValue match {
      case Some(v) =>
        withResource(v) { _ =>
          GpuColumnVector.from(gatherMask.ifElse(gather, v.getBase), dataType)
        }
      case _ =>
        GpuColumnVector.from(gather.incRefCount(), dataType)
    }
  }
}

[...]

currentValue = Some(updateCumulativeResult(currentValue, whenBoolAdjusted.getBase, thenValues))
```

Based on the `thenValues` lifetime comment, you may want this method to be responsible for closing the `thenValue` if refactoring in this way.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +306,139 @@ case class GpuCaseWhen(
...
+
+                  cumulativePred = Some(cumulativePred match {
```
jlowe, line 380: Similar nit-and-refactor suggestion here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +306,139 @@ case class GpuCaseWhen(
...
+                    // no need to process any more branches or the else condition
+                    return currentValue.get.incRefCount()
+                  }
+                }
+              }
+            }
+        }
+
+        // invert the cumulative predicate to get the ELSE predicate
+        withResource(cumulativePred.get.getBase.not()) { elsePredicate =>
```
jlowe, line 400: We should extract the predicate, throw if it's not there, and close it early after evaluating the `not`.  Otherwise we hold onto it during the entire `ELSE` evaluation for no benefit.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +306,139 @@ case class GpuCaseWhen(
...
+          // replace null predicates with true (because this is an inverted predicate)
+          val elsePredNoNulls = withResource(Scalar.fromBool(true)) { t =>
+            GpuColumnVector.from(elsePredicate.replaceNulls(t), DataTypes.BooleanType)
```
jlowe, line 403: Similarly we're holding onto `elsePredicate` here longer than necessary.  It can be closed after this is evaluated.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +306,139 @@ case class GpuCaseWhen(
...
+                  // if this WHEN predicate is true for all rows and no previous predicate has
+                  // been true then we can return immediately
+                  return GpuExpressionsUtils.columnarEvalToColumn(thenExpr, batch)
+                }
+                val thenValues = withResource(filterBatch(tbl,
+                    whenBoolAdjusted.getBase, colTypes)) {
+                  trueBatch => GpuExpressionsUtils.columnarEvalToColumn(thenExpr, trueBatch)
+                }
+                withResource(thenValues) { _ =>
+                  withResource(gather(whenBoolAdjusted.getBase, thenValues)) { gather =>
```
jlowe, line 368: `thenValues` should be closed after evaluating the `gather` as it's no longer needed after that.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +306,139 @@ case class GpuCaseWhen(
...
+                  }
+                }
+              }
+            }
+        }
+
+        // invert the cumulative predicate to get the ELSE predicate
+        withResource(cumulativePred.get.getBase.not()) { elsePredicate =>
+          // replace null predicates with true (because this is an inverted predicate)
+          val elsePredNoNulls = withResource(Scalar.fromBool(true)) { t =>
```
jlowe, line 402: Nit: If we're going to create scalars anyway, seems like we could avoid some extra kernels and intermediate values here by creating an inverted form of `boolToInt`, e.g.:
```scala
  private def boolToIntInverted(cv: ColumnVector): ColumnVector = {
    withResource(GpuScalar.from(1, DataTypes.IntegerType)) { one =>
      withResource(GpuScalar.from(0, DataTypes.IntegerType)) { zero =>
        cv.ifElse(zero, one)
      }
    }
  }
```
Then we can call `boolToIntInverted` on `cumulativePred` and directly get the else mask we want rather than needing to invert and overwrite nulls to get there.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +306,139 @@ case class GpuCaseWhen(
...
+              case Some(expr) =>
+                if (isAllTrue(elsePredNoNulls)) {
+                  GpuExpressionsUtils.columnarEvalToColumn(expr, batch)
+                } else {
+                  val elseValues = withResource(
+                      filterBatch(tbl, elsePredNoNulls.getBase, colTypes)) {
+                    elseBatch => GpuExpressionsUtils.columnarEvalToColumn(expr, elseBatch)
+                  }
+                  withResource(elseValues) { _ =>
+                    withResource(gather(elsePredNoNulls.getBase, elseValues)) { gather =>
```
jlowe, line 416: `elseValues` should be closed after the `gather`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +306,139 @@ case class GpuCaseWhen(
...
+        // invert the cumulative predicate to get the ELSE predicate
+        withResource(cumulativePred.get.getBase.not()) { elsePredicate =>
+          // replace null predicates with true (because this is an inverted predicate)
+          val elsePredNoNulls = withResource(Scalar.fromBool(true)) { t =>
+            GpuColumnVector.from(elsePredicate.replaceNulls(t), DataTypes.BooleanType)
+          }
+          withResource(elsePredNoNulls) { _ =>
+            elseValue match {
+              case Some(expr) =>
+                if (isAllTrue(elsePredNoNulls)) {
```
jlowe, line 408: We can make this early-out even cheaper by moving it up earlier and avoiding the `elsePredicate` computation shenanigans.  We can check for `isAllFalse(cumulativePred)` to know that everything is the `ELSE` clause (or nulls if no `ELSE` clause).

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +306,139 @@ case class GpuCaseWhen(
...
+                        gather, currentValue.get.getBase), dataType)
+                    }
+                  }
+                }
+
+              case None =>
+                // if there is no ELSE condition then we return NULL for any rows not matched by
+                // previous branches
+                withResource(GpuScalar.from(null, dataType)) { nullScalar =>
+                  if (isAllTrue(elsePredNoNulls) && elsePredNoNulls.getRowCount <= Int.MaxValue) {
```
jlowe, line 427: I don't understand the reason for the `getRowCount` check here or how it could ever evaluate to false.

andygrove: My rationale here was that `getRowCount` returns a `Long` and we need to pass it to `GpuColumnVector.from` which requires `Int`. If in the future we supported row counts > `Int.MaxValue` then the call to `toInt` could be unsafe?

jlowe: If we support row counts > `Int.MaxValue` in the future then `GpuColumnVector.from` would also need to support the same.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +317,170 @@ case class GpuCaseWhen(
...
+  /**
+   * Determine for each row whether this is the first WHEN predicate so far to evaluate to true
+   */
+  private def isFirstTrueWhen(
+      cumulativePred: Option[GpuColumnVector],
+      whenBool: GpuColumnVector): GpuColumnVector = {
+    cumulativePred match {
+      case Some(prev) =>
+        withResource(prev.getBase.not()) { notPreviouslyTrue =>
+          GpuColumnVector.from(whenBool.getBase.and(notPreviouslyTrue), DataTypes.BooleanType)
```
jlowe, line 480: Similar concern with nulls here.  If `prev` could contain nulls then `not()` will still be null and then `and` will result in a null which is undesired if `whenBool` contains a `true` in the corresponding row.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +317,170 @@ case class GpuCaseWhen(
...
+   * `prevPredicate OR thisPredicate`.
+   */
+  private def calcCumulativePredicate(
+      cumulativePred: Option[GpuColumnVector],
+      whenBool: GpuColumnVector,
+      firstTrueWhen: GpuColumnVector): GpuColumnVector = {
+    cumulativePred match {
+      case Some(prev) =>
+        withResource(prev) { _ =>
+          GpuColumnVector.from(whenBool.getBase.or(prev.getBase), DataTypes.BooleanType)
```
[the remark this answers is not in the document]
andygrove: I've pushed a fix for this. This is the first time I have used the cuDF AST support so hopefully, I am using it correctly.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -301,12 +315,191 @@ case class GpuCaseWhen(
...
+      whenBool: GpuColumnVector): GpuColumnVector = {
+    cumulativePred match {
+      case Some(prev) =>
+        val notPreviouslyTrue = withResource(prev.getBase.not()) { previouslyFalse =>
+          withResource(prev.getBase.isNull) { previouslyNull =>
+            previouslyFalse.or(previouslyNull)
+          }
+        }
+        val result = withResource(new Table(whenBool.getBase, notPreviouslyTrue)) { t =>
+          val expr = new ast.BinaryOperation(ast.BinaryOperator.NULL_LOGICAL_AND,
```
jlowe, line 492: This won't work because `whenBool` can contain nulls, and null AND something is always null, even with the AST LOGICAL_AND.  I think we can avoid AST and leverage what we've already done previously by doing something like this.  boolInverted has the nice property of treating nulls like they were originally false.
```scala
  withResource(boolInverted(prev.getBase)) { notPrev =>
    val whenReplaced = withResource(Scalar.fromBool(false)) { falseScalar =>
      whenBool.getBase.replaceNulls(falseScalar)
    }
    GpuColumnVector.from(whenReplaced.and(notPrev), DataTypes.BooleanType)
  }
```

andygrove: Thanks. I added tests to hit the cases where predicates can evaluate to null and confirmed that they failed with the original code and pass with the latest code (including the above suggestion).

jlowe (approved):

Issue #4104, opened 2021-11-15
[FEA] Support org.apache.spark.sql.catalyst.expressions.ReplicateRows


viadea:
**Is your feature request related to a problem? Please describe.**

This is a feature request to support `org.apache.spark.sql.catalyst.expressions.ReplicateRows`.
ReplicateRows is an internal function used by optimizer to rewrite EXCEPT ALL and INTERSECT ALL queries as per the [Spark source code](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/generators.scala).

Below is a mini repro:
```
import org.apache.spark.sql.Row
import org.apache.spark.sql.types._
import org.apache.spark.sql.functions._

val data = Seq(
    Row(Row("Adam ","","Green"),"1","M",1000.1, "2019-01-01",List("Java","Scala")),
    Row(Row("Bob ","Middle","Green"),"2","M",2000.2, "2019-01-02",List("Java","Python")),
    Row(Row("Cathy ","","Green"),"3","F",3000.3, "2019-01-03",List())
)

val schema = (new StructType()
  .add("name",new StructType()
    .add("firstname",StringType)
    .add("middlename",StringType)
    .add("lastname",StringType)) 
  .add("id",StringType)
  .add("gender",StringType)
  .add("salary",DoubleType)
  .add("birthdayStr",StringType)
  .add("language",ArrayType(StringType))
             )

val df = spark.createDataFrame(spark.sparkContext.parallelize(data),schema)
df.withColumn("birthday", to_date(col("birthdayStr"))).write.format("parquet").mode("overwrite").save("/tmp/testparquet")
val df2 = spark.read.parquet("/tmp/testparquet")
df2.createOrReplaceTempView("df2")
df2.printSchema

val querytext="""SELECT gender from df2 except all (select gender from df2 where salary <> 10)"""
sql(querytext).explain
```

The Spark Driver log messages:
```
  !Exec <GenerateExec> cannot run on GPU because not all expressions can be replaced
    !NOT_FOUND <ReplicateRows> replicaterows(sum#99L, gender#76) cannot run on GPU because no GPU enabled version of expression class org.apache.spark.sql.catalyst.expressions.ReplicateRows could be found
```

revans2:
The APIS already exist in CUDF `TableView.repeat`. The hardest part will be memory management and combining it with the existing code that is rather explode and pos_explode specific.

PR #4388, opened 2021-12-20
Add support for ReplicateRows

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/4104 .  ReplicateRows is used as optimization for exceptAll and intersectAll. The final plan includes HashAggregateExec followed by replicateRows. 
Current limitation of ReplicateRows would be that we could support only those datatypes which are supported by HashAggregateExec. Added integration tests for both exceptAll and intersectAll


nartal1:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3104,6 +3104,22 @@ object GpuOverrides extends Logging {
         override val supportOuter: Boolean = true
         override def convertToGpu(): GpuExpression = GpuPosExplode(childExprs.head.convertToGpu())
       }),
+    expr[ReplicateRows](
+      "Given an input row replicates the row N times",
+      ExprChecks.projectOnly(
+        // The plan is optimized to run HashAggregate on the rows to be replicated. So this
```
revans2, line 3110: Please file a follow on issue for this. I don't understand why decimal128 is a problem for generate. Even if it is being fed into an aggregation we still support aggregating on decimal128 values. It is just a bit slower.  If these are all going to be used as keys to the hash aggregate, then it is a bit different, but that still needs a follow on issue to make it work.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -164,6 +180,63 @@ trait GpuGenerator extends GpuUnevaluable {
   }
 }
 
+case class GpuReplicateRows(children: Seq[Expression]) extends GpuGenerator {
```
revans2, line 183: You are getting a compile error in some cases.

```
Error: :29:25.509Z] [ERROR] [Error] /home/jenkins/agent/workspace/jenkins-rapids_premerge-github-3584/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala:183: class GpuReplicateRows needs to be abstract, since method withNewChildrenInternal in class TreeNode of type (newChildren: IndexedSeq[org.apache.spark.sql.catalyst.expressions.Expression])org.apache.spark.sql.catalyst.expressions.Expression is not defined
```

I think because of the shim layer you need to have this 

```
extends GpuUnevaluableUnaryExpression with GpuGenerator {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
nartal1: Sorry, decimal128 isn't the problem for generate here but all the `project` columns become keys for HashAggregateExec. So we end up running HashAggregate on CPU and then ReplicateRows on GPU in decimal128 case. 
I will update ReplicateRows to support  decimal 128(keeping the comment that grouping by 128-decimal is not supported yet).  
Could you please let me know if the follow on issue should track supporting grouping by 128-decimal in general OR it should be for ReplicateRows ?

I ran a simple query and the explain shows that HashAggregate is on CPU followed by ReplicateRows on GPU:

Dataframe:
+---+------------------------------+
| id|               amount           |
+---+------------------------------+
|100|100.3300000000000...|
|200|200.6600000000000...|
|300|300.9900000000000...|
+---+------------------------------+

Result of explain:
`testDf.select($"id", $"amount").exceptAll(testDf.select($"id", $"amount").filter("id > 100")).explain`

```
== Physical Plan ==
GpuColumnarToRow false
+- GpuProject [id#100, amount#101]
   +- GpuGenerate gpureplicaterows(sum#118L, id#100, amount#101), [id#100, amount#101], false, [id#100, amount#101]
      +- GpuCoalesceBatches targetsize(2147483647)
         +- GpuFilter (gpuisnotnull(sum#118L) AND (sum#118L > 0)), true
               +- *(2) HashAggregate(keys=[id#100, amount#101], functions=[sum(vcol#115L)])
                  +- Exchange hashpartitioning(id#100, amount#101, 200), ENSURE_REQUIREMENTS, [id=#511]
                     +- *(1) HashAggregate(keys=[id#100, amount#101], functions=[partial_sum(vcol#115L)])
                        +- GpuColumnarToRow false
                           +- GpuUnion
                              :- GpuRowToColumnar targetsize(2147483647)
                              :  +- LocalTableScan [vcol#115L, id#100, amount#101]
                                 +- LocalTableScan [vcol#116L, id#111, amount#112]

```

nartal1:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -164,6 +180,63 @@ trait GpuGenerator extends GpuUnevaluable {
...
+
+  override def inputSplitIndices(inputBatch: ColumnarBatch,
+      generatorOffset: Int,
+      outer: Boolean,
+      targetSizeBytes: Long): Array[Int] = {
+    val vectors = GpuColumnVector.extractBases(inputBatch)
+    val inputRows = inputBatch.numRows()
+    if (inputRows == 0) return Array()
+
+    // Calculate the number of rows that needs to be replicated. Here we find the mean of the
```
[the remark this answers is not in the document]
nartal1: Filed follow on issue - https://github.com/NVIDIA/spark-rapids/issues/4405

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -164,6 +180,63 @@ trait GpuGenerator extends GpuUnevaluable {
...
+      case (e, index) => StructField(s"col$index", e.dataType)
+    })
+
+  override def generate(inputBatch: ColumnarBatch,
+      generatorOffset: Int,
+      outer: Boolean): ColumnarBatch = {
+
+    val schema = GpuColumnVector.extractTypes(inputBatch)
+    val vectors = GpuColumnVector.extractBases(inputBatch)
+    val replicateVector = vectors(generatorOffset)
```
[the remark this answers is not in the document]
nartal1: Thanks! You are right. Changed it to get only the required ColumnVector.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -164,6 +180,63 @@ trait GpuGenerator extends GpuUnevaluable {
...
+
+    def replicateRows(inputTable: Table, genOffset: ColumnVector): Table = {
```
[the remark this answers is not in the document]
nartal1: done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -164,6 +180,63 @@ trait GpuGenerator extends GpuUnevaluable {
...
+      outer: Boolean,
+      targetSizeBytes: Long): Array[Int] = {
+    val vectors = GpuColumnVector.extractBases(inputBatch)
+    val inputRows = inputBatch.numRows()
+    if (inputRows == 0) return Array()
+
+    // Calculate the number of rows that needs to be replicated. Here we find the mean of the
+    // generator column. Multiplying the mean with size of projected columns would give us the
+    // approximate memory required.
+    val meanOutputRows = math.ceil(vectors(generatorOffset).mean().getDouble)
```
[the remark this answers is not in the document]
nartal1: generatorOffset is the index of ColumnVector which has `sum` (for each row) that needs to be projected. 
That column is of type long. So calculating mean() over a long type should be okay, right? I don't see how we there could be a overflow. 
Please let me know if your concern was something else.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -164,6 +180,63 @@ trait GpuGenerator extends GpuUnevaluable {
...
+    }
+  }
+
+  override def inputSplitIndices(inputBatch: ColumnarBatch,
+      generatorOffset: Int,
+      outer: Boolean,
+      targetSizeBytes: Long): Array[Int] = {
+    val vectors = GpuColumnVector.extractBases(inputBatch)
+    val inputRows = inputBatch.numRows()
+    if (inputRows == 0) return Array()
```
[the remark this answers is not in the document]
nartal1: I couldn't get it to 0 inputRows.  It was mostly copied from Explode code which does similar comparison.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
nartal1: fixed this compile error.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuGenerateExec.scala
```diff
@@ -164,6 +180,61 @@ trait GpuGenerator extends GpuUnevaluable {
...
+    StructType(children.tail.zipWithIndex.map {
+      case (e, index) => StructField(s"col$index", e.dataType)
+    })
+
+  override def generate(inputBatch: ColumnarBatch,
+      generatorOffset: Int,
+      outer: Boolean): ColumnarBatch = {
+
+    val schema = GpuColumnVector.extractTypes(inputBatch)
+    val replicateVector = GpuColumnVector.extractBases(inputBatch)(generatorOffset)
```
revans2, line 195: nit: I was thinking more of something like

```
val replicateVector = inputBatch.column(generatorOffset).asInstanceOf[GpuColumnVector]
```

Or if we are converting `inputBatch` to a table anyways...

```
    withResource(GpuColumnVector.from(inputBatch)) { table =>
      val replicateVector = table.getColumn(generatorOffset)
      withResource(table.repeat(replicateVector)) { replicatedTable =>
        GpuColumnVector.from(replicatedTable, schema)
      }
    }
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
revans2: I still would like a follow on issue for decimal 128 so we are sure we don't miss it.

revans2 (approved):
Just a very minor nit, and a follow on issue that I want to be sure we don't forget to do.

nartal1:
@revans2 Filed follow on issue for decimal 128 https://github.com/NVIDIA/spark-rapids/issues/4410 . PTAL

nartal1:
build

revans2 (approved):

Issue #4386, opened 2021-12-20
[BUG] conditionals_test.py FAILED with side_effects_cast[Integer/Long] on Databricks 9.1 Runtime


NvTimLiu:
**Describe the bug**
[BUG] conditionals_test.py FAILED with side_effects_cast[Integer/Long], details as below `rapids_databricks_nightly-dev-github/255`, 

```
=================================== FAILURES ===================================
_______________ test_conditional_with_side_effects_cast[Integer] _______________
[gw0] linux -- Python 3.8.12 /databricks/conda/envs/cudf-udf/bin/python

data_gen = Integer

    @pytest.mark.parametrize('data_gen', int_n_long_gens, ids=idfn)
    def test_conditional_with_side_effects_cast(data_gen):
        gen = mk_str_gen('[0-9]{1,20}')
>       assert_gpu_and_cpu_are_equal_collect(
                lambda spark : two_col_df(spark, data_gen, gen).selectExpr(
                    'IF(a RLIKE "^[0-9]{1,5}$", CAST(a AS INT), 0)'),
                conf = {'spark.sql.ansi.enabled':True,
                        'spark.rapids.sql.expression.RLike': True})

../../src/main/python/conditionals_test.py:209: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../src/main/python/asserts.py:505: in assert_gpu_and_cpu_are_equal_collect
    _assert_gpu_and_cpu_are_equal(func, 'COLLECT', conf=conf, is_cpu_first=is_cpu_first)
../../src/main/python/asserts.py:424: in _assert_gpu_and_cpu_are_equal
    run_on_cpu()
../../src/main/python/asserts.py:410: in run_on_cpu
    from_cpu = with_cpu_session(bring_back, conf=conf)
../../src/main/python/spark_session.py:86: in with_cpu_session
    return with_spark_session(func, conf=copy)
../../src/main/python/spark_session.py:70: in with_spark_session
    ret = func(_spark)
../../src/main/python/asserts.py:198: in <lambda>
    bring_back = lambda spark: limit_func(spark).collect()
../../src/main/python/conditionals_test.py:210: in <lambda>
    lambda spark : two_col_df(spark, data_gen, gen).selectExpr(
/databricks/spark/python/pyspark/sql/dataframe.py:1709: in selectExpr
[...]

PR #4394, opened 2021-12-20
Use correct column name in GpuIf test [databricks]

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/4386

The first commit updates the test `test_conditional_with_side_effects_cast` to fix an error where the test was referencing the wrong column. This worked fine in Apache Spark presumably due to implicit casts being added. The test failed in Databricks.

The second commit simplifies the side-effect tests to use `unary_op_df` instead of `two_col_df` since these tests only reference a single column.


andygrove:
build

revans2 (approved):
oops thanks for catching this.

Issue #4395, opened 2021-12-20
[FEA] acquire the semaphore after concatToHost in GpuShuffleCoalesceIterator


abellina:
There is an opportunity in `GpuShuffleCoalesceIterator` for optimization where we are holding onto the semaphore before we concat on the host.

The cuDF call we are currently calling is: `JCudfSerialization.concatToContiguousTable`, but it's trivial to split it up since what it calls inside of cuDF is public already. The proposal is to:

1. Without holding the semaphore, call: `JCudfSerialization.concatToHostBuffer` to get a `HostConcatResult`.
2. Acquire the semaphore
3. Get the contiguous table on the GPU by calling: `HostConcatResult.toContiguousTable`.

Results with this change are promising, saving for all queries about 1 minute of runtime when adding everything up. Most queries are above the 1x line. The queries at or below 0.9 were: q52, q46, q68, q45, and q42. When executed multiple times they all went above 1x (these are single-digit second queries).

abellina:
An even better change may be for us to delay putting these batches on the GPU until we really need them. This would be specific for joins, where the build side could rest on the GPU for a while as we wait for the stream side. In these cases what you almost want is to `concatToHostBuffer`, but keep that `HostConcatResult` around and not need to acquire the GPU at all until the stream side is also ready to go.

jlowe:
> In these cases what you almost want is to concatToHostBuffer, but keep that HostConcatResult around and not need to acquire the GPU at all until the stream side is also ready to go.

This is a join-specific optimization, and I don't think we should get too far ahead of ourselves there.  The optimization described above should apply to all cases where we are doing host-side concat, and thus I think is a good optimization as-is.  We can always add a more sophisticated optimization for the shuffle-into-a-join case, but let's get this one done first.

PR #4396, opened 2021-12-20
GpuShuffleCoalesceIterator acquire semaphore after host concat

abellina:
Closes https://github.com/NVIDIA/spark-rapids/issues/4395

This is a small optimization that was spotted while looking into NDS Q64 traces. With the change, Q64 can save up to 3 seconds (though it changes quite a bit from run to run). When executing this over all of NDS, it saved ~1 minute for the whole run chipping away at times a few hundred ms, up to 5 seconds for q94.

I saw some queries being slower, with the worst case being q42 (which for 1 sample out of 10 was 2x slower). I have not been able to reproduce this case, with all subsequent runs at 1x or above. This was a 3.8 second in the last weekly run, with the patch it's hovering between 3.6 and 4.5.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffleCoalesceExec.scala
```diff
@@ -151,9 +151,6 @@ class GpuShuffleCoalesceIterator(
   }
 
   private def concatenateBatch(): ColumnarBatch = {
-    // about to start using the GPU in this task
-    GpuSemaphore.acquireIfNecessary(TaskContext.get(), semWaitTime)
```
jlowe, line 155: The new code no longer acquires the semaphore when there's no columns as it did before.  Is that a change we really want to make here?  I can think of cases where we could end up exploding the row-count-only batch into GPU data without holding the semaphore.

abellina: Oh right that's a great catch. We should add an unconditional acquire for the empty column case.

abellina: Handled here: https://github.com/NVIDIA/spark-rapids/pull/4396/commits/bc2a56cc55789ddeb9c8b516d3506f0057b3a9ad. I ran a few pointed queries that saw benefits with this new commit (like q94), I don't see it negate the benefit of the original patch, so I think it's still a go.

jlowe (approved):

jlowe:
build

PR #4397, opened 2021-12-20
Ignore cufile.log

andygrove:

Ignore `cufile.log`.


revans2 (approved):
okay, but we should really also not be producing it.

jlowe:
> okay, but we should really also not be producing it.

Filed #4404 to track this.

andygrove:
build

PR #4408, opened 2021-12-21
Use the ORC version that corresponds to the Spark version [databricks]

### LINKED ISSUE #4031: [BUG] Spark 3.3.0 test failure: NoSuchMethodError org.apache.orc.TypeDescription.getAttributeValue
the nightly build fails on the Spark 3.3.0 shim layer tests:

09:03:31  OrcScanSuite:
09:03:32  *** RUN ABORTED ***
09:03:32    java.lang.NoSuchMethodError: org.apache.orc.TypeDescription.getAttributeValue(Ljava/lang/String;)Ljava/lang/String;
09:03:32    at org.apache.spark.sql.execution.datasources.orc.OrcUtils$.toCatalystType$1(OrcUtils.scala:103)
09:03:32    at org.apache.spark.sql.execution.datasources.orc.OrcUtils$.$anonfun$toCatalystSchema$1(OrcUtils.scala:118)
09:03:32    at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
09:03:32    at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
09:03:32    at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
09:03:32    at org.apache.spark.sql.execution.datasources.orc.OrcUtils$.toStructType$1(OrcUtils.scala:116)
09:03:32    at org.apache.spark.sql.execution.datasources.orc.OrcUtils$.toCatalystSchema(OrcUtils.scala:138)
09:03:32    at org.apache.spark.sql.execution.datasources.orc.OrcUtils$$anonfun$readSchema$5.applyOrElse(OrcUtils.scala:148)
09:03:32    at org.apache.spark.sql.execution.datasources.orc.OrcUtils$$anonfun$readSchema$5.applyOrElse(OrcUtils.scala:145)
09:03:32    at scala.collection.TraversableOnce.collectFirst(TraversableOnce.scala:148)
09:03:32    ...

res-life:
[BUG] Spark 3.3.0 test failure: NoSuchMethodError org.apache.orc.TypeDescription.getAttributeValue #4031

**Root cause**
Rapids plugin uses a constant ORC version 1.5.10 and Spark 3.3.0 begins to use ORC 1.7.x which is not compatible with 1.5.10. 
For the unit test cases, it uses ORC 1.5.10, because of directly specified the dependency which overides the ORC 1.7.x dependency, Spark invoking the "getAttributeValue" which resides in ORC 1.7.x will fail.
But for the integration test cases, because of it uses the jars resides in the $SPARK_HOME/jars which includes ORC 1.7.x, it will not fail.
There are different behaviors between integration tests and unit tests.

**Solution**
1. Use the ORC version that corresponds to the Spark version and stop shading ORC. 
    This will make the behaviors of IT and UT consistent.
    Will make the aggregator jar small because of stopping shading ORC.
    Will upgrade ORC accordingly with Spark. 
2. Move the failed test case into IT and keep shading ORC 1.5.10
    This will make us confusing why we can't put this case into UT.

Here we use option 1.

Another change, added a common module:
In this PR it contains a util class ThreadFactoryBuilder that is used to replace the Guava dependency. The guava dependency is removed because of it's a messy jar in practice.
The ThreadFactoryBuilder is used in both sql-plugin and tools modules, so put it into the new common module.

This fixes #4031


res-life:
Use the ORC version that corresponds to the Spark version.
Make the test module depend on the aggregator jar instead of the dist jar.
Temp comment a test case in OrcScanSuite, will check it later.
Currently keep shading the guava, orc, etc, will update the shade scope in the issue #1398 .
Will add some comments later.

res-life:
build

res-life:
build

res-life:
build

sql-plugin/pom.xml
```diff
@@ -49,45 +49,457 @@
...
+                <dependency>
+                    <groupId>org.apache.orc</groupId>
+                    <artifactId>orc-mapreduce</artifactId>
+                    <version>1.5.10</version>
+                </dependency>
+                <dependency>
+                    <groupId>org.apache.hive</groupId>
+                    <artifactId>hive-storage-api</artifactId>
+                    <version>2.7.1</version>
+                </dependency>
```
jlowe, line 167: If the intent is to use the same version of ORC and Hive that Spark is using, why are we pulling in these dependencies explicitly rather than getting them transitively from the Spark artifacts?  The risk of putting them explicitly here is that they change versions for Spark snapshot artifacts and we fail to notice.

jlowe (commented):
I'm not clear on the end goal of this PR.  It looks like it is going to compile against different versions of ORC yet still pull ORC into the dist jar.  How is that going to work in practice -- won't the different ORC versions from different aggregator jars conflict when we try to pull it all together into the dist jar?

jlowe:
Additionally, how are the concerns about varying ORC classifiers pulled in by different Spark builds, as detailed in https://github.com/NVIDIA/spark-rapids/issues/4031#issuecomment-991157488, being addressed?

res-life:
>I'm not clear on the end goal of this PR. It looks like it is going to compile against different versions of ORC yet still pull ORC into the dist jar. How is that going to work in practice -- won't the different ORC versions from different aggregator jars conflict when we try to pull it all together into the dist jar?

"binary-dedupe.sh" is used to compare the class binary, so there is no conflict for multiple ORC versions. Of cause, there will be more class files. 


>If the intent is to use the same version of ORC and Hive that Spark is using, why are we pulling in these dependencies explicitly rather than getting them transitively from the Spark artifacts? The risk of putting them explicitly here is that they change versions for Spark snapshot artifacts and we fail to notice.

Sometimes the transitive jar is not the same as the Spark runtime, so explicitly specify. 
You are right, for the Spark snapshot, it's better to use the transitive jar.

>Additionally, how are the concerns about varying ORC classifiers pulled in by different Spark builds, as detailed in #4031 (comment), being addressed?

To address this concern,  I have no good idea, it's better to use the current strategy of shading the ORC with the hive code.
The PR of #1398 will create a new module to collect the necessary ORC and hive classes to minimum the shading scope.
Seems the rapids plugin only used the orc-core, hive-storage-api and protobuf-java, I'll shade the ORC input/output class.

shims/pom.xml
```diff
@@ -85,6 +85,10 @@
             <id>release321</id>
             <modules><module>spark321</module></modules>
         </profile>
+        <profile>
+            <id>release301db</id>
+            <modules><module>spark301db</module></modules>
+        </profile>
```
jlowe, line 91: Why is this change here, and if it is really needed, why only 301db and not the other Databricks profiles?

shims/spark311cdh/src/main/scala/com/nvidia/spark/rapids/shims/spark311cdh/Spark311CDHShims.scala
```diff
@@ -41,4 +49,107 @@ class Spark311CDHShims extends Spark31XShims with Spark30Xuntil33XShims {
     new ParquetFilters(schema, pushDownDate, pushDownTimestamp, pushDownDecimal, pushDownStartWith,
       pushDownInFilterThreshold, caseSensitive)
   }
+
+  /**
+   * create Orc OutStream
+   * CDH version use the shaded Orc 1.5.10 code
+   * @return
+   */
+  override def newOrcOutStream(name: String, orcBufferSize: Int, codec: CompressionCodec,
```
jlowe, line 58: We have been moving away from using a runtime lookup of the shims (i.e.: methods in the shim interface as done here) to a compile-time lookup of the shims (i.e.: the shims v2 approach).  I'd prefer we use that here.  For example, we could create an OrcShim or OrcWrapper object class under com.nvidia.spark.rapids.shims.v2 that contains these ORC functions.  We would then create separate ones based on ORC version under the sql-plugin  directories and add those source paths to the appropriate profiles based on which ORC version the corresponding Spark version uses.  The common code would then simply use the OrcShim or OrcWrapper object directly rather than using the shim loader at runtime to lookup the shims.

sql-plugin/pom.xml
jlowe: Still waiting for an answer to this question.

sql-plugin/pom.xml
```diff
@@ -49,45 +49,457 @@
...
-            </exclusion>
-          </exclusions>
-        </dependency>
         <dependency>
           <groupId>com.google.protobuf</groupId>
           <artifactId>protobuf-java</artifactId>
         </dependency>
     </dependencies>
 
+    <dependencyManagement>
```
jlowe, line 58: dependencyManagement is useful for parent poms, but this isn't a parent pom.  It's also odd to have a dependencyManagement section with no versions listed, so I'm not sure what the point of this is.  All of the dependencies used in the normal dependency section below have versions as expected, so this seems unnecessary.

res-life:
build

sql-plugin/pom.xml
res-life: Dependencies without version are used by different profiles, specified the exclusions,  so the dependencies in profiles can omit the exclusions.

shims/pom.xml
res-life: Added for all the spark releases. These are only used by idea, idea can add or remove submodule dynamically when switching the profiles in Maven panel.

shims/spark311cdh/src/main/scala/com/nvidia/spark/rapids/shims/spark311cdh/Spark311CDHShims.scala
res-life: Yes, it is right, updated.

sql-plugin/pom.xml
res-life: The scope of the transitive dependencies of Spark are provided, the maven shade plugin only shade compile jars, so should  explicitly specify orc jar as compile scope, and then should add the version.
And sometimes the transitive jars are not the same as the Spark runtime(spark-home/jars).

Your are right, for the Spark snapshot, the transitive jar will change without notice. Seems no good idea but mannully update the version.

res-life:
build

sql-plugin/pom.xml
res-life: For the Spark snapshot, seems it's better to add a daily crontab job to check the version.

jlowe: An alternative would be to have the build run a script to compute the version by examining the Spark jar dependencies to find the ORC version.  It could then fail the build (or emit a loud warning) if it finds the computed ORC version doesn't match the manually specified version.  Even better would be to use the computed ORC version directly as the plugin's ORC dependency version if that's possible.

sql-plugin/src/main/320+/scala/com/nvidia/spark/rapids/shims/v2/OrcShims.scala
```diff
@@ -0,0 +1,102 @@
...
+    val backupSchema = readerOpts.getSchema
+    readerOpts.schema(updatedReadSchema)
+    withResource(new RecordReaderImplSub(orcReader.asInstanceOf[ReaderImpl], readerOpts)) {
+      recordReader =>
+        stripes.foreach { stripe =>
+          val stripeFooter = dataReader.readStripeFooter(stripe)
+
+          recordReader.readRowIndex(stripe.getStripeId.intValue(), fileIncluded, sargColumns)
+          // use returnNone = false underline
+          val rowGroups = recordReader.pickRowGroups()
```
jlowe, line 89: `SargApplier` still appears to have the same `pickRowGroups` method in the newer ORC versions, and the old ORC versions have a `RecordReaderImpl.pickRowGroups`, so wondering why this is done differently in Spark 3.2.0+.  Should we be using the record reader approach in previous Spark versions as well, or should Spark 3.2.0+ simply keep using `SargApplier`?  I'm not seeing why there should be a different approach between the two versions.

Note that the predicate pushdown filter is encoded in `sargApp` and otherwise not passed to this method, so curious how predicate pushdown is working with the `RecordReaderImplSub` approach.

sql-plugin/src/main/320+/scala/com/nvidia/spark/rapids/shims/v2/OrcShims.scala
```diff
@@ -0,0 +1,102 @@
...
+    // backup the schema
+    val backupSchema = readerOpts.getSchema
+    readerOpts.schema(updatedReadSchema)
+    withResource(new RecordReaderImplSub(orcReader.asInstanceOf[ReaderImpl], readerOpts)) {
+      recordReader =>
+        stripes.foreach { stripe =>
+          val stripeFooter = dataReader.readStripeFooter(stripe)
+
+          recordReader.readRowIndex(stripe.getStripeId.intValue(), fileIncluded, sargColumns)
+          // use returnNone = false underline
```
jlowe, line 88: I don't understand what this comment means.  Is there some missing context?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/VersionUtils.scala
```diff
@@ -24,6 +24,8 @@ object VersionUtils {
 
   lazy val isSpark320OrLater: Boolean = cmpSparkVersion(3, 2, 0) >= 0
 
+  lazy val isSpark311Before: Boolean = !isSpark311OrLater
```
jlowe, line 27: This should be named `isPriorToSpark311` or `isBeforeSpark311` if we keep it, otherwise one could misread that as "is Spark 3.1.1 before the current version."  However I would argue it's not necessary.  IMHO `!isSpark311OrLater` is pretty clear at the call site.  Otherwise one would argue we need equivalent "before" vals for all the other Spark versions.

tests/pom.xml
```diff
@@ -54,12 +54,6 @@
             <classifier>${cuda.version}</classifier>
             <scope>provided</scope>
         </dependency>
-        <dependency>
```
jlowe, line 57: 2022 copyrights, and this seems like an unrelated change?

tests/src/test/scala/com/nvidia/spark/rapids/OrcScanSuite.scala
```diff
@@ -122,7 +125,7 @@ class OrcScanSuite extends SparkQueryCompareTestSuite {
       // should not have data, if orc.proleptic.gregorian is specified
       val ret = spark.sql("select c1 from df1 where c1 >= to_date('1582-10-03', 'yyyy-MM-dd') ")
           .collect()
-      assert(ret.isEmpty)
+       assert(ret.isEmpty)
```
jlowe, line 128:
```suggestion
      assert(ret.isEmpty)
```

tests/src/test/scala/com/nvidia/spark/rapids/SparkQueryCompareTestSuite.scala
```diff
@@ -1835,6 +1835,9 @@ trait SparkQueryCompareTestSuite extends FunSuite with Arm {
   def assumeSpark320orLater: Assertion =
     assume(VersionUtils.isSpark320OrLater, "Spark version not 3.2.0+")
 
+  def assumeSpark311Before: Assertion =
```
jlowe, line 1838: Name should be `assumePriorToSpark311` to match precedent set by `assumePriorToSpark320`.

sameerz:
Can we retarget this to 22.04?  Reason being we ought to wait for the fix for https://github.com/rapidsai/cudf/issues/9964 to go in before we merge this.

sql-plugin/src/main/320+/scala/com/nvidia/spark/rapids/shims/v2/OrcShims.scala
res-life: DataReader.readRowIndex was removed after ORC 1.6.11, so used a different approach.
Actually buildOrcReaderOpts in GpuOrcScanBase created the SearchArgument and then RecordReaderImplSub generated a sargApp.
Now aligned, replaced  DataReader.readRowIndex with StripePlanner.readRowIndex

tests/pom.xml
res-life: Commit time is before 2021-01-01, so did not add 2022 copyright.

res-life:
Updated, but still a draft, need to find a way to compute the version by examining the Spark jar dependencies.
Sent a mail to discuss If shade is needed.

res-life:
build

tgravescs:
@res-life  Please put the issue description for the PR in the first post.

res-life:
build

res-life:
Changed the solution of aligning Orc versions PR after investigated the history of shading ORC.
Now do not shade ORC again, updated for all the Spark versions: 3.0.1 to 3.3.0, 301db, 312db and 311cdh.
After this change, the aggregator jar reduced size from 13M to 6.8M.
And do not need to handle the refine Orc shade issue.

res-life:
build

res-life:
After 11/24/2019 and from ORC-1.5.7 Spark no longer use "nohive" classifier ORC jar again for Hive 2.0+.
If our product aims Spark v3.0.0+ and Hive 2.0+ without Hive 1.x, then it's safe to remove the shade configuration of ORC.
So, let's trying to remove it and do some regression tests before merging the code.

Details:
cd spark-source-path
git checkout v3.0.0
git log -p ./dev/deps | grep -20 orc-core-1.5

res-life:
ORC 1.6.11+ failed to prune when reading ORC file in Proleptic calendar which was written in Hybrid calendar.
The ORC file in the issue was created by Spark CPU version and ORC 1.5.10 behavior is correct.
Filed a ORC bug: https://issues.apache.org/jira/browse/ORC-1083

res-life:
build

pom.xml
```diff
@@ -899,88 +900,6 @@
...
-            <dependency>
-              <groupId>com.google.protobuf</groupId>
-              <artifactId>protobuf-java</artifactId>
-              <version>${protobuf.java.version}</version>
-            </dependency>
-            <dependency>
-              <groupId>com.google.flatbuffers</groupId>
-              <artifactId>flatbuffers-java</artifactId>
-              <version>${flatbuffers.java.version}</version>
-            </dependency>
```
jlowe, line 983: Was this deleted accidentally?  The plugin uses flatbuffers directly, and I don't think this is related to ORC.

sql-plugin/pom.xml
```diff
@@ -49,41 +44,13 @@
             <artifactId>scalatest_${scala.binary.version}</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <!--Note Spark 3.0.1~3.2.2 use 1.9; 3.3.0 use 1.12; but we use 1.11-->
+        <!--TODO investigate: flat buffers is a little jar, just use version 1.9 is sufficient? better not shade it-->
```
jlowe, line 49: IIRC we needed version 11 for an API we were using.  As you said, it's small and I don't see any reason to use the Spark version here.  The reason we're using flatbuffers has nothing to do with what version Spark is using.  We're not passing flatbuffers to/from Spark, this is for communication in our UCX-based shuffle and therefore it seems appropriate to use a fixed version so we don't have to worry about flatbuffer version-specific bugs across Spark versions.

sql-plugin/pom.xml
```diff
@@ -49,41 +44,13 @@
...
-              <artifactId>slf4j-api</artifactId>
-            </exclusion>
-          </exclusions>
-        </dependency>
-        <dependency>
-          <groupId>com.google.protobuf</groupId>
-          <artifactId>protobuf-java</artifactId>
+            <groupId>com.google.flatbuffers</groupId>
+            <artifactId>flatbuffers-java</artifactId>
+            <version>${flatbuffers.java.version}</version>
```
jlowe, line 53: This should use the dependency management from the parent pom (which should not have been deleted).

shuffle-plugin/pom.xml
```diff
@@ -124,6 +124,12 @@
                     <version>${spark.version}</version>
                     <scope>provided</scope>
                 </dependency>
+	        <dependency>
+                    <groupId>com.google.guava</groupId>
```
jlowe, line 128: Why are we adding a guava dependency only in the databricks profile?  Seems like we're missing a general guava dependency instead.  Applies to the sql-plugin guava dependency as well.  It looks like we're only using guava for `ThreadFactoryBuilder` which we may want to reconsider, as guava is a notoriously messy dependency in practice.

res-life:
build

res-life:
Added flat buffers dependency into dependency management.
Removed Guava dependency across the whole project.
Rebased for branch 22.04.

jlowe:
Since this PR stops shading ORC, it should also update [NOTICE-binary](https://github.com/NVIDIA/spark-rapids/blob/branch-22.02/NOTICE-binary) to remove the sections that we're no longer shading.

res-life:
build

res-life:
build

res-life:
build

res-life:
build

common/src/test/scala/com/nvidia/spark/rapids/ThreadFactoryBuilderTest.scala
```diff
@@ -0,0 +1,59 @@
...
+package com.nvidia.spark.rapids
+
+import java.util.concurrent.Executors
+
+import org.scalatest.FunSuite
+
+class ThreadFactoryBuilderTest extends FunSuite {
+
+  test("test thread factory builder") {
+    var pool = Executors.newFixedThreadPool(2,
```
jlowe, line 26: The test should ensure `pool` is shutdown.

common/src/test/scala/com/nvidia/spark/rapids/ThreadFactoryBuilderTest.scala
```diff
@@ -0,0 +1,59 @@
...
+      new ThreadFactoryBuilder().setNameFormat("Thread %s").setDaemon(true).build())
+
+    pool.submit(new Runnable {
+      override def run(): Unit = {
+        assert(Thread.currentThread().isDaemon)
+        assert(Thread.currentThread().getName == "Thread 0")
+      }
+    })
```
jlowe, line 34: I don't think this does what we want.  The main thread, i.e: the one responsible for the test, will submit a bunch of tasks to executors and does _not_ wait for them to complete nor evaluate any `Future` result from their run.  In other words, the main thread could race through and "succeed" before the first submitted task is even run.  In addition, the tasks will throw exceptions if they assert, but these will not be seen by the main test thread.  Instead the executor service will catch these exceptions and hold onto them, waiting for some task to examine the `Future` that was returned by the `submit` so it can propagate the exception that was caught when someone tries to get the result from the `submit`.  However since that is thrown away, so its the exception -- it will be silently eaten.

For this test to be valid, the main thread minimally needs to wait for the submitted tasks by examining the `Future` results so any exceptions thrown by the worker threads will be propagated to the main thread and thus fail the test.

jlowe (commented):
The shuffle plugin pom should depend on the common module since the code directly accesses classes in that module.

common/src/test/scala/com/nvidia/spark/rapids/ThreadFactoryBuilderTest.scala
```diff
@@ -26,34 +26,43 @@ class ThreadFactoryBuilderTest extends FunSuite {
...
+    ret = pool.submit(new Callable[String] {
+      override def call(): String = {
         assert(Thread.currentThread().isDaemon)
-        assert(Thread.currentThread().getName == "Thread 1")
+        assert(Thread.currentThread().getName() == "Thread 1")
+        ""
       }
     })
+    ret.get()
+    pool.shutdown()
```
jlowe, line 46: `pool` won't be shutdown if there's an exception.  This needs to be in a `finally` block.  Same applies to the shutdown call below.

jlowe (commented):
I think this is pretty close, just a small nit in the test.

@tgravescs it would be good to have you take another look.

res-life:
build

tools/pom.xml
```diff
@@ -40,6 +40,11 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>com.nvidia</groupId>
+            <artifactId>rapids-4-spark-common_${scala.binary.version}</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
jlowe, line 47: Adding this dependency is causing the following warning during the build:
```
[WARNING] rapids-4-spark-common_2.12-22.02.0-SNAPSHOT.jar, rapids-4-spark-tools_2.12-22.02.0-SNAPSHOT.jar define 1 overlapping resource: 
[WARNING]   - META-INF/MANIFEST.MF
[WARNING] maven-shade-plugin has detected that some class files are
[WARNING] present in two or more JARs. When this happens, only one
[WARNING] single version of the class is copied to the uber jar.
[WARNING] Usually this is not harmful and you can skip these warnings,
[WARNING] otherwise try to manually exclude artifacts based on
[WARNING] mvn dependency:tree -Ddetail=true and the above output.
[WARNING] See http://maven.apache.org/plugins/maven-shade-plugin/
```
The shade directives will need to be refined.

res-life:
build

tools/pom.xml
res-life: Updated

common/src/test/scala/com/nvidia/spark/rapids/ThreadFactoryBuilderTest.scala
```diff
@@ -0,0 +1,72 @@
...
+package com.nvidia.spark.rapids
+
+import java.util.concurrent.{Callable, Executors, ExecutorService}
+
+import org.scalatest.FunSuite
+
+class ThreadFactoryBuilderTest extends FunSuite {
+
+  test("test thread factory builder") {
+    var pool1: ExecutorService = null
```
jlowe, line 26: Nit: There's no need for `var` or `null` here.  This should simply be:
```scala
val pool1 = Executors.newFixedThreadPool(...)
try {
  ...
} finally {
  pool1.shutdown()
}
```
Applies to the other instances of this.

common/src/test/scala/com/nvidia/spark/rapids/ThreadFactoryBuilderTest.scala
```diff
@@ -0,0 +1,72 @@
...
+    try {
+      pool1 = Executors.newFixedThreadPool(2,
+        new ThreadFactoryBuilder().setNameFormat("thread-pool1-1 %s").setDaemon(true).build())
```
jlowe, line 29: I found it odd that the code below looks exactly like this code.  We're testing setting `setDaemon` to the default value, which means if the code completely botched the handling of that setting and the method did nothing, this test would still pass.

jlowe (commented):
Just minor nits remain, so this looks good to me.  Would like to hear from @tgravescs before merging.

common/src/test/scala/com/nvidia/spark/rapids/ThreadFactoryBuilderTest.scala
res-life: The default value of setDaemon is false.
I've added the cases for both setDaemon(true) and setDaemon(false)

res-life:
build

common/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+<!-- Filter unused classes for shade purpose, generate a jar with shaded classifier -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.nvidia</groupId>
+        <artifactId>rapids-4-spark-parent</artifactId>
+        <version>22.02.0-SNAPSHOT</version>
```
jlowe, line 27:
```suggestion
        <version>22.04.0-SNAPSHOT</version>
```

common/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+    </parent>
+
+    <artifactId>rapids-4-spark-common_2.12</artifactId>
+    <name>RAPIDS Accelerator for Apache Spark Common</name>
+    <description>Utility code that is common across the RAPIDS Accelerator projects</description>
+    <version>22.02.0-SNAPSHOT</version>
```
jlowe, line 33:
```suggestion
    <version>22.04.0-SNAPSHOT</version>
```

res-life:
build

jlowe (dismissed):
Thanks for all the hard work, @res-life!  This looks good to me.  @tgravescs can you take a look?

jenkins/databricks/build.sh
```diff
@@ -87,10 +87,18 @@ then
...
     PARQUETCOMMONJAR=----workspace_${SPARK_MAJOR_VERSION_STRING}--maven-trees--hive-2.3__hadoop-2.7--org.apache.parquet--parquet-common--org.apache.parquet__parquet-common__1.10.1-databricks9.jar
     PARQUETCOLUMNJAR=----workspace_${SPARK_MAJOR_VERSION_STRING}--maven-trees--hive-2.3__hadoop-2.7--org.apache.parquet--parquet-column--org.apache.parquet__parquet-column__1.10.1-databricks9.jar
+    ORC_CORE_JAR=----workspace_${SPARK_MAJOR_VERSION_STRING}--maven-trees--hive-2.3__hadoop-2.7--org.apache.orc--orc-core--org.apache.orc__orc-core__1.5.12.jar
+    ORC_SHIM_JAR=----workspace_${SPARK_MAJOR_VERSION_STRING}--maven-trees--hive-2.3__hadoop-2.7--org.apache.orc--orc-shims--org.apache.orc__orc-shims__1.5.12.jar
+    ORC_MAPREDUCE_JAR=----workspace_${SPARK_MAJOR_VERSION_STRING}--maven-trees--hive-2.3__hadoop-2.7--org.apache.orc--orc-mapreduce--org.apache.orc__orc-mapreduce__1.5.12.jar
+    PROTOBUF_JAR=----workspace_${SPARK_MAJOR_VERSION_STRING}--maven-trees--hive-2.3__hadoop-2.7--com.google.protobuf--protobuf-java--com.google.protobuf__protobuf-java__2.6.1.jar
```
tgravescs, line 93: this jar looks the same the one in the else statement, if so move out

sql-plugin/src/main/301until320-all/scala/com/nvidia/spark/rapids/shims/v2/OrcShims301until320Base.scala
```diff
@@ -0,0 +1,88 @@
...
+import scala.collection.mutable.ArrayBuffer
+
+import com.nvidia.spark.rapids.OrcOutputStripe
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.hive.common.io.DiskRangeList
+import org.apache.orc.{CompressionCodec, CompressionKind, DataReader, OrcFile, OrcProto, PhysicalWriter, Reader, StripeInformation}
+import org.apache.orc.impl.{DataReaderProperties, OutStream, SchemaEvolution}
+import org.apache.orc.impl.RecordReaderImpl.SargApplier
+
+// [301, 320) ORC shims
```
tgravescs, line 27: nit - not sure if the ) is supposed to be until,  I think we can just remove this comment as the name should be pretty clear

tgravescs (commented):
overall looks good, a couple of nits. It would also be nice to update the description to talk about the new common module and why it was added.

res-life:
build

tgravescs (approved):

jlowe (approved):

Issue #4378, opened 2021-12-17
[BUG] udf_test udf_cudf_test failed require_minimum_pandas_version check in spark 320+


pxLi:
**Describe the bug**
seems pandas lib in spark 320+ is incompatible w/ the one installed as cudf's dep

```
[2021-12-17T11:14:48.244Z] ==================================== ERRORS ====================================
[2021-12-17T11:14:48.244Z] ______________ ERROR collecting src/main/python/udf_cudf_test.py _______________
[2021-12-17T11:14:48.244Z] ../../src/main/python/udf_cudf_test.py:21: in <module>
[2021-12-17T11:14:48.244Z]     require_minimum_pandas_version()
[2021-12-17T11:14:48.244Z] /home/jenkins/agent/workspace/jenkins-rapids_cudf_udf-dev-github-33-cuda11.2/jars/spark-3.2.0-bin-hadoop3.2/python/lib/pyspark.zip/pyspark/sql/pandas/utils.py:27: in require_minimum_pandas_version
[2021-12-17T11:14:48.244Z]     import pandas
[2021-12-17T11:14:48.244Z] ../../../spark-3.2.0-bin-hadoop3.2/python/pyspark/pandas/__init__.py:31: in <module>
[2021-12-17T11:14:48.244Z]     require_minimum_pandas_version()
[2021-12-17T11:14:48.244Z] /home/jenkins/agent/workspace/jenkins-rapids_cudf_udf-dev-github-33-cuda11.2/jars/spark-3.2.0-bin-hadoop3.2/python/lib/pyspark.zip/pyspark/sql/pandas/utils.py:35: in require_minimum_pandas_version
[2021-12-17T11:14:48.244Z]     if LooseVersion(pandas.__version__) < LooseVersion(minimum_pandas_version):
[2021-12-17T11:14:48.244Z] E   AttributeError: partially initialized module 'pandas' has no attribute '__version__' (most likely due to a circular import)
[2021-12-17T11:14:48.244Z] 
[2021-12-17T11:14:48.244Z] During handling of the above exception, another exception occurred:
[2021-12-17T11:14:48.244Z] ../../src/main/python/udf_cudf_test.py:24: in <module>
[2021-12-17T11:14:48.244Z]     raise AssertionError("incorrect pandas version during required testing " + str(e))
[...]

PR #4419, opened 2021-12-22
Import the right pandas from conda

NvTimLiu:
There is a pandas package in the spark v3.2.0 or later binary dir: python/pyspark/pandas,
    the udf-cudf test includes this dir via PYTHONPATH, which causes test failure as below:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/jars/spark-3.2.0-bin-hadoop3.2/python/pyspark/pandas/__init__.py", line 31, in <module>
    require_minimum_pandas_version()
  File "/jars/spark-3.2.0-bin-hadoop3.2/python/pyspark/sql/pandas/utils.py", line 35, in require_minimum_pandas_version
    if LooseVersion(pandas.__version__) < LooseVersion(minimum_pandas_version):
AttributeError: partially initialized module 'pandas' has no attribute '__version__' (most likely due to a circular import)
```

To fix, put conda package path ahead of the env 'PYTHONPATH', to import the right pandas
    from conda instead of spark3.2.0 or later binary path.

### DIFF
--- a/jenkins/spark-tests.sh
+++ b/jenkins/spark-tests.sh
@@ -111,9 +111,6 @@ export PYTHONPATH=$SPARK_HOME/python:$SPARK_HOME/python/pyspark/:$SPARK_HOME/pyt
 IS_SPARK_311_OR_LATER=0
 [[ "$(printf '%s\n' "3.1.1" "$SPARK_VER" | sort -V | head -n1)" = "3.1.1" ]] && IS_SPARK_311_OR_LATER=1
 
-IS_SPARK_320_OR_LATER=0
-[[ "$(printf '%s\n' "3.2.0" "$SPARK_VER" | sort -V | head -n1)" = "3.2.0" ]] && IS_SPARK_320_OR_LATER=1
-
 export SPARK_TASK_MAXFAILURES=1
 [[ "$IS_SPARK_311_OR_LATER" -eq "0" ]] && SPARK_TASK_MAXFAILURES=4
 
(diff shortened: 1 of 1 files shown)


NvTimLiu:
build

NvTimLiu:
To fix https://github.com/NVIDIA/spark-rapids/issues/4378

jenkins/spark-tests.sh
```diff
@@ -276,6 +279,14 @@ fi
 
 # cudf_udf_test
 if [[ "$TEST_MODE" == "ALL" || "$TEST_MODE" == "CUDF_UDF_ONLY" ]]; then
+  if [[ "$IS_SPARK_320_OR_LATER" -eq "1" ]]; then
```
jlowe, line 282: Does this change need to be specific to Spark 3.2?  If pulling in the conda version of pandas works across all Spark versions, I'd rather reduce the number of special setups we have to maintain.

GaryShen2008: I think it's because there's no pandas package in the python folder of the previous version of spark 3.2.
But yes, I think it should be no harm to do in all the spark.

NvTimLiu: There is another spark `pandas` folder in spark-3.2.0 or later versions, so we use conda `pandas` to overwrite spark `pandas`.

No harm to do in all the spark, updated

jlowe (approved):

jlowe:
build

GaryShen2008:
build

PR #4420, opened 2021-12-22
Fix up incorrect results of rounding past the max digits of data type [databricks]

### LINKED ISSUE #4273: [BUG] Rounding past the size that can be stored in a type produces incorrect results
There are some odd corner cases when rounding for integers and floating point values.  If you try to round past what the number can hold Spark will typically produce a 0 all of the time. For integers there is still some overflow issues possible, so it is not 100% that way.  For CUDF with integers we can get numbers that are way off and for Floating point values we can get back a NaN.  This is a real corner case though because we only support a constant scale value so someone would have to get things rather messed up.

We have fixed this for Decimal though.

sperlingxx:

Fixes #4273

Current PR is to fix up integral values rounded by a scale exceeding/reaching the max digits of data type. Under this circumstance, cuDF may produce different results to Spark. The general strategy applied in this PR:
* For scales exceeding max digits, we can simply return zero values.
* For scales equaling to max digits, we need to perform round. Fortunately, round up will NOT occur on the max digits of numeric types except LongType. Therefore, we only need to handle round down for most of types, through returning zero values. For LongType rounded by scale -19, we handle it in a specialized way.


sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -453,13 +453,160 @@ abstract class GpuRoundBase(child: Expression, scale: Expression) extends GpuBin
...
+    // type. Under this circumstance, cuDF may produce different results to Spark.
+    //
+    // In this method, we handle round overflow, aligning the inconsistent results to Spark.
+    //
+    // For scales exceeding max digits, we can simply return zero values.
+    //
+    // For scales equaling to max digits, we need to perform round. Fortunately, round up
+    // will NOT occur on the max digits of numeric types except LongType. Therefore, we only
+    // need to handle round down for most of types, through returning zero values.
+    def fixUpOverflowInts(zeroFn: () => Scalar): ColumnVector = {
```
firestarman, line 469: I would prefer defining these functions out of this `doColumnar` to making them as the internal functions.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -453,13 +453,160 @@ abstract class GpuRoundBase(child: Expression, scale: Expression) extends GpuBin
...
+    //  the max value of Float64: 1.7976931348623157E308
+    //  The first digit is up to 1, which can't be rounded up.
+    //  the max value of Int64: 9223372036854775807
+    //  The first digit is up to 9, which can be rounded up.
+    //
+    // When rounding up 19-digits long values on the first digit, the result can be 1e19 or -1e19.
+    // Since LongType can not hold these two values, the 1e19 overflows as -8446744073709551616L,
+    // and the -1e19 overflows as 8446744073709551616L. The overflow happens in the same way for
+    // HALF_UP (round) and HALF_EVEN (bround).
+    def fixUpInt64OnBounds(zeroFn: () => Scalar): ColumnVector = {
```
firestarman, line 514: Here only handles the long type, so seems this `zeroFn` is not necessary.

It should always be a zero column of the long type, right ?

BTW, do we really need to take care of this round up/down of long ? Can cudf `round` handle it the same as what you are doing here ?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -453,13 +453,160 @@ abstract class GpuRoundBase(child: Expression, scale: Expression) extends GpuBin
...
+
+    // Fixes up float points rounded by a scale exceeding the max digits of data type. Under this
+    // circumstance, cuDF produces different results to Spark.
+    // Compared to integral values, fixing up round overflow of float points needs to take care
+    // of some special values: nan, inf, -inf.
+    def fpZeroReplacement(zeroFn: () => Scalar,
+        infFn: () => Scalar,
+        negInfFn: () => Scalar): ColumnVector = {
+      val scaleVal = scale.getValue.asInstanceOf[Int]
+      val maxDigits = if (dataType == FloatType) 39 else 309
```
firestarman, line 559: I prefer to have a similar API to `DecimalUtil.getPrecisionForIntegralType`, e.g. `DecimalUtil.getPrecisionForFloatType`

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -430,6 +430,31 @@ def test_decimal_round(data_gen):
...
+            'round(double_c, -308)', 'round(double_c, -309)',
+            'round(double_c, 309)', 'round(double_c, 309)',
+            'bround(byte_c, -2)', 'bround(byte_c, -3)',
+            'bround(short_c, -4)', 'bround(short_c, -5)',
+            'bround(int_c, -9)', 'bround(int_c, -10)',
+            'bround(long_c, -19)', 'bround(long_c, -20)',
+            'bround(float_c, -38)', 'bround(float_c, -39)',
+            'bround(float_c, 38)', 'bround(float_c, 39)',
+            'bround(double_c, -308)', 'bround(double_c, -309)',
+            'bround(double_c, 309)', 'bround(double_c, 309)'))
```
firestarman, line 456:
```suggestion
            'bround(double_c, 308)', 'bround(double_c, 309)'))
```

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -430,6 +430,31 @@ def test_decimal_round(data_gen):
...
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: gen_df(spark, gen).selectExpr(
+            'round(byte_c, -2)', 'round(byte_c, -3)',
+            'round(short_c, -4)', 'round(short_c, -5)',
+            'round(int_c, -9)', 'round(int_c, -10)',
+            'round(long_c, -19)', 'round(long_c, -20)',
+            'round(float_c, -38)', 'round(float_c, -39)',
+            'round(float_c, 38)', 'round(float_c, 39)',
+            'round(double_c, -308)', 'round(double_c, -309)',
+            'round(double_c, 309)', 'round(double_c, 309)',
```
firestarman, line 448:
```suggestion
            'round(double_c, 308)', 'round(double_c, 309)',
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -453,13 +453,160 @@ abstract class GpuRoundBase(child: Expression, scale: Expression) extends GpuBin
...
+        withResource(Seq(zeroFn(), infFn(), negInfFn())) { case Seq(zero, inf, negInf) =>
+          // builds joined predicate: !Null AND !Nan AND !Inf And !-Inf
+          val joinedPredicate = {
+            val conditions = Seq(() => lhsValue.isNotNan,
+                                 () => lhsValue.notEqualTo(inf),
+                                 () => lhsValue.notEqualTo(negInf))
+            conditions.foldLeft(lhsValue.isNotNull) { case (buffer, builder) =>
+              withResource(buffer) { _ =>
+                withResource(builder()) { predicate =>
+                  buffer.and(predicate)
```
revans2, line 572: Where are the intermediate values closed?

sperlingxx: If I am not wrong, the output of Nth foldLeft iteration will be closed in N+1th iteration as buffer.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
sperlingxx: Done.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
sperlingxx: You are right. I removed the argument `zeroFn`. 
When rounding by scales exceeding the max digits, the results of `cudf::round` are inconsistent with Spark. And in terms of cuDF, it is a undefined behavior. So, technically cuDF can return any value under this situation. That's why we need to override the round of extreme large scales.

sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
sperlingxx: Strictly speaking, the `39` and `309` are numeric bounds rather than precision of float types. These numeric bounds are rarely used in spark-rapids.  And they have little relation with `DecimalUtil`. So, I don't think we need to make it a common method.

revans2 (approved):

PR #4440, opened 2021-12-29
Enable GPU broadcast exchange reuse for DPP when AQE enabled [databricks]

### LINKED ISSUE #4439: [FEA] Enable GPU broadcast exchange reuse for DPP when AQE enabled
With #4150, we implemented the `GpuSubqueryBroadcastExec` to enable GPU broadcast exchange reuse for DPP. But we don't support it when AQE is enabled. We should support it with AQE, since it may bring up a performance improvement for Spark 320+.

sperlingxx:

Closes #4439

Current PR leveraged "columnar adaptive plan", the new feature of Spark 320, to support GPU replacement of `SubqueryBroadcastExec` when AQE enabled.

In this PR, we removed the `GpuBroadcastToCpu` since it was only for SubqueryBroadcast. To be specific, Broadcast Exchange are inserted under two conditions in terms of Spark SQL: broadcast joins and SubqueryBroadcast in DPP.
* For broadcast joins (BroadcastHashJoin and BroadcastNestedLoopJoin), the build side (BroadcastExchange) will fall back to CPU once the Join plan falls back. There is no need for `GpuBroadcastTocpu` since the BroadcastExchange itself is on the CPU.
* For SubqueryBroadcast, current PR will replace original `AdaptiveSparkPlanExec` with columnar adaptive plan if possible, which eliminate the appearance of `ColumnarToRowExec(e: BroadcastQueryStageExec)`.


sperlingxx:
build

sperlingxx:
build

sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -37,8 +37,7 @@ import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.catalyst.trees.TreeNodeTag
 import org.apache.spark.sql.catalyst.util.ArrayData
 import org.apache.spark.sql.connector.read.Scan
-import org.apache.spark.sql.execution._
-import org.apache.spark.sql.execution.ScalarSubquery
+import org.apache.spark.sql.execution.{ScalarSubquery, _}
```
revans2, line 40: nit

```
import org.apache.spark.sql.execution._
```

revans2 (dismissed):
I really like the comments in GpuSubqueryBroadcastExec. They help to explain what is happening. But just to be sure that I understand everything.  `GpuSubqueryBroadcastExec` takes columnar as input and returns rows.  `GpuBroadcastExchangeExec` always returns columnar data so it can be reused with anything else that needs columnar data.  When AQE is enabled Spark knows that the output is columnar or not so it can adjust without us having to do it, and all we have to do is to match the pattern to know if we can replace this or not.

If I got anything wrong please correct me.  I would also like to see a few others like @andygrove review this too before it is merged in.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuSubqueryBroadcastExec.scala
```diff
@@ -20,18 +20,124 @@ import scala.collection.JavaConverters.asScalaIteratorConverter
...
+    //          +- [GPU overrides of executed subquery...]
+    //
+    case a: AdaptiveSparkPlanExec =>
+      ShimLoader.getSparkShims.getAdaptiveInputPlan(a) match {
+        case ex: BroadcastExchangeExec =>
+          val exMeta = new GpuBroadcastMeta(ex, conf, p, r)
+          exMeta.tagForGpu()
+          if (exMeta.canThisBeReplaced) {
+            broadcastBuilder = () =>
+              ShimLoader.getSparkShims.columnarAdaptivePlan(a, null)
```
andygrove, line 117: I am slightly nervous about passing a `null` for the goal here, even though it is technically safe because the goal isn't currently referenced in the shims for Spark 3.2.0 and later. Perhaps we could use a default goal instead?

```suggestion
              ShimLoader.getSparkShims.columnarAdaptivePlan(a, 
                TargetSize(conf.gpuTargetBatchSizeBytes))
```

andygrove (approved):
LGTM with one nit. Thank you for the comprehensive documentation. That made the PR much easier to review.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: Done

> I really like the comments in GpuSubqueryBroadcastExec. They help to explain what is happening. But just to be sure that I understand everything. `GpuSubqueryBroadcastExec` takes columnar as input and returns rows. `GpuBroadcastExchangeExec` always returns columnar data so it can be reused with anything else that needs columnar data. When AQE is enabled Spark knows that the output is columnar or not so it can adjust without us having to do it, and all we have to do is to match the pattern to know if we can replace this or not.
> 
> If I got anything wrong please correct me. I would also like to see a few others like @andygrove review this too before it is merged in.

Yes

sperlingxx:
> I really like the comments in GpuSubqueryBroadcastExec. They help to explain what is happening. But just to be sure that I understand everything. `GpuSubqueryBroadcastExec` takes columnar as input and returns rows. `GpuBroadcastExchangeExec` always returns columnar data so it can be reused with anything else that needs columnar data. When AQE is enabled Spark knows that the output is columnar or not so it can adjust without us having to do it, and all we have to do is to match the pattern to know if we can replace this or not.
> 
> If I got anything wrong please correct me. I would also like to see a few others like @andygrove review this too before it is merged in.

Yes

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuSubqueryBroadcastExec.scala
sperlingxx: Done.

sperlingxx:
build

revans2 (approved):

andygrove (approved):

PR #4449, opened 2021-12-31
Add recursive type checking and fallback tests for casting array with unsupported element types to string

### LINKED ISSUE #4446: [BUG] GpuCast crashs when casting from Array with unsupportable child type
**Describe the bug**
Currently, GpuCast will crash with cuDF exception when we try to cast from ArrayType which holds an inconvertable child type towards the cast target.

Cast Array of Map to String with GPU enabled

**Expected behavior**
In my opinion, compared with adding a P.S. note in the user documentation, a better approach is to verify child types of Array in `CastExprMeta.recursiveTagExprForGpuCheck`, falling back the Cast expression to CPU if there exists non-replaceable type pairs. It will prevent the crash of  GpuCast from the misbehaviors of users.

HaoYang670:
close #4446 

1. Add recursive type checking for casting array to string.
2. Add `MapType` into unsupported element types in integration tests
3. Replace `Xfail` by `fallback` when casting unsupported types to string in integration tests


HaoYang670:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -124,6 +124,9 @@ final class CastExprMeta[INPUT <: CastBase](
           case (fromChild, toChild) =>
             recursiveTagExprForGpuCheck(fromChild.dataType, toChild.dataType, depth + 1)
         }
+      case (ArrayType(element_type, _), StringType) =>
```
revans2, line 127: nit `element_type` does not follow the convention. Could we make it `elementType`?

revans2 (approved):
Great catch/fix. Just a minor nit

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
HaoYang670: Thank you, I will fix it

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -1270,10 +1270,11 @@ class CastChecks extends ExprChecks {
   val calendarChecks: TypeSig = none
   val sparkCalendarSig: TypeSig = CALENDAR + STRING
 
-  val arrayChecks: TypeSig = STRING + ARRAY.nested(commonCudfTypes + DECIMAL_128_FULL + NULL +
+  val arrayChecks: TypeSig = psNote(TypeEnum.STRING, "the array's children must also support " +
+    "being cast to string") + ARRAY.nested(commonCudfTypes + DECIMAL_128_FULL + NULL +
       ARRAY + BINARY + STRUCT + MAP) +
-      psNote(TypeEnum.ARRAY, "The array's child type must also support being cast to " +
-          "the desired child type")
+      psNote(TypeEnum.ARRAY, "The array's children must also support being cast to the " +
```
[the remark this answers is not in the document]
HaoYang670: Thank you, I will fix it

HaoYang670:
build

HaoYang670:
build

revans2 (approved):

PR #4451, opened 2022-01-03
Some metrics improvements and timeline reporting

revans2:
This is a result of trying to find a heuristic to optimize parquet/orc splits and also looking at buffering times to try and understand if there are more optimizations we could do for HDFS/other distributed file systems.

It fixes some metrics and offers a way to visualize the metrics in the timeline view from the profiling tool. The colors on the timeline were already really bad, and this does not help at all.

Screenshot from 2022-01-03 08-09-22

I don't consider this completely done because I have not documented the metrics reporting yet. I have not done this because I wasn't sure if the colors I have picked are okay. Also I was not sure if we wanted to put in a pattern in addition to a color to make it simpler to see. I also am not sure if this is something we want to have on by default, especially because the semaphore time only happens when debug metrics are enabled. Here is a high level overview.

The bottom half of each task shows the amount of time taken as reported by various metrics.

  * yellow is the deserialization time for the task as reported by Spark.  This works on both CPU and GPU tasks.
  * white is the read time for a task. This is a combination of the "buffer time" SQL metric and the shuffle read time as reported by Spark. The shuffle data works on both CPU and GPU, but the buffer time metric is GPU only.
  * red is the semaphore wait time. This only shows up on GPU tasks when DEBUG metrics are enabled. It does not apply to CPU tasks.
  * green is the "op time" SQL metrics. This is GPU task specific. I am also a little concerned about this because I have seen it be longer than the total time for a task. I fixed one issue with it where the op time included the semaphore time for shuffle coalesce.  But I can still see it for tasks with lots of large joins in them.
  * blue is the write time for a task. This is the "write time" SQL metric and the shuffle write time as reported by Spark. Like with the read time the shuffle metrics work for both GPU and CPU, but the write time metrics are GPU specific.

feedback is appreciated.


tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/GenerateTimeline.scala
```diff
@@ -154,6 +156,25 @@ object GenerateTimeline {
...
+  private def subTimingMark(
+      color: String,
+      startTime: Long,
+      endTime: Long,
+      slot: Int,
+      xStart: Long,
+      yStart: Long,
+      minStart: Long,
+      fileWriter: ToolTextFileWriter): Unit = {
+    val x = xStart + (startTime - minStart)/MS_PER_PIXEL
```
gerashegalov, line 168: nit: spaces around `/`

docs/tuning-guide.md
```diff
@@ -306,32 +306,34 @@ performance.
...
-| numOutputRows    | output rows              | Number of rows that the operator outputs.         |
-| numPartitions    | partitions               | Number of output partitions from a file scan or shuffle exchange. |
+| Key              | Name                     | Description                                                                                                                                                                         |
+|------------------|--------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| bufferTime       | buffer time              | Time spent buffering input from file data sources. This buffering time happens on the CPU, typically with no GPU semaphore held.                                                    |
+| readFsTime       | time to read fs data     | Time spent actually reading the data and writing it to on heap memory. This is a part of `bufferTime`                                                                               |
```
gerashegalov, line 312: hyphenated spelling `on-heap` , `off-heap` is easier to parse

gerashegalov (approved):
LGTM, minor comments

revans2:
build

revans2:
Converting to draft because I found some issues with op time for join that I want to understand better.

revans2:
build

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/GenerateTimeline.scala
```diff
@@ -27,7 +27,9 @@ abstract class TimelineTiming(
     val endTime: Long)
 
 class TimelineTaskInfo(val stageId: Int, val taskId: Long,
-    startTime: Long, endTime: Long, val duration: Long)
+    startTime: Long, endTime: Long, val duration: Long,
```
jlowe, line 30: With all the `val` parameters, should this just be a `case class`?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarOutputWriter.scala
```diff
@@ -102,7 +102,7 @@ abstract class ColumnarOutputWriter(path: String, context: TaskAttemptContext,
       }
```
jlowe, line 102: 2022 copyrights.

revans2: Every file that was touched in 2022 got a 2022 copyright update.
```
$ git log sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarOutputWriter.scala
commit 3869bde288bbf178b56c2310b22d9b4b61009743
Author: Robert (Bobby) Evans <bobby@apache.org>
Date:   Fri Dec 31 09:14:06 2021 -0600

    Fix a metric and clean up timeline a little
```

I didn't think there was a requirement to go back and update the copyright for older files.

tools/src/main/scala/com/nvidia/spark/rapids/tool/profiling/GenerateTimeline.scala
revans2: `TimelineTiming` would then have to change from an abstract class to a trait. I can do it, but it does not feel directly related to the issue being dealt with.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -36,7 +36,7 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
 abstract class AbstractGpuJoinIterator(
     gatherNvtxName: String,
     targetSize: Long,
-    opTime: GpuMetric,
+    val opTime: GpuMetric,
```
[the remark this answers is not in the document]
revans2: I made it a val so that anything that inherits from this can get access to it.  I thought it was a little cleaner when I did it, but I am not sold on it and I am happy to change it back.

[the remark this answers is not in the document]
jlowe: I'm fine with it as-is, mostly was curious on the reasoning since it didn't seem necessary as of now.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarOutputWriter.scala
jlowe: But this wasn't even updated for 2021 copyrights when it was touched last year, and since we have to touch it this year to fix that, I just went with 2022.  If you want to make it 2021 I'm fine with that too.

jlowe (approved):

jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ColumnarOutputWriter.scala
revans2: That is a great point. I missed that we missed the 2021 update. And now I have touched the file in 2022 so...

Issue #4455, opened 2022-01-05
[BUG] join_test.py::test_struct_self_join[IGNORE_ORDER({'local': True})] failed in spark330


pxLi:
**Describe the bug**
**pyspark.sql.utils.IllegalArgumentException: Part of the plan is not columnar class org.apache.spark.sql.execution.datasources.v2.ShowTablesExec**

```
22/01/05 00:21:37 ERROR GpuOverrideUtil: Encountered an exception applying GPU overrides java.lang.IllegalArgumentException: Part of the plan is not columnar class org.apache.spark.sql.execution.datasources.v2.ShowTablesExec

[2022-01-05T00:21:40.201Z] ShowTables [namespace#66, tableName#67, isTemporary#68], V2SessionCatalog(spark_catalog), [default]

[2022-01-05T00:21:40.201Z] 

[2022-01-05T00:21:40.201Z] java.lang.IllegalArgumentException: Part of the plan is not columnar class org.apache.spark.sql.execution.datasources.v2.ShowTablesExec

[2022-01-05T00:21:40.201Z] ShowTables [namespace#66, tableName#67, isTemporary#68], V2SessionCatalog(spark_catalog), [default]

[2022-01-05T00:21:40.201Z] 

[2022-01-05T00:21:40.201Z] 	at com.nvidia.spark.rapids.GpuTransitionOverrides.assertIsOnTheGpu(GpuTransitionOverrides.scala:503) ~[spark330/:?]

[2022-01-05T00:21:40.201Z] 	at com.nvidia.spark.rapids.GpuTransitionOverrides.$anonfun$apply$3(GpuTransitionOverrides.scala:569) ~[spark330/:?]

[2022-01-05T00:21:40.201Z] 	at com.nvidia.spark.rapids.GpuOverrides$.logDuration(GpuOverrides.scala:456) ~[spark3xx-common/:?]

[2022-01-05T00:21:40.201Z] 	at com.nvidia.spark.rapids.GpuTransitionOverrides.$anonfun$apply$1(GpuTransitionOverrides.scala:549) ~[spark330/:?]

[2022-01-05T00:21:40.202Z] 	at com.nvidia.spark.rapids.GpuOverrideUtil$.$anonfun$tryOverride$1(GpuOverrides.scala:3944) ~[spark3xx-common/:?]

[...]

PR #4458, opened 2022-01-05
Exclude some metadata operators when checking GPU replacement

firestarman:
This fixes https://github.com/NVIDIA/spark-rapids/issues/4455 .

The exec for `SHOW TABLES` in Spark 330+ is changed from `ExecutedCommandExec` to `ShowTablesExec`, so add it as an exclusion, along with the `DropTableExec` which also used in the tests.


firestarman:
build

revans2 (approved):
Happy to see that this was a simple fix.

jlowe (approved):

Issue #2252, opened 2021-04-23
[TASK] Use merge sort for nested types when cudf supports it


revans2:
When https://github.com/rapidsai/cudf/issues/8050 is done we should remove all of the work that was done in https://github.com/NVIDIA/spark-rapids/pull/2251 to work around the lack of cudf support.

firestarman:
Reopen the issue [#8050](https://github.com/rapidsai/cudf/issues/8050) since cudf::merge is still missing the array and map type support.

PR #4459, opened 2022-01-05
Use merge sort for struct types in non-key columns

firestarman:
This partially addresses the issue https://github.com/NVIDIA/spark-rapids/issues/2252, since [cudf merge only supports structs](https://github.com/rapidsai/cudf/pull/8422) now for nested types.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/SortUtils.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/SortUtils.scala
@@ -23,7 +23,7 @@ import ai.rapids.cudf.{ColumnVector, NvtxColor, OrderByArg, Table}
 
 import org.apache.spark.sql.catalyst.expressions.{Alias, Attribute, AttributeReference, BoundReference, Expression, NullsFirst, NullsLast, SortOrder}
 import org.apache.spark.sql.rapids.execution.TrampolineUtil
-import org.apache.spark.sql.types.{ArrayType, DataType, MapType, StructType}
+import org.apache.spark.sql.types.{ArrayType, DataType, MapType}
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
 object SortUtils extends Arm {
@@ -211,12 +211,9 @@ class GpuSorter(
     }
   }
 
-  private[this] lazy val hasNestedInKeyColumns = cpuOrderingInternal.exists { order =>
-    projectedBatchTypes(order.child.asInstanceOf[BoundReference].ordinal) match {
-      case _: ArrayType | _: StructType | _: MapType => true
-      case _ => false
-    }
-  }
+  private[this] lazy val hasNestedInKeyColumns = cpuOrderingInternal.exists ( order =>
+    DataTypeUtils.isNestedType(order.child.dataType)
+  )
 
   /** (This can be removed once https://github.com/rapidsai/cudf/issues/8050 is addressed) */
   private[this] lazy val hasUnsupportedNestedInRideColumns = {
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/SortUtils.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/SortUtils.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


firestarman:
Marking as draft because I am doing some bencmarks.

firestarman:
build

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SortUtils.scala
```diff
@@ -226,15 +237,16 @@ class GpuSorter(
...
-          // In the current version of cudf merge does not work for structs or lists (nested types)
+          // In the current version of cudf merge does not work for lists
           // This should be fixed by https://github.com/rapidsai/cudf/issues/8050
-          val hasNested = {
+          val hasUnsupportedNested = {
             val tab = tabs.head
             (0 until tab.getNumberOfColumns).exists { i =>
-              tab.getColumn(i).getType.isNestedType
+              // Map is list of struct in cudf, so checking list type covers the map type.
+              isOrHasListType(tab.getColumn(i))
```
jlowe, line 246: Walking the cudf type is way more expensive, as we have to traverse the JNI and track closing native resources.  Instead we can leverage the Spark type and `TrampolineUtil.dataTypeExistsRecursively`, e.g.:
```scala
          val hasUnsupportedNested = (0 until batches.head.numCols()).exists { i =>
            TrampolineUtil.dataTypeExistsRecursively(batches.head.column(i).dataType(),
              t => t.isInstanceOf[ArrayType] || t.isInstanceOf[MapType])
          }
```

firestarman:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SortUtils.scala
firestarman: Updated

firestarman:
I am going to close this since there is no perf improvment with this PR.

firestarman:
I am going to close this since the perf did not get improved with this PR, according to my local benchmarks as below.

File Size(Parquet)(:GB) | Merge-sort struct(:sec) | Concat-sort struct(:sec)
-- | -- | --
5.7 | 24.78 | 25.29
5.7 | 25.16 | 25.13
15 | 68.07 | 68.01
15 | 70.50 | 71.32

The test is just to read the data, sort the data and write the data.
`spark.read.parquet(dataPath).sort('id').select('id', 'name', 'person').write.mode("overwrite").parquet(outPath)`

Data schema
```
id                         long
name                   struct<firstname:string, middlename:string, lastname:string>
gender                string  
age                      int   
person                 struct<name:struct, gender:string, age:int>    
```

firestarman:
ready for review

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SortUtils.scala
```diff
@@ -210,6 +211,23 @@ class GpuSorter(
     }
   }
 
+  private[this] lazy val hasNestedInKeyColumns = cpuOrderingInternal.exists { order =>
+    projectedBatchTypes(order.child.asInstanceOf[BoundReference].ordinal) match {
+      case _: ArrayType | _: StructType | _: MapType => true
+      case _ => false
+    }
+  }
```
jlowe, line 219: This should leverage `DataTypeUtils.isNestedType`.  Also, do we need to go all the way through the cpuOrdering, downcasting to a bound reference, and indexing the projected batches?  It seems like we can simply examine the original sort order directly, but maybe I'm missing something.  e.g.:
```suggestion
  private[this] lazy val hasNestedInKeyColumns = sortOrder.exists { order =>
    DataTypeUtils.isNestedType(order.child.dataType)
  }
```

jlowe (commented):
Since this is adding new support for a type, are there tests to verify the new functionality is working (and continues to work going forward)?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SortUtils.scala
firestarman: Thanks for review. Updated to leverage the `DataTypeUtils.isNestedType`. And yes, you are right, we'd better to go into a bound reference.

As far as I know, since here we check nested existence in the keys for the `Table.merge`, ideally `cudfOrdering` should be used, instead of the original sort order.

However the `OerderByArg`  has no type info, so use the `cpuOrderingInternal` instead which has the data type, and contains the same key columns with `cudfOrdering`.

firestarman:
> Since this is adding new support for a type, are there tests to verify the new functionality is working (and continues to work going forward)?

Sorry for the confusion. This only changes the path where the code will run into when there is a struct column in the ride along columns, not add a new type support, so the original tests will cover this change.

firestarman:
build

jlowe (approved):

PR #4460, opened 2022-01-05
Abstract the text based PartitionReader

wbo4958:
This PR tries to abstract the text-based PartitionReader which will be applied to CSV reader (This PR added) and the in-coming JSON reader


wbo4958:
build

revans2 (approved):
This looks fine to me. I am less sure about the `hasHeader` flag in the generic text reader. Just because JSON is not going to need it.  But it should be fine.

wbo4958:
> This looks fine to me. I am less sure about the `hasHeader` flag in the generic text reader. Just because JSON is not going to need it. But it should be fine.

Thx, You're right, the headerFlag should not in the abstract class. Removed it. Thx

wbo4958:
build

revans2 (approved):

Issue #3890, opened 2021-10-21
[FEA] Decimal 128 Support: Audit/Cleanup


revans2:
As a part of adding in support for Decimal 128. We did a most of it one operator at a time. Once this is merged into the main branch again. 

- [x] We should go back through and verify that we didn't miss any operators, that all of the `TypeSig` helpers like `TypeSig.gpuNumeric` are either replaced with `numeric` or at least refer to `DECIMAL_128_FULL`.
- [x] Might be good to rename `DECIMAL_128_FULL` to just `DECIMAL_128`. 
- [x] We should go through the tests and make sure that the data gens are all updated to have decimal support correctly in them.
- [x] We should also go through all of the documentation and make sure it has been properly updated to refer to what we can support now.
Just generally go through and clean up.

kuhushukla:
I have been going through the execs and exprs and will address other issues mentioned above shortly.

kuhushukla:
@sameerz I reopened this since we can use this to get even a couple more changes in when hash partitioning is in and rework that I suggested above. Please feel free to close and open just a smaller task if you would prefer.

nartal1:
We are testing decimal_128 in join_tests by including  `decimal_128_gens` in `all_gen` in this PR https://github.com/NVIDIA/spark-rapids/pull/4470 and order by tests are testing decimal_128. Closing.

PR #4462, opened 2022-01-05
Rename DECIMAL_128_FULL and rework usage of TypeSig.gpuNumeric

kuhushukla:
Partially addresses #3890 , specifically the first two sub-tasks.


kuhushukla:
build

kuhushukla:
@revans2 , @sperlingxx thoughts?

revans2:
I am fine with renaming DECIMAL_128_FULL to DECIMAL_128. It had FULL on it before because I thought we would use it to restrict the maximum precision allowed for things like multiply and divide, but the computation was complicated enough that there is not simple cutoff for most operations.

I do not want to mix GPU and non GPU TypeSig categories.  TypeSig.numeric should never be used in a GPU type signature. This is about future proofing things. So if someone adds in a new DECIMAL_256 or something like that in the future we can update one of them and know that it matches the CPU, and not have to worry about it impacting anything on the GPU. It might be nice to rename numeric to be cpuNumber instead so it is clearer.  Again they started out as common, and then when we started to run into decimal situations I separated them, but I didn't want to change the name all at once for backwards compatibility reasons, but I should have.

kuhushukla:
> It might be nice to rename numeric to be cpuNumberic instead so it is clearer. 


@revans2 , that makes sense. I can do that here shortly but then based on the comment I think we may just want to keep the occurrences of `gpuNumeric + DECIMAL_128` as is. Am I right?

revans2:
> @revans2 , that makes sense. I can do that here shortly but then based on the comment I think we may just want to keep the occurrences of `gpuNumeric + DECIMAL_128` as is. Am I right?

I think we have more places that use `gpuNumeric + DECIMAL_128` than we do just `gpuNumeric`. I think it would be cleaner to make `gpuNumeric` include DECIMAL_128 and then in the few places that cannot support it yet, we can have them create a version for what they do support.  And with the latest hashing PR #4470 there should be almost none left.

kuhushukla:
> > @revans2 , that makes sense. I can do that here shortly but then based on the comment I think we may just want to keep the occurrences of `gpuNumeric + DECIMAL_128` as is. Am I right?
> 
> I think we have more places that use `gpuNumeric + DECIMAL_128` than we do just `gpuNumeric`. I think it would be cleaner to make `gpuNumeric` include DECIMAL_128 and then in the few places that cannot support it yet, we can have them create a version for what they do support. And with the latest hashing PR #4470 there should be almost none left.

Ok yes I also think that makes most sense.

kuhushukla:
build

kuhushukla:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/TypeChecks.scala
```diff
@@ -635,7 +635,7 @@ object TypeSig {
   /**
    * All values that correspond to Spark's AtomicType
    */
-  val atomics: TypeSig = numeric + BINARY + BOOLEAN + DATE + STRING + TIMESTAMP
+  val atomics: TypeSig = cpuNumeric + BINARY + BOOLEAN + DATE + STRING + TIMESTAMP
```
revans2, line 638: nit: Could we make this `cpuAtomics` also? Just because it inherits from cpuNumeric.

revans2 (dismissed):
Just one nit.

revans2 (approved):

kuhushukla:
build

PR #4463, opened 2022-01-05
Avoid replacing a hash join if build side is unsupported by the join type

jlowe:
This adds a sanity check to the hash join tagging to ensure that we never try to replace a hash join where the build side matches the join side.  Such join setups do not directly support batching the non-build side and concatenating the results as equivalent to a single batch join.


jlowe:
build

tgravescs (approved):

jlowe:
Odd, looks like it failed to start the CI job for some reason.  Trying again.

jlowe:
build

jbrennan333 (approved):
+1 this looks good to me

PR #4473, opened 2022-01-07
Implement GpuAnd/GpuOr using AST expressions

jbrennan333:

As part of performance analysis for tpc-ds queries, we saw in traces for q28 there were repeating sections with a large number of binary operations.   @revans2 said that some of our binary operations cause a lot of kernel calls, and suggested we convert GpuAnd and GpuOr to use AST expressions.

In initial testing on my desktop, it does look like this shaves > 1 seconds off the queryTime for q28.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
@@ -56,11 +56,6 @@ case class GpuNot(child: Expression) extends CudfUnaryExpression
   }
 }
 
-object GpuLogicHelper {
-  def eqNullAware(lhs: GpuScalar, rhs: Boolean): Boolean =
-    lhs.isValid && (lhs.getValue.asInstanceOf[Boolean] == rhs)
-}
-
 case class GpuAnd(left: Expression, right: Expression) extends CudfBinaryOperator with Predicate {
   override def inputType: AbstractDataType = BooleanType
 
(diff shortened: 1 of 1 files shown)


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -70,63 +65,52 @@ case class GpuAnd(left: Expression, right: Expression) extends CudfBinaryOperato
...
-      }
-      withResource(firstPass) { firstPass =>
-        withResource(r.equalToNullAware(falseVal)) { rIsFalse =>
-          rIsFalse.ifElse(falseVal, firstPass)
+    withResource(new Table(l, r)) {
+      table => {
+        val astOp = new ast.BinaryOperation(ast.BinaryOperator.NULL_LOGICAL_AND,
+          new ast.ColumnReference(0),
+          new ast.ColumnReference(1))
+        withResource(astOp.compile()) { compiledAstExpr =>
```
revans2, line 76: Do you know how much time is spent in compiling the op? I don't have a good way to cache it and clean up afterwards properly, but I would like to know how much it might be contributing?

abellina:
Are pretty sure this is always a win? If not, would we like to keep the binops around behind a setting or perhaps detect that for batch X the binops or AST are faster... then use one or the other.

revans2 (approved):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -70,63 +65,52 @@ case class GpuAnd(left: Expression, right: Expression) extends CudfBinaryOperato
...
+          compiledAstExpr.computeColumn(table)
```
andygrove, line 77: I wrote identical code to this in https://github.com/NVIDIA/spark-rapids/commit/ee162b52d2c12702de5df9af53e1ac5e21563748 and I wonder if we should consider having some common utility code around building these AST expressions. Another option would be exposing this logic in object `GpuLogicHelper` so that it can more easily be re-used elsewhere.

jbrennan333:
@abellina before putting this in, I would definitely like to do a run of all queries on spark2a to see how much this helps across all queries.

revans2:
> Are we pretty sure this is always a win? If not, would we like to keep the binops around behind a setting or perhaps detect that for batch X the binops or AST are faster... then use one or the other.

I don't see how it could not be a win. In my AST testing boolean AND/OR is not a win to replace.  But we are replacing 3 boolean bin ops and two copy if/else calls with a single AST operation. I would expect this to be at least 3x-4x faster compared to the existing implementation. If we could get actual NULL_LOGICAL_AND/NULL_LOGICAL_OR binary operations, then we probably would want to go back to the binops. This is because NULL_EQUALS for a single operation is about 30% faster as a binop than an AST.  But once we combine a few NullEquals together in a single AST call, then the AST starts to win by about 30%

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -70,63 +65,52 @@ case class GpuAnd(left: Expression, right: Expression) extends CudfBinaryOperato
...
-    withResource(Scalar.fromBool(false)) { falseVal =>
-      if (eqNullAware(lhs, false)) {
-        ColumnVector.fromScalar(falseVal, r.getRowCount.toInt)
-      } else {
-        withResource(l.and(r)) { lAndR =>
-          withResource(r.equalToNullAware(falseVal)) { rIsFalse =>
-            rIsFalse.ifElse(falseVal, lAndR)
-          }
+    val lhsLit = if (lhs.isValid) {
+      ast.Literal.ofBoolean(lhs.getValue.asInstanceOf[Boolean])
```
jlowe, line 86: It's mildly unfortunate we don't have a way to directly pass the cudf `Scalar` directly to the AST, as this is unnecessary GPU synchronization (both for the `isValid` and the `getValue`).  Given that ultimately when the AST is compiled it generates a cudf `Scalar` for the literal, (with corresponding GPU synchronization there), it's a lot of unnecessary (if likely very small) synchronizations.  I'll file a cudf issue to support passing cudf `Scalar` instances to an AST directly.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -70,63 +65,52 @@ case class GpuAnd(left: Expression, right: Expression) extends CudfBinaryOperato
...
     val r = rhs.getBase
-    withResource(Scalar.fromBool(false)) { falseVal =>
-      if (eqNullAware(lhs, false)) {
-        ColumnVector.fromScalar(falseVal, r.getRowCount.toInt)
-      } else {
-        withResource(l.and(r)) { lAndR =>
-          withResource(r.equalToNullAware(falseVal)) { rIsFalse =>
-            rIsFalse.ifElse(falseVal, lAndR)
-          }
+    val lhsLit = if (lhs.isValid) {
```
jlowe, line 85: Getting an AST Literal from a `GpuScalar` should be a utility method, as it appears multiple times here and we may want to update all occurrences in the future if Java cudf ASTs are updated to take `Scalar` instances directly.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
jbrennan333: @revans2 does creating common utility code for this fall under your larger task to use AST for more operations?  I can certainly move this to a utility function that accepts the op as an argument.  Also - my current patch removes `GpuLogicHelper`, because I removed the use of `egNullAware()` by GpuAnd and GpuOr.  Do we need to keep it around in case others need it?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
jlowe: Filed rapidsai/cudf#10002 for this.

jlowe:
> I would definitely like to do a run of all queries on spark2a to see how much this helps across all queries.

Marking this as a draft so it isn't approved and merged before those results are obtained.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
jbrennan333: Could GpuLiteral be used here?

jlowe: GpuLiteral is something that can appear in the query plans and represents a query literal that we could translate to the GPU.  However in this case we're handed a GpuScalar as a method parameter, not a GpuLiteral.  If you're asking simply where would the utility method be placed, I think GpuScalar is more appropriate than GpuLiteral since the former is a much closer match to a libcudf AST literal (which is just a wrapper for a GPU scalar value like GpuScalar is).

jbrennan333:
I am going to close this. It is superseded by @revans2's implementation of the NULL_LOGICAL_AND/OR binops.
CUDF: https://github.com/rapidsai/cudf/pull/10016
plugin: https://github.com/NVIDIA/spark-rapids/pull/4501

PR #4476, opened 2022-01-07
Don't acquire the semaphore for empty input while scanning

abellina:

This PR removes instances where we acquired the GpuSemaphore within the scan code, specifically those scans that are `PartitionReader`s. The change is to stop acquiring the semaphore here, because these cases will not be consumed downstream, given how the `PartitionReader` is used. 

`PartitionReader` provides a `next()` method that returns false when there isn't work left. The `PartitionReaderIterator` wraps these readers, as far as I can tell, and it exposes the iterator interface (`hasNext()` returns false if the `PartitionReader` returned a false in `next()`). This allows us to stop requiring us to acquire the semaphore in the `next()` method, when there isn't any work left in the reader.

The PR also adds an acquire in the aggregate where we are producing rows out of nothing. This should be a noop, and likely just a miss that it wasn't there before (protected by upstream nodes doing the acquire for the agg).

The change has modest gains in NDS so far. I want to run a few more times, but so far I see: 10 queries that are 1.20x+ faster, 39 queries that are 1.10x+, and most queries (92) are above 1. I don't see a lot of regression here, except for q94 with 2 seconds, but q94 is falling back to the CPU and can be unpredictable. In other words, it seems to be mostly good, but I want to run a few more tests. 

@revans2 wanted comments explaining why we are not acquiring the semaphore. I wasn't sure where to put them, and they are the same each time. I could see us moving them to the `PartitionReaderIterator` or somewhere else more appropriate.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMultiFileReader.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMultiFileReader.scala
@@ -308,7 +308,6 @@ abstract class MultiFileCloudPartitionReaderBase(
   private var filesToRead = 0
   protected var currentFileHostBuffers: Option[HostMemoryBuffersWithMetaDataBase] = None
   private var isInitted = false
-  private var isFirstBatch = true
   private val tasks = new ConcurrentLinkedQueue[Future[HostMemoryBuffersWithMetaDataBase]]()
   private val tasksToRun = new Queue[Callable[HostMemoryBuffersWithMetaDataBase]]()
   private[this] val inputMetrics = Option(TaskContext.get).map(_.taskMetrics().inputMetrics)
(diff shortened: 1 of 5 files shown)


abellina:
hmm looks like the build is failing on not being able to create the docker container.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMultiFileReader.scala
```diff
@@ -308,7 +308,6 @@ abstract class MultiFileCloudPartitionReaderBase(
   private var filesToRead = 0
```
jlowe, line 308: 2022 copyrights

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScanBase.scala
```diff
@@ -611,7 +611,6 @@ class GpuOrcPartitionReader(
     override val execMetrics : Map[String, GpuMetric],
```
jlowe, line 611: 2022 copyrights

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScanBase.scala
```diff
@@ -1458,7 +1458,6 @@ class ParquetPartitionReader(
   with ParquetPartitionReaderBase {
```
jlowe, line 1458: 2022 copyrights

jlowe (commented):
Looks good to me other than copyright nits.

jlowe (dismissed):

jlowe:
build

abellina:
I ran another iteration of NDS, with similar results. Most queries are slightly faster, with some more significant gains in those single digit queries like q2 q53 q55 where we are gaining up to 3 seconds.

Overall I see ~42 seconds over all of NDS at 2TB.

jlowe (approved):

jlowe:
build

jbrennan333 (approved):
+1 this looks good to me

Issue #2441, opened 2021-05-18
[FEA] Use AST for filter in join APIs


revans2:
**Is your feature request related to a problem? Please describe.**
This is follow on work for #2433 and #2440. The goal is to take the AST join work from CUDF when it is ready and use it. We might want to remove the split check that happens before doing a join if it was not removed as a part of #2440. This is because that check cannot take into account the inequality join parts that can happen, so it will be really inaccurate in those cases, not just the fairly inaccurate that it is now.

sameerz:
Will move this to 22.04 for the remaining task related to mixed joins for conditional hash semi and anti joins.

PR #4477, opened 2022-01-07
Use libcudf mixed joins for conditional hash joins [databricks]

jlowe:
Related to #2441.  Depends on rapidsai/cudf#9941.

This implements GPU support for LeftOuter, RightOuter, FullOuter, and Inner joins using libcudf's mixed join feature.  As before, inner joins will fallback to a post-filter approach for the condition if the condition cannot be expressed as a cudf AST.


jbrennan333 (commented):
+1 this looks good to me.

jlowe:
cudf dependency has been merged and published, so this should be ready to go.

jlowe:
build

jbrennan333 (approved):
+1 lgtm

PR #4471, opened 2022-01-07
Disable orc write by default because of https://issues.apache.org/jira/browse/ORC-1075

wbo4958:
Spark 3.1.1+ failed to read the orc file written by cudf when filter is pushed down.

This PR disables orc write by default and gives some information about the reason.

For more details, please refer to  https://issues.apache.org/jira/browse/ORC-1075

Signed-off-by: Bobby Wang <wbo4958@gmail.com>

PR #4481, opened 2022-01-10
Disable orc write by default because of https://issues.apache.org/jira/browse/ORC-1075

wbo4958:
This is duplicated PR of https://github.com/NVIDIA/spark-rapids/pull/4471 which can't be merged because of CI issue.


wbo4958:
build

PR #4484, opened 2022-01-10
Update GpuFileFormatWriter to stay in sync with recent Spark changes, but still not support writing Hive bucketed table on GPU.

### LINKED ISSUE #3949: [Audit][SPARK-32709][SQL] Support writing Hive bucketed table (Parquet/ORC format with Hive hash)
We have copied some code(def wrtie) from FileFormatWriter.scala. We need to keep GpuFileFormatWriter.scala in sync with Apache Spark's file.
PR: https://github.com/apache/spark/commit/4a34db9a17

HaoYang670:
Signed-off-by: remzi 13716567376yh@gmail.com
close #3949

In this PR, we:
1. Copy some code from  apache/spark@4a34db9 which supports Hive bucketed writing. However, as we do not support Hive hash partition on GPU, we throw an exception in Rapids.
2. Create a new file, datasourcev2_write_test.py and add a test here to make sure we are falling back to CPU when someone tries to do a Hive bucketed write.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
@@ -17,7 +17,6 @@
 package org.apache.spark.sql.rapids
 
 import java.util.{Date, UUID}
-
 import ai.rapids.cudf.ColumnVector
 import com.nvidia.spark.TimingUtils
 import com.nvidia.spark.rapids._
@@ -26,7 +25,6 @@ import org.apache.hadoop.fs.Path
 import org.apache.hadoop.mapreduce._
 import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat
 import org.apache.hadoop.mapreduce.task.TaskAttemptContextImpl
-
 import org.apache.spark.{SparkException, TaskContext}
 import org.apache.spark.internal.Logging
 import org.apache.spark.internal.io.{FileCommitProtocol, SparkHadoopWriterUtils}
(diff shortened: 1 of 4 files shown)


HaoYang670:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
```diff
@@ -17,7 +17,6 @@
 package org.apache.spark.sql.rapids
 
 import java.util.{Date, UUID}
-
```
jlowe, line 20: This whitespace should not have been deleted per our style guidelines.  See https://github.com/NVIDIA/spark-rapids/blob/branch-22.02/docs/dev/idea-code-style-settings.xml#L11-L17

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
```diff
@@ -26,7 +25,6 @@ import org.apache.hadoop.fs.Path
 import org.apache.hadoop.mapreduce._
 import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat
 import org.apache.hadoop.mapreduce.task.TaskAttemptContextImpl
-
```
jlowe, line 29: Restore whitespace per style guide.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
```diff
@@ -128,19 +126,17 @@ object GpuFileFormatWriter extends Logging {
...
+    val writerBucketSpec = bucketSpec.map { spec =>
+      // TODO: Cannot support this until we:
+      // support Hive hash partitioning on the GPU
       throw new UnsupportedOperationException("GPU hash partitioning for bucketed data is not "
-          + "compatible with the CPU version")
+        + "compatible with the CPU version")
+
+      // add this to let the compiler infer
+      // the type of writerBucketSpec as `Option[GpuWriterBucketSpec]`
+      GpuWriterBucketSpec(null, null)
```
jlowe, line 137: This should be getting a dead code warning from the compiler.  If we want the compiler to know that `writerBucketSpec` is of type `GpuWriterBucketSpec` then a much better way is to simply declare it as such, i.e.:
```scala
val writerBucketSpec: GpuWriterBucketSpec = bucketSpec.map { spec =>
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
```diff
@@ -128,19 +126,17 @@ object GpuFileFormatWriter extends Logging {
...
-      // TODO: Cannot support this until we either:
-      // Guarantee join and bucketing are both on the GPU and disable GPU-writing if join not on GPU
-      //   OR
-      // Guarantee GPU hash partitioning is 100% compatible with CPU hashing
+    val writerBucketSpec = bucketSpec.map { spec =>
+      // TODO: Cannot support this until we:
+      // support Hive hash partitioning on the GPU
       throw new UnsupportedOperationException("GPU hash partitioning for bucketed data is not "
-          + "compatible with the CPU version")
+        + "compatible with the CPU version")
```
jlowe, line 133: Nit: Style guide is 4-space indent on continuation, so the original code was properly indented.

jlowe (commented):
Headline implies that this adds support for writing Hive bucketed tables, but it is not supported even after this PR is merged.  We weren't accidentally trying to support this before the PR either, so maybe headline should just state we're updating GpuWriteJobDescription to stay in sync with recent Spark changes.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
HaoYang670: > Headline implies that this adds support for writing Hive bucketed tables, but it is not supported even after this PR is merged. We weren't accidentally trying to support this before the PR either, so maybe headline should just state we're updating GpuWriteJobDescription to stay in sync with recent Spark changes.

Thank you! I have modified the headline

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
HaoYang670: Thank you! I have fixed it

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
HaoYang670: Thank you! I have fixed it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
HaoYang670: Thank you! I have fixed it.

HaoYang670:
build

HaoYang670:
build

jlowe (commented):
Looks better, but tests are failing with:
```
The format of the existing table default.tmp_table_574308_0 is `HiveFileFormat`. It doesn't match the specified format `ParquetDataSourceV2`.
```

We may need a `.format("hive")` when writing the dataframe, but I'm not an expert on Spark's Hive support.

revans2:
Should we update the title of this PR as we are not adding in support for hived bucketed table writes.

HaoYang670:
build

HaoYang670:
The title has been updated to tell users that we do not support writing hive bucketed table on GPU so far.

revans2 (approved):

jlowe (approved):

PR #4485, opened 2022-01-10
Add json reader support

wbo4958:
This PR is trying to support the json reader on basic type support.


wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -901,6 +901,17 @@ object RapidsConf {
       .booleanConf
       .createWithDefault(false)
 
+  val ENABLE_JSON = conf("spark.rapids.sql.format.json.enabled")
+    .doc("When set to false disables all json input and output acceleration. " +
+      "(only input is currently supported anyways)")
+    .booleanConf
+    .createWithDefault(true)
```
revans2, line 908: These should not be true by default. There are just too many problems with the parsing support right now for us to have them be true.

wbo4958:
build

wbo4958:
build

wbo4958:
build

revans2 (commented):
It is looking good. My main concerns at this point is adding some documentation to docs/compatibility.md. Even if that is just that this is a very experimental feature and has a lot of problems.  More specifics about what does not work is better.

I also wanted to make sure that @tgravescs and @nartal1 are okay with JSON reads not being flagged in the qualification tool any more.  We generate that file automatically, but I am not 100% sure with our current level of support that we want JSON to be on the truly supported list.

tgravescs:
since the supported data sources file has it as CO the qualification tool should still report it as not supported properly, but we should test to make sure @nartal1 can you test?

tgravescs:
sorry, I didn't fully look at the PR, looks like the qualification results were update, only diff is it reports each type. They are properly listed under not supported still:
=> Unsupported Read File Formats and Types,Unsupported Write Data Format

nartal1:
> sorry, I didn't fully look at the PR, looks like the qualification results were update, only diff is it reports each type. They are properly listed under not supported still: => Unsupported Read File Formats and Types,Unsupported Write Data Format

Thanks @tgravescs for looking into it. Yes, Instead of json[*], it's now reporting the types which are not supported which is good.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
wbo4958: Done

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
```diff
@@ -0,0 +1,238 @@
...
+        // fallback to cpu to throw exception
+        meta.willNotWorkOnGpu("GpuJsonScan does not support Corrupt Record which must " +
+          "be string type and nullable")
+      }
+    }
+
+    if (readSchema.length == 1 &&
+      readSchema.head.name == parsedOptions.columnNameOfCorruptRecord) {
+      // fallback to cpu to throw exception
+      meta.willNotWorkOnGpu("GpuJsonScan dose not support Corrupt Record")
```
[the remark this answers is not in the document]
wbo4958: Thx Done.

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    // {name: 'Reynold Xin'} is not supported by CUDF
+    if (parsedOptions.allowUnquotedFieldNames) {
+      meta.willNotWorkOnGpu("GpuJsonScan does not support allowUnquotedFieldNames")
+    }
+
+    // {'name': 'Reynold Xin'} is not supported by CUDF
+    if (options.get("allowSingleQuotes").map(_.toBoolean).getOrElse(false)) {
+      meta.willNotWorkOnGpu("GpuJsonScan dose not support allowSingleQuotes")
```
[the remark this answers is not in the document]
wbo4958: Thx Done

wbo4958:
build

wbo4958:
build

docs/compatibility.md
```diff
@@ -434,6 +434,49 @@ The plugin supports reading `uncompressed`, `snappy` and `gzip` Parquet files an
 fall back to the CPU when reading an unsupported compression format, and will error out in that
 case.
 
+## JSON
+
+The JSON format read is a very experimental feature which is expected to have some issues, so we disable 
+it by default. If you would like to test it, you can enable `spark.rapids.sql.format.json.enabled` and 
+`spark.rapids.sql.format.json.read.enabled`.
+
+Currently, the JSON reader doesn't support the column prune when reading JSON files, which is the blocker
```
revans2, line 443: nit: can we reword it as

```
Currently, the GPU accelerated JSON reader doesn't support column pruning, which will likely make
this difficult to use or even test.  The user must specify the full schema ...
```

Also what happens with predicate push down?  If we don't use one of those columns in a query I thought that Spark would push down to skip reading it, even if we said it was in the schema. So in the example

```
val df = spark.read.json("people.json")
df.selectExpr("SUM(age)").show()
```

would that work or not?  If not then we need to document that part too.

docs/compatibility.md
```diff
@@ -434,6 +434,49 @@ The plugin supports reading `uncompressed`, `snappy` and `gzip` Parquet files an
...
+  val df = spark.read.schema(schema).json("people.json")
+  ```
+
+While the below code will not work in the current version,
+
+``` scala
+val schema = StructType(Seq(StructField("name", StringType)))
+val df = spark.read.schema(schema).json("people.json")
+```
+
```
revans2, line 475: Can we mention that Arrays, Structs and Maps are not supported?

revans2 (commented):
Just a few comments about the docs.

docs/compatibility.md
wbo4958: It will work since this PR passes the whole dataSchema to cudf to read all columns, and then do the column pruning according to the readSchema

docs/compatibility.md
wbo4958: Done

wbo4958: Done

docs/compatibility.md
wbo4958: Done, changed the doc

wbo4958:
build

wbo4958:
build

docs/compatibility.md
```diff
@@ -434,6 +434,54 @@ The plugin supports reading `uncompressed`, `snappy` and `gzip` Parquet files an
...
+### JSON supporting types
+
+The nested types(array, map and struct) are not supported yet in current version.
```
andygrove, line 479: Does this mean that it is not possible to access primitive types stored within an array, map, and struct, or just that we cannot return a column of type array, map, or struct?

revans2: >  Does this mean that it is not possible to access primitive types stored within an array, map, and struct,
Yes
> or just that we cannot return a column of type array, map, or struct?
Spark does not treat them differently as far as I can tell. CUDF sort of does but not in any way that is compatible with what Spark does and it is only for top level arrays.

revans2 (approved):
But I would like others to look at this too

wbo4958:
@tgravescs Could you help to review this PR ?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTextBasedPartitionReader.scala
```diff
@@ -212,6 +207,24 @@ abstract class GpuTextBasedPartitionReader(
...
+  /**
+   * Handle the table decoded by GPU
+   * @param readDataSchema the Spark schema describing what will be read
+   * @param table the table decoded by GPU
+   * @return the new optional Table
+   */
+  def handleResult(readDataSchema: StructType, table: Table): Option[Table] = {
+    val numColumns = table.getNumberOfColumns
+
+    closeOnExcept(table) { _ =>
```
jlowe, line 219: This is odd.  Normally there are two ways to handle a GPU resource being passed to a method:
1. The method owns it and is responsible for closing the GPU resource, exceptions or not.
2. The caller owns it and is responsible for closing it, exceptions or not.

However this method closes it only when there's an exception but otherwise does not.  For this to work properly, that implies the caller is not protecting the GPU resource it owns from an exception.  If it is properly doing that, the caller will end up double-closing the GPU resource because this will have closed it first.  IMO this method should not own the table and rely on the caller to properly handle exceptions, from here or anywhere else while it is responsible for the table.

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
```diff
@@ -0,0 +1,326 @@
...
+      return Some(table)
+    }
+
+    // JSON will read all columns in dataSchema due to https://github.com/rapidsai/cudf/issues/9990,
+    // but actually only the columns in readDataSchema are needed, we need to do the "column" prune,
+    // Once the FEA is shipped in CUDF, we should remove this hack.
+
+    val prunedCols = readDataSchema.fieldNames.map { name =>
+      val optionIndex = dataSchema.getFieldIndex(name)
+      closeOnExcept(table) { _ =>
```
jlowe, line 313: Same "who really owns the table?" comments apply here as well.

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTextBasedPartitionReader.scala
wbo4958: This case should match the scenario `The method owns it and is responsible for closing the GPU resource, exceptions or not.`

For the GPU resource handling convention, we should close the input table and return a new
table just like the below code. But considering the code simplicity and optimization, we just return the input table.

withResource(table) { _
   val cols = (0 until  table.getNumberOfColumns).map(i => table.getColumn(i))
  Some(new Table(cols: _*))
}

wbo4958: Added some comments in the code.

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
wbo4958: same as above reply

jlowe (approved):

PR #4491, opened 2022-01-10
Skip the test for spark versions 3.1.1, 3.1.2 and 3.2.0 only [databricks]

### LINKED ISSUE #4488: [BUG] isCastingStringToNegDecimalScaleSupported seems set wrong for some Spark versions
**Describe the bug**
In https://github.com/NVIDIA/spark-rapids/pull/4172/files we added a shim function isCastingStringToNegDecimalScaleSupported.
Based on the resolution of the Spark issue https://issues.apache.org/jira/browse/SPARK-37451 some of those settings seem wrong.

for instance it seems like its fixed in 3.1.3 and 3.2.1 and then also 3.3.0 so we should update the shims to match.

razajafri:
The fix was posted for 3.1.3+ and 3.2.1+ which makes the bug valid only in version 3.1.1, 3.1.2 and 3.2.0

Fixes #4488 

### DIFF
--- a/integration_tests/src/main/python/spark_session.py
+++ b/integration_tests/src/main/python/spark_session.py
@@ -111,10 +111,6 @@ def is_before_spark_320():
 def is_before_spark_330():
     return spark_version() < "3.3.0"
 
-# The bug SPARK-37451 only affects the following versions
-def is_neg_dec_scale_bug_version():
-    return ("3.1.1" <= spark_version() < "3.1.3") or ("3.2.0" <= spark_version() < "3.2.1")
-
 def is_databricks91_or_later():
     spark = get_spark_i_know_what_i_am_doing()
     return spark.conf.get("spark.databricks.clusterUsageTags.sparkVersion", "") >= "9.1"
--- a/integration_tests/src/main/python/spark_session.py
+++ b/integration_tests/src/main/python/spark_session.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2021, NVIDIA CORPORATION.
+# Copyright (c) 2020-2022, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -110,6 +110,10 @@ def is_before_spark_320():
 
 def is_before_spark_330():
     return spark_version() < "3.3.0"
+
+# The bug SPARK-37451 only affects the following versions
+def is_neg_dec_scale_bug_version():
+    return ("3.1.1" <= spark_version() < "3.1.3") or ("3.2.0" <= spark_version() < "3.2.1")
 
 def is_databricks91_or_later():
     spark = get_spark_i_know_what_i_am_doing()
--- a/sql-plugin/src/main/31xdb/scala/com/nvidia/spark/rapids/shims/v2/Spark31XdbShimsBase.scala
+++ b/sql-plugin/src/main/31xdb/scala/com/nvidia/spark/rapids/shims/v2/Spark31XdbShimsBase.scala
@@ -128,6 +128,4 @@ trait Spark31XdbShimsBase extends SparkShims {
   }
 
   override def shouldFallbackOnAnsiTimestamp(): Boolean = false
-
-  override def isCastingStringToNegDecimalScaleSupported: Boolean = false
 }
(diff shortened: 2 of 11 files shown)


integration_tests/src/main/python/spark_session.py
```diff
@@ -111,6 +111,10 @@ def is_before_spark_320():
 def is_before_spark_330():
     return spark_version() < "3.3.0"
 
+# The bug SPARK-37451 only affects the following versions
+def is_neg_dec_scale_bug_version():
```
jlowe, line 115: Shouldn't this definition be in the one file that needs it, cast_test.py?  I don't see us needing to check for this one very specific bugfix across tests in general.

jlowe (commented):
I expected to see a change to Spark31XShims to have `isCastingStringToNegDecimalScaleSupported` return false, but it's not here.  I don't see how the spark311 and spark312 shims are picking up the changes to mark them properly.

integration_tests/src/main/python/spark_session.py
```diff
@@ -111,6 +111,10 @@ def is_before_spark_320():
...
+    return ("3.1.1" <= spark_version() < "3.1.3") or ("3.2.0" <= spark_version() < "3.2.1")
```
tgravescs, line 116: the last part can just be spark_version == 3.2.0

tgravescs:
31xShim already has it set to false:
https://github.com/NVIDIA/spark-rapids/blob/branch-22.02/sql-plugin/src/main/311until320-nondb/scala/com/nvidia/spark/rapids/shims/v2/Spark31XShims.scala#L546

I believe you can remove it from Spark330Shims as well now.

sql-plugin/src/main/31xdb/scala/com/nvidia/spark/rapids/shims/v2/Spark31XdbShimsBase.scala
```diff
@@ -129,5 +129,5 @@ trait Spark31XdbShimsBase extends SparkShims {
 
   override def shouldFallbackOnAnsiTimestamp(): Boolean = false
 
-  override def isCastingStringToNegDecimalScaleSupported: Boolean = true
+  override def isCastingStringToNegDecimalScaleSupported: Boolean = false
```
tgravescs, line 132: this does seem a bit dangerous in the sense if we add a databricks 3.1.3 shim we could easily miss this.
perhaps at least put a comment here.  I assume the test won't fail if we get it wrong?

razajafri:
> I expected to see a change to Spark31XShims to have `isCastingStringToNegDecimalScaleSupported` return false, but it's not here. I don't see how the spark311 and spark312 shims are picking up the changes to mark them properly.

It's already there in the Spark31XShims https://github.com/razajafri/spark-rapids-1/blob/9e601d639caafef010e33717a7e46f82dd2c92e5/sql-plugin/src/main/311until320-nondb/scala/com/nvidia/spark/rapids/shims/v2/Spark31XShims.scala#L546. That should cover it?

integration_tests/src/main/python/spark_session.py
razajafri: Sure, I can move this but I wasn't sure we like accessing `spark_version()` method outside of `spark_session.py`

integration_tests/src/main/python/spark_session.py
razajafri: What if we have `3.2.0-cdh` or whatever?

tgravescs: ah that is a good point, could have a minor version past the 3.2.0, so leave it as you have it.

sql-plugin/src/main/31xdb/scala/com/nvidia/spark/rapids/shims/v2/Spark31XdbShimsBase.scala
razajafri: The test isn't skipped for 3.1.3, it will blow up saying `part of the plan isn't columnar` as there is a check in `GpuCast` for it if we forget to override it for Spark-3.1.3-db. 

I can move this to Spark312dbShims.

razajafri:
build

tgravescs (approved):

jlowe (approved):

razajafri:
build

jlowe (approved):

Issue #4486, opened 2022-01-10
[BUG] Regular expressions with hex digits not working as expected


andygrove:
**Describe the bug**

I am seeing cuDF fail to match some hex digits correctly. Here is one example. I do not yet know if this is related to https://github.com/NVIDIA/spark-rapids/issues/4409 or whether this is a separate issue.

```
javaPattern=\xA9, cudfPattern=\xA9, input='©', cpu=true, gpu=false
```

**Steps/Code to reproduce bug**

```scala
  test("compare CPU and GPU: hex") {
    val patterns = Seq(raw"\x61", raw"\xA9", raw"\xa9")
    val inputs = Seq("a", "b", "\u00A9")
    assertCpuGpuMatchesRegexpFind(patterns, inputs)
  }
```

**Expected behavior**
Test should pass.

**Environment details (please complete the following information)**
N/A

**Additional context**
N/A

andygrove:
This is perhaps due to cuDF requiring UTF-8 and the fact that the copyright symbol has bytes 0xC2 0xA9 or 11000010 10101001 in UTF-8 encoding, and byte 0xA9 in ANSI encoding.

andygrove:
Now that https://github.com/NVIDIA/spark-rapids/pull/4492 is merged, we fall back to CPU for now for hex digits.

NVnavkumar:
With cuDF PR [#10220](https://github.com/rapidsai/cudf/pull/10220), at least 127 characters should be supported by the cuDF regex hex escape, so this should be re-enabled and tested.

PR #4492, opened 2022-01-10
Fall back to CPU for regular expressions containing hex digits

andygrove:

cuDF and Spark are not consistent in handling certain ranges of hex digits, so we fall back to the CPU for now.

Issue https://github.com/NVIDIA/spark-rapids/issues/4486 has been filed for fixing the hex digit support in the future.

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
@@ -148,6 +148,7 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
   }
 
   ignore("known issue - octal digit") {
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -220,7 +221,8 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
     assertCpuGpuMatchesRegexpFind(patterns, inputs)
   }
 
-  test("compare CPU and GPU: hex") {
+  ignore("compare CPU and GPU: hex") {
```
jlowe, line 224: I thought these tests wouldn't be turned off but instead turned into assertions that we do not support such patterns and thus should fallback.

jlowe (approved):

jlowe (approved):

jlowe:
build

PR #4495, opened 2022-01-10
Enable approx_percentile by default

andygrove:

Enable approx_percentile by default.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -785,7 +785,7 @@ The GPU implementation of `approximate_percentile` uses
 [t-Digests](https://arxiv.org/abs/1902.04023) which have high accuracy, particularly near the tails of a
 distribution. Because the results are not bit-for-bit identical with the Apache Spark implementation of
 `approximate_percentile`, this feature is disabled by default and can be enabled by setting
-`spark.rapids.sql.expression.ApproximatePercentile=true`.
+`spark.rapids.sql.incompatibleOps.enabled=true`.
 
 There is also a known issue ([issue #4060](https://github.com/NVIDIA/spark-rapids/issues/4060)) where
 incorrect results are produced intermittently.
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -3249,6 +3249,11 @@ object GpuOverrides extends Logging {
       (c, conf, p, r) => new TypedImperativeAggExprMeta[ApproximatePercentile](c, conf, p, r) {
 
         override def tagAggForGpu(): Unit = {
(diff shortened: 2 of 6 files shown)


tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3249,6 +3249,11 @@ object GpuOverrides extends Logging {
       (c, conf, p, r) => new TypedImperativeAggExprMeta[ApproximatePercentile](c, conf, p, r) {
 
         override def tagAggForGpu(): Unit = {
```
revans2, line 3251: If you look at `Upper` or `Lower` as an example there is a built in way to mark an entire operator as incompat and it lets us put in a doc string with it to explain how it is incompat.

```
}).incompat("The GPU implementation of approx_percentile is not bit-for-bit " +
    "compatible with Apache Spark.")
```

revans2 (approved):
Just a small nit, but it looks good even without it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
andygrove: Ah, I didn't realize we had that. That looks much better and I will update this.

docs/compatibility.md
```diff
@@ -785,7 +785,7 @@ The GPU implementation of `approximate_percentile` uses
 [t-Digests](https://arxiv.org/abs/1902.04023) which have high accuracy, particularly near the tails of a
 distribution. Because the results are not bit-for-bit identical with the Apache Spark implementation of
 `approximate_percentile`, this feature is disabled by default and can be enabled by setting
-`spark.rapids.sql.expression.ApproximatePercentile=true`.
+`spark.rapids.sql.incompatibleOps.enabled=true`.
```
jlowe, line 788: This says that approx percentile is disabled by default and can be enabled by setting `incompatibleOps.enabled=true`, yet the configs.md document states that it _is enabled by default_ and does not mention anything about being incompatible with Spark's implementation.  Looks like the meta class for approx percentile in GpuOverrides should have been marked with `.incompat` and text explaining what is not 100% compatible with Spark.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3249,6 +3249,11 @@ object GpuOverrides extends Logging {
...
+          if (!conf.isIncompatEnabled) {
```
jlowe, line 3252: Rather than doing this check explicitly, we should be using the `incompat` method on the meta to do this check and also generate documentation.

jlowe (approved):

revans2 (approved):

PR #4501, opened 2022-01-11
Switch and/or to use new cudf binops to improve performance

revans2:
This depends on https://github.com/rapidsai/cudf/pull/10016

Seeing some good performance improvements, but need to run some more tests.


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -56,81 +56,14 @@ case class GpuNot(child: Expression) extends CudfUnaryExpression
   }
```
jlowe, line 56: 2022 copyrights

abellina (commented):
looks good (other than copyright)

jbrennan333 (commented):
Looks good to me as well.

jlowe:
I forgot there are cases where code was using the AST version of the NULL_LOGICAL_OR operator, see https://github.com/NVIDIA/spark-rapids/blob/branch-22.02/sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala#L445-L451. It would be good to have that code switch to using this new direct binop instead.

revans2:
I ran the tests manually and fixed an issue with AST not being translated properly because it was separate from the normal code.  That is fixed now. 

I ran TPC-DS at scale factor 3k and it looks to be generally a win, 1.1% faster. But there are about as many tests that are slower 51 as there are that are faster 54. I am going to see if I can reproduce some of the slowness locally to try and dig in a little deeper.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
```diff
@@ -282,7 +282,9 @@ object CudfBinaryExpression {
     BinaryOp.LESS -> ast.BinaryOperator.LESS,
     BinaryOp.LESS_EQUAL -> ast.BinaryOperator.LESS_EQUAL,
     BinaryOp.LOGICAL_AND -> ast.BinaryOperator.NULL_LOGICAL_AND,
+    BinaryOp.NULL_LOGICAL_AND -> ast.BinaryOperator.NULL_LOGICAL_AND,
```
jlowe, line 285: Nit: It would be nice if this was kept in lexicographical order as it was before.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuExpressions.scala
```diff
@@ -282,7 +282,9 @@ object CudfBinaryExpression {
     BinaryOp.LESS -> ast.BinaryOperator.LESS,
```
jlowe, line 282: 2022 copyrights

revans2:
Dependency just got merged into cudf. I'll move this out of draft when the dependency goes through CI

revans2:
build

jlowe (approved):

Issue #4506, opened 2022-01-12
[FEA]Improve GpuSequence


firestarman:
This is to track the bugs related to GpuSequence

1. https://github.com/NVIDIA/spark-rapids/issues/4499
2. When start < stop but with negative step, CPU blows up, but GPU outputs empty list.
3. GPU is missing max length check for arrays

Issue #4499, opened 2022-01-11
[BUG]GpuSequence blows up when nulls exist in any of the inputs (start, stop, step)


firestarman:
**Describe the bug**
GpuSequence blows up when nulls exist in any of the inputs (start, stop, step), e.g.
```
Caused by: ai.rapids.cudf.CudfException: cuDF failure at: ../src/lists/sequences.cu:148: steps input column must not have nulls.
  at ai.rapids.cudf.ColumnVector.sequences(Native Method)
  at ai.rapids.cudf.ColumnVector.sequence(ColumnVector.java:533)

```
but Spark just returns nulls. 
```
+---+---+----------------------+
| _1| _2|sequence(_1, _2, NULL)|
+---+---+----------------------+
|  1|  2|                  null|
|  3|  4|                  null|
+---+---+----------------------+

```

**Steps/Code to reproduce bug**
Launch a Spark shell with GPU support, and run
`Seq((1, 2), (3, 4)).toDF.repartition(1).selectExpr("*", "sequence(_1, _2, null)").show`

**Expected behavior**
We should keep align with what Spark does for null cases.

firestarman:
Filed the FEA issue #10012 to cudf for this

PR #4520, opened 2022-01-13
GpuSequence refactor[databricks]

firestarman:
This refactor mainly covers:

- Merge the `GpuSequence` and `GpuSequenceWithStep` into one `GpuSequence` to avoid some duplicate code. And get ready for the shim layers that will come with time/date types support.
- Let `GpuSequence` extends from `GpuExpression` instead of the original binary or ternary expression to handle the children evaluation directly.
- Follow the Spark algorithm to compute the sequence lengths, supporting the same illegal inputs check.
- Update the tests, and add the tests for illegal cases.

Fixes https://github.com/NVIDIA/spark-rapids/issues/4506
Fixes https://github.com/NVIDIA/spark-rapids/issues/4499

### DIFF
--- a/integration_tests/src/main/python/collection_ops_test.py
+++ b/integration_tests/src/main/python/collection_ops_test.py
@@ -14,7 +14,7 @@
 
 import pytest
 
-from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_and_cpu_are_equal_sql
+from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_and_cpu_error
 from data_gen import *
 from pyspark.sql.types import *
 from spark_session import with_cpu_session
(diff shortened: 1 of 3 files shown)


firestarman:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -407,260 +408,284 @@ class GpuSequenceMeta(
...
+      return true
+    }
+    if (col.hasNulls) {
+      return false
+    }
+    withResource(col.all()) { allTrue =>
+      // Guaranteed there is at least one row and no nulls so result must be valid
+      allTrue.getBoolean
+    }
   }
```
jlowe, line 436: This is the same as `GpuConditionalExpression.isAllTrue`.  Please refactor to a utility object class, like BoolUtils or something, so we're not copying this code everywhere.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -407,260 +408,284 @@ class GpuSequenceMeta(
...
+    // For example, if the predicate evaluates to [F, null, T, F, T] then this
+    // gets translated first to [0, 0, 1, 0, 1] and then the scan operation
+    // will perform an exclusive sum on these values and produce [0, 0, 0, 1, 1].
+    // Combining this with the original predicate boolean array results in the two
+    // T values mapping to indices 0 and 1, respectively.
+    val boolAsInt = withResource(Scalar.fromInt(1)) { one =>
+      withResource(Scalar.fromInt(0)) { zero =>
+        predicate.ifElse(one, zero)
+      }
+    }
```
jlowe, line 452: This should be refactored as part of the new utilities class, same as `boolToInt` in `GpuConditionalExpression`.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -407,260 +408,284 @@ class GpuSequenceMeta(
...
-
-  override def left: Expression = start
-
-  override def right: Expression = stop
-
-  override def dataType: DataType = ArrayType(start.dataType, containsNull = false)
 
-  override def withTimeZone(timeZoneId: String): TimeZoneAwareExpression =
-    copy(timeZoneId = Some(timeZoneId))
+  /** (Copied most of the part from GpuConditionalExpression) */
```
jlowe, line 438: Please refactor to reduce code duplication, as this seems to be essentially the same.  If the original `GpuConditionalExpression` code was refactored into a utility method that took `ColumnVector` arguments then I think it could be reused verbatim.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -407,260 +408,284 @@ class GpuSequenceMeta(
...
+    withResource(Scalar.fromByte(0.toByte)) { zero =>
+      withResource(new Table(start, stop)) { startStopTable =>
+        // (step > 0 && start <= stop)
+        withResource(step.greaterThan(zero)) { positiveStep =>
+          withResource(startStopTable.filter(positiveStep)) { upTbl =>
+            withResource(upTbl.getColumn(0).lessOrEqualTo(upTbl.getColumn(1))) { allUp =>
+              require(isAllTrue(allUp), "Illegal sequence boundaries: step > 0 but start > stop")
             }
           }
         }
```
jlowe, line 492: `positiveStep` should be closed earlier, e.g.:
```suggestion
        val upTbl = withResource(step.greaterThan(zero)) { positiveStep =>
          startStopTable.filter(positiveStep)
        }
        withResource(upTbl) { upTbl =>
          withResource(upTbl.getColumn(0).lessOrEqualTo(upTbl.getColumn(1))) { allUp =>
            require(isAllTrue(allUp), "Illegal sequence boundaries: step > 0 but start > stop")
          }
        }
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -407,260 +408,284 @@ class GpuSequenceMeta(
...
-        val size = closeOnExcept(step) { _ =>
-          withResource(difference.abs()) { absDifference =>
-            absDifference.add(one)
+        // (step < 0 && start >= stop)
+        withResource(step.lessThan(zero)) { negativeStep =>
```
jlowe, line 494: `negativeStep` should be closed earlier, similar to previous comment.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -407,260 +408,284 @@ class GpuSequenceMeta(
...
+          withResource(startStopTable.filter(negativeStep)) { downTbl =>
+            withResource(downTbl.getColumn(0).greaterOrEqualTo(downTbl.getColumn(1))) { allDown =>
+              require(isAllTrue(allDown),
+                "Illegal sequence boundaries: step < 0 but start < stop")
+            }
+          }
+        }
+        // (step == 0 && start == stop)
+        withResource(step.equalTo(zero)) { zeroStep =>
```
jlowe, line 503: Close `zeroStep` earlier.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -407,260 +408,284 @@ class GpuSequenceMeta(
...
+    // ```
+    val sizeAsLong = withResource(Scalar.fromLong(1L)) { one =>
+      val diff = withResource(stop.castTo(DType.INT64)) { stopAsLong =>
+        withResource(start.castTo(DType.INT64)) { startAsLong =>
+          stopAsLong.sub(startAsLong)
+        }
+      }
+      // actualSize = 1L + (stop.toLong - start.toLong) / estimatedStep.toLong
+      val actualSize = withResource(diff) { _ =>
+        withResource(step.castTo(DType.INT64)) { stepAsLong =>
```
jlowe, line 528: `stepAsLong` should be closed immediately after computing `quotient`.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -407,260 +408,284 @@ class GpuSequenceMeta(
...
+
+  override def columnarEval(batch: ColumnarBatch): Any = {
+    // 1 Compute the array size for all the rows
+    val (startCol, sizeCol, stepCol, nonNullPredOpt) =
+      withResource(columnarEvalToColumn(start, batch)) { startGpuCol =>
+        withResource(columnarEvalToColumn(stop, batch)) { stopGpuCol =>
+          withResource(stepOpt.map(columnarEvalToColumn(_, batch))) { stepColOpt =>
+            // Filter out the nulls before any computation, since cudf 'sequences' does
+            // not allow nulls.(Pls refer to https://github.com/rapidsai/cudf/issues/10012)
+            // Then no need to do computation for the null rows.
```
jlowe, line 631: I didn't get the impression that we were actually supposed to `filter` the nulls out here.  Per the comment at https://github.com/rapidsai/cudf/issues/10012#issuecomment-1010302972 I was under the impression we're simply doing a logical AND of the start/stop/step validity masks to determine which rows will be valid in the output.  For every row that is not going to be valid, we update the `sizeCol` to be zero for those corresponding entries (can build up a column view with the AND'd validity mask and use `replaceNulls` to do this easily).  Then we execute the sequence with the validity masks stripped off, and then use the AND'd validity mask as the mask of the final resulting column.

That should preclude the need to do any fancy prefix sum scan, gather shenanigans, etc.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -407,260 +408,284 @@ class GpuSequenceMeta(
...
+            val (starts, stops, stepsOption, nonNullOption) = filterOutNullsIfNeeded(
+                startGpuCol.getBase, stopGpuCol.getBase, stepColOpt.map(_.getBase))
+
+            // Get the array size of each valid row, along with the steps
+            withResource(stops) { _ =>
+              closeOnExcept(starts) { _ =>
+                closeOnExcept(nonNullOption) { _ =>
+                  closeOnExcept(stepsOption.getOrElse(defaultSteps(starts, stops))) { steps =>
```
wbo4958, line 639: Do we really want to keep the defaultSteps in the memory?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: Thanks for review.
Personally I like the "filter-out nulls" way which follows the processing in Spark, and the logic seems less cudf specific. 
But I am trying your suggestion locally, still struggling with many failures in tests due to different output between CPU and GPU, will continue it next week.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: It is not a must, I think.
If there is an improvement, I want to handle it in the following PR.

integration_tests/src/main/python/collection_ops_test.py
```diff
@@ -116,90 +116,127 @@ def test_sort_array_lit(data_gen, is_ascending):
...
+#     "Too long sequence: xxxxxxxxxx. Should be <= 2147483632"
+sequence_too_long_length_gens = [
+    IntegerGen(min_val=2147483633, max_val=2147483633, special_cases=[]),
+    LongGen(min_val=2147483635, max_val=2147483635, special_cases=[None])
+]
+
+@pytest.mark.parametrize('stop_gen', sequence_too_long_length_gens, ids=idfn)
+def test_sequence_too_long_sequence(stop_gen):
+    assert_gpu_and_cpu_error(
+        # To avoid OOM, reudce the row number to 2, it is enough to verify this case.
```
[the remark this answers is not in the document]
firestarman: Nice catch, updated

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: No longer needed.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: This is no longer needed

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: This is no longer needed

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: Done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: Eventually, I have made it works. Updated

firestarman:
build

firestarman:
build

firestarman:
Remove databricks due to https://github.com/NVIDIA/spark-rapids/issues/4548

firestarman:
build

firestarman:
build

firestarman:
Add `[databricks]` back which was removed before due to https://github.com/NVIDIA/spark-rapids/issues/4548

integration_tests/src/main/python/collection_ops_test.py
```diff
@@ -116,90 +116,127 @@ def test_sort_array_lit(data_gen, is_ascending):
...
+    (ByteGen(min_val=-10, max_val=20, special_cases=[]),
+        ByteGen(min_val=20, max_val=50, special_cases=[]),
+        ByteGen(min_val=1, max_val=5, special_cases=[])),
+    (ShortGen(min_val=-10, max_val=20, special_cases=[]),
+        ShortGen(min_val=20, max_val=50, special_cases=[]),
+        ShortGen(min_val=1, max_val=5, special_cases=[])),
+    (IntegerGen(min_val=-10, max_val=20, special_cases=[]),
+        IntegerGen(min_val=20, max_val=50, special_cases=[]),
+        IntegerGen(min_val=1, max_val=5, special_cases=[])),
+    (LongGen(min_val=-10, max_val=20, special_cases=[None]),
```
wbo4958, line 137: Looks like your test case can't cover below scenario?

start, stop, step
2,  10, 1
10, 1, -1
2, 2, 0

firestarman: Yes, we do. 
You can refer to the gens in sequence_normal_integral_gens. There are comments to describe each case indivdually.

pxLi:
failed blossom PVC lock. let me retrigger

pxLi:
build

integration_tests/src/main/python/collection_ops_test.py
firestarman: OK, you mean the 3 cased mixed in a single dataset. Added

firestarman:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -407,260 +411,237 @@ class GpuSequenceMeta(
...
+        }
+        val allEq = withResource(equalTbl) { _ =>
+          equalTbl.getColumn(0).equalTo(equalTbl.getColumn(1))
+        }
+        withResource(allEq) { _ =>
+          require(isAllValidTrue(allEq),
+              "Illegal sequence boundaries: step == 0 but start != stop")
+        }
+      }
+    } // end of zero
```
jlowe, line 484: This method is pretty big, and this is just pre-checks to the interesting part of the algorithm.  Recommend factoring this out into a private method to make the main method easier to follow.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -407,260 +411,237 @@ class GpuSequenceMeta(
...
-      ColumnVector.sequence(start.getBase, ret(0), ret(1))
-    }
-  }
 
-  override def doColumnar(start: GpuScalar, stop: GpuColumnVector): ColumnVector = {
-    withResource(calculateSizeAndStep(start.getBase, stop.getBase, stop.dataType())) { ret =>
-      withResource(ColumnVector.fromScalar(start.getBase, stop.getRowCount.toInt)) { startV =>
-        ColumnVector.sequence(startV, ret(0), ret(1))
+    // check size
+    closeOnExcept(sizeAsLong) { _ =>
```
jlowe, line 524: Rather than have a `closeOnExcept(sizeAsLong)` immediately followed by a `withResource(sizeAsLong)`, we can just combine the two blocks together in a single `withResource(sizeAsLong)`.  Is there any advantage to separate the `closeOnExcept` block?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -407,260 +411,237 @@ class GpuSequenceMeta(
...
-  override def doColumnar(
-      start: GpuScalar,
-      stop: GpuColumnVector,
-      step: GpuColumnVector): ColumnVector = {
-    withResource(calculateSize(start.getBase, stop.getBase, step.getBase, stop.getRowCount.toInt,
-        start.dataType)) { size =>
-      withResource(ColumnVector.fromScalar(start.getBase, stop.getRowCount.toInt)) { startV =>
-        ColumnVector.sequence(startV, size, step.getBase)
+  @transient
+  private lazy val defaultSteps: (ColumnView, ColumnView) => ColumnVector =
```
jlowe, line 588: This cannot be a `lazy val` and we cannot in general use `lazy val` on GPU resources.  If it is needed twice, the first time it will create a `ColumnVector` instance and then return it, and the caller will eventually close it.  Then the second time we will _not_ create a new instance, because it's a `lazy val`, and we'll end up returning an instance that corresponds to a closed column vector.  This needs to be a `def` not a `val`, as we need to generate the column each time since it will be closed each time.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/BoolUtils.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import ai.rapids.cudf.{ColumnVector, DType}
+
+object BoolUtils extends Arm {
+
+  /**
+   * Whether all the valid rows in 'col' are true. An empty column will get true.
+   * null rows are skipped.
+   */
+  def isAllValidTrue(col: ColumnVector): Boolean = {
```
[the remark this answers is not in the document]
firestarman: `isAllValidTrue` is not a copy of `isAllTrue`. They have difference.

As its name, `isAllValidTrue` will ignore the null rows, but `isAllTrue` return false if there is any null row.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: This `defaultSteps` is just a function and its type is `(ColumnView, ColumnView) => ColumnVector`, not a GPU resource.

OK, the name is confusable, updated. Sorry for that.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: No, just thought it was clearer to read. Updated

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: Good suggestion, updated.

firestarman:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
jlowe: Apologies for misreading this initially, but I'm still not seeing a compelling reason to use a `lazy val` here instead of a more readable `def` approach.  It seems like this is all to save a single match on `dataType`, otherwise it's just like any other def (but harder to read).

jlowe (approved):
Default steps handling via a lazy val of a function is a bit odd, but I don't think it's a blocker for this to be merged.

firestarman:
> Default steps handling via a lazy val of a function is a bit odd, but I don't think it's a blocker for this to be merged.

The first thought of this is to avoid running the `case match` for each batch. Different element types may have different default steps. I will update this accordingly in the future when supporting date/timestamp.

PR #4523, opened 2022-01-13
Add more integration tests for decimal 128

### LINKED ISSUE #3890: [FEA] Decimal 128 Support: Audit/Cleanup

kuhushukla:
Fixes #3890 
Once hash partitioning is fully in, we can also rework how `all_gen` in join tests and otherwise is used to include decimal128.


kuhushukla:
build

revans2 (approved):

PR #4525, opened 2022-01-13
Properly close data writer in GpuFileFormatWriter

### LINKED ISSUE #1866: [BUG] GpuFileFormatWriter does not close the data writer
GpuFileFormatWriter has code derived from Spark's FileFormatWriter, and [SPARK-30227](https://issues.apache.org/jira/browse/SPARK-30227) added code to ensure the data writer is unconditionally closed when the write completes (either successfully or due to an error).  GpuFileFormatWriter should do the same.

I don't believe our GPU writers rely on the `close()` method being invoked to free resources, but it would be good to fix this for both consistency with the Spark code and to cover the case where someone adds cleanup logic to a data writer `close()` method in the future.

NVnavkumar:
Fixes #1866

Make this more consistent with implementation of FileFormatWriter change implemented in SPARK-30227 by adding finallyBlock with close() call on dataWriter


NVnavkumar:
build

jlowe (approved):

Issue #4360, opened 2021-12-14
[FEA] Add explain api for Spark 2.X


tgravescs:
**Is your feature request related to a problem? Please describe.**
Investigate adding an explain api that would run on a Spark 2.x plan.  

This would be a user callable api because there is no plugin point in catalyst until Spark 3.0.

Task breakdown:

- [x] Core initial implementation using separate module
- [ ] update deploy scripts

Follow ons for another release: https://github.com/NVIDIA/spark-rapids/issues/4522

tgravescs:
still has deploy scripts that need to be done.

PR #4529, opened 2022-01-13
Add support for Spark 2.x Explain Api

tgravescs:
contributes to https://github.com/NVIDIA/spark-rapids/issues/4360

This PR adds a new module that is a copy of a bunch of the code from sql-plugin but only keeps the tagging functionality. So all convert functions and dependencies on cudf were removed. This is to allow running with a single jar (rapids-4-spark-sql-meta_2.11-22.02.0-SNAPSHOT-spark24.jar) against Spark 2.x and call the explain api val output=com.nvidia.spark.rapids.ExplainPlan.explainPotentialGpuPlan(mydf)
to get an idea of how the users application may run on the GPU.  It also gives us an idea if the plugin doesn't support some functions the users application is using.

I unfortunately was unable to find a good way to split the main sql-plugin code into tagging and convert before 22.02 so that is why this is a separate module with a lot of copied code. In the next release we will look at commonizing the code.

Note I named the jar rapids-4-spark-sql-meta_2.11-22.02.0-SNAPSHOT-spark24.jar thinking if we are able to commonize the code perhaps the rapids-4-spark-sql-meta name would apply there.  If people have other ideas please let me know.
As of testing this one jar worked for all the Spark 2.4.x versions I have tried. It even works against CDH which has a 2.4.X jar version.  I only tested one version of CDH though as well.  Note that management said we only have to support Spark 2.4 and newer.
Also note that Spark 2.4 by default builds with scala 2.11 so that is what I use here. They did optionally support 2.12 but I haven't supported there here. scala 2.11 doesn't support -Ywarn-unused:imports so I had to copy some build code into the module pom file.

Since its completely separate code it won't break if someone changes the sql-plugin in an incompatible way. so I don't have it build by default.  It can be built in 2 ways:
- top level run: mvn  clean -Dbuildver=24X package -DskipTests
- or cd spark2-sql-plugin and mvn clena package -DskipTests

I tried to make comments about what 2.x supported in various places to make diffing the code easier. Its also possible I missed some unused imports as intellij doesn't seem to like to import that module properly.

All testing is manual right now. Testing involved running all nds queries with Spark 2.4.X and getting explain output and comparing it to the explain output run on Spark 3.x. Very little differences were there, the diffs were all because Spark 2.x doesn't support the exact same functions the same way in catalyst.  I also manually hacked the integration tests to run the explain to look for it causing any exceptions. This only kind of worked. Lots of tests failed due to it relying on features not there in 2.x. I manually went through a bunch of the output, but want to see if I can make another pass and get rid of things not supported.

Some things Spark 2 doesn't support or are different (I'm sure there are things I'm missing):
1. MapInPandasExec, Acosh, Asinh, Atanh, GetTimestamp, IntegralDivide, MapEntries, TransformValues, NormalizeNaNAndZero, KnownFloatingPointNormalized, DateAddInterval, IntegralDivide, TransformKeys
2.  Adaptive Execution is not supported in 2.x
3. 2.x doesn't support the ANSI stuff
4. SortMergeJoinExec is functionally different with how it optimizes which causes some diffs in plans
5. Aggregate don’t support TypeImperativeAgg - means GpuObjectHashAggregateExecMeta not supported
6. There are just some class changes like missing BaseAggregateExec
7. Datasource V2 is completely different.
8. 3.x switched to Proleptic Gregorian calendar and 2.x is not for file reading (https://issues.apache.org/jira/browse/SPARK-31404)

once this is merged people will have to update the meta information in 2 places until we can get it commonized. I'm sure I will have to make another pass to get things up to date after this PR goes in.  

If all this goes in I will have to modify deploy scripts to release the jar.

I added in a script that attempts to diff all the classes and functions added here.  Its in the scripts directory.  It also has a directory scripts/spark2diffs that have a bunch of diff files used by that script.  I did use the script to upmerge to the latest and it worked in the sense it found a diff, which I manually resolved.  I may try to run the script in the premerge but can do taht in a separate PR.


tgravescs:
build

tgravescs:
I upmerged to the latest and added a script for diffing.

tgravescs:
build

tgravescs:
NOte this still needs upmerged once https://github.com/NVIDIA/spark-rapids/pull/4322 is merged and I also have 1 TODO in the diff script.

tgravescs:
Upmerged to the latest again to pull the 3.x explain changes, also picked up another config change.  This is ready and nothing else to update at the moment.

tgravescs:
build

scripts/spark2diffs/RapidsMeta.diff
```diff
@@ -0,0 +1,394 @@
...
+69c74
+< abstract class RapidsMeta[INPUT <: BASE, BASE](
+---
+> abstract class RapidsMeta[INPUT <: BASE, BASE, OUTPUT <: BASE](
+72c77
+<     val parent: Option[RapidsMeta[_, _]],
+---
+>     val parent: Option[RapidsMeta[_, _, _]],
+88c93
+<   // val childScans: Seq[ScanMeta[_]]
```
jlowe, line 34: There's a mix of deleting stuff 2.x explain does not need vs. commenting it out.  Was there a reason this file has a lot of comments instead of deletions as was done most places elsewhere?

jlowe (commented):
I'm through almost all the diffs but skipped the huge GpuOverrides changes for now, posting what I have so far.  Wondering if this would be less delta change from the baseline if we had the spark2 jar pull in the few cudf classes it needs or even encapsulate/require the cudf jar.  We already do not require CUDA or a GPU on the driver, so even if we pull in cudf classes during planning it shouldn't try to load native code or find CUDA or a GPU during planning.  That might simplify quite a bit since we wouldn't have to hack out anything that might touch a cudf class (even one as benign as `DType`).

spark2-sql-plugin/src/main/scala/com/nvidia/spark/rapids/DateUtils.scala
```diff
@@ -0,0 +1,212 @@
...
+      }
+      val startIndex = format.length - word.length
+      patterns += FormatKeywordToReplace(word, startIndex, startIndex + word.length)
+    }
+    patterns
+  }
+
+  case class TimestampFormatConversionException(reason: String) extends Exception
+
+  // pull in some unit and DateTimeUtils stuff because 2.3.X doesn't have it
```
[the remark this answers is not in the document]
tgravescs: I'll remove, I started 2.3 support until I heard back that we didn't need

scripts/spark2diffs/RapidsMeta.diff
tgravescs: I'll remove these

spark2-sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCastMeta.scala
```diff
@@ -0,0 +1,145 @@
...
+        if (!conf.isCastStringToTimestampEnabled) {
+          willNotWorkOnGpu("the GPU only supports a subset of formats " +
+              "when casting strings to timestamps. Refer to the CAST documentation " +
+              "for more details. To enable this operation on the GPU, set" +
+              s" ${RapidsConf.ENABLE_CAST_STRING_TO_TIMESTAMP} to true.")
+        }
+      case (_: StringType, _: DateType) =>
+        // NOOP for anything prior to 3.2.0
+      case (_: StringType, dt:DecimalType) =>
+        // Spark 2.x  since bug  and we don't know what version of Spark 3 they will be using
```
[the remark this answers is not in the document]
tgravescs: I'll clarify

spark2-sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCastMeta.scala
```diff
@@ -0,0 +1,145 @@
...
+    toTypeOverride: Option[DataType] = None)
+  extends UnaryExprMeta[INPUT](cast, conf, parent, rule) {
+
+  def withToTypeOverride(newToType: DecimalType): CastExprMeta[INPUT] =
+    new CastExprMeta[INPUT](cast, ansiEnabled, conf, parent, rule,
+      doFloatToIntCheck, stringToAnsiDate, Some(newToType))
+
+  val fromType: DataType = cast.child.dataType
+  val toType: DataType = toTypeOverride.getOrElse(cast.dataType)
+  // 2.x doesn't have config set to true
```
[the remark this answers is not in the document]
tgravescs: yes, I'll update

spark2-sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBroadcastJoinMeta.scala
```diff
@@ -0,0 +1,35 @@
...
+    conf: RapidsConf,
+    parent: Option[RapidsMeta[_, _]],
+    rule: DataFromReplacementRule)
+  extends SparkPlanMeta[INPUT](plan, conf, parent, rule) {
+
+  def canBuildSideBeReplaced(buildSide: SparkPlanMeta[_]): Boolean = {
+    buildSide.wrapped match {
+      case reused: ReusedExchangeExec =>
+        reused.child.getClass.getCanonicalName().contains("GpuBroadcastExchangeExec")
+      case f if f.getClass.getCanonicalName().contains("GpuBroadcastExchangeExec") => true
```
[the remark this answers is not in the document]
tgravescs: because we don't have GpuBroadcastExchangeExec in the 2.x source base, these are really for AQE so I'll just remove and comment

tgravescs:
Pulling in just DType might reduce the diffs quite a bit.  Now that I know the diffs I have some other ideas for integrating it with the sql-plugin code for next release.

tgravescs:
build

spark2-sql-plugin/pom.xml
```diff
@@ -0,0 +1,256 @@
...
+        <spark.test.version>${spark.version}</spark.test.version>
+        <spark.version.classifier>spark24</spark.version.classifier>
+        <target.classifier>spark24</target.classifier>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>com.google.flatbuffers</groupId>
+            <artifactId>flatbuffers-java</artifactId>
+            <scope>compile</scope>
+        </dependency>
```
jlowe, line 45: This dependency seems unnecessary, and I couldn't find any usage of it in spark2-sql-plugin sources.

spark2-sql-plugin/pom.xml
```diff
@@ -0,0 +1,256 @@
...
+        <dependency>
+          <groupId>org.apache.hive</groupId>
+          <artifactId>hive-storage-api</artifactId>
+          <exclusions>
+            <exclusion>
+              <groupId>org.slf4j</groupId>
+              <artifactId>slf4j-api</artifactId>
+            </exclusion>
+          </exclusions>
+        </dependency>
```
jlowe, line 86: Should we be using the same version as is used by Spark 2.4 here?  If not, don't we need to shade the ORC/Hive classes to avoid interfering with the versions Spark 2.4 is using?  I'm wondering if we really need to explicitly list ORC/Hive here or can just get the transitive dependency via the spark 2.4 jars.  Since we're not going to need to parse/create ORC footers or perform predicate pushdown as part of an explain-only API, I'll be a bit surprised if we really need ORC jars at all.  We may need to use some Hive classes to examine Hive UDFs to see if they are RAPIDS accelerated.

spark2-sql-plugin/pom.xml
```diff
@@ -0,0 +1,256 @@
...
+                  </goals>
+                  <phase>process-resources</phase>
+                  <configuration>
+                    <target>
+                      <!-- copy NOTICE-binary to NOTICE -->
+                      <copy
+                          todir="${project.build.directory}/classes/META-INF/"
+                          verbose="true">
+                        <fileset dir="${project.basedir}/..">
+                          <include name="NOTICE-binary"/>
```
jlowe, line 160: Depending on how we resolve the ORC/Hive thing, the parent NOTICE-binary may not be appropriate for this artifact.

spark2-sql-plugin/pom.xml
tgravescs: copy and paste and I missed removing

spark2-sql-plugin/pom.xml
tgravescs: yeah we can use the same version I think, 2.4.1-2.4.8 use orc 1.5.5 and 2.4.0 uses orc 1.5.2, I'll check on the hive classes

spark2-sql-plugin/pom.xml
tgravescs: updated to just pull NOTICE since no orc and no ucx

tgravescs:
build

pom.xml
```diff
@@ -1208,6 +1328,7 @@
                             </goals>
                             <configuration>
                                 <classifier>${spark.version.classifier}tests</classifier>
+                                <skipIfEmpty>true</skipIfEmpty>
```
tgravescs, line 1331: Note, I changed this to note generate the tests jar is empty, this will affect other modules as well if you want me to split it out let me know.

tgravescs:
build

tgravescs:
build

spark2-sql-plugin/pom.xml
```diff
@@ -0,0 +1,211 @@
...
+                  <configuration>
+                    <target>
+                      <!-- copy NOTICE, this jar doesn't ship with ORC or UCX -->
+                      <copy
+                          todir="${project.build.directory}/classes/META-INF/"
+                          verbose="true">
+                        <fileset dir="${project.basedir}/..">
+                          <include name="NOTICE"/>
+                        </fileset>
+                        <mapper type="glob" from="*" to="*"/>
```
jlowe, line 117: If we're not going to rename or otherwise change anything about the file as we add it, then we can simply handle this file like we do the LICENSE file up above, eliminating the need to use the antrun plugin for this.

spark2-sql-plugin/pom.xml
```diff
@@ -0,0 +1,211 @@
...
+          </resource>
+        </resources>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                       <!-- transient jar, writing compressed can take several x time -->
+                        <compress>false</compress>
```
jlowe, line 94: This isn't a transient jar, correct?

tgravescs: yep, updated

tgravescs:
build

spark2-sql-plugin/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+              <include>NOTICE</include>
+            </includes>
+          </resource>
+        </resources>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <classifier>${spark.version.classifier}</classifier>
```
jlowe, line 93: Do we want/need the classifier on this jar?

jlowe (approved):

spark2-sql-plugin/pom.xml
tgravescs: Yes, at least my intention was to leave it to be able to clearly see this was for spark 2.4.  Also the plan was if we needed to support 2.3 for this release it would just be a separate jar rather then doing an entire parallel world like setup again.  if you have other thoughts let me know

Issue #4521, opened 2022-01-13
[BUG] Inconsistencies in handling of newline characters and string and line anchors


andygrove:
**Describe the bug**
Although we have implemented transpiler logic to correctly rewrite patterns for cuDF for many cases, we still do not cover all cases, and adding `\r` and `\n` to the set of characters we use for fuzzing causes test failures.

This is quite a broad issue and will likely be split into multiple issues once more analysis has been done.

**Steps/Code to reproduce bug**
Add `\r\n` to the `REGEXP_LIMITED_CHARS_COMMON` constant in `RegularExpressionTranspilerSuite` and run the tests.

**Expected behavior**
Test should pass and we should fall back to CPU for anything that we cannot support correctly.

**Environment details (please complete the following information)**
N/A

**Additional context**
None

PR #4530, opened 2022-01-13
Fix correctness issues in regexp and add `\r` and `\n` to fuzz tests

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/4521

Recent testing has exposed a number of compatibility issues related to end of string/line anchors and newlines. This PR fixes these correctness issues but also is a step backward in terms of what we support on the GPU. I am now working on re-implementing the features that have regressed to follow on from this PR, now that we have more comprehensive fuzz testing that now includes the `\r` and `\n` characters.

Changes in this PR:

- `\Z` and `$` now fall back to CPU, but `\z` is still supported on the GPU
- Fuzz tests now include `\r` and `\n`
- Enabled a test that uses fully random characters that was previously disabled
- Fixes a bug where the transpiler did not detect choice expressions that cuDF is not capable of compiling
- Updated the compatibility guide to remove out of date information and group related items together
- Updated some tests to use `\z` instead of `$`

Follow-on issues:

- https://github.com/NVIDIA/spark-rapids/issues/4532
- https://github.com/NVIDIA/spark-rapids/issues/4533


tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -193,50 +197,19 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
     assert(transpiled === expected)
   }
 
-  test("transpile complex regex 1") {
```
andygrove, line 196: This regex is not supported by cuDF but the transpiler previously did not detect this

docs/compatibility.md
```diff
@@ -465,25 +465,22 @@ CPU when the RAPIDS Accelerator determines that a pattern is either unsupported
 
 Here are some examples of regular expression patterns that are not supported on the GPU and will fall back to the CPU.
 
+- Line anchor `^` is not supported in some contexts, such as when combined with a choice (`^|a`).
+- Line anchor `$`
+- String anchor `\Z`
+- String anchors `\z` is not supported by `regexp_replace`
```
jlowe, line 471:
```suggestion
- String anchor `\z` is not supported by `regexp_replace`
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -441,16 +441,7 @@ class CudfRegexTranspiler(replace: Boolean) {
...
-            RegexRepetition(
-              RegexCharacterClass(negated = false,
-                characters = ListBuffer(RegexChar('\r'))),
-              SimpleQuantifier('?')),
-            RegexRepetition(
-              RegexCharacterClass(negated = false,
-                characters = ListBuffer(RegexChar('\n'))),
-              SimpleQuantifier('?')),
-            RegexChar('$')))
+          throw new RegexUnsupportedException("line anchor $ is not supported")
```
jlowe, line 444: Should this have a link to the issue as is done below for the same case?

andygrove: Yes, and the other case was actually a bug .. an escaped `$` should have been dealt with as a literal `$` and not an anchor.

jlowe (approved):

jlowe:
build

andygrove:
build

jlowe (approved):

jlowe:
build

Issue #4514, opened 2022-01-12
[BUG] test_hash_reduction_pivot_without_nans failed


jlowe:
Nightly Dataproc tests failed:
```
[2022-01-12T20:49:46.955Z] =================================== FAILURES ===================================
[2022-01-12T20:49:46.955Z] _ test_hash_reduction_pivot_without_nans[{'spark.rapids.sql.variableFloatAgg.enabled': 'true', 'spark.rapids.sql.hasNans': 'true', 'spark.rapids.sql.castStringToFloat.enabled': 'true', 'spark.rapids.sql.hashAgg.replaceMode': 'partial'}-[('a', Long), ('b', Integer), ('c', Long)]] _
[2022-01-12T20:49:46.955Z] 
[2022-01-12T20:49:46.955Z] data_gen = [('a', Long), ('b', Integer), ('c', Long)]
[2022-01-12T20:49:46.955Z] conf = {'spark.rapids.sql.castStringToFloat.enabled': 'true', 'spark.rapids.sql.hasNans': 'true', 'spark.rapids.sql.hashAgg.replaceMode': 'partial', 'spark.rapids.sql.variableFloatAgg.enabled': 'true'}
[2022-01-12T20:49:46.955Z] 
[2022-01-12T20:49:46.955Z]     @approximate_float
[2022-01-12T20:49:46.955Z]     @ignore_order(local=True)
[2022-01-12T20:49:46.955Z]     @incompat
[2022-01-12T20:49:46.955Z]     @pytest.mark.parametrize('data_gen', _init_list_no_nans, ids=idfn)
[2022-01-12T20:49:46.955Z]     @pytest.mark.parametrize('conf', get_params(_confs_with_nans, params_markers_for_confs_nans), ids=idfn)
[2022-01-12T20:49:46.955Z]     def test_hash_reduction_pivot_without_nans(data_gen, conf):
[2022-01-12T20:49:46.955Z] >       assert_gpu_and_cpu_are_equal_collect(
[2022-01-12T20:49:46.955Z]             lambda spark: gen_df(spark, data_gen, length=100)
[2022-01-12T20:49:46.955Z]                 .groupby()
[2022-01-12T20:49:46.955Z]                 .pivot('b')
[2022-01-12T20:49:46.955Z]                 .agg(f.sum('c')),
[2022-01-12T20:49:46.955Z]             conf=conf)
[2022-01-12T20:49:46.955Z] 
[...]

PR #4534, opened 2022-01-14
Temporarily disable integration test - test_hash_reduction_pivot_without_nans

nartal1:
This PR is to temporarily disable test_hash_reduction_pivot_without_nans test until we have a proper fix for the issue 
https://github.com/NVIDIA/spark-rapids/issues/4514.
I am not able to repro the issue in YARN cluster so it might take sometime to reproduce and fix the bug.
This would help the CI pipelines to be green.

### DIFF
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -533,20 +533,20 @@ def test_hash_pivot_reduction_nan_fallback(data_gen):
         "PivotFirst",
         conf=_nans_float_conf)
 
-# https://github.com/NVIDIA/spark-rapids/issues/4514
-# Disabling below test temporarily until we have a fix. Disabling fixes the CI pipeline failures.
-#@approximate_float
-#ignore_order(local=True)
-#@incompat
-#@pytest.mark.parametrize('data_gen', _init_list_no_nans, ids=idfn)
-#@pytest.mark.parametrize('conf', get_params(_confs_with_nans, params_markers_for_confs_nans), ids=idfn)
-#def test_hash_reduction_pivot_without_nans(data_gen, conf):
-#    assert_gpu_and_cpu_are_equal_collect(
-#        lambda spark: gen_df(spark, data_gen, length=100)
-#            .groupby()
-#            .pivot('b')
-#            .agg(f.sum('c')),
-#        conf=conf)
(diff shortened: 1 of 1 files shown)


integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -533,18 +533,20 @@ def fetch_pivot_values(spark):
...
-@pytest.mark.parametrize('data_gen', _init_list_no_nans, ids=idfn)
-@pytest.mark.parametrize('conf', get_params(_confs_with_nans, params_markers_for_confs_nans), ids=idfn)
-def test_hash_reduction_pivot_without_nans(data_gen, conf):
-    assert_gpu_and_cpu_are_equal_collect(
-        lambda spark: gen_df(spark, data_gen, length=100)
-            .groupby()
-            .pivot('b')
-            .agg(f.sum('c')),
-        conf=conf)
+# https://github.com/NVIDIA/spark-rapids/issues/4514
```
andygrove, line 536: It would be better to skip the test rather than comment it out:

```python
@pytest.mark.skip(reason='https://github.com/NVIDIA/spark-rapids/issues/4514')
```

jlowe: `xfail` seems appropriate here, as this is code we fully expect to pass but is failing due to known issues.  `skip` should be used for tests that are never expected to pass if attempted, even after bugfixes (e.g.: tests for a feature that a particular Spark version does not support, etc.)

abellina: It doesn't seem to be failing in github CI, so we probably want to skip?

jlowe: xfail is what we want.  xfail is fine if the test actually passes (it reports it as an XPASS result).

revans2: There is a config for that I wish we had done a good enough job with things that it would fail. But that is not how it is right now :(

nartal1:
Thanks for the review. I have changed the test to xfail. PTAL.

nartal1:
build

abellina (approved):

jlowe (approved):

andygrove (approved):

PR #4535, opened 2022-01-14
Revert "Disable orc write by default because of https://issues.apache.org/jira/browse/ORC-1075 (#4471)"

wbo4958:
Depends on https://github.com/rapidsai/cudf/pull/10041 which has added the column statistic in RowIndex when writing ORC files by CUDF. I tested it locally, it can work. We need to enable GPU accelerated ORC write by default.


wbo4958:
build

tgravescs:
converted to draft since cudf pr still draft

jlowe (dismissed):
@wbo4958 please elaborate a bit in the PR description why we're able to revert the disable commit because of the cudf changes, as this would help anyone that needs to investigate this change in the future.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuOrcFileFormat.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2022, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2021, NVIDIA CORPORATION.
```
tgravescs, line 2: should leave 2022 in there I think

wbo4958:
build

jlowe (approved):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuOrcFileFormat.scala
wbo4958: Done, Thx

wbo4958:
build

PR #4545, opened 2022-01-17
Fix CVE-2021-22569

pxLi:

pre-merge CI is blocked by vulnerability scan https://github.com/advisories/GHSA-wrvw-hg22-4m67

try update protobuf-java version to suggested version

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -848,7 +848,7 @@
         <shim.module.name>${spark.version.classifier}</shim.module.name>
         <slf4j.version>1.7.30</slf4j.version>
         <hive.storage.api.version>2.7.1</hive.storage.api.version>
-        <protobuf.java.version>2.5.0</protobuf.java.version>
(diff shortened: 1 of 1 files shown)


pxLi:
build

pxLi:
Spark runtime use 2.5.0 version, so we use the specific version till now.

NVIDIA blackduck scan does not allow vulnerability of direct dependencies, so I bumped the version here to unblock the pre-merge CI if test could pass.

The real fix will be https://github.com/NVIDIA/spark-rapids/pull/4408 which will remove several direct dependencies including protobuf-java

zhanga5 (approved):

pxLi:
> so I'm not sure I like changing this right before the 22.02 release. #4408 is scheduled for 22.04. It builds and tests run so maybe its ok but we going from a 2.x to a 3.x and I think we should look at compatibility or what exactly we are using this for. #4551

yeah, this should just be a temp fix to unblock CI. We should go for some mature fix

Issue #4507, opened 2022-01-12
[DOC] udf-examples.md links broken on nvidia.github.io


mattf:
## Report incorrect documentation

**Location of incorrect documentation**

https://nvidia.github.io/spark-rapids/docs/additional-functionality/rapids-udfs.html

broken links for -

- RapidsUDF
- udf-examples
- URLDecode x3
- URLEncode x3
- CosineSimilarity
- StringWordCount

**Describe the problems or issues found in the documentation**

all these links use `../..` relative paths for their destination, which works when browsing the source in the github.com/NVIDIA/spark-rapids repository but does not work when browsing the documentation published to nvidia.github.io/spark-rapids

**Suggested fix for documentation**

i expect links to work in both github.com/NVIDIA/spark-rapids as well as on nvidia.github.io/spark-rapids

viadea:
@jlowe @sameerz 
To fix this issue, I plan to use the branch link for the source code in the dev branch(say branch-22.02). 
For example:
```
[udf-examples](https://github.com/NVIDIA/spark-rapids/tree/branch-22.02/udf-examples)
```
How do you think?

If you agree, it means, each time when creating a new dev branch, we just change the version at that time.

How do you think about this solution?

jlowe:
> I plan to use the branch link for the source code in the dev branch(say branch-22.02).

There are two significant drawbacks to that approach:
- We would need to remember to update the link every time we create a new release
- For those browsing the docs via github pages, the link would take them out of github pages rather than to the corresponding documentation on github pages.

I hope we can find a better solution that links to github if you're browsing on github and github pages if you're browsing on github pages and does not require an update every release.

viadea:
closed.

PR #4554, opened 2022-01-18
Update gh-pages UDF doc to correct links for 21.12 version[skip ci]

viadea:
Update gh-pages UDF doc to correct links to address https://github.com/NVIDIA/spark-rapids/issues/4507.

Note: we need to update this page once 22.02 is about to be released.


docs/additional-functionality/rapids-udfs.md
```diff
@@ -52,7 +52,7 @@ Other forms of Spark UDFs are not supported, such as:
 
 For supported UDFs, the RAPIDS Accelerator will detect a GPU implementation
 if the UDF class implements the
-[RapidsUDF](../../sql-plugin/src/main/java/com/nvidia/spark/RapidsUDF.java)
+[RapidsUDF](https://github.com/NVIDIA/spark-rapids/blob/branch-21.12/sql-plugin/src/main/java/com/nvidia/spark/RapidsUDF.java)
```
jlowe, line 55: If this is just for github pages, shouldn't these links all point to the gh-pages version, i.e.: under https://nvidia.github.io/spark-rapids/, rather than the github markdown from the source repository?

jlowe (commented):
Commit needs signoff per [the contributor guidelines.](https://github.com/NVIDIA/spark-rapids/blob/branch-22.02/CONTRIBUTING.md#sign-your-work)

pxLi:
going to close and reopen this PR to retrigger signoff check

docs/additional-functionality/rapids-udfs.md
viadea: @jlowe gh-pages branch does not have source code so i have to link it to the github source code for that specific released branch.

jlowe: Ah right, these are pointing to source.  If we're going to point to source from the Github pages, then this should point to the `main` branch since gh-pages should reflect what's been released (i.e.: on `main`).  This also has the benefit of not needing to be updated in most cases.  It only needs to be updated when the link changes or is removed, and we will get a merge conflict on the gh-pages branch if that occurs.

viadea: Actually I am fine with main. Actually that is easier.
The only concern is that, before the official release say 22.02, the main branch merge may be done before that.
If we are fine with that, i can update all links to point to main branch.

viadea: @jlowe  how about now? i changed everything to main branch.

jlowe (approved):

Issue #4503, opened 2022-01-11
[BUG] regexp_replace with back references produces incorrect results on GPU


andygrove:
**Describe the bug**

Spark supports back references in `regexp_replace` but this functionality does not work correctly on the GPU.

## Example

```scala
val df = List("a1","b22","333").toDF("str").repartition(2)
val df2 = df.withColumn("repBackRef",regexp_replace('str,"(\\d+)$",";$1"))
df2.show
```

## Expected Results

```
+---+----------+
|str|repBackRef|
+---+----------+
| a1|       a;1|
|b22|      b;22|
|333|      ;333|
+---+----------+
```

## Actual Results

```
+---+----------+
|str|repBackRef|
+---+----------+
| a1|      a;$1|
|b22|      b;$1|
|333|       ;$1|
+---+----------+
```

**Steps/Code to reproduce bug**
See above.

**Expected behavior**
We should fall back to CPU, or produce the correct results.

**Environment details (please complete the following information)**
N/A

**Additional context**
None

PR #4556, opened 2022-01-18
regexp_replace with back-references should fall back to CPU

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/4503 and https://github.com/NVIDIA/spark-rapids/issues/4559

Follow-on issue to support back-references on the GPU: https://github.com/NVIDIA/spark-rapids/issues/4557

### DIFF
--- a/sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
+++ b/sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
@@ -15,12 +15,10 @@
  */
 package com.nvidia.spark.rapids.shims.v2
 
-import java.util.regex.Pattern
-
 import com.nvidia.spark.rapids.{CudfRegexTranspiler, DataFromReplacementRule, GpuExpression, GpuOverrides, RapidsConf, RapidsMeta, RegexUnsupportedException, TernaryExprMeta}
 
 import org.apache.spark.sql.catalyst.expressions.{Expression, Literal, RegExpReplace}
-import org.apache.spark.sql.rapids.{GpuRegExpReplace, GpuStringReplace}
+import org.apache.spark.sql.rapids.{GpuRegExpReplace, GpuRegExpUtils, GpuStringReplace}
 import org.apache.spark.sql.types.DataTypes
 import org.apache.spark.unsafe.types.UTF8String
 
@@ -32,6 +30,7 @@ class GpuRegExpReplaceMeta(
   extends TernaryExprMeta[RegExpReplace](expr, conf, parent, rule) {
 
   private var pattern: Option[String] = None
+  private var replacement: Option[String] = None
 
   override def tagExprForGpu(): Unit = {
     expr.regexp match {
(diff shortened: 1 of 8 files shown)


revans2 (approved):
I am not thrilled with using a regular expression to find the back references, but I could not figure out a better way to do it either.

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
```diff
@@ -48,6 +50,14 @@ class GpuRegExpReplaceMeta(
       case _ =>
         willNotWorkOnGpu(s"only non-null literal strings are supported on GPU")
     }
+
+    expr.rep match {
+      case Literal(s: UTF8String, DataTypes.StringType) if s != null =>
+        val backrefPattern = Pattern.compile("\\$[0-9]")
```
jlowe, line 56: Note that this can create false-positives.  A Spark regex replace string of `"\\$1"` isn't asking for a group but this pattern matcher will think it is.

andygrove: I was working on a fix for this but that exposed [another issue](https://github.com/NVIDIA/spark-rapids/issues/4559) that I am going to have to fix first, so I will mark this PR as a draft for now.

andygrove:
@revans2 @jlowe This is ready for another review when you have time

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -744,6 +744,48 @@ case class GpuLike(left: Expression, right: Expression, escapeChar: Char)
...
+   * replacement string before passing to cuDF.
+   */
+  def unescapeReplaceString(s: String): String = {
+    val b = new StringBuilder
+    var i = 0
+    while (i < s.length) {
+      if (s.charAt(i) == '\\' && i+1 < s.length) {
+        i += 1
+      }
+      b.append(s.charAt(i))
```
jlowe, line 781: What about escape characters that evaluate to other characters besides the character after the backslash, e.g.: \t, \n, etc.?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -863,15 +906,16 @@ case class GpuRegExpReplace(
...
+      cudfReplacementString: String) = {
+    this(srcExpr, searchExpr, GpuLiteral("", StringType), cudfRegexPattern, cudfReplacementString)
   }
 
   override def doColumnar(
       strExpr: GpuColumnVector,
       searchExpr: GpuScalar,
       replaceExpr: GpuScalar): ColumnVector = {
-    strExpr.getBase.replaceRegex(cudfRegexPattern, replaceExpr.getBase)
+    strExpr.getBase.replaceRegex(cudfRegexPattern, Scalar.fromString(cudfReplacementString))
```
jlowe, line 918: The scalar is leaked here.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
jlowe: Ah, nevermind.  These types of escapes should not be appearing in the string once it arrives to this point, so I think we're fine here.

jlowe (approved):

jlowe:
build

Issue #4022, opened 2021-11-03
[BUG] Ensure shims.v2.ParquetCachedBatchSerializer and similar classes are at most package-private  


gerashegalov:
**Describe the bug**
This is a follow-up issue to https://github.com/NVIDIA/spark-rapids/pull/4021#issuecomment-960142547 . We have a few classes that exist merely to be proxied by the publicly documented facade classes. Consider reflecting this in the access level of those classes or their constructors to make sure that the user cannot reference / configure them directly in Apache Spark.

PR #4560, opened 2022-01-19
Make shims.v2.ParquetCachedBatchSerializer as protected

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/4022. 
Making com.nvidia.spark.rapids.shims.v2.ParquetCachedBatchSerializer as protected since that class is not configurable by SparkConf/RapidsConf.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/ExclusiveModeGpuDiscoveryPlugin.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/ExclusiveModeGpuDiscoveryPlugin.scala
@@ -35,7 +35,7 @@ import org.apache.spark.resource.{ResourceInformation, ResourceRequest}
  *  This plugin can be activated in spark with the configuration:
  *  `--conf spark.resources.discoveryPlugin=com.nvidia.spark.ExclusiveModeGpuDiscoveryPlugin`
  */
-protected class ExclusiveModeGpuDiscoveryPlugin extends ResourceDiscoveryPlugin with Proxy {
+class ExclusiveModeGpuDiscoveryPlugin extends ResourceDiscoveryPlugin with Proxy {
   override def discoverResource(
     request: ResourceRequest,
     sparkConf: SparkConf
--- a/sql-plugin/src/main/scala/com/nvidia/spark/ExclusiveModeGpuDiscoveryPlugin.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/ExclusiveModeGpuDiscoveryPlugin.scala
@@ -35,7 +35,7 @@ import org.apache.spark.resource.{ResourceInformation, ResourceRequest}
  *  This plugin can be activated in spark with the configuration:
  *  `--conf spark.resources.discoveryPlugin=com.nvidia.spark.ExclusiveModeGpuDiscoveryPlugin`
  */
-class ExclusiveModeGpuDiscoveryPlugin extends ResourceDiscoveryPlugin with Proxy {
+protected class ExclusiveModeGpuDiscoveryPlugin extends ResourceDiscoveryPlugin with Proxy {
   override def discoverResource(
     request: ResourceRequest,
     sparkConf: SparkConf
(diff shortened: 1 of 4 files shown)


nartal1:
build

sql-plugin/src/main/scala/com/nvidia/spark/ExclusiveModeGpuDiscoveryPlugin.scala
```diff
@@ -35,7 +35,7 @@ import org.apache.spark.resource.{ResourceInformation, ResourceRequest}
  *  This plugin can be activated in spark with the configuration:
  *  `--conf spark.resources.discoveryPlugin=com.nvidia.spark.ExclusiveModeGpuDiscoveryPlugin`
  */
-class ExclusiveModeGpuDiscoveryPlugin extends ResourceDiscoveryPlugin with Proxy {
+protected class ExclusiveModeGpuDiscoveryPlugin extends ResourceDiscoveryPlugin with Proxy {
```
jlowe, line 38: This is the public class that the user needs to use [per the documentation](https://github.com/NVIDIA/spark-rapids/blob/branch-22.02/docs/get-started/getting-started-on-prem.md#yarn-without-isolation).  Any class directly under `com.nvidia.spark`, as opposed to under `com.nvidia.sparki.rapids`, should be public.

jlowe (commented):
I'm not clear on the goal of this PR.  IIUC, marking these as Scala protected is only going to enforce something at compile time, not runtime.  I don't think it is going to "hide" them in the resulting jar, and they would still be directly accessible if this were a normal jar.   Therefore I don't think this accomplishes anything if we're worried about users inadvertently or directly accessing these classes in the jar _if_ the dist jar was a normal jar.

But the dist jar _is not a normal jar_.  Almost all classes are "hidden" in the "parallel-world" classloader setup.  Users that attempt to compile against this jar will not be able to find classes in their normal package locations, and users asking for these classes on the spark command-line will similarly not be able to locate them at their expected locations.  Only the truly public classes, of which there are very few, are placed in their expected locations in the jar and therefore are directly accessible.

gerashegalov:
The goal of  #4022  is to eliminate a source of confusion for developers of spark-rapids internals, inspired by some discussions with @razajafri . The current PR description is not quite accurate. If we go ahead with this, the PR should mark the ancestor classes / traits or otherwise internal classes that are not *settable* via config differently than public classes. I think the point @jlowe's point about runtime implications of the parallel worlds jar is valid but it is orthogonal to the issue.

sql-plugin/src/main/scala/com/nvidia/spark/ExclusiveModeGpuDiscoveryPlugin.scala
razajafri: Why is this change a part of this PR?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/RapidsShuffleInternalManagerBase.scala
```diff
@@ -444,7 +444,7 @@ trait VisibleShuffleManager {
  * @param conf
  * @param isDriver
  */
-abstract class ProxyRapidsShuffleInternalManagerBase(
+protected abstract class ProxyRapidsShuffleInternalManagerBase(
```
razajafri, line 447: Same question?

nartal1: Removed it. I had misunderstood the bug.

sql-plugin/src/main/scala/com/nvidia/spark/ExclusiveModeGpuDiscoveryPlugin.scala
nartal1: Thanks @jlowe and @razajafri. I have kept it as public now. I misunderstood the bug and made this change.

nartal1:
@jlowe, @razajafri and @gerashegalov .  I didn't understand the bug properly earlier and made those changes. I have updated the PR description. Please take another look.

razajafri (dismissed):

jlowe (approved):

jlowe:
build

razajafri:
@nartal1 follow these steps to never having to worry about updating the copyrights manually 

https://github.com/NVIDIA/spark-rapids/blob/branch-22.02/CONTRIBUTING.md#pre-commit-hooks

gerashegalov (approved):

Issue #4454, opened 2022-01-04
[BUG] Shade warnings when building the tools artifact


jlowe:
**Describe the bug**
Building produces the following warning:
```
[WARNING] rapids-4-spark-tools_2.12-22.02.0-SNAPSHOT.jar, scala-library-2.12.15.jar define 1 overlapping resource: 
[WARNING]   - META-INF/MANIFEST.MF
[WARNING] maven-shade-plugin has detected that some class files are
[WARNING] present in two or more JARs. When this happens, only one
[WARNING] single version of the class is copied to the uber jar.
[WARNING] Usually this is not harmful and you can skip these warnings,
[WARNING] otherwise try to manually exclude artifacts based on
[WARNING] mvn dependency:tree -Ddetail=true and the above output.
[WARNING] See http://maven.apache.org/plugins/maven-shade-plugin/
```

**Steps/Code to reproduce bug**
`mvn clean package -DskipTests`

**Expected behavior**
Build should not emit shade warnings.

PR #4566, opened 2022-01-19
remove scala-library from combined tools jar

tgravescs:
Fixes https://github.com/NVIDIA/spark-rapids/issues/4454

build was reporting a shade warnings when building the tools artifact

```
[WARNING] rapids-4-spark-tools_2.12-22.02.0-SNAPSHOT.jar, scala-library-2.12.15.jar define 1 overlapping resource: 
[WARNING]   - META-INF/MANIFEST.MF
[WARNING] maven-shade-plugin has detected that some class files are
```

I did some testing and it doesn't look like we need to include the scala-library in the combined tools jar.  For the tools the user has to include the jars directory for Spark so we can pick up the scala library from there. I tested against a bunch of different spark versions from 3.0.1 and up and it ran fine.

### DIFF
--- a/tools/pom.xml
+++ b/tools/pom.xml
@@ -99,17 +99,10 @@
                         <configuration>
                             <artifactSet>
                                 <includes>
-                                    <include>org.scala-lang:scala-library</include>
                                     <include>org.rogach:scallop_${scala.binary.version}</include>
                                 </includes>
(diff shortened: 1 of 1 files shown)


tgravescs:
build

jlowe (approved):

Issue #4567, opened 2022-01-19
[BUG] Profile tool hangs in compare mode


tgravescs:
**Describe the bug**
Running the profile tool in compare mode against all the test profile files hangs:

`java -cp tools/target/spark311/rapids-4-spark-tools_2.12-22.02.0-SNAPSHOT.jar:$SPARK_HOME/jars/* com.nvidia.spark.rapids.tool.profiling.ProfileMain --compare ./tools/src/test/resources/spark-events-profiling/*`

Seems like its in an infinite loop printing out "empty sql id arr".

PR #4569, opened 2022-01-19
Fix infinite loop with Profiling tool compare mode and app with no sql ids

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/4567

The profiling tool when run in compare mode and an application doesn't have sqlIds can end up in an infinite loop.
Here we fix that just by removing that appId so we continue on to the next one.


tgravescs:
build

nartal1 (approved):

Issue #4315, opened 2021-12-07
[BUG] test_hash_reduction_decimal_overflow_sum[30] failed OOM in integration tests


pxLi:
**Describe the bug**
related to #4272 

FAILED ../../src/main/python/hash_aggregate_test.py::test_hash_reduction_decimal_overflow_sum[30]

```
java.lang.OutOfMemoryError: Could not allocate native memory: std::bad_alloc: out_of_memory: RMM failure at:/home/jenkins/agent/workspace/jenkins-cudf_nightly-dev-github-556-cuda11/cpp/build/_deps/rmm-src/include/rmm/mr/device/arena_memory_resource.hpp:157: Maximum pool size exceeded
[2021-12-07T06:04:57.784Z] 	at ai.rapids.cudf.ColumnView.binaryOpVS(Native Method)
[2021-12-07T06:04:57.784Z] 	at ai.rapids.cudf.ColumnView.binaryOp(ColumnView.java:1205)
[2021-12-07T06:04:57.784Z] 	at ai.rapids.cudf.ColumnView.binaryOp(ColumnView.java:1195)
[2021-12-07T06:04:57.784Z] 	at ai.rapids.cudf.BinaryOperable.div(BinaryOperable.java:171)
[2021-12-07T06:04:57.784Z] 	at org.apache.spark.sql.rapids.GpuDecimalSumHighDigits.$anonfun$columnarEval$10(AggregateFunctions.scala:720)
[2021-12-07T06:04:57.784Z] 	at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
[2021-12-07T06:04:57.784Z] 	at com.nvidia.spark.rapids.Arm.withResource$(Arm.scala:26)
[2021-12-07T06:04:57.784Z] 	at org.apache.spark.sql.rapids.GpuDecimalSumHighDigits.withResource(AggregateFunctions.scala:690)
[2021-12-07T06:04:57.784Z] 	at org.apache.spark.sql.rapids.GpuDecimalSumHighDigits.$anonfun$columnarEval$9(AggregateFunctions.scala:719)
[2021-12-07T06:04:57.784Z] 	at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
[2021-12-07T06:04:57.784Z] 	at com.nvidia.spark.rapids.Arm.withResource$(Arm.scala:26)
[...]

PR #4570, opened 2022-01-19
A few quick fixes to try to reduce max memory usage in the tests

revans2:
This helps with #4315 and #4542

I didn't take the time to measure the actual maximum memory usage by these tests. I plan on making some changes to do that soon.  For now this just does the most obvious things. It disables the test for #4315 as it is not critical. It also splits up the test for #4542 into smaller tests to try to reduce the total memory usage.


revans2:
build

jlowe (approved):

PR #4585, opened 2022-01-20
Spark2 module upmerge, deploy script, and updates for Jenkins

tgravescs:
Upmerge spark2 module to latest, pulls changes from:
https://github.com/NVIDIA/spark-rapids/commit/0d236d887e37c66593ebf67525ec5894753295b6

Update the deploy script to deploy the new spark2 jar.  Unfortunately I haven't tested them because the test deploy build is failing for something different, working with @NvTimLiu 

Added some docs to the script for developers to give them an idea on how to deal with it.

Update the diff script to error out properly only when actual diff in diff files.
Changed the aggregate diff to not use git so we could more easily run from jenkins build.  

fixes https://github.com/NVIDIA/spark-rapids/issues/4360

### DIFF
--- a/scripts/rundiffspark2.sh
+++ b/scripts/rundiffspark2.sh
@@ -25,7 +25,7 @@
 # just using interface, and we don't really expect them to use it on 2.x so just skip diffing
 #  ../spark2-sql-plugin/src/main/java/com/nvidia/spark/RapidsUDF.java
 
-# If this script fails then a developer should do soemthing like:
+# If this script fails then a developer should do something like:
 #  1. Look at each file with a diff output from the script
 #  2. Look at the commits for that file in sql-plugin module and see what changed dealing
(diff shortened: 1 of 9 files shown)


tgravescs:
build

scripts/rundiffspark2.sh
```diff
@@ -25,44 +25,59 @@
 # just using interface, and we don't really expect them to use it on 2.x so just skip diffing
 #  ../spark2-sql-plugin/src/main/java/com/nvidia/spark/RapidsUDF.java
 
+# If this script fails then a developer should do soemthing like:
```
jlowe, line 28:
```suggestion
# If this script fails then a developer should do something like:
```

tgravescs:
build

jlowe (approved):

Issue #4539, opened 2022-01-14
[FEA] semaphore optimization in shuffled hash join


abellina:
This is an optimization we identified while looking into q23a/b, q24a/b and several other TPCDS queries. Currently, the shuffled hash join code follows these steps in order to materialize build and stream side when it first starts working a partition:

1. It fetches the build side, concatenates it on the host, grabs the semaphore and puts it on the GPU 
2. It fetches the first stream side batch, concatenates it on the host, and puts it on the GPU (while continuing to hold on to the semaphore)
3. performs the join

The observation in traces is is that grabbing the semaphore in (1) means we are holding onto the semaphore while the IO parts of the first stream batch are taking place in (2). This is CPU work and we should be able to do this outside of the semaphore, moreover it can be very much non-trivial amounts of time. In q23a/b there are several seconds spent in this mode.

A proof of concept was coded that does this instead:

1. It fetches the build side and concatenates it on the host.
2. The stream side is allowed to fetch that first batch, concatenates it on the host, acquires the semaphore, and puts the stream batch on the GPU.
3. The build side is allowed to go to the GPU.
4. performs the join

This has lead to savings of ~10s in q23a/b, q24a/b and others. Overall we see close to 2 minutes worth of time spent.

The complicated part about this change is that It adds more host memory pressure, since many tasks would be doing their IO and keeping a host-side batch. One approach to keep this in check is using the batch size goal as a limit, so if we reach the limit we'd start copying batch-sized batches to the GPU and grabbing the semaphore. I am still working through this part of it.

PR #4588, opened 2022-01-20
Optimize semaphore acquisition in GpuShuffledHashJoinExec

abellina:
Closes https://github.com/NVIDIA/spark-rapids/issues/4539.

This PR adds an optimization for the shuffled hash join where the semaphore is not taken after the build side is fetched to the host (since we attempt to keep the build side on the host while we load the first batch from the stream side).

After the first stream batch is loaded, the semaphore has been acquired. We then proceed to bring the build batch to the GPU.

This code falls back to the old behavior in cases where batches are not serialized (our input is not a shuffle) and when the streams are empty. 

It is in draft mode because I need to run more testing in NDS. I don't think the case where I need to loop over the build batch multiple times has been hit with my runs so far (other than unit tests), so I need to figure out a way to trigger that case.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -129,22 +133,22 @@ case class GpuShuffledHashJoinExec(
...
       (streamIter, buildIter) => {
-        val stIt = new CollectTimeIterator("shuffled join stream", streamIter, streamTime)
-        val startTime = System.nanoTime()
-
-        withResource(ConcatAndConsumeAll.getSingleBatchWithVerification(buildIter,
-          localBuildOutput)) { builtBatch =>
+        val (builtBatch, maybeBufferedStreamIter) =
+          GpuShuffledHashJoinExec.getBuiltBatchAndStreamIter(
+            targetSize,
+            buildPlan.output,
```
revans2, line 142: Go back to `localBuildOutput`. It is there to avoid needing to serialize the entire buildPlan when shipping the zipPartitions function.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTransitionOverrides.scala
```diff
@@ -237,6 +237,27 @@ class GpuTransitionOverrides extends Rule[SparkPlan] {
       p.withNewChildren(p.children.map(optimizeCoalesce))
   }
 
+  /**
+   * Removes `GpuCoalesceBatches(GpuShuffleCoalesceExec(build side))` for the build side
+   * for the shuffled hash join. The coalesce logic has been moved to the
+   * `GpuShuffleCoalesceExec` class, and is handled differently to prevent holding onto the
```
revans2, line 243: This is fine for now, but I really would prefer to have us just build the plan the right way from the beginning.  Perhaps we can have a flag in GpuExec that says for this input I know how to handle raw shuffle data so then the rules that insert the host side coalesce can deal with it there.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +159,169 @@ case class GpuShuffledHashJoinExec(
...
+   * @param buildTime build GPU metric
+   * @return a pair of `ColumnarBatch` and streamed iterator that can be
+   *         used for the join
+   */
+  def getBuiltBatchAndStreamIter(targetBatchSize: Long,
+                                 buildOutput: Seq[Attribute],
+                                 buildIter: Iterator[ColumnarBatch],
+                                 streamIter: Iterator[ColumnarBatch],
+                                 semWait: GpuMetric,
+                                 buildTime: GpuMetric): (ColumnarBatch, Iterator[ColumnarBatch]) = {
```
revans2, line 217: nit: if it is simple it would be great to add some more metrics here for the operations that we combined into this.  We no longer know how long it took to concat on the GPU. At a minimum that should be added to the opTime metric.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +159,169 @@ case class GpuShuffledHashJoinExec(
...
+    closeOnExcept(bufferedBuildIterator) { _ =>
+      val startTime = System.nanoTime()
+      // find if the build side is non-empty, and if the first batch is
+      // a serialized batch. If neither condition is met, we fallback to the
+      // `getSingleBatchWithVerification` method.
+      val firstBatchIsSerialized = {
+        if (!buildIter.hasNext) {
+          false
+        } else {
+          bufferedBuildIterator = new CloseableBufferedIterator(buildIter.buffered)
```
revans2, line 228: How much is buffered when we do this?

revans2:
So just to be clear this optimization can never work with UCX right?

abellina:
> So just to be clear this optimization can never work with UCX right?

Great point. I have missed a case where we need to ignore the config when UCX is turned on and we have compression on. If we don't do this, the fallback case right now doesn't handle compressed batches from the shuffle.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +159,169 @@ case class GpuShuffledHashJoinExec(
...
+   * @param targetBatchSize configured target batch size
+   * @param buildOutput output attributes of the build plan
+   * @param buildIter build iterator
+   * @param streamIter stream iterator
+   * @param semWait wait GPU metric
+   * @param buildTime build GPU metric
+   * @return a pair of `ColumnarBatch` and streamed iterator that can be
+   *         used for the join
+   */
+  def getBuiltBatchAndStreamIter(targetBatchSize: Long,
```
revans2, line 212: We cannot as any of the iterators if they `haveNext` unless we hold the GPU semaphore or we know 100% that they will not put something on the GPU.  I the general case we cannot know that. So we cannot call it on the stream iterator and we cannot call it on the build iterator to discover what type of iterator it is.

revans2 (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
revans2: I might not understand how/when the semaphore is grabbed in these cases. If I am wrong, which I hope I am, then we need some more comments explaining how this all works.

revans2: Ya I rethought it. Can we have some comments around the hasNext calls saying that they will grab the semaphore if they are going to put something on the GPU and we don't need to worry about it in those cases.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +159,169 @@ case class GpuShuffledHashJoinExec(
...
+      }
+
+      if (!firstBatchIsSerialized || !streamIter.hasNext) {
+        // fallback if we failed to find serialized build batches, or if the
+        // stream iterator is empty
+        val builtBatch =
+        ConcatAndConsumeAll.getSingleBatchWithVerification(
+          Option(bufferedBuildIterator).getOrElse(buildIter), buildOutput)
+        val delta = System.nanoTime() - startTime
+        buildTime += delta
```
revans2, line 248: So does the build time in this case also include the time to check to see if the streamIter is empty? If so I am fine with it, we just might want to document it because it could be really confusing.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
abellina: Yeah, if hasNext is called in the buffered iterator a single batch is getting loaded. That batch may be on the GPU. The expectation is that the iterator that gave us that batch via the buffered iterator is getting on the GPU, if not that's a bug. I'll put a comment around this to make this clear.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +159,169 @@ case class GpuShuffledHashJoinExec(
...
+        (builtBatch, streamIter)
+      } else {
+        val dataTypes = buildOutput.map(_.dataType).toArray
+        val bufferedStreamIter = new CloseableBufferedIterator(streamIter.buffered)
+        val hostBuffs = new ArrayBuffer[SerializedTableColumn]()
+        var buildBatch: ColumnarBatch = null
+        closeOnExcept(bufferedStreamIter) { _ =>
```
revans2, line 255: nit: why make this two lines of code?

```
closeOnExcept(new CloseableBufferedIterator(streamIter.buffered)) { bufferedStreamIter =>
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +159,169 @@ case class GpuShuffledHashJoinExec(
...
+                while (bufferedBuildIterator.hasNext && sizeInHost <= targetBatchSize) {
+                  val hostBatch = bufferedBuildIterator.next()
+                  val serializedTableColumn =
+                    hostBatch.column(0).asInstanceOf[SerializedTableColumn]
+                  hostBuffs += serializedTableColumn
+                  sizeInHost += serializedTableColumn.hostBuffer.getLength
+                }
+
+                // concat our host batches on the host
+                val hostConcat = withResource(hostBuffs) { _ =>
```
revans2, line 269: We need to move the close block for hostConcat higher up. There is a require in between hostConcat being created and when it is safe to throw an exception.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +159,169 @@ case class GpuShuffledHashJoinExec(
...
+        // stream iterator is empty
+        val builtBatch =
+        ConcatAndConsumeAll.getSingleBatchWithVerification(
+          Option(bufferedBuildIterator).getOrElse(buildIter), buildOutput)
+        val delta = System.nanoTime() - startTime
+        buildTime += delta
+        (builtBatch, streamIter)
+      } else {
+        val dataTypes = buildOutput.map(_.dataType).toArray
+        val bufferedStreamIter = new CloseableBufferedIterator(streamIter.buffered)
```
jlowe, line 252: Do we really need a buffered iterator on the stream-side batch?  We just need to call `hasNext` which will grab the semaphore but I don't see the need for us to call `next` on the stream-side iterator.  That also side-steps another bug here where we create a closeable buffered iterator but pass it to code that will never call close on it since it's just a regular iterator to that code.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +159,169 @@ case class GpuShuffledHashJoinExec(
...
+        val hostBuffs = new ArrayBuffer[SerializedTableColumn]()
+        var buildBatch: ColumnarBatch = null
```
jlowe, line 254: I'm not seeing the need to declare this at this relatively high scope.  IIUC there are two basic cases here:
- We hit the end of the iterator before the target size, we can then build the batch immediately and return it, so no need to track this batch outside of the while loop.
- We hit the target size before the end of the iterator, which then we need to coalesce these all together into a single batch.  That can be accomplished by creating a new iterator by tacking our host-concat-and-put-on-GPU batch on the front of a `GpuShuffleCoalesceIterator` that wraps the build iterator we were using.  We then pass this to `ConcatAndConsumeAll.getSingleBatchWithVerification` and return it immediately.

In other words, this while loop should never be building batches on the GPU iteratively, only fetching host batches until one of two things happen which will be the last iteration of this while loop.

abellina:
I added a commit that I think addresses many of the points, enough has changed that I think this needs another pass.

One comment from Jason wasn't addressed for sure:

> Do we really need a buffered iterator on the stream-side batch? We just need to call hasNext which will grab the semaphore but I don't see the need for us to call next on the stream-side iterator. That also side-steps another bug here where we create a closeable buffered iterator but pass it to code that will never call close on it since it's just a regular iterator to that code.

We need _something_ to hold on to that first stream batch. That `doJoin` pays no attention to this, so if there is an exception before that the first batch is popped, then the batch will be leaked.

I think one ugly option is to attach the closeable iterator to the task completion logic. I think we discussed doing this, but I am confirming that's the approach or if there's something better.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
```diff
@@ -46,15 +46,28 @@ object ConcatAndConsumeAll {
    * @return a single batch with all of them concated together.
```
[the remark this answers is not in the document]
abellina: I should be good now.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
abellina: done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
abellina: Ok I still have this. Let me handle.

abellina: This is no longer an issue.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
abellina: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -470,6 +470,15 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(false)
 
+  val OPTIMIZE_SHUFFLED_HASH_JOIN = conf("spark.rapids.sql.optimizeShuffledHashJoin")
```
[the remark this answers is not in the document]
abellina: Changed the config name an the function name for the optimization.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -129,22 +133,22 @@ case class GpuShuffledHashJoinExec(
     val joinOutputRows = gpuLongMetric(JOIN_OUTPUT_ROWS)
     val targetSize = RapidsConf.GPU_BATCH_SIZE_BYTES.get(conf)
     val spillCallback = GpuMetric.makeSpillCallback(allMetrics)
-    val localBuildOutput: Seq[Attribute] = buildPlan.output
```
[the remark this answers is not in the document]
abellina: Yeap back to `localBuildOutput`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +159,169 @@ case class GpuShuffledHashJoinExec(
...
+          } else {
+            firstBatch.column(0) match {
+              case _: SerializedTableColumn => true
+              case _ => false
+            }
+          }
+        }
+      }
+
+      if (!firstBatchIsSerialized || !streamIter.hasNext) {
```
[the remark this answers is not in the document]
abellina: Yes absolutely. This should be fixed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -123,28 +127,35 @@ case class GpuShuffledHashJoinExec(
...
-    val buildTime = gpuLongMetric(BUILD_TIME)
     val streamTime = gpuLongMetric(STREAM_TIME)
     val joinTime = gpuLongMetric(JOIN_TIME)
     val joinOutputRows = gpuLongMetric(JOIN_OUTPUT_ROWS)
-    val targetSize = RapidsConf.GPU_BATCH_SIZE_BYTES.get(conf)
+    val batchSizeBytes = RapidsConf.GPU_BATCH_SIZE_BYTES.get(conf)
     val spillCallback = GpuMetric.makeSpillCallback(allMetrics)
-    val localBuildOutput: Seq[Attribute] = buildPlan.output
+    val localBuildOutput = buildPlan.output
+    // stomp on metrics that don't make sense in the join
```
jlowe, line 136: This comment is very confusing.  All of these metrics are very relevant for joins.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +166,160 @@ case class GpuShuffledHashJoinExec(
...
+object GpuShuffledHashJoinExec extends Arm {
+  /**
+   * Helper iterator that wraps a BufferedIterator of AutoCloseable subclasses.
+   * This iterator also implements AutoCloseable, so it can be closed in case
+   * of exceptions.
+   *
+   * @param wrapped the buffered iterator
+   * @tparam T an AutoCloseable subclass
+   */
+  class CloseableBufferedIterator[T <: AutoCloseable](wrapped: BufferedIterator[T])
```
jlowe, line 179: Does this need a task completion listener?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +166,160 @@ case class GpuShuffledHashJoinExec(
...
+        } else {
+          bufferedBuildIterator = new CloseableBufferedIterator(buildIter.buffered)
+          val firstBatch = bufferedBuildIterator.head
+          if (firstBatch.numCols() != 1) {
+            false
+          } else {
+            firstBatch.column(0) match {
+              case _: SerializedTableColumn => true
+              case _ => false
+            }
```
jlowe, line 247:
```suggestion
            firstBatch.column(0).isInstanceOf[SerializedTableColumn]
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +166,160 @@ case class GpuShuffledHashJoinExec(
...
+        buildTime += delta
+        (builtBatch, streamIter)
+      } else {
+        val dataTypes = buildOutput.map(_.dataType).toArray
+        closeOnExcept(new CloseableBufferedIterator(streamIter.buffered)) { bufferedStreamIter =>
+          val hostConcatIter = new MaybeHostConcatResultIterator(bufferedBuildIterator,
+              hostTargetBatchSize, dataTypes, coalesceMetricsMap)
+          closeOnExcept(hostConcatIter.next()) { maybeHostConcatResult =>
+            val buildBatch = if (!hostConcatIter.hasNext()) {
+              val singleBatchBufferTimeDelta = System.nanoTime() - startTime
```
jlowe, line 267: It's difficult to follow code when there's a `val somevar = if (...) {` followed by screens of code for the `if` body.  Suggest factoring this into a private method that is responsible for getting the build batch, whether that's using the single host buffer or falling back.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
```diff
@@ -155,3 +166,160 @@ case class GpuShuffledHashJoinExec(
...
+            }
+          }
+        }
+      }
+
+      if (!firstBatchIsSerialized) {
+        // fallback if we failed to find serialized build batches
+        val builtBatch =
+          ConcatAndConsumeAll.getSingleBatchWithVerification(
+            Option(bufferedBuildIterator).getOrElse(buildIter), buildOutput)
```
jlowe, line 256: There should be no need to emulate a fallback here.  If the first batch isn't serialized then that means we're not a join that was directly connected to a shuffle on the build-side.  That means the coalesce should have already occurred, otherwise there's something wrong with the planning.  May be prudent to `assert(!buildIter.hasNext)` but otherwise we should be able to just use the build batch and stream iter directly at this point.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffleCoalesceExec.scala
```diff
@@ -75,55 +113,76 @@ case class GpuShuffleCoalesceExec(child: SparkPlan, targetBatchByteSize: Long)
...
       while (batchCanGrow && iter.hasNext) {
         closeOnExcept(iter.next()) { batch =>
           inputBatchesMetric += 1
           // don't bother tracking empty tables
           if (batch.numRows > 0) {
-            inputRowsMetric += batch.numRows
             val tableColumn = batch.column(0).asInstanceOf[SerializedTableColumn]
             batchCanGrow = canAddToBatch(tableColumn.header)
             serializedTables.addLast(tableColumn)
+            inputRowsMetric += tableColumn.header.getNumRows
```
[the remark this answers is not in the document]
abellina: I thought I had reverted this change, but I didn't. Nice catch

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
abellina: Making the comment better, thanks for pointing it out

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
abellina: It's getting one!

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
abellina: OK, sounds like a plan

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffleCoalesceExec.scala
```diff
@@ -75,55 +113,76 @@ case class GpuShuffleCoalesceExec(child: SparkPlan, targetBatchByteSize: Long)
  * to the target batch size and then concatenated on the host before the data
  * is transferred to the GPU.
  */
-class GpuShuffleCoalesceIterator(
+class MaybeHostConcatResultIterator(
```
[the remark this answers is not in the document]
abellina: This class went away.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffleCoalesceExec.scala
```diff
@@ -61,10 +61,48 @@ case class GpuShuffleCoalesceExec(child: SparkPlan, targetBatchByteSize: Long)
...
+        dataTypes, metricsMap)
+    }
+  }
+}
+
+/**
+ * A helper (but ugly) way of encapsulating the output of the
+ * `MaybeHostConcatResultIterator`. Batches fetched from the shuffle may or may
+ * not be empty, yet still on the host. There isn't a `HostConcatResult` for an empty
+ * batch, so this case class exists to expose a `getColumnarBatch` that always returns
```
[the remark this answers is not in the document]
abellina: This class went away.

sql-plugin/src/main/scala/ai/rapids/cudf/HostConcatResultUtil.scala
```diff
@@ -0,0 +1,63 @@
...
+
+object HostConcatResultUtil extends Arm {
+  /**
+   * Create a `HostConcatResult` that only contains number of rows.
+   *
+   * In order to create an instance of `HostConcatResult` that is closeable,
+   * we allocate a 0-byte `HostMemoryBuffer`. Because this is requesting memory
+   * from the system, it becomes: UNSAFE.allocateMemory(0) which returns 0.
+   * The HostMemoryBuffer.close() logic skips such cases where the address
+   * is 0, so the resulting `HostConcatResult` can be used in withResources easily.
```
jlowe, line 33: Nit: This is a lot of detail that IMO is unnecessary.  We're allocating an empty buffer, and we release that buffer when we're done.  How empty buffers are handled underneath is not this code's concern since it is adhering to the ownership convention of `HostMemoryBuffer`.

sql-plugin/src/main/scala/ai/rapids/cudf/HostConcatResultUtil.scala
```diff
@@ -0,0 +1,63 @@
...
+      hostConcatResult: HostConcatResult,
+      sparkSchema: Array[DataType]): ColumnarBatch = {
+    if (hostConcatResult.getTableHeader.getNumColumns == 0) {
+      // We expect the caller to have acquired the GPU unconditionally before calling
+      // `getColumnarBatch`, as a downstream exec may need the GPU, and the assumption is
+      // that it is acquired in the coalesce code.
+      new ColumnarBatch(Array.empty, hostConcatResult.getTableHeader.getNumRows)
+    } else {
+      withResource(hostConcatResult.toContiguousTable) { ct =>
+        GpuColumnVector.from(ct.getTable, sparkSchema)
```
jlowe, line 59:
```suggestion
        GpuColumnVectorFromBuffer.from(ct, sparkSchema)
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
abellina: Thinking about this one more, I am not seeing the issue, I am probably missing something. The call to `getSingleBatchWithVerification` is building an empty batch (if the build itereator is empty) and is checking that it gets a single GPU batch in the non-empty build iterator case.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
abellina: Ok this should be better here: https://github.com/NVIDIA/spark-rapids/pull/4588/commits/a607e73887ffd05ad863d1af4a464b2a3999ea4a.

jlowe (approved):

abellina:
build

abellina:
Looking into the test failures.

abellina:
build

abellina:
@jlowe @revans2 could you take another look when you have a chance?

jlowe (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCoalesceBatches.scala
```diff
@@ -448,9 +458,84 @@ class GpuCoalesceIterator(iter: Iterator[ColumnarBatch],
     batches.append(SpillableColumnarBatch(batch, SpillPriorities.ACTIVE_BATCHING_PRIORITY,
       spillCallback))
 
+  protected def popAll(): Array[ColumnarBatch] = {
+    closeOnExcept(batches.map(_.getColumnarBatch())) { wip =>
```
revans2, line 462: nit: does this need to be a safeMap?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuShuffledHashJoinExec.scala
revans2: Still would like to know the answer to this. The things in the iterator need to be closed and can potentially be very large.

abellina: When we call `.buffered` on the iterator it won't actively buffer anything, it's wrapping the original iterator in an interator that would buffer 1 `ColumnarBatch` in this case.

When we call `head` that batch is stored in the `BufferedIterator`. The `CloseableBufferedIterator` was added to actively close that if there is an exception.

abellina:
@revans2 I replied to your comment just now in the original request, but adding a link here so we can continue if there's more I need to look into: 

> Still would like to know the answer to this. The things in the iterator need to be closed and can potentially be very large.


https://github.com/NVIDIA/spark-rapids/pull/4588#discussion_r799836101

abellina:
build

revans2 (approved):

abellina:
Ok I have a bug with `safeMap`, I thought I had built this before I pushed it last, but obviously I didn't. I am going to retest the patch against real queries to be sure.

abellina:
build

abellina:
@jlowe @revans2 ok apologies. Now this should be ready.

Issue #33, opened 2020-05-28
[FEA] hypot SQL function


revans2:
**Is your feature request related to a problem? Please describe.**
would be really great to support the SQl function hypot.  We can probably do it initially with just 
```
sqrt((lhs * lhs) + (rhs * rhs))
```

This might not be good enough with overflows etc and we may need to look a little deeper in how to support it.

NVnavkumar:
There are some equivalent forms that avoid overflow/underflow (see https://en.wikipedia.org/wiki/Pythagorean_addition). Will explore.

revans2:
To be clear we have not tested for overflow etc in this at all to know what we can/should do.  The Spark code is using the java built in `Math.hypot`

https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#hypot-double-double-

Which indicates that it takes care of avoiding intermediate overflow and underflow. I am not sure how critical it is when Spark only supports Doubles as input, but it would be good to know.

NVnavkumar:
It does look like the naive algorithm is much more prone to overflow because of the squares, I think we can use:

x * sqrt(1 + (y/x)^2), where x = max(abs(lhs), abs(rhs)) and y = min(abs(lhs), abs(rhs))

PR #4592, opened 2022-01-21
Adds SQL function HYPOT using the GPU

NVnavkumar:
Implements #33. 

This implements the SQL HYPOT function in class GpuHypot, and substitutes the original definition:

```
HYPOT(lhs, rhs) = sqrt(lhs^2 + rhs^2)
```

with this version:

```
HYPOT(lhs, rhs) = r
WHERE:
r = x * sqrt(1 + (y / x)^2)
x = max(abs(lhs), abs(rhs)), y = min(abs(lhs), abs(rhs))
```

This is done to avoid overflow/underflow situations like the Spark implementation, which relies on `java.lang.Math.hypot(double, double)`.  It also implements the following corner cases that are defined in the java version [documentation](https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#hypot-double-double-):
 
 * If either argument is infinite, then the result is positive infinity.
 * If either argument is `NaN` and neither argument is infinite, then the result is `NaN`.

Also it implements another special case:
 * If both arguments are `0.0`, then the result is `0.0`.


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -419,6 +419,81 @@ case class GpuCot(child: Expression) extends GpuUnaryMathExpression("COT") {
...
+  // r = sqrt(lhs^2 + rhs^2) 
+  // This is prone to overflow with regards either square term
+  // However, we can reduce it to
+  // r = sqrt(x^2 + y^2) = x * sqrt(1 + (y/x)^2)
+  // where x = max(abs(lhs), abs(rhs)), y = min(abs(lhs), abs(rhs))
+  // which will only overflow if both terms are near the maximum representation space
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector = {
+    // Using this form, because we really need a max and min that aren't NULL-aware
+    // In that case, then result will always be NULL, which is consistent with Spark
+    val (x, y) = withResource(Seq(lhs.getBase.abs, rhs.getBase.abs)) {
```
revans2, line 436: x and y can leak if an exception is thrown at any point between when they are assigned and line 466 for y and 477 for x. This is also true of many of the scalar values zero, inf, and nan just below this.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -419,6 +419,81 @@ case class GpuCot(child: Expression) extends GpuUnaryMathExpression("COT") {
...
+          returnZero.ifElse(ColumnVector.fromScalar(zero, x.getRowCount.toInt),
+            x.mul(onePlusTermSqrt)
+          )
+        )
+      )
+    }
+  }
+
+  override def doColumnar(lhs: GpuScalar, rhs: GpuColumnVector): ColumnVector = {
+    doColumnar(GpuColumnVector.from(lhs, rhs.getRowCount.toInt, left.dataType), rhs)
```
revans2, line 489: We need to make sure that the new ColumnVector being created is closed too.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -419,6 +419,81 @@ case class GpuCot(child: Expression) extends GpuUnaryMathExpression("COT") {
...
+  }
+
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuScalar): ColumnVector = {
+    doColumnar(lhs, GpuColumnVector.from(rhs, lhs.getRowCount.toInt, right.dataType))
```
revans2, line 493: Need to make sure the new column vector being created is closed too.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -419,6 +419,81 @@ case class GpuCot(child: Expression) extends GpuUnaryMathExpression("COT") {
...
+
+  // naive implementation of hypot
+  // r = sqrt(lhs^2 + rhs^2) 
+  // This is prone to overflow with regards either square term
+  // However, we can reduce it to
+  // r = sqrt(x^2 + y^2) = x * sqrt(1 + (y/x)^2)
+  // where x = max(abs(lhs), abs(rhs)), y = min(abs(lhs), abs(rhs))
+  // which will only overflow if both terms are near the maximum representation space
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector = {
+    // Using this form, because we really need a max and min that aren't NULL-aware
```
revans2, line 434: This is large and complicated enough of a method that I think it would be better if we broke it up into smaller functions in an object.  Just to make the code more readable, and simpler to reason about.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -419,6 +419,81 @@ case class GpuCot(child: Expression) extends GpuUnaryMathExpression("COT") {
...
+    // In that case, then result will always be NULL, which is consistent with Spark
+    val (x, y) = withResource(Seq(lhs.getBase.abs, rhs.getBase.abs)) { 
+      case Seq(lhsAbs, rhsAbs) => {
+        withResource(lhsAbs.greaterThan(rhsAbs)) { lhsGreater => 
+          (lhsGreater.ifElse(lhsAbs, rhsAbs), lhsGreater.ifElse(rhsAbs, lhsAbs))
```
revans2, line 439: nit: We are doing two computations here, and in a number of other places too. If the second computation threw an exception the result of the first one would leak.  It would be nice to clean it up a bit. Something like.

```scala
withResource(lhsAbs.greaterThan(rhsAbs)) { lhsGreater => 
    closeOnExcept(lhsGreater.ifElse(lhsAbs, rhsAbs)) { => x
        (x, lhsGreater.ifElse(rhsAbs, lhsAbs))
    }
}
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
revans2, line 436: Just like the other comment. There are multiple places in the code where you try to keep the line count small my doing two operations in a Sequence or something similar and then putting both in a `withResource` call to make it safe.  That is technically not safe, because if the second one fails the first one can leak.  We have a `safeMap` API that can handle the corner cases for you, but I personally think it would be cleaner to just have multiple withResource calls in the right places and to split things up into smaller methods where possible.

revans2:
Also I just want to be sure we don't forget to move this to be based off of the 22.04 branch instead of the 22.02 branch like it is now.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -419,6 +419,68 @@ case class GpuCot(child: Expression) extends GpuUnaryMathExpression("COT") {
...
+                 y: ColumnVector): ColumnVector = {
+    withResource(x.isNull) { xIsNull =>
+      withResource(y.isNull) { yIsNull =>
+        xIsNull.or(yIsNull)
+      }
+    }
+  }
+
+  def computeHypot(x: ColumnVector, y: ColumnVector): ColumnVector = {
+    withResource(y.div(x)) { yOverX =>
```
revans2, line 470: Nit: there are a lot of intermediate values that live longer than they need to.  Could we do it a bit more like.

```
    val yOverXSquared = withResource(y.div(x)) { yOverX =>
        yOverX.mul(yOverX)
    }
    val onePlusTerm = withResource(yOverXSquared) { yOverXSquared =>
      withResource(Scalar.fromDouble(1)) { one =>
        one.add(yOverXSquared)
      }
    }
    val onePlusTermSqrt = withResource(onePlusTerm) { onePlusTerm =>
      onePlusTerm.sqrt
    }
    withResource(onePlusTermSqrt) { onePlusTermSqrt =>
      x.mul(onePlusTermSqrt)
    }
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -419,6 +419,128 @@ case class GpuCot(child: Expression) extends GpuUnaryMathExpression("COT") {
...
+                            Scalar.fromDouble(Double.PositiveInfinity),
+                            Scalar.fromDouble(Double.NaN))
+    withResource(litForSpecial) { case Seq(zero, inf, nan) =>
+      withResource(GpuHypot.chooseXandY(lhs.getBase, rhs.getBase)) { case Seq(x, y) =>
+        withResource(Scalar.fromNull(x.getType)) { nullS => 
+          withResource(GpuHypot.either(inf, x, y)) { eitherInf =>
+            withResource(GpuHypot.eitherNull(x, y)) { eitherNull =>
+              withResource(GpuHypot.eitherNan(x, y)) { eitherNan =>
+                withResource(GpuHypot.both(zero, x, y)) { bothZero =>
+                  withResource(GpuHypot.computeHypot(x, y)) { hypotComputed =>
```
revans2, line 513: nit: here we could do the same king of thing to reduce total memory usage.

```
withResource(litForSpecial) { case Seq(zero, inf, nan) =>
  withResource(GpuHypot.chooseXandY(lhs.getBase, rhs.getBase)) { case Seq(x, y) =>
    val zeroOrBase = withResource(GpuHypot.both(zero, x, y)) { bothZero =>
      withResource(GpuHypot.computeHypot(x, y)) { hypotComputed =>
        bothZero.ifElse(zero, hypotComputed)
      }
    }
    val nanOrRest = withResource(zeroOrBase) { zeroOrBase =>
      withResource(GpuHypot.eitherNan(x, y)) { eitherNan =>
        eitherNan.ifElse(nan, zeroOrBase)
      }
    }
    val infOrRest = withResource(nanOrRest) { nanOrRest =>
      withResource(GpuHypot.either(inf, x, y)) { eitherInf =>
        eitherInf.ifElse(inf, nanOrRest)
      }
    }
    withResource(infOrRest) { infOrRest =>
      withResource(GpuHypot.eitherNull(x, y)) { eitherNull =>
        withResource(Scalar.fromNull(x.getType)) { nullS => 
          eitherNull.ifElse(nullS, infOrRest)
        }
      }
    }
  }
}
```

revans2 (commented):
Just some nits to improve the worst case memory usage. Looking really good

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
NVnavkumar: Cleaned up in latest commit

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
NVnavkumar: Cleaned up in latest commit

NVnavkumar:
build

revans2 (approved):

PR #4596, opened 2022-01-21
Enable aggregate push down for parquet on GPU

### LINKED ISSUE #3951: [FEA] Aggregate (Min/Max/Count) push down for Parquet
https://issues.apache.org/jira/browse/SPARK-36645 added min/max push downs to parquet in 3.3.0.  This feature is off by default, but it allows a metadata only query to run on the CPU.  It would be great if we could implement this feature, but at a minimum if we see it being used we should fall back to the CPU because there is little performance improvement that we will be able to do, as it is a meta data only query.

HaoYang670:
Signed-off-by: remzi 13716567376yh@gmail.com
close #3951

In this PR, we:
1. add the aggregation push down information in `GpuParquetScan`
2. Create parquet scan reader for aggregation push down: read metadata on CPU and convert to GPU column batch
3. Update shim layers:
- move `GpuParquetScan` from 311+ to 311until330 (new created) and 330+ (add new parameter `pushedAggregate`
- move `GetScan` from 320+ to 320until322, 322until330(new created) and 330+
4. update pom file


HaoYang670:
build

HaoYang670:
build

HaoYang670:
build

HaoYang670:
build

sameerz (changes requested):
Please retarget to 22.04, as we are in burndown for 22.02.

integration_tests/src/main/python/parquet_test.py
```diff
@@ -12,13 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from distutils.log import FATAL
+from sqlite3 import Timestamp
```
jlowe, line 16: These odd imports appear to be unused

integration_tests/src/main/python/parquet_test.py
```diff
@@ -685,3 +688,351 @@ def test_parquet_reading_from_unaligned_pages_basic_filters_with_nulls(spark_tmp
...
+        spark.createDataFrame(data).write.mode('overwrite').parquet(data_path)
+        
+        df = spark.read.parquet(data_path).selectExpr("Max(_1)")
+        explain = df._sc._jvm.PythonSQLUtils.explainString(df._jdf.queryExecution(), "simple")
+
+        assert "PushedAggregation: []" in explain
+        return df
+
+    assert_gpu_and_cpu_are_equal_collect(do_explain, {"spark.sql.parquet.aggregatePushdown": "true", 
+                                  "spark.sql.sources.useV1SourceList": ""})
```
jlowe, line 711: Nit: It would be nice to factor out the configs specific to aggregate pushdown into a variable that can be reused in the various aggregate pushdown tests to simplify those tests, simplify maintenance, and reduce the chance of config typos.

pom.xml
```diff
@@ -732,7 +732,7 @@
                                         <source>${project.basedir}/src/main/311+-nondb/scala</source>
                                         <source>${project.basedir}/src/main/311until330-all/scala</source>
                                         <source>${project.basedir}/src/main/320+/scala</source>
-                                        <source>${project.basedir}/src/main/322+/scala</source>
+                                        <source>${project.basedir}/src/main/322until330/scala</source>
```
jlowe, line 735: Why remove 322+?  That forces ShimVectorizedColumnReader to be copied essentially verbatim when it otherwise would not need to be.  Seems like we want both a 322+, to house code that truly is common for everything >= 3.2.2, and a separate 322until320 for code that only applies to Spark >= 3.2.2 and < 3.3.0.

sql-plugin/src/main/311until330-all/scala/com/nvidia/spark/rapids/shims/v2/GpuParquetScan.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2022, NVIDIA CORPORATION.
```
jlowe, line 2: Updating the copyright on a file that is otherwise untouched.

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuParquetScan.scala
```diff
@@ -0,0 +1,230 @@
...
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.sources.Filter
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.util.CaseInsensitiveStringMap
+import org.apache.spark.sql.vectorized.ColumnarBatch
+import org.apache.spark.util.SerializableConfiguration
+
+
+
+
```
jlowe, line 50:
```suggestion
```

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuParquetScan.scala
```diff
@@ -0,0 +1,230 @@
...
+case class GpuParquetPartitionReaderFactoryForAggrerationPushdown(
```
jlowe, line 51:
```suggestion
case class GpuParquetPartitionReaderFactoryForAggregationPushdown(
```

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuParquetScan.scala
```diff
@@ -0,0 +1,230 @@
...
+  private val maxReadBatchSizeBytes = rapidsConf.maxReadBatchSizeBytes
+  private val datetimeRebaseModeInRead = parquetOptions.datetimeRebaseModeInRead
+
+  private val filterHandler = GpuParquetFileFilterHandler(sqlConf)
+
+  private def getFooter(file: PartitionedFile): ParquetMetadata = {
+    val conf = broadcastedConf.value.value
+    val filePath = new Path(new URI(file.filePath))
+
+    if (aggregation.isEmpty) {
```
jlowe, line 74: What's the point of instantiating this class with empty aggregations?  Spark has these checks because it doesn't have a separate factory for aggregation pushdown, but since this is using a separate class, why have the extra checks here?  Seems like we should simply put a `require(aggregation.nonEmpty)` at the top of this class, document it in the scaladoc for this class, and never expect someone to try to use this when there are no aggregation filters.

This would simplify the class and also reduce the number of parameters needed to be passed to this class.

integration_tests/src/main/python/parquet_test.py
HaoYang670: Oh, my mistake! Have removed them.

integration_tests/src/main/python/parquet_test.py
HaoYang670: Thank you for the advice! I'll refactor them.

HaoYang670: done!

pom.xml
HaoYang670: Sorry, my mistake! I will restore 322+ and let 322until330 extend 322+.

sql-plugin/src/main/311until330-all/scala/com/nvidia/spark/rapids/shims/v2/GpuParquetScan.scala
HaoYang670: Fixed

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuParquetScan.scala
HaoYang670: fixed

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuParquetScan.scala
HaoYang670: OH! My spelling mistake, has been fixed!

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuParquetScan.scala
HaoYang670: You are right! I will fix it.

HaoYang670:
build

HaoYang670:
build

HaoYang670:
> Please retarget to 22.04, as we are in burndown for 22.02.

Done

integration_tests/src/main/python/parquet_test.py
```diff
@@ -685,3 +686,339 @@ def test_parquet_reading_from_unaligned_pages_basic_filters_with_nulls(spark_tmp
...
+
+    conf = {"spark.sql.parquet.aggregatePushdown": "true", 
+            "spark.sql.sources.useV1SourceList": "",
+            "spark.sql.parquet.datetimeRebaseModeInWrite": "CORRECTED",
+            "spark.sql.parquet.int96RebaseModeInWrite": "CORRECTED",
+            "spark.rapids.sql.format.parquet.writer.int96.enabled": "false"}
+
+    
+    assert_gpu_and_cpu_are_equal_collect(timestamp_not_push_down, conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(without_timestamp_and_binary, conf=conf)
```
jlowe, line 1024: This looks like two unrelated tests crammed into one, as they have separate local functions being called.  Only significant thing being shared here is the data generator which can be factored out into a module variable and used in each separate test.

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuParquetScan.scala
```diff
@@ -0,0 +1,203 @@
...
+import org.apache.spark.sql.execution.datasources.v2.{FilePartitionReaderFactory, FileScan}
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.sources.Filter
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.util.CaseInsensitiveStringMap
+import org.apache.spark.sql.vectorized.ColumnarBatch
+import org.apache.spark.util.SerializableConfiguration
+
+/**
+ * Users should make sure that `aggregation` is valid when using this factory.
```
jlowe, line 49: This comment doesn't add any value.  It's clear from the signature that an aggregation needs to be passed since it's not `Option[Aggregation]`, otherwise what makes it important to not pass `null` for an aggregation vs. passing `null` for `broadcastedConf` or the other parameters?

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuParquetScan.scala
```diff
@@ -0,0 +1,203 @@
...
+    @transient sqlConf: SQLConf,
+    broadcastedConf: Broadcast[SerializableConfiguration],
+    dataSchema: StructType,
+    readDataSchema: StructType,
+    partitionSchema: StructType,
+    filters: Array[Filter],
+    aggregation: Aggregation,
+    parquetOptions: ParquetOptions,
+    @transient rapidsConf: RapidsConf,
+  metrics: Map[String, GpuMetric]) extends FilePartitionReaderFactory with Arm with Logging {
```
jlowe, line 61: There's a bunch of unused parameters here, e.g.:
```suggestion
    broadcastedConf: Broadcast[SerializableConfiguration],
    dataSchema: StructType,
    readDataSchema: StructType,
    partitionSchema: StructType,
    aggregation: Aggregation,
    parquetOptions: ParquetOptions)
    extends FilePartitionReaderFactory with Arm with Logging {
```

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuParquetScan.scala
```diff
@@ -0,0 +1,203 @@
...
+            partitionedFile.filePath,
+            dataSchema,
+            partitionSchema,
+            aggregation,
+            readDataSchema,
+            partitionedFile.partitionValues,
+            getDatetimeRebaseSpec(footer.getFileMetaData)
+          )
+
+          new GpuRowToColumnConverter(readDataSchema).convertBatch(Array(row), readDataSchema)
```
jlowe, line 103: If I'm reading this code correctly, it seems like we're going out of our way to translate a Parquet scan to the GPU even though at planning time we know that this will just be a CPU-only read of the footer and agg statistics from it.  The GPU "implementation" is just calling the underlying Spark code and then following it up with a RowToColumnar transition.  Why not simply have the tagging code _not_ replace a Parquet scan when we know that the Parquet scan will simply be reading aggregation values from the footer?  Then the RowToColumnar will be put into the plan automatically when necessary.

This also can avoid performance issues where it looks like the GPU is performing the Parquet scan (when it really isn't) and we convert it from row to columnar here internally, but then in the query plan this node is followed by something we cannot do on the GPU, so then we follow up this internal RowToColumn conversion with a ColumnarToRow conversion in the query plan to continue execution on the CPU when we should have remained on the CPU in row mode all along.

HaoYang670: Yes, you are right. Fallback to CPU is the reasonable solution. I will close this PR and reopen a new one.

Issue #4536, opened 2022-01-14
[BUG] Investigate odd AST slowness with ASYNC allocator.


revans2:
**Describe the bug**
I am having some really odd problems with the ASYNC allocator. So it could just be related to my setup specifically. I have been testing the performance of AST against binops in a number of situations. I found that when using the ASYNC allocator and a large number of rows (100000000) many of the AST runs get really slow. Like on the order of seconds slow.  I have seen this happen with BOOL8 doing (LOGICAL_AND, LOGICAL_OR) and (FLOAT32, FLOAT64, INT32, INT64) doing (ADD, DIV, MUL, SUB, EQUAL, GREATER, GREATER_EQUAL, LESS, LESS_EQUAL, and NOT_EQUAL)

It is especially odd because it does not happen if there are nulls in the data.  It does not happen if AST runs before binops, only after. It does not happen with smaller amounts of data. And for some of the operators it only happens when there are a small number of operations being done. It does not happen with no polling, and it does not happen with ARENA.

I have also seen the test hang on shutdown, and the only way to stop it is with a kill -9, but that may be a separate issue.

**Steps/Code to reproduce bug**
[DataTest.scala.zip](https://github.com/NVIDIA/spark-rapids/files/7872187/DataTest.scala.zip)

**Expected behavior**
AST is not 500 times slower in this odd corner case.

rongou:
Was able to reproduce this on my desktop. The profile looks pretty funky:
```console
CUDA API Statistics:

 Time (%)  Total Time (ns)  Num Calls    Avg (ns)       Med (ns)      Min (ns)      Max (ns)      StdDev (ns)                  Name                
 --------  ---------------  ---------  -------------  -------------  -----------  -------------  -------------  -----------------------------------
     49.3    1,722,493,522         10  172,249,352.2        1,302.5          871  1,722,475,770  544,694,041.0  cudaStreamSynchronize_ptsz         
     27.9      975,786,305      1,605      607,966.5        4,288.0        3,737      2,062,738      817,300.2  cudaLaunchKernel_ptsz              
     16.9      590,838,738          1  590,838,738.0  590,838,738.0  590,838,738    590,838,738            0.0  cudaFree                           
      5.1      177,420,710          1  177,420,710.0  177,420,710.0  177,420,710    177,420,710            0.0  cudaMemPoolDestroy_v11020          
[...]

PR #4606, opened 2022-01-21
Set default RMM pool to ASYNC for cuda 11.2+

### LINKED ISSUE #4515: [FEA] Set RMM async allocator as default
Currently the RMM arena allocator is the default, but in some circumstances it can still run into OOM errors due to memory fragmentation. The async allocator, relying on `cudaMallocAsync` and `cudaFreeAsync`, can remap physical pages when running of GPU memory, thus is more resistant to memory fragmentation.

Set the async allocator as the default.

Continue improving the arena allocator is possible, but since it doesn't remap memory, probably won't ever be as fragmentation resistance as the async allocator.

**Additional context**
Set this at the beginning of 22.04 for more testing.

rongou:
Fixes #4515 


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -404,10 +406,11 @@ object RapidsConf {
...
       "\"ARENA\", the RMM arena allocator is used; with \"ASYNC\", the new CUDA stream-ordered " +
       "memory allocator in CUDA 11.2+ is used. If set to \"NONE\", pooling is disabled and RMM " +
-      "just passes through to CUDA memory allocation directly. Note: \"ARENA\" is the " +
-      "recommended pool allocator if CUDF is built with Per-Thread Default Stream (PTDS).")
+      "just passes through to CUDA memory allocation directly. When unset, if the CUDA driver is " +
+      "newer than 11.2, the \"ASYNC\" allocator is used; otherwise, the \"ARENA\" allocator is " +
+      "used.")
     .stringConf
-    .createWithDefault("ARENA")
+    .createWithDefault("")
```
jlowe, line 413: Given the number of users on CUDA < 11.2 will be dwindling over time quickly, wondering if we should just have `ASYNC` automatically fallback to `ARENA` with a log message on startup when the fallback occurs.

jlowe (commented):
Are we really ready for this?  It would be good to have benchmark results showing this is not going to be a regression for some use-cases.  It seems like we still have unresolved performance issues that appear to be tied to this allocator, e.g.:#4536.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
rongou: Done.

rongou:
We do have benchmarks on TPCDS at various scale factors. The idea is turn this on early in 22.04 and get it more widely tested and benchmarked. If we do run into some block we can always revert it back before the release is cut.

jlowe:
> We do have benchmarks on TPCDS at various scale factors.

Do we have those metrics isolated to just this change?  There are quite a few performance improvements going in recently, and I want to make sure we don't end up masking a regression from the async allocator with a performance improvement from elsewhere.

I would really like to see what this allocator costs for queries that don't spill, and last I knew we only had metrics at scale for UCX shuffle which of course can spill quite a lot due to its aggressive caching of shuffle outputs.  The only metrics I've seen for a query that doesn't spill is from #4536, and I find it odd that we want to check this in without first understanding the impact footprint for a 500X slowdown.

revans2:
> I would really like to see what this allocator costs for queries that don't spill, and last I knew we only had metrics at scale for UCX shuffle which of course can spill quite a lot due to its aggressive caching of shuffle outputs. The only metrics I've seen for a query that doesn't spill is from #4536, and I find it odd that we want to check this in without first understanding the impact footprint for a 500X slowdown.

To be clear most of the tests didn't show that slowness. I would need to do some more analysis to see if there was any real difference that I saw beyond this one corner case.

rongou:
This is the result for TPCDS at 1TB. I'll dig more into #4536.
ARENA vs ASYNC allocator
.

rongou:
#4536 turns out to be a red herring.

jlowe (approved):

rongou:
build

abellina (approved):

PR #4611, opened 2022-01-24
Support BitLength and OctetLength

### LINKED ISSUE #4142: [Audit][FEA][SPARK-36751] Accelerate more flavors of string length like bit_length
The plugin lacks overrides for various string length functions initially introduced by 
SPARK-20749

sperlingxx:

Closes #4142

This PR is to provide GPU overrides for Spark SQL functions: `BitLength` and `OctetLength`


sperlingxx:
build

jlowe (approved):

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
```diff
@@ -56,4 +57,24 @@ trait Spark33XShims extends Spark33XFileOptionsShims {
     new ParquetFilters(schema, pushDownDate, pushDownTimestamp, pushDownDecimal, pushDownStartWith,
       pushDownInFilterThreshold, caseSensitive, datetimeRebaseMode)
   }
+
+  override def getExprs: Map[Class[_ <: Expression], ExprRule[_ <: Expression]] = {
+    (super.getExprs.values ++ Seq(
+      GpuOverrides.expr[BitLength](
```
revans2, line 63: BitLength has been in Spark since 2.3.0. Why are we only putting it into the 3.3 shim?

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
```diff
@@ -56,4 +57,24 @@ trait Spark33XShims extends Spark33XFileOptionsShims {
...
+        "The bit length of string data",
+        ExprChecks.unaryProject(TypeSig.INT, TypeSig.INT,
+                                TypeSig.STRING, TypeSig.STRING + TypeSig.BINARY),
+        (a, conf, p, r) => new UnaryExprMeta[BitLength](a, conf, p, r) {
+          override def convertToGpu(child: Expression): GpuExpression = GpuBitLength(child)
+        }),
+      GpuOverrides.expr[OctetLength](
```
revans2, line 70: OctetLength has been in Spark since 2.3.0. Why are we only putting it into the 3.3 shim?

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
sperlingxx: Because they were not available for Spark users through any explicit way before Spark 330.

revans2: https://github.com/apache/spark/blob/3fdfce3120f307147244e5eaf46d61419a723d50/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala#L347 and https://github.com/apache/spark/blob/3fdfce3120f307147244e5eaf46d61419a723d50/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala#L329

They are in the SQL function registry for v3.0.0. They appear to be available for all versions of Spark that we support. I don't think you would even need to change the test because you are calling them from the SQL functions.  In 3.3 they scala APIs were added to make it simpler to call them, but they have been available for a long time.

sperlingxx:
build

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
sperlingxx: I moved these bindings into the common section.

integration_tests/src/main/python/string_test.py
```diff
@@ -401,6 +402,14 @@ def test_length():
                 'CHAR_LENGTH(a)',
                 'CHARACTER_LENGTH(a)'))
 
+@pytest.mark.skipif(is_before_spark_330(),
+                    reason="octet_length and bit_length only available in 330+")
```
jlowe, line 406:
```suggestion
```
Also should remove the (now unnecessary) import change.

sperlingxx: Yes, it was removed now.

sperlingxx:
build

sperlingxx:
build

jlowe (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3352,7 +3352,23 @@ object GpuOverrides extends Logging {
...
+    expr[BitLength](
+      "The bit length of string data",
+      ExprChecks.unaryProject(
+        TypeSig.INT, TypeSig.INT,
+        TypeSig.STRING, TypeSig.STRING + TypeSig.BINARY),
+      (a, conf, p, r) => new UnaryExprMeta[BitLength](a, conf, p, r) {
+        override def convertToGpu(child: Expression): GpuExpression = GpuBitLength(child)
+      }),
+    expr[OctetLength](
+      "The bit length of string data",
```
revans2, line 3365: Sorry one more thing. This is the **byte** length not the **bit** length.

revans2 (changes requested):

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala

revans2 (approved):

revans2:
@tgravescs do you plan on handling updating the 2.x explain or do we need to start enforcing it on new PRs?

PR #4613, opened 2022-01-24
fallback to cpu when detecting column pruning in json reader

wbo4958:
It's really a bad user experience when a user tries to read JSON files with partial data schema (AKA column pruning), since the CUDF will throw exception.

So as an improvement, this PR added inferring JSON schema when tagging. If the data schema is not the same with the schema inferred, we will fallback the JSON reader to run on GPU instead of throwing an exception. But the extra inferring schema may have caused performance issues.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
@@ -32,8 +32,8 @@ import org.apache.spark.sql.catalyst.json.{JSONOptions, JSONOptionsInRead}
 import org.apache.spark.sql.catalyst.util.PermissiveMode
 import org.apache.spark.sql.connector.read.{PartitionReader, PartitionReaderFactory}
 import org.apache.spark.sql.execution.QueryExecutionException
-import org.apache.spark.sql.execution.datasources.json.JsonDataSource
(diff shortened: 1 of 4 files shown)


wbo4958:
build

wbo4958:
build

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
```diff
@@ -74,21 +78,38 @@ object GpuJsonScan {
...
+      files: Seq[FileStatus]): Unit = {
 
     val parsedOptions = new JSONOptionsInRead(
       options,
       sparkSession.sessionState.conf.sessionLocalTimeZone,
       sparkSession.sessionState.conf.columnNameOfCorruptRecord)
 
+    // Since cudf (https://github.com/rapidsai/cudf/issues/9990) does not support
+    // column pruning for now, we just fallback to CPU.
+    val optSchema = JsonDataSource(parsedOptions).inferSchema(sparkSession, files, parsedOptions)
```
revans2, line 100: This is crazy expensive to do. By default it will read all of the json files and not even sub-sample them. It will be as expensive as parsing the JSON on the CPU. It will never speed up processing. I would rather keep the bad user experience and document it instead.

revans2 (changes requested):

wbo4958:
Close this PR since it will have terrible performance issue.

Issue #4599, opened 2022-01-21
[BUG] test_group_apply_udf and test_group_apply_udf_more_types hangs on Databricks 9.1


jlowe:
Running `test_group_apply_udf` or `test_group_apply_udf_more_types` hangs in the Databricks 9.1 environment.  There's no CPU utilization, so it is not an infinite loop.  From a stack trace, it appears the code is waiting for data from Python that never arrives:
```
"Executor task launch worker for task 1.0 in stage 6.0 (TID 45)" #79 daemon prio=5 os_prio=0 tid=0x00007fab6c140000 nid=0x5876 runnable [0x00007faaa43be000]
   java.lang.Thread.State: RUNNABLE
	at java.net.SocketInputStream.socketRead0(Native Method)
	at java.net.SocketInputStream.socketRead(SocketInputStream.java:116)
	at java.net.SocketInputStream.read(SocketInputStream.java:171)
	at java.net.SocketInputStream.read(SocketInputStream.java:141)
	at java.io.BufferedInputStream.fill(BufferedInputStream.java:246)
	at java.io.BufferedInputStream.read(BufferedInputStream.java:265)
	- locked <0x00000000fa10ad40> (a java.io.BufferedInputStream)
	at java.io.DataInputStream.readInt(DataInputStream.java:387)
	at org.apache.spark.sql.rapids.execution.python.GpuPythonArrowOutput$$anon$1.read(GpuArrowEvalPythonExec.scala:328)
	at org.apache.spark.sql.rapids.execution.python.GpuPythonArrowOutput$$anon$1.read(GpuArrowEvalPythonExec.scala:285)
```

firestarman:
There is a DB specific config `spark.databricks.execution.pandasZeroConfConversion.groupbyApply.enabled` which is 'false' by default.
The test can pass after setting this conf to `true`.

Seems DB9.1 supports disabling this feature and disables it by default.

PR #4618, opened 2022-01-24
[DB31x] Pick the correct Python runner for flatmap-group Pandas UDF[databricks]

firestarman:
This PR follows the change introduced in Databricks 9.1 to choose the correct Python runner (`GpuGroupUDFArrowPythonRunner` or `GpuArrowPythonRunner`) for the `GpuFlatMapGroupsInPandasExec` according to two Databricks specific configs.

These two Python runners use different protocols to communicate with the Python side. And the protocol choice is made by the two related configs.

Fixes https://github.com/NVIDIA/spark-rapids/issues/4599

### DIFF
--- a/integration_tests/src/main/python/udf_test.py
+++ b/integration_tests/src/main/python/udf_test.py
@@ -213,6 +213,29 @@ def test_window_aggregate_udf_array_from_python(data_gen, window):
 
 
 # ======= Test flat map group in Pandas =======
+
+# seperate the tests into before and after db 91. To verify
+# the new "zero-conf-conversion" feature introduced from db 9.1.
+@pytest.mark.skipif(not is_databricks91_or_later(), reason="zero-conf is supported only from db9.1")
+@ignore_order(local=True)
+@pytest.mark.parametrize('zero_enabled', [False, True])
+@pytest.mark.parametrize('data_gen', [LongGen()], ids=idfn)
+def test_group_apply_udf_zero_conf(data_gen, zero_enabled):
+    def pandas_add(data):
+        data.sum = data.b + data.a
+        return data
+
+    conf_with_zero = arrow_udf_conf.update({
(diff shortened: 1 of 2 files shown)


firestarman:
Still working on the tests, so mark it as draft

firestarman:
build

firestarman:
I am wondering whether we need to test the case for enabling this 'zero-conf-conversion-group'.

jlowe (commented):
I do think we need explicit tests for zero conf enabled and disabled to make sure we don't regress.  If we don't test it, we won't know whether there's a problem that crops up later, and hanging is a pretty bad problem.

firestarman:
> I do think we need explicit tests for zero conf enabled and disabled to make sure we don't regress. If we don't test it, we won't know whether there's a problem that crops up later, and hanging is a pretty bad problem.

Added the tests.

I thought the DB changes too much without any rule, so the DB-specific tests should be fragile and may fail unexpectedly one day. Besides this config seems not easy to be understood by users, and it is disabled by default, so I guess there are few cases to enable it in practice. Then I thought no need to test it.

firestarman:
build

firestarman:
build

integration_tests/src/main/python/udf_test.py
```diff
@@ -213,8 +213,31 @@ def pandas_sum(to_process: pd.Series) -> list:
 
 
 # ======= Test flat map group in Pandas =======
-@pytest.mark.skipif(is_databricks91_or_later(), reason="https://github.com/NVIDIA/spark-rapids/issues/4599")
-@ignore_order
+
+# seperate the tests into before and after db 91. To verify
```
jlowe, line 217:
```suggestion
# separate the tests into before and after db 91. To verify
```

jlowe (approved):
Minor typo but otherwise lgtm.

firestarman:
build

pxLi (approved):
Approve for typo fix

Issue #4253, opened 2021-12-01
[BUG] Dependencies missing of spark-rapids v21.12.0 release jars


NvTimLiu:
**Describe the bug**
We use “mvn dependency:get” command to download the rapids-4-spark jar from the OSS staging repo for the release test [spark-tests.sh#L33-L34](https://github.com/NVIDIA/spark-rapids/blob/branch-22.02/jenkins/spark-tests.sh#L33-L34).
What we want is to download the rapids-4-spark jar only, maven tries to get its dependency jars as blow, which are not available, because we will NOT release these dependencies.

We used `denpendency-reduce` before in dist/pom.xml to void some these denpendencies, but not on branch-21.12, we remove the `denpendency-reduce`  from dist/pom.xml

Related issue on branch-21.10:  #3929 #3927

```
mvn dependency:get -B -Ddest=snap -DremoteRepositories=https://oss.sonatype.org/content/repositories/staging -DgroupId=com.nvidia -DartifactId=rapids-4-spark_2.12 -Dversion=21.12.0 -s spark-rapids/jenkins/settings.xml -Dpackaging=jar

22:57:31  [INFO] ------------------------------------------------------------------------
22:57:31  [INFO] Reactor Summary for RAPIDS Accelerator for Apache Spark Root Project 21.12.0-SNAPSHOT:
22:57:31  [INFO] 
22:57:31  [INFO] RAPIDS Accelerator for Apache Spark Root Project ... FAILURE [ 14.257 s]
22:57:31  [INFO] RAPIDS Accelerator for Apache Spark SQL Plugin ..... SKIPPED
22:57:31  [INFO] RAPIDS Accelerator for Apache Spark Shuffle Plugin . SKIPPED
22:57:31  [INFO] RAPIDS Accelerator for Apache Spark SQL Plugin Shims SKIPPED
22:57:31  [INFO] RAPIDS Accelerator for Apache Spark SQL Plugin Spark 3.0.1 Shim SKIPPED
22:57:31  [INFO] RAPIDS Accelerator for Apache Spark Scala UDF Plugin SKIPPED
22:57:31  [INFO] RAPIDS Accelerator for Apache Spark Aggregator ..... SKIPPED
22:57:31  [INFO] RAPIDS Accelerator for Apache Spark Distribution ... SKIPPED
22:57:31  [INFO] RAPIDS Accelerator for Apache Spark UDF Examples ... SKIPPED
[...]

Issue #3971, opened 2021-10-29
[BUG] udf-examples dependencies are incorrect


jlowe:
The `udf-examples` subproject depends on the unpublished `rapids-4-spark-sql` rather than the published distribution jar, `rapids-4-spark`.  Simply updating the dependency to be what is expected will break the `test` build phase as described in #3966.

Ideally the UDF examples project should be migrated to an external project repository so it more accurately reflects a true example that users can replicate and is not involved in the build process of the RAPIDS Accelerator directly.  By building the examples in the same repository as the RAPIDS Accelerator, we risk accidentally relying on an environment that an external project cannot replicate without also building the RAPIDS Accelerator.

Issue #3966, opened 2021-10-29
[BUG] udf-examples module fails on `mvn compile` and `mvn test`


firestarman:
**Describe the bug**
 `mvn compile` and `mvn test` are failing due to the error as below

```
[INFO] --- scala-maven-plugin:4.3.0:compile (scala-compile-first) @ rapids-4-spark-udf-examples_2.12 ---
[INFO] Using incremental compilation using Mixed compile order
[INFO] Compiler bridge file: /home/liangcail/.sbt/1.0/zinc/org.scala-sbt/org.scala-sbt-compiler-bridge_2.12-1.3.1-bin_2.12.8__52.0-1.3.1_20191012T045515.jar
[INFO] Compiling 2 Scala sources and 7 Java sources to /home/liangcail/work/projects/on_github/spark-rapids/udf-examples/target/classes ...
[ERROR] [Error] /home/liangcail/work/projects/on_github/spark-rapids/udf-examples/src/main/scala/com/nvidia/spark/rapids/udf/scala/URLDecode.scala:22: object RapidsUDF is not a member of package com.nvidia.spark
[ERROR] [Error] /home/liangcail/work/projects/on_github/spark-rapids/udf-examples/src/main/scala/com/nvidia/spark/rapids/udf/scala/URLDecode.scala:30: not found: type RapidsUDF
[ERROR] [Error] /home/liangcail/work/projects/on_github/spark-rapids/udf-examples/src/main/scala/com/nvidia/spark/rapids/udf/scala/URLEncode.scala:22: object RapidsUDF is not a member of package com.nvidia.spark
[ERROR] [Error] /home/liangcail/work/projects/on_github/spark-rapids/udf-examples/src/main/scala/com/nvidia/spark/rapids/udf/scala/URLEncode.scala:30: not found: type RapidsUDF
[ERROR] four errors found
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for RAPIDS Accelerator for Apache Spark Root Project 21.12.0-SNAPSHOT:
[INFO] 
[INFO] RAPIDS Accelerator for Apache Spark Root Project ... SUCCESS [  0.636 s]
[INFO] RAPIDS Accelerator for Apache Spark SQL Plugin ..... SUCCESS [ 26.145 s]
[INFO] RAPIDS Accelerator for Apache Spark Shuffle Plugin . SUCCESS [  4.883 s]
[...]

PR #4619, opened 2022-01-24
Move the udf-examples module to the external repository spark-rapids-examples [databricks]

res-life:
[BUG] udf-examples dependencies are incorrect #3971

Move the the whole udf-examples module to the external spark-rapids-examples repository.
The moved udf-examples depends on Spark Rapids repository and Spark Rapids repository should not depends on the external udf-examples. This will avoid circular dependency between the two repositories.
Done: Update the jenkens pipeline to depend on the external udf-examples for test purpose.

Background:
- mvn compile or mvn test for a clean build will fail, because of dist.jar is generated in the package phase and compile phase can't access the classes in the dist.jar. See issue [#3966](https://github.com/NVIDIA/spark-rapids/issues/3966)
- Dependency dist.jar is not the same with the published one. The published one always removes the rapids-4-spark-aggregator dependency,  see issue [#4253](https://github.com/NVIDIA/spark-rapids/issues/4253)

Contributes to #3971


res-life:
build

udf-examples/pom.xml
```diff
@@ -32,13 +32,6 @@
   <version>22.02.0-SNAPSHOT</version>
```
jlowe, line 32: I don't understand why this pom is still here.  It claims to be the UDF examples artifact, but supposedly we're moving that jar to another repository?  Also I'm not a fan of having a circular dependency in this repository where this project depends on the UDF examples jar to test (but oddly, those tests are not associated with the UDF examples code that was moved?), and the external UDF examples project depends on this project to build.

IMO everything associated with udf-examples should be moved to the external repository, code _and_ tests.  If the desire is to ensure UDF example tests are passing, then we can always update premerge and nightly build scripts to build and locally install the artifacts from this repository then subsequently build and test the udf-examples project.

udf-examples/pom.xml
```diff
@@ -58,14 +51,8 @@
     </dependency>
     <dependency>
         <groupId>com.nvidia</groupId>
-        <!--
-        This should depend on rapids-4-spark_${scala.binary.version} instead, but that dependency
-        only exists after the package phase. External projects should depend on
-        rapids-4-spark_${scala.binary.version}.
-        -->
-        <artifactId>rapids-4-spark-sql_${scala.binary.version}</artifactId>
+        <artifactId>rapids-4-spark_${scala.binary.version}</artifactId>
```
jlowe, line 54: The original complaint when I tried to make this change earlier (see #3945) is that it breaks `mvn test` on a clean build because this artifact does not exist until the `package` phase is executed.  That's one, but not all, of the main motivations to move this module into a separate build.

jlowe (changes requested):
I'm not a fan of removing some, but not all, of the udf examples.  They should be bundled together, along with their tests, and built separately from the plugin build to ensure it is compiling and testing against the plugin dist artifact.

res-life:
build

res-life:
build

res-life:
Move the whole udf-examples module to external. 
The moved udf-examples depends on Spark Rapids repository and Spark Rapids repository should not depends on the external udf-examples. This will avoid circular dependency.

res-life:
For the spark-rapids-examples code, refer to https://github.com/NVIDIA/spark-rapids-examples/pull/94

res-life:
Convert to draft because of the build is not passed.

docs/additional-functionality/rapids-udfs.md
```diff
@@ -135,43 +135,7 @@ type `DECIMAL64(scale=-2)`.
...
-to decode URL-encoded strings
-- [URLEncode](../../udf-examples/src/main/java/com/nvidia/spark/rapids/udf/hive/URLEncode.java)
-implements a Hive generic UDF using the
-[Java APIs of RAPIDS cudf](https://docs.rapids.ai/api/cudf-java/stable)
-to URL-encode strings
-- [StringWordCount](../../udf-examples/src/main/java/com/nvidia/spark/rapids/udf/hive/StringWordCount.java)
-implements a Hive simple UDF using
-[native code](../../udf-examples/src/main/cpp/src) to count words in strings
-
+in the [udf-examples](https://github.com/NVIDIA/spark-rapids-examples/tree/branch-22.04/examples/Spark-Rapids/udf-examples) project.
```
jlowe, line 138: It's unfortunate there isn't a `main` branch in the spark-rapids-examples repository we can point to so we're always pointing to the latest stable release code.  Otherwise we're going to forget about this and then point to possibly very stale examples, and the user would miss any new examples that were added in later versions of the spark-rapids-examples code.  @GaryShen2008 should we be doing a merge to main on the spark-rapids-examples repository as we do for this one?

integration_tests/src/main/python/row-based_udf_test.py
```diff
@@ -1,43 +0,0 @@
...
-from asserts import assert_gpu_and_cpu_are_equal_sql
-from data_gen import *
-from spark_session import with_spark_session
-from rapids_udf_test import skip_if_no_hive, load_hive_udf_or_skip_test
-
-def test_hive_empty_simple_udf():
-    with_spark_session(skip_if_no_hive)
-    data_gens = [["i", int_gen], ["s", string_gen]]
-    def evalfn(spark):
-        load_hive_udf_or_skip_test(spark, "emptysimple", "com.nvidia.spark.rapids.udf.hive.EmptyHiveSimpleUDF")
```
jlowe, line 26: This file should not be deleted.  It is not testing RAPIDS accelerated UDFs but CPU UDFs accelerated by a feature built into the RAPIDS Accelerator plugin.  The degenerate Hive UDFs used for these tests should be moved to the integration_tests jar, as they are not real-world examples we would want users to emulate.

tests/src/test/scala/com/nvidia/spark/rapids/udf/URLEncode.scala
```diff
@@ -32,9 +32,9 @@ class URLEncode extends Function[String, String] with RapidsUDF with Serializabl
   override def apply(s: String): String = {
     Option(s).map { s =>
       URLEncoder.encode(s, "utf-8")
-          .replace("+", "%20")
-          .replace("*", "%2A")
-          .replace("%7E", "~")
+        .replace("+", "%20")
+        .replace("*", "%2A")
+        .replace("%7E", "~")
```
jlowe, line 37: Scala style is [4 spaces on continuation](https://github.com/NVIDIA/spark-rapids/blob/branch-22.02/docs/dev/idea-code-style-settings.xml#L35), so this should not have been changed.
```suggestion
          .replace("+", "%20")
          .replace("*", "%2A")
          .replace("%7E", "~")
```

tests/src/test/scala/com/nvidia/spark/rapids/udf/URLDecode.scala
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.nvidia.spark.rapids.udf.scala
+package com.nvidia.spark.rapids.udf
```
jlowe, line 17: I assume this is getting moved to avoid colliding with the "real" UDFs in the udf-examples module, so since we're moving, let's use a package name to make it clear these are just test UDFs and not intended for real use, e.g.: `com.nvidia.spark.rapids.test.udf`, `com.nvidia.spark.rapids.udf.test`, or something similar.

docs/additional-functionality/rapids-udfs.md
GaryShen2008: Sure, we can create a main branch for the latest release version.

res-life:
Updated the code for the rapids_build-it-UDF-native Jenkins job.
The Jenkes pipeline uses the following command to test native UDF test cases.
So although the Java and Scala code no longer depend on the external udf-examples, but the rapids_udf_test.py and run_pyspark_from_build.sh depends on the external udf-examples.
```
LOCAL_JAR_PATH=/path-to/local/jars \
bash ./integration_tests/run_pyspark_from_build.sh \
-m rapids_udf_example_native --rapids_udf_example_native

```

res-life:
build

res-life:
build

integration_tests/src/main/python/rapids_udf_test.py
```diff
@@ -12,6 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+#
+# This file depends on external udf-examples jar built from
+# https://github.com/NVIDIA/spark-rapids-examples/tree/branch-xx.xx/examples/Spark-Rapids/udf-examples.
+# By default the cases are disable, are only used by Jenkins jobs to test JNI UDF test cases.
```
jlowe, line 18: I'm not a fan of this approach, but I agree with the overall goal of regularly testing the plugin against the UDF examples.

IMO the tests for the udf-examples should be with the udf-examples, not spread across multiple repositories.  If we want the nightly and premerge plugin builds to test the udf-examples, then those builds should run the tests in the udf-examples repository after installing the plugin jar that was just built.

Otherwise if we follow the precedent of putting tests here, then the plugin repository will accumulate tests for _everything_ that might be built on top of the plugin in the future (e.g.: ML libraries, etc.).  It's cleaner to keep the tests in their respective separate repositories since they're already setup and expecting to use a plugin jar.  Only thing that needs to change is the ability to feed those tests a locally built jar, which should be solved by a simple 'mvn install' of the plugin before performing the test builds in the other repositories.

integration_tests/src/main/python/rapids_udf_test.py
```diff
@@ -21,104 +26,48 @@
...
-    def evalfn_decimal(spark):
-        load_hive_udf_or_skip_test(spark, "fraction", "com.nvidia.spark.rapids.udf.hive.DecimalFraction")
-        return gen_df(spark, [["dec", DecimalGen(38, 18)]])
-    assert_gpu_and_cpu_are_equal_sql(
-        evalfn_decimal,
-        "hive_generic_udf_test_table",
-        "SELECT fraction(dec) FROM hive_generic_udf_test_table")
+    except AnalysisException as e:
+        # rethrow this error, or test case will be skipped silently
+        raise RuntimeError("UDF {} failed to load, check if {} is in the class path, error is {}".format(udfname, udfclass, repr(e)))
```
jlowe, line 42: Curious, why not let the original exception be thrown directly?  Won't it be obvious from the original exception which UDF class cannot be found in the classpath?

res-life:
build

res-life:
build

integration_tests/src/main/python/rapids_udf_test.py
res-life: Updated

integration_tests/src/main/python/rapids_udf_test.py
res-life: Updated

res-life: Moved, spark-rapids-examples PR : https://github.com/NVIDIA/spark-rapids-examples/pull/94

docs/additional-functionality/rapids-udfs.md
```diff
@@ -134,44 +134,10 @@ type `DECIMAL64(scale=-2)`.
...
-[Java APIs of RAPIDS cudf](https://docs.rapids.ai/api/cudf-java/stable)
-to decode URL-encoded strings
-- [URLEncode](../../udf-examples/src/main/java/com/nvidia/spark/rapids/udf/hive/URLEncode.java)
-implements a Hive generic UDF using the
-[Java APIs of RAPIDS cudf](https://docs.rapids.ai/api/cudf-java/stable)
-to URL-encode strings
-- [StringWordCount](../../udf-examples/src/main/java/com/nvidia/spark/rapids/udf/hive/StringWordCount.java)
-implements a Hive simple UDF using
-[native code](../../udf-examples/src/main/cpp/src) to count words in strings
+Source code for examples of RAPIDS accelerated UDFs is provided
```
jlowe, line 138: The extra whitespace ends up disconnecting this sentence when it's rendered.
```suggestion
```

res-life:
Filed an issue to track the premerge and nightly tests: https://github.com/NVIDIA/spark-rapids/issues/4704

res-life:
build

res-life:
build

res-life:
build

res-life:
The nightly builds are not ready, convert to draft to prevent it from being accidentally merged.

res-life:
Pipelines created by Tim are ready: Rapids-examples/examples-udf-examples-nightly, Rapids-examples/examples-udf-examples-native.
Pipeline issue is: https://github.com/NVIDIA/spark-rapids/issues/4704
@jlowe  Please review

jlowe:
I mentioned in [this comment](https://github.com/NVIDIA/spark-rapids/issues/4704#issuecomment-1047903862) that we shouldn't lose premerge testing against at least some external UDFs for this repository.  Having separate nightly tests is fine, I don't care much either way as long as we get nightly coverage of the UDF examples working against updated snapshot versions of the RAPIDS Accelerator, but as it is now I think we're losing premerge tests against truly external UDFs.

jlowe:
This hasn't been built in a while, so getting a fresh build to verify it's still ready to go.

jlowe:
build

res-life:
build

PR #4622, opened 2022-01-25
Fallback to CPU when encoding is not supported for JSON reader

wbo4958:
Falllback to CPU when encoding is not UTF-8 or US-ASCII

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
@@ -16,6 +16,8 @@
 
 package org.apache.spark.sql.catalyst.json.rapids
 
+import java.nio.charset.StandardCharsets
+
 import scala.collection.JavaConverters._
 
 import ai.rapids.cudf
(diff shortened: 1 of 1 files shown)


wbo4958:
build

revans2 (approved):
This is a little late to be doing this for 22.02. But it is a simple bug fix.

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
```diff
@@ -138,6 +140,11 @@ object GpuJsonScan {
       meta.willNotWorkOnGpu("GpuJsonScan only supports \"\\n\" as a line separator")
     }
 
+    parsedOptions.encoding.foreach(enc =>
+      if (enc != StandardCharsets.UTF_8.name() && enc != StandardCharsets.US_ASCII.name()) {
+      meta.willNotWorkOnGpu("GpuJsonScan only supports UTF8 or US-ASCII encoded data")
```
andygrove, line 145: Could we have the error message include the encoding that is causing us to fall back? This will make it easier for the user to debug.

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
```diff
@@ -138,6 +140,11 @@ object GpuJsonScan {
       meta.willNotWorkOnGpu("GpuJsonScan only supports \"\\n\" as a line separator")
     }
 
+    parsedOptions.encoding.foreach(enc =>
```
andygrove, line 143: nit: could we use `exists` here instead since this is an `Option`? Something like:

```scala
  if (parsedOptions.encoding.exists(enc =>
      enc != StandardCharsets.UTF_8.name() && enc != StandardCharsets.US_ASCII.name())) {
```

wbo4958: Thx for the comment, I'd like to create a follow-up targeting 22.04

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
wbo4958: Thx for the comment, I'd like to create a follow-up targeting 22.04

PR #4623, opened 2022-01-25
Enable aggregate push down for parquet

### LINKED ISSUE #3951: [FEA] Aggregate (Min/Max/Count) push down for Parquet

HaoYang670:
close #3951 
In this PR, we:
1. Implement aggregate push down for parquet, which is a new feature of Spark 330,
2. Fallback to CPU when an aggregation is pushed down because this is a metadata scan,
3. Add tests

### DIFF
--- a/integration_tests/src/main/python/parquet_test.py
+++ b/integration_tests/src/main/python/parquet_test.py
@@ -693,7 +693,6 @@ conf_for_parquet_aggregate_pushdown = {
 }
 
 @pytest.mark.skipif(is_before_spark_330(), reason='Aggregate push down on Parquet is a new feature of Spark 330')
-@allow_non_gpu(any = True)
 def test_parquet_scan_without_aggregation_pushdown_not_fallback(spark_tmp_path):
     """
     No aggregation will be pushed down in this test, so we should not fallback to CPU
--- a/integration_tests/src/main/python/parquet_test.py
+++ b/integration_tests/src/main/python/parquet_test.py
@@ -704,11 +704,10 @@ def test_parquet_scan_without_aggregation_pushdown_not_fallback(spark_tmp_path):
         df = spark.read.parquet(data_path).selectExpr("Max(p)")
         return df
 
-    assert_cpu_and_gpu_are_equal_collect_with_capture(
+    assert_gpu_and_cpu_are_equal_collect(
         do_parquet_scan,
-        exist_classes= "GpuBatchScanExec",
-        non_exist_classes= "BatchScanExec",
-        conf= conf_for_parquet_aggregate_pushdown)
+        conf_for_parquet_aggregate_pushdown
+    )
 
 
 @pytest.mark.skipif(is_before_spark_330(), reason='Aggregate push down on Parquet is a new feature of Spark 330')
--- a/integration_tests/src/main/python/parquet_test.py
+++ b/integration_tests/src/main/python/parquet_test.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2021, NVIDIA CORPORATION.
+# Copyright (c) 2020-2022, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,11 +14,11 @@
 
 import pytest
 
-from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_fallback_collect, assert_equal
+from asserts import assert_cpu_and_gpu_are_equal_collect_with_capture, assert_gpu_and_cpu_are_equal_collect, assert_gpu_fallback_collect
 from data_gen import *
 from marks import *
 from pyspark.sql.types import *
-from spark_session import with_cpu_session, with_gpu_session
+from spark_session import with_cpu_session, with_gpu_session, is_before_spark_330
 
 def read_parquet_df(data_path):
     return lambda spark : spark.read.parquet(data_path)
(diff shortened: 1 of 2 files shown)


HaoYang670:
build

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
```diff
@@ -56,4 +60,86 @@ trait Spark33XShims extends Spark33XFileOptionsShims {
...
+      (a, conf, p, r) => new ScanMeta[ParquetScan](a, conf, p, r) {
+        override def tagSelfForGpu(): Unit = {
+          GpuParquetScanBase.tagSupport(this)
+          // we are being overly cautious and that Parquet does not support this yet
+          if (a.isInstanceOf[SupportsRuntimeFiltering]) {
+            willNotWorkOnGpu("Parquet does not support Runtime filtering (DPP)" +
+              " on datasource V2 yet.")
+          } else if (a.pushedAggregate.nonEmpty) {
+            willNotWorkOnGpu(
+              "Implementing aggregate push down on GPU gains little performance improvement"
```
revans2, line 76: nit: If it gains some then we should do it, but I don't think it would gain any, and we cannot make that statement until we have built and tested it. I would say something more like.

"Aggregates pushed into parquet read, which is a metadata only operation"

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
```diff
@@ -56,4 +60,86 @@ trait Spark33XShims extends Spark33XFileOptionsShims {
...
+            a.readDataSchema,
+            a.readPartitionSchema,
+            a.pushedFilters,
+            a.options,
+            a.partitionFilters,
+            a.dataFilters,
+            conf)
+        }
+      }),
+    GpuOverrides.scan[OrcScan](
```
revans2, line 95: Do we really need to override the Orc and CSV Scans to be able to replace the Parquet `ScanMeta`? Would it have been better to have created a `ParquetScanMeta` and placed it in the v2 shims. One for everything up to before 3.3 and one for 3.3+?

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
jlowe: Capitalization nits: the string passed to `willNotWorkOnGpu` is appended to form the sentence that will be shown to the user, so it should not normally start with a capitalized word.  Also Parquet is a proper noun, so it should be capitalized.

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
HaoYang670: I find that aggregate push down for Orc is also supported in Spark 330+, which may let us update the `GpuOrcScan` API in the future. So I copy the OrcScan to Spark33XShims.
https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2/orc/OrcScan.scala#L33-L44

HaoYang670: For `CSVScan`, I find there is a parameter `pushedFilter` in Spark, which is not added in `GpuCSVScan`. I am not sure whether we will add it in the future, so I also copy the function to 33X shims.

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
HaoYang670: Thanks for your reviews, I will update it.

HaoYang670:
build

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
```diff
@@ -56,4 +60,86 @@ trait Spark33XShims extends Spark33XFileOptionsShims {
...
+    GpuOverrides.scan[ParquetScan](
+      "Parquet parsing",
+      (a, conf, p, r) => new ScanMeta[ParquetScan](a, conf, p, r) {
+        override def tagSelfForGpu(): Unit = {
+          GpuParquetScanBase.tagSupport(this)
+          // we are being overly cautious and that Parquet does not support this yet
+          if (a.isInstanceOf[SupportsRuntimeFiltering]) {
+            willNotWorkOnGpu("Parquet does not support Runtime filtering (DPP)" +
+              " on datasource V2 yet.")
+          } else if (a.pushedAggregate.nonEmpty) {
```
wbo4958, line 74: The basic principle of tagging is to record more information about not running on GPU. So here, let's move the pushedAggregate checking out of the `else if`

HaoYang670: Thank you, I will change it.

HaoYang670:
build

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
HaoYang670: I have created the issue https://github.com/NVIDIA/spark-rapids/issues/4638

integration_tests/src/main/python/parquet_test.py
```diff
@@ -685,3 +685,48 @@ def test_parquet_reading_from_unaligned_pages_basic_filters_with_nulls(spark_tmp
...
                 all_confs)
+
+
+conf_for_parquet_aggregate_pushdown = {
+    "spark.sql.parquet.aggregatePushdown": "true", 
+    "spark.sql.sources.useV1SourceList": ""
+}
+
+@pytest.mark.skipif(is_before_spark_330(), reason='Aggregate push down on Parquet is a new feature of Spark 330')
+@allow_non_gpu(any = True)
```
jlowe, line 696: If this test is _not_ supposed to fallback, why are we allowing non-GPU nodes in the plan?

integration_tests/src/main/python/parquet_test.py
```diff
@@ -685,3 +685,48 @@ def test_parquet_reading_from_unaligned_pages_basic_filters_with_nulls(spark_tmp
...
+    def do_parquet_scan(spark):
+        spark.range(10).selectExpr("id", "id % 3 as p").write.partitionBy("p").mode("overwrite").parquet(data_path)
+        df = spark.read.parquet(data_path).selectExpr("Max(p)")
+        return df
+
+    assert_cpu_and_gpu_are_equal_collect_with_capture(
+        do_parquet_scan,
+        exist_classes= "GpuBatchScanExec",
+        non_exist_classes= "BatchScanExec",
+        conf= conf_for_parquet_aggregate_pushdown)
```
jlowe, line 712: This should just be `assert_cpu_and_gpu_are_equal_collect`.  We're honestly not that interested in exactly which GPU nodes are involved here, rather that the plan is all on the GPU.  That's what `assert_cpu_and_gpu_are_equal_collect` already checks, once we remove the `@allow_non_gpu` decorator.

HaoYang670:
build

integration_tests/src/main/python/parquet_test.py
HaoYang670: Thank you! I have updated

jlowe (approved):

Issue #4538, opened 2022-01-14
Premerge build takes too long again (over 3 hours)


tgravescs:
**Describe the bug**
The premerge build for pull requests takes over 3 hours again, it would be nice if it took less time.

investigate why its taking so long.
Perhaps we can drop support for some versions?  Spark should be releasing some new minor versions soon.

pxLi:
For most spark versions in pre-merge we only do build w/o test, so they should not cost too much time.

IMO most time comes from new tests/cases we added to integration tests, may be we can shrink the data_gen size in pre-merge only?

jlowe:
We've added a lot of tests that have tons of dimensions, and that adds a ton of tests to run.  For example, some of the slower tests are joins, and those test across the dimensions of join types, data types, and batch sizes (if not even more dimensions in some specific cases).  We may want to consider if some of these test combinations are not worth the cost it is taking to run them.  For example, do we really need to test every join type and every data type for a batch size of 100 (which is quite slow to run), or is it sufficient just to test one or just a handful of join types and data types for that extreme configuration?

Executive summary is that I think we need to audit which test files and individual tests are taking the longest and then see if we really need to run all the dimensions on those tests to get a satisfactory level of coverage from them.

revans2:
I think that there are a few things happening.  We have some tests that are slow because we are doing exhaustive combinations.  We have other tests that are slow because we are doing a lot of computation. I totally agree that we probably don't need all combinations. We probably also don't need to cover as much data in the pre-merge which might reduce the amount of computation. But before we get into any of this we probably want to measure to see what each problem really is instead of just guessing.

Do we have stats somewhere about how long each test is taking? It looks like we can add `--durations=N` to the command line when running the tests and it will tell us the top `N` longest running tests, or 0 if we want to see all of them, but with a min time cutoff that is set by `--durations-min`. So we could get an idea about what is actually happening, and it can let us also test on a subset of the tests to see if our efforts would help without having to run all 3 hours of tests each time we want to make a change and see if it helped.

My guess is that both of you are right.  We have some tests that we know are very long running and we can probably turn off some versions of them without much hit to coverage. There are probably also a lot of really small tests. I am not sure if we drop the size of the data how big of an impact that will have. 

[...]

PR #4624, opened 2022-01-25
Enable integration test results report on Jenkins [skip ci]

zhanga5:

The test result would be collected by using [JUnit plugin](https://plugins.jenkins.io/junit) on Jenkins for helping evaluate individual test case duration, e.g, https://github.com/NVIDIA/spark-rapids/issues/4538#issuecomment-1017154557

In case of any test case was failed, the error message or stacktrace would be collected for further troubleshooting, e.g, http://nv/2sU (need vpn)

image

### DIFF
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -154,6 +154,7 @@ else
           $TEST_TYPE_PARAM
           "$TEST_ARGS"
           $RUN_TEST_PARAMS
+          --junitxml=TEST-pytest-`date +%s%N`.xml
           "$@")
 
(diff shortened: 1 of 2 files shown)


jlowe (approved):
Changes look OK, but I noticed that this is linked to close #4538 which seems incorrect.  This doesn't make the integration tests run faster, just gathers reporting on it.

pxLi (approved):

pxLi:
build

zhanga5:
> Changes look OK, but I noticed that this is linked to close #4538 which seems incorrect. This doesn't make the integration tests run faster, just gathers reporting on it.

agree. Removed now

PR #4629, opened 2022-01-26
Remove shims module [databricks]

razajafri:
This is the first commit in a series of commit. The goal is to get rid of v1 shims usage and replace it with v2 usage 

* Source files from shims module were moved to sql-plugin module
* Test files were moved to test folder inside sql-plugin module
* Added the ability to select test files based on profile for testing shims
* Moved README.md from shims module to docs/dev/ and renamed it to shims.md.
* Modified the relative links in the doc to match the new location


razajafri:
build

razajafri:
build

docs/dev/shims.md
```diff
@@ -16,7 +16,7 @@ common code, maximize reuse, and minimize logic duplication.
 This is achieved by using a ServiceProvider pattern. All Shims implement the same API, 
 the suitable Shim implementation is loaded after detecting the current Spark build version
 attempting to instantiate our plugin. We use the
-[ShimLoader](../sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala)
+[ShimLoader](../../sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala)
```
jlowe, line 19: These pages are published to Github Pages, and this relative link will not work once published to that site.  We may want to point to this class on the `main` branch using an absolute link to the code, e.g.: https://github.com/NVIDIA/spark-rapids/blob/main/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala

This applies to all relative links to code in this document.

pom.xml
```diff
@@ -114,6 +114,7 @@
                                 <configuration>
                                     <sources>
                                         <source>${project.basedir}/src/main/301+-nondb/scala</source>
+                                        <source>${project.basedir}/src/main/shim301/scala</source>
```
jlowe, line 117: Nit: Shouldn't this just be `301` to follow the convention set by the other directories?  Similar comment for the other `shim*` directories that were added.

sql-plugin/src/main/shim301/scala/com/nvidia/spark/rapids/shims/spark301/RapidsShuffleInternalManager.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2022, NVIDIA CORPORATION.
```
jlowe, line 2: Copyright update on a file whose contents were not modified.  Comment applies to many other files.

razajafri: Are we OK with moving files without changing the copyright? I can revert this change.

pom.xml
razajafri: I named them this way so reviewers can clearly distinguish between files that were moved from the shims module. These folders will go away and everything will be merged into the existing folders i.e. `301db`, `301until310-all` etc

sql-plugin/src/main/shim301/scala/com/nvidia/spark/rapids/shims/spark301/RapidsShuffleInternalManager.scala
jlowe: The copyright is on the contents of the file, not the name of the file.

razajafri: Done

sql-plugin/src/main/shim301db/scala/com/nvidia/spark/rapids/shims/spark301db/RapidsShuffleInternalManager.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, NVIDIA CORPORATION.
+ * Copyright (c) 2021-2021, NVIDIA CORPORATION.
```
razajafri, line 2: ugh! let me fix this

pom.xml
jlowe: > These folders will go away

I still think these should be named appropriately.  It's 301db, not shim301db, for example.  Someone could decide to use these directories for a file unrelated to this shim refactoring work that requires compiling only on a very specific version.  If that happens then the folder will not go away, and we'll wonder why the naming was not consistent.

razajafri: gotcha! done

razajafri:
build

sql-plugin/src/main/303/scala/com/nvidia/spark/rapids/shims/spark303/RapidsShuffleInternalManager.scala
```diff
@@ -77,4 +77,4 @@ class ProxyRapidsShuffleInternalManager(conf: SparkConf, isDriver: Boolean)
     self.getReaderForRange(handle, startMapIndex, endMapIndex, startPartition, endPartition,
       context, metrics)
   }
-}
\ No newline at end of file
+}
```
jlowe, line 80: File shouldn't be modified, otherwise we need copyrights.  Yay, copyrights!  Applies to some other files as well that had trivial end-of-file whitespace changes.

razajafri: I don't know how to get rid of this. any ideas?

razajafri: wow this was trickier than I thought. Even after removing the eof, github wouldn't accept it.

I had to change the base branch and revert back to branch-22.04 base for it to work

razajafri:
build

jlowe (dismissed):

razajafri:
CI failed due to `[ERROR] Failed to execute goal on project rapids-4-spark-udf_2.12: Could not resolve dependencies for project com.nvidia:rapids-4-spark-udf_2.12:jar:22.04.0-SNAPSHOT: Could not find artifact com.nvidia:rapids-4-spark-shims-spark312db_2.12:jar:22.04.0-SNAPSHOT in snapshots-repo (https://oss.sonatype.org/content/repositories/snapshots) -> [Help 1]`

Are we missing some jars for 22.04 for db?

pxLi:
> CI failed due to `[ERROR] Failed to execute goal on project rapids-4-spark-udf_2.12: Could not resolve dependencies for project com.nvidia:rapids-4-spark-udf_2.12:jar:22.04.0-SNAPSHOT: Could not find artifact com.nvidia:rapids-4-spark-shims-spark312db_2.12:jar:22.04.0-SNAPSHOT in snapshots-repo (https://oss.sonatype.org/content/repositories/snapshots) -> [Help 1]`
> 
> Are we missing some jars for 22.04 for db?

nope, the rapids-4-spark-shims-spark301db&rapids-4-spark-shims-spark312db_2.12 should be built with in this CI (but this PR remove the 301db and 312db shims), and we never deploy nightly shims to sonatype snapshot repo.

jlowe:
Ah, right, I totally missed in the review that a number of other poms (aggregator, tests, etc.) need to be updated, as they explicitly depend on the shims artifact being removed.

@razajafri you can build in a clean repo, either by cleaning out your local ~/.m2 or using something like `-Dmaven.repo.local=/tmp/m2temprepo` to reproduce these types of errors locally.

razajafri:
> Ah, right, I totally missed in the review that a number of other poms (aggregator, tests, etc.) need to be updated, as they explicitly depend on the shims artifact being removed.
> 
> @razajafri you can build in a clean repo, either by cleaning out your local ~/.m2 or using something like `-Dmaven.repo.local=/tmp/m2temprepo` to reproduce these types of errors locally.

aah! of course I should've removed the dependency from other modules but I don't get how the CI was passing until I built on databricks. Does the CI reuse previously used maven repo?

razajafri:
build

sql-plugin/pom.xml
```diff
@@ -393,6 +393,32 @@
...
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>net.alchim31.maven</groupId>
+                <artifactId>scala-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.scalatest</groupId>
+                <artifactId>scalatest-maven-plugin</artifactId>
+            </plugin>
```
jlowe, line 403: These are already specified a few lines earlier, not sure why they were added again?

sql-plugin/pom.xml
```diff
@@ -393,6 +393,32 @@
...
+            <plugin>
+                <artifactId>maven-antrun-plugin</artifactId>
```
jlowe, line 405: There's already an antrun plugin section in this pom, so this should be collected there to reduce boilerplate and keep the usage of a particular plugin centralized.

sql-plugin/pom.xml
razajafri: copy paste error

sql-plugin/pom.xml
razajafri: merged with the one above it

razajafri:
build

jlowe (approved):

tgravescs:
build

Issue #4003, opened 2021-11-02
[FEA] Add regular expression support to GPU implementation of StringSplit


andygrove:
**Is your feature request related to a problem? Please describe.**
We currently implement StringSplit but only support non-regexp patterns.

**Describe the solution you'd like**
We should support regular expressions in the split pattern, compatible with Spark.

**Describe alternatives you've considered**
None

**Additional context**
None

andygrove:
Depends on https://github.com/rapidsai/cudf/issues/9862

ttnghia:
Depends on https://github.com/rapidsai/cudf/pull/10128 and https://github.com/rapidsai/cudf/pull/10139.

sameerz:
Example from https://github.com/NVIDIA/spark-rapids/issues/4658#issue-1117811691 

> I wish we support split with regular expressions.
> 
> Mini repro:
> 
> val address = Seq((1,"abc.com"),
> (2,"...abc"),
> (3,".a.b.c"))
> 
> import spark.implicits._
> val df = address.toDF("id","txt")
> df.write.mode("overwrite").format("parquet").save("/tmp/testparquet")
> val df2=spark.read.parquet("/tmp/testparquet")
> df2.createOrReplaceTempView("df2")
> spark.sql("select id,txt, split(txt, '\\\\.')[0] AS new_txt from df2").show()
> 
> Not-supported-messages:
> 
> !Expression <StringSplit> split(txt#209, \., -1) cannot run on GPU because regular expressions are not supported yet
>

PR #4636, opened 2022-01-26
Support `str_to_map`

ttnghia:
This adds supports for `str_to_map` SQL function.

Note that fully support for this function requires the ability to split string by regular expression delimiter. As such, this PR depends on https://github.com/NVIDIA/spark-rapids/issues/4003, which also depends on https://github.com/rapidsai/cudf/pull/10128 and https://github.com/rapidsai/cudf/pull/10139.

If the dependencies cannot be resolved upon releasing of 24.04, we will just merge this with basic support (i.e., string split with only literal delimiters).

 - [ ] Implementation for basic support
 - [ ] Implementation with regex support
 - [ ] Integration tests

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -2648,6 +2648,13 @@ object GpuOverrides extends Logging {
         override def convertToGpu(child: Expression): GpuExpression =
           GpuMapEntries(child)
       }),
+    expr[StringToMap](
+      "Creates a map after splitting the input into key-value pairs using delimiters",
+      ExprChecks.projectOnly(TypeSig.STRING, TypeSig.STRING,
+        Seq(ParamCheck("str", TypeSig.STRING, TypeSig.STRING),
+          ParamCheck("pairDelim", TypeSig.lit(TypeEnum.STRING), TypeSig.STRING),
+          ParamCheck("keyValueDelim", TypeSig.lit(TypeEnum.STRING), TypeSig.STRING))),
+      (in, conf, p, r) => new GpuStringToMapMeta(in, conf, p, r)),
     expr[ArrayMin](
       "Returns the minimum value in the array",
       ExprChecks.unaryProject(
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -2650,7 +2650,7 @@ object GpuOverrides extends Logging {
       }),
     expr[StringToMap](
       "Creates a map after splitting the input into key-value pairs using delimiters",
-      ExprChecks.projectOnly(TypeSig.STRING, TypeSig.STRING,
+      ExprChecks.projectOnly(TypeSig.MAP.nested(TypeSig.STRING), TypeSig.MAP.nested(TypeSig.STRING),
         Seq(ParamCheck("str", TypeSig.STRING, TypeSig.STRING),
           ParamCheck("pairDelim", TypeSig.lit(TypeEnum.STRING), TypeSig.STRING),
           ParamCheck("keyValueDelim", TypeSig.lit(TypeEnum.STRING), TypeSig.STRING))),
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
@@ -17,7 +17,7 @@
 package org.apache.spark.sql.rapids
 
 import ai.rapids.cudf.{ColumnVector, DType}
-import com.nvidia.spark.rapids.{GpuColumnVector, GpuExpression, GpuExpressionsUtils, GpuMapUtils}
+import com.nvidia.spark.rapids.{Arm, GpuColumnVector, GpuExpression, GpuExpressionsUtils, GpuMapUtils}
 import com.nvidia.spark.rapids.RapidsPluginImplicits.AutoCloseableProducingSeq
 import com.nvidia.spark.rapids.shims.v2.ShimExpression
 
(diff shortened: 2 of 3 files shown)


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1379,3 +1379,107 @@ case class GpuStringSplit(str: Expression, regex: Expression, limit: Expression)
...
+case class GpuStringToMap(strs: Expression, pairDelim: Expression, keyValueDelim: Expression)
+    extends GpuExpression with ExpectsInputTypes with NullIntolerant {
+
+  def this(str: Expression, pairDelim: Expression) = {
+    this(str, pairDelim, GpuLiteral(":", StringType))
+  }
+
+  def this(str: Expression) = {
+    this(str, GpuLiteral(",", StringType), GpuLiteral(":", StringType))
+  }
```
ttnghia, line 1432: Not sure if these constructors are needed, because they are never be called.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1379,3 +1379,107 @@ case class GpuStringSplit(str: Expression, regex: Expression, limit: Expression)
...
+
+  private def toMap(strs: GpuColumnVector,
+                    pairDelim: GpuScalar,
+                    keyValueDelim: GpuScalar): GpuColumnVector = {
+    // Firstly, split the input strings into lists of strings.
+    withResource(strs.getBase.stringSplitRecord(pairDelim.getBase)) { listsOfStrings =>
+      // Extract strings column from the output lists column.
+      withResource(listsOfStrings.getChildColumnView(0)) { stringsCol =>
+        // Split the key-value strings into pairs of strings of key-value (using maxSplit = 1).
+        // withResource(stringsCol.stringSplit(keyValueDelim.getBase, 1)) { keyValuePairs =>
```
ttnghia, line 1468: This will be enabled after cudf snapshot has been updated to accept the `maxSplit` parameter.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2648,6 +2648,13 @@ object GpuOverrides extends Logging {
         override def convertToGpu(child: Expression): GpuExpression =
           GpuMapEntries(child)
       }),
+    expr[StringToMap](
+      "Creates a map after splitting the input into key-value pairs using delimiters",
+      ExprChecks.projectOnly(TypeSig.MAP.nested(TypeSig.STRING), TypeSig.MAP.nested(TypeSig.STRING),
+        Seq(ParamCheck("str", TypeSig.STRING, TypeSig.STRING),
+          ParamCheck("pairDelim", TypeSig.lit(TypeEnum.STRING), TypeSig.STRING),
```
revans2, line 2655: nit: Actually it is a little hard to tell from the Spark code but it is also verifying that only literal values are used for the delims.

https://github.com/apache/spark/blob/4920014cc3a14019a2584c1059bc3d7b75426635/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/complexTypeCreator.scala#L552

so you can have the Spark once marked as `TypeSig.lit(STRING)` also. This will let our docs be a bit more accurate.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeCreator.scala
```diff
@@ -16,8 +16,8 @@
 
 package org.apache.spark.sql.rapids
```
revans2, line 17: 2022 copyright

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1379,3 +1379,107 @@ case class GpuStringSplit(str: Expression, regex: Expression, limit: Expression)
...
+    rule: DataFromReplacementRule)
+    extends TernaryExprMeta[StringToMap](expr, conf, parent, rule) {
+  import GpuOverrides._
+
+  private def checkRegex(delimExpr: Expression) : Unit = {
+    extractLit(delimExpr).foreach { delim =>
+      val str = delim.value.asInstanceOf[UTF8String]
+      if (str != null){
+        //
+        // Currently, we don't support regex. However, we can't check for regex very accurately.
```
revans2, line 1396: I personally would check for the default string delims and add them to an allow list. That way it will still work by default and we can at least check it in with everything turned on, then in a follow on issue we can add in support for regular expressions.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
revans2: They are not needed. Some times it is nice to match what Spark does, but here it is just dead code.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1379,3 +1379,107 @@ case class GpuStringSplit(str: Expression, regex: Expression, limit: Expression)
...
+
+  override def dataType: MapType = MapType(StringType, StringType)
+  override def inputTypes: Seq[AbstractDataType] = Seq(StringType, StringType, StringType)
+  override def prettyName: String = "str_to_map"
+  override def children: Seq[Expression] = Seq(strs, pairDelim, keyValueDelim)
+  override def nullable: Boolean = children.head.nullable
+  override def foldable: Boolean = children.forall(_.foldable)
+
+  override def columnarEval(batch: ColumnarBatch): Any = {
+    withResourceIfAllowed(strs.columnarEval(batch)) { val0 =>
```
revans2, line 1442: nit: can we have more descriptive names than val0, val1, etc.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1298,9 +1298,9 @@ class GpuStringSplitMeta(
       if (str != null) {
         if (RegexParser.isRegExpString(str.toString)) {
           willNotWorkOnGpu("regular expressions are not supported yet")
-        }
-        if (str.numChars() == 0) {
-          willNotWorkOnGpu("An empty regex is not supported yet")
+          if (str.numChars() == 0) {
```
revans2, line 1301: Why did this change? Do we have tests for split on empty String that will make sure that this is covered?

revans2 (commented):
If you want this to go in as two separate PRs. One that can go in right now and another follow on PR to match the changes you are making in CUDF I am fine with that. But if the CUDF changes are going to be breaking changes we need a PR that is 100% ready to go that is based off of the changes to CUDF.  Just waiting for CI to complete means that we are likely going to have to wait at least 4 hours from the time that the CUDF change goes in to the time that the plugin build is no longer broken. If the plugin PR is not 100% ready to go then the plugin build will be broken for much longer.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
ttnghia: This is a bug. String splitting can be done with an empty delimiter. In addition, an empty string should not be a regex-it is just a literal string.
According to string splitting API, empty delimiter means splitting by whitespace.

ttnghia: I just checked again and realized that the splitting behavior of libcudf and Java for empty delimiter is inconsistent. For empty delimiter, Java splits the input into an array of individual chars while libcudf splits by whitespace.

Then we indeed can't support empty delimiter.

ttnghia: I just filed an issue for it: https://github.com/rapidsai/cudf/issues/10251

ttnghia:
build

ttnghia:
@revans2 @andygrove Should we test build this with [databrick]?

ttnghia:
build

revans2 (approved):

ttnghia:
build

ttnghia:
build

revans2 (dismissed):

ttnghia:
build

revans2 (approved):

Issue #124, opened 2020-06-08
[BUG] CSV/JSON Parsing some float values results in overflow


revans2:
**Describe the bug**
Parsing `-3.4028234663852886e+38` as a  float32 results in `-Inf`.

**Steps/Code to reproduce bug**
There is an integration test added for this.

**Expected behavior**
It should parse to the corresponding supported value.

wbo4958:
Json reader also has the same issue. Changed the title

Issue #1986, opened 2021-03-22
[BUG] CSV reading null inconsistent between spark.rapids.sql.format.csv.enabled=true&false


mattf:
**Describe the bug**

input...
```
empty-0space,,end
empty-0space-quoted,"",end
empty-1space, ,end
empty-1space-quoted," ",end
empty-2space,  ,end
empty-2space-quoted,"  ",end
no-null,3.14,end
```
schema...
```
StructType([StructField("firstField", StringType()),
            StructField("dblField", DoubleType(), True),
            StructField("lastField", StringType())])
```

config spark.rapids.sql.format.csv.enabled = true ...
```
+-------------------+--------+---------+
|         firstField|dblField|lastField|
+-------------------+--------+---------+
|       empty-0space|    null|      end|
|empty-0space-quoted|    null|      end|
|       empty-1space|     0.0|      end|
|empty-1space-quoted|    null|      end|
|       empty-2space|     0.0|      end|
|empty-2space-quoted|    null|      end|
|            no-null|    3.14|      end|
+-------------------+--------+---------+
```

config spark.rapids.sql.format.csv.enabled = false ...
```
+-------------------+--------+---------+
|         firstField|dblField|lastField|
+-------------------+--------+---------+
|       empty-0space|    null|      end|
|empty-0space-quoted|    null|      end|
|       empty-1space|    null|      end|
|empty-1space-quoted|    null|      end|
|       empty-2space|    null|      end|
|empty-2space-quoted|    null|      end|
|            no-null|    3.14|      end|
+-------------------+--------+---------+
```

**Expected behavior**
results should be the same

**Environment details (please complete the following information)**
 - Environment location: local
 - Spark configuration settings related to the issue: 0.4 w/ libcudf 0.18.1

mattf:
```
from pyspark.sql import SparkSession
from pyspark.sql.types import StructType, StructField, StringType, DoubleType

INPUT = \
"""
empty-0space,,end
empty-0space-quoted,"",end
empty-1space, ,end
empty-1space-quoted," ",end
empty-2space,  ,end
empty-2space-quoted,"  ",end
no-null,3.14,end
"""

with open("repro-data.csv", "w") as fp:
    fp.write(INPUT)

schema = StructType([StructField("firstField", StringType()), StructField("dblField", DoubleType(), True), StructField("lastField", StringType())])

[...]

Issue #126, opened 2020-06-08
[BUG] CSV parsing large number values overflow


revans2:
**Describe the bug**
When parsing large float number values in CSV they overflow to 'inf' or '-inf' instead of returning null as the CPU does.

For int numbers it is the same but the value literally overflows instead of becoming Inf or -Inf.

**Steps/Code to reproduce bug**
have a CSV file with a value too large to fit in the type being used.

Issue #125, opened 2020-06-08
[BUG] CSV Parseing Inf, -Inf, and NaN not supported


revans2:
**Describe the bug**
Parsing these values results in a null.

**Steps/Code to reproduce bug**
There are integration tests added .

**Expected behavior**
Ideally we would be able to support `positiveInf`, `negativeInf`, and `nanValue` CSV parser options.

revans2:
Int and -Inf are now partially supported, but they are not configurable.

wbo4958:
JSON also can't parse NaN.

andygrove:
I filed https://github.com/NVIDIA/spark-rapids/issues/4644 for supporting `positiveInf`, `negativeInf`, and `nanValue` CSV parser options to reduce the scope of this issue.

PR #4637, opened 2022-01-26
Improve support for reading CSV and JSON floating-point values

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/124 and partially addresses https://github.com/NVIDIA/spark-rapids/issues/125, https://github.com/NVIDIA/spark-rapids/issues/126, and https://github.com/NVIDIA/spark-rapids/issues/1986 (only for floating-point types)

This PR affects both CSV and JSON readers and reads floating-point columns as strings and then casts them to a floating-point type. This means that we can now support special values such as NaN and Inf more consistently with Spark.

If this approach is acceptable then I will follow up with PRs to do this for other data types.

Note that this isn't perfect and there are still some follow-on issues (some existing and some new):

- https://github.com/NVIDIA/spark-rapids/issues/125
- https://github.com/NVIDIA/spark-rapids/issues/4644
- https://github.com/NVIDIA/spark-rapids/issues/4647
- https://github.com/NVIDIA/spark-rapids/issues/4646

Out of these, issue 4647 bothers me the most.

Status:

- [x] Implement basic functionality
- [x] Enable related tests that were previously XFAIL
- [x] Fix resource leaks
- [x] Add new tests for better coverage of edge cases
- [x] Add support (and tests) for JSON option `allowNonNumericNumbers`
- [x] Add tests with ansi mode enabled for JSON and CSV with invalid inputs


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTextBasedPartitionReader.scala
```diff
@@ -175,7 +189,25 @@ abstract class GpuTextBasedPartitionReader(
...
+        // parse floating-point columns that were read as strings
+        val castTable = withResource(table) { _ =>
+          val columns = new ListBuffer[ColumnVector]()
+          val ansiEnabled = SQLConf.get.ansiEnabled
+          for (i <- 0 until table.getNumberOfColumns) {
+            val castColumn = dataSchema.fields(i).dataType match {
+              case DataTypes.FloatType =>
+                GpuCast.castStringToFloats(table.getColumn(i), ansiEnabled, DType.FLOAT32)
+              case DataTypes.DoubleType =>
+                GpuCast.castStringToFloats(table.getColumn(i), ansiEnabled, DType.FLOAT64)
```
jlowe, line 201: These columns are never closed.  `Table` increments the refcount, so they need to be closed after being added to the table.  Probably can use `withResource` on an `ArrayBuffer[ColumnVector]` for this.

revans2 (commented):
This is a really great step. Ultimately I would like to use this same trick to support other types, like decimal and date/time. But I am not convinced that we want to make it common everywhere because JSON has some odd configs/requirements with how it parses some numbers, which will be different from CSV and from casting. But these are corner cases and it might not matter that much anyways.

On a side note @GaryShen2008 we need to make sure that we take this into account when we are gathering requirements for JSON parsing in CUDF. We already know that they are not going to want to provide all of the parsing options that Spark supports, so we probably want to instead work on supporting the parsing ourselves as much as we can.

integration_tests/src/main/python/csv_test.py
```diff
@@ -224,19 +224,23 @@ def read_impl(spark):
...
+    pytest.param('simple_float_values.csv', _double_schema, {'header': 'true'}, marks=pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/126')),
     pytest.param('simple_boolean_values.csv', _bool_schema, {'header': 'true'}, marks=pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/2071')),
     pytest.param('ints_with_whitespace.csv', _number_as_string_schema, {'header': 'true'}, marks=pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/2069')),
     pytest.param('ints_with_whitespace.csv', _byte_schema, {'header': 'true'}, marks=pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/130'))
     ], ids=idfn)
 @pytest.mark.parametrize('read_func', [read_csv_df, read_csv_sql])
 @pytest.mark.parametrize('v1_enabled_list', ["", "csv"])
-def test_basic_read(std_input_path, name, schema, options, read_func, v1_enabled_list):
-    updated_conf=copy_and_update(_enable_all_types_conf, {'spark.sql.sources.useV1SourceList': v1_enabled_list})
+@pytest.mark.parametrize('ansi_enabled', ["true", "false"])
```
revans2, line 238: I was hoping for a test that had bad data and would trigger an exception in ANSI mode, because from reading the CSV code they do not care about ANSI mode at all and will not throw an exception for data like this.

andygrove: I've just added a new `floats_invalid.json` so that we cover some invalid inputs in this test.

We now always specify `ansiEnabled=false` in the CSV/JSON code that is calling `CAST` so that we return null on bad data rather than throwing an exception, to match Spark's current behavior (at least, in the versions of Spark that I have tested so far).

revans2: okay, but some invalid CSV tests would be good to, just to verify that path also.

docs/compatibility.md
```diff
@@ -377,13 +377,6 @@ date. Typically, one that overflowed.
...
 ### CSV Floating Point
 
-The CSV parser is not able to parse `NaN` values.  These are
-likely to be turned into null values, as described in this
-[issue](https://github.com/NVIDIA/spark-rapids/issues/125).
-
-Some floating-point values also appear to overflow but do not for the CPU as described in this
-[issue](https://github.com/NVIDIA/spark-rapids/issues/124).
-
 Any number that overflows will not be turned into a null value.
```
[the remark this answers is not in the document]
andygrove: I've removed this and added the disclaimer that we have the same issues as casting from float to string. I have enabled tests for overflow but had to remove the NaN / Inf values from `simple_floats_values.csv` but I think that is ok since we already have those values in `nans_and_inf.csv` which is covered by the same test but currently XFAILs due to known issues.

integration_tests/src/test/resources/simple_float_values.csv
```diff
@@ -16,12 +16,4 @@ bad
...
 1.7976931348623158E308
-1.2e-234
-NAN
-nan
-NaN
-Inf
--Inf
-INF
--INF
-
```
andygrove, line 27: Nan and Inf values are already covered in nan_and_inf.csv

revans2 (dismissed):

andygrove:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTextBasedPartitionReader.scala
```diff
@@ -175,7 +188,30 @@ abstract class GpuTextBasedPartitionReader(
...
+              case _ =>
+                table.getColumn(i).incRefCount()
+            }
+            columns += castColumn
+          }
+          // Table increases the ref counts on the columns so we have
+          // to close them after creating the table
+          withResource(columns) { _ =>
```
jlowe, line 209: This `withResource` should include the `for` loop, otherwise we leak columns if we've built some, but not all, of the columns before something throws.

jlowe (approved):

andygrove:
build

andygrove:
There are test failures in `HashAggregatesSuite` when running against Spark 3.1.1 - I am investigating.

andygrove:
build

tests/src/test/scala/com/nvidia/spark/rapids/HashAggregatesSuite.scala
```diff
@@ -637,6 +638,7 @@ class HashAggregatesSuite extends SparkQueryCompareTestSuite {
   FLOAT_TEST_testSparkResultsAreEqual(
       "doubles basic aggregates group by doubles",
       doubleCsvDf,
+      maxFloatDiff = 0.0001,
```
jlowe, line 641: Aggregation ordering isn't deterministic, so I understand why this change could be necessary, but I'm not sure why we weren't seeing this before.  Are we sure the exact same values from the CSV file are being loaded as before this PR and this change to the test isn't masking another problem?

andygrove: For some reason, this seems to happen pretty reliably with Spark 3.1.1 (I had previously only tested locally with 3.0.1 and 3.2.0). Here is a subset of the results showing the difference between CPU and GPU.

## CPU

```
WrappedArray(-100.0, 456.0, 23.0, 129.0, -600.0, -106.0, 106.0, -94.0, -100.0, 1), 
WrappedArray(100.0, 456.0, 223.0, 124.0, 100.0, 99.0, -99.0, 101.0, 100.0, 1), 
WrappedArray(200.0, 456.0, 323.0, 125.0, 400.0, 198.0, -198.0, 202.0, 200.0, 1), 
WrappedArray(300.0, 456.0, 423.0, 126.0, 900.0, 297.0, -297.0, 303.0, 300.0, 1), 
WrappedArray(400.0, 456.0, 523.0, 127.0, 1600.0, 396.0, -396.0, 404.0, 400.0, 1), 
```

## GPU

```
WrappedArray(-100.0, 456.0, 23.0, 129.0, -599.9999999999999, -106.0, 106.0, -94.0, -100.0, 1), 
WrappedArray(100.0, 456.0, 223.0, 124.0, 99.99999999999999, 99.0, -99.0, 101.0, 100.0, 1), 
WrappedArray(200.0, 456.0, 323.0, 125.0, 399.99999999999994, 198.0, -198.0, 202.0, 200.0, 1), 
WrappedArray(300.0, 456.0, 423.0, 126.0, 899.9999999999999, 297.0, -297.0, 303.0, 300.0, 1), 
WrappedArray(400.0, 456.0, 523.0, 127.0, 1599.9999999999998, 396.0, -396.0, 404.0, 400.0, 1),
```

andygrove: No, this is definitely related to this PR. The affected column is `max(col("doubles") * col("more_doubles"))` ... I will dig into this more and see if I can determine why this is only causing issues with Spark 3.1.1

jlowe: I think the first thing to check is what I asked earlier:
> Are we sure the exact same values from the CSV file are being loaded as before this PR

revans2: > Are we sure the exact same values from the CSV file are being loaded as before this PR

We know that they will not be the same.  We know that CUDF is not parsing them the same as Spark is, and we also taking steps to change how we parse the numbers when casting to try to make it closer to Spark. Until we have code to parse it exactly the same as Spark, I don't know how much we care except that it is close enough.

We had a huge discussion earlier about what close enough means for floating point.  In the pytest code we typically go with the pytest default relative tolerance of `1e-6`. Or we can be off by one part per million.

I am much more interested to understand if both we and Spark are producing the same results each time? And if Spark is changing (because it looks like 3.1.1 is different from the others) why is it changing?

andygrove: So it turns out that this is not specific to 3.1.1 (I saw the same failure with 3.0.1) and is an intermittent issue, presumably dependent on the order that accumulations are happening during the aggregate. I added debug logging and confirmed that the values read from CSV and the resulting CAST values are the same for 3.0.1 and 3.1.1.

andygrove: This PR was originally motivated for JSON support (especially for handling of out-of-range numbers and NaN/Inf parsing) but also impacts CSV since this is common code. Technically, this is a regression in this case because of the small difference between native cuDF parsing vs casting from string, but there are other cases where these changes make CSV support more compatible with Spark. We could consider putting this code being a configuration option, or just having this change apply to JSON and not CSV if this is a concern.

jlowe: Yes, ultimately this will be much better for some values and slightly worse for others.  It is quite odd that cudf is parsing CSV floating point values directly differently than casting the strings as floating point.

Once the tolerances are lowered to match pytest, I'm good to go on this.

andygrove:
build

jlowe (approved):

PR #4652, opened 2022-01-28
Remove non-deterministic test order in nightly [skip ci]

pxLi:

remove the `shuf` to avoid non-deterministic test order. (was added to discover if any better ordering in random runs)

tested locally


pxLi:
build

GaryShen2008 (approved):

Issue #4664, opened 2022-01-31
[BUG] MortgageAdaptiveSparkSuite failed with duplicate buffer exception


jlowe:
Nightly tests failed on Spark 3.0.2 with this error:
```
[2022-01-29T13:14:16.578Z]   org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 5.0 failed 1 times, most recent failure: Lost task 0.0 in stage 5.0 (TID 6, nightly-test-339-9q329-fq4cf, executor driver): com.nvidia.spark.rapids.DuplicateBufferException: Buffer ID ShuffleBufferId(shuffle_2_6_0,2) at tier device memory already registered device memory buffer size=273792
[2022-01-29T13:14:16.578Z] 	at com.nvidia.spark.rapids.RapidsBufferCatalog$$anon$1.apply(RapidsBufferCatalog.scala:120)
[2022-01-29T13:14:16.578Z] 	at com.nvidia.spark.rapids.RapidsBufferCatalog$$anon$1.apply(RapidsBufferCatalog.scala:111)
[2022-01-29T13:14:16.578Z] 	at java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1877)
[2022-01-29T13:14:16.578Z] 	at com.nvidia.spark.rapids.RapidsBufferCatalog.registerNewBuffer(RapidsBufferCatalog.scala:126)
[2022-01-29T13:14:16.578Z] 	at com.nvidia.spark.rapids.RapidsBufferStore.addBuffer(RapidsBufferStore.scala:218)
[2022-01-29T13:14:16.578Z] 	at com.nvidia.spark.rapids.RapidsDeviceMemoryStore.$anonfun$addContiguousTable$1(RapidsDeviceMemoryStore.scala:117)
[2022-01-29T13:14:16.578Z] 	at com.nvidia.spark.rapids.RapidsDeviceMemoryStore.$anonfun$addContiguousTable$1$adapted(RapidsDeviceMemoryStore.scala:113)
[2022-01-29T13:14:16.578Z] 	at com.nvidia.spark.rapids.Arm.freeOnExcept(Arm.scala:142)
[2022-01-29T13:14:16.578Z] 	at com.nvidia.spark.rapids.Arm.freeOnExcept$(Arm.scala:140)
[...]

PR #4669, opened 2022-01-31
Prevent deadlock between RapidsBufferStore and RapidsBufferBase on close

abellina:

Targeting this to 22.04 but I can target to 22.02 if desired.

Closes https://github.com/NVIDIA/spark-rapids/issues/4664.

Prevents a deadlock between `RapidsBufferStore` and `RapidsBufferBase` on shutdown, removing code that was added before that was causing #4664 (this patch no longer stops the context -> no longer causes shuffle id conflicts on register)

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferCatalog.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferCatalog.scala
@@ -149,6 +149,13 @@ class RapidsBufferCatalog extends Logging {
     }
   }
 
+  /** Remove all buffers. Only to be called on shutdown. */
+  def removeAllBuffersOnShutdown(): Unit = {
+    val existingBuffers = bufferMap.keySet()
+    logWarning(s"Removing all buffers (${existingBuffers.size()}) from catalog when shutting down.")
+    existingBuffers.forEach(removeBuffer(_))
+  }
+
   /** Return the number of buffers currently in the catalog. */
   def numBuffers: Int = bufferMap.size()
 }
@@ -232,6 +239,8 @@ object RapidsBufferCatalog extends Logging with Arm {
       gdsStorage.close()
       gdsStorage = null
     }
+
+    singleton.removeAllBuffersOnShutdown()
   }
 
   def getDeviceStorage: RapidsDeviceMemoryStore = deviceStorage
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferCatalog.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferCatalog.scala
@@ -149,13 +149,6 @@ class RapidsBufferCatalog extends Logging {
     }
   }
 
-  /** Remove all buffers. Only to be called on shutdown. */
-  def removeAllBuffersOnShutdown(): Unit = {
-    val existingBuffers = bufferMap.keySet()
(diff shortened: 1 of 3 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferCatalog.scala
```diff
@@ -232,6 +239,8 @@ object RapidsBufferCatalog extends Logging with Arm {
       gdsStorage.close()
       gdsStorage = null
     }
+
+    singleton.removeAllBuffersOnShutdown()
```
jlowe, line 243: Wondering if this will just move the exception somewhere else.  For example, could we now get "unknown buffer ID" if an outstanding buffer gets unregistered after catalog shutdown?

It feels like we're patching a symptom rather than fixing the root cause of why all the buffers didn't get unregistered for the first context.

abellina:
This PR: https://github.com/NVIDIA/spark-rapids/pull/4374, added a `.close` for the `SparkSession` in the mortgage suite, saying that there was a hang, but not explaining anything more. I removed the `.close` and that uncovered a deadlock, handled in https://github.com/NVIDIA/spark-rapids/pull/4669/commits/4e302d2fdd2c8dccd611615e5725f204d1ffc664. I reverted my other change to remove all buffers because this keeps the `SparkContext` alive, and so the conflict of shuffle ids is no longer seen.

That said, I now see device memory leaks. I'll dig at that next. I am not sure if those are related to my change.

abellina:
Ok the reason for the memory leaks is because of the shutdown hook logic in the JVM. The shutdown hooks, as far as I understand, run in parallel, so a registered hook for the `MemoryCleaner` runs at any random time, even before Spark is able to stop it's context, unregister shuffles, etc. One approach that would remove ambiguity would be to change the `MemoryCleaner` to not hook directly to the JDK, but instead provide the API so that a custom shutdown order could be imposed (potentially registering the `MemoryCleaner` against Spark's `ShutdownHookManager`).

I think this is a nice to have but it would make the code easier to follow. It took me a while to realize that the cuDF MemoryCleaner is racing against spark's ShutdownHookManager.

abellina:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
```diff
@@ -70,11 +70,17 @@ abstract class RapidsBufferStore(
       }
```
jlowe, line 70: 2022 copyrights

jlowe (commented):
Would be good to file an issue about the underlying problem that still isn't addressed, re: closing and restarting the Spark session

abellina:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
abellina: done

jlowe (approved):

abellina:
build

abellina:
It is not clear to me why the previous one failed. I am looking to get some logs out of this to see if I can figure it out.

abellina:
build

abellina:
I am going to merge this. The previous time this failed, it seemed to die at `GpuDeviceManagerSuite`. Looking at the test, it seems to be a really simple test, that is just checking values constructed in the suite. So I am not sure why it failed, and I've had successful CI with this patch a few times.

PR #4681, opened 2022-02-02
Reimplement check for non-regexp strings using RegexParser

andygrove:
This PR reimplements the check used in GpuStringSplit to determine if a string is a regular expression or not. The previous approach had limitations and now that we have a regular expression parser we can improve the check so that there are fewer false positives.

This change is needed to support the work for https://github.com/NVIDIA/spark-rapids/issues/3542

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionParserSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionParserSuite.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, NVIDIA CORPORATION.
+ * Copyright (c) 2021-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionParserSuite.scala
```diff
@@ -21,6 +21,24 @@ import org.scalatest.FunSuite
...
+      "\f", "\\a", "\\e", "\\cx", "[abc]", "^", "[a-z&&[def]]", ".", "*", "\\d", "\\D",
+      "\\h", "\\H", "\\s", "\\S", "\\v", "\\V", "\\w", "\\w", "\\p", "$", "\\b", "\\B",
+      "\\A", "\\G", "\\Z", "\\z", "\\R", "?", "|", "(abc)", "a{1,}", "\\k", "\\Q", "\\E")
+    for (string <- strings) {
+      assert(RegexParser.isRegExpString(string))
+    }
+  }
+
+  test("detect non-regexp strings") {
+    val strings = Seq("\\.", "A", ",", "\t")
```
ttnghia, line 36: We also need to treat `":"` as non-regex.
```suggestion
    val strings = Seq("\\.", "A", ",", "\t", ":")
```

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionParserSuite.scala
```diff
@@ -21,6 +21,24 @@ import org.scalatest.FunSuite
...
+    for (string <- strings) {
+      assert(!RegexParser.isRegExpString(string))
+    }
+  }
+
   test("empty pattern") {
     assert(parse("") === RegexSequence(ListBuffer()))
```
ttnghia, line 43: Sorry, I can't get it: what will `RegexParser.isRegExpString("")` be?

andygrove: That should return `false` due to `case RegexSequence(parts) => parts.exists(isRegExpString)` but I will add it to the test to be sure.

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionParserSuite.scala
```diff
@@ -21,6 +21,24 @@ import org.scalatest.FunSuite
...
+      "\f", "\\a", "\\e", "\\cx", "[abc]", "^", "[a-z&&[def]]", ".", "*", "\\d", "\\D",
+      "\\h", "\\H", "\\s", "\\S", "\\v", "\\V", "\\w", "\\w", "\\p", "$", "\\b", "\\B",
+      "\\A", "\\G", "\\Z", "\\z", "\\R", "?", "|", "(abc)", "a{1,}", "\\k", "\\Q", "\\E")
+    for (string <- strings) {
+      assert(RegexParser.isRegExpString(string))
+    }
+  }
+
+  test("detect non-regexp strings") {
+    val strings = Seq("\\.", "A", ",", "\t", ":", "")
```
revans2, line 36: This looks wrong.  For example if I did a split with `"\\."` as the separator on `"a.b.c"` I would expect an output of `["a", "b", "c"]`. But I cannot take `"\\."` and pass it into a non-regexp string split, because it will try to look for a `"\\"` character in front of the '`.'`.  I am I wrong/missing something?  I think we could still support it, but we would have to transpile the regular expression into a constant string.

andygrove: Yes, I agree, this isn't what we want here. Moving this to draft for now and will rethink this approach.

andygrove: I updated to treat any string containing escaped characters as a regular expression.

andygrove:
build

Issue #4426, opened 2021-12-22
Handle hidden metadata columns


jlowe:
[SPARK-37273](https://issues.apache.org/jira/browse/SPARK-37273) added support for reading hidden metadata columns.  We either need to support reading these columns or falling back to the CPU if these columns are requested.

wbo4958:
I will check this issue

PR #4686, opened 2022-02-03
Fall back to CPU for queries referencing hidden metadata columns

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/4426

[SPARK-37273](https://issues.apache.org/jira/browse/SPARK-37273) added support for reading hidden metadata columns and we do not support this on GPU.

This PR adds checks to see if the user is attempting to read any of the metadata columns (`file_path`, `file_name`, `file_size`, `file_modification_time`) and will fall back to CPU in this case.


integration_tests/src/main/python/parquet_test.py
```diff
@@ -728,3 +728,39 @@ def do_parquet_scan(spark):
...
+    with_cpu_session(lambda spark : spark.range(10) \
+                     .selectExpr("id", "id % 3 as p") \
+                     .write \
+                     .partitionBy("p") \
+                     .mode("overwrite") \
+                     .parquet(data_path))
+
+    def do_parquet_scan(spark):
+        df = spark.read.parquet(data_path).selectExpr("id", "_metadata.file_path", "_metadata.file_name",
+                                                      "_metadata.file_size", "_metadata.file_modification_time")
```
jlowe, line 745: It might be good to iterate the test through the possible metadata names, otherwise just recognizing a single metadata column name but missing the others would still allow this test to pass.

integration_tests/src/main/python/parquet_test.py
```diff
@@ -728,3 +728,39 @@ def do_parquet_scan(spark):
...
+def test_parquet_scan_with_hidden_metadata_column_names(spark_tmp_path):
+    """
+    If the parquet file has column names file_path, file_name, file_size, file_modification_time then these
+    override the hidden metadata columns and we can run on GPU
+    """
+    data_path = spark_tmp_path + "/hidden_metadata_column_names.parquet"
+
+    def do_parquet_scan(spark):
+        spark.range(10).selectExpr("id", "id % 3 as file_size").write.partitionBy("id").mode("overwrite").parquet(data_path)
+        df = spark.read.parquet(data_path).selectExpr("Max(file_size)")
```
jlowe, line 763: I'm confused by this test, as it's not trying to access a metadata column via the `_metadata.` prefix so I'm not sure how it would potentially trigger the metadata path.  I'd actually be more curious what would happen if the Parquet file had a column named `_metadata`, especially if it is a struct column with fields named `file_size`, etc.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuReadParquetFileFormat.scala
```diff
@@ -57,7 +57,20 @@ class GpuReadParquetFileFormat extends ParquetFileFormat with GpuReadFileFormatW
 }
 
 object GpuReadParquetFileFormat {
+
+  private val hiddenMetadataFields = Set(
```
jlowe, line 61: Rather than manually maintain a list of metadata columns and hope we're keeping it in sync with Spark, I think it would be a lot easier to just check the `metadata` field of the original `FileSourceScan` (in Spark 3.3.x+) and make sure we do not convert it to the GPU if it is non-empty.  Spark is already determining whether there are metadata fields, and we can reuse that logic to stay in sync.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuReadParquetFileFormat.scala
```diff
@@ -57,7 +58,19 @@ class GpuReadParquetFileFormat extends ParquetFileFormat with GpuReadFileFormatW
 }
 
 object GpuReadParquetFileFormat {
+
+  // copied from src/main/scala/org/apache/spark/sql/catalyst/util/package.scala
+  private val METADATA_COL_ATTR_KEY = "__metadata_col"
```
jlowe, line 63: I'd personally rather lean more into leveraging what Spark is doing to identify the metadata columns.  Rather than copying the logic, and hoping it stays in sync, we can let Spark identify the metadata columns for us.  When a metadata column is used, it shows up as a special type of attribute, `MetadataAttribute`.  Similar to what is done in 3.3's `FileSourceScanExec`, we could check if any the `output` attributes in `FileSourceScanExec` are actually `MetadataAttribute`.  If so, we avoid converting it to GPU.  Converting from `FileSourceScanExec` is already shimmed, so this would be a matter of putting this logic in the 3.3 shims for execs.

This also has the benefit of applying to _any_ file input format that is doing this, not just Parquet.  ORC, CSV, etc. aren't being handled in this PR currently, but ultimately they need to be.

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
```diff
@@ -143,4 +143,16 @@ trait Spark33XShims extends Spark33XFileOptionsShims {
...
   ).map(r => (r.getClassFor.asSubclass(classOf[Scan]), r)).toMap
+
+  override def tagHiddenMetadataExpressions(meta: SparkPlanMeta[FileSourceScanExec]): Unit = {
+    val metadataColumns = meta.wrapped.expressions.filter { expr =>
+      expr match {
+        case MetadataAttribute(expr) => true
+        case _ => false
+      }
+    }
+    if (metadataColumns.nonEmpty) {
```
jlowe, line 154: Nit: We don't actually care which columns are the metadata columns, just if there is at least one.
```suggestion
    val hasMetadataColumns = meta.wrapped.expressions.exists { expr =>
      expr.isInstanceOf[MetadataAttribute]
    }
    if (hasMetadataColumns) {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileSourceScanExec.scala
```diff
@@ -589,6 +589,7 @@ case class GpuFileSourceScanExec(
 
 object GpuFileSourceScanExec {
   def tagSupport(meta: SparkPlanMeta[FileSourceScanExec]): Unit = {
+    ShimLoader.getSparkShims.tagHiddenMetadataExpressions(meta)
```
jlowe, line 592: We've been moving away from v1 style runtime-loaded shims and instead doing v2 compile-time shims for any new shims.  This could be handled in one of two ways, either adding a v2 shim utility method to check for metadata columns (e.g.: com.nvidia.spark.rapids.shims.v2.DataSourceUtils or whatever) or it can be handled in the meta wrapper for FileSourceScanExec which is already shim-specific (e.g.: see the custom tagging code at https://github.com/NVIDIA/spark-rapids/blob/branch-22.04/sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/Spark30XdbShims.scala#L179-L187), however we'd need to move FileSourceScanExec wrapping to Spark33XShims in that case.

integration_tests/src/main/python/parquet_test.py
```diff
@@ -728,3 +729,44 @@ def do_parquet_scan(spark):
...
+
+    assert_cpu_and_gpu_are_equal_collect_with_capture(
+        do_parquet_scan,
+        exist_classes= "FileSourceScanExec",
+        non_exist_classes= "GpuBatchScanExec")
+
+@pytest.mark.skipif(is_before_spark_330(), reason='Hidden file metadata columns are a new feature of Spark 330')
+def test_parquet_scan_with_hidden_metadata_column_names(spark_tmp_path):
+    """
+    If the parquet file has column names file_path, file_name, file_size, file_modification_time then these
```
jlowe, line 757: I'm not sure we need this test anymore.  It's testing Spark more than the plugin now that we're just keying off of whether Spark thinks there's a metadata column or not.

integration_tests/src/main/python/parquet_test.py
```diff
@@ -728,3 +729,44 @@ def do_parquet_scan(spark):
         exist_classes= "BatchScanExec",
         non_exist_classes= "GpuBatchScanExec",
         conf = conf_for_parquet_aggregate_pushdown)
+
+@pytest.mark.skipif(is_before_spark_330(), reason='Hidden file metadata columns are a new feature of Spark 330')
```
jlowe, line 733: It would be good to have a smoke test for metadata columns in the ORC and CSV tests as well, as that should be covered by the updated changes.

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
andygrove: I will make the `exists` change but we need to use `MetadataAttribute.unapply` here rather than `.isInstanceOf[MetadataAttribute]` since the expressions themselves are `AttributeReference`

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileSourceScanExec.scala
andygrove: I actually started out by copying the `FileSourceScanExec` wrapping into the `Spark33XShims` but it copied a lot of code without modification. I will look into the v2 approach.

andygrove: I removed the call to the shim loader but I'm not sure if my change is exactly what you were looking for. Let me know if I should refactor this to a call to a separate util class.

jlowe (commented):
> I removed the call to the shim loader but I'm not sure if my change is exactly what you were looking for.

I was expecting a v2 shim object class that provided a utility method that the existing `GpuFileSourceScanExec.tagSupport` would call internall to tag for metadata columns.  The pre-330 version of that class would do nothing, while the 330+ version of that class would check for `MetadataAttribute`.

However what's here works too.  I mainly didn't want to add yet another for-all-shims interface and runtime-lookup of the shim, and this accomplishes that goal.

integration_tests/src/main/python/csv_test.py
```diff
@@ -14,13 +14,13 @@
 
 import pytest
 
-from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_fallback_collect, assert_gpu_fallback_write
+from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_fallback_collect, assert_gpu_fallback_write, assert_cpu_and_gpu_are_equal_collect_with_capture
```
ttnghia, line 17: I prefer to break line within 100 chars :smiley:

andygrove:
build

PR #4688, opened 2022-02-03
Optimize DECIMAL128 sum aggregations [databricks]

jlowe:
Depends on rapidsai/cudf#10201.

This accelerates sum aggregations on DECIMAL128 by splitting up the 128-bit values into 32-bit chunks, summing the chunks separately into 64-bit accumulated values, and then reassembling the 128-bit value from the accumulated chunks (with overflow checking).  This changes what would normally force a cudf sort-based aggregation into one that can be hash-based which can significantly improve performance.  This also allows us to remove some of the DECIMAL128 overflow code.


jlowe:
build

revans2 (dismissed):
The main thing I would like to see is follow on work to do the same kind of thing for average. We also should look at SUM for window operations. I don't think it will be needed there because the data comes in sorted, but some of the cleanup that has been done to split decimal from others would be good there too.

jlowe:
build

revans2 (approved):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -734,285 +804,355 @@ case class GpuDecimalSumHighDigits(
...
+  override lazy val preMerge: Seq[Expression] = aggBufferAttributes
+
+  protected lazy val mergeSum: CudfAggregate = new CudfSum(resultType)
 
-  private lazy val zeroDec = {
-    val dt = resultType.asInstanceOf[DecimalType]
-    GpuLiteral(Decimal(0, dt.precision, dt.scale), dt)
+  override lazy val mergeAggregates: Seq[CudfAggregate] =Seq(mergeSum)
+
+  override lazy val postMerge: Seq[Expression] = postMergeAttr
```
abellina, line 858: this is the default value for `postMerge`, so we should not need this override.

abellina: same question for `postUpdate`

abellina (changes requested):

jlowe:
Tracked down the CI failure to a libcudf issue with sort-based sum aggregations not performing aggregations in the result type as hash-based aggregations do.  Filed rapidsai/cudf#10246.  In the meantime, I'll update this to pre-cast the inputs to avoid the issue.

jlowe:
@abellina your comments should now be addressed.

Note that this also includes a change to lower the batch size being used in hash_aggregate_tests for exercising out-of-core hash aggregate processing, as the 312db failure from the previous CI run was triggered by such processing.  We were _not_ seeing it in other CI runs because we weren't regularly exercising this code path, but we do with the new lower batch size value.  This adds approx 5 minutes of test time on my desktop (with 4-way parallelism), but it seems worth it for the extra coverage.

jlowe:
build

revans2 (approved):

abellina:
One more nit on an override

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala
```diff
@@ -734,285 +809,350 @@ case class GpuDecimalSumHighDigits(
...
 
-  // For some operations we need to SUm the higher digits in addition to the regular value so
-  // we can detect overflow. This is the type of the higher digits SUM value.
-  private lazy val higherDigitsCheckType: DecimalType = {
-    val t = resultType.asInstanceOf[DecimalType]
-    DecimalType(t.precision - GpuDecimalSumOverflow.updateCutoffPrecision, 0)
-  }
+  override lazy val aggBufferAttributes: Seq[AttributeReference] = sum :: Nil
+
+  override lazy val preMerge: Seq[Expression] = aggBufferAttributes
```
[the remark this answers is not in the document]
jlowe: Fixed.

jlowe:
build

abellina (approved):

Issue #589, opened 2020-08-19
[FEA] Support ExistenceJoin


revans2:
**Is your feature request related to a problem? Please describe.**
Spark will some times use an optimized join type called an `ExistenceJoin`.  It is not a direct join type you can just do but instead an optimized version of the `IN` or `EXISTS` operators that lets spark remove a subquery.  This shows up in TPC-DS query 10, and probably others.  It is very similar to a left semi join, but instead of filtering out columns that don't match instead it adds a new column with a boolean true or false to indicate if the join would have matched. 

**Describe the solution you'd like**
I think we might need some support from cudf for this.

**Describe alternatives you've considered**
There isn't a lot here. I think we need some help from cudf.

- [x] add an integration test exhibiting ExistenceJoin  #4697
- [ ] implement a gatherer for ExistenceJoin #4796

viadea:
mini repro:
```
select
  count(*)
from
  customer c
where
  exists (select * from store_sales
          where c.c_customer_sk = ss_customer_sk) 
          AND
   (exists (select * from web_sales
            where c.c_customer_sk = ws_bill_customer_sk) or
    exists (select * from catalog_sales
            where c.c_customer_sk = cs_ship_customer_sk))
```

Driver Log message:
```
scala> spark.sql(query).explain()
21/11/03 23:19:56 WARN GpuOverrides:
          !Exec <SortMergeJoinExec> cannot run on GPU because ExistenceJoin(exists#225) currently is not supported
            @Expression <AttributeReference> c_customer_sk#0 could run on GPU
            @Expression <AttributeReference> cs_ship_customer_sk#156 could run on GPU
            !Exec <SortMergeJoinExec> cannot run on GPU because ExistenceJoin(exists#224) currently is not supported
              @Expression <AttributeReference> c_customer_sk#0 could run on GPU
              @Expression <AttributeReference> ws_bill_customer_sk#85 could run on GPU
```

revans2:
We might be able to do this without any changes from CUDF.  ExistenceJoin is not a regular join type, as was stated before. As such cudf might push back a bit if we try to implement it there.  Happily ExistanceJoin really just produces the left output columns, unchanged, and adds in an "exists" boolean column. The exists boolean columns is the same thing that is used by Spark to filter the left table on a semi-join.

As such we can produce the gather map for a left semi-join, and instead of gathering the results, we can look to see if the column would have been gathered.

[...]

PR #4697, opened 2022-02-05
Add preliminary test and test framework changes for ExistanceJoin

gerashegalov:
1. Allow matching short TreeNode string against a regex. 
1. which enables  us to ensure that the test exhibits ExistenceJoin
    
Contributes to #589

### DIFF
--- a/integration_tests/src/main/python/join_test.py
+++ b/integration_tests/src/main/python/join_test.py
@@ -15,7 +15,7 @@
 import pytest
 from pyspark.sql.functions import broadcast
 from pyspark.sql.types import *
-from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_fallback_collect
+from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_fallback_collect, assert_cpu_and_gpu_are_equal_collect_with_capture
 from conftest import is_databricks_runtime, is_emr_runtime
 from data_gen import *
 from marks import ignore_order, allow_non_gpu, incompat, validate_execs_in_gpu_plan
@@ -361,7 +361,7 @@ def test_broadcast_nested_loop_join_special_case_group_by(data_gen, batch_size):
 def test_right_broadcast_nested_loop_join_with_ast_condition(data_gen, join_type, batch_size):
     def do_join(spark):
         left, right = create_df(spark, data_gen, 50, 25)
-        # This test is impacted by https://github.com/NVIDIA/spark-rapids/issues/294 
+        # This test is impacted by https://github.com/NVIDIA/spark-rapids/issues/294
         # if the sizes are large enough to have both 0.0 and -0.0 show up 500 and 250
         # but these take a long time to verify so we run with smaller numbers by default
         # that do not expose the error
@@ -651,7 +651,7 @@ def test_half_cache_join(join_type, cache_side, cpu_side):
 
         if (cache_side == 'cache_left'):
             # Try to force the shuffle to be split between CPU and GPU for the join
-            # by default if the operation after the shuffle is not on the GPU then 
+            # by default if the operation after the shuffle is not on the GPU then
             # don't do a GPU shuffle, so do something simple after the repartition
             # to make sure that the GPU shuffle is used.
             left = left.repartition('a').selectExpr('b + 1 as b', 'a').cache()
@@ -659,7 +659,7 @@ def test_half_cache_join(join_type, cache_side, cpu_side):
         else:
             #cache_right
             # Try to force the shuffle to be split between CPU and GPU for the join
-            # by default if the operation after the shuffle is not on the GPU then 
+            # by default if the operation after the shuffle is not on the GPU then
             # don't do a GPU shuffle, so do something simple after the repartition
             # to make sure that the GPU shuffle is used.
             right = right.repartition('r_a').selectExpr('c + 1 as c', 'r_a').cache()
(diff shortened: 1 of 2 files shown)


gerashegalov:
build

revans2 (commented):
The code looks fine. I am just a little confused why this needs to be split off from the rest of the ExistanceJoin code. I guess adding in the regular expression check is technically unrelated, but then I would prefer to see that as a separate PR without the xfail test.

If I saw "Add an xfail'd test for ExistenceJoin" in the changelog I would be confused and not know why we did that or what it was all about.

integration_tests/src/main/python/join_test.py
```diff
@@ -785,3 +785,37 @@ def do_join(spark):
...
+# which feeds into a filter "exists OR someOtherPredicate"
+# If the condition is something like an AND, it makes the result a subset of a SemiJoin, and
+# the optimizer won't use ExistenceJoin.
+@ignore_order(local=True)
+@pytest.mark.parametrize(
+    "allowFallback", [
+        pytest.param('true',
+            marks=pytest.mark.allow_non_gpu('SortMergeJoinExec')),
+        pytest.param('false',
+            marks=pytest.mark.xfail(reason="https://github.com/NVIDIA/spark-rapids/issues/589"))
```
jlowe, line 803: I find it a bit odd to add a test that tests for both all-on-GPU and not-all-on-GPU, and only one is ever expected to pass.  Tests should cover what is expected, and when a PR later changes the behavior, that PR also changes the test accordingly.  Having a test do it both ways like this is confusing and wastes test time.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -358,6 +359,7 @@ object ExecutionPlanCaptureCallback {
   }
 
   def assertContains(gpuPlan: SparkPlan, className: String): Unit = {
+    implicit val regexMap = scala.collection.mutable.Map.empty[String, Try[Regex]]
```
jlowe, line 362: The use of implicit parameter passing seems unnecessary and makes the code harder to read.  It looks like everywhere that calls `containsPlan` now needs to make sure they create an implicit map val before calling, so I'm wondering why we don't just pass the map explicitly?  My Scala-fu is not the best, so maybe I'm missing something.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -408,8 +415,16 @@ object ExecutionPlanCaptureCallback {
...
     case p: ReusedExchangeExec =>
       containsPlan(p.child, className)
-    case p =>
-      p.expressions.exists(containsExpression(_, className))
+    case p if p.expressions.exists(containsExpression(_, className)) =>
+      true
+    case p: SparkPlan =>
+      regexMap.getOrElseUpdate(className,
+        Try(className.r).recoverWith { case t: Throwable =>
+          logWarning(s"Regex broken in a simpleStringWithNodeId match, $t")
```
jlowe, line 423: Why make this a log rather than actually throw?  If the regex is broken we want the test to fail so the regex can be fixed.  Logging a message in a test is very likely to be completely ignored, as most developers aren't scrubbing the test logs.

integration_tests/src/main/python/join_test.py
gerashegalov: The not-all-on-GPU test is there to prove that we are executing the right plan. The GPU-only test with the existing constructs already bails on the fact "SortMergeJoinExec" was not on GPU without verifying it had ExistenceJoin as joinType

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
gerashegalov: This API does not handle regexes exclusively (and I did not want to create a new one).

gerashegalov:
> The code looks fine. I am just a little confused why this needs to be split off from the rest of the ExistanceJoin code. I guess adding in the regular expression check is technically unrelated, but then I would prefer to see that as a separate PR without the xfail test.

It does not have to, but I thought I create it as a tiny milestone in the test-first approach. I can keep working on this branch until the feature is done and there is no need for xfail. 

> 
> If I saw "Add an xfail'd test for ExistenceJoin" in the changelog I would be confused and not know why we did that or what it was all about.

I will make it more clear if we are going to keep the test-only PR

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
jlowe: And what if the test passes when it shouldn't, because the regex was broken and didn't find what it was supposed to that would have caused the test to properly fail?  I'm really not a fan of this.

gerashegalov:
build

integration_tests/src/main/python/join_test.py
jlowe: I still don't understand why we would run the test twice.  The test that is verifying we fallback to the CPU can check the SortMergeExec node to ensure it's an existence join.  Otherwise with this kind of precedent, we'd be running all of our tests twice to verify we're getting the right plan before we run it to put it on the GPU.

Not going to block this, since you're actively working on ExistenceJoin. I'm assuming this will soon revert back to just being a single test to make sure it's on the GPU in that PR.

jlowe (approved):

revans2 (approved):

Issue #4614, opened 2022-01-24
[FEA]JSON reader: support "allowNumericLeadingZero" option


GaryShen2008:
From the [Spark JSON option](https://spark.apache.org/docs/latest/sql-data-sources-json.html#data-source-option),
"allowNumericLeadingZero",  Allows leading zeros in numbers (e.g. 00012).

Here the parser strips the leading zeros for numeric JSON types instead of marking them as corrupt entries, 
so if I had something like {"a": 0001, "b": 01.1000000000} both entries would cause the row to be marked as corrupt. 
Even if we provide a schema an ask for both a and b to be strings they show up as nulls because the values are technically corrupt. If I set the option to true they are parsed, but the leading and in the case of the float, 
trailing zeros are stripped from the resulting string.

Comments from Bobby:
My guess is that it is actually the other way around.  CUDF probably strips off the leading zeros and would not return an error if it saw them. If we ask CUDF to return them all as strings we might be able to implement a leading zero check, but I don't know if we want to. This is because of the oddness I talked abut in the last comment. I think if we have {"a": "01.100"} and ask Spark for a double it will not complain, but if we have the quotes off it will.  I am not 100% sure about that though.

nartal1:
As per Bobby's observation, CUDF strips off the leading zeros by default. So the behavior is same as Spark's when this option is set to true.


When it is set to `false`(which is the default), there are couple of scenarios:

1. If there is only column in the json file and that column has leading zeros. AnalysisException is thrown :
```
scala> val df = spark.read.option("allowNumericLeadingZeros", "false").json("/home/nartal/json-reader/people-numericLeadingZero-1.json")
df: org.apache.spark.sql.DataFrame = [_corrupt_record: string]

scala> df.show
org.apache.spark.sql.AnalysisException: Since Spark 2.3, the queries from raw JSON/CSV files are disallowed when the
referenced columns only include the internal corrupt record column
(named _corrupt_record by default). For example:
spark.read.schema(schema).json(file).filter($"_corrupt_record".isNotNull).count()
and spark.read.schema(schema).json(file).select("_corrupt_record").show().
Instead, you can cache or save the parsed results and then send the same query.
For example, val df = spark.read.schema(schema).json(file).cache() and then
df.filter($"_corrupt_record".isNotNull).count().
```
2. If there is  more than 1 column in the json file, 
CUDF throws error that types for all columns must be specified(as column pruning is not supported yet for json reader in CUDF).

```
[...]

Issue #4612, opened 2022-01-24
[FEA]JSON reader: support unquoted field name and control chars


GaryShen2008:
From the [Spark JSON option](https://spark.apache.org/docs/latest/sql-data-sources-json.html#data-source-option), 
there're 2 options.
1. allowUnquotedFieldNames: Allows unquoted JSON field names.
2. allowUnquotedControlChars: Allows JSON Strings to contain unquoted control characters (ASCII characters with value less than 32, including tab and line feed characters) or not.

revans2:
allowUnquotedControlChars is what CUDF supports by default already. We need a way to disable it because that is what Spark has by default.

revans2:
This depends on https://github.com/rapidsai/cudf/issues/15222. It may be broken up into smaller pieces in the future, and there are a lot of things that could be validated.

PR #4698, opened 2022-02-05
Add tests for few json options

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/4614 and partially addresses https://github.com/NVIDIA/spark-rapids/issues/4612.

Added tests to verify `allowNumericLeadingZeros` and `allowUnquotedControlChars` options. Have documented the differences between Spark and plugin.


docs/compatibility.md
```diff
@@ -489,6 +489,14 @@ these formats when unquoted, will produce `null` on the CPU and may produce vali
 Another limitation of the GPU JSON reader is that it will parse strings containing floating-point values where
 Spark will treat them as invalid inputs and will just return `null`.
 
+### json options
```
revans2, line 492: nit: for consistency I Think this should be 
```
### JSON Options
```

docs/compatibility.md
```diff
@@ -489,6 +489,14 @@ these formats when unquoted, will produce `null` on the CPU and may produce vali
...
+
+Below are some of the options that are supported by the plugin with minor differences when compared to Spark:
```
revans2, line 494: Can we explain this a little more. Perhaps something like.

```
Spark supports passing options to the JSON parser when reading a dataset.  In most cases if the RAPIDS Accelerator sees one of these options that it does not support it will fall back to the CPU. In some cases we do not. The following options are documented below.
```

docs/compatibility.md
```diff
@@ -489,6 +489,14 @@ these formats when unquoted, will produce `null` on the CPU and may produce vali
...
+
+- `allowNumericLeadingZeros`  - Allows leading zeros in numbers (e.g. 00012). Spark throws exception when the option
```
revans2, line 496: Again I think we could be more clear here.

```
- `allowNumericLeadingZeros`  - Allows leading zeros in numbers (e.g. 00012). By default this is set to false. When it is false Spark throws an exception when if it encounters this type of number.  The RAPIDS Accelerator strips off leading zeros from all numbers and this config has no impact on it. 
```

docs/compatibility.md
```diff
@@ -489,6 +489,14 @@ these formats when unquoted, will produce `null` on the CPU and may produce vali
 Another limitation of the GPU JSON reader is that it will parse strings containing floating-point values where
 Spark will treat them as invalid inputs and will just return `null`.
 
+### json options
+
+Below are some of the options that are supported by the plugin with minor differences when compared to Spark:
+
+- `allowNumericLeadingZeros`  - Allows leading zeros in numbers (e.g. 00012). Spark throws exception when the option 
+is set as false whereas plugin always strips off the leading zeros.
+- `allowUnquotedControlChars` - Allows JSON Strings to contain unquoted control characters. This matches with Spark's behavior.
```
revans2, line 498: Can we list what the default is, what Spark does when it encounters unquoted control characters that where not expected and what the RAPIDS accelerator does?

integration_tests/src/main/python/json_test.py
```diff
@@ -170,17 +173,33 @@ def test_json_ts_formats_round_trip(spark_tmp_path, date_format, ts_part, v1_ena
     'nan_and_inf.json',
     pytest.param('nan_and_inf_edge_cases.json', marks=pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/4646')),
     'floats.json',
+    pytest.param('floats_leading_zeros.json', marks=pytest.mark.xfail(reason="cudf always strips off leading zeros")),
```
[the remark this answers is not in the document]
nartal1: Thanks @revans2 for reviewing this. The reason I had put the xfail is the test fails when those options are not set to `true`. Plugin behavior is same as Spark when the option is set to `true` so those test will get XPASS. 
If the option is `false`(default), it's when it would XFAIL:
- `allowNumericLeadingZeros` - fails on CPU  with an exception where as GPU parses it correctly.
- `allowUnquotedControlChars` - fails on both CPU and GPU.

I was testing with both the options (true, false). Would it be okay to add tests only with `true` and document the default behavior(false) in the .md file below?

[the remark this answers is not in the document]
revans2: Yes I am fine with a test that is only run when `allowNumericLeadingZeros` is true.

docs/compatibility.md
```diff
@@ -489,13 +489,21 @@ these formats when unquoted, will produce `null` on the CPU and may produce vali
...
-- `allowUnquotedControlChars` - Allows JSON Strings to contain unquoted control characters. This matches with Spark's behavior.
+- `allowNumericLeadingZeros`  - Allows leading zeros in numbers (e.g. 00012). By default this is set to false. 
+When it is false Spark throws an exception if it encounters this type of number. The RAPIDS Accelerator 
+strips off leading zeros from all numbers and this config has no impact on it.
+
+- `allowUnquotedControlChars` - Allows JSON Strings to contain unquoted control characters (ASCII characters with 
+value less than 32, including tab and line feed characters) or not. By default this is set to false. If the schema 
+is provided while reading JSON file, then this flag has no impact on the RAPIDS Accelerator as it always allows 
+unquoted control characters but Spark reads these entries incorrectly as null. However, if the schema is not provided
+and when the option is false, then RAPIDS Accelerator's behavior is same as Spark where an exception is thrown.
```
revans2, line 506: Can we explain why this happens when no schema is provided?  Could we just have a separate section about schema discovery and JSON?

nartal1: Added separate section about JSON schema discovery.

docs/compatibility.md
```diff
@@ -489,6 +489,34 @@ these formats when unquoted, will produce `null` on the CPU and may produce vali
...
 
+### JSON Schema discovery
+
+Spark SQL can automatically infer the schema of a JSON dataset if schema is not provided explicitly. In certain 
+cases it cannot infer the schema correctly. For example if JSON dataset has values such as 00012 which is read 
+with default options then those are marked as corrupt entries with a dedicated column name specified by 
+`columnNameOfCorruptRecord`. An exception is thrown if there is column with name`columnNameOfCorruptRecord` in the
+dataframe. 
+However if the schema is provided with default options while reading for such entries they show up as nulls 
+because the values are technically corrupt.
```
revans2, line 500: I was thinking more along the lines of 

```
Spark SQL can automatically infer the schema of a JSON dataset if schema is not provided explicitly. The CPU handles schema discovery and there is no GPU acceleration of this. By default Spark will read/parse the entire 
dataset to determine the schema. This means that some options/errors which are ignored by the GPU may still
result in an exception if used with schema discovery.
```

nartal1:
build

PR #4701, opened 2022-02-07
[BUG] Decrease nighlty tests upper limitation to 7 [skip ci]

### LINKED ISSUE #4675: jenkins integration build timed out at 10 hours
**Describe the bug**
It looks like our integration build that tests the stable versions of Spark timed out after 10 hours. It looks like we do builds in 3 different stages and within those stages things are parallel. But I think that will means we get 3 times the max test time. It looks like each of those is taking like 3.5 hours so we hit the timeout.  I will increase the timeout for now but we may want to find a better way to do this.

pxLi:

fix #4675, 8 as the upper bound of parallelism is not good any more. Decrease it to 7 to save around 40-60 mins

### DIFF
--- a/jenkins/spark-tests.sh
+++ b/jenkins/spark-tests.sh
@@ -258,8 +258,8 @@ if [[ $TEST_MODE == "ALL" || $TEST_MODE == "IT_ONLY" ]]; then
       PARALLELISM=$(nvidia-smi --query-gpu=memory.free --format=csv,noheader | \
                     awk '{if (MAX < $1){ MAX = $1}} END {print int(MAX / (2 * 1024))}')
     fi
-    # parallelism > 8 could slow down the whole process, so we have a limitation for it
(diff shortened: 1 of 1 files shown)


pxLi:
build

GaryShen2008 (approved):

tgravescs:
I'm a bit unclear by the description here, decreasing the parallelism makes it faster why?  We are allowing more gpu memory and not spilling as much?

pxLi:
> I'm a bit unclear by the description here, decreasing the parallelism makes it faster why? We are allowing more gpu memory and not spilling as much?

Per the test results, its not always the larger parallelism, the faster total tests run. Even we have more gpu mem to split tests, the parallel run could much slower when too many cases were sharing the cuda cores (competing the SM? context switching? sry, I am not a cuda expert)

PR #4714, opened 2022-02-07
Add regular expression support to string_split

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/4003

Depends on https://github.com/rapidsai/cudf/pull/10139

Follow-on issues:
- https://github.com/NVIDIA/spark-rapids/issues/4719
- https://github.com/NVIDIA/spark-rapids/issues/4720

Status:

- [x] Draft implementation
- [x] Move some logic from GpuStringSplit to GpuStringSplitMeta
- [x] File follow-on issues for limit = 0 or 1, and supporting line and string anchors


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1333,7 +1330,21 @@ case class GpuStringSplit(str: Expression, regex: Expression, limit: Expression)
...
+        // TODO this parameter has different meaning between Java and cuDF (limit vs maxSplit)
+        intLimit,
+        true)
+    } else {
+      str.getBase.stringSplitRecord(
+        pattern,
+        // TODO this parameter has different meaning between Java and cuDF (limit vs maxSplit)
+        intLimit,
+        false)
+    }
```
ttnghia, line 1347:
```suggestion
val splitByRegex = RegexParser.isRegExpString(pattern);
str.getBase.stringSplitRecord(pattern, intLimit, splitByRegex)
```

ttnghia: Did I miss something? Why do we need to generate `cudfPattern`?

andygrove: We need to translate Java regexp to cuDF regexp to address compatibility issues

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1289,51 +1289,67 @@ class GpuStringSplitMeta(
...
-        if (RegexParser.isRegExpString(str.toString)) {
-          willNotWorkOnGpu("regular expressions are not supported yet")
-        }
         if (str.numChars() == 0) {
           willNotWorkOnGpu("An empty regex is not supported yet")
         }
+        isRegExp = RegexParser.isRegExpString(str.toString)
+        if (isRegExp) {
+          try {
+            pattern = new CudfRegexTranspiler(RegexSplitMode).transpile(str.toString)
```
ttnghia, line 1308: Is `tagExprForGpu` guaranteed to be called before `convertToGpu`? I'm a bit nervous with the `pattern` variable not being initialized at the beginning.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1289,51 +1289,67 @@ class GpuStringSplitMeta(
...
       regexp: Expression,
       limit: Expression): GpuExpression =
-    GpuStringSplit(str, regexp, limit)
+    GpuStringSplit(str, regexp, limit, isRegExp, pattern)
 }
 
-case class GpuStringSplit(str: Expression, regex: Expression, limit: Expression)
-    extends GpuTernaryExpression with ImplicitCastInputTypes {
+case class GpuStringSplit(str: Expression, regex: Expression, limit: Expression,
+    isRegExp: Boolean, pattern: String)
```
ttnghia, line 1338: Can we get rid of the `regex` expression completely? It is now useless since we use `pattern` instead.

ttnghia: Maybe this can be achieved if you override `columnarEval` instead of `doColumnar` similar to https://github.com/NVIDIA/spark-rapids/pull/4636/files#diff-a12810882b81a4eb395c03a80951f96ec080db793ffed6755739eeb2122840ccR1432

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1289,51 +1289,67 @@ class GpuStringSplitMeta(
...
     val regexp = extractLit(expr.regex)
     if (regexp.isEmpty) {
       willNotWorkOnGpu("only literal regexp values are supported")
     } else {
       val str = regexp.get.value.asInstanceOf[UTF8String]
       if (str != null) {
-        if (RegexParser.isRegExpString(str.toString)) {
-          willNotWorkOnGpu("regular expressions are not supported yet")
-        }
         if (str.numChars() == 0) {
```
ttnghia, line 1302: This should be removed. libcudf supports empty delimiter.

sql-plugin/src/main/301until310-nondb/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceMeta.scala
```diff
@@ -39,7 +39,7 @@ class GpuRegExpReplaceMeta(
           // use GpuStringReplace
         } else {
           try {
-            pattern = Some(new CudfRegexTranspiler(replace = true).transpile(s.toString))
+            pattern = Some(new CudfRegexTranspiler(RegexReplaceMode).transpile(s.toString))
```
ttnghia, line 42: Note that this also needs to be modified for all shim.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
andygrove: It is, but it doesn't hurt to add some validation here. I will add this.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1289,51 +1289,69 @@ class GpuStringSplitMeta(
...
     }
   }
   override def convertToGpu(
       str: Expression,
       regexp: Expression,
-      limit: Expression): GpuExpression =
-    GpuStringSplit(str, regexp, limit)
+      limit: Expression): GpuExpression = {
+    GpuStringSplit(str, regexp, limit, isRegExp, pattern.getOrElse(
+      throw new IllegalStateException("Expression has not been tagged with cuDF regex pattern")))
```
ttnghia, line 1335: Can we do this?
```suggestion
      {
        tagExprForGpu()
        pattern.get
      }
```

integration_tests/src/main/python/string_test.py
```diff
@@ -25,15 +27,100 @@
...
+            'split(a, "C", 3)',
+            'split(a, "_", 999)'))
+
+def test_split_re_negative_limit():
+    data_gen = mk_str_gen('([bf]o{0,2}:){1,7}') \
+        .with_special_case('boo:and:foo')
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark : unary_op_df(spark, data_gen).selectExpr(
+            'split(a, ":", -1)',
+            'split(a, "o", -2)'))
```
revans2, line 80: Why is "o" an re?  Would also like to understand about ':' being an RE too, because eventually we will want to trans-pile the re into a static string if we can get away with it, and both of these appear to be candidates for that.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
revans2: It would be switching this from a `GpuTernaryExpression` to a `GpuBinaryExpression`. I personally don't see it as a big deal either way.

revans2 (commented):
Nothing big, this is looking good.

integration_tests/src/main/python/string_test.py
andygrove: Thanks for catching this. I took these examples from Spark's documentation for string_split when debugging differences in `limit` and forgot about the regexp aspect while working on that. I will update these to use character classes instead.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
andygrove: I would prefer to stick with `GpuTernaryExpression` to match Spark

andygrove:
build

sql-plugin/src/main/311+-nondb/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
```diff
@@ -18,7 +18,7 @@ package com.nvidia.spark.rapids.shims.v2
 import com.nvidia.spark.rapids.{CudfRegexTranspiler, DataFromReplacementRule, GpuExpression, GpuOverrides, QuaternaryExprMeta, RapidsConf, RapidsMeta, RegexUnsupportedException}
 
 import org.apache.spark.sql.catalyst.expressions.{Expression, Literal, RegExpReplace}
-import org.apache.spark.sql.rapids.{GpuRegExpReplace, GpuRegExpUtils, GpuStringReplace}
+import org.apache.spark.sql.rapids.{GpuRegExpReplace, GpuRegExpUtils, GpuStringReplace, RegexReplaceMode}
```
ttnghia, line 21: Similar to above.

sql-plugin/src/main/31xdb/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
```diff
@@ -18,7 +18,7 @@ package com.nvidia.spark.rapids.shims.v2
 import com.nvidia.spark.rapids.{CudfRegexTranspiler, DataFromReplacementRule, GpuExpression, GpuOverrides, QuaternaryExprMeta, RapidsConf, RapidsMeta, RegexUnsupportedException}
 
 import org.apache.spark.sql.catalyst.expressions.{Expression, Literal, RegExpReplace}
-import org.apache.spark.sql.rapids.{GpuRegExpReplace, GpuRegExpUtils, GpuStringReplace}
+import org.apache.spark.sql.rapids.{GpuRegExpReplace, GpuRegExpUtils, GpuStringReplace, RegexReplaceMode}
```
ttnghia, line 21: Similar to above.

sql-plugin/src/main/301db/scala/com/nvidia/spark/rapids/shims/v2/GpuRegExpReplaceExec.scala
```diff
@@ -18,7 +18,7 @@ package com.nvidia.spark.rapids.shims.v2
 import com.nvidia.spark.rapids.{CudfRegexTranspiler, DataFromReplacementRule, GpuExpression, GpuOverrides, RapidsConf, RapidsMeta, RegexUnsupportedException, TernaryExprMeta}
 
 import org.apache.spark.sql.catalyst.expressions.{Expression, Literal, RegExpReplace}
-import org.apache.spark.sql.rapids.{GpuRegExpReplace, GpuRegExpUtils, GpuStringReplace}
+import org.apache.spark.sql.rapids.{GpuRegExpReplace, GpuRegExpUtils, GpuStringReplace, RegexReplaceMode}
```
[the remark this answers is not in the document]
andygrove: Thanks. Fixed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -121,6 +121,10 @@ trait GpuConditionalExpression extends ComplexTypeMergingExpression with GpuExpr
     // predicate boolean array results in the two T values mapping to
     // indices 0 and 1, respectively.
 
+    // [F, null, T, F, T]
+    // [0, 0, 0, 1, 1]
+    [ 0, 1 ]
+
```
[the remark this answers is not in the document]
andygrove: I didn't realize I had local changes from a pairing session yesterday. I reverted the commit.

revans2 (approved):
Looks good to me

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
ttnghia: But we don't have to, right? I don't see any benefit from keeping it a `TernaryExpr` instead of just `UnaryExpr`/`GpuExpr`. I tried to implement `GpuStringToMap` to inherit `GpuExpression` and the evaluation function is super short: https://github.com/NVIDIA/spark-rapids/pull/4636/files#diff-a12810882b81a4eb395c03a80951f96ec080db793ffed6755739eeb2122840ccR1507-R1518

ttnghia: Basically we have evaluated the literal delimiter pattern before calling to the Gpu override, thus we only pass in ONE input string expression.

andygrove: The delimiter expression already isn't used in the evaluation. It is only referenced in `override def second: Expression = regex` which is just used to construct `children` in `final def children: Seq[Expression] = IndexedSeq(first, second, third)`.

I'm not against making the change and am curious to see what the benefits are but I would rather do this as a follow-on issue and review how similar regexp expressions are implemented since they all follow this same pattern.

revans2:
build

PR #4715, opened 2022-02-07
Patterns such (3?)+ should now fall back to CPU

### LINKED ISSUE #4487: [BUG] regexp_find hangs with some patterns
**Describe the bug**

Add this test to `RegularExpressionTranspilerSuite`.

```scala
test("adhoc") {
  val patterns = Seq("(3?)+")
  val inputs = Seq("3aZ")
}
```

**Expected behavior**
This should not hang. We either need to fix the underlying bug in cuDF, or we need to detect cases where this happens and fall back to CPU.

None

**Additional context**
Related cuDF Issue: https://github.com/rapidsai/cudf/issues/10006

NVnavkumar:
Partial fix for #4487.

Previously patterns like `(3?)+` or `(\A)+` would hang in `libcudf`. In some cases, that has now been corrected upstream and will now thrown an exception as an unsupported pattern. This fix will intercept these patterns in the transpiler to throw the appropriate `RegexUnsupportedException` 

Fuzz testing currently brings up more possible edge cases that break in the plugin, so this is only considered a partial fix at moment. Further testing can be performed to finally handle the remaining edge cases.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -610,6 +610,23 @@ class CudfRegexTranspiler(replace: Boolean) {
           // example: "a*+"
           throw new RegexUnsupportedException(nothingToRepeat)
 
+        case (RegexGroup(capture, term), SimpleQuantifier(ch)) if "+*".contains(ch) =>
+          // example: "(3?)+"
+          def isSimpleRepetition(e: RegexAST):Boolean = {
+            e match {
+              case RegexRepetition(term, quantifier) =>
+                term.isInstanceOf[RegexCharacterClassComponent]
```
andygrove, line 618: We can use pattern matching more fully here and avoid the `isInstanceOf`. Also, `quantifier` is unused here.

```suggestion
              case RegexRepetition(_: RegexCharacterClassComponent, _) => true
```

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -119,6 +119,13 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
     )
   }
 
+  test("cuDF does not support single repetition both inside and outside of capture groups") {
+    // see https://github.com/NVIDIA/spark-rapids/issues/4487
+    val patterns = Seq("(3?)+", "(3?)*", "(3*)+")
```
andygrove, line 124: Could we test with nested groups as well, like `((3?))+` ?

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -489,10 +509,18 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
       val cpu = cpuReplace(javaPattern, input)
       val cudfPattern = new CudfRegexTranspiler(RegexReplaceMode).transpile(javaPattern)
       val gpu = try {
-        gpuReplace(cudfPattern, input)
+        val result = Future {
+          gpuReplace(cudfPattern, input)
+        }
+        Await.result(result, 10 second)
```
andygrove, line 515: Is there a reason why the earlier test has a different timeout from this test? Are these timeouts lenient enough to cope with temporary slowdowns (perhaps due to garbage collection pauses) when these tests are running in CI?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -467,6 +467,42 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
+      case RegexChar(a) if "$^".contains(a) =>
+        // example: "$*"
+        false
+
+      case RegexRepetition(_, _) =>
+        // example: "a*+"
+        false
+
+      case RegexSequence(parts) =>
+        parts.forall(x => isSupportedRepetitionBase(x))
```
andygrove, line 496: nit: we should be able to simplify this

```suggestion
        parts.forall(isSupportedRepetitionBase)
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -628,20 +664,29 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
-        case (RegexRepetition(_, _), _) =>
-          // example: "a*+"
+        case (RegexGroup(_, term), SimpleQuantifier(ch))
+            if "+*".contains(ch) && !isSupportedRepetitionBase(term) =>
           throw new RegexUnsupportedException(nothingToRepeat)
-
-        case _ =>
+        case (RegexGroup(_, term), QuantifierVariableLength(_,None))
+            if !isSupportedRepetitionBase(term) =>
+            // specifically this variable length repetition: \A{2,}
```
andygrove, line 682: nit: indent is off for this comment and it should be aligned with the `throw` rather than the match guard

NVnavkumar:
build

andygrove (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -467,6 +467,39 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
+      case RegexRepetition(_, _) => true
+      case RegexGroup(_, term) => isRepetition(term)
+      case RegexSequence(parts) if parts.nonEmpty => isRepetition(parts.last)
+      case _ => false
+    }
+  }
+
+  private def isSupportedRepetitionBase(e: RegexAST): Boolean = {
+    e match {
+      case RegexEscaped(ch) if ch != 'd' && ch != 'w' => // example: "\B?"
```
jlowe, line 481: Nit: Comments should be consistent
```suggestion
      case RegexEscaped(ch) if ch != 'd' && ch != 'w' =>
        // example: "\B?"
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -628,20 +661,29 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
-          throw new RegexUnsupportedException(nothingToRepeat)
+        case (_, QuantifierVariableLength(0,None)) if mode == RegexReplaceMode =>
+          // see https://github.com/NVIDIA/spark-rapids/issues/4468
+          throw new RegexUnsupportedException(
+            "regexp_replace on GPU does not support repetition with {0,}")
 
-        case (RegexChar(a), _) if "$^".contains(a) =>
-          // example: "$*"
-          throw new RegexUnsupportedException(nothingToRepeat)
+        case (_, QuantifierFixedLength(0)) | (_, QuantifierVariableLength(0,Some(0)))
```
jlowe, line 669:
```suggestion
        case (_, QuantifierFixedLength(0)) | (_, QuantifierVariableLength(0, Some(0)))
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -628,20 +661,29 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
+            "regex_replace and regex_split on GPU do not support repetition with {0} or {0,0}")
 
-        case (RegexRepetition(_, _), _) =>
-          // example: "a*+"
+        case (RegexGroup(_, term), SimpleQuantifier(ch))
+            if "+*".contains(ch) && !isSupportedRepetitionBase(term) =>
           throw new RegexUnsupportedException(nothingToRepeat)
-
-        case _ =>
+        case (RegexGroup(_, term), QuantifierVariableLength(_,None))
```
jlowe, line 677:
```suggestion
        case (RegexGroup(_, term), QuantifierVariableLength(_, None))
```

jlowe (approved):
Minor style nits that aren't must-fix, lgtm.

Issue #4186, opened 2021-11-22
[FEA] combine compatible scalar subqueries


revans2:
This is something that we really should look into doing for Spark itself and not as a one off in the plugin.

TPC-DS query 9 will do a large number of scalar sub-queries to produce the final output.  It actually ends up running about 16 jobs in total, 15 of which are scalar sub-queries.

A ScalarSubquery is a subquery that produces a single result (one column and one row).  In most cases these are things like.

```sql
select IF(A > select avg(A) from my_table, "ABOVE AVERAGE", "NORMAL") from my_table
```

Here it is doing a reduction for on column A to find the average value and then trying to see if this particular column is above the average for the entire table.  To do this Spark will run `select avg(A) from my_table` as a separate query, and then collect the result back to the driver and insert it into a `ScalarSubquery` expression that returns the result of running the sub-query.

This is all great when there is just one sub-query. But if there are a lot of them, like with TPC-DS query 9. You can end up with a lot of overhead launching multiple jobs. This includes reading the footers on the input files. Possibly reading in overlapping column data, etc.

I have found that if I manually combine the sub-queries based off of the input data and the where clause in the sub-query I can reduce the number of sub-query jobs that would need to run 6 instead of 16. i.e.

```sql
select IF(select COUNT(A) from my_table > 100, select AVG(B) from my_table, select AVG(c) from my_table) from my_table
```

Instead of running 3 separate sub-queries `select COUNT(A) from my_table`, `select AVG(B) from my_table`, and `select AVG(c) from my_table` we would re-write it to be a single query. `select COUNT(A), AVG(B), AVG(c) from my_table`, and then we could have an alternative version of ScalarSubquery that knows how to access different columns of the returned data.  I predict that this would be able to cut the time of query 9 in half.

There are more complex things that we can do to combine more sub-queries together if the where clause of a reduction is not the same, and some things depending on how they overlap. This can involve things like putting an IF expression in the reductions based off of the where clause.  The problem that I run into with this is that on the GPU we end up computing the where clause multiple times and it can be very slow compared to the CPU.  I will file a separate issue to understand that.

PR #4726, opened 2022-02-09
Optimizer rule to combine compatible scalar subqueries

sperlingxx:
As mentioned in #4186, the propose of current PR is to bring up a new optimizer rule to combine compatible scalar subqueries, which is supposed to speed up specific queries by a large scale, such as TPC-DS query 9.

This PR is only a prototype for verification and further discussion. Ideally, the rule is supposed to be added as a common rule of Apache Spark if it is indeed necessary. 

The primary work of this PR is the rule `CombineScalarSubquery`, whose basic idea is combining compatible aggregate plans served for scalar-subquery as much as possible. Aggregates sharing the same "base plan" with other aggregates are considered as compatible. This rule is supposed to  combine all groups of compatible aggregates. And create a unified fused aggregate for each group, which can serve multiple scalar subqueries. The "base plan" refers to the top non-project descendant of the Aggregate, which means it could be either the child of Aggregate or the child of Aggregate(Project(...)).

In general, `CombineScalarSubquery` consists of three steps:
1. Collect combination candidates from scalar sub-queries. Assign these candidates into different groups according to the canonicalized base plan.
2. Combine collected candidates of each group with an unified base plan to build fused aggregates.
3. Replace ScalarSubqueries with SharedScalarSubqueries based on fused aggregates.

Here is an example:
```sql
SELECT SUM(i) FROM t
WHERE l > (SELECT MIN(l2) FROM t) AND l2 < (SELECT MAX(l) FROM t)
AND AND i2 <> (SELECT MAX(i2) FROM t) AND i2 <> (SELECT MIN(i2) FROM t)
```
Before the apply of `CombineScalarSubquery`, the optimized logicial plan looks like:
```
Aggregate [sum(i)]
+- Project [i]
  +- Filter (((l > scalar-subquery#1) AND (l2 < scalar-subquery#2)) AND (NOT (i2 = scalar-subquery#3) AND NOT (i2 = scalar-subquery#4)))
     :  :- Aggregate [min(l2)]
     :  :  +- Project [l2]
     :  :     +- Relation [l,l2,i,i2]
     :  +- Aggregate [max(l)]
     :     +- Project [l]
     :        +- Relation [l,l2,i,i2]
     :  +- Aggregate [max(i2)]
     :     +- Project [l]
     :        +- Relation [l,l2,i,i2]
     :  +- Aggregate [min(i2)]
     :     +- Project [l]
     :        +- Relation [l,l2,i,i2]
     +- Relation [l,l2,i,i2]
```
After the apply of `CombineScalarSubquery`:
```
 Aggregate [sum(i)]
 +- Project [i]
   +- Filter (((l > shared-scalar-subquery#1) AND (l2 < shared-scalar-subquery#2)) AND (NOT (i2 = shared-scalar-subquery#3) AND NOT (i2 = shared-scalar-subquery#4)))
      :  :- Aggregate [min(l2),max(l),max(i2),min(i2)]
      :  :  +- Project [l2,l,i2]
      :  :     +- Relation [l,l2,i,i2]
      :  :- Aggregate [min(l2),max(l),max(i2),min(i2)]
      :  :  +- Project [l2,l,i2]
      :        +- Relation [l,l2,i,i2]
      :  :- Aggregate [min(l2),max(l),max(i2),min(i2)]
      :  :  +- Project [l2,l,i2]
      :        +- Relation [l,l2,i,i2]
      :  :- Aggregate [min(l2),max(l),max(i2),min(i2)]
      :  :  +- Project [l2,l,i2]
      :        +- Relation [l,l2,i,i2]
      +- Relation [l,l2,i,i2]
```
In this case, all four sub-queries are compatible. Therefore, they can share the same fused aggregate. During the preparation of physical plan, these fuesd aggregates will be detected and reused by the rule  `ReuseExchangeAndSubquery`.


sperlingxx:
Hi @revans2, how do you feel about the current approach? And I am not sure if I should create a Spark issue about this topic.

PR #4731, opened 2022-02-09
Trim join tests to improve runtime of tests

jlowe:
Relates to #4538.

This trims quite a bit of test combinations from the join integration tests.  Joins test on many dimensions such as join type, data type, and batch size.  Some joins can take a long time, and we run them twice (CPU+GPU), so doing every combination contributes to the very long runtime for all the join tests.

The tests were trimmed with the following guidelines:
- It does not seem necessary for special-case count joins to test every possible input type since it's just a count, so those tests were updated to use a single input type.
- It does not seem necessary to test every combination of join type and data type on small batch sizes when the same combination of join type and data type are being tested at larger sizes.  Dealing with small batch sizes is more about managing gather maps and calling gather appropriately than the actual data types themselves.  In addition, small batch sizes can take quite a long time for some joins. Therefore the tests were updated to only test a small subset of input types on small batch combinations.
- There were many decimal types being tested that added no extra code coverage, since joins don't care about decimal scale settings.  The decimal types were trimmed to test one precision/scale combination for 32-bit, 64-bit, and 128-bit decimal types, respectively.

### DIFF
--- a/integration_tests/src/main/python/join_test.py
+++ b/integration_tests/src/main/python/join_test.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import pytest
+from _pytest.mark.structures import ParameterSet
 from pyspark.sql.functions import broadcast
 from pyspark.sql.types import *
 from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_fallback_collect, assert_cpu_and_gpu_are_equal_collect_with_capture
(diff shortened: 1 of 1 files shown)


revans2 (approved):
This looks good to me. We still cover all of the types we covered before (with some fewer decimal types). We also cover the smaller number of batches better.

PR #4735, opened 2022-02-09
Re-enables partial regex support for octal digits on the GPU

### LINKED ISSUE #4409: [BUG] Possible race condition in regular expression support for octal digits
**Describe the bug**

I am seeing odd results with regular expressions containing octal digits. During testing I seem to be able to trigger a race condition where the first call to `containsRe` will produce `false` and subsequent calls will return `true`. Here is a test that reproduces the issue (to be added to `RegularExpressionTranspilerSuite`).

```scala
test("possible race condition") {
  for (_ <- 0 until 100) {
    println(gpuContains("\\101", Seq("A")).mkString(", "))
  }
}
```

For reference, here is the `gpuContains` method. I Added some `println` statements here.

```scala
private def gpuContains(cudfPattern: String, input: Seq[String]): Array[Boolean] = {
  println(cudfPattern)
  println(input.mkString(", "))
  val result = new Array[Boolean](input.length)
  withResource(ColumnVector.fromStrings(input: _*)) { cv =>
    withResource(cv.containsRe(cudfPattern)) { c =>
      withResource(c.copyToHost()) { hv =>
        result.indices.foreach(i => result(i) = hv.getBoolean(i))
      }
    }
  }
  result
}
```

Test output:

```
\101
A
false
\101
A
true
\101
A
true
...
```

If I remove the `println` statements from the `gpuContains` method then the result is always `true`.

See above.

**Expected behavior**
Result should be the same on every call.

N/A

**Additional context**
N/A

NVnavkumar:
Fixes #4409.

Since a couple of underlying `libcudf` issues with octal digit support in regular expression have now been resolved, this code re-enables octal digit support with the appropriate transpiling from Java Pattern support to libcudf regular expression format.  A couple of known limitations:

* octal digits from `\200` to `\377` are not supported by `libcudf`, so this will still fallback to CPU
* octal digits are not currently supported in character classes (e.g. `[\022]`) in `libcudf`, so this will also fallback to CPU

Documentation has been updated to reflect these limitations.


revans2:
Just a small question. If we can parse them properly why not switch them over to something that CUDF does support? like hex digits?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -730,6 +747,8 @@ sealed case class QuantifierVariableLength(minLength: Int, maxLength: Option[Int
 
 sealed trait RegexCharacterClassComponent extends RegexAST
 
+// TODO: Consider merging some of these subclasses to handling escaping
```
revans2, line 750: Is this TODO needed?  Are we doing to do something with it? Is there a follow on issue that needs to be filed?

NVnavkumar:
> Just a small question. If we can parse them properly why not switch them over to something that CUDF does support? like hex digits?

CUDF actually also doesn't support hex digits in the range that corresponds to 128-255

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
NVnavkumar: oops, forgot to remove an old comment, fixed

revans2 (approved):

NVnavkumar:
build

andygrove (approved):
LGTM. I am not sure why the build failed but it seems unrelated.

andygrove:
build

PR #4740, opened 2022-02-09
Enable regular expressions on GPU by default

andygrove:

This PR adds a new config option `spark.rapids.sql.regexp.enabled` which defaults to `true` and enables all existing regular expression support on GPU, such as `RLike`, `RegExp`, `RegExpReplace`, and `RegExpExtract`. This replaces the previous approach of disabling individual expression classes by default.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
@@ -821,6 +821,10 @@ class GpuRLikeMeta(
     private var pattern: Option[String] = None
 
     override def tagExprForGpu(): Unit = {
+      if (!conf.isRegExpEnabled) {
+        willNotWorkOnGpu(s"Regular expression support is disabled. " +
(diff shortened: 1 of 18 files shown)


revans2 (dismissed):
Just a nit on the doc comments.  Also the check appears to be the same everywhere and it might be nice to have common code for it, but that is really minor.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -959,6 +967,10 @@ class GpuRegExpExtractMeta(
   private var numGroups = 0
 
   override def tagExprForGpu(): Unit = {
+    if (!conf.isRegExpEnabled) {
+      willNotWorkOnGpu(s"Regular expression support is disabled. " +
+        s"Set ${RapidsConf.ENABLE_REGEXP.key}=true to enable it")
+    }
```
jlowe, line 973:
```suggestion
    GpuRegExpUtils.tagForRegExpEnabled(this)
```

jlowe (approved):

Issue #4727, opened 2022-02-09
[BUG] Read Parquet does not ignoreCorruptFiles


chawkm:
**Describe the bug**
On a CPU cluster when we set the following:
spark.conf.set("spark.sql.files.ignoreCorruptFiles", true)
Parquet files are then read and corrupt files are skipped over.

On GPU cluster with the same config corrupt files cause an exception to occur:
java.lang.RuntimeException:
`... is not a Parquet file expected magic number at tail 80 65 82 49 but found 00 00 00 00`

**Steps/Code to reproduce bug**
Please provide a list of steps or a code sample to reproduce the issue.
Avoid posting private or sensitive data.
1. Add a corrupt file to a directory of parquet files
2. Try and read directory of parquet files with ignoreCorruptFiles=true

**Expected behavior**
Corrupt files should be ignored.

**Environment details (please complete the following information)**
 - Environment location: [Standalone, YARN, Kubernetes, Cloud(specify cloud provider)]
 - Spark configuration settings related to the issue

Databricks: 9.1 LTS ML (includes Apache Spark 3.1.2, GPU, Scala 2.12)
Init Script:
#!/bin/bash
sudo wget -O /databricks/jars/rapids-4-spark_2.12-21.12.0.jar https://repo1.maven.org/maven2/com/nvidia/rapids-4-spark_2.12/21.12.0/rapids-4-spark_2.12-21.12.0.jar
sudo wget -O /databricks/jars/cudf-21.12.2-cuda11.jar https://repo1.maven.org/maven2/ai/rapids/cudf/21.12.2/cudf-21.12.2-cuda11.jar

spark config:
spark.task.resource.gpu.amount 0.0625
spark.sql.files.ignoreCorruptFiles true
executor.extraJavaOptions -Dlog4j2.formatMsgNoLookups=true
spark.plugins com.nvidia.spark.SQLPlugin
fs.s3a.acl.default BucketOwnerFullControl
spark.locality.wait 0s
fs.s3n.impl.disable.cache true
driver.extraJavaOptions -Dlog4j2.formatMsgNoLookups=true
spark.rapids.memory.pinnedPool.size 2G
spark.sql.adaptive.enabled false
fs.s3.impl.disable.cache true
fs.s3a.impl.disable.cache true
spark.databricks.delta.optimizeWrite.enabled false
spark.rapids.sql.concurrentGpuTasks 2

**Additional context**

jlowe:
[...]

PR #4742, opened 2022-02-10
Add ignoreCorruptFiles feature for Parquet reader [databricks]

wbo4958:
To close https://github.com/NVIDIA/spark-rapids/issues/4727.

GPU parquet reader will throw an exception when reading some corrupt parquet files even if the user has set `"spark.sql.files.ignoreCorruptFiles"` to true for `coalescing` or `multi-threaded` reading. While `PERFILE` reading is ok since it is in the wrapper of `FilePartitionReader` which has handled the exceptions. see https://github.com/apache/spark/blob/branch-3.2/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2/FilePartitionReader.scala#L39-L59

For multi-threaded reading, when detecting the `wrong footer` or `failing to copy the files` into the HostMemoryBuffer, it will totally ignore the whole file. For example, assuming there are 3 row groups,  the multi-threaded reader has successfully copied the first 2 row groups into the HostMemoryBuffer, and it failed to copy the 3rd row group caused by IOException. The multi-threaded reader will ignore the first loaded 2 row groups,  return empty HostMemoryBuffer. But for CPU, it may have returned the first 2 row groups to users. 

For coalescing reading, we have first calculated the offset and size for each file and launched multi-threads to copy the row groups into the `fixed location` of the HostMemoryBuffer. Assuming we're coalescing 1.parquet, 2.parquet and 3.parquet files. First, we have allocated a total HostMemoryBuffer for 1.parquet, 2.parquet and 3.parquet and launched 3 threads to copy 1.parquet, 2.parquet and 3.parquet into the total HostMemoryBuffer separately. Asssuming the 2.parquet is failed to copying or something. Then the re-generated parquet file will only contain 1.parquet, empty hole (for 2.parquet, but failed to copy, we just left the empty hole), 3.parquet

This PR only considers the exceptions that happened in reading footer and copying row groups. It didn't catch the exception when decoding in GPU. So, even if ignoreCorruptFiles is enabled, it will throw an exception when cudf failed to decode the host buffer.

### DIFF
--- a/integration_tests/src/main/python/parquet_test.py
+++ b/integration_tests/src/main/python/parquet_test.py
@@ -20,7 +20,7 @@ from marks import *
 from pyspark.sql.types import *
 from pyspark.sql.functions import *
 from spark_session import with_cpu_session, with_gpu_session, is_before_spark_330
-from src.main.python.conftest import is_databricks_runtime
+from conftest import is_databricks_runtime
 
 
 def read_parquet_df(data_path):
--- a/integration_tests/src/main/python/parquet_test.py
+++ b/integration_tests/src/main/python/parquet_test.py
@@ -20,6 +20,8 @@ from marks import *
 from pyspark.sql.types import *
 from pyspark.sql.functions import *
 from spark_session import with_cpu_session, with_gpu_session, is_before_spark_330
+from src.main.python.conftest import is_databricks_runtime
+
 
 def read_parquet_df(data_path):
     return lambda spark : spark.read.parquet(data_path)
@@ -755,6 +757,7 @@ def test_parquet_scan_with_hidden_metadata_fallback(spark_tmp_path, metadata_col
 @ignore_order
 @pytest.mark.parametrize('v1_enabled_list', ["", "parquet"])
 @pytest.mark.parametrize('reader_confs', reader_opt_confs, ids=idfn)
+@pytest.mark.skipif(is_databricks_runtime(), reason="Databricks does not support ignoreCorruptFiles")
 def test_parquet_read_with_corrupt_files(spark_tmp_path, reader_confs, v1_enabled_list):
     first_data_path = spark_tmp_path + '/PARQUET_DATA/first'
     with_cpu_session(lambda spark : spark.range(1).toDF("a").write.parquet(first_data_path))
--- a/integration_tests/src/main/python/parquet_test.py
+++ b/integration_tests/src/main/python/parquet_test.py
@@ -750,3 +750,23 @@ def test_parquet_scan_with_hidden_metadata_fallback(spark_tmp_path, metadata_col
         do_parquet_scan,
         exist_classes= "FileSourceScanExec",
         non_exist_classes= "GpuBatchScanExec")
+
+
+@ignore_order
+@pytest.mark.parametrize('v1_enabled_list', ["", "parquet"])
+@pytest.mark.parametrize('reader_confs', reader_opt_confs, ids=idfn)
+def test_parquet_read_with_corrupt_files(spark_tmp_path, reader_confs, v1_enabled_list):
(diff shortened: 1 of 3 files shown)


wbo4958:
build

wbo4958:
We still need to do the same thing for text-based file formats like ORC/JSON

wbo4958:
build

wbo4958:
CI failed on Databricks env. will check it

jlowe (changes requested):
There's quite a bit missing from this PR.  It currently only addresses Parquet, but there are many other readers that need to support this config setting (or preclude being on the GPU if it is set).  In addition, the error handling only covers the small section of the read where we are examining the footer, but we could get I/O errors or other exceptions when trying to read the row group data, decode it on the GPU, etc.

wbo4958:
> There's quite a bit missing from this PR. It currently only addresses Parquet, but there are many other readers that need to support this config setting (or preclude being on the GPU if it is set). In addition, the error handling only covers the small section of the read where we are examining the footer, but we could get I/O errors or other exceptions when trying to read the row group data, decode it on the GPU, etc.

Thx Jason. The new commits have considered the situation reading footer and copying row group, but it didn't catch the exception when CUDF failed to decode the buffer. The exception CUDF failed to decode may caused by CUDF code itself, so I tend to throw this exception.

And as to other readers like ORC/JSON, I'd like to have some follow ups.

jlowe:
> The exception CUDF failed to decode may be caused by CUDF code itself, so I tend to throw this exception.

The same "but it might be a bug in the code" logic applies whether we're talking about cudf or the Spark/Parquet code.  I'm OK with addressing this in a followup because it's a bit tricky to do (and may require some assistance within cudf), but we need to file a tracking issue for it since this isn't completely covered.

> And as to other readers like ORC/JSON, I'd like to have some follow ups.

That's fine, but then please file the tracking issues and link to the original issue (or a new epic issue per discussion below).

This PR is marked as closing the original issue, so please either unlink the issue as we will not properly support ignoring corrupt files with just this PR or file a separate issue (maybe an epic) to track fully implementing this feature.

wbo4958:
Yeah, It's kind of tricky, but I don' have more ideas on how to do that. I think it's really not reasonable to consider a "correct" file into a corrupted file caused by cudf/spark-rapids code itself.

Hi @jlowe, I'd like to hear your opinions.

wbo4958:
build

wbo4958:
After discussion with @GaryShen2008, we think we should fall back coalescing reading to cloud reading when ignoreCorruptedFile is set. 

Coalescing reading is trying hard to coalesce many parquet files at a time. If one row group of one file is corrupted, we will ignore the whole coalesced files if we decide to try-catch cudf decoding. While PERFILE/Multithread readings only ignore the corrupted file.

wbo4958:
build

wbo4958:
build

jlowe (commented):
Overall the code looks OK for Parquet, but there still needs to be followup issues for ORC, ?CSV, JSON, etc.  Also it looks like Databricks may not support `ignoreCorruptFiles`, as the test failed on the _CPU_, not the GPU.  We may need to skip that test on Databricks.

wbo4958:
> Overall the code looks OK for Parquet, but there still needs to be followup issues for ORC, CSV, JSON, etc. Also it looks like Databricks may not support `ignoreCorruptFiles`, as the test failed on the _CPU_, not the GPU. We may need to skip that test on Databricks.

Yeah, just tested on databricks env, which will throw exception on CPU. But with this PR, the corrupt files will be ignored when ignoreCorruptFiles is set to true.

wbo4958:
build

wbo4958:
build

jlowe (approved):

Issue #4748, opened 2022-02-10
[BUG] Update qualification tool to remove DECIMAL 128 as potential problem


tgravescs:
**Describe the bug**
22.02 supports decimal 128, we need to remove it from the qualification tool as potential problem.

PR #4752, opened 2022-02-11
Remove Decimal as potential problem in qualification tool

tgravescs:
We now support decimal 128 and decimals are on by default (as of 22.02).  The qualification tool shouldn't report them as potential problem anymore.

Tests updated to reflect this.

fixes https://github.com/NVIDIA/spark-rapids/issues/4748


jlowe (dismissed):
Looks OK to me, but should this target branch-22.02?

nartal1 (approved):

PR #4754, opened 2022-02-11
Close SpillableBatch to avoid leaks

### LINKED ISSUE #4730: [BUG] python UDF tests are leaking
**Describe the bug**
As a part of trying to debug host memory usage I found that the UDF tests are leaking badly. Specifically udf_test.py and udf_cudf_test.py

Build the plugin and run some of the tests in there with `-Dai.rapids.refcount.debug=true` then look at the results of what happens.

For example:
```
PYSP_TEST_spark_master='local[1]' COVERAGE_SUBMIT_FLAGS=-Dai.rapids.refcount.debug=true TEST_PARALLEL=0 SPARK_HOME=$HOME/src/spark_3.2.0/ ./run_pyspark_from_build.sh -k 'udf_test.py and test_group_agg and not more_typesregate_udf and Byte and not more_types' -s 2>&1 | tee single_udf.txt
```
Showed that 48 column vectors were leaked and one host buffer was leaked.  This is just one test, there are 137 tests in those two files and when I run all of them I see over 20,615 items being leaked.

**Expected behavior**
There are no items that are leaked.

revans2:
this fixes #4730 
I did a bit of cleanup for some of the code, but the leaks really were that the SpillableColumnBatch instances that were inserted into the queue were never closed when they were dequeued.

### DIFF
--- a/integration_tests/run_pyspark_from_build.sh
+++ b/integration_tests/run_pyspark_from_build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2020-2021, NVIDIA CORPORATION.
+# Copyright (c) 2020-2022, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


revans2:
build

jlowe (approved):

PR #4757, opened 2022-02-11
Update qualification tool to remove DECIMAL 128 as potential problem

tgravescs:
We now support decimal 128 and decimals are on by default (as of 22.02). The qualification tool shouldn't report them as potential problem anymore.

Tests updated to reflect this.

fixes https://github.com/NVIDIA/spark-rapids/issues/4748


tgravescs:
actually there is probably doc updates I need here too.

nartal1 (approved):

tgravescs:
build

revans2 (approved):

jlowe (approved):

PR #4760, opened 2022-02-11
Short circuit AND/OR in ANSI mode

### LINKED ISSUE #4526: [BUG] Short circuit AND/OR in ANSI mode
**Describe the bug**
AND/OR on the CPU in Spark will process the [LHS unconditionally](https://github.com/apache/spark/blob/99805558fc80743747f32c7008cb7cc99c1cda01/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala#L740).  But they will only process [the RHS](https://github.com/apache/spark/blob/99805558fc80743747f32c7008cb7cc99c1cda01/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala#L743-L744) if they cannot figure out the result from just the lhs.

On the GPU we process both unconditionally, but in ANSI mode we can run into problems just like with #3849. We should do a similar fix for AND/OR like we did for IF/ELSE, and follow what Spark does on the CPU. For AND if the RHS has side effects we only process it for rows where the LHS is not false (this includes nulls).  For OR if the RHS has side effects we only process it for rows where the LHS is not true (this includes nulls).

amahussein:
fixes #4526 

**Compatibility:**

- For AND if the RHS has side effects we only process it for rows where the LHS is not false (this includes nulls).
- For OR if the RHS has side effects we only process it for rows where the LHS is not true (this includes nulls).

**Testing:**

- Added two integration tests in logic_test.py

**Code changes:**

- Override the implementation of `columnarEval` for both `GpuAnd` and `GpuOR`
- The above change requires some refactoring to `ConditionalExpressions` since the new code needs to use some of the methods defined there such as: `gather`, `filterBatch`,


integration_tests/src/main/python/logic_test.py
```diff
@@ -14,36 +14,87 @@
...
+# AND/OR on the CPU in Spark will process the LHS unconditionally. But they will only
+# process the RHS if they cannot figure out the result from just the LHS.
+# Tests the GPU short-circuits the predicates without throwing Exception in ANSI mode.
+@pytest.mark.skip
+@pytest.mark.parametrize('ansi_enabled', ['false', 'true'])
+@pytest.mark.parametrize('lhs_predicate', [False, True])
+@pytest.mark.parametrize('expr', _get_arithmetic_overflow_expr('OR'))
+def test_or_with_side_effect(expr, ansi_enabled, lhs_predicate):
+    ansi_conf = {'spark.sql.ansi.enabled': ansi_enabled}
+    if ansi_enabled == 'true' and (not(lhs_predicate)):
```
jlowe, line 92:
```suggestion
    if ansi_enabled == 'true' and not lhs_predicate:
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -66,6 +67,125 @@ case class GpuAnd(left: Expression, right: Expression) extends CudfBinaryOperato
 
   override def binaryOp: BinaryOp = BinaryOp.NULL_LOGICAL_AND
   override def astOperator: Option[BinaryOperator] = Some(ast.BinaryOperator.NULL_LOGICAL_AND)
+
+  protected def filterBatch(
```
jlowe, line 71: Rather than copying all this code, I expected `GpuAnd` to use the existing `GpuConditionalExpression` trait.

amahussein: Yes absolutely agree.
I was thinking to create a new trait for helpers related to the ColumnVector since those methods are not really specific to "ConditionalExpression". WDYT?

integration_tests/src/main/python/logic_test.py
```diff
@@ -14,36 +14,87 @@
...
+    ).selectExpr(expr)
+
+def _get_arithmetic_overflow_expr(op) :
+    return ['a {} (CAST(b as INT) + 2) > 0'.format(op)]
+
+# AND/OR on the CPU in Spark will process the LHS unconditionally. But they will only
+# process the RHS if they cannot figure out the result from just the LHS.
+# Tests the GPU short-circuits the predicates without throwing Exception in ANSI mode.
+@pytest.mark.parametrize('ansi_enabled', ['true', 'false'])
+@pytest.mark.parametrize('lhs_predicate', [True, False])
```
gerashegalov, line 69: can we add a Null lhs predicate for completeness?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
jlowe: Maybe, although it starts to get outside the scope of the PR a bit.  If we are moving stuff around, `filterBatch` should be part of `GpuFilter` which does something almost identical to this method.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -66,6 +67,125 @@ case class GpuAnd(left: Expression, right: Expression) extends CudfBinaryOperato
...
+          val finalRet = withResourceIfAllowed(rEval) { _ =>
+            (lhsBool, rEval) match {
+              case (t: GpuColumnVector, f: GpuColumnVector) =>
+                withResource(gather(t.getBase, f)) { combinedVector =>
+                  t.getBase.and(combinedVector)
+                }
+              case (t: GpuColumnVector, f: GpuScalar) =>
+                t.getBase.and(f.getBase)         
+            }
+          }
```
andygrove, line 173: This can be simplied since `lhsBool` is of a fixed type, so need to match on that. We can also the [Pattern Matching Anonymous Function](https://www.scala-lang.org/files/archive/spec/2.13/08-pattern-matching.html#pattern-matching-anonymous-functions) syntax to make it a little more concise - this basically removes the need for the explicit `rEval match`.

```suggestion
          val finalRet = withResourceIfAllowed(rEval) {
            case r: GpuColumnVector =>
              withResource(gather(lhsBool.getBase, r)) { combinedVector =>
                lhsBool.getBase.and(combinedVector)
              }
            case r: GpuScalar =>
              lhsBool.getBase.and(r.getBase)
          }
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -57,26 +90,135 @@ case class GpuNot(child: Expression) extends CudfUnaryExpression
...
+          // replace nulll values
+          val lhsNoNulls = withResource(Scalar.fromBool(true)) { trueScalar =>
+            lhsBool.getBase.replaceNulls(trueScalar)
+          }
+          withResource(lhsNoNulls) { _ =>
+            val rEval = withResource(filterBatch(tbl, lhsNoNulls, colTypes)) { leftTrueBatch =>
+              rightExpr.columnarEval(leftTrueBatch)
+            }
+            val finalRet = withResourceIfAllowed(rEval) { _ =>
+              (rEval) match {
```
andygrove, line 156: The parentheses here are creating a tuple containing `rEval` and are not required. They will need removing from the `case` lines as well.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -35,6 +36,38 @@ trait GpuPredicateHelper {
   }
 }
 
+trait GpuCVPredicateHelper extends GpuColumnVectorHelper {
+  /**
+   * Overrides the default behavior in which null values are considered false.
+   */
+  override protected def isAllFalse(col: GpuColumnVector): Boolean = {
+    assert(BooleanType == col.dataType())
+    if (col.hasNull) {
```
revans2, line 45: Do we need the row count == 0 here too? If not why?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -35,6 +36,38 @@ trait GpuPredicateHelper {
...
+      return false
+    }
+    withResource(col.getBase.all()) { allFalse =>
```
revans2, line 48: The name of `allFalse` looks wrong. It looks like it should be `allTrue`, so we can say `isAllFalse = !allTrue`

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -57,26 +90,135 @@ case class GpuNot(child: Expression) extends CudfUnaryExpression
...
+              val rEval = withResource(filterBatch(tbl, lhsNoNulls, colTypes)) { leftInverseBatch =>
+                rightExpr.columnarEval(leftInverseBatch)
+              }
+              val finalRet = withResourceIfAllowed(rEval) { _ =>
+                (rEval) match {
+                  case (f: GpuColumnVector) =>
+                    withResource(gather(lhsNoNulls, f)) { combinedVector =>
+                      doColumnar(lhsBool, GpuColumnVector.from(combinedVector, dataType))
+                    }
+                  case (t: GpuColumnVector, f: GpuScalar) =>
```
andygrove, line 211: This seems incorrect and I think this was intended to be `case (f: GpuScalar)` ?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -57,26 +90,135 @@ case class GpuNot(child: Expression) extends CudfUnaryExpression
...
+              rightExpr.columnarEval(leftTrueBatch)
+            }
+            val finalRet = withResourceIfAllowed(rEval) { _ =>
+              (rEval) match {
+                case (f: GpuColumnVector) =>
+                  withResource(gather(lhsNoNulls, f)) { combinedVector =>
+                    doColumnar(lhsBool, GpuColumnVector.from(combinedVector, dataType))
+                  }
+                case (f: GpuScalar) =>
+                  doColumnar(lhsBool, f)
```
andygrove, line 162: This path is not covered by the current tests. Would we ever hit this path? I don't think that scalar expressions can have side-effects and we would only call this method if the RHS has side-effects.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
andygrove: This path is not covered by the current tests. Would we ever hit this path? I don't think that scalar expressions can have side-effects and we would only call this method if the RHS has side-effects.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
revans2: Wow that is an interesting situation. The only place right now where we can return a Scalar from processing when the input is not a foldable constant is some corner cases with Coalesce, and Coalesce can totally have side effects associated with it for down stream tasks that are a part of it.  Perhaps we should try to write a test for this just to be sure...

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
amahussein: That's a good question. By checking the implementation in `conditionalExpressions.scala::isAllTrue`, the case when rowCount is 0 is considered "True". For `conditionalExpressions.scala::isAllFalse`, when the number of rows is 0, then `if (col.getRowCount == col.numNulls())`  will be true.
In other words, empty rows will cause both isAllTrue and isAllFalse both to be true.

Back to GpuAnd/GpuOR, should I consider an empty column True or False?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -57,26 +90,135 @@ case class GpuNot(child: Expression) extends CudfUnaryExpression
...
+
+    withResource(GpuColumnVector.from(batch)) { tbl =>
+      withResource(GpuExpressionsUtils.columnarEvalToColumn(leftExpr, batch)) { lhsBool =>
+        if (isAllFalse(lhsBool)) {
+          withResource(GpuScalar.from(false, dataType)) { falseScalar =>
+            GpuColumnVector.from(falseScalar, lhsBool.getRowCount.toInt, BooleanType)
+          }
+        } else {
+          // replace nulll values
+          val lhsNoNulls = withResource(Scalar.fromBool(true)) { trueScalar =>
```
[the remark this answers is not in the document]
amahussein: I looked up all references to `columnVector.replaceNulls()` and I did not see it guarded by check for nulls. So, I decided to follow the same pattern thinking that `replaceNulls()` is not expensive.
I fixed that in the recent commits.

integration_tests/src/main/python/logic_test.py
amahussein: Good Suggestion! Done!

integration_tests/src/main/python/logic_test.py
```diff
@@ -14,36 +14,87 @@
...
+            conf=ansi_conf)
+
+# AND/OR on the CPU in Spark will process the LHS unconditionally. But they will only
+# process the RHS if they cannot figure out the result from just the LHS.
+# Tests the GPU short-circuits the predicates without throwing Exception in ANSI mode.
+@pytest.mark.skip
+@pytest.mark.parametrize('ansi_enabled', ['false', 'true'])
+@pytest.mark.parametrize('lhs_predicate', [False, True])
+@pytest.mark.parametrize('expr', _get_arithmetic_overflow_expr('OR'))
+def test_or_with_side_effect(expr, ansi_enabled, lhs_predicate):
```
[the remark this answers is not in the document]
amahussein: Done!
Initially, when I looked at the existing python tests, I thought that there is preference to keep separate tests for readability.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
amahussein: I could not add `GpuConditionalExpression ` out of the box because of conflicts between the traits tree. So, I pulled the generic methods into a new trait that can be extended by the GpuAnd/GpuOr.
Another approach would be to move that to Helper object, but I did not like that because it would cause more scattered changes in `conditionalExpressions.scala`

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -57,26 +90,135 @@ case class GpuNot(child: Expression) extends CudfUnaryExpression
...
+            val lhsNoNulls = withResource(Scalar.fromBool(true)) { trueScalar =>
+              lhsInverted.replaceNulls(trueScalar)
+            }
+            // filter batch based on the inverse of the LHS
+            withResource(lhsNoNulls) { _ =>
+              val rEval = withResource(filterBatch(tbl, lhsNoNulls, colTypes)) { leftInverseBatch =>
+                rightExpr.columnarEval(leftInverseBatch)
+              }
+              val finalRet = withResourceIfAllowed(rEval) { _ =>
+                (rEval) match {
```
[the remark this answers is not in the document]
amahussein: Done!

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
amahussein: Done!

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
andygrove: I've been trying to write a test that hits the scalar path and have failed to come up with anything so far. It looks like Coalesce will only return a scalar if the first non-null parameter is a scalar, but Spark will optimize the Coalesce out in that case. I am back to wondering if the scalar path is even possible, but maybe I am missing the edge case here.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -57,26 +88,123 @@ case class GpuNot(child: Expression) extends CudfUnaryExpression
...
   override def binaryOp: BinaryOp = BinaryOp.NULL_LOGICAL_OR
   override def astOperator: Option[BinaryOperator] = Some(ast.BinaryOperator.NULL_LOGICAL_OR)
+
+  override def columnarEvalWithSideEffects(batch: ColumnarBatch): Any = {
+    val leftExpr = left.asInstanceOf[GpuExpression]
+    val rightExpr = right.asInstanceOf[GpuExpression]
+    val colTypes = GpuColumnVector.extractTypes(batch)
+
+    withResource(GpuColumnVector.from(batch)) { tbl =>
+
```
andygrove, line 181: nit: remove blank line here

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -57,26 +88,123 @@ case class GpuNot(child: Expression) extends CudfUnaryExpression
...
+    val leftExpr = left.asInstanceOf[GpuExpression]
+    val rightExpr = right.asInstanceOf[GpuExpression]
+    val colTypes = GpuColumnVector.extractTypes(batch)
+
+    withResource(GpuColumnVector.from(batch)) { tbl =>
+      withResource(GpuExpressionsUtils.columnarEvalToColumn(leftExpr, batch)) { lhsBool =>
+        if (isAllFalse(lhsBool)) {
+          shortCircuitWithBool(lhsBool, false)
+        } else {
+          // Replace nulll values (if any) with true.
```
andygrove, line 143:
```suggestion
          // Replace null values (if any) with true.
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
revans2: I don't think Spark had an optimization in older versions, so you might try that.  Also this is enough of a corner case that I am fine if it as is.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
amahussein: Done!

amahussein:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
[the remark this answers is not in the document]
revans2: I mostly wanted some kind of evidence if it is slow or not.  I think in this case my gut is probably okay, but it might be nice to have something better than that.  Do you have any benchmarks that we could test this with?  I know that AND/OR ended up being very slow int he past and caused measurable degradation in some TPC-DS queries.  So it should be possible to have a project with lots of ANDS/ORs in it and see.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -25,36 +25,8 @@ import org.apache.spark.sql.catalyst.expressions.{ComplexTypeMergingExpression,
...
-              throw new IllegalStateException(s"Unexpected inputs" +
-                s" ($t: ${t.getClass}, $f: ${f.getClass})")
-          }
-          GpuColumnVector.from(finalRet, dataType)
-        }
-      }
-    }
-  }
 
+trait GpuColumnVectorHelper extends Arm {
```
revans2, line 29: Can we find a better name for this? It was `GpuConditionalExpression` so it was clear that it was supported to be for use with those, but now it is a generic utility library with an even more generic name and no clear contract on how these APIs should be used.  Could we change this to be an object instead of a trait? Could we restore the name of `GpuConditionalExpression` so it is still clear that they are intended to be used with these types of classes? Also I really dislike that we are overriding `isAllTrue` that just does not appear to make and since to me.  If we have different ways of calculating `isAllTrue` then we should either have separate utility APIs with a name that makes it clear how each works, or we should have a flag passed to the API to allow you to select how it works.  Does that make since to you?

amahussein: Yes, that makes sense.
I initially thought to use the existing Object `GpuExpressionsUtils`. Should that be fine? Or do you prefer to create a new Object helper , and what would be its name?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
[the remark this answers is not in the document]
amahussein: I was hoping you never say that :-)
I will take a look into benchmarking. It may take me some time to get back with meaningful data though.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala
```diff
@@ -25,37 +25,9 @@ import org.apache.spark.sql.catalyst.expressions.{ComplexTypeMergingExpression,
...
-          }
-          GpuColumnVector.from(finalRet, dataType)
-        }
-      }
-    }
-  }
+object GpuExpressionWithSideEffectUtils extends Arm {
 
-  protected def isAllTrue(col: GpuColumnVector): Boolean = {
+  def isAllTrue(col: GpuColumnVector): Boolean = {
```
revans2, line 30: nit: can we document that nulls are considered to be false here?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -57,26 +58,110 @@ case class GpuNot(child: Expression) extends CudfUnaryExpression
...
+  /**
+   * When computing logical expressions on the CPU, the true and false
+   * expressions are evaluated lazily, meaning that the RHS expression
+   * of logical-AND is not evaluated when LHS is False. For logical-OR,
+   * RHS is not evaluated when LHS is True.
+   * This is important in the case where the expressions can have
+   * side-effects, such as throwing exceptions for invalid inputs.
+   *
+   * This method performs lazy evaluation on the GPU by first filtering
+   * the input batch a  where the LHS predicate is True.
```
revans2, line 83: This line feels like it needs to be looked at again. the "a " in between batch and where feels off.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -57,26 +58,110 @@ case class GpuNot(child: Expression) extends CudfUnaryExpression
...
+              rightExpr.columnarEval(leftTrueBatch)
+            }
+            val finalRet = withResourceIfAllowed(rEval) { _ =>
+              rEval match {
+                case f: GpuColumnVector =>
+                  withResource(gather(lhsNoNulls, f)) { combinedVector =>
+                    doColumnar(lhsBool, GpuColumnVector.from(combinedVector, dataType))
+                  }
+                case f: GpuScalar =>
+                  doColumnar(lhsBool, f)
```
revans2, line 110: Like was stated before this code is not covered by any test. I have my doubts that it can be covered, but I think the simplest solution to this is to call columnarEvalToColumn which will return the scalar as a column and not worry about it.

```
withResource(GpuExpressionsUtils.columnarEvalToColumn(rightExpr, leftTrueBatch)) { rEval =>
    withResource(gather(lhsNoNulls, rEval)) { combinedVector =>
        doColumnar(lhsBool, GpuColumnVector.from(combinedVector, dataType))
    }
}
```

It also makes the code much smaller with the expense of extra memory in a case we don't know if it is possible to hit.

revans2 (approved):
This looks good. I have a few nits that would be nice to address, but nothing that I think is required.

amahussein:
> This looks good. I have a few nits that would be nice to address, but nothing that I think is required.

Thanks @revans2 !
I pushed one final commit to address the remaining comments.
Does it look good to you?

revans2 (approved):

Issue #4393, opened 2021-12-20
[BUG] Columnar to Columnar transfers are very slow


revans2:
**Describe the bug**
When exploring #4164 I tested what the performance impact is of staying on the CPU for parquet parsing. I found that for many of the queries in question (like NDS query 9) the CPU was also spending all of it's time parsing parquet so it was not a win at all, especially if we use less cores compared to the CPU.  But while doing that I noticed that we were wasting a lot of time on columnar to columnar data transfers. When profiling I found that we were spending a lot of time in the cudf `HostBuffer.Builder` code checking if we needed to grow the buffers.  This was for buffers where we know up front exactly the maximum amount of memory needed to hold the data.  There were also some issues with the plugin code itself that was not using ideal loops because scala does not make that simple.

I was able to speed up the transfers massively by hacking up the Builder to avoid checking at all. This is not good, but gives us a proof of concept.  We need to.

1. Cache in the builder the maximum number of rows/bytes before we have to grow a buffer instead of re-computing it each time. (this might be enough on the CUDF side, but we need to check)
2. Don't call a JNI function to get the size of the validity buffer we need to allocate. We can do the same computation in pure java and it saves a lot of time.
3. Update all of the for loops in the columnar to columnar transfer code to be while loops because they are much faster.
4. Reduce the number of method calls in the CUDF data path as much as possible. (inline things if we can)
5. This is a bit separate but I noticed that in the columnar to columnar transfer code we compute an approximate batch size at the beginning, but then when we have finished a batch we calculate a better approximation bases off of the data we actually cached. Then we stomp over the top of it for the very next batch. This is not great, although we rarely get more than one batch so it might not matter that much.

We really need some kind of micro benchmarks that we can use to measure how fast this and other operations are that we can run regularly and see if we are regressing. Also once we have these changes made we should do some more profiling to see if there is more that we can do to speed this up.

sperlingxx:
I would like to try this optimization work if no one takes.

revans2:
Sure go ahead

sperlingxx:
https://github.com/NVIDIA/spark-rapids/issues/4565

revans2:
Great work @sperlingxx do you have any benchmark results about how much this helped? If not please don't spend any time creating them. I am just curious.

sperlingxx:
> Great work @sperlingxx do you have any benchmark results about how much this helped? If not please don't spend any time creating them. I am just curious.

[...]

PR #4770, opened 2022-02-14
Improve columnarCopy for HostColumnarToGpu

sperlingxx:

Following the suggestion in #4393, this PR replaced inefficient Scala FOR loops for columnar copy with Java ones, in order to speed up Columnar to Columnar transfers. Besides, it uses `ColumnVector.hasNull` to determine whether the host column data contains any null value, which is more accurate than nullable information from read schema.

According to a non-comprehensive test, the new implementation saves approximately half of time on the columnar copy, comparing to the original implementation.

Here is the test script, which generates 1 billion rows Long data with null:
```python
@allow_non_gpu('FileSourceScanExec')
def test_hc2dc_perf():
    from pyspark.sql.functions import col, sum

    def gen_data(spark, data_path, n_part=10, rows_per_part=100000):
        from pyspark.sql.types import LongType, StructField, StructType

        rdd = spark.sparkContext.parallelize(list(range(n_part)), numSlices=n_part)

        def rand_gen(seed_iter):
            from random import Random
            from pyspark.sql import Row
            rd = Random(next(seed_iter))
            for _ in range(rows_per_part):
                if rd.randint(0, 10) == 0:
                    yield Row(a=None)
                yield Row(a=rd.randint(0, 100000000000000000))

        rows = rdd.mapPartitions(rand_gen)
        df = spark.createDataFrame(rows, StructType([StructField('a', LongType(), True)]))
        df.write.parquet(data_path)

    path = 'PARQUET_DATA_1234'
    with_cpu_session(lambda spark: gen_data(spark, path, n_part=1000, rows_per_part=1000000))
    with_gpu_session(
        lambda spark: spark.read.parquet(path).agg(sum(col('a') % 3)).collect(),
        conf={'spark.rapids.sql.format.parquet.enabled': 'false'})
```

The performance of original implementation (the new metric "buffer time" recording the exact time cost of columnar copy):
```
HostColumnarToGpu

concat batch time total (min, med, max )
5.4 s (18 ms, 42 ms, 208 ms )
stream time total (min, med, max )
58.5 s (278 ms, 609 ms, 1.2 s )
output columnar batches: 91
peak device memory: 8,863,707,456
op time total (min, med, max )
1.7 m (565 ms, 1.1 s, 1.6 s )
output rows: 1,090,917,512
buffer time total (min, med, max )
37.5 s (266 ms, 407 ms, 608 ms )
```

The performance of new implementation:
```
HostColumnarToGpu

concat batch time total (min, med, max )
4.8 s (17 ms, 44 ms, 137 ms )
stream time total (min, med, max )
43.8 s (240 ms, 437 ms, 1.0 s )
output columnar batches: 91
peak device memory: 8,863,707,456
op time total (min, med, max )
1.1 m (339 ms, 711 ms, 1.2 s )
output rows: 1,090,917,512
buffer time total (min, med, max )
21.1 s (102 ms, 227 ms, 320 ms )
```


sperlingxx:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
```diff
@@ -0,0 +1,196 @@
...
+
+  public static void nullCopy(ColumnBuilder b, int rows) {
+    for (int i = 0; i < rows; i++) {
+      b.appendNull();
+    }
+  }
+
+  public static void booleanCopy(ColumnVector cv, ColumnBuilder b, int rows) {
+    if (!cv.hasNull()) {
+      for (int i = 0; i < rows; i++) b.append(cv.getBoolean(i));
```
revans2, line 36: nit: I would prefer to split this line up for formatting purposes.

```java
for (int i = 0; i < rows; i++) {
  b.append(cv.getBoolean(i));
}
```

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
```diff
@@ -0,0 +1,196 @@
...
+      return;
```
revans2, line 37: nit: Is there any performance difference between using an early return like this and an else statement? If not I would prefer the else.

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+    for (int i = 0; i < rows; i++) {
+      if (cv.isNullAt(i)) {
+        b.appendNull();
+        continue;
```
revans2, line 42: nit: Is there any performance difference between using the continue here and putting the `b.append(cv.getBoolean(i));` in an else? if not I would prefer the else.

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
```diff
@@ -0,0 +1,196 @@
...
+        b.append((int) cv.getDecimal(i, precision, scale).toUnscaledLong());
+      }
+      return;
+    }
+    for (int i = 0; i < rows; i++) {
+      if (cv.isNullAt(i)) {
+        b.appendNull();
+        continue;
+      }
+      b.append((int) cv.getDecimal(i, precision, scale).toUnscaledLong());
```
revans2, line 159: Can we file a follow on issue for the decimal types, or possibly put in the improvements now? If we know that the `ColumnVector` is a `WritableColumnVector` we could bypass the need to create a `Decimal` object at all and copy the data/bytes directly.

For decimal32 we could call into `intCopy`. For decimal64 we could call into `longCopy`, and for 128-bit decimal we could get the raw binary bytes.

https://github.com/apache/spark/blob/c8b34ab7340265f1f2bec2afa694c10f174b222c/sql/core/src/main/java/org/apache/spark/sql/execution/vectorized/WritableColumnVector.java#L393-L397

and the open up an API in CUDF JNI to set those raw bytes.  This is what CUDF is doing already

https://github.com/rapidsai/cudf/blob/7025c40c16bfea9b37ac9110f62b7b23a2c51725/java/src/main/java/ai/rapids/cudf/HostColumnVector.java#L1313-L1315

Which is why I said it might be good as a follow on issue.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
```diff
@@ -407,6 +291,7 @@ case class HostColumnarToGpu(child: SparkPlan, goal: CoalesceSizeGoal)
     OP_TIME -> createNanoTimingMetric(MODERATE_LEVEL, DESCRIPTION_OP_TIME),
     STREAM_TIME -> createNanoTimingMetric(MODERATE_LEVEL, DESCRIPTION_STREAM_TIME),
     CONCAT_TIME -> createNanoTimingMetric(MODERATE_LEVEL, DESCRIPTION_CONCAT_TIME),
+    BUFFER_TIME -> createNanoTimingMetric(MODERATE_LEVEL, DESCRIPTION_BUFFER_TIME),
```
revans2, line 294: If we are going to use buffer time here we are going to need to update the docs https://github.com/NVIDIA/spark-rapids/blob/branch-22.04/docs/tuning-guide.md#metrics. I think it fits, but the docs make it specific for buffering input data. There is also a tool that will use bufferTime to graph how long each process took to buffer input data. Because of both of these things I think it might be better to add in a new metric. But even then we would need to update the docs to explain that new metric.

I also would prefer to see this at a DEBUG_LEVEL. I don't think we need this all the time.

revans2 (commented):
Just some nits. Nothing that is a blocker.  Great work.

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
```diff
@@ -0,0 +1,196 @@
...
+ */
+
+package com.nvidia.spark.rapids;
+
+import ai.rapids.cudf.HostColumnVector.ColumnBuilder;
+
+import org.apache.spark.sql.vectorized.ColumnVector;
+
+/**
+ * A helper class which efficiently transfers different types of host columnar data into cuDF.
```
tgravescs, line 24: it would be nice to explain this more and mention specifically not use to scala for loops.

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
gerashegalov: It looks like we could benefit from an efficient bulk API in HostMemoryBuffer

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
gerashegalov: Note that using Scala primitive loops is fine
```
var i = 0
while (i < n) {
   /
   i = i + 1
}
```
Performance degradation comes from Boxing/Unboxing ints in the for comprehension since for comprehensions are syntactic sugar for function application to each element in a range.

See https://stackoverflow.com/questions/21373514/why-are-scala-for-loops-slower-than-logically-identical-while-loops

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
revans2: This `ColumnVector` is the Spark class, not the CUDF class.  In some cases we can know what the layout is, but not all and in many of those cases the underlying data is not exposed, so there is no good way to do a bulk data copy. We could make changes to Spark to try and improve that, but this only really shows up in cases when we are reading an input format like Parquet or ORC.

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
sperlingxx: I dumped JVM bytecode for both approaches. Compared with the else approach,  the return approach saves one `goto` instruction. And the `continue` approach is identical to the `else` approach in terms of bytecode:
The "return" approach:
```java
  public static void shortCopy(org.apache.spark.sql.vectorized.ColumnVector, ai.rapids.cudf.HostColumnVector$ColumnBuilder, int);
    Code:
       0: aload_0
       1: invokevirtual #3                  // Method org/apache/spark/sql/vectorized/ColumnVector.hasNull:()Z
       4: ifne          31
       7: iconst_0
       8: istore_3
       9: iload_3
      10: iload_2
      11: if_icmpge     30
      14: aload_1
      15: aload_0
      16: iload_3
      17: invokevirtual #9                  // Method org/apache/spark/sql/vectorized/ColumnVector.getShort:(I)S
      20: invokevirtual #10                 // Method ai/rapids/cudf/HostColumnVector$ColumnBuilder.append:(S)Lai/rapids/cudf/HostColumnVector$ColumnBuilder;
      23: pop
      24: iinc          3, 1
      27: goto          9
      30: return
      31: iconst_0
      32: istore_3
      33: iload_3
      34: iload_2
      35: if_icmpge     70
      38: aload_0
      39: iload_3
      40: invokevirtual #6                  // Method org/apache/spark/sql/vectorized/ColumnVector.isNullAt:(I)Z
      43: ifeq          54
      46: aload_1
      47: invokevirtual #2                  // Method ai/rapids/cudf/HostColumnVector$ColumnBuilder.appendNull:()Lai/rapids/cudf/HostColumnVector$ColumnBuilder;
      50: pop
      51: goto          64
      54: aload_1
      55: aload_0
      56: iload_3
      57: invokevirtual #9                  // Method org/apache/spark/sql/vectorized/ColumnVector.getShort:(I)S
      60: invokevirtual #10                 // Method ai/rapids/cudf/HostColumnVector$ColumnBuilder.append:(S)Lai/rapids/cudf/HostColumnVector$ColumnBuilder;
      63: pop
      64: iinc          3, 1
      67: goto          33
      70: return
```

The "else" approach:
```java
  public static void shortCopy(org.apache.spark.sql.vectorized.ColumnVector, ai.rapids.cudf.HostColumnVector$ColumnBuilder, int);
    Code:
       0: aload_0
       1: invokevirtual #3                  // Method org/apache/spark/sql/vectorized/ColumnVector.hasNull:()Z
       4: ifne          33
       7: iconst_0
       8: istore_3
       9: iload_3
      10: iload_2
      11: if_icmpge     30
      14: aload_1
      15: aload_0
      16: iload_3
      17: invokevirtual #9                  // Method org/apache/spark/sql/vectorized/ColumnVector.getShort:(I)S
      20: invokevirtual #10                 // Method ai/rapids/cudf/HostColumnVector$ColumnBuilder.append:(S)Lai/rapids/cudf/HostColumnVector$ColumnBuilder;
      23: pop
      24: iinc          3, 1
      27: goto          9
###       30: goto          72                     // the extra goto instruction of the "else" approach
      33: iconst_0
      34: istore_3
      35: iload_3
      36: iload_2
      37: if_icmpge     72
      40: aload_0
      41: iload_3
      42: invokevirtual #6                  // Method org/apache/spark/sql/vectorized/ColumnVector.isNullAt:(I)Z
      45: ifeq          56
      48: aload_1
      49: invokevirtual #2                  // Method ai/rapids/cudf/HostColumnVector$ColumnBuilder.appendNull:()Lai/rapids/cudf/HostColumnVector$ColumnBuilder;
      52: pop
      53: goto          66
      56: aload_1
      57: aload_0
      58: iload_3
      59: invokevirtual #9                  // Method org/apache/spark/sql/vectorized/ColumnVector.getShort:(I)S
      62: invokevirtual #10                 // Method ai/rapids/cudf/HostColumnVector$ColumnBuilder.append:(S)Lai/rapids/cudf/HostColumnVector$ColumnBuilder;
      65: pop
      66: iinc          3, 1
      69: goto          35
      72: return
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/HostColumnarToGpu.scala
sperlingxx: Done.

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
sperlingxx: Done.

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
sperlingxx: Done.

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
sperlingxx: DOne.

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
sperlingxx: I put the explanation into the comment. And I tested with Scala while-loops. Its performance was roughly same as Java loops.

sperlingxx:
build

revans2 (approved):

Issue #4753, opened 2022-02-11
[FEA] Profiling tool should report all spark configs


tgravescs:
**Is your feature request related to a problem? Please describe.**
profiling tool should report all the spark configs (not just the rapids ones) so that those configs can be easily compared between runs and it gives similar functionality to the UI.

PR #4781, opened 2022-02-14
Add Spark properties table to profiling tool output

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/4753

The profiling tool now prints all the Spark properties into a separate table.  It still prints just the rapids related ones into one table so you can easily see those and then all the other Spark properties into a separate table. The spark properties are very useful for comparing applications.

This also has a random change to change a log info message when we see events we don't handle to a log debug message.

 examples:

rapids properties no change:
```
Spark Rapids parameters set explicitly:
+-----------------------------------+----------+----------+
|propertyName                       |appIndex_2|appIndex_1|
+-----------------------------------+----------+----------+
|spark.rapids.memory.pinnedPool.size|2G        |2G        |
|spark.rapids.sql.concurrentGpuTasks|2         |2         |
|spark.rapids.sql.explain           |NOT_ON_GPU|NOT_ON_GPU|
+-----------------------------------+----------+----------+
```

Added spark properties:

```
Spark Properties:
+---------------------------------+------------------------------------------------------------------------------------------+------------------------------------------------------------------------------------------+
|propertyName                     |appIndex_2                                                                                |appIndex_1                                                                                |
+---------------------------------+------------------------------------------------------------------------------------------+------------------------------------------------------------------------------------------+
|spark.app.id                     |local-1622821994212                                                                       |local-1622814619968                                                                       |
|spark.app.name                   |Spark shell                                                                               |Spark shell                                                                               |
....
```


nartal1:
LGTM. Do we need to update in spark-profiling-tool.md that spark properties are also covered along with rapids properties?

tgravescs:
yes, thanks for pointing that out, I added it to list.  I didn't put example as it seems like its pretty self explanatory.

nartal1 (approved):

tgravescs:
build

PR #4785, opened 2022-02-15
Support casting Map to String

### LINKED ISSUE #4696: [FEA] Support casting map to string
Currently, we don't support casting map to string yet. As such, the following command can't work on GPU:
```
spark.sql("map(a, b)").show()
```

In order to show the content of the created map, I have to call either `map_keys` or `map_values` on the created map then show. It's inconvenient.

We should better support casting map to string. Assuming that we have supported casting array/struct to string, casting a map to string is just as simple as concatenating `cast_to_string(map_keys) + " -> " + cast_to_string(map_values)` pair by pair.

HaoYang670:
close #4696

In this PR, we:
1. refactor the function `cast_array_to_string` by abstracting the logic of concatenating String Array to a new function `concatenateStringArrayElements` which is also used by `cast_map_to_string`
2. create the function `cast_map_to_string` which casts MapType column to StringType column. Also add some tests.
3. deleted some outdated tests.

### DIFF
--- a/integration_tests/src/main/python/cast_test.py
+++ b/integration_tests/src/main/python/cast_test.py
@@ -56,15 +56,6 @@ def test_cast_nested(data_gen, to_type):
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark : unary_op_df(spark, data_gen).select(f.col('a').cast(to_type)))
 
-@allow_non_gpu('ProjectExec', 'Cast', 'Alias')
-@pytest.mark.parametrize('data_gen,to_type', [
-    # maps are not supported for casting to a String, but structs are, so we need to verify this
-    (StructGen([('structF1', StructGen([('structF11', MapGen(ByteGen(nullable=False), byte_gen))]))]), StringType())])
-def test_cast_nested_fallback(data_gen, to_type):
-    assert_gpu_fallback_collect(
-            lambda spark : unary_op_df(spark, data_gen).select(f.col('a').cast(to_type)),
-            'Cast')
-
 def test_cast_string_date_valid_format():
     # In Spark 3.2.0+ the valid format changed, and we cannot support all of the format.
     # This provides values that are valid in all of those formats.
--- a/integration_tests/src/main/python/cast_test.py
+++ b/integration_tests/src/main/python/cast_test.py
@@ -236,6 +236,17 @@ def test_cast_array_with_unsupported_element_to_string_fallback(data_gen, legacy
     )
 
 
+basic_map_gens_for_cast_to_string = [MapGen(f(nullable=False), f()) for f in [BooleanGen, ByteGen, ShortGen, IntegerGen, LongGen, DateGen, TimestampGen]] + [simple_string_to_string_map_gen]
+
+@pytest.mark.parametrize('data_gen', basic_map_gens_for_cast_to_string, ids=idfn)
+@pytest.mark.parametrize('legacy', ['true', 'false'])
+def test_cast_map_to_string(data_gen, legacy):
+    _assert_cast_to_string_equal(
+        data_gen, 
+        {"spark.rapids.sql.castDecimalToString.enabled"    : 'true', 
+        "spark.sql.legacy.castComplexTypesToString.enabled": legacy})
+
+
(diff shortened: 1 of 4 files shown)


HaoYang670:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -732,34 +750,88 @@ object GpuCast extends Arm {
...
+    val (leftStr, rightStr, nullStr) =
+      if (legacyCastToString) ("[", "]", "") else ("{", "}", "null")
+
+    val (strKey, strValue) = withResource(input.getChildColumnView(0)) { kvStructColumn =>
+      val strKey = withResource(kvStructColumn.getChildColumnView(0)) { keyColumn =>
+        doCast(
+          keyColumn, from.keyType, StringType, ansiMode,
+          legacyCastToString, stringToDateAnsiModeEnabled)
+      }
+      val strValue = withResource(kvStructColumn.getChildColumnView(1)) { valueColumn =>
```
revans2, line 789: nit: Technically if an exception happens when computing `strValue` `strKey` will be leaked. Can we have a `closeOnExcept` for strKey while computing `strValue`?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -732,34 +750,88 @@ object GpuCast extends Arm {
...
+          valueColumn, from.valueType, StringType, ansiMode,
+          legacyCastToString, stringToDateAnsiModeEnabled)
+      }
+      (strKey, strValue)
+    }
+
+    withResource(
+      Seq(leftStr, rightStr, arrowStr, emptyStr, nullStr, spaceStr).safeMap(Scalar.fromString)
+    ) { case Seq(leftScalar, rightScalar, arrowScalar, emptyScalar, nullScalar, spaceScalar) =>
+      val strElements = withResource(Seq(strKey, strValue)) { case Seq(strKey, strValue) =>
```
revans2, line 800: nit: again we could leak strKey and strValue if we get an exception while trying to create the scalar values above.  This is a little harder to fix because strKey and strValue do not outlive the scalars. This one is a much smaller nit because it is much less likely to happen.

integration_tests/src/main/python/cast_test.py
```diff
@@ -167,21 +158,27 @@ def test_cast_long_to_decimal_overflow():
         conf={'spark.sql.legacy.allowNegativeScaleOfDecimal': True})
 
 # casting these types to string should be passed
-basic_gens_for_cast_to_string = [byte_gen, short_gen, int_gen, long_gen, string_gen, boolean_gen, date_gen, null_gen, timestamp_gen] + decimal_gens_no_neg
+basic_gens_for_cast_to_string = [ByteGen, ShortGen, IntegerGen, LongGen, StringGen, BooleanGen, DateGen, TimestampGen]
```
revans2, line 161: Why exactly what this change made? It feels a little confusing to have the classes in here, when the other similarly named variables have the instances of the gen in them. It also is not clear what value this is providing over having just the gens in here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
HaoYang670: Done

integration_tests/src/main/python/cast_test.py
HaoYang670: Because it is not easy to build the `data_gens` for `Map` from the provided instances of gens as the `_key_gen`   should be created with the declaration `nullable = False`.  
```python
basic_array_struct_gens_for_cast_to_string = [f() for f in basic_gens_for_cast_to_string] + [null_gen] + decimal_gens_no_neg
basic_map_gens_for_cast_to_string = [
    MapGen(f(nullable=False), f()) for f in basic_gens_for_cast_to_string] + [
    MapGen(DecimalGen(nullable=False), DecimalGen(precision=7, scale=3)), MapGen(DecimalGen(precision=7, scale=7, nullable=False), DecimalGen(precision=12, scale=2))]
```

revans2 (approved):

Issue #4761, opened 2022-02-11
[BUG] test_simple_partitioned_read failed on Spark 3.3


jlowe:
From a recent nightly test run:
```
[2022-02-11T18:42:38.233Z] _ test_simple_partitioned_read[{'spark.rapids.sql.format.orc.reader.type': 'MULTITHREADED'}-] _
[2022-02-11T18:42:38.233Z] 
[2022-02-11T18:42:38.233Z] spark_tmp_path = '/tmp/pyspark_tests//310736/', v1_enabled_list = ''
[2022-02-11T18:42:38.233Z] reader_confs = {'spark.rapids.sql.format.orc.reader.type': 'MULTITHREADED'}
[2022-02-11T18:42:38.233Z] 
[2022-02-11T18:42:38.233Z]     @pytest.mark.parametrize('v1_enabled_list', ["", "orc"])
[2022-02-11T18:42:38.233Z]     @pytest.mark.parametrize('reader_confs', reader_opt_confs, ids=idfn)
[2022-02-11T18:42:38.233Z]     def test_simple_partitioned_read(spark_tmp_path, v1_enabled_list, reader_confs):
[2022-02-11T18:42:38.233Z]         # Once https://github.com/NVIDIA/spark-rapids/issues/131 is fixed
[2022-02-11T18:42:38.233Z]         # we should go with a more standard set of generators
[2022-02-11T18:42:38.233Z]         orc_gens = [byte_gen, short_gen, int_gen, long_gen, float_gen, double_gen,
[2022-02-11T18:42:38.233Z]         string_gen, boolean_gen, DateGen(start=date(1590, 1, 1)),
[2022-02-11T18:42:38.233Z]         TimestampGen(start=datetime(1590, 1, 1, tzinfo=timezone.utc))]
[2022-02-11T18:42:38.233Z]         gen_list = [('_c' + str(i), gen) for i, gen in enumerate(orc_gens)]
[2022-02-11T18:42:38.233Z]         first_data_path = spark_tmp_path + '/ORC_DATA/key=0/key2=20'
[2022-02-11T18:42:38.233Z]         with_cpu_session(
[2022-02-11T18:42:38.233Z]                 lambda spark : gen_df(spark, gen_list).write.orc(first_data_path))
[2022-02-11T18:42:38.233Z]         second_data_path = spark_tmp_path + '/ORC_DATA/key=1/key2=21'
[2022-02-11T18:42:38.233Z]         with_cpu_session(
[...]

PR #4786, opened 2022-02-15
Add shims to compare ORC TypeDescription

wbo4958:
To fix https://github.com/NVIDIA/spark-rapids/issues/4761

Orc has defaulted to check Attribute when comparing two TypeDescription
from Spark 3.2.x, which causes some issues, since the catalyst schema
does not include this kind of information.


wbo4958:
build

revans2 (approved):

PR #4794, opened 2022-02-16
Decimal-128 support for mod and pmod

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/1572. 
cudf has added decimal128 support for pmod and remainder operators. This PR is to add decimal-128 support for the same.
Updated the integration tests.


nartal1:
build

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -236,7 +236,7 @@ def test_int_division_mixed(lhs, rhs):
                 'a DIV b'),
             conf=allow_negative_scale_of_decimal_conf)
 
-@pytest.mark.parametrize('data_gen', numeric_gens, ids=idfn)
+@pytest.mark.parametrize('data_gen', numeric_gens + decimal_128_gens, ids=idfn)
```
jlowe, line 239: I think we should add `decimal_gen_scale_precision` and `decimal_gen_64bit` so the DECIMAL32 and DECIMAL64 precision ranges are also tested.  Applies to the pmod test below as well.

nartal1: Done

jlowe (approved):

nartal1:
build

PR #4796, opened 2022-02-16
Implement ExistenceJoin Iterator using an auxiliary left semijoin 

### LINKED ISSUE #589: [FEA] Support ExistenceJoin

gerashegalov:
This PR implements an iterator for ExistenceJoin 

1. This PR computes ExistenceJoin by executing  left semijoin via cuDF. The lhs GatherMap is scattering `true` into a Boolean column with all lhs.numRows  being initially`false` . The rhs data is not gathered. 

1. The PR also fixes regex matching against SparkPlan node strings. The previously used simple String mentions ExistenceJoin only in the CPU plan but does not print ExistenceJoin type as part of the Join exec string in the GPU plan. 

Closes #589
 

gerashegalov:
build

gerashegalov:
build

build/buildall
```diff
@@ -50,7 +57,7 @@ function bloopInstall() {
       mkdir -p "$bloop_config_dir"
       rm -f "$bloop_config_dir"/*
 
-      mvn install ch.epfl.scala:maven-bloop_2.13:1.4.9:bloopInstall -pl dist -am \
+      mvn install ch.epfl.scala:maven-bloop_${BLOOP_SCALA_VERSION}:${BLOOP_VERSION}:bloopInstall -pl dist -am \
```
jlowe, line 60: These changes have nothing to do with ExistenceJoin and really should be a separate PR.  Otherwise if we need to revert one we end up reverting them both, and the PR summary "hides" the unrelated changes bundled with it.

integration_tests/src/main/python/join_test.py
```diff
@@ -802,27 +802,23 @@ def do_join(spark):
...
-            "   OR l._1 in (select * from {} as r)"
+            "   or exists (select * from {} as r where r._2 = l._2 AND r._3 = l._3)"
         ).format(leftTable, rightTable))
+        res.explain(True)
         return res
-    assert_cpu_and_gpu_are_equal_collect_with_capture(do_join, r".+Join ExistenceJoin\(exists#[0-9]+\).+")
-
+    assert_cpu_and_gpu_are_equal_collect_with_capture(do_join, r"ExistenceJoin\(exists#[0-9]+\)",
+        conf={
+            "spark.rapids.sql.explain"                  : "ALL",
```
jlowe, line 822: Debugging code inadvertently left here?

integration_tests/src/main/python/join_test.py
```diff
@@ -802,27 +802,23 @@ def do_join(spark):
...
+            "spark.sql.adaptive.enabled"                : False
```
jlowe, line 823: Do we need this because it fails to capture properly without it?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -506,7 +514,6 @@ class ConditionalHashJoinIterator(
     }
   }
 }
-
```
jlowe, line 509: Nit: This whitespace should be restored.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -117,6 +120,8 @@ object GpuHashJoin extends Arm {
         if (keyDataTypes.exists(_.isInstanceOf[StructType])) {
           meta.willNotWorkOnGpu(s"$joinType joins currently do not support with struct keys")
         }
+      case ExistenceJoin(_) =>
+        // TODO enabling existence join
```
jlowe, line 124: This TODO needs to be addressed.  If the join has a condition then it needs to be applied when on the GPU, and if it cannot be processed by the GPU then we need to fallback to the CPU.  Tests need to be added for these cases.  If this is not going to be addressed in this PR then this needs to do the safe thing, which is fallback to the CPU.  As it is now, it will allow some existence joins that cannot be processed on the GPU to be on the GPU which will either result in a crash or silent data corruption.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -259,7 +264,7 @@ abstract class BaseHashJoinIterator(
     // TODO: Replace this estimate with exact join row counts using the corresponding cudf APIs
     //       being added in https://github.com/rapidsai/cudf/issues/9053.
     joinType match {
-      case _: InnerLike | LeftOuter | RightOuter =>
+      case _: InnerLike | LeftOuter | RightOuter | ExistenceJoin(_) =>
```
jlowe, line 267: ExistenceJoin cannot explode like Inner/LeftOuter/RightOuter.  This should be treated more like a semi join, as the number of output rows should be the same as the number of left table input rows.

integration_tests/src/main/python/join_test.py
```diff
@@ -802,27 +802,23 @@ def do_join(spark):
...
-def test_existence_join(allowFallback, spark_tmp_table_factory):
+def test_existence_join(spark_tmp_table_factory):
     leftTable = spark_tmp_table_factory.get()
     rightTable = spark_tmp_table_factory.get()
     def do_join(spark):
         # create non-overlapping ranges to have a mix of exists=true and exists=false
-        spark.createDataFrame([v] for v in range(2, 10)).createOrReplaceTempView(leftTable)
-        spark.createDataFrame([v] for v in range(0, 8)).createOrReplaceTempView(rightTable)
+        spark.createDataFrame([f"left_{v}", v * 10, v * 100] for v in range(2, 10)).createOrReplaceTempView(leftTable)
+        spark.createDataFrame([f"right_{v}", v * 10, v * 100] for v in range(0, 8)).createOrReplaceTempView(rightTable)
```
jlowe, line 811: There should be duplicate values in the right table, as I believe the tests will start failing for the current approach once they are added.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -411,7 +416,10 @@ class HashJoinIterator(
       rightData: LazySpillableColumnarBatch): Option[JoinGatherer] = {
     withResource(new NvtxWithMetrics("hash join gather map", NvtxColor.ORANGE, joinTime)) { _ =>
       val maps = joinType match {
-        case LeftOuter => leftKeys.leftJoinGatherMaps(rightKeys, compareNullsEqual)
+        case LeftOuter | ExistenceJoin(_) =>
+          // existence join is implemented as a left outer join with GatherMap converted
+          // to the Boolean column "exists"
+          leftKeys.leftJoinGatherMaps(rightKeys, compareNullsEqual)
```
jlowe, line 422: An existence join is not like a left outer join.  A left outer join will replicate rows when there are multiple matches in the right table.

jlowe (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
```diff
@@ -532,12 +551,21 @@ class JoinGathererImpl(
...
+        val batch = data.getBatch
+        val gatheredTable = withResource(GpuColumnVector.from(batch)) { table =>
+          table.gather(gatherView, boundsCheckPolicy)
+        }
+        withResource(gatheredTable) { gt =>
+          GpuColumnVector.from(gt, GpuColumnVector.extractTypes(batch))
+        }
+      }.getOrElse {
+        withResource(gatherView.binaryOp(BinaryOp.GREATER,
+          Scalar.fromInt(Int.MinValue), DType.BOOL8)) { existenceColumn =>
```
revans2, line 564: This scalar is going to be leaked and needs to be closed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
```diff
@@ -532,12 +551,21 @@ class JoinGathererImpl(
...
+      dataOpt.map { data =>
+        val batch = data.getBatch
+        val gatheredTable = withResource(GpuColumnVector.from(batch)) { table =>
+          table.gather(gatherView, boundsCheckPolicy)
+        }
+        withResource(gatheredTable) { gt =>
+          GpuColumnVector.from(gt, GpuColumnVector.extractTypes(batch))
+        }
+      }.getOrElse {
+        withResource(gatherView.binaryOp(BinaryOp.GREATER,
```
revans2, line 563: Can we add some documentation about what happens if the JoinGatherer does not have a dataColumn, and what that means. It is not an obvious API change.

integration_tests/src/main/python/join_test.py
```diff
@@ -802,27 +802,46 @@ def do_join(spark):
...
+])
+def test_existence_join(numComplementsToExists, aqeEnabled, spark_tmp_table_factory):
     leftTable = spark_tmp_table_factory.get()
     rightTable = spark_tmp_table_factory.get()
     def do_join(spark):
         # create non-overlapping ranges to have a mix of exists=true and exists=false
-        spark.createDataFrame([v] for v in range(2, 10)).createOrReplaceTempView(leftTable)
-        spark.createDataFrame([v] for v in range(0, 8)).createOrReplaceTempView(rightTable)
+        # duplicate every row in the rhs to verify it does not affect
+        # the the number of the output rows, which should be equal to the number of the
```
jlowe, line 816:
```suggestion
        # the number of the output rows, which should be equal to the number of the
```

integration_tests/src/main/python/join_test.py
```diff
@@ -802,27 +802,46 @@ def do_join(spark):
...
-            "   OR l._1 in (select * from {} as r)"
+            f"where l._2 >= {10 * (lhs_upper_bound - numComplementsToExists)}"
+            "   or exists (select * from {} as r where r._2 = l._2 AND r._3 = l._3)"
         ).format(leftTable, rightTable))
         return res
-    assert_cpu_and_gpu_are_equal_collect_with_capture(do_join, r".+Join ExistenceJoin\(exists#[0-9]+\).+")
 
+    assert_cpu_and_gpu_are_equal_collect_with_capture(do_join, r"ExistenceJoin\(exists#[0-9]+\)",
+        conf={
+            "spark.sql.adaptive.enabled": aqeEnabled,
```
jlowe, line 846: Is it important to vary AQE here (i.e.: are we getting improved existence join test coverage by this)?  Not sure why AQE would apply to this type of join test but not the tests of other join types.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -298,14 +299,20 @@ abstract class SplittableJoinIterator(
         None
       }
 
-      val lazyLeftMap = LazySpillableGatherMap(leftMap, spillCallback, "left_map")
+      // TODO refactor lazyLeftMap out of the way, lazy workaround
+      lazy val lazyLeftMap = LazySpillableGatherMap(leftMap, spillCallback, "left_map")
       val gatherer = rightMap match {
         case None =>
           // When there isn't a `rightMap` we are in either LeftSemi or LeftAnti joins.
```
jlowe, line 306: Would be good to update this comment block to explain what's going on with existence joins through here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -344,4 +351,38 @@ abstract class SplittableJoinIterator(
...
+
+  private def createExistenceJoinGatherer(
+    lhs: LazySpillableColumnarBatch,
+    existenceGatherMap: LazySpillableGatherMap
+  ): JoinGatherer = {
+    // cuDF executes left semijoin, the gatherer is constructed with a new
+    // gather to gather every row from lhs
+    //
+    // we build a new rhs with a the "exists" Boolean column that has as many rows
+    // as the input from from a false-Scalar, then scatter true-Scalar using the original
```
jlowe, line 363:
```suggestion
    // we build a new rhs with an "exists" Boolean column that has as many rows
    // as the input from a false-Scalar, then scatter true-Scalar using the original
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -344,4 +351,38 @@ abstract class SplittableJoinIterator(
...
+    // semijoin lhs-GatherMap labeling rows that have at least one match in the original
+    // rhs
+    //
+    val rhsExistsCB = withResource(Scalar.fromBool(false)) { falseScalar =>
+      withResource(Scalar.fromBool(true)) { trueScalar =>
+        withResource(ColumnVector.fromScalar(falseScalar, lhs.numRows)) { falseCV =>
```
jlowe, line 369: Nit: `falseScalar` should be freed after this, as it's no longer needed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -344,4 +351,38 @@ abstract class SplittableJoinIterator(
...
+    // cuDF executes left semijoin, the gatherer is constructed with a new
+    // gather to gather every row from lhs
+    //
+    // we build a new rhs with a the "exists" Boolean column that has as many rows
+    // as the input from from a false-Scalar, then scatter true-Scalar using the original
+    // semijoin lhs-GatherMap labeling rows that have at least one match in the original
+    // rhs
+    //
+    val rhsExistsCB = withResource(Scalar.fromBool(false)) { falseScalar =>
+      withResource(Scalar.fromBool(true)) { trueScalar =>
```
jlowe, line 368: Nit: Related to the `falseScalar` lifespan comment, this should be created closer to the scope in which it is needed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -344,4 +351,38 @@ abstract class SplittableJoinIterator(
...
+        withResource(ColumnVector.fromScalar(falseScalar, lhs.numRows)) { falseCV =>
+          withResource(new Table(falseCV)) { existsNothingTable =>
+            // auto close original gather map
+            withResource(existenceGatherMap) { egMap =>
```
jlowe, line 372: This should be the first `withResource` in the entire chain, as any exception that occurs in this method should close it.

integration_tests/src/main/python/join_test.py
```diff
@@ -802,27 +802,46 @@ def do_join(spark):
...
-            marks=pytest.mark.allow_non_gpu('SortMergeJoinExec')),
-        pytest.param('false',
-            marks=pytest.mark.xfail(reason="https://github.com/NVIDIA/spark-rapids/issues/589"))
-    ], ids=idfn
-)
-def test_existence_join(allowFallback, spark_tmp_table_factory):
+@pytest.mark.parametrize('numComplementsToExists', [0, 1, 2])
+@pytest.mark.parametrize('aqeEnabled', [
+    pytest.param(False),
+    pytest.param(True, marks=pytest.mark.allow_non_gpu('ShuffleExchangeExec'))
```
revans2, line 808: Why is the shuffle not on the GPU? and do we need a follow on issue to put it on the GPU?

integration_tests/src/main/python/join_test.py
```diff
@@ -802,27 +802,46 @@ def do_join(spark):
...
+        rhs_data.append(('right_null', None, None))
+        rhs_data_with_dupes=[]
+        for dupe in rhs_data:
+            rhs_data_with_dupes.extend([dupe, dupe])
+
+        df_right = spark.createDataFrame(rhs_data_with_dupes)
+        df_right.createOrReplaceTempView(rightTable)
+
+        assert df_right.count() == df_left.count() * 2
+        assert df_right.distinct().count() == df_left.count()
```
revans2, line 834: Do we need these asserts still? It shows that the input data matches what is expected, but after we have done it once do we need to do it every time?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -298,14 +299,20 @@ abstract class SplittableJoinIterator(
         None
       }
 
-      val lazyLeftMap = LazySpillableGatherMap(leftMap, spillCallback, "left_map")
+      // TODO refactor lazyLeftMap out of the way, lazy workaround
```
revans2, line 302: Please file a follow on issue to do this, or just remove the TODO

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -344,4 +351,38 @@ abstract class SplittableJoinIterator(
...
+                withResource(
+                  Table.scatter(Array(trueScalar), existsView, existsNothingTable, false)
+                ) { existsTable =>
+                  withResource(
+                    GpuColumnVector.from(existsTable, Array[DataType](BooleanType))
+                  ) { existsBatch =>
+                    LazySpillableColumnarBatch.apply(existsBatch, spillCallback, "right_data")
+                  }}}}}}}}
+
+    val lazyRightMap: LazySpillableGatherMap = LazySpillableGatherMap.identity(lhs.numRows)
```
jlowe, line 384: This is pretty wasteful.  As it is now, we'll create the following unnecessary outputs:
- a gather map with an INT32 for every row in the left table
- a manifest of that gather, which is a copy of every column in the left table

The existence join should be space and computationally efficient.  For the left table columns, we should simply increment the refcounts on the left table columns, i.e.: zero-copy.  Then we can build a new columnar batch with the added existence column computed via the scatter approach above.  Trying to use JoinGatherer does not seem like the best fit, as we don't actually want to gather for this type of join.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -344,4 +351,38 @@ abstract class SplittableJoinIterator(
...
   }
+
+  private def createExistenceJoinGatherer(
+    lhs: LazySpillableColumnarBatch,
+    existenceGatherMap: LazySpillableGatherMap
+  ): JoinGatherer = {
+    // cuDF executes left semijoin, the gatherer is constructed with a new
+    // gather to gather every row from lhs
+    //
+    // we build a new rhs with a the "exists" Boolean column that has as many rows
```
revans2, line 362: This feels off.  `with a the "exists"` I think the `a` is a typo

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
revans2, line 363: Double from

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -344,4 +351,38 @@ abstract class SplittableJoinIterator(
...
+    // as the input from from a false-Scalar, then scatter true-Scalar using the original
+    // semijoin lhs-GatherMap labeling rows that have at least one match in the original
+    // rhs
+    //
+    val rhsExistsCB = withResource(Scalar.fromBool(false)) { falseScalar =>
```
revans2, line 367: This is way too deeply nested for me. Could we try to break it up some? The `falseScalar` is only used to create `falseCV`. It might also be nice to create a method for Table.scatter that takes the columnView and a single Scalar as input, and does all of the wrapping/unwrapping. to make this code that much more readable.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
```diff
@@ -344,4 +351,38 @@ abstract class SplittableJoinIterator(
...
+            withResource(existenceGatherMap) { egMap =>
+              val numExists = egMap.getRowCount.toInt
+              withResource(egMap.toColumnView(0, numExists)) { existsView =>
+                withResource(
+                  Table.scatter(Array(trueScalar), existsView, existsNothingTable, false)
+                ) { existsTable =>
+                  withResource(
+                    GpuColumnVector.from(existsTable, Array[DataType](BooleanType))
+                  ) { existsBatch =>
+                    LazySpillableColumnarBatch.apply(existsBatch, spillCallback, "right_data")
```
revans2, line 381: Why call apply directly instead of just calling it as a function?

integration_tests/src/main/python/join_test.py
revans2: That is fine. Might be nice to have a comment explaining it.

integration_tests/src/main/python/join_test.py
gerashegalov: added this because I see varying behavior with RDDScanExec. Once this is fixed it will produce a shuffle hash join plan that I don't see without AQE:
```
Exec <SortExec> could run on GPU but is going to be removed because replacing sortMergeJoin with shuffleHashJoin
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AbstractGpuJoinIterator.scala
gerashegalov: yes, maybe not to the last detail but I was concious that this is an inneficient solution, only efficient in the code amount needed to implement it

scalastyle-config.xml
```diff
@@ -94,7 +94,7 @@ You can also disable only one rule, by specifying its rule id, as specified in:
 
     <!-- ??? usually shouldn't be checked into the code base. -->
     <check level="error" class="org.scalastyle.scalariform.NotImplementedErrorUsage"
-           enabled="true"/>
+           enabled="false"/>
```
revans2, line 97: Why is this change being made? it feels odd to remove this check?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -671,6 +673,33 @@ trait GpuHashJoin extends GpuExec {
       new ConditionalHashJoinIterator(spillableBuiltBatch, boundBuildKeys, lazyStream,
         boundStreamKeys, streamedPlan.output, compiledCondition,
         realTarget, joinType, buildSide, compareNullsEqual, spillCallback, opTime, joinTime)
+    } else if (joinType.isInstanceOf[ExistenceJoin]) {
+      new Iterator[ColumnarBatch]() {
```
jlowe, line 677: Nit: I think this is enough code for this iterator to have a class for it, or at least a private method that creates this iterator, to make this `val joinIterator = if (...` block more compact and readable.

gerashegalov: agreed. This is just a WIP prototype code with everything leaking but passing the test, which I wanted to show you to get a confirmation whether this is close to the solution you suggest.

gerashegalov:
conditional join is not ready yet, hitting
```
Caused by: ai.rapids.cudf.CudfException: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
```

gerashegalov:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -622,7 +628,114 @@ trait GpuHashJoin extends GpuExec {
...
+        Table.mixedLeftSemiJoinGatherMap(
+          leftfKeysTab,
+          rightKeysTab,
+          leftTab,
+          rightTab,
+          compiledCondition,
+          if (compareNullsEqual) NullEquality.EQUAL else NullEquality.UNEQUAL))
+    }
+
+    private def existsScatterMap(leftColumnarBatch: ColumnarBatch): GatherMap = {
```
jlowe, line 701: Would be good to comment this function or have a comment somewhere that documents how the existence join column is computed.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -622,7 +628,114 @@ trait GpuHashJoin extends GpuExec {
...
+        })
+    }
+
+    private def falseColumnTable(numLeftRows: Int): Table = {
+      withResource(ColumnVector.fromScalar(falseScalar, numLeftRows))(
+        new Table(_)
+      )
+    }
+
+    private def existsTable(leftColumnarBatch: ColumnarBatch): Table = {
```
jlowe, line 716: I suggest having this return a `ColumnVector`.  It's more confusing to return a `Table` when the `Table` only has one column, and the callers will have no use for the `Table`.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -669,6 +782,27 @@ trait GpuHashJoin extends GpuExec {
...
+      numOutputRows: GpuMetric,
+      joinOutputRows: GpuMetric,
+      numOutputBatches: GpuMetric,
+      opTime: GpuMetric,
+      joinTime: GpuMetric): Iterator[ColumnarBatch] = {
+
+    // The HashJoinIterator takes ownership of the built keys and built data. It will close
+    // them when it is done
+    val joinIterator = if (joinType.isInstanceOf[ExistenceJoin]) {
+      existenceJoinIterator(builtBatch, stream)
```
jlowe, line 801: The existence join iterator should update the appropriate metrics, otherwise they will fail to be reported properly.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -622,7 +628,114 @@ trait GpuHashJoin extends GpuExec {
...
+        use(GpuColumnVector.from(builtBatch)),
+        use(gpuExpr.convertToAst(numFirstConditionTableColumns).compile())
+      )
+    )
+
+    val rightKeysTab = use(
+      withResource(GpuProjectExec.project(builtBatch, boundBuildKeys))(GpuColumnVector.from(_)))
+
+    val falseScalar = use(Scalar.fromBool(false))
+    val trueScalar = use(Scalar.fromBool(true))
```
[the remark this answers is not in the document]
gerashegalov: they are closed when the iterator is exhausted. Why keep recreating them?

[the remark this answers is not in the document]
jlowe: Because they can create fragmentation problems by holding onto them for a long time.  The counter argument is: they are relatively cheap to create, so why cache them?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -507,6 +510,141 @@ class ConditionalHashJoinIterator(
...
+   * into the left table batch pointing to rows that have a join partner on the
+   * right-hand side.
+   *
+   * Given Boolean column FC totaling as many rows as leftColumnarBatch, all having
+   * the value "false", scattering "true" into column FC will produce the "exists"
+   * column of ExistenceJoin
+   */
+  private def existsScatterMap(leftColumnarBatch: ColumnarBatch): GatherMap = {
+    withResource(
+      new NvtxWithMetrics("existence join gather map", NvtxColor.ORANGE, joinTime)
```
jlowe, line 599:
```suggestion
      new NvtxWithMetrics("existence join scatter map", NvtxColor.ORANGE, joinTime)
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -507,6 +510,141 @@ class ConditionalHashJoinIterator(
...
+    val streamHasNext = lazyStream.hasNext
+    if (!streamHasNext) {
+      close()
+    }
+    streamHasNext
+  }
+
+  override def next(): ColumnarBatch = {
+    withResource(lazyStream.next()) { lazyBatch =>
+      existenceJoinNextBatch(lazyBatch.getBatch)
```
jlowe, line 541: `opTime` needs to include the time spent performing the join.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -507,6 +510,141 @@ class ConditionalHashJoinIterator(
...
+  numFirstConditionTableColumns: Int,
+  compareNullsEqual: Boolean,
+  opTime: GpuMetric,
+  joinTime: GpuMetric
+) extends Iterator[ColumnarBatch]()
+    with TaskAutoCloseableResource with Arm {
+
+  use(spillableBuiltBatch)
+  val compiledConditionRes: Option[CompiledExpression] = boundCondition
+    .map(gpuExpr => use(gpuExpr.convertToAst(numFirstConditionTableColumns).compile()))
```
jlowe, line 529: `opTime` needs to include time spent compiling the AST

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
jlowe, line 541: It would be nice to have an NVTX range around the existence join processing, as there may be quite a bit of GPU computation related to evaluating key columns, etc.  Can use an `NvtxWithMetrics` to handle this along with the `opTime` update.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -507,6 +510,141 @@ class ConditionalHashJoinIterator(
...
+    if (!streamHasNext) {
+      close()
+    }
+    streamHasNext
+  }
+
+  override def next(): ColumnarBatch = {
+    withResource(lazyStream.next()) { lazyBatch =>
+      existenceJoinNextBatch(lazyBatch.getBatch)
+    }
```
jlowe, line 542: Just before returning a new output batch, we need to call `spillableBuiltBatch.allowSpilling()`, since we're holding onto that GPU resource while downstream nodes continue to process on the GPU.  `LazySpillableColumnarBatch` isn't actually spillable until that is called (and we don't want to call it before we actually need to support spilling).

integration_tests/src/main/python/join_test.py
```diff
@@ -779,27 +779,65 @@ def do_join(spark):
...
+        # duplicate with a match
+        lhs_data.append(('left_2', 20, 200))
+        lhs_data.append(('left_null', None, None))
+        df_left = spark.createDataFrame(lhs_data)
+        df_left.createOrReplaceTempView(leftTable)
+
+        rhs_data = list((f"right_{v}", v * 10, v * 100) for v in range(0, 8))
+        rhs_data.append(('right_null', None, None))
+        # duplicate every row in the rhs to verify it does not affect
+        # the number of output rows, which should be equal to the number of the
```
jlowe, line 811:
```suggestion
        # the number of output rows, which should be equal to the left table row count
```

gerashegalov:
build

sameerz:
build

jlowe (approved):

PR #4805, opened 2022-02-16
Avoid output cast by using unsigned type output for GpuExtractChunk32 [databricks]

### LINKED ISSUE #4777: Remove input upcast in GpuExtractChunk32
After rapidsai/cudf#10250 we should be able to remove the [input upcast in GpuExtractChunk32](https://github.com/NVIDIA/spark-rapids/blob/branch-22.04/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/AggregateFunctions.scala#L599-L604).

As part of the change, we will need to create new unsigned DataType classes to handle `UINT32` columns for the incoming data of the three least significant chunks and `UINT64` columns for the resulting aggregations on those columns.

jlowe:
Fixes #4777.

Now that rapidsai/cudf#10246 is fixed, we can avoid upcasting the outputs of `GpuExtractChunk32` by directly using the unsigned types generated during the chunk extraction.  This also adds GPU unsigned types for integer and long for use in Catalyst expressions which are needed for this change.


jlowe:
build

revans2 (approved):

abellina (approved):

Issue #4802, opened 2022-02-16
[BUG] GPU CSV read does not honor ignoreCorruptFiles or ignoreMissingFiles


jlowe:
Relates to #4727.  If the GPU performs a CSV read operation, it should honor the `spark.sql.files.ignoreCorruptFiles` and `spark.sql.files.ignoreMissingFiles` settings.

Issue #4803, opened 2022-02-16
[BUG] GPU JSON read does not honor ignoreCorruptFiles or ignoreMissingFiles


jlowe:
Relates to #4727.  If the GPU performs a JSON read operation, it should honor the spark.sql.files.ignoreCorruptFiles and spark.sql.files.ignoreMissingFiles settings.

Issue #4795, opened 2022-02-16
[BUG] Read ORC does not ignoreCorruptFiles


wbo4958:
This is a follow-up of https://github.com/NVIDIA/spark-rapids/issues/4727. ORC mutlithreaded and coalescing reading will throw exception even ignoreCorruptFiles is set when reading the corrupt file.

PR #4809, opened 2022-02-17
Add ignoreCorruptFiles for ORC readers [databricks]

wbo4958:
This PR is to close https://github.com/NVIDIA/spark-rapids/issues/4802, https://github.com/NVIDIA/spark-rapids/issues/4803 and https://github.com/NVIDIA/spark-rapids/issues/4795.

Since ORC/Parquet PERFILE and texted-based GPU readings (CSV, JSON) are already in the scope of FileScanRDD which has done the feature ignoring missing and ignoring corrupt files. we don't need to do extra things for this.

I can't find a way to trigger the IOException or RuntimeException for CPU CSV/JSON parsing, since CPU can parse wrong files when CSV/JSON reading. 

But for JSON corrupt files, CUDF will throw an exception, I just add a test for GPU JSON reading.

### DIFF
--- a/integration_tests/src/main/python/json_test.py
+++ b/integration_tests/src/main/python/json_test.py
@@ -16,9 +16,10 @@ import pytest
 
 from asserts import assert_gpu_and_cpu_are_equal_collect
 from data_gen import *
-from src.main.python.marks import approximate_float, allow_non_gpu
+from conftest import is_databricks_runtime
+from marks import approximate_float, allow_non_gpu, ignore_order
 
-from src.main.python.spark_session import with_cpu_session
+from spark_session import with_cpu_session, with_gpu_session
 
 json_supported_gens = [
     # Spark does not escape '\r' or '\n' even though it uses it to mark end of record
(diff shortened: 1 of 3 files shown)


wbo4958:
build

jlowe (commented):
Doesn't GpuTextBasedPartitionReader need a try..catch block around `readToTable(Boolean)` to cover the case of the filesystem returning an I/O exception or runtime exception?  Spark CPU covers this in a central place via `FilePartitionReader.next` which skips files that return I/O errors or other runtime exceptions when `ignoreCorruptFiles` is true.

wbo4958:
> Doesn't GpuTextBasedPartitionReader need a try..catch block around `readToTable(Boolean)` to cover the case of the filesystem returning an I/O exception or runtime exception? Spark CPU covers this in a central place via `FilePartitionReader.next` which skips files that return I/O errors or other runtime exceptions when `ignoreCorruptFiles` is true.

Yeah, we don't need to do that, since Spark itself has helped us for this.

For V1, we will get FileScanRDD or GpuFileScanRDD, both of which has implemented ignoreCorruptFiles/ignoreMissingValues

For V2, we will get GpuDataSourceRDD, and the GpuTextBasedPartitionReader will be wrapped into FilePartitionReader which has implemented the ignoreCorruptFiles/ignoreMissingValues.

I added the `test_json_read_with_corrupt_files` for both v1 and v2 scenarios, which can tested the above scenarios.

PR #4812, opened 2022-02-17
Default integration test configs to allow negative decimal scale [databricks]

jlowe:
This changes the integration test setup to always request support for negative decimal scales by default.  If desired, tests could still override this by specifying a conf that explicitly sets the config to disallow it.  This reduces a lot of boilerplate on tests that have been copied throughout.

A reusable conf for enabling ANSI was also added to reduce the test boilerplate for this conf which appears in a number of places.


jlowe:
build

revans2 (approved):

PR #4819, opened 2022-02-17
Throw exception if casting from double to datetime 

### LINKED ISSUE #4342: Add a config to allow cast between Datetime and Numeric in ANSI mode
As part of https://github.com/apache/spark/commit/2afa52d203. Spark has added a config to cast numeric <> datetime. 

We should support this

razajafri:
For Spark 3.3.0+ there is a check added to throw an exception if casting from double to datetime in ANSI mode. 

fixes #4342 

### DIFF
--- a/integration_tests/src/main/python/cast_test.py
+++ b/integration_tests/src/main/python/cast_test.py
@@ -294,10 +294,7 @@ def test_cast_string_to_negative_scale_decimal():
 @pytest.mark.parametrize('type', [DoubleType(), FloatType()])
 def test_cast_double_to_timestamp(type):
     def fun(spark):
-        data=[float("inf"), float("-inf"), float("nan")]
-        df=spark.createDataFrame(
-            SparkContext.getOrCreate().parallelize(data),
-            StructType([StructField('a', type)])
-        )
-        return df.select(f.col('a').cast(TimestampType())).collect()
-    assert_gpu_and_cpu_error(fun, {"spark.sql.ansi.enabled": True}, "org.apache.spark.SparkException")
+        data=[float("inf"),float("-inf"),float("nan")]
+        df = spark.createDataFrame(data, DoubleType())
+        return df.select(f.col('value').cast(TimestampType())).collect()
+    assert_gpu_and_cpu_error(fun, {"spark.sql.ansi.enabled": True}, "java.time.DateTimeException")
--- a/integration_tests/src/main/python/cast_test.py
+++ b/integration_tests/src/main/python/cast_test.py
@@ -296,7 +296,7 @@ def test_cast_double_to_timestamp(type):
     def fun(spark):
         data=[float("inf"), float("-inf"), float("nan")]
         df=spark.createDataFrame(
-            SparkContext.getOrCreate().parallelize([data]),
+            SparkContext.getOrCreate().parallelize(data),
             StructType([StructField('a', type)])
         )
         return df.select(f.col('a').cast(TimestampType())).collect()
--- a/integration_tests/src/main/python/cast_test.py
+++ b/integration_tests/src/main/python/cast_test.py
@@ -16,7 +16,7 @@ import pytest
 
 from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_and_cpu_are_equal_sql, assert_gpu_and_cpu_error, assert_gpu_fallback_collect, assert_py4j_exception
 from data_gen import *
-from spark_session import is_before_spark_320, with_gpu_session
+from spark_session import is_before_spark_320, is_before_spark_330, with_gpu_session
 from marks import allow_non_gpu, approximate_float
 from pyspark.sql.types import *
 from spark_init_internal import spark_version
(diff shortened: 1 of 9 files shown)


sql-plugin/src/main/301until330-all/scala/com/nvidia/spark/rapids/shims/v2/RapidsErrorUtils.scala
```diff
@@ -30,4 +30,8 @@ object RapidsErrorUtils {
     throw new NoSuchElementException(s"Key: ${elementKey} " +
       s"does not exist in any one of the rows in the map column")
   }
+
+  def preprocessCastFloatToTimestamp(input: ColumnView): Unit = {
```
jlowe, line 34: `RapidsErrorUtils` seems like a random class to throw this method into.  I thought the point of this class is to encapsulate how exceptions are thrown once someone has decided to throw an exception rather than house every possible method that might throw on some version of Spark?

jlowe, line 34: Nit: It would be nice if this method were a bit more descriptive.  `preprocess` and no return type doesn't say much about what it's doing.  Maybe something like `checkAnsiCastFloatToTimestamp`?

integration_tests/src/main/python/cast_test.py
```diff
@@ -289,3 +289,12 @@ def test_cast_string_to_negative_scale_decimal():
...
             f.col('a').cast(DecimalType(8, -3))))
+
+@pytest.mark.skipif(is_before_spark_330(), reason="ansi cast throws exception only in 3.3.0+")
+@pytest.mark.parametrize('type', [DoubleType(), FloatType()])
+def test_cast_double_to_timestamp(type):
+    def fun(spark):
+        data=[float("inf"),float("-inf"),float("nan")]
+        df = spark.createDataFrame(data, DoubleType())
+        return df.select(f.col('value').cast(TimestampType())).collect()
+    assert_gpu_and_cpu_error(fun, {"spark.sql.ansi.enabled": True}, "java.time.DateTimeException")
```
jlowe, line 300:
```suggestion
    assert_gpu_and_cpu_error(fun, ansi_enabled_conf, "java.time.DateTimeException")
```

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/Spark33XShims.scala
```diff
@@ -33,7 +33,7 @@ import org.apache.spark.sql.execution.datasources.v2.orc.OrcScan
 import org.apache.spark.sql.execution.datasources.v2.parquet.ParquetScan
 import org.apache.spark.sql.types.StructType
 
-trait Spark33XShims extends Spark33XFileOptionsShims {
+trait Spark33XShims extends Spark33XFileOptionsShims with Arm {
```
jlowe, line 36: Is this change relevant to the PR anymore?

integration_tests/src/main/python/cast_test.py
```diff
@@ -289,3 +289,15 @@ def test_cast_string_to_negative_scale_decimal():
...
+@pytest.mark.parametrize('type', [DoubleType(), FloatType()])
+def test_cast_double_to_timestamp(type):
+    def fun(spark):
+        data=[float("inf"), float("-inf"), float("nan")]
+        df=spark.createDataFrame(
+            SparkContext.getOrCreate().parallelize([data]),
+            StructType([StructField('a', type)])
+        )
+        return df.select(f.col('a').cast(TimestampType())).collect()
+    assert_gpu_and_cpu_error(fun, {"spark.sql.ansi.enabled": True}, "org.apache.spark.SparkException")
```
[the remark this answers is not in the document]
jlowe: This comment was not addressed?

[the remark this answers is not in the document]
razajafri: This conf was removed as part of https://github.com/apache/spark/pull/34985. It now only relies on the ANSI mode

sql-plugin/src/main/301until330-all/scala/com/nvidia/spark/rapids/shims/v2/RapidsErrorUtils.scala
razajafri: I am fine with that name. I can rename it

sql-plugin/src/main/scala/com/nvidia/spark/rapids/FloatUtils.scala
```diff
@@ -63,21 +63,21 @@ object FloatUtils extends Arm {
...
   }
 
-  def infinityToNulls(vec: ColumnVector): ColumnVector = {
-    def getInfinityVector: ColumnVector = {
-      if (vec.getType == DType.FLOAT64) {
-        ColumnVector.fromDoubles(Double.PositiveInfinity, Double.NegativeInfinity)
-      } else {
-        ColumnVector.fromFloats(Float.PositiveInfinity, Float.NegativeInfinity)
-      }
+  def getInfinityVector(vec: ColumnView): ColumnVector = {
```
jlowe, line 66: This should just take a `DType`.  Requiring a `ColumnView` is a bit excessive when all it needs is a type, and it makes for a confusing read of the caller's code, since it's not immediately clear why `getInfinityVector` would need the input column to do its job, whereas passing the type as a parameter is very clear.

jlowe (approved):

jlowe:
build

andygrove (approved):

PR #4823, opened 2022-02-18
Reduce redundancy of decimal testing [databricks]

jlowe:
Relates to #4538 

Reduces the number of decimal types being tested for most tests that aren't really focused on decimal semantics.  Now most tests that test across types will test only three decimal combinations, a 32-bit decimal, a 64-bit decimal, and a 128-bit decimal.  In addition the following notable changes were made:
- `decimal_gens` now includes 128-bit decimals.
- `all_gen` now includes 128-bit decimals

### DIFF
--- a/integration_tests/src/main/python/arithmetic_ops_test.py
+++ b/integration_tests/src/main/python/arithmetic_ops_test.py
@@ -31,13 +31,36 @@ _no_overflow_multiply_gens = [
     IntegerGen(min_val = 1, max_val = 1000, special_cases=[]),
     LongGen(min_val = 1, max_val = 3000, special_cases=[])]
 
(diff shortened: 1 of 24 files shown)


jlowe:
build

revans2 (dismissed):

jlowe:
The cache failures on Databricks 9.1 were because the cache tests (which only run on Spark 3.1+) failed on the added DECIMAL128 inputs.  Filed #4826 and excluded those from the cache test types.

jlowe:
build

jlowe:
Note that this PR appears to reduce the overall python tests count during premerge by about 2000 tests, from 16550 down to 14494.

revans2 (approved):

Issue #4824, opened 2022-02-18
[FEA] Support reading decimals from JSON and CSV


andygrove:
**Is your feature request related to a problem? Please describe.**
I would like to be able to read decimal types from CSV and JSON inputs.

**Describe the solution you'd like**
I would like to be able to read decimal types from CSV and JSON  inputs.

**Describe alternatives you've considered**
None

**Additional context**
None

Issue #4646, opened 2022-01-27
[FEA] Make JSON parsing of `NaN` and `Infinity` values fully compatible with Spark


andygrove:
**Is your feature request related to a problem? Please describe.**
In https://github.com/NVIDIA/spark-rapids/pull/4637 we are adding partial support for parsing `NaN` and `Infinity` values from JSON but there are many edge cases still where we are not compatible.

**Describe the solution you'd like**
The XFAIL test in `json_test.py` that links to this issue should pass.

**Describe alternatives you've considered**
None

**Additional context**
None

Issue #4615, opened 2022-01-24
[FEA]JSON reader: support "allowNonNumericNumbers"


GaryShen2008:
From the [Spark JSON option](https://spark.apache.org/docs/latest/sql-data-sources-json.html#data-source-option),
allowNonNumericNumbers: 

Allows JSON parser to recognize set of “Not-a-Number” (NaN) tokens as legal floating number values.
+INF: for positive infinity, as well as alias of +Infinity and Infinity.
-INF: for negative infinity, alias -Infinity.
NaN: for other not-a-numbers, like result of division by zero.

nartal1:
+INF and -INF works when this option is set. `NaN` returns as Null. 
I think it would be addressed in https://github.com/NVIDIA/spark-rapids/pull/4637. 
So it would be mostly adding tests once `NaN` issue is resolved.

andygrove:
Spark's behavior is pretty odd with this option. Setting `allowNonNumericNumbers` to `false` still allows some variations of `NaN` and `Infinity` (tested with Spark 3.2.0).

I tested with this input file:

``` json
{ "number": "NaN" }
{ "number": NaN }
{ "number": "+INF" }
{ "number": +INF }
{ "number": "-INF" }
{ "number": -INF }
{ "number": "INF" }
{ "number": INF }
{ "number": Infinity }
{ "number": +Infinity }
{ "number": -Infinity }
{ "number": "Infinity" }
{ "number": "+Infinity" }
{ "number": "-Infinity" }
```

``` scala
import org.apache.spark.sql.types._
val schema = StructType(Seq(StructField("number", DataTypes.FloatType, false)))

val df = spark.read.format("json").schema(schema).option("allowNonNumericNumbers", "false").json("/home/andy/Documents/nan_valid.json")

df.show

+---------+
|   number|
+---------+
|      NaN|
|     null|
|     null|
|     null|
|     null|
|     null|
|     null|
| Infinity|
|-Infinity|
+---------+


val df = spark.read.format("json").schema(schema).option("allowNonNumericNumbers", "true").json("/home/andy/Documents/nan_valid.json")

df.show

+---------+
|   number|
+---------+
|      NaN|
|      NaN|
| Infinity|
|-Infinity|
| Infinity|
| Infinity|
|-Infinity|
| Infinity|
|-Infinity|
+---------+
```

andygrove:
[...]

PR #4825, opened 2022-02-18
Initial support for reading decimal types from JSON and CSV

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/4824 and https://github.com/NVIDIA/spark-rapids/issues/4615 and https://github.com/NVIDIA/spark-rapids/issues/4646

The main change in this PR is adding support for reading decimal values from CSV and JSON by reading them as strings and then casting them to decimal.

There is also a fix for a bug in handling invalid JSON floating-point values such as `.1` and `1.`which was discovered during this work.

There are also some updates to the compatibility guide related to previous JSON and CSV PRs. I can PR those docs changes separately if that is preferred.


docs/compatibility.md
```diff
@@ -283,14 +283,7 @@ will produce a different result compared to the plugin.
...
 
- * [spark.rapids.sql.csv.read.bool.enabled](configs.md#sql.csv.read.bool.enabled)
- * [spark.rapids.sql.csv.read.byte.enabled](configs.md#sql.csv.read.byte.enabled)
  * [spark.rapids.sql.csv.read.date.enabled](configs.md#sql.csv.read.date.enabled)
- * [spark.rapids.sql.csv.read.double.enabled](configs.md#sql.csv.read.double.enabled)
- * [spark.rapids.sql.csv.read.float.enabled](configs.md#sql.csv.read.float.enabled)
- * [spark.rapids.sql.csv.read.integer.enabled](configs.md#sql.csv.read.integer.enabled)
- * [spark.rapids.sql.csv.read.long.enabled](configs.md#sql.csv.read.long.enabled)
- * [spark.rapids.sql.csv.read.short.enabled](configs.md#sql.csv.read.short.enabled)
  * [spark.rapids.sql.csvTimestamps.enabled](configs.md#sql.csvTimestamps.enabled)
```
andygrove, line 287: These changes are not specific to this PR and could be PR'd separately

integration_tests/src/test/resources/decimals.csv
```diff
@@ -0,0 +1,14 @@
...
+0
+1.
+0.12
+.12
++.12
+-.12
+1
+1.01
+12.34
+12.3456
```
revans2, line 12: Could we have a few more numbers to see if Spark is rounding when the scale is larger than expected, or if it is truncating?

integration_tests/src/test/resources/decimals.json
```diff
@@ -0,0 +1,12 @@
...
+{ "number": 12 }
+{ "number": 12.0 }
+{ "number": 12. }
+{ "number": .34 }
+{ "number": +.34 }
+{ "number": -.34 }
+{ "number": 0.34 }
+{ "number": 12.34 }
+{ "number": 12.3456 }
+{ "number": 12.345678 }
```
revans2, line 11: Same here it would be nice to have some numbers that would possibly round down instead of up if Spark was rounding.

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
```diff
@@ -370,4 +368,46 @@ class JsonPartitionReader(
     }
   }
 
+  /**
+   * JSON has strict rules about valid numeric formats. See https://www.json.org/ for specification.
+   *
+   * Spark then has it's own rules for supporting NaN and Infinity, which are not
```
revans2, line 374: I think `it's` needs to be `its`

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
```diff
@@ -370,4 +368,46 @@ class JsonPartitionReader(
...
+      "^" +
+      "(?:" +
+        "(?:-?[0-9]+(?:\\.[0-9]+)?(?:[eE][\\-\\+]?[0-9]+)?)" +
+        "|NaN" +
+        "|(?:[\\+\\-]?INF)" +
+        "|(?:[\\-\\+]?Infinity)" +
+      ")" +
+      "$"
+    } else {
+      "^-?[0-9]+(?:\\.[0-9]+)?(?:[eE][\\-\\+]?[0-9]+)?$"
```
revans2, line 392: Is there any way we can do this without a regular expression?  If not could we file a follow on issue to figure out how to speed this up?

andygrove: Yes, I can rework this to avoid regexp. I will do so in this PR.

revans2 (approved):
Looking good.

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
andygrove: I simplified this a bit but it is still calling into `GpuCast` which applies additional complex regular expressions that are mostly redundant because we have already sanitized the input here, so I have filed https://github.com/NVIDIA/spark-rapids/issues/4837 to optimize this farther

andygrove:
build

andygrove:
@revans2 Could I get another review on this one? There were changes after your last approval, to reduce the amount of regexp in use.

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
```diff
@@ -370,4 +370,47 @@ class JsonPartitionReader(
...
+   *
+   * Spark then has its own rules for supporting NaN and Infinity, which are not
+   * valid numbers in JSON.
+   */
+  private def sanitizeNumbers(input: ColumnVector): ColumnVector = {
+    // Note that this is not 100% consistent with Spark versions prior to Spark 3.3.0
+    // due to https://issues.apache.org/jira/browse/SPARK-38060
+    // cuDF `isFloat` supports some inputs that are not valid JSON numbers, such as `.1`, `1.`,
+    // and `+1` so we use a regular expression to match valid JSON numbers instead
+    val jsonNumberRegexp = "^-?[0-9]+(?:\\.[0-9]+)?(?:[eE][\\-\\+]?[0-9]+)?$"
```
revans2, line 384: nit: can you file a follow on issue in the spark rapids repo to look into ways to remove the regular expression? I am hopeful it will be fixed as a part of the rewrite of JSON parsing, but if not I want us to track it.

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
```diff
@@ -370,4 +370,47 @@ class JsonPartitionReader(
...
+  }
+
+  override def castStringToFloat(input: ColumnVector, dt: DType): ColumnVector = {
+    withResource(sanitizeNumbers(input)) { sanitizedInput =>
+      super.castStringToFloat(sanitizedInput, dt)
+    }
+  }
+
+  override def castStringToDecimal(input: ColumnVector, dt: DecimalType): ColumnVector = {
+    withResource(sanitizeNumbers(input)) { sanitizedInput =>
```
revans2, line 411: nit: Decimal does not support NaN or Inf, etc. It would be nice to not have to to the extra checks if we are parsing decimal, but allowNonNumbericNumbers happens to be set.

revans2 (approved):
Just two nits related to performance that I can live without.

sql-plugin/src/main/scala/org/apache/spark/sql/catalyst/json/rapids/GpuJsonScan.scala
revans2: Sorry you already did this one :)

PR #4827, opened 2022-02-18
Make InternalExclusiveModeGpuDiscoveryPlugin and ExplainPlanImpl as protected class.

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/4022.  
Went through the docs and checked all the public classes which can be set. In this PR, we are making InternalExclusiveModeGpuDiscoveryPlugin as protected as it should be loaded only through reflection. 
Updated the comments for  ParquetCachedBatchSerializer as well. 
Rest of the public classes - GpuKryoRegistrator and RapidsShuffleManager doesn't require any changes.

### DIFF
--- a/sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/ParquetCachedBatchSerializer.scala
+++ b/sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/ParquetCachedBatchSerializer.scala
@@ -257,8 +257,9 @@ private case class CloseableColumnBatchIterator(iter: Iterator[ColumnarBatch]) e
 }
 
 /**
- * This class assumes, the data is Columnar and the plugin is on. This is meant to be loaded
- * by reflection only.
+ * This class assumes, the data is Columnar and the plugin is on.
+ * Note, this class should not be referenced directly in source code.
+ * It should be loaded by reflection using ShimLoader.newInstanceOf, see ./docs/dev/shims.md
  */
 protected class ParquetCachedBatchSerializer extends GpuCachedBatchSerializer with Arm {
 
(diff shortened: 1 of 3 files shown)


gerashegalov:
from the classes referenced in ShimLoader via newInstanceOf it looks we need to cover `com.nvidia.spark.rapids.ExplainPlanImpl` as well.

nartal1:
> from the classes referenced in ShimLoader via newInstanceOf it looks we need to cover `com.nvidia.spark.rapids.ExplainPlanImpl` as well.

Thanks, @gerashegalov  for pointing this out. Updated the PR.  PTAL.

sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/v2/ParquetCachedBatchSerializer.scala
```diff
@@ -257,7 +257,8 @@ private case class CloseableColumnBatchIterator(iter: Iterator[ColumnarBatch]) e
 }
 
 /**
- * This class assumes, the data is Columnar and the plugin is on
+ * This class assumes, the data is Columnar and the plugin is on. This is meant to be loaded
+ * by reflection only.
```
gerashegalov, line 261: last nit: let us make this and other comments more precise: 
> Note, this class should not referenced directly in source code. It should be loaded by reflection using ShimLoader.newInstanceOf, see  ./docs/dev/shims.md

gerashegalov (approved):
LGTM

gerashegalov:
build

nartal1:
Thanks @gerashegalov  for the review.

nartal1:
build

PR #4835, opened 2022-02-22
Trim parquet_write_test to reduce integration test runtime

jlowe:
Relates to #4538.  This significantly reduces the runtime of `parquet_write_test.py` by reducing the overhead of some types being tested and reducing redundant test configurations.  On the type overhead reduction front, one type being tested was an array-of-array-of-map-of-struct-of-map-of-string-to-string.  Each string was averaging 15 characters, and each map was 20 elements, and each array level was 10 entries, so there were quite a lot of string characters being processed on the CPU for data generation.  This was causing tests involving those types to execute on the order of minutes for that specific type.  Reducing the length of the maps and arrays solved this issue while still providing the same type coverage.

On the redundant test configuration front, there were many tests that were varying the timestamp write configs for _all_ types, even those without timestamps anywhere in the type hierarchy.  This was solved by creating a separate test to test timestamp types and varying the timestamp config for just that test, and removing that dimension on all the other tests.

The tests were also varying the reader configuration and datasource version as dimensions across many tests, but this file is for testing Parquet writes rather than reads, and these settings don't change the behavior of the write code.  Parquet read tests are already covered by `parquet_test.py`, so these test dimensions were removed.

### DIFF
--- a/integration_tests/src/main/python/parquet_write_test.py
+++ b/integration_tests/src/main/python/parquet_write_test.py
@@ -59,12 +59,18 @@ parquet_basic_map_gens = [MapGen(f(nullable=False), f()) for f in [
     limited_timestamp]] + [simple_string_to_string_map_gen,
                            MapGen(DecimalGen(20, 2, nullable=False), decimal_gen_128bit)]
 
-parquet_struct_gen = [StructGen([['child' + str(ind), sub_gen] for ind, sub_gen in enumerate(parquet_basic_gen)]),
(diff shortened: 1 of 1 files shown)


jlowe:
build

revans2 (approved):

jlowe:
Looks like it timed out connecting to the Databricks cluster.  Kicking again.

jlowe:
build

jlowe:
build

jlowe:
Databricks clusters continue to encounter rsync failures preventing CI from proceeding.

PR #4838, opened 2022-02-22
Trim cache tests to improve integration test time [databricks]

jlowe:
Relates to #4538.  The cache tests were testing a bunch of different join types, but we already have `join_test.py` to test joins.  For `cache_test.py` we're primarily interested in the data types entering the cached dataframe and what we do to the cached dataframe.  This removes all the join dimensions on the tests and reduces the join-related cache tests to just two tests, a simple inner join that is cached and a simple inner join that's cached then filtered.


revans2 (approved):

jlowe:
build

PR #4839, opened 2022-02-22
Optionally return null if element not exists map/array

### LINKED ISSUE #4668: [FEA][Audit] - [SPARK-37750][SQL] ANSI mode: optionally return null result if element not exists in array/map
This PR adds a config which controls whether it throws exceptions or returns null results  if array index is out of bound or map key does not exist for arrays and maps respectively. Have to check if we support ANSI mode for these types and if we want to match the same behavior as CPU.
PR: https://github.com/apache/spark/commit/08fd5013ad
Follow on PR to change config name: https://github.com/apache/spark/commit/0b9f120c99

amahussein:
fixes #4668 

Spark 330 Added a new configuration `spark.sql.ansi.failOnElementNotExists` which controls whether throwing exceptions or returning null results when element not exists in the `[]` operator in array/map type. [Spark-37750 ](https://issues.apache.org/jira/browse/SPARK-37750) 
The default value of the new configuration is true.

This PR adds support of the new flag for rapids.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
@@ -221,6 +221,12 @@ trait SparkShims {
    */
   def shouldFallbackOnAnsiTimestamp(): Boolean
 
+  /**
+   * This is to support ANSI mode: optionally return null result if element not exists
+   * in array/map.
+   */
+  def shouldFailOnElementNotExists(): Boolean = false
+
   def createTable(table: CatalogTable,
     sessionCatalog: SessionCatalog,
     tableLocation: Option[URI],
--- a/integration_tests/src/main/python/array_test.py
+++ b/integration_tests/src/main/python/array_test.py
@@ -120,10 +120,11 @@ def test_array_contains_for_nans(data_gen):
     assert_gpu_and_cpu_are_equal_collect(main_df)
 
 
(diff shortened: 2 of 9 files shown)


amahussein:
build

amahussein:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
```diff
@@ -221,6 +221,12 @@ trait SparkShims {
    */
   def shouldFallbackOnAnsiTimestamp(): Boolean
 
+  /**
+   * This is to support ANSI mode: optionally return null result if element not exists
+   * in array/map.
+   */
+  def shouldFailOnElementNotExists(): Boolean = false
```
revans2, line 228: So by default it does not fail, even if ANSI is enabled? 

Is the following correct for what we want?
|version | throw exception if key not found...|
|-|-|
|3.0.*| Never|
|3.1.*| If ANSI is enabled |
|3.2.*| If ANSI is enabled |
|3.3.* | If ANSI is enabled and strictIndexOperator is enabled (SQLConf includes both together in the helper method) |

revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/SparkShims.scala
amahussein: The table is correct.

`strictIndexOperator` is true by default

PR #4848, opened 2022-02-23
Profile tool: fix printing of task failed reason

tgravescs:
In the profiling tool task failed table, we were just doing a toString on the task end reason. This caused it to some times have an address in the output and that address could change between runs.  Fix it so we use the toErrorString to get the nicely formatted string.


tgravescs:
build

tools/src/test/scala/com/nvidia/spark/rapids/tool/profiling/HealthCheckSuite.scala
```diff
@@ -58,8 +58,14 @@ class HealthCheckSuite extends FunSuite {
     val healthCheck = new HealthCheck(apps)
     for (app <- apps) {
       val failedTasks = healthCheck.getFailedTasks
+      // the end reason gets the delimtier changed when writing to CSV so to compare properly
```
jlowe, line 61:
```suggestion
      // the end reason gets the delimiter changed when writing to CSV so to compare properly
```

tgravescs:
build

nartal1 (approved):

PR #4850, opened 2022-02-23
Also check cuda runtime version when using the ASYNC allocator

rongou:
It's possible to use the newest driver with a runtime < 11.2, in which case ASYNC would break.


rongou:
build

abellina (approved):

jlowe (approved):

Issue #4849, opened 2022-02-23
[FEA] Support parsing dates in JSON reader


andygrove:
**Is your feature request related to a problem? Please describe.**
We should support reading dates from JSON, respecting timeParserPolicy.

**Describe the solution you'd like**
We should support reading dates from JSON, respecting timeParserPolicy.

**Describe alternatives you've considered**
None

**Additional context**
None

Issue #1091, opened 2020-11-10
[BUG] Incorrect values when parsing dates from timestamps stored in CSV files


andygrove:
**Describe the bug**
If I specify a schema with a `DateType` when reading from a CSV file containing timestamps, I get corrupt data when the plugin is enabled.

**Steps/Code to reproduce bug**

Create `tests/src/test/resources/timestamps.csv` CSV file:

```csv
2019-01-03T12:34:56.123456,1
2019-01-03T12:34:56.123456,1
2019-01-03T12:34:56.123456,1
2019-01-05T12:34:56.123456,2
2019-01-05T12:34:56.123456,3
2019-01-06T12:34:56.123456,6
```

Add this method to `tests/src/test/scala/com/nvidia/spark/rapids/SparkQueryCompareTestSuite.scala`:

```scala
  def timestampsAsDatesCsvDf= {
    fromCsvDf("timestamps.csv", StructType(Array(
      StructField("dates", DateType, false),
      StructField("ints", IntegerType, false)
    )))(_)
  }
```

Add these tests to `tests/src/test/scala/com/nvidia/spark/rapids/CsvScanSuite.scala`:

```scala
  testSparkResultsAreEqual(
    "Test CSV parse dates",
    datesCsvDf,
    conf=new SparkConf()) {
    df => df.withColumn("next_day", date_add(col("dates"), lit(1)))
  }

  testSparkResultsAreEqual(
    "Test CSV parse timestamps as dates",
    timestampsAsDatesCsvDf,
    conf=new SparkConf()) {
    df => df.withColumn("next_day", date_add(col("dates"), lit(1)))
  }
```

The first test passes but the second test fails with:

```
CPU: WrappedArray([2019-01-03,1,2019-01-04], [2019-01-03,1,2019-01-04], [2019-01-03,1,2019-01-04], [2019-01-05,2,2019-01-06], [2019-01-05,3,2019-01-06], [2019-01-06,6,2019-01-07])

GPU: WrappedArray([0718-10-22,1,0718-10-23], [0718-10-22,1,0718-10-23], [0718-10-22,1,0718-10-23], [2280-05-20,2,2280-05-21], [2280-05-20,3,2280-05-21], [3779-09-03,6,3779-09-04])
```

**Expected behavior**
The tests should both pass because the output should be the same with or without the plugin enabled.

**Environment details (please complete the following information)**
Running tests in IDE.

**Additional context**
N/A

sameerz:
[...]

Issue #123, opened 2020-06-08
[FEA] add in support for dayfirst formats in the CSV parser


revans2:
**Is your feature request related to a problem? Please describe.**
In the cudf CSV parser there is the possibility to support formats like 'dd/MM/yyyy' or 'dd-MM-yyyy' for dates, and similarly for timestamps.  The issue is that we need to expose that config setting through the cudf java API and we need to detect it in both the date and the timestamp formats to
see if we can support it.  Because there is only one setting for this in the file we would have to make sure that both of the formats were consistent or if both are used we fall back to the CPU.

PR #4853, opened 2022-02-23
Improve date support in JSON and CSV readers

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/4849 and https://github.com/NVIDIA/spark-rapids/issues/1091 and partially fixes https://github.com/NVIDIA/spark-rapids/issues/1111 and https://github.com/NVIDIA/spark-rapids/issues/123

Adds support for reading date columns from JSON and CSV, respecting `timeParserPolicy`.

Status:
- [x] Basic functionality in place
- [x] Improve test coverage
- [x] Update documentation


integration_tests/src/main/python/csv_test.py
```diff
@@ -167,15 +168,7 @@
...
 
 _enable_all_types_conf = {'spark.rapids.sql.csvTimestamps.enabled': 'true',
-        'spark.rapids.sql.csv.read.bool.enabled': 'true',
-        'spark.rapids.sql.csv.read.date.enabled': 'true',
-        'spark.rapids.sql.csv.read.byte.enabled': 'true',
-        'spark.rapids.sql.csv.read.short.enabled': 'true',
-        'spark.rapids.sql.csv.read.integer.enabled': 'true',
-        'spark.rapids.sql.csv.read.long.enabled': 'true',
-        'spark.rapids.sql.csv.read.float.enabled': 'true',
-        'spark.rapids.sql.csv.read.double.enabled': 'true',
```
andygrove, line 177: Most of these options were removed from the configs in earlier PRs related to JSON support

andygrove:
build

docs/compatibility.md
```diff
@@ -323,17 +322,14 @@ Only a limited set of formats are supported when parsing dates.
 * `"MM/yyyy"`
 * `"MM-dd-yyyy"`
 * `"MM/dd/yyyy"`
+* `"dd-MM-yyyy"`
+* `"dd/MM/yyyy"`
 
 The reality is that all of these formats are supported at the same time. The plugin will only
```
revans2, line 328: I don't think this sentence is true still and might need to be removed

docs/compatibility.md
```diff
@@ -323,17 +322,14 @@ Only a limited set of formats are supported when parsing dates.
 * `"MM/yyyy"`
```
revans2, line 322: If we don't need a config for dates any more do we need this section of the documentation at all any more?

https://github.com/NVIDIA/spark-rapids/issues/869 was duplicated to https://github.com/NVIDIA/spark-rapids/issues/1091 so even if that is all that we keep here, it would be good to update the comment there.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/DateUtils.scala
```diff
@@ -211,4 +213,55 @@ object DateUtils {
   }
 
   case class TimestampFormatConversionException(reason: String) extends Exception
+
+  def tagAndGetCudfFormat(
+      meta: RapidsMeta[_, _, _],
+      sparkFormat: String,
+      parseString: Boolean): String = {
+    var strfFormat = ""
```
revans2, line 221: nit: I would feel better if this was set like.

```
var strfFormat: String
```

Possibly even setting it to null first. Just because an empty format might actually produce results and be harder to debug compered to a null or something the compiler checks is being set.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuBatchScanExec.scala
```diff
@@ -211,14 +212,14 @@ object GpuCSVScan {
...
+        // Spark's CSV parser will parse the string "2020-50-16" to the date 2024/02/16 when
+        // timeParserPolicy is set to LEGACY mode and we would reject this as an invalid date
+        // so we fall back to CPU
+        meta.willNotWorkOnGpu(s"GpuCSVScan does not support timeParserPolicy=LEGACY")
       }
       ShimLoader.getSparkShims.dateFormatInRead(parsedOptions).foreach { dateFormat =>
-        if (!supportedDateFormats.contains(dateFormat)) {
-          meta.willNotWorkOnGpu(s"the date format '${dateFormat}' is not supported'")
-        }
+        DateUtils.tagAndGetCudfFormat(meta, dateFormat, parseString = true)
```
revans2, line 222: So we are throwing out the strfFormat that is returned? should we cache it so we don't need to calculate it again?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTextBasedPartitionReader.scala
```diff
@@ -230,6 +235,26 @@ abstract class GpuTextBasedPartitionReader(
     }
   }
 
+  def dateFormat: String
+
+  def castStringToDate(input: ColumnVector, ansiEnabled: Boolean): ColumnVector = {
```
revans2, line 240: This looks like it is never called with ansiEnabled. Is there a reason to keep this parameter?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTextBasedPartitionReader.scala
```diff
@@ -230,6 +235,26 @@ abstract class GpuTextBasedPartitionReader(
...
+    val cudfFormat = DateUtils.toStrf(dateFormat, parseString = true)
```
revans2, line 241: nit: How expensive is this? Should we cache it?

tests/src/test/scala/com/nvidia/spark/rapids/SparkQueryCompareTestSuite.scala
```diff
@@ -150,7 +150,6 @@ trait SparkQueryCompareTestSuite extends FunSuite with Arm {
 
   def enableCsvConf(): SparkConf = {
```
revans2, line 151: Should we just delete this?

integration_tests/src/main/python/json_test.py
```diff
@@ -204,21 +207,81 @@ def test_json_ts_formats_round_trip(spark_tmp_path, date_format, ts_part, v1_ena
     'floats_invalid.json',
     pytest.param('floats_edge_cases.json', marks=pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/4647')),
     'decimals.json',
+    'dates.json',
+    'dates_invalid.json',
 ])
-@pytest.mark.parametrize('schema', [_bool_schema, _byte_schema, _short_schema, _int_schema, _long_schema, _float_schema, _double_schema, _decimal_10_2_schema, _decimal_10_3_schema])
+@pytest.mark.parametrize('schema', [_bool_schema, _byte_schema, _short_schema, _int_schema, _long_schema, _float_schema, _double_schema, _decimal_10_2_schema, _decimal_10_3_schema, _date_schema])
```
revans2, line 213: nit: can we wrap this line?

revans2 (commented):
Mostly nits outside of the documentation needing to be updated.

revans2:
It also looks like some versions of Spark do not include a dateFormat option.

docs/compatibility.md
andygrove: We still need this section until we also merge the timestamp PR (which will follow on from this one). Issue 869/1091 is now resolved and I have now added tests to confirm.

revans2 (dismissed):
The nits I have left are small enough I am happy to merge this in as is.

andygrove:
build

andygrove:
build

revans2 (approved):

andygrove:
Build failed with Spark 3.3.0

``` 
GpuJsonScan.scala:414: value dateFormat is not a member of org.apache.spark.sql.catalyst.json.JSONOptions
```

I will need to some shim work next

andygrove:
build

andygrove:
build

revans2 (approved):

andygrove:
build

andygrove:
@revans2 I fixed the issues with 3.3.0 and tests are now passing

revans2 (approved):

PR #4854, opened 2022-02-23
Avoid regexp_cost with stringSplit on the GPU using transpilation

### LINKED ISSUE #4685: [FEA] Avoid regexp cost in string_split for escaped characters
In `GpuStringSplit` we detect whether the provided pattern is a simple string or a regular expression so that we can call the appropriate cuDF method. There is an optimization that we could implement to avoid the cost of regexp in some cases. For example, given the pattern `\.` we could transpile to `.` and use the non-regex cuDF method.

As described above.

None

**Additional context**
N/A

NVnavkumar:
Fixes #4685.

This avoids the RegExp cost on the GPU by transpiling simple patterns that only contain a combination of non-special characters and escaped meta characters to simplified strings that can be passed to string split with RegExp disabled.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
@@ -277,7 +277,7 @@ class RegexParser(pattern: String) {
             // word boundaries
             consumeExpected(ch)
             RegexEscaped(ch)
-          case '[' | '\\' | '^' | '$' | '.' | '⎮' | '?' | '*' | '+' | '(' | ')' | '{' | '}' =>
+          case '[' | '\\' | '^' | '$' | '.' | '|' | '?' | '*' | '+' | '(' | ')' | '{' | '}' =>
             // escaped metacharacter
             consumeExpected(ch)
             RegexEscaped(ch)
@@ -446,6 +446,7 @@ object RegexSplitMode extends RegexMode
  *                if matching only (rlike)
  */
 class CudfRegexTranspiler(mode: RegexMode) {
+  private val regexMetaChars = ".$^[]\\|?*+(){}"
 
   // cuDF throws a "nothing to repeat" exception for many of the edge cases that are
   // rejected by the transpiler
@@ -465,6 +466,30 @@ class CudfRegexTranspiler(mode: RegexMode) {
     // write out to regex string, performing minor transformations
     // such as adding additional escaping
     cudfRegex.toRegexString
+  }
+
+  def transpileToSplittableString(e: RegexAST): Option[String] = {
+    e match {
+      case RegexEscaped(ch) if regexMetaChars.contains(ch) => Some(ch.toString)
+      case RegexChar(ch) if !regexMetaChars.contains(ch) => Some(ch.toString)
+      case RegexSequence(parts) =>
+        parts.foldLeft[Option[String]](Some("")) { (all, x) => 
+          all match {
+            case Some(current) =>
+              transpileToSplittableString(x) match {
+                case Some(y) => Some(current + y)
+                case _ => None
+              }
+            case _ => None
+          }
+        }
+      case _ => None
+    }
+  }
+
+  def transpileToSplittableString(pattern: String): Option[String] = {
+    val regex = new RegexParser(pattern).parse()
+    transpileToSplittableString(regex)
   }
 
   private def isRepetition(e: RegexAST): Boolean = {
@@ -666,7 +691,7 @@ class CudfRegexTranspiler(mode: RegexMode) {
           throw new RegexUnsupportedException(
             "regexp_replace on GPU does not support repetition with {0,}")
 
-        case (_, QuantifierFixedLength(0)) | (_, QuantifierVariableLength(0,Some(0)))
+        case (_, QuantifierFixedLength(0)) | (_, QuantifierVariableLength(0, Some(0)))
           if mode != RegexFindMode =>
           throw new RegexUnsupportedException(
             "regex_replace and regex_split on GPU do not support repetition with {0} or {0,0}")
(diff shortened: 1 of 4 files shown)


tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -436,6 +436,26 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
     doAstFuzzTest(Some(REGEXP_LIMITED_CHARS_REPLACE), RegexReplaceMode)
   }
 
+  test("string split - optimized") {
+    val patterns = Set("\\.", "\\$", "\\[", "\\(", "\\}", "\\+", "\\\\", "c\\|d")
```
andygrove, line 440: It would be good to add some patterns without escaped characters as well, such as `","`

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -610,8 +666,8 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
     input.map(s => s.split(pattern, limit))
   }
 
-  private def gpuSplit(pattern: String, input: Seq[String], limit: Int): Seq[Array[String]] = {
-    val isRegex = RegexParser.isRegExpString(pattern)
+  private def gpuSplit(pattern: String, input: Seq[String],
+                       limit: Int, isRegex: Boolean): Seq[Array[String]] = {
```
andygrove, line 670: nit: indent is off here 

```suggestion
      limit: Int, isRegex: Boolean): Seq[Array[String]] = {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1341,6 +1341,22 @@ abstract class StringSplitRegExpMeta[INPUT <: TernaryExpression](expr: INPUT,
...
+              try {
+                pattern = transpiler.transpile(utf8Str.toString)
+              } catch {
+                case e: RegexUnsupportedException =>
+                  willNotWorkOnGpu(e.getMessage)
+              }
+          }
         } else {
           val str = utf8Str.toString
           isRegExp = RegexParser.isRegExpString(str)
```
andygrove, line 1362: This line appears to be redundant. `isRegExp` is already initialized prior to the `if` statement so it is guaranteed to be `false` if we are hitting the `else` block?

``` scala
        isRegExp = RegexParser.isRegExpString(utf8Str.toString)
        if (isRegExp) {
```

NVnavkumar: In this case, I did go by code that was already in place, also `isRegExp` is declared as a variable (`var`):

```
 var isRegExp: Boolean = false
```

I could rewrite this code to be a little cleaner though

NVnavkumar: never mind, I see the redundancy. I missed this when merging in upstream code.

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
NVnavkumar: Added a few more tests. Note though that the original check for regular expression will probably supersede the check here. 

A thought might be to remove the isRegExpString call in the first place, and just treat non-regexp strings with the same code path. Would that make sense?

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
NVnavkumar: Fixed by putting parameters on each line per databricks scala style guide.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
NVnavkumar: Cleaned this up in latest commit

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
NVnavkumar: I pushed a change to consolidate this logic a bit to remove redundancy.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -277,7 +277,7 @@ class RegexParser(pattern: String) {
             // word boundaries
             consumeExpected(ch)
             RegexEscaped(ch)
-          case '[' | '\\' | '^' | '$' | '.' | '⎮' | '?' | '*' | '+' | '(' | ')' | '{' | '}' =>
+          case '[' | '\\' | '^' | '$' | '.' | '|' | '?' | '*' | '+' | '(' | ')' | '{' | '}' =>
```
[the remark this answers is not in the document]
NVnavkumar: It was some accidental unicode version (https://www.compart.com/en/unicode/U+23AE)

andygrove (approved):

andygrove:
It looks like there was a transient build failure

andygrove:
build

NVnavkumar:
build

andygrove:
The vulnerability scan is failing with `Could not communicate with Black Duck: The connection was not successful for an unknown reason. If an api token is being used, it could be incorrect.`.

sameerz:
build

sameerz:
build

sameerz:
build

Issue #4814, opened 2022-02-17
[FEA] Support element_at with non-literal index


viadea:
I wish we can support element_at with non-literal index.

For example:
```
import org.apache.spark.sql.types._
import org.apache.spark.sql.Row

val arrayData = Seq(
    Row("John",List("apple","orange","banana"),1),
    Row("David",List("apple","orange","banana"),2),
    Row("Harry",List("apple","other"),1)
)

val arraySchema = new StructType().add("name",StringType).add("fruits", ArrayType(StringType)).add("favorite",IntegerType)

val df = spark.createDataFrame(spark.sparkContext.parallelize(arrayData),arraySchema)
df.write.mode("overwrite").format("parquet").save("/tmp/testparquet")
val df2=spark.read.parquet("/tmp/testparquet")
df2.createOrReplaceTempView("df2")

spark.sql("select name,element_at(fruits,favorite) from df2").show()
```

The related unsupported messages:
```
!Expression <ElementAt> element_at(fruits#92, favorite#93, false) cannot run on GPU because ordinal only supports IntegerType if it is a literal value; index/key only supports IntegerType if it is a literal value
```

revans2:
We should update GpuGetArrayItem while we are at it, to make both consistent.

firestarman:
Depends on https://github.com/rapidsai/cudf/issues/10340

firestarman:
blossom is down, will trigger pre-merge once it is back

PR #4858, opened 2022-02-24
Support non-literal index for `GpuElementAt` and `GpuGetArrayItem`[databricks]

firestarman:
This PR is to add the non-literal index (aka indices as a column, which can let different rows have different indices) support for `GpuElementAt` and `GpuGetArrayItem`.

It also updates the relevant tests for them.

close https://github.com/NVIDIA/spark-rapids/issues/4814


firestarman:
build

firestarman:
blossom is down, will trigger pre-merge again once it is back.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -124,67 +124,122 @@ case class GpuElementAt(left: Expression, right: Expression, failOnError: Boolea
...
+            // raised if no valid entry (An entry is valid when both the array row and its index
+            // are not null), the same with what Spark does.
+            val hasLargerIndices = withResource(indices.abs()) { absIndices =>
+              withResource(array.countElements()) { numElements =>
+                absIndices.greaterThan(numElements)
+              }
+            }
+            withResource(hasLargerIndices) { _ =>
+              if(BoolUtils.isAnyValidTrue(hasLargerIndices)) {
+                throw new ArrayIndexOutOfBoundsException("Some indices are out of bound")
```
revans2, line 144: We are not using the RapidsErrorUtils. Is it worth finding the first entry that does not match after this point and getting the exact same error message?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -111,31 +102,80 @@ case class GpuGetArrayItem(child: Expression, ordinal: Expression, failOnError:
...
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector = {
+    val (array, indices) = (lhs.getBase, rhs.getBase)
+    val indicesCol = withResource(Scalar.fromInt(0)) { zeroS =>
+      withResource(indices.lessThan(zeroS)) { hasNegativeIndicesCV =>
+        val hasNegativeIndices = isAnyValidTrue(hasNegativeIndicesCV)
+        if (failOnError) {
+          // Check if any index is out of bound only when ansi mode is enabled.
+          // No exception should be raised if no valid entry (An entry is valid when both
+          // the array row and its index are not null), the same with what Spark does.
+          if(hasNegativeIndices && array.getNullCount != array.getRowCount) {
```
revans2, line 114: nit: in general I think we want to have a space after the `if`. There are several places where this is not happening, and what is more it is inconsistent in the patch.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
```diff
@@ -111,31 +102,80 @@ case class GpuGetArrayItem(child: Expression, ordinal: Expression, failOnError:
...
+            throw new ArrayIndexOutOfBoundsException("Invalid index: some indices are < 0")
+          }
+          val hasLargerIndicesCV = withResource(array.countElements()) { numElements =>
+            indices.greaterOrEqualTo(numElements)
+          }
+          withResource(hasLargerIndicesCV) { _ =>
+            if(isAnyValidTrue(hasLargerIndicesCV)) {
+              // No need to check the validity of array column here, since the validity info
+              // is included in this `hasLargerIndicesCV`.
+              throw new ArrayIndexOutOfBoundsException(
```
revans2, line 125: Same here is it worth it to make this the same as in both code paths?

jlowe (commented):
Minor nits otherwise lgtm.

integration_tests/src/main/python/data_gen.py
```diff
@@ -906,6 +906,7 @@ def gen_scalars_for_sql(data_gen, count, seed=0, force_no_nulls=False):
 
 # Some array gens, but not all because of nesting
 array_gens_sample = single_level_array_gens + nested_array_gens_sample
+array_of_map_gen = ArrayGen(MapGen(StringGen(pattern='key_[0-9]', nullable=False), StringGen(), max_length=10), max_length=10)
```
revans2, line 909: nit: this name is a little generic. Nested types makes it difficult to name things well. Perhaps `array_map_string_string_gen`. It is ugly but there is less ambiguity. Also if this only used in one file, or one place in the code we don't need to create it in data_gen.py

revans2 (approved):
Just some nits that I could live without

revans2:
build

integration_tests/src/main/python/data_gen.py
firestarman: For now, this is only used in `array_test`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/BoolUtils.scala
```diff
@@ -41,4 +41,24 @@ object BoolUtils extends Arm {
...
+  /**
+   * Whether there is any valid row in 'col' and it is true. An empty column will get false.
+   * null rows are skipped.
+   */
+  def isAnyValidTrue(col: ColumnVector): Boolean = {
+    assert(DType.BOOL8 == col.getType, "input column type is not bool")
+    if (col.getRowCount == 0) {
+      return false
+    }
+
```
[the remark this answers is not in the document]
firestarman: Thanks for review, updated

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
firestarman: I think it would be OK since we are on the way to throw an exception to abort.
The main drawback is GPU memory, we need to hold the 'numElements' column until the whole check is done.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
firestarman: updated

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/complexTypeExtractors.scala
firestarman: I add it.

integration_tests/src/main/python/data_gen.py
firestarman: updated

firestarman:
build

jlowe (dismissed):

jlowe:
build

firestarman:
build

revans2 (dismissed):

firestarman:
build

sameerz:
build

firestarman:
resolve the conflicts

firestarman:
build

sperlingxx:
LGTM

sperlingxx (approved):

PR #4872, opened 2022-02-25
Speed up copying decimal column from parquet buffer to GPU buffer

### LINKED ISSUE #4784: [FEA] Improve copying decimal data from CPU columnar data
For now, when we copy decimal column vectors during `HostColumnarToGpu`, it involves a round trip of Decimal encoding and decoding for each value: `cv.getDecimal(i, precision, scale).toUnscaledLong()`, which can be avoided under some circumstances.

Ideally, we can bypass the round trip through feeding `cuDF.HostColumnBuilder` with the underlying storage(int32/int64/binary), if we know the exact type of upstreaming ColumnVector, such as: `WritableColumnVector`.

sperlingxx:

Closes #4784

Adds specialized support for the columnar copy of  `WritableColumnVector` on decimal type. The new implementation copies the unscaled values directly to avoid the round trip of Decimal encoding/decoding.

### DIFF
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
@@ -18,10 +18,6 @@ package com.nvidia.spark.rapids;
 
 import ai.rapids.cudf.HostColumnVector.ColumnBuilder;
 
-import ai.rapids.cudf.HostColumnVectorCore;
-import org.apache.hadoop.hive.ql.exec.vector.DecimalColumnVector;
-import org.apache.spark.sql.execution.datasources.orc.OrcAtomicColumnVector;
-import org.apache.spark.sql.execution.datasources.orc.OrcColumnVector;
 import org.apache.spark.sql.execution.vectorized.WritableColumnVector;
 import org.apache.spark.sql.vectorized.ColumnVector;
 
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
@@ -167,18 +167,15 @@ public class ColumnarCopyHelper {
     }
   }
 
-  public static void decimal32CopyFromParquet(WritableColumnVector cv,
-      ColumnBuilder b, int rows) {
+  public static void decimal32Copy(WritableColumnVector cv, ColumnBuilder b, int rows) {
     intCopy(cv, b, rows);
   }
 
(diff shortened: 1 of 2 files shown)


sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
```diff
@@ -162,7 +167,33 @@ public static void stringCopy(ColumnVector cv, ColumnBuilder b, int rows) {
     }
   }
 
-  // TODO: https://github.com/NVIDIA/spark-rapids/issues/4784
+  public static void decimal32CopyFromParquet(WritableColumnVector cv,
```
jlowe, line 170: This method should not have`Parquet` in the method name.  `WritableColumnVector` may be used by Spark's Parquet code, but it's not necessarily specific to Parquet (it's also used by `InMemoryRelation` in some cases).  I'd just call this `decimal32Copy` and let the type overloading handle cases where we want to copy from different column types.

Comment applies to the other decimal copy methods that were added.

jlowe (commented):
Are there any microbenchmark numbers showing the performance effects of this change?

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
sperlingxx: Applied.

jlowe (dismissed):

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
```diff
@@ -18,6 +18,11 @@
 
 import ai.rapids.cudf.HostColumnVector.ColumnBuilder;
 
+import ai.rapids.cudf.HostColumnVectorCore;
+import org.apache.hadoop.hive.ql.exec.vector.DecimalColumnVector;
+import org.apache.spark.sql.execution.datasources.orc.OrcAtomicColumnVector;
+import org.apache.spark.sql.execution.datasources.orc.OrcColumnVector;
```
revans2, line 24: nit: I don't think that the ORC columns are used and should probably be removed.

firestarman:
build

sameerz:
build

sperlingxx:
build

sql-plugin/src/main/java/com/nvidia/spark/rapids/ColumnarCopyHelper.java
sperlingxx: Yes, I fixed it.

firestarman (approved):

Issue #4876, opened 2022-02-28
[BUG] Nightly build failed on Databricks with "pip: No such file or directory"


NvTimLiu:
**Describe the bug**

Nightly build failed on Databricks with "pip: No such file or directory", due to the update of conda/mamba default to Python 3.10 from Python3.8

```
10:13:40  + PYSPARK_PYTHON=/databricks/conda/envs/cudf-udf/bin/python
10:13:40  ++ which pip
10:13:40  + sudo /databricks/conda/envs/cudf-udf/bin/pip install pytest sre_yield requests pandas pyarrow findspark pytest-xdist pytest-order
10:13:40  sudo: unable to execute /databricks/conda/envs/cudf-udf/bin/pip: No such file or directory
```

**Steps/Code to reproduce bug**
Run nightly build with init script "init_cudf_udf.sh"

NvTimLiu:
This is due to the mismatch of `python=3.8` vs `pip-python=3.10`

When installing manba via conda `[conda install -c conda-forge mamba](https://github.com/NVIDIA/spark-rapids/blob/branch-22.04/jenkins/databricks/init_cudf_udf.sh#L30)` , the default python=3.10 alone with `pip-python=3.10` are installed.

As our integration test runs with python=3.8. Then we tried to down-grade to `[mamba install ... python=3.8 ](https://github.com/NVIDIA/spark-rapids/blob/branch-22.04/jenkins/databricks/init_cudf_udf.sh#L30)`,  the mismatch occoured.

There was no such issue before, as the default python was `python =3.8` for conda/mamba package.

PR #4878, opened 2022-02-28
Set conda/mamba default to Python version to 3.8 [skip ci]

NvTimLiu:
To fix issue: https://github.com/NVIDIA/spark-rapids/issues/4876

Set python version to 3.8 with conda/mamba installation, conda will install pip-python 3.8, instead of the default python 3.10 along with pip-python 3.10. 


jlowe (approved):

revans2 (approved):

pxLi:
Going to force merge this to unblock nightly build

PR #4881, opened 2022-02-28
Fix compilation error caused by ClusteredDistribution parameters

### LINKED ISSUE #4879: [BUG] [SPARK-38237][SQL] ClusteredDistribution clustering keys break build with wrong arguments
[SPARK-38237] Allow ClusteredDistribution to require full clustering keys](https://github.com/apache/spark/pull/35574) changes break RAPIDS build with the following error:

```bash
[INFO] Compiling 255 Scala sources and 29 Java sources to sql-plugin/target/spark330/classes ...
[ERROR] [Error] sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuHashPartitioning.scala:30: wrong number of arguments for pattern org.apache.spark.sql.catalyst.plans.physical.ClusteredDistribution(clustering: Seq[org.apache.spark.sql.catalyst.expressions.Expression],requireAllClusterKeys: Boolean,requiredNumPartitions: Option[Int])
[ERROR] [Error] sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRangePartitioning.scala:68: wrong number of arguments for pattern org.apache.spark.sql.catalyst.plans.physical.ClusteredDistribution(clustering: Seq[org.apache.spark.sql.catalyst.expressions.Expression],requireAllClusterKeys: Boolean,requiredNumPartitions: Option[Int])
[ERROR] two errors found
```

amahussein:
Fixes #4879 


sql-plugin/src/main/301until330-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRangePartitioning.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2022, NVIDIA CORPORATION.
```
jlowe, line 2:
```suggestion
 * Copyright (c) 2020-2022, NVIDIA CORPORATION.
```

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuHashPartitioning.scala
```diff
@@ -27,8 +27,12 @@ case class GpuHashPartitioning(expressions: Seq[Expression], numPartitions: Int)
   override def satisfies0(required: Distribution): Boolean = {
     super.satisfies0(required) || {
       required match {
-        case ClusteredDistribution(requiredClustering, _) =>
-          expressions.forall(x => requiredClustering.exists(_.semanticEquals(x)))
+        case c @ ClusteredDistribution(requiredClustering, requireAllClusterKeys, _) =>
```
jlowe, line 30: Note that while I was looking into this, I noticed we're also missing the changes associated with [SPARK-38124](https://issues.apache.org/jira/browse/SPARK-38124).  If we don't fix that here, we should track it in a followup.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRangePartitioningBase.scala
```diff
@@ -0,0 +1,46 @@
...
+ * sortable records by range into roughly equal ranges. The ranges are determined by sampling
+ * the content of the RDD passed in.
+ *
+ * @note The actual number of partitions created might not be the same
+ * as the `numPartitions` parameter, in the case where the number of sampled records is less than
+ * the value of `partitions`.
+ *
+ * The GpuRangePartitioner is where all of the processing actually happens.
+ */
+abstract class GpuRangePartitioningBase(
```
jlowe, line 36: I'm not sure this added class is a net good change.  There's a minimal amount of reuse happening here, and the code now follows the original Spark code less closely.

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuRangePartitioning.scala
```diff
@@ -0,0 +1,61 @@
...
+          //   then there must be a [a1, b1 c1] larger than [a2, b2, c2], which violates
+          //   RangePartitioning definition. So it's guaranteed that, any [a, b] in a previous
+          //   partition must not be greater(i.e. smaller or equal to) than any [a, b] in the
+          //   following partition. Thus `RangePartitioning(a, b, c)` satisfies
+          //   `OrderedDistribution(a, b)`.
+          val minSize = Seq(requiredOrdering.size, gpuOrdering.size).min
+          requiredOrdering.take(minSize) == gpuOrdering.take(minSize)
+        case c @ ClusteredDistribution(requiredClustering, requireAllClusterKeys, _) =>
+          if (requireAllClusterKeys) {
+            c.areAllClusterKeysMatched(children)
```
revans2, line 52: Do we know if this will work with our GPU children?

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuHashPartitioning.scala
```diff
@@ -27,8 +27,12 @@ case class GpuHashPartitioning(expressions: Seq[Expression], numPartitions: Int)
...
+          if (requireAllClusterKeys) {
+            c.areAllClusterKeysMatched(expressions)
```
revans2, line 32: Same question here. Do we have tests that cover this case? Because the reason we have to override it is so that it will play nicely with GPU expressions.

sql-plugin/src/main/301until330-all/scala/com/nvidia/spark/rapids/shims/v2/GpuRangePartitioning.scala
```diff
@@ -14,34 +14,22 @@
  * limitations under the License.
  */
 
-package com.nvidia.spark.rapids
+package com.nvidia.spark.rapids.shims.v2
 
-import com.nvidia.spark.rapids.shims.v2.ShimExpression
+import com.nvidia.spark.rapids.GpuRangePartitioningBase
```
jlowe, line 19: Should generate a compile error since the class was removed.  Was this built locally before posting?

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuHashPartitioning.scala
amahussein: I got lost inside test cases. I added debug messages inside `columnarEval` but was not able to pinpoint integration tests that cover that code.
At least `com.nvidia.spark.rapids.JoinsSuite` did.

`areAllClusterKeysMatched` is executing something like ` l.semanticEquals(r)` 
`SQLConf.REQUIRE_ALL_CLUSTER_KEYS_FOR_DISTRIBUTION` is false bydefault. Should I file a followup issue to create the tests needs for Range/hash partitioning with `REQUIRE_ALL_CLUSTER_KEYS_FOR_DISTRIBUTION` set to true?
Or do you have quick pointers for me to add basic tests within this PR?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRangePartitioningBase.scala
amahussein: I thought that we may extend that in the future, But that's a good point that it is not really needed now.

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/GpuHashPartitioning.scala
revans2: It is a little more complicated than just seeing that the code is called. I should have explained this in the original question. We do some odd things when replacing expressions. A simple attribute reference (like make sure the keys match column a) is fine. The problem shows up if we are doing computation as a part of that. So things like make sure the keys match column a + column b.  Because we replace the + with a GPU version things don't always work out. I don't know what `areAllClusterKeysMatched` does. So that is probably the first place to look. After that, yes we need tests to verify that this code is being executed and that it is doing the right thing, especially in cases when there is some kind of computation happening.

amahussein: `areAllClusterKeysMatched` in [ClusteredDistribution](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/physical/partitioning.scala#L103) checks that `for-all` expressions they semantically equal to the cluster keys. The previous code used to check `if exists`.

- One possible test is to have a new unit test GpuHashPartitioning then asserts on the return of `GpuHashPartitioning.satisfies` for a different sequence of expressions. But I was not yet sure if this is good enough and which expressions should we care about for the test.
-  Other way is to compare the captured execution plan to the expected. This is more complicated to me because I am not familiar yet with execution plans and where they would end up running GPU/CPU.

jlowe: Would it make sense to do the testing as a followup issue (still for 22.04)?  Currently the 3.3 builds are broken, and merging this would help unblock it.  We could unblock 3.3 builds in parallel with working on tests.  The code is doing the same thing wrt. `semanticEquals` that was done before, so if there's an issue there it's probably not specific to Spark 3.3.

revans2: You are correct and I think we are probably okay with this change. I am just being paranoid. `semanticEquals` involves canonicalizing the expressions and then checking for equality. It should be fine. I am going to just call it good. I just have been bit in the past when only part of something is moved to the GPU, but not all of it. Because the current code works I am going to assume that the new code will work too.

revans2 (approved):

Issue #4846, opened 2022-02-23
[Audit][FEA][SPARK-38094] GPU write Parquet failed after Spark added "spark.sql.parquet.fieldId.write.enabled" key in ParquetFileFormat in Parquet


res-life:
**Describe the bug**
[SPARK-38094](https://issues.apache.org/jira/browse/SPARK-38094) added a new key in ParquetFileFormat for Spark 3.3.0 Parquet: "spark.sql.parquet.fieldId.write.enabled".
And this causes GPU write Parquet failed, the GpuParquetFileFormat should also handle it.

**Steps/Code to reproduce bug**
Download the latest Spark 3.3.0 Snapshot tar.gz file
Run a parquet write IT like:
```
def test_parquet_write(spark_tmp_path):
    data_path = spark_tmp_path + '/PARQUET_DATA'
    gen = IntegerGen()

    assert_gpu_and_cpu_writes_are_equal_collect(
        lambda spark, path: unary_op_df(spark, gen).coalesce(1).write.parquet(path),
        lambda spark, path: spark.read.parquet(path),
        data_path)
```
Errors are:
```
Caused by: java.lang.IllegalArgumentException: For input string: "null"
	at scala.collection.immutable.StringLike.parseBoolean(StringLike.scala:330) ~[scala-library-2.12.15.jar:?]
	at scala.collection.immutable.StringLike.toBoolean(StringLike.scala:289) ~[scala-library-2.12.15.jar:?]
	at scala.collection.immutable.StringLike.toBoolean$(StringLike.scala:289) ~[scala-library-2.12.15.jar:?]
	at scala.collection.immutable.StringOps.toBoolean(StringOps.scala:33) ~[scala-library-2.12.15.jar:?]
	at org.apache.spark.sql.execution.datasources.parquet.SparkToParquetSchemaConverter.<init>(ParquetSchemaConverter.scala:455) ~[spark-sql_2.12-3.3.0-SNAPSHOT.jar:3.3.0-SNAPSHOT]
	at org.apache.spark.sql.execution.datasources.parquet.ParquetWriteSupport.init(ParquetWriteSupport.scala:114) ~[spark-sql_2.12-3.3.0-SNAPSHOT.jar:3.3.0-SNAPSHOT]
	at com.nvidia.spark.rapids.GpuParquetWriter.<init>(GpuParquetFileFormat.scala:343) ~[spark3xx-common/:?]
[...]

PR #4882, opened 2022-03-01
Disable write/read Parquet when Parquet field IDs are used

res-life:
Temporarily disable write/read parquet when schema has a specified Parquet field ID
Solution is: Fallback to CPU when writing Parquet with field id column. Fallback to CPU if 
"spark.sql.parquet.fieldId.read.enabled" is true when reading Parquet.
Contributes #4846


sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/ParquetFieldIdShims.scala
```diff
@@ -27,4 +31,18 @@ object ParquetFieldIdShims {
...
+    if (ParquetUtils.hasFieldIds(schema)) {
+      meta.willNotWorkOnGpu(
+        "Currently not support 'parquet.field.id' in parquet writer, schema is " + schema.json)
+    }
+  }
+
+  def tagGpuSupportReadForFieldId(meta: RapidsMeta[_, _, _], conf: RuntimeConfig): Unit = {
+    if(conf.get(SQLConf.PARQUET_FIELD_ID_READ_ENABLED.key, "false").toBoolean) {
+      meta.willNotWorkOnGpu("Currently not support reading field ids, " +
+          "please set spark.sql.parquet.fieldId.read.enabled as false")
```
jlowe, line 45:
```suggestion
      meta.willNotWorkOnGpu("reading by Parquet field ID is not supported, " +
          "${SQLConf.PARQUET_FIELD_ID_READ_ENABLED} is true")
```

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/ParquetFieldIdShims.scala
```diff
@@ -27,4 +31,18 @@ object ParquetFieldIdShims {
       SQLConf.PARQUET_FIELD_ID_WRITE_ENABLED.key,
       sqlConf.parquetFieldIdWriteEnabled.toString)
   }
+
+  def tagGpuSupportWriteForFieldId(meta: RapidsMeta[_, _, _], schema: StructType): Unit = {
+    if (ParquetUtils.hasFieldIds(schema)) {
```
jlowe, line 36: This check should not be performed unless `SQLConf.PARQUET_FIELD_ID_WRITE_ENABLED` is set to true.

tests/src/test/330+/scala/com/nvidia/spark/rapids/ParquetFieldIdSuite.scala
```diff
@@ -0,0 +1,78 @@
...
+  test("try to write field id") {
+    val tmpFile = File.createTempFile("field-id", ".parquet")
+    try {
+      def withId(id: Int) =
+        new MetadataBuilder().putLong(ParquetUtils.FIELD_ID_METADATA_KEY, id).build()
+      // not support writing field id
+      val schema = new StructType().add("c1", IntegerType, nullable = true, withId(1))
+      val data = (1 to 4).map(i => Row(i))
+
+      assertThrows[IllegalArgumentException] {
```
jlowe, line 42: This is not how we normally test for CPU fallbacks.  `IllegalArgumentException` could catch all sorts of exceptions that aren't actually the one we want/expect and then the test passes for the wrong reasons.  Typically we capture the plan and examine it to verify certain nodes are present and/or others are not present.

Also we prefer to put the fallback tests in the Python integration tests which have some nice framework to handle fallback testing.  See `assert_gpu_fallback_write` and the other fallback assertions in asserts.py.

As for the Parquet field ID metadata, you should be able to construct that in pyspark via something like this:
```python

schema = StructType().add("c1", IntegerType(), nullable=True, metadata={ 'parquet.field.id' : 1 })
```

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/v2/ParquetFieldIdShims.scala
res-life: done

tests/src/test/330+/scala/com/nvidia/spark/rapids/ParquetFieldIdSuite.scala
res-life: Done, moved to pyspark.

jlowe (approved):

revans2 (approved):

firestarman:
build

sameerz:
build

PR #4886, opened 2022-03-01
Fix 312db build [Databricks]

razajafri:
This PR fixes the DB build 

*  Use the `ShuffleQueryStageExec.newReuseInstance` method instead of creating a new instance of `ShuffleQueryStageExec`
* Fixed the shuffleId doesn't override anything error by removing the override


razajafri:
Closing this because I was seeing this on DB 8.2 which we don't support.

PR #4906, opened 2022-03-07
Gh page markdown link check [skip ci]

nvliyuan:
add github action to check broken markdown links.
related to #618

PR #4907, opened 2022-03-07
Add markdown check action

nvliyuan:
add github action to check broken markdown links on dev-branch 22.04
related to #4906

### DIFF
--- a/.github/workflows/markdown-links-check.yml
+++ b/.github/workflows/markdown-links-check.yml
@@ -15,7 +15,9 @@
 # A workflow to check if PR got broken hyperlinks
 name: Check Markdown links
 
-on: push
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
 
 jobs:
   markdown-link-check:
--- a/.github/workflows/markdown-links-check.yml
+++ b/.github/workflows/markdown-links-check.yml
@@ -15,9 +15,7 @@
 # A workflow to check if PR got broken hyperlinks
 name: Check Markdown links
 
-on:
-  pull_request_target:
-    types: [opened, synchronize, reopened]
+on: push
 
 jobs:
   markdown-link-check:
@@ -30,4 +28,6 @@ jobs:
         use-verbose-mode: 'yes'
         check-modified-files-only: 'yes'
         config-file: '.github/workflows/markdown-links-check/markdown-links-check-config.json'
+        base-branch: 'gh-pages'
+        
         
--- a/.github/workflows/markdown-links-check.yml
+++ b/.github/workflows/markdown-links-check.yml
@@ -28,5 +28,6 @@ jobs:
       with:
         max-depth: -1
         use-verbose-mode: 'yes'
+        check-modified-files-only: 'yes'
         config-file: '.github/workflows/markdown-links-check/markdown-links-check-config.json'
         
--- a/.github/workflows/markdown-links-check.yml
+++ b/.github/workflows/markdown-links-check.yml
@@ -15,9 +15,9 @@
 # A workflow to check if PR got broken hyperlinks
 name: Check Markdown links
 
-on:
-  pull_request_target:
-    types: [opened, synchronize, reopened]
+on: push
+#  pull_request_target:
+#    types: [opened, synchronize, reopened]
 
 jobs:
   markdown-link-check:
--- a/.github/workflows/markdown-links-check.yml
+++ b/.github/workflows/markdown-links-check.yml
@@ -15,10 +15,9 @@
 # A workflow to check if PR got broken hyperlinks
 name: Check Markdown links
 
-#on:
-#  pull_request_target:
-#    types: [opened, synchronize, reopened]
-on: push
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
 
 jobs:
   markdown-link-check:
--- a/.github/workflows/markdown-links-check.yml
+++ b/.github/workflows/markdown-links-check.yml
@@ -15,9 +15,9 @@
 # A workflow to check if PR got broken hyperlinks
 name: Check Markdown links
 
-on: push
-#  pull_request_target:
-#    types: [opened, synchronize, reopened]
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
 
 jobs:
   markdown-link-check:
--- a/.github/workflows/markdown-links-check.yml
+++ b/.github/workflows/markdown-links-check.yml
@@ -1,0 +1,14 @@
+name: Check Markdown links
+
+on: push
+
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: gaurav-nelson/github-action-markdown-link-check@v1
+      with:
+        max-depth: -1
+        use-verbose-mode: 'yes'
+        config-file: '.github/workflows/markdown-links-check/markdown-links-check-config.json'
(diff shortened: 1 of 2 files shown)


nvliyuan:
please found the detail information about this pr in #4906

.github/workflows/markdown-links-check.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: gaurav-nelson/github-action-markdown-link-check@v1
+      with:
+        max-depth: -1
+        use-verbose-mode: 'yes'
```
jlowe, line 15: Should we also specify `check-modified-files-only` here?

.github/workflows/markdown-links-check/markdown-links-check-config.json
```diff
@@ -0,0 +1,31 @@
...
+      "pattern": "http:\/\/RM_hostname"
+    },
+    {
+      "pattern": "\/docs\/img\/perf-cost.png"
+    },
+    {
+      "pattern": "\/docs\/img\/ease-of-use.png"
+    },
+    {
+      "pattern": "\/docs\/get-started\/getting-started.md"
```
jlowe, line 16: I'm confused why there are so many exceptions here.  For example, why is the link to getting-started.md an exception when there are many other instances where we are linking to other .md files with a relative path?  Similarly, why are png files an issue?  Are there clues in the verbose log as to what error the link check is getting when trying to access these links?  It shouldn't be a 404, and I'm wondering if we're missing some status codes.

.github/workflows/markdown-links-check.yml
```diff
@@ -0,0 +1,16 @@
+name: Check Markdown links
```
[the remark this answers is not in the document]
nvliyuan: thanks, added

.github/workflows/markdown-links-check.yml
nvliyuan: thanks, added

.github/workflows/markdown-links-check/markdown-links-check-config.json
nvliyuan: thanks, add http status code 0 400 403 and remove exceptions

.github/workflows/markdown-links-check/markdown-links-check-config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "timeout": "5s",
+  "retryCount":10,
+  "aliveStatusCodes": [0, 200, 302, 400, 403, 429]
```
jlowe, line 4: Error code 400 is a bad request which does not seem appropriate to consider a valid link.  From the logs, what URI is triggering that status code?

Also allowing 429 seems like it could lead to some false-positives, I'd prefer we try the `retryOn429` and `retry-after` settings to work around these error codes, otherwise we could have a bad link that the server kicks back as 429 and we think it's valid.

I'm also curious which URIs are triggering error code 403, as that implies some of our links require authentication.  I assume all of our links reference public resources, but maybe there are some exceptions.

0 is not a valid HTTP status code, so I'm also curious which URIs trigger that response.

nvliyuan: For code 400, it could be found in some changelog.md file like:
image
For code 429, all github issues links will be marked 429 because it send requests too much and fast, see:
image
For code 403, it seems the github action runner(docker) cannot reach maven repository, see:
image
For code 0, it could be found in getting-started-alluxio.md, it is a cluster url, not really need to link, see:
image
all logs can be found [here](https://github.com/nvliyuan/yuali-spark-rapids/actions)

nvliyuan: for status code 429, I also see other open source repo add it as an exception, see
https://github.com/netdata/netdata/blob/master/.mlc_config.json and retryon429 is also a workaround see
https://github.com/hashicorp/packer/blob/main/.mlc_config.json

jlowe: I'm not saying we can just ignore 429, but if `retryon429` (and possibly a `retry-after` setting to go along with it) works, that is far preferable to just taking any 429 and counting it as success.

.github/workflows/markdown-links-check.yml
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2020-2022, NVIDIA CORPORATION.
```
pxLi, line 1: `# Copyright (c) 2022, NVIDIA CORPORATION.`

.github/workflows/markdown-links-check.yml
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# A workflow to check if PR got broken hyperlinks
+name: Check Markdown links
+
+on:
+  pull_request_target:
```
pxLi, line 19: `pull_request_target` -> `pull_request` to checkout the "refs/pull/:prNumber/merge"
otherwise the 
`- uses: actions/checkout@master`
will check out the BASE ref here (branch-22.04)

FYI we use pull_request_target in signoff-check is only because that the check is based on BASE branch instead of relying on developer's feature branch

.github/workflows/markdown-links-check.yml
nvliyuan: thanks, updated

.github/workflows/markdown-links-check.yml
nvliyuan: thanks and good to know that, one question is that do we need a GITHUB_TOKEN when we use `pull_request `?
If we don't add a GITHUB_TOKEN, will we hit `Error: Parameter token or opts.auth is required`?

pxLi: GITHUB_TOKEN is the default token of repo and is maintained by github action itself as ENV.

> If we don't add a GITHUB_TOKEN, will we hit Error: Parameter token or opts.auth is required?

For actions/checkout, no. 
I am not sure where you did get this error, if the some specific action require adding specific arg, then you can try add 
```
      with:
        token: ${{ secrets.GITHUB_TOKEN }}
```
above was just a example, this really depends on the action provider, if they require extra tokens (or access), they should have documented somewhere in their repo

pxLi: image
after changing to pull_request, you can see result in this PR's check status

.github/workflows/markdown-links-check/markdown-links-check-config.json
nvliyuan: Added `retryon429`.

Exclude pattern `https://github.com/NVIDIA/spark-rapids/pull/*` and `https://github.com/NVIDIA/spark-rapids/issues/*` because there are thousands of rapids PRs and issues URL links we don't really need to check, and this will spend a long time to check, and it cannot be avoid by `check-modified-files-only` because these thousands of links are in CHANGELOG file.

Exclude pattern `/docs/archives`, this relative link can only be found in CHANGELOG, and the link is not broken, please see https://github.com/nvliyuan/yuali-spark-rapids/blob/branch-22.04-add-markdown-action/CHANGELOG.md   I don't know why the link checker mark it as broken, maybe a bug?

Remove status code 0(status code 0 is defined by the link checker, if we ignore some links, these links will be mark as status 0)

pxLi (approved):
LGTM. Let's merge this first and let it run for several days before marking it as a required check

pxLi:
build

pxLi:
build

PR #4917, opened 2022-03-08
Default some configs to protect against cluster settings in integration tests

### LINKED ISSUE #4893: integration tests assume incompat config is false, we should verify this
**Describe the bug**
I was seeing test failures in one of our integration test environment where we were running on an actual cluster.  Specifically with test test_hash_groupby_typed_imperative_agg_without_gpu_implementation_fallback. This test assumes that the spark.rapids.sql.incompatibleOps.enabled=false.  In this test environment that flag had been turned to true in the global spark configs which caused the test to fail. In this case it wasn't obvious to me that this test relied on that behavior.

If the tests are assuming this is false we should have a global check for it and fail with useful message if it finds its not set that way.

Note that we have a check https://github.com/NVIDIA/spark-rapids/blob/branch-22.04/integration_tests/src/main/python/asserts.py#L219 that doesn't like setting the conf manually.

jlowe:
Fixes #4893.

Sets up some configs to defaults in the integration tests to protect against cluster settings that could override the defaults and surprise tests that assume the defaults.

### DIFF
--- a/integration_tests/src/main/python/spark_session.py
+++ b/integration_tests/src/main/python/spark_session.py
@@ -33,7 +33,29 @@ _orig_conf_keys = _orig_conf.keys()
 
 # Default settings that should apply to CPU and GPU sessions.
 # These settings can be overridden by specific tests if necessary.
-_default_conf = { 'spark.sql.legacy.allowNegativeScaleOfDecimal': 'true' }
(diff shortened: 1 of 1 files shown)


jlowe:
build

tgravescs (approved):

PR #4857, opened 2022-02-24
Transition to v2 shims [Databricks]

razajafri:
This PR in a nutshell 

* Moved all the SparkShims to a standard package `com.nvidia.spark.rapids.shims.v2` so we don't have to dynamically look them up at runtime
* Combined duplicate classes to reduce code redundancy
* Removed the need to call the `ShimLoader` when accessing the Shims

fixes #4474

PR #4921, opened 2022-03-09
Add DatabricksShimVersion to unshimmed class list [databricks]

### LINKED ISSUE #4918: [BUG] databricks nightly build failed
**Describe the bug**
introduced in https://github.com/NVIDIA/spark-rapids/pull/4857

spark301db
```
scala.MatchError: 3.0.1-databricks (of class com.nvidia.spark.rapids.DatabricksShimVersion)
```

spark312db
```
java.lang.LinkageError: loader constraint violation: when resolving method "com.nvidia.spark.rapids.shims.spark312db.SparkShimServiceProvider$.VERSION()Lcom/nvidia/spark/rapids/DatabricksShimVersion;" the class loader (instance of org/apache/spark/util/MutableURLClassLoader) of the current class, com/nvidia/spark/rapids/shims/spark312db/SparkShimServiceProvider, and the class loader (instance of sun/misc/Launcher$AppClassLoader) for the method's defining class, com/nvidia/spark/rapids/shims/spark312db/SparkShimServiceProvider$, have different Class objects for the type com/nvidia/spark/rapids/DatabricksShimVersion used in the signature
```

razajafri:
~Reverting changes to the DB SparkShims as part of the #4857. This change was originally envisioned to avoid referencing the ServiceProvider, but in case of DB we are not reusing the SparkShimImpl so we can directly reference it.~

Talked to @tgravescs about this and confirmed the correct fix is to "unshim" the `DatabricksShimVersion` that way the class loader will pick up the correct class. 

fixes #4918 

### DIFF
--- a/dist/unshimmed-common-from-spark301.txt
+++ b/dist/unshimmed-common-from-spark301.txt
@@ -19,6 +19,7 @@ com/nvidia/spark/rapids/ShimLoader*
 com/nvidia/spark/rapids/ShimVersion*
 com/nvidia/spark/rapids/SparkShimServiceProvider*
 com/nvidia/spark/rapids/SparkShimVersion*
+com/nvidia/spark/rapids/DatabricksShimVersion*
 com/nvidia/spark/rapids/SparkShims*
 com/nvidia/spark/udf/Plugin*
 libjucx.so
--- a/dist/unshimmed-common-from-spark301.txt
+++ b/dist/unshimmed-common-from-spark301.txt
@@ -6,6 +6,7 @@ com/nvidia/spark/ExclusiveModeGpuDiscoveryPlugin*
 com/nvidia/spark/RapidsUDF*
 com/nvidia/spark/SQLPlugin*
 com/nvidia/spark/rapids/ColumnarRdd*
+com/nvidia/spark/rapids/DatabricksShimVersion*
 com/nvidia/spark/rapids/ExecutionPlanCaptureCallback*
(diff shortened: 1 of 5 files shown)


razajafri:
@pxLi PTAL

razajafri:
build

razajafri:
build

tgravescs:
I'm confused by the description here. I don't think this is the right fix. it appears the issue is with classloader and having 2 versions of the DatabricksSparkVersion, I think it needs to be added to the classes we don't shim.   We have com/nvidia/spark/rapids/SparkShimVersion and ShimVersion in there already.

razajafri:
> 

You are right, I wanted to put up this patch to unblock the CI last night. I will put up another PR

razajafri:
build

dist/unshimmed-common-from-spark301.txt
```diff
@@ -19,6 +19,7 @@ com/nvidia/spark/rapids/ShimLoader*
 com/nvidia/spark/rapids/ShimVersion*
 com/nvidia/spark/rapids/SparkShimServiceProvider*
 com/nvidia/spark/rapids/SparkShimVersion*
+com/nvidia/spark/rapids/DatabricksShimVersion*
```
jlowe, line 22: Nit: This file is lexicographically ordered, would be good to keep that precedent.

razajafri:
build

jlowe (approved):

tgravescs (approved):

Issue #4148, opened 2021-11-18
[Audit][FEA][SPARK-36854] Add ANSI interval types to PCBS 


gerashegalov:
ParquetCachedBatchSerializer needs to support ANSI interval types

PR #4926, opened 2022-03-10
Support `DayTimeIntervalType` in `ParquetCachedBatchSerializer`[databricks]

firestarman:
This PR is to add the new `DayTimeIntervalType` support in `ParquetCachedBatchSerializer`  for Spark v3.3.0+, along with some tests.

closes https://github.com/NVIDIA/spark-rapids/issues/4148
closes https://github.com/NVIDIA/spark-rapids/issues/4931

### DIFF
--- a/integration_tests/src/main/python/cache_test.py
+++ b/integration_tests/src/main/python/cache_test.py
@@ -17,7 +17,7 @@ import pytest
 from asserts import assert_gpu_and_cpu_are_equal_collect, assert_equal, assert_collection_equal_ignore_order
 from data_gen import *
 import pyspark.sql.functions as f
-from spark_session import with_cpu_session, with_gpu_session, is_before_spark_330, is_with_rapids_cache_serializer
+from spark_session import with_cpu_session, with_gpu_session, is_before_spark_330
 from join_test import create_df
 from marks import incompat, allow_non_gpu, ignore_order
 
@@ -287,11 +287,7 @@ def test_cache_map_and_array(data_gen, enable_vectorized):
     assert_gpu_and_cpu_are_equal_collect(helper)
 
 
-# No need to run these tests without rapids cache serializer since the 'InMemoryTableScanExec'
-# node will fall back to CPU for such case. Then there is no GPU things joined in for the
-# cache/uncache operations.
-@pytest.mark.skipif(is_before_spark_330() or not is_with_rapids_cache_serializer(),
-    reason='DayTimeInterval is not supported before Spark3.3.0 or rapids cache serializer is not used')
+@pytest.mark.skipif(is_before_spark_330(), reason='DayTimeInterval is not supported before Spark3.3.0')
 @ignore_order(local=True)
 def test_cache_daytimeinterval_input_row():
     assert_gpu_and_cpu_are_equal_collect(
@@ -299,8 +295,7 @@ def test_cache_daytimeinterval_input_row():
             DayTimeIntervalGen(), int_gen, null_gen).cache().selectExpr('b', 'a'))
 
 
-@pytest.mark.skipif(is_before_spark_330() or not is_with_rapids_cache_serializer(),
-    reason='DayTimeInterval is not supported before Spark3.3.0 or rapids cache serializer is not used')
+@pytest.mark.skipif(is_before_spark_330(), reason='DayTimeInterval is not supported before Spark3.3.0')
 @allow_non_gpu("FileSourceScanExec", "ColumnarToRowExec")
 @pytest.mark.parametrize('alongside_gen', [int_gen, ArrayGen(int_gen)], ids=idfn)
 @pytest.mark.parametrize('with_rapids_memoryscan', ['true', 'false'],
--- a/integration_tests/src/main/python/cache_test.py
+++ b/integration_tests/src/main/python/cache_test.py
@@ -288,38 +288,12 @@ def test_cache_map_and_array(data_gen, enable_vectorized):
 
 
 @pytest.mark.skipif(is_before_spark_330(), reason='DayTimeInterval is not supported before Spark3.3.0')
+@pytest.mark.parametrize('enable_vectorized_conf', enable_vectorized_confs, ids=idfn)
 @ignore_order(local=True)
-def test_cache_daytimeinterval_input_row():
-    assert_gpu_and_cpu_are_equal_collect(
-        lambda spark: three_col_df(spark,
-            DayTimeIntervalGen(), int_gen, null_gen).cache().selectExpr('b', 'a'))
-
-
-@pytest.mark.skipif(is_before_spark_330(), reason='DayTimeInterval is not supported before Spark3.3.0')
-@ignore_order(local=True)
-@allow_non_gpu("FileSourceScanExec", "ColumnarToRowExec")
-@pytest.mark.parametrize('alongside_gen', [int_gen, ArrayGen(int_gen)], ids=idfn)
-@pytest.mark.parametrize('with_rapids_memoryscan', ['true', 'false'],
-                         ids=["rapids_memoryscan_on", "rapids_memoryscan_off"])
-@pytest.mark.parametrize('with_rapids_reader', ['true', 'false'],
-                         ids=["rapids_reader_on", "rapids_reader_off"])
-def test_cache_daytimeinterval_input_columnar(spark_tmp_path, alongside_gen,
-                                              with_rapids_memoryscan, with_rapids_reader):
-    tmp_data_path = spark_tmp_path + '/PARQUET_DATA'
+def test_cache_daytimeinterval(enable_vectorized_conf):
     def test_func(spark):
-        two_col_df(spark,
-            DayTimeIntervalGen(), alongside_gen).write.mode("overwrite").parquet(tmp_data_path)
-        df = spark.read.parquet(tmp_data_path)
+        df = two_col_df(spark, DayTimeIntervalGen(), int_gen)
         df.cache().count()
-        return df.selectExpr('b', 'a')
+        return df.selectExpr("b", "a")
+    assert_gpu_and_cpu_are_equal_collect(test_func, enable_vectorized_conf)
 
-    test_conf = {
-        # rapids-spark doesn't support LEGACY read for parquet, also set the int96 rebase mode
-        # values because LEGACY in databricks which will preclude this op from running on GPU.
-        'spark.sql.parquet.datetimeRebaseModeInWrite': 'CORRECTED',
-        'spark.sql.parquet.datetimeRebaseModeInRead' : 'CORRECTED',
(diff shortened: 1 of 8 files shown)


firestarman:
build

integration_tests/src/main/python/cache_test.py
```diff
@@ -285,3 +285,67 @@ def helper(spark):
         return df.selectExpr("a")
 
     assert_gpu_and_cpu_are_equal_collect(helper)
+
+
+# No need to run these tests without rapids cache serializer since the 'InMemoryTableScanExec'
+# node will fall back to CPU for such case. Then there is no GPU things joined in for the
+# cache/uncache operations.
+@pytest.mark.skipif(is_before_spark_330() or not is_with_rapids_cache_serializer(),
```
razajafri, line 293: Why the extra check "is_with_rapids_cache_serializer()"? Does the test fail with the default serializer?

integration_tests/src/main/python/cache_test.py
```diff
@@ -285,3 +285,67 @@ def helper(spark):
...
+    reason='DayTimeInterval is not supported before Spark3.3.0 or rapids cache serializer is not used')
+@ignore_order(local=True)
+def test_cache_daytimeinterval_input_row():
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: three_col_df(spark,
+            DayTimeIntervalGen(), int_gen, null_gen).cache().selectExpr('b', 'a'))
+
+
+@pytest.mark.skipif(is_before_spark_330() or not is_with_rapids_cache_serializer(),
```
razajafri, line 302: Same comment as above

integration_tests/src/main/python/cache_test.py
firestarman: Thanks for review.
No, it passes. As said in the comment, with the default serializer, `InMemoryTableScanExec` will run on CPU, then no GPU things here for cache. So do we still need to test ?

integration_tests/src/main/python/cache_test.py
firestarman: Same as the above

integration_tests/src/main/python/cache_test.py
razajafri: So, the original intent for cache_test.py was not to test PCBS or GpuInMemoryTableScanExec. It's testing if output from GPU exec's can successfully be cached.

firestarman: Good to know, thanks.

firestarman: I removed the serializer check

sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/ParquetCachedBatchSerializer.scala
```diff
@@ -334,10 +335,14 @@ protected class ParquetCachedBatchSerializer extends GpuCachedBatchSerializer wi
         isSchemaSupportedByCudf(schema)) {
       def putOnGpuIfNeeded(batch: ColumnarBatch): ColumnarBatch = {
         if (!batch.column(0).isInstanceOf[GpuColumnVector]) {
-          val s: StructType = structSchema
-          val gpuCB = new GpuColumnarBatchBuilder(s, batch.numRows()).build(batch.numRows())
-          batch.close()
-          gpuCB
+          // The input batch from CPU must NOT be closed, because the columns inside it
+          // will be reused, and Spark expects the producer to close its batches.
+          val rowNum = batch.numRows()
```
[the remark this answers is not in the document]
firestarman: updated

firestarman:
build

integration_tests/src/main/python/cache_test.py
firestarman: removed the serializer check

integration_tests/src/main/python/cache_test.py
```diff
@@ -285,3 +285,62 @@ def helper(spark):
...
+
+
+@pytest.mark.skipif(is_before_spark_330(), reason='DayTimeInterval is not supported before Spark3.3.0')
+@allow_non_gpu("FileSourceScanExec", "ColumnarToRowExec")
+@pytest.mark.parametrize('alongside_gen', [int_gen, ArrayGen(int_gen)], ids=idfn)
+@pytest.mark.parametrize('with_rapids_memoryscan', ['true', 'false'],
+                         ids=["rapids_memoryscan_on", "rapids_memoryscan_off"])
+@pytest.mark.parametrize('with_rapids_reader', ['true', 'false'],
+                         ids=["rapids_reader_on", "rapids_reader_off"])
+def test_cache_daytimeinterval_input_columnar(spark_tmp_path, alongside_gen,
```
razajafri, line 305: Why is this test method so complicated? What are we doing in this test that we aren't doing in any other test? We have a DF that we cache and then pull it from the cache, compare the result from the CPU and GPU

In other words, why don't we just do the following? 

```
def test_cache_daytimeinterval_input_columnar():
    def func(spark):
        df = two_col_df(spark, DayTimeIntervalGen(), alongside_gen)
        df.cache().count()
        return df.selectExpr("a") // or whatever the column name is

    assert_gpu_and_cpu_are_equal_collect(func, conf={YOUR CONF})

```

I could be missing something

firestarman: I thought since we implement all the paths (both CPU and GPU) in the PCBS, I think the output for all of the paths should be equal to the original data.
Comparing the GPU output to CPU output is based on the CPU output of the PCBS is reliable and equal to the Spark output.
Maybe it is over designed.

firestarman: Reading data from parquet is to get a columnar input to be cached. 
I do it in the `test_cache_daytimeinterval_input_row` test, it only checks the path of converting internal rows to cached batches.

firestarman: Simplified this  test

firestarman:
build

sql-plugin/src/main/311+-all/scala/com/nvidia/spark/rapids/shims/ParquetCachedBatchSerializer.scala
```diff
@@ -334,10 +335,14 @@ protected class ParquetCachedBatchSerializer extends GpuCachedBatchSerializer wi
...
         if (!batch.column(0).isInstanceOf[GpuColumnVector]) {
-          val s: StructType = structSchema
-          val gpuCB = new GpuColumnarBatchBuilder(s, batch.numRows()).build(batch.numRows())
-          batch.close()
-          gpuCB
+          // The input batch from CPU must NOT be closed, because the columns inside it
+          // will be reused, and Spark expects the producer to close its batches.
+          val numRows = batch.numRows()
+          val gcbBuilder = new GpuColumnarBatchBuilder(structSchema, numRows)
+          for (i <- 0 until batch.numCols()) {
```
res-life, line 342: Can use the following to improve the performance:
```
var rowIndex = 0
while (rowIndex < batch.numRows()) {

    ......
    rowIndex += 1
}
```

A similar PR: https://github.com/NVIDIA/spark-rapids/pull/4770

firestarman: Number of rows is always quite large, so this change can improve some performance. However here is for columns, changing to this will get little benfit for performance.

firestarman:
@razajafri  Could you review this again ?

integration_tests/src/main/python/cache_test.py
razajafri: Reading/writing to parquet isn't necessary to write/read columnar cache. 

If you look at `InMemoryTableScanExec` it calls the `convertColumnarBatchToCachedBatch` if the spark plan supports columnar input and if the serializer supports columnar input (which it always does [here](https://github.com/NVIDIA/spark-rapids/blob/611f54dce73a8e42c575e7ae2f95b1d729085179/sql-plugin/src/main/311%2B-all/scala/com/nvidia/spark/rapids/shims/ParquetCachedBatchSerializer.scala#L266)) 

Reading cache columnar in PCBS depends on three variables, whether the conf `spark.sql.inMemoryColumnarStorage.enableVectorizedReader` is enabled and the plan has 100 or less columns and the plan's output is AtomicType or NullType

firestarman: Good to know.
Just got home and updated the test. Now it does almost the same with `test_cache_columnar`, except the additional data type `DayTimeIntervalType`.

firestarman:
build

firestarman:
build

firestarman:
build

integration_tests/src/main/python/cache_test.py
```diff
@@ -285,3 +285,15 @@ def helper(spark):
         return df.selectExpr("a")
 
     assert_gpu_and_cpu_are_equal_collect(helper)
+
```
razajafri, line 288: extra space

integration_tests/src/main/python/cache_test.py
```diff
@@ -285,3 +285,15 @@ def helper(spark):
...
+
+@pytest.mark.skipif(is_before_spark_330(), reason='DayTimeInterval is not supported before Spark3.3.0')
+@pytest.mark.parametrize('enable_vectorized_conf', enable_vectorized_confs, ids=idfn)
+@ignore_order(local=True)
+def test_cache_daytimeinterval(enable_vectorized_conf):
```
razajafri, line 293: Now, it begs a question, do we need a separate test or can we add an additional type to one (all?) of the existing test? 

You can use a pytest.param(..., marks=[skipif]) and add it to an existing test. The case may be that you need an additional test, I am not sure

firestarman: I tried this `pytest.param(..., marks=[skipif])` but got the below error on Spark before 3.3.0.

```

==================================== ERRORS ====================================
________________ ERROR collecting src/main/python/cache_test.py ________________
../../src/main/python/cache_test.py:60: in <module>
    daytime_interval_gen = pytest.param(DayTimeIntervalGen(),
../../src/main/python/data_gen.py:626: in __init__
    super().__init__(DayTimeIntervalType(), nullable=nullable, special_cases=special_cases)
E   NameError: name 'DayTimeIntervalType' is not defined
- generated xml file: /home/liangcail/work/projects/on_github/spark-rapids/integration_tests/target/run_dir/TEST-pytest-1647331437827840861.xml -
ERROR ../../src/main/python/cache_test.py - NameError: name 'DayTimeIntervalT...
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
====================== 11715 deselected, 1 error in 2.51s ======================

```

razajafri (approved):

Issue #4146, opened 2021-11-18
[Audit][FEA][SPARK-36831] Support ANSI interval types from CSV source


gerashegalov:
Make sure the plugin can read ANSI interval types from CSV source

res-life:
@revans2  Please help review the solution:


Interval types can be found in https://spark.apache.org/docs/latest/sql-ref-datatypes.html

Currently plugin do not support write for csv, so let's talk about reading interval type from CSV.

Spark read interval code is in:
[IntervalUtils.fromDayTimeString](https://github.com/apache/spark/blob/v3.2.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/IntervalUtils.scala#L430)

There are legacy form and normal form switched by SQLConf.LEGACY_FROM_DAYTIME_STRING.

**legacy form**
By default, legacy from daytime string is disable, so we may not support this.

SQLConf.LEGACY_FROM_DAYTIME_STRING See: https://github.com/apache/spark/blob/v3.2.1/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L3042

parseDayTimeLegacy see:
https://github.com/apache/spark/blob/v3.2.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/IntervalUtils.scala#L463toL475

**normal form**
By default, Spark use this form,  some examples see below table, we will support this form only.

|type|example|valid|comment|
|:-|:-|:-|:-|
|INTERVAL DAY | INTERVAL '100' DAY  |true    |    |
|INTERVAL DAY | 100 DAY  |true    |    |
|INTERVAL DAY TO HOUR | INTERVAL '100 10' DAY TO HOUR  |true    |    |
|INTERVAL DAY TO HOUR | 100 10   |true    |    |
|INTERVAL DAY TO SECOND | INTERVAL '100 10:30:40.999999' DAY TO SECOND  |true    |    |
|INTERVAL DAY TO SECOND | 100 10:30:40.999999  |true    |    |
|INTERVAL DAY TO SECOND | INTERVAL '100 10:30:40. &nbsp;  &nbsp;999999' DAY TO SECOND  |false    | has extra spaces   |
|INTERVAL DAY TO SECOND | INTERVAL '100  &nbsp;  &nbsp; 10:30:40.999999' DAY TO SECOND  |false    |  has extra spaces     |
[...]

Issue #1948, opened 2021-03-16
[FEA]Audit - [SPARK-34605][SQL] Support `java.time.Duration` as an external type of the day-time interval type


tgravescs:
**Is your feature request related to a problem? Please describe.**
spark made a change to support java.time.Duration for SPARK-27790 Support ANSI SQL INTERVAL types

https://github.com/apache/spark/commit/17601e0

It looks like we copied the getConverterForType function in GpuRowToColumnarExec so we should look closer at pulling that in.

https://issues.apache.org/jira/browse/SPARK-34605

sameerz:
@tgravescs will review with @revans2

tgravescs:
getConverterFortype is used in GpuRowToColumnarExec:
https://github.com/NVIDIA/spark-rapids/blob/branch-0.5/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala#L94

I don't know if there is already another lira to add support for DayTimeIntervalType. I assume we need to add it to our isSupportedType checks and so forth as well.   @revans2  do you happen to know?

sameerz:
P1 for whether we support time.Duration for the cached batch serializer.  P3 for whether we are going to support time.Duration.

revans2:
There are a number of new types that have been added into spark and if we run into any of them we fall back to the CPU for it. Eventually we want to add in more types, but that should not block any of the existing code/queries from working.

res-life:
Duration is an external type, Plugin do not need to handle it.
The Duration is used when creating a `df`, see below:
```
schema = StructType([
    StructField("c1", DayTimeIntervalType(
                DayTimeIntervalType.DAY, DayTimeIntervalType.DAY
            ), False),
])
data = [(timedelta( days=50, seconds=27), ),]
df = spark.createDataFrame(
        SparkContext.getOrCreate().parallelize(data, numSlices=2),
        schema)

```


day-time interval type is stored as int64 internally in Spark,
and Spark also stores it as int64 to parquet file along with an metada which indicating the mapping between `day-time interval` and `int64`

This issue supports ANSI intervals to/from Parquet:
https://github.com/NVIDIA/spark-rapids/issues/4145

Can we close it?  @revans2

revans2:
I think we can close this because we can do columnar to row and row to columnar on these types.

PR #4927, opened 2022-03-10
Support reading ANSI day time interval type from CSV source

res-life:
Contributes #4146 

Support reading ANSI day time interval type from CSV source.
This PR is an implementation on the plugin side currently.
If cuDF supports this in future, should update, see https://github.com/rapidsai/cudf/issues/10356


res-life:
build

res-life:
build

res-life:
build

res-life:
build

res-life:
Filed 2 Spark issues:
Overflow occurs when reading ANSI day time interval from CSV file
https://issues.apache.org/jira/browse/SPARK-38520

The second range is not [0, 59] in the day time ANSI interval
https://issues.apache.org/jira/browse/SPARK-38324

integration_tests/src/main/python/data_gen.py
```diff
@@ -613,31 +613,112 @@ def make_null():
...
-#   DAY, days in the range [0..106751991].
+#   If hour is start field, it's max value is long.max / microseconds in one hour
+#   DAY, days in the range [0..106751991]. 106751991 is long.max / microseconds in one day
 # For more details: https://spark.apache.org/docs/latest/sql-ref-datatypes.html
-# Note: 106751991/365 = 292471 years which is much bigger than 9999 year, seems something is wrong
+# Note: 106751991/365 = 292471 years which is much bigger than 9999 year
 class DayTimeIntervalGen(DataGen):
     """Generate DayTimeIntervalType values"""
-    def __init__(self, max_days = None, nullable=True, special_cases =[timedelta(seconds = 0)]):
+    def __init__(self, max_days=None, start_field="day", end_field="second", allow_negative=False, nullable=True,
```
firestarman, line 630: Better to cover the cases as more as possible, so I prefer to 'true' being the default value of `allow_negative`  here.

integration_tests/src/main/python/data_gen.py
```diff
@@ -613,31 +613,112 @@ def make_null():
...
+                 special_cases=[timedelta(seconds=0)]):
```
firestarman, line 631: How about adding None to the default special cases list ?

sql-plugin/src/main/301until330-all/scala/com/nvidia/spark/rapids/shims/GpuTypeShims.scala
```diff
@@ -46,4 +47,9 @@ object GpuTypeShims {
    * @return the cuDF type if the Shim supports
    */
   def toRapidsOrNull(t: DataType): DType = null
+
+  def supportCsvRead(dt: DataType) : Boolean = false
+
+  def csvRead(cv: ColumnVector, dt: DataType): ColumnVector =
```
firestarman, line 53:
```suggestion
  def csvRead(cv: cudf.ColumnVector, dt: DataType): cudf.ColumnVector =
```
Otherwise, it will conflict with the PR https://github.com/NVIDIA/spark-rapids/pull/4926 who imports the `org.apache.spark.sql.vectorized.ColumnVector`

sql-plugin/src/main/301until330-all/scala/com/nvidia/spark/rapids/shims/GpuTypeShims.scala
```diff
@@ -15,6 +15,7 @@
  */
 package com.nvidia.spark.rapids.shims
 
+import ai.rapids.cudf.ColumnVector
```
firestarman, line 18:
```suggestion
import ai.rapids.cudf
```

integration_tests/src/main/python/data_gen.py
res-life: nullable = true will handle the None case, is it right? see: data.gen.py 57 line:
```
        if self.nullable:
            self.with_special_case(None, weight)
```

res-life:
build

res-life:
**Performance test result**
|GPU|CPU|radio|
|-|-|-|
|5,057ms|30,569ms|6|

Test file is 1.7G and has 12320000 rows and 10 columns, it's duplicated from day-time-interval.csv by many times. The day-time-interval.csv is in the tests module.

The sql is:
spark.sql("select count(c01)  + count(c02) + count(c03) + count(c04)  + count(c05) + count(c06) + count(c07) + count(c08)  + count(c09) + count(c10) from tbl").show()

Details code:
```
import org.apache.spark.sql.{DataFrame, SparkSession}
import org.apache.spark.sql.types.{DayTimeIntervalType, StructField, StructType}

def dayTime(s: Byte, e: Byte): DayTimeIntervalType = DayTimeIntervalType(s, e)

val schema = StructType(Seq(
  StructField("c01", dayTime(DayTimeIntervalType.DAY, DayTimeIntervalType.DAY)),
  StructField("c02", dayTime(DayTimeIntervalType.DAY, DayTimeIntervalType.HOUR)),
  StructField("c03", dayTime(DayTimeIntervalType.DAY, DayTimeIntervalType.MINUTE)),
  StructField("c04", dayTime(DayTimeIntervalType.DAY, DayTimeIntervalType.SECOND)),
  StructField("c05", dayTime(DayTimeIntervalType.HOUR, DayTimeIntervalType.HOUR)),
  StructField("c06", dayTime(DayTimeIntervalType.HOUR, DayTimeIntervalType.MINUTE)),
  StructField("c07", dayTime(DayTimeIntervalType.HOUR, DayTimeIntervalType.SECOND)),
  StructField("c08", dayTime(DayTimeIntervalType.MINUTE, DayTimeIntervalType.MINUTE)),
  StructField("c09", dayTime(DayTimeIntervalType.MINUTE, DayTimeIntervalType.SECOND)),
  StructField("c10", dayTime(DayTimeIntervalType.SECOND, DayTimeIntervalType.SECOND))
))

val path = "/tmp/my-tmp/large.csv"

spark.conf.set("spark.rapids.sql.enabled", false)
var start = System.currentTimeMillis()
spark.read.schema(schema).csv(path).createOrReplaceTempView("tbl")
spark.sql("select count(c01)  + count(c02) + count(c03) + count(c04)  + count(c05) + count(c06) + count(c07) + count(c08)  + count(c09) + count(c10) from tbl").show()
println((System.currentTimeMillis() - start) + "ms")


spark.conf.set("spark.rapids.sql.enabled", true)
start = System.currentTimeMillis()
spark.read.schema(schema).csv(path).createOrReplaceTempView("tbl")
spark.sql("select count(c01)  + count(c02) + count(c03) + count(c04)  + count(c05) + count(c06) + count(c07) + count(c08)  + count(c09) + count(c10) from tbl").show()
println((System.currentTimeMillis() - start) + "ms")

```

res-life:
build

integration_tests/src/main/python/data_gen.py
```diff
@@ -613,31 +613,112 @@ def make_null():
...
+        max_micros = self._max_days * micros_per_day
+
+        # set default value
+        days = 0
+        hours = 0
+        minutes = 0
+        seconds = 0
+        microseconds = 0
+
+        if (start_field, end_field) == ("day", "day"):
```
firestarman, line 665: I think it would be better to use the below structure, which should get a better perf.
```
if:
    ...
elif:
    ...
elif:
    ...
else:
    ...
```

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/GpuIntervalUtils.scala
```diff
@@ -0,0 +1,556 @@
...
+   * Return sign * base * multiple.
+   *
+   * @param sign         long column
+   * @param groupInTable string column contains positive long
+   * @param multiple     const long value
+   * @return sign * base * multiple
+   */
+  private def multiple(
+      sign: ColumnVector, groupInTable: ColumnVector, multiple: Long): ColumnVector = {
+    withResource(groupInTable.castTo(DType.INT64)) { baseLong =>
```
firestarman, line 548: `baseLong` can be closed right away after getting `baseWithSign`.
One more thing, could we only compute the `baseLong.mul(sign)` when the `sign` is `-1`. For sign being `+1`, we could just increase the reference count of `baseLong` by 1.

tests/src/test/resources/day-time-interval-to-be-fix.csv
```diff
@@ -0,0 +1,2 @@
+interval '106751992' day,,,,interval '2562047789' hour,,,interval '153722867281' minute,,
```
firestarman, line 1: May I know what this `to be fix` mean ?

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/GpuIntervalUtils.scala
```diff
@@ -0,0 +1,556 @@
...
+   * @return
+   */
+  private def multiple(
+      sign: ColumnVector, base: ColumnVector, multiple: Long, maxInBase: Long): ColumnVector = {
+    withResource(base.castTo(DType.INT64)) { baseLong =>
+      withResource(Scalar.fromLong(maxInBase)) { maxScalar =>
+        withResource(baseLong.greaterThan(maxScalar)) { greater =>
+          withResource(Scalar.fromNull(DType.INT64)) { nullScalar =>
+            withResource(greater.ifElse(nullScalar, baseLong)) { baseWithFix =>
+              withResource(baseWithFix.mul(sign)) { baseWithSign =>
```
firestarman, line 526: Better to try to close the columns once they are no longer needed, to reduce the memory peak. e.g. `baseWithFix` can be closed just after getting `baseWithSign`.
```
val baseWithSign = withReource(greater.ifElse(nullScalar, baseLong)) { baseWithFix =>
  baseWithFix.mul(sign)
}
withResource(baseWithSign) { _ =>
   withResource(Scalar.fromLong(multiple)) { multipleScalar =>
     baseWithSign.mul(multipleScalar)
   }
}
```

res-life:
@revans2  help to review

integration_tests/src/main/python/data_gen.py
```diff
@@ -613,31 +613,112 @@ def make_null():
...
+        fields = ["day", "hour", "minute", "second"]
+        start_index = fields.index(start_field)
+        end_index = fields.index(end_field)
+        if start_index > end_index:
+            raise RuntimeError('Start field {}, end field {}, valid fields is {}, start field should <= end field'.format(start_field, end_field, fields))
+
+        super().__init__(DayTimeIntervalType(start_index, end_index), nullable=nullable, special_cases=special_cases)
+
+    def _gen_random(self, rand, start_field, end_field):
+        micros_per_second = 1000 * 1000
```
revans2, line 651: nit make these members, or better hidden constants in the file so we don't have to re-calculate them each time.

integration_tests/src/main/python/data_gen.py
```diff
@@ -613,31 +613,112 @@ def make_null():
...
-#   DAY, days in the range [0..106751991].
+#   If hour is start field, it's max value is long.max / microseconds in one hour
+#   DAY, days in the range [0..106751991]. 106751991 is long.max / microseconds in one day
 # For more details: https://spark.apache.org/docs/latest/sql-ref-datatypes.html
-# Note: 106751991/365 = 292471 years which is much bigger than 9999 year, seems something is wrong
+# Note: 106751991/365 = 292471 years which is much bigger than 9999 year
 class DayTimeIntervalGen(DataGen):
     """Generate DayTimeIntervalType values"""
-    def __init__(self, max_days = None, nullable=True, special_cases =[timedelta(seconds = 0)]):
+    def __init__(self, max_days=None, start_field="day", end_field="second", allow_negative=True, nullable=True,
```
revans2, line 630: This is a little confusing to me. max_days really means the total number of days that can be selected. So if I want a delta that just has seconds in it (i.e. 0-59 seconds as the case may be) then I can only limit it by days to 1 day? This is the typical use case that I would expect and I think it is what SQL expects, but here it appears to be ignored.

integration_tests/src/main/python/data_gen.py
revans2: Can we do this totally differently? There is just so much copy/paste between each part of the code. We know that the python code really is only converting it to micro seconds. So can we just generate a random number for the microseconds, up to `max_micros`, and then truncate it to the proper time? Also can we convert the "string" start/end fields into something simpler to write code for?

```
# in __init__
DAY = 0
HOUR = 1
MIN = 2
SEC = 3
fields_to_look_at = ["day", "hour", "minute", "second"]
si = fields_to_look_at.index(start_field)
ei = fileds_to_look_at.index(end_field)
assert si <= ei

self.hasDays = si <= DAY and ei >= DAY
self.hasHours = si <= HOUR and ei >= HOUR
self.hasMin = si <= MIN and ei >= MIN
self.hasSec = si <= SEC and ei >= SEC
```

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/GpuIntervalUtils.scala
```diff
@@ -0,0 +1,556 @@
...
+ * Parse DayTimeIntervalType string column to long column of micro seconds
+ * Spark DayTimeIntervalType type:
+ * https://spark.apache.org/docs/latest/sql-ref-datatypes.html
+ * Spark parse DayTime:
+ * https://github.com/apache/spark/blob/v3.2.1/
+ * sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/IntervalUtils.scala#L275
+ *
+ * DayTimeIntervalType have 10 sub types, 10 types and examples are as following:
+ *
+ * INTERVAL DAY: INTERVAL '100' DAY
```
revans2, line 35: What happens if Spark sees `'100' DAY` on a type it expects to be INTERVAL DAY TO HOUR?

integration_tests/src/main/python/data_gen.py
res-life: done

sql-plugin/src/main/301until330-all/scala/com/nvidia/spark/rapids/shims/GpuTypeShims.scala
res-life: done

tests/src/test/resources/day-time-interval-to-be-fix.csv
res-life: Refer to the last test case in CsvScanForIntervalSuite.
A Spark issue blocked this test case: https://issues.apache.org/jira/browse/SPARK-38520

integration_tests/src/main/python/csv_test.py
```diff
@@ -266,7 +266,29 @@ def test_basic_csv_read(std_input_path, name, schema, options, read_func, v1_ena
         pytest.param(double_gen),
         pytest.param(FloatGen(no_nans=False)),
         pytest.param(float_gen),
-        TimestampGen()]
+        TimestampGen(),
+        # 365 days * 5000 is about 5000 years
+        DayTimeIntervalGen(start_field="day", end_field="day", max_days=365 * 5000, allow_negative=True),
```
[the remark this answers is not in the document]
res-life: done

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/GpuIntervalUtils.scala
res-life: `interval '100' DAY` does not match the regexp of `INTERVAL DAY TO HOUR` type and Spark return `NULL` when reading the CSV file. There are 10 regexps for each sub `DayTimeIntervalType` type.  Should explicitly specify the sub types in the schema when reading CSV files.

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/GpuIntervalUtils.scala
res-life: >baseLong can be closed right away after getting baseWithSign

Yes.

> For sign being +1, we could just increase the reference count of baseLong by 1. Doing this may get a better perf for positive sign.

The `sign` is a column with -1 or 1 in it, it's not a scalar.

integration_tests/src/main/python/data_gen.py
```diff
@@ -613,31 +613,112 @@ def make_null():
...
+#   If second is start field, it's max value is long.max / microseconds in one second
 #   MINUTE, minutes within hours [0..59],
+#   If minute is start field, it's max value is long.max / microseconds in one minute
 #   HOUR, hours within days [0..23],
-#   DAY, days in the range [0..106751991].
+#   If hour is start field, it's max value is long.max / microseconds in one hour
+#   DAY, days in the range [0..106751991]. 106751991 is long.max / microseconds in one day
 # For more details: https://spark.apache.org/docs/latest/sql-ref-datatypes.html
-# Note: 106751991/365 = 292471 years which is much bigger than 9999 year, seems something is wrong
+# Note: 106751991/365 = 292471 years which is much bigger than 9999 year
```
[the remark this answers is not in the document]
res-life: I originally think the max interval day should be 9999.
I'll delete this line

integration_tests/src/main/python/data_gen.py
res-life: Yes, this argument is some what confusing, but if adding aratuments for each field there will be too many.

integration_tests/src/main/python/data_gen.py
```diff
@@ -613,31 +613,112 @@ def make_null():
         self._start(rand, make_null)
 
 # DayTimeIntervalGen is for Spark 3.3.0+
-# DayTimeIntervalType(startField, endField): Represents a day-time interval which is made up of a contiguous subset of the following fields:
+# DayTimeIntervalType(startField, endField):
+# Represents a day-time interval which is made up of a contiguous subset of the following fields:
 #   SECOND, seconds within minutes and possibly fractions of a second [0..59.999999],
+#   Note Spark now uses 99 as max second, see issue https://issues.apache.org/jira/browse/SPARK-38324
```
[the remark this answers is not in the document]
res-life: The finer granularity you mentioned will be trancated,  code is: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/IntervalUtils.scala#L1037toL1056

see below:
```
from datetime import date, datetime, timedelta, timezone
schema = StructType([
    StructField("c1", DayTimeIntervalType(
                DayTimeIntervalType.DAY, DayTimeIntervalType.DAY
            ), False),
])
data = [(timedelta( days=50, seconds=27), ),]
df = spark.createDataFrame(
        SparkContext.getOrCreate().parallelize(data, numSlices=2),
        schema)
df.show()
+-----------------+
|               c1|
+-----------------+
|INTERVAL '50' DAY|
+-----------------+

```

integration_tests/src/main/python/data_gen.py
revans2: I agree that all of the fields would be too much, but if you passed in a timedelta for the max/min, then you get all of the fields in a simple to use package.

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/GpuIntervalUtils.scala
res-life: Finally, `DayTimeIntervalGen` is simplified in the last code commit.

res-life:
build

integration_tests/src/main/python/data_gen.py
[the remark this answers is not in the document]
res-life: - Yes, casting to string truncates by looking at start and end fields, calling the following code
```
  def toDayTimeIntervalString(
      micros: Long,
      style: IntervalStyle,
      startField: Byte,
      endField: Byte): String = {
```

- After debug, I found the internal long value is not truncated when creating `df`, it stores the micros long value.
The `endField` is always `SECOND` in `durationToMicros`, see below:
```
  def durationToMicros(duration: Duration): Long = {
    durationToMicros(duration, DT.SECOND)
  }

  def durationToMicros(duration: Duration, endField: Byte)
```
- In summary:
The external type of `DayTimeInterval` is Duration, it's converted to micros long value regardless of what start and end fields are.
When casting to string, Spark looks at start and end fields to tuncate the micros.
The pyspark `DayTimeIntervalGen` can just pass a random micros.

- The below code will call the following:
`durationToMicros(duration, DT.SECOND)` create `df`,  the interal long is: **micros of 1 day + micros of 1 second**
Cast to string, calls `toDayTimeIntervalString`,  **truncating the 1 second**
```
  test("my test") {
    val data = Seq(Row(Duration.ofDays(1).plusSeconds(1)))
    val schema = StructType(Array(
      StructField("t", DayTimeIntervalType(DayTimeIntervalType.DAY, DayTimeIntervalType.DAY))
    ))
    val df = spark.createDataFrame(spark.sparkContext.parallelize(data), schema)
    df.show()
  }
```

We can simplify the `DayTimeIntervalGen` by only generate a random micros regardless of (start, end) field, see the current implimetation of `DayTimeIntervalGen`

BTW Because of Duration is external type,  the plugin do not need to handle this type, so this issue  https://github.com/NVIDIA/spark-rapids/issues/1948 can be closed.

res-life:
The last comment: https://github.com/NVIDIA/spark-rapids/pull/4927#discussion_r827648974

res-life:
Will resolve the conflict after https://github.com/NVIDIA/spark-rapids/pull/4946 is merged, because of 4946 is also conflict with this.

res-life:
Filed a Spark issue Interval types are not truncated to the expected endField when creating a DataFrame via Duration https://issues.apache.org/jira/browse/SPARK-38577

res-life:
build

res-life:
build

integration_tests/src/main/python/data_gen.py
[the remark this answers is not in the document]
res-life: Filed a Spark issue: Interval types are not truncated to the expected endField when creating a DataFrame via Duration https://issues.apache.org/jira/browse/SPARK-38577[](https://github.com/res-life)

integration_tests/src/main/python/data_gen.py
res-life: Done, passed in a timedelta for the max/min.

tests/src/test/330/scala/com/nvidia/spark/rapids/CsvScanForIntervalSuite.scala
```diff
@@ -0,0 +1,637 @@
...
+
+  testSparkResultsAreEqual(
+    "test read day-time interval csv file",
+    spark => readCsv(spark, "day-time-interval.csv")
+  ) {
+    df => df
+  }
+
+  /**
+   * TODO: Blocked by Spark overflow issue: https://issues.apache.org/jira/browse/SPARK-38520
```
revans2, line 578: Do we do the right thing with the overflow? If so then we need to document this.

revans2 (approved):

tests/src/test/330/scala/com/nvidia/spark/rapids/CsvScanForIntervalSuite.scala
res-life: Yes, should update doc, another followup PR: https://github.com/NVIDIA/spark-rapids/pull/4981

Issue #4900, opened 2022-03-03
[FEA] - Some unsupported operations on the GPU


eyalhir74:
This epic is to track the following operations for @eyalhir74

 - https://github.com/NVIDIA/spark-rapids/issues/4656 is for being able to have arrays in the grouping key. It already existed.
 - https://github.com/NVIDIA/spark-rapids/issues/4932 is to support the array intersect. Be aware that this is likely not going to support nesting arrays under the arrays, at least for a while.
 - https://github.com/NVIDIA/spark-rapids/issues/4929 is for the min aggregations on arrays.
 - https://github.com/NVIDIA/spark-rapids/pull/4930 is a PR to turn on support for Structs in sort_array

The following is the original text for this request.

While running a few queries we have (in a huge spark to gpu conversion effort), I see a lot of the issues below.
Would be great if we could get some feedback on what can we do next. Currently the performance we get from the GPU is not justifing the migration to RAPIDS :(

`!Expression <Cast> cast(xxx#1105 as string) cannot run on GPU because Cast from ArrayType(StringType,true) to StringType is not supported`

`!Expression <Cast> cast(xxx#1114 as string) cannot run on GPU because Cast from ArrayType(LongType,true) to StringType is not supported`

`!Expression <Cast> cast(yyyy#1108 as string) cannot run on GPU because Cast from ArrayType(StructType(StructField(key,StringType,true), StructField(value_pairsList,ArrayType(StructType(StructField(key,StringType,true), StructField(value,DoubleType,true)),true),true)),true) to StringType is not supported
`
     
`!Expression <Cast> cast(xxx#1109 as string) cannot run on GPU because Cast from ArrayType(StructType(StructField(count,LongType,true), StructField(publisherId,LongType,true)),true) to StringType is not supported
`
      
`!Expression <Min> min(yyyyy#504) cannot run on GPU because expression Min min(pv_userVisitedPublishers#504) produces an unsupported type ArrayType(StructType(StructField(count,LongType,true), StructField(publisherId,LongType,true)),true); input expression AttributeReference pv_userVisitedPublishers#504 (ArrayType(StructType(StructField(count,LongType,true), StructField(publisherId,LongType,true)),true) is not supported)
`

[...]

PR #4930, opened 2022-03-10
Add in support for sorting arrays with structs in sort_array

revans2:
This is a part of #4900 but is only a small part of it. CUDF already supported everything so it was only a few more mins to implement this patch with proper tests than it was to verify that it was working manually.

### DIFF
--- a/docs/supported_ops.md
+++ b/docs/supported_ops.md
@@ -11316,7 +11316,7 @@ are limited.
 <td> </td>
 <td> </td>
 <td> </td>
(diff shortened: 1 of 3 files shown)


revans2:
build

jlowe (commented):
Need to generate the updated supported ops documentation, but otherwise lgtm.

revans2:
build

revans2:
@jlowe I updated the docs. I forgot that buildall does not do that for us anymore, and I have to manually build with the default version to make it happen

gerashegalov:
>  forgot that buildall does not do that for us anymore, and I have to manually build with the default version to make it happen

we could change CI to push the result of `antrun:run@update_config_docs` to the PR branch instead of having it complain that the developer forgot to run it for the default version

jlowe (approved):

Issue #4863, opened 2022-02-24
[FEA] Improve timestamp support in JSON and CSV readers


andygrove:
**Is your feature request related to a problem? Please describe.**
Reading of timestamps in JSON and CSV are not currently consistent with Spark. 

**Describe the solution you'd like**
We should read the values as strings in cuDF and then cast to timestamps in a way that is compatible with Spark, including respecting `timeParserPolicy`.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #4938, opened 2022-03-11
Improve compatibility when reading timestamps from JSON and CSV sources

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/4863 and closes https://github.com/NVIDIA/spark-rapids/issues/123

Improves timestamp support in JSON and CSV to match Spark, by reading from cuDF as strings and then converting to timestamps in the plugin.

There is one follow-on issues:
- https://github.com/NVIDIA/spark-rapids/issues/4940

Status

- [x] Basic functionality in place
- [x] Fix regression in CsvScanSuite
- [x] Improve code documentation
- [x] Review/update compatibility guide
- [x] Review tests


andygrove:
build

andygrove:
build

andygrove:
build

andygrove:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTextBasedPartitionReader.scala
```diff
@@ -236,20 +242,136 @@ abstract class GpuTextBasedPartitionReader(
...
+    // fix timestamps that have milliseconds but no microseconds
+    // example ".296" => ".296000"
+    val placeholder = "@@@"
+    val sanitized = withResource(regexpFiltered) { _ =>
+      // cannot replace with back-refs directly because cuDF cannot support "\1000\2"
+      withResource(regexpFiltered.stringReplaceWithBackrefs(
+        raw"(\.\d{3})(Z?)\Z", raw"\1$placeholder\2")) { tmp =>
+        withResource(Scalar.fromString(placeholder)) { from =>
+          withResource(Scalar.fromString("000")) { to =>
+            tmp.stringReplace(from, to)
```
revans2, line 325: Can we please have a follow on issue to support parsing timestamps ourselves without regular expressions when we can have custom kernels for it? Just because this works, but it is just a bit too much for me.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTextBasedPartitionReader.scala
```diff
@@ -236,20 +242,136 @@ abstract class GpuTextBasedPartitionReader(
...
+      withResource(stripped.matchesRe(regex)) { matchesRe =>
+        withResource(Scalar.fromNull(DType.STRING)) { nullString =>
+          matchesRe.ifElse(stripped, nullString)
+        }
+      }
+    }
+
+    // fix timestamps that have milliseconds but no microseconds
+    // example ".296" => ".296000"
+    val placeholder = "@@@"
```
revans2, line 318: Can you add a comment about why `@@@` is an okay sequence to use here an will never interfere with a real timestamp.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTextBasedPartitionReader.scala
```diff
@@ -359,3 +481,150 @@ abstract class GpuTextBasedPartitionReader(
...
+   * following cuDF formats being returned, in this order:
+   *
+   * - `%Y-%m-%d`
+   * - `%Y-%m-%dT%H:%M`
+   * - `%Y-%m-%dT%H:%M:%S`
+   * - `%Y-%m-%dT%H:%M:%S.%f`
+   */
+  def toCudfFormats(sparkFormat: String, parseString: Boolean): Seq[String] = {
+
+    val optionalFractional = Seq("[.SSS][XXX]", "[.SSS]", "[.SSSSSS]", "[.SSS][XXX]",
```
revans2, line 578: nit: can we move this closer to where it is used?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTextBasedPartitionReader.scala
```diff
@@ -359,3 +481,150 @@ abstract class GpuTextBasedPartitionReader(
...
+
+    val hasZsuffix = sparkFormat.endsWith("Z")
+    val formatRoot = if (hasZsuffix) {
+      sparkFormat.substring(0, sparkFormat.length-1)
+    } else {
+      sparkFormat
+    }
+
+    // strip off suffixes that cuDF will not recognize
+    val str = formatRoot
```
revans2, line 589: nit `str` is not that descriptive

revans2 (commented):
I think the code is okay, but it is really complicated and a lot of assumptions that only a very specific set of formats are allowed. I keep thinking that there might be a simpler way to make it more data driven with look up tables instead of transpiling everything. But then I see we convert the format to both regular expressions and to the CUDF format and I just don't know if the lookup table will actually be small/less error prone or not.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuTextBasedPartitionReader.scala
andygrove: I filed https://github.com/NVIDIA/spark-rapids/issues/4969 for the follow on.

andygrove:
build

revans2 (dismissed):

andygrove:
build

andygrove:
build

andygrove:
build

andygrove:
@revans2 could you re-approve this one, please. I had to upmerge since your last approval.

revans2 (approved):

PR #4947, opened 2022-03-14
Warn if using CUDA driver less than 11.5.0 with ASYNC

abellina:

This overrides the choice of pool to ARENA if the user chose ASYNC and is using a driver that is earlier than 11.5.0. This was added due to some incompatibilities we have seen in 11.4.x.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
@@ -1492,14 +1492,23 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
   lazy val isPooledMemEnabled: Boolean = get(POOLED_MEM)
 
   lazy val rmmPool: String = {
-    val pool = get(RMM_POOL)
-    if ("ASYNC".equalsIgnoreCase(pool) &&
-        (Cuda.getRuntimeVersion < 11020 || Cuda.getDriverVersion < 11020)) {
-      logWarning("CUDA runtime/driver does not support the ASYNC allocator, falling back to ARENA")
-      "ARENA"
-    } else {
-      pool
+    var pool = get(RMM_POOL)
+    val driverVersion = Cuda.getDriverVersion
+    val runtimeVersion = Cuda.getRuntimeVersion
+    var fallbackMessage: Option[String] = None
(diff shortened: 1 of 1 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1492,14 +1492,23 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
...
-        (Cuda.getRuntimeVersion < 11020 || Cuda.getDriverVersion < 11020)) {
-      logWarning("CUDA runtime/driver does not support the ASYNC allocator, falling back to ARENA")
-      "ARENA"
-    } else {
-      pool
+    var pool = get(RMM_POOL)
+    val driverVersion = Cuda.getDriverVersion
+    val runtimeVersion = Cuda.getRuntimeVersion
+    var fallbackMessage: Option[String] = None
+    if ("ASYNC".equalsIgnoreCase(pool)) {
```
jlowe, line 1499: Nit: These are not needed except in the async allocator case.
```suggestion
    if ("ASYNC".equalsIgnoreCase(pool)) {
      val driverVersion = Cuda.getDriverVersion
      val runtimeVersion = Cuda.getRuntimeVersion
      var fallbackMessage: Option[String] = None
```

revans2 (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1492,14 +1492,23 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
...
-      "ARENA"
-    } else {
-      pool
+    var pool = get(RMM_POOL)
+    if ("ASYNC".equalsIgnoreCase(pool)) {
+      val driverVersion = Cuda.getDriverVersion
+      val runtimeVersion = Cuda.getRuntimeVersion
+      var fallbackMessage: Option[String] = None
+      if (runtimeVersion < 11020 || driverVersion < 11020) {
+        fallbackMessage = Some("CUDA runtime/driver does not support the ASYNC allocator")
```
tgravescs, line 1501: I know this was here, but do we want to add version into the message?

jlowe (approved):

Issue #4936, opened 2022-03-11
[BUG]databricks nightly window_function_test failures


tgravescs:
**Describe the bug**
```
[2022-03-11T03:51:14.568Z] FAILED ../../src/main/python/window_function_test.py::test_window_running_rank[Decimal(20,2)][IGNORE_ORDER({'local': True})]
[2022-03-11T03:51:14.568Z] FAILED ../../src/main/python/window_function_test.py::test_window_running[Decimal(20,2)-Long-1000][IGNORE_ORDER({'local': True})]
[2022-03-11T03:51:14.568Z] FAILED ../../src/main/python/window_function_test.py::test_window_running[Decimal(20,2)-Long-1g][IGNORE_ORDER({'local': True})]
```

Perhaps related to https://github.com/NVIDIA/spark-rapids/pull/4924?

```
[2022-03-11T03:51:14.566Z] java.lang.IllegalArgumentException: Part of the plan is not columnar class com.databricks.sql.execution.window.RunningWindowFunctionExec
[...]

PR #4949, opened 2022-03-14
XFail the running window rank tests on [databricks] when using Decimal 128

NVnavkumar:
Test update for #4936

This fix will xfail the tests when using Databricks to test running window rank() aggregations and using Decimal 128 as the partition by parameter type.


tgravescs:
build

Issue #4911, opened 2022-03-08
[FEA][Audit] [SPARK-38314] - Fail to read parquet files after writing the hidden file metadata


nartal1:
**Is your feature request related to a problem? Please describe.**
We have copied FileFormatWriter code in GpuFileFormatWriter.scala so I suspect the bug exists in our code as well. Need to update the file to keep it in sync with Spark's code.

PR: https://github.com/apache/spark/commit/50520fe3eb

firestarman:
GPU does not support writing hidden file meta, and will fall back to CPU writing. Could we have the case to reproduce this failure on GPU ?

PR #4954, opened 2022-03-15
Fix of failing to read parquet files after writing the hidden file metadata in

firestarman:
This change is to follow the update in Spark PR: apache/spark@50520fe3eb

Now GPU writing will fall back to CPU for the hidden file metadata case. So we can not reproduce this failure on GPU, then no need additional tests for this change now.

closes https://github.com/NVIDIA/spark-rapids/issues/4911


firestarman:
build

nartal1 (approved):

PR #4958, opened 2022-03-16
Add independent mvn verify check [skip ci]

### LINKED ISSUE #4939: [FEA]  Add a quick PR check focussing on missed doc generation against the default Spark version
>  forgot that buildall does not do that for us anymore, and I have to manually build with the default version to make it happen

we could change CI to push the result of `antrun:run@update_config_docs` to the PR branch instead of having it complain that the developer forgot to run it for the default version

_Originally posted by @gerashegalov in https://github.com/NVIDIA/spark-rapids/issues/4930#issuecomment-1064402605_

pxLi:

fix #4939 

The check is running in github runners, so the triggering does not require manual operation which is the same as signoff check (PR open, reopen, new commit)

Verified in forked repo, and working in this PR. the check would normally take 2-3 minutes. 
If this looks OK, I will merge and let it run for some days before mark it as a required check, thanks!

Success
image

### DIFF
--- a/.github/workflows/docgen-check.yml
+++ b/.github/workflows/docgen-check.yml
@@ -1,0 +1,45 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# A workflow to check if generated docs are up to date
+name: doc-gen check
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  docgen-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2 # refs/pull/:prNumber/merge
+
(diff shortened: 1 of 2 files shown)


pxLi:
build

pxLi:
@gerashegalov @revans2 can you help take a look if the command and time of check looks OK? Thanks!

.github/workflows/docgen-check.yml
```diff
@@ -0,0 +1,45 @@
...
+          java-version: 8
+
+      - name: build deps
+        run: mvn install -pl 'aggregator' -am -DskipTests
+
+      - name: doc-gen
+        run: mvn antrun:run@update_config_docs -pl dist
+
+      - name: check if existing modified files
+        run: mvn exec:exec@if_modified_files -P pre-merge -pl dist
```
gerashegalov, line 45: these three invocations of maven should be just one maven step to the tune of 
```
 mvn verify -pl dist -am -DskipTests -Dskip -Dmaven.javadoc.skip
 ```
 
 that takes in 2+ minutes. It will take care of style checks, default shim version and conf generation.

pxLi: sounds good, I will refactor the check to cover more

pxLi: I was intentional to separate stages, so people could quickly figure out which stage their job failed. But as they still need to check the log, leverage a single command to cover should be fine

pxLi:
refactor to replace doc-gen only check w/ mvn-verify check (include rat, code style and doc-gen)

pxLi:
build

tgravescs:
so to be clear, this check will run before it kicks off the build and test, or in parallel?  It just fails vs updating as issue title implies.  Lets update the issue title to match.

gerashegalov (approved):

pxLi:
> so to be clear, this check will run before it kicks off the build and test, or in parallel? It just fails vs updating as issue title implies. Lets update the issue title to match.

Its a test run in parallel which cost much less time (2-3 mins), and it does not require to be manually triggered (gets trigger when PR open/reopen, or new commits pushed).

Thanks Gera for updating the issue title~

Issue #4745, opened 2022-02-10
[DOC] add example python api call for ExplainPlan.explainPotentialGpuPlan


tgravescs:
## Report incorrect documentation

**Location of incorrect documentation**
We currently don't have an example calling the python api, this might be a bit confusing to users who aren't used to it so lets add an example.

spark._jvm.com.nvidia.spark.rapids.ExplainPlan.explainPotentialGpuPlan(df3._jdf, "ALL")

PR #4965, opened 2022-03-16
Documentation: add example python api call for ExplainPlan.explainPotentialGpuPlan [skip ci]

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/4745

### DIFF
--- a/docs/get-started/getting-started-workload-qualification.md
+++ b/docs/get-started/getting-started-workload-qualification.md
@@ -158,7 +158,7 @@ with adaptive query execution enable, it will not be able to show any changes
 in the plan due to that.
 
 This is very similar output you would get by running the query with the
-Rapids Accelerator enabled and with the config `spark.rapids.sql.enabled` enabled.
+RAPIDS Accelerator enabled and with the config `spark.rapids.sql.enabled` enabled.
 
 Requires the RAPIDS Accelerator for Apache Spark jar and RAPIDS cudf jar be included
 in the classpath but the RAPIDS Accelerator for Apache Spark should be disabled.
--- a/docs/get-started/getting-started-workload-qualification.md
+++ b/docs/get-started/getting-started-workload-qualification.md
@@ -175,7 +175,7 @@ output = sc._jvm.com.nvidia.spark.rapids.ExplainPlan.explainPotentialGpuPlan(df.
 
 Parameters:  
 `df` - The Spark DataFrame to get the query plan from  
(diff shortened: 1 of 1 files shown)


tgravescs:
build

tgravescs:
build

docs/get-started/getting-started-workload-qualification.md
```diff
@@ -188,7 +188,20 @@ which is the same as the driver logs with `spark.rapids.sql.explain=all`.
...
+   val output=com.nvidia.spark.rapids.ExplainPlan.explainPotentialGpuPlan(df)
+   ```
+
+   Python API:
+   ```python
+   output = sparkSession._jvm.com.nvidia.spark.rapids.ExplainPlan.explainPotentialGpuPlan(df._jdf, "ALL")
+   ```
+     - First parameter, `df` is the Spark DataFrame to get the query plan from.
+     - Second parameter, `ALL` returns all the explain data, you can change to `NOT_ON_GPU` to return just what
+       does not work on the GPU. The Scala API, ALL is the default when second parameter is not specified.
```
jlowe, line 202: "The Scala API," is a bit confusing here, was it intended to say "For the Scala API," or was this supposed to be in a separate section like "Python API:" above?

tgravescs: really I just want the java doc of this function which we don't publish so I changed this to just have a function documentation section

tgravescs:
IO need to fix a few formatting things

tgravescs:
ok formatting is fixed

tgravescs:
build

docs/get-started/getting-started-workload-qualification.md
```diff
@@ -145,6 +145,47 @@ which is the same as the driver logs with `spark.rapids.sql.explain=all`.
...
+```
+
+Looks at the CPU plan associated with the dataframe and outputs information
+about which parts of the query the RAPIDS Accelerator for Apache Spark
+could place on the GPU. This only applies to the initial plan, so if running
+with adaptive query execution enable, it will not be able to show any changes
+in the plan due to that.
+
+This is very similar output you would get by running the query with the
+Rapids Accelerator enabled and with the config `spark.rapids.sql.enabled` enabled.
```
jlowe, line 161:
```suggestion
RAPIDS Accelerator enabled and with the config `spark.rapids.sql.enabled` enabled.
```

tgravescs:
build

jlowe (approved):

Issue #4913, opened 2022-03-08
[BUG] Fall back to the CPU if we see a scale on Ceil or Floor


revans2:
**Describe the bug**
#4910 showed that a new parameter, scale, was added to floor and ceil in 3.3. If we see a a scale that is not 0 we cannot support this and must fall back to the CPU to avoid data corruption. At least until we can add in support for it properly.

tgravescs:
> At least until we can add in support for it properly.

Do we have a follow on issue to add support for it or we were just going to wait for customer to request?

Issue #4910, opened 2022-03-07
[FEA] [Audit] [SPARK-37475][SQL] Add scale parameter to floor and ceil functions


nartal1:
**Is your feature request related to a problem? Please describe.**
Below PR adds `scale` parameter to `floor`/`ceil` functions in order to allow users to control the rounding position. Need to evaluate if we have to pull in this change for Spark-3.3.

Spark PR: https://github.com/apache/spark/commit/62421455e5

revans2:
It is unlikely that we are going to be able to support this fully without some changes to CUDF.

CUDF's round only supports HALF_UP and HALF_EVEN.

https://github.com/rapidsai/cudf/blob/57ff6f55b9fd44e8a8e10282d3f95d5f38e299ef/cpp/include/cudf/round.hpp#L36

Floor and Ceil are unary operators and do not support setting a scale.

So in the sort term we are probably going to have to fall back to the CPU if we see a floor or ceil with a scale that is not 0, and then later on put in support to CUDF for what we need.

PR #4971, opened 2022-03-16
Fall back to the CPU for non-zero scale on Ceil or Floor functions [databricks]

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/4913.
Apache Spark has added scale parameter for Ceil/Floor functions in Spark-3.3 https://github.com/NVIDIA/spark-rapids/issues/4910. When ceil/floor functions are called with scale parameter, the physical plan changes compared to earlier. It adds `RoundCeil`/`RoundFloor` and the result type is decimal type.

In this PR, we have added support to RoundCeil and RoundFloor if the scale is `0`. i.e we call GpuCeil if ceil is called with scale 0. For scale values other than `0`, we fallback to CPU.


nartal1:
build

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -351,11 +351,32 @@ def test_floor(data_gen):
...
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : unary_op_df(spark, data_gen).selectExpr('ceil(a, 0)'),
+            conf={'spark.rapids.sql.castFloatToDecimal.enabled':'true'})
+
+@pytest.mark.skipif(is_before_spark_330(), reason='scale parameter in Ceil function is not supported before Spark 3.3.0')
+@allow_non_gpu('ProjectExec')
+@pytest.mark.parametrize('data_gen', double_n_long_gens + _arith_decimal_gens_no_neg_scale, ids=idfn)
+def test_ceil_scale_nonzero(data_gen):
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : unary_op_df(spark, data_gen).selectExpr('ceil(a, -1)'))
```
razajafri, line 378: would it be an overkill to ask for a int_gen for the scale value even if we limit it to -5, 5 range?

razajafri:
@gerashegalov @jlowe what do you guys think about adding utility classes, like `RapidsFloorCeilUtils` in this PR and `RapidsErrorUtils` previously, as we move forward as opposed to just adding methods in the `SparkShims` trait? 
My vote is for having individual modular classes like the ones I mentioned above in favor of just adding methods to `SparkShimsImpl`.

integration_tests/src/main/python/arithmetic_ops_test.py
nartal1: Sorry, I did not understand your question completely. 
Cudf doesn't support `scale` for Ceil/Floor functions. Those are currently unaryExpressions. So until we have cudf implementation of scale for these functions plan is to fall back to CPU. Scale with value `0` is an exception as it is same as calling Ceil/Floor functions. Just that the physical plan changes.

razajafri: Then what are we really testing in this test? Would it make sense to add this test when we actually support ceil/floor with scale?

nartal1: Had added this test for completion i.e to show that we fall back to cpu for non zero scale by having allow_non_gpu. Your suggestion makes sense. I can remove this test and just have scale=0 test.

nartal1: Removed this test. PTAL.

razajafri (dismissed):

nartal1:
build

sql-plugin/src/main/301until330-all/scala/org/apache/spark/sql/rapids/shims/RapidsFloorCeilUtils.scala
```diff
@@ -0,0 +1,32 @@
...
+
+package org.apache.spark.sql.rapids.shims
+
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.rapids.GpuFloorCeil
+import org.apache.spark.sql.types.{DataType, DecimalType, LongType}
+
+object RapidsFloorCeilUtils {
+
+  def getDataType(child: Expression): DataType = {
```
revans2, line 25: nit: can we rename this to outputDataType? It make it much more clear what is happening in my opinion. 

Also I think I would prefer to have the child's data type passed in instead of the child expression itself.  It makes slightly more code common and also I think makes this method more specific.

But these are both nits that I can live without.

sql-plugin/src/main/301until330-all/scala/org/apache/spark/sql/rapids/shims/RapidsFloorCeilUtils.scala
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.rapids.shims
+
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.rapids.GpuFloorCeil
+import org.apache.spark.sql.types.{DataType, DecimalType, LongType}
+
+object RapidsFloorCeilUtils {
```
revans2, line 23: Why create a new class instead of using GpuFloorCeil?

revans2 (commented):
Just some nits and a question. Looks good.

sql-plugin/src/main/301until330-all/scala/org/apache/spark/sql/rapids/shims/RapidsFloorCeilUtils.scala
revans2: That is fine, I just wanted to understand it.

nartal1:
build

sql-plugin/src/main/301until330-all/scala/org/apache/spark/sql/rapids/shims/RapidsFloorCeilUtils.scala
nartal1: done. PTAL

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -351,11 +351,32 @@ def test_floor(data_gen):
...
+def test_floor_scale_zero(data_gen):
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark : unary_op_df(spark, data_gen).selectExpr('floor(a, 0)'),
+            conf={'spark.rapids.sql.castFloatToDecimal.enabled':'true'})
+
+@pytest.mark.skipif(is_before_spark_330(), reason='scale parameter in Floor function is not supported before Spark 3.3.0')
+@allow_non_gpu('ProjectExec')
+@pytest.mark.parametrize('data_gen', double_n_long_gens + _arith_decimal_gens_no_neg_scale, ids=idfn)
+def test_floor_scale_nonzero(data_gen):
+    assert_gpu_and_cpu_are_equal_collect(
```
revans2, line 365: There is a specific API to verify that we fell back to the CPU. `assert_gpu_fallback_collect`

revans2 (approved):

revans2:
build

revans2 (approved):

nartal1:
CI failed due to databricks test failure. Fix is in https://github.com/NVIDIA/spark-rapids/pull/4998. Will upmerge once that PR is merged.

nartal1:
build

Issue #1922, opened 2021-03-12
[FEA] Support ORC forced positional evolution


nartal1:
**Is your feature request related to a problem? Please describe.**
We have copied code from OrcUtils.scala to GpuOrcScan.scala. We need to pull in the changes from https://github.com/apache/spark/commit/00d43b1f82

PR #4972, opened 2022-03-16
Add support for ORC forced positional evolution

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/1922

This is to keep OrcUtils in sync with Spark's code. In the Spark PR https://github.com/apache/spark/pull/29737/ they have added a check for 'orc.force.positional.evolution` for matching top level column by position. 
Updated the code so that when this config is set, we read the columns based on positions and not the column names.
Added test to verify the behavior.


razajafri:
Adding a test is always a good idea, even if it's a simple one just to do a sanity check

jlowe:
This looks good, but agree it would be nice to have a test to show this works.  The original Spark PR could be a reference of how to approach testing this.

nartal1:
build

nartal1:
build

jlowe (approved):

razajafri (approved):

PR #4973, opened 2022-03-17
Implement support for ArrayExists expression

### LINKED ISSUE #4815: [FEA] Support org.apache.spark.sql.catalyst.expressions.ArrayExists
I wish we can support org.apache.spark.sql.catalyst.expressions.ArrayExists.

Mini repro:
```
val arrayData = Seq(
    Row("John",List("apple","orange","banana"),1),
    Row("David",List("apple","orange","banana"),2),
    Row("Harry",List("apple","other"),1)
)

val arraySchema = new StructType().add("name",StringType).add("fruits", ArrayType(StringType)).add("favorite",IntegerType)

val df = spark.createDataFrame(spark.sparkContext.parallelize(arrayData),arraySchema)
df.write.mode("overwrite").format("parquet").save("/tmp/testparquet")
val df2=spark.read.parquet("/tmp/testparquet")
df2.createOrReplaceTempView("df2")

spark.sql("select name,exists(fruits,x -> x == 'other') as if_love_apple from df2").show()
```

Unsupported messages:
```
! <ArrayExists> exists(fruits#92, lambdafunction((lambda x#207 = other), lambda x#207, false)) cannot run on GPU because GPU does not currently support the operator class org.apache.spark.sql.catalyst.expressions.ArrayExists
```

gerashegalov:
This PR implements ArrayExists, it has two major phases
1.  first apply function to produce array of Booleans 
2. run segmented reduce ANY to if any of the values are true

Spark 3.x default is the 3VL logic:
- if any element is true the array maps to true
- if no element is true and there is at least one null, the array maps to null
- if no element is true and none is null, the array maps to false

Legacy mode 2VL:
- if any element is true the array maps to true
- if no element is true , the array maps to false

Closes #4815


gerashegalov:
build

integration_tests/src/main/python/array_test.py
```diff
@@ -331,3 +331,29 @@ def test_get_array_struct_fields(data_gen):
...
+        element_type = data_gen.data_type.elementType
+        if isinstance(element_type, IntegralType):
+            columns.extend([
+                'exists(a, item -> item % 2 = 0) as exists_even',
+                'exists(a, item -> item < 0) as exists_negative',
+                'exists(a, item -> item >= 0) as exists_non_negative'
+            ])
+
+        if isinstance(element_type, StringType):
+            columns.extend(['exists(a, entry -> length(entry) > 5) as exists_upper_case'])
```
revans2, line 352: nit: exists_upper_case is checking for length, not upper case.

integration_tests/src/main/python/array_test.py
```diff
@@ -331,3 +331,29 @@ def test_get_array_struct_fields(data_gen):
...
+
+        return unary_op_df(spark, data_gen).selectExpr(columns)
+
+    assert_gpu_and_cpu_are_equal_collect(do_it, conf= {
+        'spark.sql.legacy.followThreeValuedLogicInArrayExists' : threeVL,
+        'spark.rapids.sql.regexp.enabled'                      : True,
```
revans2, line 358: Why do we need regular expressions turned on here?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2857,6 +2857,29 @@ object GpuOverrides extends Logging {
...
+      "Return true if any element satisfies the predicate LambdaFunction",
+      ExprChecks.projectOnly(TypeSig.BOOLEAN,
+        TypeSig.ARRAY.nested(TypeSig.all),
+        Seq(
+          ParamCheck("argument",
+            TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL_128 + TypeSig.NULL +
+                TypeSig.ARRAY + TypeSig.STRUCT + TypeSig.MAP),
+            TypeSig.ARRAY.nested(TypeSig.all)),
+          ParamCheck("function",
+            (TypeSig.commonCudfTypes + TypeSig.DECIMAL_128 + TypeSig.NULL +
```
revans2, line 2870: Actually I think this is the return type of the function, so I think this should be TypeSig.BOOLEAN for both.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/higherOrderFunctions.scala
```diff
@@ -277,20 +265,148 @@ case class GpuArrayTransform(
     }
   }
 
+  protected def transformListColumnView(
```
revans2, line 268: Can we have a few comments here explaining what this does and what the life cycle of `lambdaTransformedCV` is. i.e. it is required the the implementation of this close it?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/higherOrderFunctions.scala
```diff
@@ -277,20 +265,148 @@ case class GpuArrayTransform(
...
+  }
+
+  private def existsReduce(columnView: cudf.ColumnView, nullPolicy: cudf.NullPolicy) = {
+    columnView.listReduce(
+      cudf.SegmentedReductionAggregation.any(),
+      nullPolicy,
+      DType.BOOL8)
+  }
+
+  private def replaceChildNullsByFalse(cv: cudf.ColumnView): cudf.ColumnView = {
```
revans2, line 351: nit: can we rename this to `replaceChildNullsByFalseView` or at least add in a comment to make sure that it is clear that the input cv cannot be closed until the output of this function is also closed.

revans2 (dismissed):
Overall it looks really good. Just some nits and docs to clean up.

revans2 (approved):

revans2 (approved):

revans2:
build

PR #4977, opened 2022-03-17
Add missing 314 to buildall script

jlowe:
Fixes the `buildall` script when building with the `snapshots` profile, since it was not building for Spark 3.1.4 but the snapshot profile expected the aggregator artifacts to have been locally installed.


jlowe:
build

tgravescs (approved):

PR #4981, opened 2022-03-18
Update doc for CSV reading interval

res-life:
Contributes #4146

This is a followup for PR https://github.com/NVIDIA/spark-rapids/pull/4927 to update doc.
This PR describes the compatibility for CSV ANSI day time interval.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -358,19 +358,30 @@ They are within round-off errors except when they are close enough to overflow t
 then results in a number being returned when the CPU would have returned null.
 
 ### CSV ANSI day time interval
-Apache Spark has an overflow issue when reading ANSI day time interval, the plugin fixed it on GPU.
-The issue is https://issues.apache.org/jira/browse/SPARK-38520.   
-e.g.:   
+This type was added in as a part of Spark 3.3.0, and it's not supported on Spark versions before 3.3.0.
+Apache Spark can [overflow](https://issues.apache.org/jira/browse/SPARK-38520) when reading ANSI day time interval values.
+The RAPIDS Accelerator does not overflow and as such is not bug for bug compatible with Spark in this case.
 
-interval string in csv | Spark reads to | Plugin reads to | comments
------|---------------------------|-----------------|-----------
-interval '106751992' day| INTERVAL '-106751990' DAY | NULL| Spark issue
-interval '2562047789' hour| INTERVAL '-2562047787' HOUR | NULL| Spark issue
+Interval string in csv|Spark reads to|The RAPIDS Accelerator reads to|Comments|
+-----|-------------------------|-----------------|-----------|
+interval '106751992' day| INTERVAL '-106751990' DAY | NULL| Spark issue|
+interval '2562047789' hour| INTERVAL '-2562047787' HOUR | NULL| Spark issue|
 
-There are two valid textual representations in CSV, take DAY TO SECOND interval for example:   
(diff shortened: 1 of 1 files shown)


res-life:
build

docs/compatibility.md
```diff
@@ -357,6 +357,21 @@ Also parsing of some values will not produce bit for bit identical results to wh
 They are within round-off errors except when they are close enough to overflow to Inf or -Inf which
 then results in a number being returned when the CPU would have returned null.
 
+### CSV ANSI day time interval
+Apache Spark has an overflow issue when reading ANSI day time interval, the plugin fixed it on GPU.
```
revans2, line 361: Can we use "the RAPIDS Accelerator" instead of "the plugin"?

i.e. 

```
Apache Spark can [overflow](https://issues.apache.org/jira/browse/SPARK-38520) when reading ANSI 
day time interval values. The RAPIDS Accelerator does not overflow and as such is not bug for bug 
compatible with Spark in this case.
```

docs/compatibility.md
```diff
@@ -357,6 +357,21 @@ Also parsing of some values will not produce bit for bit identical results to wh
...
+
+interval string in csv | Spark reads to | Plugin reads to | comments
+-----|---------------------------|-----------------|-----------
+interval '106751992' day| INTERVAL '-106751990' DAY | NULL| Spark issue
+interval '2562047789' hour| INTERVAL '-2562047787' HOUR | NULL| Spark issue
+
+There are two valid textual representations in CSV, take DAY TO SECOND interval for example:   
+INTERVAL '100 10:30:40.999999' DAY TO SECOND   
+100 10:30:40.999999  
+Currently only supports the first one which is the default when Apache Spark writing.
```
revans2, line 373: Currently the RAPIDS Accelerator only supports the first ...

Also if this is true why is this type enabled by default for CSV reading. Shouldn't we have a config and have it disabled for DAY TO SECOND?

docs/compatibility.md
```diff
@@ -357,6 +357,21 @@ Also parsing of some values will not produce bit for bit identical results to wh
 They are within round-off errors except when they are close enough to overflow to Inf or -Inf which
 then results in a number being returned when the CPU would have returned null.
 
+### CSV ANSI day time interval
```
revans2, line 360: nit: can we explain that this type was added in as a part of 3.3.0?

docs/compatibility.md
res-life: The short one is style of HIVE.
See: https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/HiveResult.scala#L130
```
    case (duration: Duration, DayTimeIntervalType(startField, endField)) =>
      toDayTimeIntervalString(
        durationToMicros(duration, endField),
        HIVE_STYLE,
        startField,
        endField)
// Calls IntervalUtils.toDayTimeIntervalString
```
Spark does not generate this kind of representation.

docs/compatibility.md
res-life: done

docs/compatibility.md
res-life: done

res-life:
build

docs/compatibility.md
revans2: I understand, but that means that someone can still run into it. From a documentation standpoint I think this is fine. But I would like to see follow work to enable support for this, or have us fall back to the CPU if we cannot.

PR #4992, opened 2022-03-21
Support Collect-like Reduction Aggregations

sperlingxx:
Current PR is the last piece of the puzzle of the [collect aggregations supporting](https://github.com/NVIDIA/spark-rapids/issues/2062), which enables collect aggregations under reduction context in spark-rapids. 

Signed-off-by: sperlingxx <lovedreamf@gmail.com>

PR #4996, opened 2022-03-21
Fix collect_set on struct type

sperlingxx:
During the development of collect-like reduction aggregation #4992, I found that `CudfCollectSet` and `CudfMergeSets` apply wrong `NullEquality`, which should be `NullEquality.EQUAL`. This problem only affects struct types because collect ops apply `NullPolicy.EXCLUDE`, which means null entries will be excluded during the collection. However, with struct type, nulls   wrapped with outer-level struct will trigger the NullEquality problem.

Since #4992 is targetted to release 22.06 , I create a separate PR to fix this problem. 

### DIFF
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -588,7 +588,8 @@ _gen_data_for_collect_list_op = _full_gen_data_for_collect_op + [[
 
 _repeat_agg_column_for_collect_set_op = [
     RepeatSeqGen(all_basic_struct_gen, length=15),
-    RepeatSeqGen(StructGen([['child0', all_basic_struct_gen]]), length=15)]
+    RepeatSeqGen(StructGen([
+        ['c0', all_basic_struct_gen], ['c1', int_gen]]), length=15)]
 
 _gen_data_for_collect_set_op_for_unique_group_by_key = [[
     ('a', LongRangeGen()),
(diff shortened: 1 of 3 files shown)


sperlingxx:
build

firestarman:
I saw https://github.com/NVIDIA/spark-rapids/pull/4992 is targeted to release 22.04.

sperlingxx:
> I saw #4992 is targeted to release 22.04.

The branch-22.06 has not been created yet.

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -656,25 +654,11 @@ def test_hash_groupby_collect_set(data_gen):
...
-@approximate_float
-@ignore_order(local=True)
-@incompat
-@pytest.mark.parametrize('data_gen', _gen_data_for_collect_set_op_for_unique_group_by_key, ids=idfn)
-def test_hash_groupby_collect_set_on_nested_type_for_unique_group_by(data_gen):
-    assert_gpu_and_cpu_are_equal_collect(
-        lambda spark: gen_df(spark, data_gen, length=100)
-            .groupby('a')
-            .agg(f.collect_set('b')))
+            .agg(f.sort_array(f.collect_set('b'))))
```
abellina, line 661: this test lost the `f.count('b')` was that intentional?

sperlingxx: Yes, because we want to test the case of nested struct, while nested struct has not been supported in `count` yet.

abellina (approved):
LGTM, but I'd like a +1 from @revans2 or @jlowe also. I am not an expert at nested types.

revans2 (approved):

PR #4955, opened 2022-03-15
Add UDT support to ParquetCachedBatchSerializer (CPU)

razajafri:
This PR adds UDT support for the case when PCBS fallsback to CPU and make sure we don't put stuff on the GPU if the output of `InMemoryRelation` is not supported. 

fixes #4806

PR #4998, opened 2022-03-21
Added the id back for struct children to make them unique [databricks]

razajafri:
The children of the struct also need to have a unique name which was erroneously removed as part of the #4955 

Also, removed the unnecessary concurrent AtomicLong and use a simple stack variable. Also added a test to see if the order of the structs matters when writing to caching and pulling from cache
  
### DIFF
--- a/integration_tests/src/main/python/cache_test.py
+++ b/integration_tests/src/main/python/cache_test.py
@@ -115,6 +115,17 @@ def test_cache_partial_load(data_gen, enable_vectorized_conf):
     assert_gpu_and_cpu_are_equal_collect(partial_return(f.col("a")), conf=enable_vectorized_conf)
     assert_gpu_and_cpu_are_equal_collect(partial_return(f.col("b")), conf=enable_vectorized_conf)
 
+@pytest.mark.parametrize('enable_vectorized_conf', enable_vectorized_confs, ids=idfn)
+@allow_non_gpu('CollectLimitExec')
+def test_cache_reverse_order(enable_vectorized_conf):
+    col0 = StructGen([['child0', StructGen([['child1', byte_gen]])]])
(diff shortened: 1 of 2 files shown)


razajafri:
build

firestarman (approved):

razajafri:
@pxLi how the PR #4955 was able to pass CI? while the PR #4971 created by @nartal1 failed CI?

pxLi:
> @pxLi how the PR #4955 was able to pass CI? while the PR #4971 created by @nartal1 failed CI?

#4955 was merged 3 days ago, premerge CI was just testing against merged code of PR at that moment, I am not sure if any following commits break the tests

pxLi:
@razajafri also https://github.com/NVIDIA/spark-rapids/pull/4955 run w/o databricks while #4971 failed only w/ databricks

razajafri:
@pxLi how the PR #4955 was able to pass CI? while the PR #4971 created by @nartal1 failed CI? 

> @razajafri also #4955 run w/o databricks while #4971 failed only w/ databricks
> 
> And IIRC, we do not have cache test w/ PCBS enabled in pre-merge CI non-databricks part (we have for databricks and nightly CI though). As our base shim has become spark311, If you need this to be tested in premerge non-db env, either make the PCBS as default config or add a separate test in the https://github.com/NVIDIA/spark-rapids/blob/branch-22.04/jenkins/spark-premerge-build.sh#L105 ci_2 part~

The premerge doesn't run PCBS but I wasn't aware that we run PCBS tests for DB. Thanks for letting me know. 
@tgravescs I know we decided against running premerge for PCBS in the past but given that we have 31x as the default version should we turn the premerge with PCBS on even for nondb Spark versions?

Issue #4821, opened 2022-02-18
JSON reader: Collect more corner cases in JSON reader parsing


HaoYang670:
Track https://github.com/NVIDIA/spark-rapids/issues/9#issuecomment-959803463 and https://github.com/NVIDIA/spark-rapids/issues/4138
There are some odd corner cases in JSON reader parsing in Spark, which are difficult to match the behaviour on GPU. We need to find more of them.

HaoYang670:
# case sensitivity
```json
{"a": 1, "b": 2}
{"A": 3, "B": 4}
```
produces
```scala
+----+----+
|   a|   b|
+----+----+
|   1|   2|
|null|null|
+----+----+
```
in Spark

HaoYang670:
# String

## 1
```json
{"a":::::::::::}
```
is parsed as
```scala
org.apache.spark.sql.AnalysisException:
Since Spark 2.3, the queries from raw JSON/CSV files are disallowed when the
referenced columns only include the internal corrupt record column
(named _corrupt_record by default).
```
in Spark, and
```python
            a
0  ::::::::::
```
in CUDF

HaoYang670:
# How many lines are there?

## 1
```json
{"a": "This is the first line"}






```
is parsed as 
```scala
+--------------------+
|                   a|
+--------------------+
|This is the first...|
+--------------------+
```
in Spark, and
```python
                        a
0  This is the first line
1                    <NA>
2                    <NA>
3                    <NA>
4                    <NA>
5                    <NA>
```
in CUDF

HaoYang670:
# Array and Struct

## 1
```json
{"a": [1,2]}
```
is parsed as 
```scala
+------+
|     a|
+------+
|[1, 2]|
+------+
```
in Spark, and 
```python
    a
0  [1
```
in CUDF.

HaoYang670:
@revans2 Could you please help to review? Are these the `corner cases` in your mind?

HaoYang670:
Where should we document them? 
Github issues, source code or the official spark-rapids docs, such as https://nvidia.github.io/spark-rapids/docs/compatibility.html?

GaryShen2008:
We'll need to create a CI job to run fuzzing test frequently to see if we can find more corner cases.
Move it to 22.08 since it's not in 22.06's target.

PR #5001, opened 2022-03-22
Add fuzzing test for JSON reader

### LINKED ISSUE #4138: [FEA] JSON: fuzz testing
After we have a working JSON reader, #4135 and further extended in #4136, we should try to do some fuzz testing on this.  There are already a lot of corner cases that don't appear to be JSON spec compatible https://github.com/NVIDIA/spark-rapids/issues/9#issuecomment-959803463 and we should look at doing fuzz testing to see what other types of issues we might come up with.

HaoYang670:
Add fuzz tests for JSON reading.
close #4138

This PR contains 3 parts:
1. Schema generator: randomly generate a JSON schema described in PySpark's DataType
2. JSON generator: randomly generator a JSON string based on the given schema
3. test_json_read_fuzz: test the behavior of CPU and GPU on reading a random json file (marked Xfail, so it will not impact the CI job)

Data Type supported so far:
1. primitive types: int, long, float, double, string
2. nested types: struct and array

Plan to support in the future:
1. primitive types: date, timestamp
2. nested types: map


HaoYang670:
build

integration_tests/src/main/python/json_fuzzing_test.py
```diff
@@ -0,0 +1,417 @@
...
+    yield random.choice(["true", "null", "false"])
+
+
+
+_enable_all_types_conf = {
+    'spark.rapids.sql.format.json.enabled': 'true',
+    'spark.rapids.sql.format.json.read.enabled': 'true'}
+
+@approximate_float
+@pytest.mark.xfail(reason = "fuzz test")
```
abellina, line 397: should there be at least a second test that is always expected to succeed?

Naively I am also not sure why this test would be expected to fail. I guess the other point to make, would be to add a github issue link on what is failing. Perhaps that's well known to others, but it isn't obvious from this change to me.

HaoYang670: The aim of this fuzzy test is to help find more corner cases in JSON reading. Link: https://github.com/NVIDIA/spark-rapids/issues/4821

Initially, my thought was that it would spoil our CI tests if a corner case was found. So I marked it as `xfail`. However, after thinking carefully about your feedback, I remove the `xfail`. Maybe we should let the CI fail when we find a new corner case, although the possibility is low.

HaoYang670:
build

integration_tests/src/main/python/json_fuzzing_test.py
```diff
@@ -0,0 +1,415 @@
...
+
+import random
+from typing import List
+from data_gen import *
+from asserts import assert_gpu_and_cpu_are_equal_collect
+from marks import approximate_float
+
+_name_gen = StringGen(pattern= "[a-z]{1,30}",nullable= False)
+_name_gen.start(random)
+_string_gen = StringGen(pattern= "[a-z]{1,30}",nullable= False)
```
andygrove, line 25: We should include a wider range of characters in the generated strings. Spark supports parsing dates and times from JSON strings so we should include characters such as digits and symbols such as ':', ".", and '-'. We may want to also have specific fuzzer code for generating dates and timestamps as strings.

andygrove: We should also include whitespace characters such as `\n`, `\r`, `\t`, and ' '. Perhaps we can just use the full ASCII character set.

integration_tests/src/main/python/json_fuzzing_test.py
```diff
@@ -0,0 +1,415 @@
...
+_string_gen.start(random)
```
andygrove, line 26: Are these tests deterministic, using a fixed seed?

integration_tests/src/main/python/json_fuzzing_test.py
HaoYang670: @andygrove  The current generator just support generating integer, fp, and string. In the future, I will add more data types such as date, timestamp and map.

HaoYang670: > We should also include whitespace characters such as `\n`, `\r`, `\t`, and ' '. Perhaps we can just use the full ASCII character set.

Done

integration_tests/src/main/python/json_fuzzing_test.py
HaoYang670: Done

integration_tests/src/main/python/json_fuzzing_test.py
HaoYang670: Done

HaoYang670:
build

HaoYang670:
build

HaoYang670:
build

HaoYang670:
build

pxLi:
please check the failed log before re-trigger,
```
Error: 3-25T06:40:13.649Z] [ERROR] Failed to execute goal org.apache.rat:apache-rat-plugin:0.13:check (default) on project rapids-4-spark-parent: Too many files with unapproved license: 1 See RAT report in: /home/jenkins/agent/workspace/jenkins-rapids_premerge-github-4260/target/spark24X/rat.txt -> [Help 1]
```

exclude related json filed at https://github.com/NVIDIA/spark-rapids/blob/branch-22.04/pom.xml#L1154-L1178 to pass the license test

HaoYang670:
build

HaoYang670:
build

integration_tests/src/main/python/json_fuzzing_test.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+
+_enable_all_types_conf = {
+    'spark.rapids.sql.format.json.enabled': 'true',
+    'spark.rapids.sql.format.json.read.enabled': 'true'}
+
+@approximate_float
+@allow_non_gpu('FileSourceScanExec')
+@pytest.mark.xfail(reason="fuzz test may randomly fail")
```
revans2, line 398: To be clear fuzz testing is intended to find issues and should not be running by default as a part of integration tests. It should be something that we can enable and run for a configurable number of iterations to find problems. Once we find the problem we need to have a way to debug/reproduce this issue. Then we can file a follow on issue/bug to fix it. 

I would much rather have us skip this test by default unless a flag is set to enable fuzz testing, preferably with a number of iterations.  With xfail we want to get to the point where if a test passes when we expect it to fail then it will be marked as a failure, `xfail_strict=true`. Because this is totally random I really don't want to mark it as xfail.

integration_tests/src/main/python/json_fuzzing_test.py
```diff
@@ -0,0 +1,413 @@
...
+def test_json_read_fuzz(spark_tmp_path):
+    depth = random.randint(1, 5)
```
revans2, line 400: How do we reproduce/debug an error when it happens? Everything is random and based off of a regular random. Typically in python we would want to create an instance of Random with a seed, and then pass it everywhere so the results can be reproduced. We do this in data_gen already.

https://github.com/NVIDIA/spark-rapids/blob/00630537365c034186fa9031694edb503e4341a6/integration_tests/src/main/python/data_gen.py#L671

I am fine if we randomly generate a seed based off of a timestamp or something like that. But at a minimum we need a way to log what that seed was so if/when it does fail we can debug things.  With xdist this is not super simple, so I would suggest that we catch all Exceptions and then wrap them with something new that includes the seed that triggered the problem.

integration_tests/src/main/python/json_fuzzing_test.py
HaoYang670: Done. Fuzz tests are skipped by defualt.
We can enable fuzz tests and store the tested data by running
```
> bash run_pyspark_from_build.sh -k test_json_read_fuzz --fuzz_test --debug_tmp_path
```

integration_tests/src/main/python/json_fuzzing_test.py
HaoYang670: Add the flag `--debug_tmp_path` can save the test data

integration_tests/src/main/python/json_fuzzing_test.py
```diff
@@ -0,0 +1,413 @@
...
+    path = random.randint(0, 2)
+    if path == 0:
+        for t in gen_digit():
+            yield t
+    elif path == 1:
+        yield chr(random.randint(0x41, 0x46))
+    else:
+        yield chr(random.randint(0x61, 0x66))
+
+def gen_number():
```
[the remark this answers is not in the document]
HaoYang670: Thank you @andygrove . Really nice reminder! I will add these and other types supported by Spark in following PRs.

HaoYang670:
build

integration_tests/README.md
```diff
@@ -378,6 +378,15 @@ As an example, here is the `spark-submit` command with the cudf_udf parameter on
 $SPARK_HOME/bin/spark-submit --jars "rapids-4-spark_2.12-22.06.0-SNAPSHOT.jar,cudf-22.06.0-SNAPSHOT-cuda11.jar,rapids-4-spark-tests_2.12-22.06.0-SNAPSHOT.jar" --conf spark.rapids.memory.gpu.allocFraction=0.3 --conf spark.rapids.python.memory.gpu.allocFraction=0.3 --conf spark.rapids.python.concurrentPythonWorkers=2 --py-files "rapids-4-spark_2.12-22.06.0-SNAPSHOT.jar" --conf spark.executorEnv.PYTHONPATH="rapids-4-spark_2.12-22.06.0-SNAPSHOT.jar" ./runtests.py --cudf_udf
 ```
 
+### Enabling fuzz tests
+
+Fuzz tests are intended to find more corner cases in testing. We disable them by default because they might randomly fail. 
+The tests can be enabled by appending th option `--fuzz_test` to the command.
```
revans2, line 384: nit: You have "th" instead of "the"

The tests can be enabled by appending the option `--fuzz_test` to the command.

integration_tests/README.md
```diff
@@ -378,6 +378,15 @@ As an example, here is the `spark-submit` command with the cudf_udf parameter on
...
+
+   * `--fuzz_test` (enable the fuzz tests when provided, and remove this option if you want to disable the tests)
+
+To reproduce an error appearing in the fuzz tests, you also need to add the flag `--debug_tmp_path` to save the test data.
```
revans2, line 388: Could we combine these two together? So if you pass in `--fuzz_test_debug_path` or something like that the tests are enabled? To be clear I am fine with keeping the debug_tmp_path as it is. If this is really hard you don't need to do it.

integration_tests/src/main/python/json_fuzz_test.py
```diff
@@ -0,0 +1,419 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# A JSON generator built based on the context free grammar from https://www.json.org/json-en.html
+
+from cgi import test
```
revans2, line 17: Why do we need this? We are not doing CGI are we?

integration_tests/README.md
HaoYang670: Done!

nartal1 (approved):
LGTM but would be good to get review from others who had reviewed earlier.

HaoYang670:
build

sameerz:
Does this close issue https://github.com/NVIDIA/spark-rapids/issues/4138 ?

HaoYang670:
> Does this close issue #4138 ?

Yes. Have updated.

HaoYang670:
Hi @revans2. Do you think we could merge this PR or not?

revans2:
@HaoYang670 I am very busy with other things. Please don't let me block this from going in. If others have approved it then we can merge it in and if there are issues we can iterate on them as we find them.

andygrove (approved):

Issue #5003, opened 2022-03-22
[BUG] Cloudera 3.1.1 tests fail due to ClouderaShimVersion


tgravescs:
**Describe the bug**
```
[2022-03-18T07:41:31.843Z] FAILED integration_tests/src/main/python/arithmetic_ops_test.py::test_multiplication_fallback_when_ansi_enabled[Byte][ALLOW_NON_GPU(ProjectExec,Alias,Multiply,Cast)]
[2022-03-18T07:41:31.843Z] FAILED integration_tests/src/main/python/arithmetic_ops_test.py::test_multiplication_fallback_when_ansi_enabled[Short][ALLOW_NON_GPU(ProjectExec,Alias,Multiply,Cast)]
[2022-03-18T07:41:31.843Z] FAILED integration_tests/src/main/python/arithmetic_ops_test.py::test_multiplication_fallback_when_ansi_enabled[Integer][ALLOW_NON_GPU(ProjectExec,Alias,Multiply,Cast)]
[2022-03-18T07:41:31.843Z] FAILED integration_tests/src/main/python/arithmetic_ops_test.py::test_multiplication_fallback_when_ansi_enabled[Long][ALLOW_NON_GPU(ProjectExec,Alias,Multiply,Cast)]
[2022-03-18T07:41:31.843Z] FAILED integration_tests/src/main/python/arithmetic_ops_test.py::test_division_fallback_on_decimal[Decimal(38,21)][ALLOW_NON_GPU(ProjectExec,Alias,Divide,Cast,PromotePrecision,CheckOverflow)]
[2022-03-18T07:41:31.843Z] FAILED integration_tests/src/main/python/arithmetic_ops_test.py::test_division_fallback_on_decimal[Decimal(21,17)][ALLOW_NON_GPU(ProjectExec,Alias,Divide,Cast,PromotePrecision,CheckOverflow)]
```

The error:
```
  py4j.protocol.Py4JJavaError: An error occurred while calling z:com.nvidia.spark.rapids.ExecutionPlanCaptureCallback.assertDidFallBack.[0m
[...]

PR #5004, opened 2022-03-22
Add ClouderaShimVersion to unshimmed files

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/5003

This fixes test failures with error:
 ` : java.lang.LinkageError: loader constraint violation: when resolving method "com.nvidia.spark.rapids.shims.spark311cdh.SparkShimServiceProvider$.VERSION()Lcom/nvidia/spark/rapids/ClouderaShimVersion;" the class loader (instance of org/apache/spark/util/MutableURLClassLoader) of the current class, com/nvidia/spark/rapids/shims/spark311cdh/SparkShimServiceProvider, and the class loader (instance of org/apache/spark/util/MutableURLClassLoader) for the method's defining class, com/nvidia/spark/rapids/shims/spark311cdh/SparkShimServiceProvider$, have different Class objects for the type com/nvidia/spark/rapids/ClouderaShimVersion used in the signature�[0m`

We did the same thing with the Databricks shim version in earlier PR.

### DIFF
--- a/dist/unshimmed-common-from-spark311.txt
+++ b/dist/unshimmed-common-from-spark311.txt
@@ -6,6 +6,7 @@ com/nvidia/spark/ExclusiveModeGpuDiscoveryPlugin*
 com/nvidia/spark/RapidsUDF*
 com/nvidia/spark/SQLPlugin*
 com/nvidia/spark/rapids/ColumnarRdd*
+com/nvidia/spark/rapids/ClouderaShimVersion*
 com/nvidia/spark/rapids/DatabricksShimVersion*
 com/nvidia/spark/rapids/ExecutionPlanCaptureCallback*
 com/nvidia/spark/rapids/ExplainPlan*


tgravescs:
build

jlowe (approved):

gerashegalov:
I would like to emphasize the need for consistency because it avoids dealing with such problems one by one. I think ShimVersion and all its descendant classes should be unshimmed together. It should be no issue  to unshim them because they are unrelated to the Spark codebase.

tgravescs:
I understand what you are saying but I'm not sure if you are proposing changing  it here or really what exactly you are proposing to change?  At this point all the ShimVersions are included here, perhaps you are saying if we can include something like com/nvidia/spark/rapids/*ShimVersion* and hope that the name doesn't conflict with any other classes?

Either way since this should fix issue and we are in burn down lets decide and file an issue to change in 22.06 if we want.

gerashegalov:
> I understand what you are saying but I'm not sure if you are proposing changing it here or really what exactly you are proposing to change? At this point all the ShimVersions are included here, perhaps you are saying if we can include something like com/nvidia/spark/rapids/_ShimVersion_ and hope that the name doesn't conflict with any other classes?
> 

Yes, my suggestion boils down to having 
SparkShimVersion, ClouderaShimVersion, DatabricksShimVersion, EMRShimVersion, ShimVersion in the unshimmed.txt. 

So IIUC we still have EMRShimVersion shimmed even after this PR.

> Either way since this should fix issue and we are in burn down lets decide and file an issue to change in 22.06 if we want.

We can handle this inconsistency in a follow-up issue.

gerashegalov (approved):
LGTM, please file an issue for EMRShimVersion if you merge as is

tgravescs:
we don't have a EMR shim in the code anymore, so really could just remove that, I can file followup

Issue #5005, opened 2022-03-22
Remove unnecessary shims after Spark 3.1.x is the minimum supported version


jlowe:
#4988 removed support for Spark 3.0.x but left the shimmed interfaces in place.  This issue tracks removing shim interfaces for code that is now common across all supported Spark versions.

gerashegalov:
this issue probably subsumes #4304

PR #5012, opened 2022-03-22
Unshim GpuOrcScan and GpuParquetScan [databricks]

jlowe:
This is part of #5005.  This unshims `GpuOrcScan` and `GpuParquetScan` which now each only have one implementation in the shim layer.


jlowe:
build

revans2 (approved):

PR #5013, opened 2022-03-22
Unshim GpuRowBasedScalaUDF [databricks]

jlowe:
This is part of #5005.  This unshims `GpuRowBasedScalaUDF` which only had one shim implementation.


jlowe:
build

Issue #4808, opened 2022-02-17
[FEA] Investigate Spark UI tab with profiling tool like info


tgravescs:
**Is your feature request related to a problem? Please describe.**
Investigate if we can add our own Spark UI tab that does similar things to the profiling tool.  Basically prints interesting information for GPU. 
It may also be interesting to see if we could put a compare option on there, where it had a text box to compare against another application.  We should be able to use the rest api to get that information as long as user has permissions.

PR #5032, opened 2022-03-23
[dummy] Simple integration with Spark history server

amahussein:
Sharing POC of integrating rapids profiling tool with SHS.

This is part of #4808


abellina:
I know this is just a draft to get things going, do you have a screenshot of what this looks like?

amahussein:
> I know this is just a draft to get things going, do you have a screenshot of what this looks like?

That was a very simple step. Add a tab to display environment information.
The example here extracts the RAPIDS Jar and put them in a separate table.

The next step for me is when I get more familiar with the Profiling-tool:

- Aside from the "compare-mode", enumerate any information that cannot be visualized by the current SparkUI tab. For example, do we have custom metrics or aggregations that we need to hook into the SHS tabs?
- I believe that running the tool from cmdline is still valuable. So, on the long run I should refactor the code to support both modes (SHS and cmdLine).
- I will add the compare-mode once I have a single App running smoothly. My intuition is that the compare mode will be easier to implement once we have all the entries stored into the SHS KVStore. Once that is done, Comparing two applications will require pulling the information from the store for comparison.

gerashegalov:
I think it might be more useful to display summarized info such as spark-rapids and cudf jar versions in the RapidsEnv tab instead of repeating the jars in the classpath you can already see in the environment tab. And add some key active configurations with links to our docs, e.g. Shuffle Manager configured, UCX is on, PCBS is used, ...

Issue #5023, opened 2022-03-23
[BUG] When+sequence could trigger "Illegal sequence boundaries" error


viadea:
Below is a minimum repro in pyspark:

**Env:**
22.02GA jars
or
22.04Snapshot jars(20220322)

Reproduced on Databricks 9.1ML GPU and also local workstation.

**Repro**
```
data = [
 ("James,,Smith",["Java","Scala","C++"],["Spark","Java"],"OH","CA"),
 ("Michael,Rose,",["Spark"],["Spark","Java"],"NY","NJ"),
 ("Robert,,Williams",["CSharp","VB"],[],"UT","NV")
]

from pyspark.sql.types import StringType, ArrayType,StructType,StructField
schema = StructType([ 
    StructField("name",StringType(),True), 
    StructField("languagesAtSchool",ArrayType(StringType()),True), 
    StructField("languagesAtWork",ArrayType(StringType()),True), 
    StructField("currentState", StringType(), True), 
    StructField("previousState", StringType(), True)
  ])

df = spark.createDataFrame(data=data,schema=schema)
df.write.format("parquet").mode("overwrite").save("/tmp/testparquet")
df2 = spark.read.parquet("/tmp/testparquet")
df2.show()

spark.conf.set("spark.rapids.sql.enabled",True)
df2.withColumn("new_column", when (  (size(col("languagesAtWork")) > 0 ) , sequence(lit(1), size(col("languagesAtWork")),lit(1))) ).show()
```

**Error:**
```
java.lang.IllegalArgumentException: requirement failed: Illegal sequence boundaries: step &gt; 0 but start &gt; stop
```

**Full stacktrace:**
```
Caused by: java.lang.IllegalArgumentException: requirement failed: Illegal sequence boundaries: step > 0 but start > stop
	at scala.Predef$.require(Predef.scala:281)
	at org.apache.spark.sql.rapids.GpuSequenceUtil$.$anonfun$checkSequenceInputs$5(collectionOperations.scala:558)
	at org.apache.spark.sql.rapids.GpuSequenceUtil$.$anonfun$checkSequenceInputs$5$adapted(collectionOperations.scala:557)
	at com.nvidia.spark.rapids.Arm.withResource(Arm.scala:28)
	at com.nvidia.spark.rapids.Arm.withResource$(Arm.scala:26)
	at org.apache.spark.sql.rapids.GpuSequenceUtil$.withResource(collectionOperations.scala:533)
[...]

Issue #5041, opened 2022-03-24
[BUG] Implement hasSideEffects for all expressions that have side-effects


andygrove:
**Describe the bug**
In https://github.com/NVIDIA/spark-rapids/issues/3849 we implemented special handling in CASE WHEN for expressions with side-effects but this relies on all expressions with side-effects overriding the `hasSideEffects` method and we have not done this for all expressions (see https://github.com/NVIDIA/spark-rapids/issues/5023 for one example).

We should audit all expressions and implement `hasSideEffects` where appropriate.

Candidates:

- [x] GpuGetArrayItem - 
- [x] GpuElementAt
- [x]  GpuGetMapValue
- [x] GpuUnaryMinus
- [x] GpuAbs
- [ ] GpuSum

**Steps/Code to reproduce bug**
N/A

**Expected behavior**
Conditional expressions such as `IF` and `CASE` should not fail due to expressions with side-effects being evaluated and throwing exceptions.

**Environment details (please complete the following information)**
N/A

**Additional context**
N/A

revans2:
I just did a test with pure spark 3.2.0 and I can confirm that case/when does not stop a SUM in ansi from throwing an exception if the case/when is outside of the SUM.

ANSI DISABLED:
```
scala>  Seq((0L, Long.MaxValue), (0L, 100L), (1L, 100L)).toDF.groupBy("_1").agg(when(col("_1") === 1, sum("_2")).otherwise(0).alias("result")).show()
+---+------+
| _1|result|
+---+------+
|  0|     0|
|  1|   100|
+---+------+
```

ANSI ENABLED:
```
scala>  Seq((0L, Long.MaxValue), (0L, 100L), (1L, 100L)).toDF.groupBy("_1").agg(when(col("_1") === 1, sum("_2")).otherwise(0).alias("result")).show()
22/04/15 19:08:34 ERROR Executor: Exception in task 0.0 in stage 10.0 (TID 29)
java.lang.ArithmeticException: long overflow
	at java.lang.Math.addExact(Math.java:809)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.agg_doConsume_0$(Unknown Source)
...
```

If I move the case/when inside the SUM, then it does the right thing in both places...
```
[...]

PR #5042, opened 2022-03-24
Implement hasSideEffects for GpuSequence

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/5023

This PR implements `hasSideEffects` for `GpuSequence` to fix a bug where a conditional expression such as `CASE .. WHEN` could result in `GpuSequence` throwing an exception because we were evaluating for all rows and not just the rows that met the condition.

I filed a follow on issue https://github.com/NVIDIA/spark-rapids/issues/5041 to do the same for all other expressions that could potentially have side-effects.


jlowe (approved):

andygrove:
build

pxLi:
build

Issue #4908, opened 2022-03-07
Update spark 2.x explain code for 22.04


tgravescs:
**Describe the bug**
Update the spark 2 code based on changes in 22.04 near when we freeze.

PR #5062, opened 2022-03-25
Update Spark 2.x explain API with changes in 22.04

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/4908

This updates all the Spark 2.x explain api code to pull over checks from the changes to the 22.04 code.

I updated diff scripts and checks all the Overrides to try to catch anything new added.

I tested using the NDS queries. I diff'd the new output of 22.04 vs the 22.02 explain api output and the main differences now seem to be with ExistenceJoin being supported and then LeftSemi join. Unfortunately I also found a case the Spark 2.x planner inserts EqualNullSafe, which we don't support on the GPU, but in  Spark 3.x it doesn't generate that, it uses other expressions so in Spark 3.x it does say it will run on the GPU.  I documented that one.
I also diff'd the results vs explain on 3.1 and didn't see anything we missed.


tgravescs:
build

Issue #5049, opened 2022-03-25
Premerge CI should automatically run on Databricks if code touches Databricks shims


jlowe:
We currently have the ability to opt-in for Databricks runs in premerge CI with the `[databricks]` tag, but the developer has to remember to use it.  It would be nice if premerge automatically ran on Databricks if there are any changes to files in the Databricks shims, i.e.: code that only builds on Databricks instances.  Currently this would apply to any PR that has file paths matching against these regexp patterns:
- `sql-plugin/src/main/.*[0-9x-]db/`
- `databricks`

pxLi:
nice, this is on my TODO list. I will enable the auto-trigger of above patterns later

pxLi:
https://github.com/NVIDIA/spark-rapids/pull/5067

Issue #4920, opened 2022-03-09
There are references to non-existent shims dir that needs to be updated


razajafri:
```
git grep '[^/]shims/'
CONTRIBUTING.md:There is a known issue with the shims/spark3xx submodules. After being enabled once, a module such as shims/spark312
docs/dev/shims.md:(i.e., in `sql-plugin/src/main/scala` as opposed to be duplicated either in `shims/spark3xx` submodules or over
jenkins/Jenkinsfile-blossom.premerge:                                                                      'sql-plugin/src/main/scala/,shims/spark311/src/main/scala/,' +
jenkins/Jenkinsfile-blossom.premerge:                                                                      'shims/spark301db/src/main/scala/,shims/spark301/src/main/scala/,' +
jenkins/Jenkinsfile-blossom.premerge:                                                                      'shims/spark302/src/main/scala/,shims/spark303/src/main/scala/,' +
jenkins/Jenkinsfile-blossom.premerge:                                                                      'shims/spark304/src/main/scala/,shims/spark312/src/main/scala/,' +
jenkins/Jenkinsfile-blossom.premerge:                                                                      'shims/spark313/src/main/scala/'
```

_Originally posted by @gerashegalov in https://github.com/NVIDIA/spark-rapids/issues/4857#issuecomment-1062160432_

razajafri:
@pxLi is assigned to an issue #4991 which should handle the references to non-existent `shims/spark*` folders. I will address the changes in `.md` file

razajafri:
I am limiting this issue to only address the references in `.md` files 

The jacoco references in Jenkinsfile-blossom.premerge will be addressed by #4991

pxLi:
https://github.com/NVIDIA/spark-rapids/pull/5067

PR #5067, opened 2022-03-28
Premerge databricks CI autotrigger [skip ci]

pxLi:
fix https://github.com/NVIDIA/spark-rapids/issues/5049 https://github.com/NVIDIA/spark-rapids/issues/4920 

verified in forked repo. This change will take effect after merge

### DIFF
--- a/jenkins/Jenkinsfile-blossom.premerge
+++ b/jenkins/Jenkinsfile-blossom.premerge
@@ -44,20 +44,6 @@ def PREMERGE_CI_2_ARGUMENT // argument for 'spark-premerge-build.sh' running fro
 def TestResult1 // keep ci_1 pytest result files that's in JUnit-XML style
 def TestResult2 // keep ci_2 pytest result files that's in JUnit-XML style
 
-def sourcePattern30X =  'shims/spark301db/src/main/scala/,' +
-    'shims/spark301/src/main/scala/,' +
-    'shims/spark302/src/main/scala/,' +
-    'shims/spark303/src/main/scala/,' +
-    'shims/spark304/src/main/scala/'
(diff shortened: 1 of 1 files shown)


jenkins/Jenkinsfile-blossom.premerge
```diff
@@ -685,3 +654,34 @@ String stashPytestResult(String name) {
...
+    )
+
+    stash(name: "source_tree", includes: "**,.git/**", useDefaultExcludes: false)
+}
+
+boolean databricksCodeChanged() {
+    def output = sh(script: '''
+            # get merge BASE from merged pull request. Log message e.g. "Merge HEAD into BASE"
+            BASE_REF=$(git --no-pager log --oneline -1 | awk '{ print $NF }')
+            git --no-pager diff --name-only ${BASE_REF} HEAD | grep -E 'sql-plugin/src/main/.*[0-9x-]db/|databricks || true'
```
jlowe, line 679: The single-quotes on the grep argument are grabbing more than desired.  In addition we don't need the actual lines that matched, just whether or not something matched.
```suggestion
            git --no-pager diff --name-only ${BASE_REF} HEAD | grep -lE 'sql-plugin/src/main/.*[0-9x-]db/|databricks' || true
```

pxLi: Good catch, thanks! I was intentionally to list the changed files so we could print out the matched lines (https://github.com/NVIDIA/spark-rapids/pull/5067/files#diff-6fcadc51af71fb51e6c978181ec240c5497d31a3af5c79d8baecaeea278b4fdaR683) to help developer understand why the CI got triggered w/ databricks build. But yeah, we may not need that until some debug scenarios actually occur.

I will refactor this part

pxLi:
build

Issue #4986, opened 2022-03-18
Update ide code style for continuation indentation


tgravescs:
**Describe the bug**
It looks like https://github.com/NVIDIA/spark-rapids/blob/branch-22.04/docs/dev/idea-code-style-settings.xml#L28 is  using 4-space continuation indent but each level of indentation in scala style is 2 spaces.
https://docs.scala-lang.org/style/indentation.html

I think the only 4 space indentation is for parameters. (https://github.com/databricks/scala-style-guide#indent)

I'm not sure the specifics it might be if we change that to 2 it breaks the parameter indentation, might be something to pay attention to

PR #5077, opened 2022-03-28
Change Scala style continuation indentation to be 2 spaces to match guide [skip ci]

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/4986

The scala style guide calls for 2 spaces of indentation at each level (https://docs.scala-lang.org/style/indentation.html).
Only exception to this that I know of is method declaration (https://github.com/databricks/scala-style-guide#indent).  So here change our code style template to use the default 2 spaces for continuation identation.  We can just remove the explicit 4 spaces and it goes back to default to 2.


tgravescs:
build

jlowe (approved):

Issue #5035, opened 2022-03-24
[BUG] Different CSV parsing behavior between 22.04 and 22.02


wbo4958:
I have a test.csv file with below content
``` console
0.0
1.0
2.0
```

I got different results between 22.04 and 22.02. 

Below test code is copied from https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/classification/Classifier.scala#L142, if we have different result, then some ML application will fail on 22.04 version
  
## 22.04 version

``` scala
scala>     import org.apache.spark.sql.functions._
import org.apache.spark.sql.functions._

scala>     import org.apache.spark.sql.types._
import org.apache.spark.sql.types._

scala>     import org.apache.spark.sql.{Row}
import org.apache.spark.sql.Row

scala>     val schema = new StructType(Array(StructField("class", DoubleType, true)))
schema: org.apache.spark.sql.types.StructType = StructType(StructField(class,DoubleType,true))

scala>     val rawInput = spark.read.schema(schema).csv("/home/bobwang/tmp/max/test.csv")
rawInput: org.apache.spark.sql.DataFrame = [class: double]

scala>     val maxLabelRow: Array[Row] = rawInput.select(max("class")).take(1)
maxLabelRow: Array[org.apache.spark.sql.Row] = Array([1.9999999999999998])

scala>     val maxDoubleLabel: Double = maxLabelRow.head.getDouble(0)
maxDoubleLabel: Double = 1.9999999999999998

scala>     maxDoubleLabel.toInt + 1
res3: Int = 2

```

## 22.02

``` scala

scala>     import org.apache.spark.sql.functions._
import org.apache.spark.sql.functions._

scala>     import org.apache.spark.sql.types._
import org.apache.spark.sql.types._

scala>     import org.apache.spark.sql.{Row}
import org.apache.spark.sql.Row

scala>     val schema = new StructType(Array(StructField("class", DoubleType, true)))
schema: org.apache.spark.sql.types.StructType = StructType(StructField(class,DoubleType,true))

scala>     val rawInput = spark.read.schema(schema).csv("/home/bobwang/tmp/max/test.csv")
rawInput: org.apache.spark.sql.DataFrame = [class: double]

scala>     val maxLabelRow: Array[Row] = rawInput.select(max("class")).take(1)
maxLabelRow: Array[org.apache.spark.sql.Row] = Array([2.0])

[...]

PR #5078, opened 2022-03-28
Disable JSON and CSV floating-point reads by default

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/5035

This PR reverts recent changes that enabled float/double/decimal reads by default for the CSV and JSON readers. The motivation for this change is that we are not bit-for-bit compatible with Spark when reading floating-point types.


andygrove:
build

andygrove:
build

pxLi:
build

revans2 (approved):

PR #5079, opened 2022-03-29
Disable spark snapshot shims pre-merge build in 22.04

pxLi:

disable some shims build for spark snapshot versions in pre-merge.
**NOTE:** after merge, we will need to revert the auto-merged commit in branch-22.06

nightly build and tests have be taken care of internally.

### DIFF
--- a/jenkins/spark-premerge-build.sh
+++ b/jenkins/spark-premerge-build.sh
@@ -43,17 +43,20 @@ mvn_verify() {
     env -u SPARK_HOME mvn -U -B $MVN_URM_MIRROR -Dbuildver=311cdh clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -Dmaven.scalastyle.skip=true -Dcuda.version=$CUDA_CLASSIFIER -pl aggregator -am
     env -u SPARK_HOME mvn -U -B $MVN_URM_MIRROR -Dbuildver=312 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -Dmaven.scalastyle.skip=true -Dcuda.version=$CUDA_CLASSIFIER -pl aggregator -am
     env -u SPARK_HOME mvn -U -B $MVN_URM_MIRROR -Dbuildver=313 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -Dmaven.scalastyle.skip=true -Dcuda.version=$CUDA_CLASSIFIER -pl aggregator -am
-    env -u SPARK_HOME mvn -U -B $MVN_URM_MIRROR -Dbuildver=314 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -Dmaven.scalastyle.skip=true -Dcuda.version=$CUDA_CLASSIFIER -pl aggregator -am
+    # skip spark snapshot shims build in 22.04. (TODO) revert this commit in 22.06 after auto-merged
+    # env -u SPARK_HOME mvn -U -B $MVN_URM_MIRROR -Dbuildver=314 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -Dmaven.scalastyle.skip=true -Dcuda.version=$CUDA_CLASSIFIER -pl aggregator -am
 
     # don't skip tests
     env -u SPARK_HOME mvn -U -B $MVN_URM_MIRROR -Dbuildver=320 clean install -Drat.skip=true -Dmaven.javadoc.skip=true -Dskip -Dmaven.scalastyle.skip=true -Dcuda.version=$CUDA_CLASSIFIER -Dpytest.TEST_TAGS='' -pl '!tools'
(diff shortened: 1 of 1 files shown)


pxLi:
build

jenkins/spark-premerge-build.sh
```diff
@@ -43,17 +43,20 @@ mvn_verify() {
     env -u SPARK_HOME mvn -U -B $MVN_URM_MIRROR -Dbuildver=311cdh clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -Dmaven.scalastyle.skip=true -Dcuda.version=$CUDA_CLASSIFIER -pl aggregator -am
     env -u SPARK_HOME mvn -U -B $MVN_URM_MIRROR -Dbuildver=312 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -Dmaven.scalastyle.skip=true -Dcuda.version=$CUDA_CLASSIFIER -pl aggregator -am
     env -u SPARK_HOME mvn -U -B $MVN_URM_MIRROR -Dbuildver=313 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -Dmaven.scalastyle.skip=true -Dcuda.version=$CUDA_CLASSIFIER -pl aggregator -am
-    env -u SPARK_HOME mvn -U -B $MVN_URM_MIRROR -Dbuildver=314 clean install -Drat.skip=true -DskipTests -Dmaven.javadoc.skip=true -Dskip -Dmaven.scalastyle.skip=true -Dcuda.version=$CUDA_CLASSIFIER -pl aggregator -am
+    # skip spark snapshot shims build in 22.04. (TODO) revert this commit in 22.06 after auto-merged
```
gerashegalov, line 46: Let us formalize when the snapshots build are disabled so we don't have to do it over and over manually by commenting and uncommenting. This could be an environment variable or a dedicated configuration file in the repo. 

Maybe we can reuse .github/workflows/auto-merge.yml ? Whether the snapshot builds are needed is directly related to auto-merge as this comment indicates. Could we  make the current branch name's equality to the `BASE` in .github/workflows/auto-merge.yml

```bash
BASE_BRANCH=$(grep 'BASE: ' .github/workflows/auto-merge.yml | grep -o branch-.*)
CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)

if [[ "$CURRENT_BRANCH" == "$BASE_BRANCH" ]]; then
  # build snapshots
  ...
fi
```

pxLi: If so, we could simply based on https://github.com/NVIDIA/spark-rapids/blob/branch-22.06/jenkins/version-def.sh#L31 instead of a workflow file. let me refactor this

pxLi:
build

pxLi:
build

jenkins/spark-premerge-build.sh
```diff
@@ -121,6 +121,13 @@ nvidia-smi
 
 . jenkins/version-def.sh
 
+BUILD_SNAPSHOTS="false"
+PREMERGE_PROFILES="-PnoSnapshots,pre-merge"
+if [[ ${PROJECT_VER} =~ ^22\.06\. ]]; then # enable snapshot builds for active development branch only
```
gerashegalov, line 126: The proposal to reuse auto-merge has the advantage is that it makes auto-merge config the single source of truth. And as part of our standard practice we never forget to update it .
 
With this solution there will be two places to update when the current branch enters the burn-down state. On the other hand it gives us the flexibility to manage auto-merge and BUILD_SHIMS independently but in a simplified form. I am not sure we have a need for this flexibility.

pxLi: > And as part of our standard practice we never forget to update it .

Actually version-def.sh is the a source of truth for testing. auto-merge BASE could probably be following other naming patterns that do not reflect versions

tgravescs:
I agree it would be nice to have one mechanism, although it would also be nice to combine build scripts as well.  We have build all and then we have separate things for premerge and nightly. That might be bigger changes though.
Talking about snapshots we also have the fact we don't package them in the release version and we should switch to do the same here.  We already have snapshot versions in the dist pom:

```
<snapshot.buildvers>
            314,
            322,
            330
</snapshot.buildvers>
```
can we reuse that.  If we can minimize where we have to change in adding/removing snapshots that would be ideal.
 I'm not sure off the top of my head where all version_def.sh is used, could it use maven commands or is it run with integration tests?.  if it can be the source of truth and used by both that would be fine with me.  Although I see a separate version-def.sh in the blossom-jenkins repo as well.  

Perhaps we need a short term solution for this and then longer term bigger change.

pxLi:
> I'm not sure off the top of my head where all version_def.sh is used

It has been used in all jenkins build and test scripts to simply export helpful ENVs

> Although I see a separate version-def.sh in the blossom-jenkins repo as well

That file was for tests that do not need to clone the plugin github repo (just need some ENVs), usually used in pipelines which just required to download artifacts and run tests.

> can we reuse that. If we can minimize where we have to change in adding/removing snapshots that would be ideal.
> Perhaps we need a short term solution for this and then longer term bigger change

@tgravescs yes, created a ticket to track https://github.com/NVIDIA/spark-rapids/issues/5086. For now, I think we can leverage version-de.sh here. Please let me know if you have any other concern, thanks!

tgravescs:
> I think we can leverage version-de.sh here. Please let me know if you have any other concern, thanks!

What exactly do you mean by this?  Adding snapshot versions to version-def.sh? is there other changes to this PR to do that?

tgravescs (approved):

PR #5087, opened 2022-03-30
Add support for regexp_replace with back-references

### LINKED ISSUE #4557: [FEA] Add support for regexp_replace with back-references
Once https://github.com/NVIDIA/spark-rapids/pull/4556 is merged, we will fall back to CPU for `regexp_replace` with back-references. This is issue is for supporting this feature on GPU.

Support `regexp_replace` with back-references on the GPU.

None

**Additional context**
None

sperlingxx:

Closes #4557

Support for regexp_replace with back-references. 
In spark-rapids, we transform replacement pattern from Spark style to cuDF style, like: `\1` `$1` -> `${1}`.
In cuDF, we adjust the valid range of group index to accommodate Spark.  So, current PR relies on cuDF [PR](https://github.com/rapidsai/cudf/pull/10530)


sperlingxx:
build

integration_tests/src/main/python/string_test.py
```diff
@@ -498,16 +498,26 @@ def test_re_replace():
                 'REGEXP_REPLACE(a, "TEST", NULL)'),
         conf=_regexp_conf)
 
-@allow_non_gpu('ProjectExec', 'RegExpReplace')
 def test_re_replace_backrefs():
-    gen = mk_str_gen('.{0,5}TEST[\ud720 A]{0,5}')
-    assert_gpu_fallback_collect(
+    gen = mk_str_gen('.{0,5}TEST[\ud720 A]{0,5}TEST')
```
andygrove, line 502: Why is the test modified to remove the `TEST` at the end of the generating string?

integration_tests/src/main/python/string_test.py
```diff
@@ -498,16 +498,26 @@ def test_re_replace():
...
+    assert_gpu_and_cpu_are_equal_collect(
         lambda spark: unary_op_df(spark, gen).selectExpr(
```
andygrove, line 504: Could you add the test case `'REGEXP_REPLACE(a, "(TEST)", "$1")'` since this fails for me locally

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -773,24 +773,44 @@ case class GpuLike(left: Expression, right: Expression, escapeChar: Char)
...
+      // match $group_index or \group_index
+      if (Seq('$', '\\').contains(rep.charAt(i))
+        && i + 1 < rep.length && rep.charAt(i + 1).isDigit) {
+
+        b.append("${")
+        var j = i + 1
+        while (j + 1 < rep.length && rep.charAt(j).isDigit) {
+          b.append(rep.charAt(j))
+          j += 1
+        }
```
andygrove, line 799: This code seems to be incorrect and does not handle the case where the pattern ends with `$1` because `j` would the index of the `1` and `j+1 < rep.length` would be false so it does not append the `1` to the cuDF pattern and produces `${}` instead.

sperlingxx: Yes, it's a blunder, which leads to the failure of test cases like `'REGEXP_REPLACE(a, "(TEST)", "$1")'`. 
I have fixed it.

sperlingxx:
build

andygrove:
build

andygrove (approved):

PR #5093, opened 2022-03-30
Fix leak when GDS buffer store closes

jlowe:
Fixes a buffer leak in the GDS buffer store where the `BatchSpiller` was creating a CuFile buffer and never releasing it.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -49,6 +49,11 @@ class RapidsGdsStore(
         singleShotSpill(other, deviceBuffer)
       }
     }
+  }
+
+  override def close(): Unit = {
+    super.close()
+    batchSpiller.close()
   }
 
   abstract class RapidsGdsBuffer(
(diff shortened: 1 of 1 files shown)


jlowe:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
```diff
@@ -120,14 +125,20 @@ class RapidsGdsStore(
...
+  class BatchSpiller() extends AutoCloseable {
     private val blockSize = 4096
     private[this] val spilledBuffers = new ConcurrentHashMap[File, Set[RapidsBufferId]]
     private[this] val pendingBuffers = ArrayBuffer.empty[RapidsGdsBatchedBuffer]
     private[this] val batchWriteBuffer = CuFileBuffer.allocate(batchWriteBufferSize, true)
     private[this] var currentFile = TempSpillBufferId().getDiskPath(diskBlockManager)
     private[this] var currentOffset = 0L
 
+    override def close(): Unit = {
+      pendingBuffers.foreach(_.free())
```
abellina, line 137: I am no expert here. `spill` has a synchronized around it. Should we synchronize `close`? and maybe throw if we try to spill from a closed batch spiller?

jlowe: In general we don't synchronize on closing buffer stores.  If you're trying to close a store while it's actively trying to spill then there's larger-scale issues that would need to be solved that adding `synchronize` will not solve.  It will only serialize the contention between `spill` and `close` but not guarantee that `spill` does not get invoked after `close`.

If we need to deal with trying to spill on a closed store IMO that's a larger scale issue that needs to be addressed in all the stores, not just this one.

abellina (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsGdsStore.scala
gerashegalov, line 137: Do we need a notion of safeFree to avoid leaking buffers when an  in-between free fails?

Do we need batchWriteBuffer.close() in a finally to prevent it from being leaked if  a `free` above throws?

jlowe: Filed #5098 to track hardening the code to potential exceptions from `RapidsBuffer.free`.

Issue #5091, opened 2022-03-30
[BUG] Profiling tool fails process custom task accumulators of type CollectionAccumulator


tgravescs:
**Describe the bug**
The profiling tool can fail to process an event log if it has custom task accumulables of type CollectionAccumulator

The code assumes it can do a toString.toLong but this is an array type and it fails:

```
java.lang.NumberFormatException: For input string: "[]"
        at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.lang.Long.parseLong(Long.java:589)
        at java.lang.Long.parseLong(Long.java:631)
        at scala.collection.immutable.StringLike.toLong(StringLike.scala:309)
        at scala.collection.immutable.StringLike.toLong$(StringLike.scala:309)
        at scala.collection.immutable.StringOps.toLong(StringOps.scala:33)
        at org.apache.spark.sql.rapids.tool.profiling.EventsProcessor.$anonfun$doSparkListenerTaskEnd$4(EventsProcessor.scala:197)
        at org.apache.spark.sql.rapids.tool.profiling.EventsProcessor.$anonfun$doSparkListenerTaskEnd$4$adapted(EventsProcessor.scala:197)
```

tgravescs:
I'll actually close this one with pr and open separate one to handle better

sameerz:
Closing for 22.04, will follow up on remaining tasks in issue #5101

PR #5094, opened 2022-03-30
Fix profiling tool reading collectionAccumulator

tgravescs:
The profiling tool can fail to process an event log if it has custom task accumulables of type CollectionAccumulator, this just changes it to catch the exception and ignore those accumulables.

This is a minimal fix for https://github.com/NVIDIA/spark-rapids/issues/5091 but I want to leave that open and handle these better.

### DIFF
--- a/tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/EventsProcessor.scala
+++ b/tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/EventsProcessor.scala
@@ -20,6 +20,7 @@ import java.util.concurrent.TimeUnit.NANOSECONDS
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
+import scala.util.control.NonFatal
 
 import com.nvidia.spark.rapids.tool.profiling._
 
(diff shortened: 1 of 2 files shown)


tgravescs:
build

tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/EventsProcessor.scala
```diff
@@ -404,8 +404,8 @@ class EventsProcessor(app: ApplicationInfo) extends EventProcessorBase[Applicati
         app.accumIdToStageId.put(res._2.id, event.stageInfo.stageId)
         arrBuf += thisMetric
       } catch {
-        case e: ClassCastException =>
-          logWarning("ClassCastException when parsing accumulables for task " +
+        case e: Exception =>
```
gerashegalov, line 407: should we catch exceptions more exhaustively with matching on `Throwable`?

tgravescs: if it's an Error, I think we just want to fail as that could be something more than just related to the accumulable, what I actually probably should do is make this match on NonFatal(e)

tgravescs:
build

revans2 (approved):

tools/src/main/scala/org/apache/spark/sql/rapids/tool/profiling/EventsProcessor.scala
gerashegalov: +1 for NonFatal, it's consistent with similar usages in our code base

gerashegalov (approved):

PR #5097, opened 2022-03-30
Implement hasSideEffects for GpuGetArrayItem, GpuElementAt, GpuGetMapValue, GpuUnaryMinus, and GpuAbs

andygrove:
Addresses most of the issues in https://github.com/NVIDIA/spark-rapids/issues/5041 but the issue will be left open and the remaining work completed in 22.06

This PR overrides `hasSideEffects` for the following expressions so that we evaluate them safely in conditional expressions.

- GpuGetArrayItem
- GpuElementAt
- GpuGetMapValue
- GpuUnaryMinus
- GpuAbs

It may be easiest to review this PR one commit at a time since each commit has the test and the implementation change for a subset of these expressions.


andygrove:
build

revans2 (approved):

Issue #4151, opened 2021-11-18
[Audit][FEA][SPARK-36920] Enable GPU overrides for ANSI interval types in arithmetics expressions.


gerashegalov:
Enable ANSI interval types for ABS, DIV ...

res-life:
- [ ] Abs: abs(interval)
- [ ] UnaryMinus: -interval
- [ ] Add: interval + interval
- [ ] Subtract: interval - interval
- [ ] Positive: +interval

// 2 new added operators
- [ ] interval * number
- [ ] interval / number

PR #5105, opened 2022-03-31
Support multiplication on ANSI interval types

res-life:
Contributes https://github.com/NVIDIA/spark-rapids/issues/4151

Supports 
- Year-month interval * (byte, short, int, long, float, double)
- Day-time interval * (byte, short, int, long, float, double)

Note: not support decimal type because of CUDF Decimal128 does not handle overflow, like:
```
CUDF positive decimal * positive decimal can generate negative value.
val d1 = ColumnVector.fromDecimals(new java.math.BigDecimal("123456789012345678901234567890123456.11"))
val d2 = Scalar.fromDecimal(new java.math.BigDecimal("12345678901234567890123456789012345611"))
val r = d1.mul(d2)
println(r)

123456789012345678901234567890123456.11 * 12345678901234567890123456789012345611 = -1053265448675579329726886670761974900.55
```


res-life:
build

res-life:
build

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
```diff
@@ -0,0 +1,497 @@
...
+
+import org.apache.spark.sql.catalyst.expressions.{Expression, ImplicitCastInputTypes, NullIntolerant}
+import org.apache.spark.sql.types._
+
+object IntervalUtils extends Arm {
+
+  /**
+   * compute interval.toLong * num
+   */
+  def mulIntervalAndNum(interval: BinaryOperable, num: BinaryOperable): ColumnVector = {
```
firestarman, line 30: I am not sure, but did you try
```
interval.mul(num, INT64)
```
instead ?

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
```diff
@@ -0,0 +1,497 @@
...
+    }
+  }
+
+  /**
+   * throws exception if any value in `longCv` exceeds the int limits
+   */
+  def toIntWithCheck(longCv: ColumnVector): ColumnVector = {
+    withResource(longCv.castTo(DType.INT32)) { intResult =>
+      withResource(longCv.notEqualTo(intResult)) { notEquals =>
+        if (notEquals.any().getBoolean) {
```
firestarman, line 49: There is already an utils method in [BoolUtils](https://github.com/NVIDIA/spark-rapids/blob/branch-22.06/sql-plugin/src/main/scala/com/nvidia/spark/rapids/BoolUtils.scala#L48).
And the scalar returned from `notEquals.any()` is leaked.

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
```diff
@@ -0,0 +1,497 @@
...
+    // check (y != 0) && (r/y != x)
+    checkMultiplyOverflowSimpleImp(x, y, r)
+
+    // check (x == Long.MIN_VALUE && y == -1)
+    x match {
+      case scalar: Scalar =>
+        if (getLong(scalar) == Long.MinValue) { // x == Long.MIN_VALUE
+          withResource(Scalar.fromLong(-1L)) { negOne =>
+            withResource(y.equalTo(negOne)) { negOneBool =>
+              if (negOneBool.any().getBoolean) { // y = -1L
```
firestarman, line 101: There is already an utils method in [BoolUtils](https://github.com/NVIDIA/spark-rapids/blob/branch-22.06/sql-plugin/src/main/scala/com/nvidia/spark/rapids/BoolUtils.scala#L48).
And the scalar returned from `negOneBool.any()` is leaked.

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
```diff
@@ -0,0 +1,497 @@
...
+          }
+        }
+      case _ =>
+        // check (x == Long.MIN_VALUE && y == -1)
+        withResource(Scalar.fromLong(Long.MinValue)) { minLong =>
+          withResource(Scalar.fromLong(-1L)) { negOneLong =>
+            withResource(x.equalTo(minLong)) { minBool =>
+              withResource(y.equalTo(negOneLong)) { oneBool =>
+                withResource(minBool.add(oneBool)) { overflow =>
+                  if (overflow.any().getBoolean) {
```
firestarman, line 114: There is already an utils method in [BoolUtils](https://github.com/NVIDIA/spark-rapids/blob/branch-22.06/sql-plugin/src/main/scala/com/nvidia/spark/rapids/BoolUtils.scala#L48).
And the scalar returned from `overflow.any()` is leaked.
There are still some places to be updated for this issue.

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
```diff
@@ -0,0 +1,497 @@
...
+        // compute interval(long) * num
+        val longResultCv: ColumnVector = interval.mul(numOperable)
+
+        withResource(longResultCv) { longResult =>
+          // check overflow
+          IntervalUtils.checkMultiplyOverflow(interval, numOperable, longResult)
+
+          longResult.incRefCount()
+        }
+      case _: FloatType => // num is float
```
firestarman, line 461: Did you try
```
case FloatType | DoubleType => interval.mul(numOperable, DType.FLOAT64)
```
instead of separating the float and double ?

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
```diff
@@ -0,0 +1,497 @@
...
+   * @param rightLong long cv or scalar, will not be scalar if leftLong is scalar
+   * @param result    is the result of x * y
+   */
+  def checkMultiplyOverflow(leftLong: BinaryOperable, rightLong: BinaryOperable,
+      result: ColumnVector): Unit = {
+    leftLong match {
+      case leftVector: ColumnVector =>
+        checkMultiplyOverflowImp(rightLong, leftVector, result)
+      case _ =>
+        checkMultiplyOverflowImp(leftLong, rightLong.asInstanceOf[ColumnVector], result)
```
firestarman, line 75: How do we know `rightLong` is a column vector here ?

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
```diff
@@ -0,0 +1,497 @@
...
+
+  /**
+   * Check if the long result of long * long is overflow
+   * See Math.multiplyExact:
+   * r = x * y;
+   * if (y != 0) && (r/y != x) throw exception;
+   * if(x == Long.MIN_VALUE && y == -1) throw exception
+   *
+   * @param x long cv or scalar
+   * @param y long cv or scalar, will not be scalar if x is scalar
```
firestarman, line 87: y is declared to be a column vector in the below function, can not be a scalar.

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
```diff
@@ -0,0 +1,497 @@
...
+   * @param r is the result of x * y
+   */
+  def checkMultiplyOverflowImp(x: BinaryOperable, y: ColumnVector,
```
firestarman, line 90: Why do we have so many functions for the overflow check ? Could we merge them into single one ?

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
```diff
@@ -0,0 +1,497 @@
...
+   * Here is a little different with `Math.multiplyExact`,
+   * omits the `check x == Long.MIN_VALUE && y == -1`,
+   * Because of the `toIntWithCheck` will throw exception
+   * for the value of `Long.MIN_VALUE * -1` exceeds int limits
+   *
+   * @param x long cv or scalar
+   * @param y long cv or scalar, will not be scalar if x is scalar
+   * @param r is the result of x * y
+   */
+  def checkMultiplyOverflowSimpleImp(x: BinaryOperable, y: ColumnVector,
```
firestarman, line 167: The postfix `Imp`, personally , is not suitable here. Usually it is used in a class declaration for implementing a interface, or a concrete method declaration implementing a abstract method.

res-life:
Waiting #5150 to merge, has some conflict on `Spark33XShims`.

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
res-life: If `left` and `right` are both `Scalar` will invoke the following method.

For more detail, see `GpuBinaryExpression`
```
  override def doColumnar(numRows: Int, intervalScalar: GpuScalar,
      numScalar: GpuScalar): ColumnVector = {
    withResource(GpuColumnVector.from(intervalScalar, numRows, interval.dataType)) { expandedLhs =>
      doColumnar(expandedLhs, numScalar)
    }
  }
```

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
res-life: Updated, good idea.

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
res-life: done.

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
res-life: done

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
res-life: done

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
res-life: Updated the comments.

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
res-life: There are four types of checking,  seems can not merge to one.

**`day-time interval` * num**
The internal type of `day-time interval` is Long
The check of `r = x(long day-time  value) * y( (byte, short, int or long) value)` is:

```
   * r = x * y;
   * if (y != 0) && (r/y != x) throw exception;
   * if(x == Long.MIN_VALUE && y == -1) throw exception   
```

The check of r = x(long day-time  value) * y(double value) is:
```
See: com.google.common.math.roundToLong
   * r = x * y;
   * has no Inf, -Inf, NaN
   * z = roundIntermediate(x, mode)
   * checkInRange(MIN_LONG_AS_DOUBLE - z < 1.0 & z < MAX_LONG_AS_DOUBLE_PLUS_ONE)
   * return z.toLong
```



**`year-month interval` * num**
The internal type of `year-month interval` is int
The check of `r = x(int year-month value) * y( (byte, short, int or long) value)` is:

```
   * r = x * y;
   * if (y != 0) && (r/y != x) throw exception;
   * check (int) r == r 
```
Note:
~~if(x == Long.MIN_VALUE && y == -1) throw exception~~
No need to check the `if(x == Long.MIN_VALUE && y == -1) throw exception`, 
replaced by checking `(int) r == r`, because of expecting int value.
If (x,y) is (Long.MIN_VALUE, -1) or (-1, Long.MIN_VALUE) `check (int) r == r` will cover

The check of r = x(double value) * y(double value) is:
```
   * r = x * y;
   * has no Inf, -Inf, NaN
   * z = roundIntermediate(x, mode)
   * checkInRange(MIN_LONG_AS_DOUBLE - z < 1.0 & z < MAX_LONG_AS_DOUBLE_PLUS_ONE)
   * return z.toInt.    // Note this is toInt
```

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
res-life: Updated

res-life:
build

res-life:
build

res-life:
build

res-life:
build

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
```diff
@@ -0,0 +1,442 @@
...
+        checkMultiplyOverflowSimpleWithRightCv(leftLong, rightVector, result)
+      case _ =>
+        checkMultiplyOverflowSimpleWithRightCv(rightLong, leftLong.asInstanceOf[ColumnVector],
+          result)
+    }
+  }
+
+  /**
+   * Check if the long result of long * long is overflow
+   * See Math.multiplyExact:
```
jlowe, line 138: The overflow checks look pretty expensive to evaluate.  Would it be faster to perform a multiply with a cast to DECIMAL128 (only on the output, if possible) and then check if the upper 64-bits do not match the expected sign bit of the result (i.e.: sign-bit of left XOR sign-bit of right)?  This is something that could be investigated as a followup.

res-life:
build

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
res-life: Good idea.

I tested the new implementation:
`decimal128_ret = left_long.mul(right_long, DECIMAL128)`  
then check `decimal128_ret < min_long or decimal128_ret > max_long`
then `decimal128_ret.toLong`

|original implementation|new implementation|
|:-:|:-:|
|40ms|25ms|

`DECIMAL_128` fits the results of `Long.MaxValue * Long.MaxValue` and  `Long.MinValue * Long.MinValue`,
and the overflow check `decimal128_ret < min_long or decimal128_ret > max_long` is equavalent to the original implementation.
I think this implementation has no bug, and let's use this.


Details code:
```
  def genLongs(): Array[Long] = {
    val rows = 100 * 1024 * 1024
    val ret = new Array[Long](rows)
    var i = 0
    while (i < rows) {
      ret(i) = Long.MaxValue - i
      i += 1
    }
    ret
  }

  test("multiple decimal 128") {
    val l = genLongs
    val r = genLongs
    val lcv = ColumnVector.fromLongs(l: _*)
    val rcv = ColumnVector.fromLongs(r: _*)

    val start = System.currentTimeMillis()
    val dT = DType.create(DType.DTypeEnum.DECIMAL128, 0)
    withResource(lcv.mul(rcv, dT)) { ret =>
      withResource(Scalar.fromLong(Long.MaxValue)) { maxLong =>
        withResource(Scalar.fromLong(Long.MinValue)) { minLong =>
          withResource(ret.greaterThan(maxLong)) { gtm =>
            withResource(ret.lessThan(minLong)) { ltm =>
              if (BoolUtils.isAnyValidTrue(gtm) || BoolUtils.isAnyValidTrue(ltm)) {
                println("overflow occurs")
              }
              ret.castTo(DType.INT64)
            }
          }
        }
      }
    }
    println((System.currentTimeMillis() - start) + " ms")
  }

  test("multiple long") {
    val l = genLongs
    val r = genLongs
    val lcv = ColumnVector.fromLongs(l: _*)
    val rcv = ColumnVector.fromLongs(r: _*)

    val start = System.currentTimeMillis()
    withResource(lcv.mul(rcv)) { r =>
      // only print if overflow
      IntervalUtils.checkMultiplyOverflow(lcv, rcv, r)
      r
    }

    println((System.currentTimeMillis() - start) + " ms")
  }
```

res-life:
build

res-life:
build

firestarman (commented):
LGTM

jlowe (approved):

PR #5107, opened 2022-03-31
Fix GpuFileFormatDataWriter failing to stat file after commit

### LINKED ISSUE #5084: [BUG] GpuWriteTaskStatsTracker complains for all writes in local mode
**Describe the bug**
I am not sure 100% what is happening, but I see the following warning every time I write a file out in local mode on 22.04 or 22.06.
```
22/03/29 18:27:32 WARN GpuWriteTaskStatsTracker: Expected 1 files, but only saw 0. This could be due to the output format not writing empty files, or files being not immediately visible in the filesystem.
```

The files are not empty. I can see the file that is written.  The CPU does nothing like this either.

jlowe:
Fixes #5084.

#3345 moved the location of stats collection to after file commit, but with some file committers the file is renamed by the commit and therefore trying to stat the original path can fail.  #3345 simply ported a change from Spark, but the reason Spark didn't have the same issue is because it also had the changes from apache/spark@7f51106c0df190b3364848f8808e03105da63681.

This PR makes similar changes to the task stats write tracker to remove the unused `newBucket` method and adds the new `closeFile` method to handle updating file stats _before_ commit.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/BasicColumnarWriteStatsTracker.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/BasicColumnarWriteStatsTracker.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,9 +17,12 @@
 package org.apache.spark.sql.rapids
 
 import java.io.FileNotFoundException
+import java.nio.charset.StandardCharsets
+
+import scala.collection.mutable
 
 import org.apache.hadoop.conf.Configuration
-import org.apache.hadoop.fs.Path
+import org.apache.hadoop.fs.{FileSystem, Path}
 
 import org.apache.spark.{SparkContext, TaskContext}
 import org.apache.spark.internal.Logging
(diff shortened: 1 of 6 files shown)


jlowe:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/BasicColumnarWriteStatsTracker.scala
```diff
@@ -67,37 +70,86 @@ class BasicColumnarWriteTaskStatsTracker(
...
     val path = new Path(filePath)
     val fs = path.getFileSystem(hadoopConf)
+    getFileSize(fs, path)
+  }
+
+  /**
+   * Get the size of the file expected to have been written by a worker.
+   * This supports the XAttr in HADOOP-17414 when the "magic committer" adds
+   * a custom HTTP header to the a zero byte marker.
+   * If the output file as returned by getFileStatus > 0 then the length if
```
gerashegalov, line 80:
```suggestion
   * If the output file length as returned by getFileStatus > 0 then the length is
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/BasicColumnarWriteStatsTracker.scala
```diff
@@ -67,37 +70,86 @@ class BasicColumnarWriteTaskStatsTracker(
...
+      if (len > 0) {
+        return Some(len)
+      }
     } catch {
       case e: FileNotFoundException =>
-        // may arise against eventually consistent object stores
+        // may arise against eventually consistent object stores.
         logDebug(s"File $path is not yet visible", e)
-        None
+        return None
```
gerashegalov, line 97: If we split `getFileSize` method in two

```scala
def getFileSizeFromStatus(...): Option[Long] = ??? // no special handling for len==0 as before
def getFileSizeFromXattrs(...): Long = ???

// getFileSize could look like 
def getFileSize(...) = getFileSizeFromStatus(... ).filter(_ > 0)
   .getOrElse(getFileSizeFromXattrs(...))
```

gerashegalov (commented):
Logic LGTM, some nits

jlowe:
Thanks for the review, @gerashegalov.  I'd like to leave the nits as-is, as it reflects the existing Spark code in `BasicWriteTaskStatsTracker` (and thus minimizes the diffs).  If you feel more strongly that we should address them anyway, Iet me know.

gerashegalov (approved):

Issue #4980, opened 2022-03-18
[BUG] window_function_test FAILED on PASCAL GPU


NvTimLiu:
**Describe the bug**

This failure seems only occur on PASCAL GPU.

Got diff window_function_test() data output between CPU and on PASCAL GPU.

Job: #rapids_it-PASCAL/2

```
=========================== short test summary info ============================

FAILED ../../src/main/python/window_function_test.py::test_window_running[Byte-Long-1000][IGNORE_ORDER({'local': True})]

FAILED ../../src/main/python/window_function_test.py::test_window_running[Byte-Long-1g][IGNORE_ORDER({'local': True})]

FAILED ../../src/main/python/window_function_test.py::test_window_running[Short-Long-1000][IGNORE_ORDER({'local': True})]

FAILED ../../src/main/python/window_function_test.py::test_window_running[Short-Long-1g][IGNORE_ORDER({'local': True})]

FAILED ../../src/main/python/window_function_test.py::test_window_running[Boolean-Long-1g][IGNORE_ORDER({'local': True})]


_____________________ test_window_running[Boolean-Long-1g] _____________________

b_gen = Boolean, c_gen = Long, batch_size = '1g'

    @ignore_order(local=True)
    @pytest.mark.parametrize('batch_size', ['1000', '1g'], ids=idfn) # set the batch size so we can test multiple stream batches
    @pytest.mark.parametrize('b_gen, c_gen', [(long_gen, x) for x in running_part_and_order_gens] +
            [(x, long_gen) for x in all_basic_gens_no_nans + [decimal_gen_32bit]], ids=idfn)
    def test_window_running(b_gen, c_gen, batch_size):
        conf = {'spark.rapids.sql.batchSizeBytes': batch_size,
                'spark.rapids.sql.hasNans': False,
                'spark.rapids.sql.variableFloatAgg.enabled': True,
                'spark.rapids.sql.castFloatToDecimal.enabled': True}
        query_parts = ['b', 'a', 'row_number() over (partition by b order by a rows between UNBOUNDED PRECEDING AND CURRENT ROW) as row_num',
                'rank() over (partition by b order by a rows between UNBOUNDED PRECEDING AND CURRENT ROW) as rank_val',
                'dense_rank() over (partition by b order by a rows between UNBOUNDED PRECEDING AND CURRENT ROW) as dense_rank_val',
[...]

PR #5122, opened 2022-04-01
Disable GPU accelerated row-column transpose for Pascal GPUs:

mythrocks:
Depends on rapidsai/cudf/pull/10568.
Works around the failures described in #4980. 

This commit bypasses GPU accelerated row-column conversion for Pascal GPUs. 
Other architectures should remain unaffected by this shunt.

Further investigation might be required to find the actual problem in
CUDF's `fixed_width_convert_to_rows()` implementation.

Signed-off-by: MithunR <mythrocks@gmail.com>

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
@@ -19,7 +19,7 @@ package com.nvidia.spark.rapids
 import scala.annotation.tailrec
 import scala.collection.mutable.Queue
 
-import ai.rapids.cudf.{HostColumnVector, NvtxColor, Table}
+import ai.rapids.cudf.{Cuda, HostColumnVector, NvtxColor, Table}
 import com.nvidia.spark.rapids.GpuColumnarToRowExecParent.makeIteratorFunc
 import com.nvidia.spark.rapids.shims.ShimUnaryExecNode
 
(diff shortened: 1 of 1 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
```diff
@@ -343,6 +343,16 @@ object GpuColumnarToRowExecParent {
     Option(Tuple2(arg.child, arg.exportColumnarRdd))
   }
 
+  /**
+   * Helper to check if GPU accelerated row-column transpose is supported
+   */
+  private def isAcceleratedTransposeSupported: Boolean = {
```
jlowe, line 349: We can cache the value rather than continually detecting the version:
```suggestion
  private lazy val isAcceleratedTransposeSupported: Boolean = {
```

mythrocks: That's where I first started. I might have overthought it:

Could the Exec instance be shared in case the same machine had two GPUs of different architectures?

jlowe: The RAPIDS Accelerator only supports one GPU per executor.  The executor picks a GPU device on startup and sticks with it.

mythrocks: Fixed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuColumnarToRowExec.scala
```diff
@@ -356,7 +366,9 @@ object GpuColumnarToRowExecParent {
         // This number includes the 1-bit validity per column, but doesn't include padding.
         // We are being conservative by only allowing 100M columns until we feel the need to
         // increase this number
-        output.length <= 100000000) {
+        output.length <= 100000000 &&
+        // Check if the current CUDA device supports accelerated transpose.
```
bdice, line 370: Is accelerated transpose a hardware feature? A driver feature? Can you clarify what causes this to fail on Pascal?

mythrocks: I've been chasing the problem for a while now. I think this comment belies that fact. 
I should mention that this is a workaround.

mythrocks: The description has been modified to highlight that this is a workaround.

jlowe (approved):

mythrocks:
Build

Issue #5133, opened 2022-04-02
[BUG] Compile error for Spark330 because of Spark changed the method signature: QueryExecutionErrors.mapKeyNotExistError


res-life:
**Describe the bug**

mvn clean install -Dbuildver=330 -DskipTests

[INFO] Compiling 267 Scala sources and 29 Java sources to /home/chong/code/2-spark-rapids/sql-plugin/target/spark330/classes ...
[ERROR] [Error] /home/chong/code/2-spark-rapids/sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/RapidsErrorUtils.scala:33: not enough arguments for method mapKeyNotExistError: (key: Any, isElementAtFunction: Boolean, context: String)NoSuchElementException.
Unspecified value parameter context.
[ERROR] one error found

Spark added a new parameter in `QueryExecutionErrors.mapKeyNotExistError`

**Additional context**
related to https://issues.apache.org/jira/browse/SPARK-38716 which affected 330+

PR #5124, opened 2022-04-01
Fix ShimVectorizedColumnReader construction for recent Spark 3.3.0 changes

jlowe:
Fixes #5123.
Fixes #5133.

Adds in a `ParsedVersion` parameter which is new in the Spark 3.3.0 constructor for `VectorizedColumnReader`.  Creation of the parameter mirrors the same logic implemented in [SPARK-37974](https://issues.apache.org/jira/browse/SPARK-37974) and [SPARK-34863](https://issues.apache.org/jira/browse/SPARK-34863).

In addition this contains a build fix for the `QueryExecutionErrors` changes in [SPARK-38716](https://issues.apache.org/jira/browse/SPARK-38716).

Issue #5123, opened 2022-04-01
[BUG] Compile error for Spark330 because of VectorizedColumnReader constructor added a new parameter.


res-life:
**Describe the bug**

mvn clean install -Dbuildver=330 -DskipTests
**not enough arguments for constructor VectorizedColumnReader**

```
[INFO] Compiler bridge file: /home/chong/.sbt/1.0/zinc/org.scala-sbt/org.scala-sbt-compiler-bridge_2.12-1.3.1-bin_2.12.15__52.0-1.3.1_20191012T045515.jar
[INFO] Compiling 267 Scala sources and 29 Java sources to /home/chong/code/2-spark-rapids/sql-plugin/target/spark330/classes ...
[ERROR] [Error] /home/chong/code/2-spark-rapids/sql-plugin/src/main/321+/scala/org/apache/spark/sql/execution/datasources/parquet/rapids/shims/ShimVectorizedColumnReader.scala:58: not enough arguments for constructor VectorizedColumnReader: (x$1: org.apache.parquet.column.ColumnDescriptor, x$2: org.apache.parquet.schema.LogicalTypeAnnotation, x$3: org.apache.parquet.column.page.PageReader, x$4: java.util.PrimitiveIterator.OfLong, x$5: java.time.ZoneId, x$6: String, x$7: String, x$8: String, x$9: String, x$10: org.apache.parquet.VersionParser.ParsedVersion)org.apache.spark.sql.execution.datasources.parquet.VectorizedColumnReader.
Unspecified value parameter x$10.
[ERROR] [Error] /home/chong/code/2-spark-rapids/sql-plugin/src/main/321+/scala/org/apache/spark/sql/execution/datasources/parquet/rapids/shims/ShimVectorizedColumnReader.scala:20: Unused import
[ERROR] two errors found
```

pxLi:
related to https://github.com/apache/spark/pull/35262

PR #5020, opened 2022-03-23
Support arithmetic operators on ANSI interval types

res-life:
Contributes #4151

Support arithmetic operators on ANSI interval types(year month interval and day time interval)
- Abs: abs(interval)
- UnaryMinus: -interval
- Add: interval + interval
- Subtract: interval - interval
- UnaryPositive: +interval

Signed-off-by: Chong Gao <res_life@163.com>

PR #5134, opened 2022-04-02
Support division on ANSI interval types

res-life:
Contributes https://github.com/NVIDIA/spark-rapids/issues/4151

Supports 
- Year-month interval / (byte, short, int, long, float, double)
- Day-time interval / (byte, short, int, long, float, double)

Note: not support decimal type because of CUDF Decimal128 does not handle overflow, like:
```
CUDF positive decimal * positive decimal can generate negative value.
val d1 = ColumnVector.fromDecimals(new java.math.BigDecimal("123456789012345678901234567890123456.11"))
val d2 = Scalar.fromDecimal(new java.math.BigDecimal("12345678901234567890123456789012345611"))
val r = d1.mul(d2)
println(r)

123456789012345678901234567890123456.11 * 12345678901234567890123456789012345611 = -1053265448675579329726886670761974900.55
```


res-life:
Blocked by 2 Spark330 build issues:
https://github.com/NVIDIA/spark-rapids/issues/5123
https://github.com/NVIDIA/spark-rapids/issues/5133

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/Spark33XShims.scala
```diff
@@ -312,6 +312,39 @@ trait Spark33XShims extends Spark321PlusShims with Spark320PlusNonDBShims {
         (a, conf, p, r) => new BinaryAstExprMeta[LessThanOrEqual](a, conf, p, r) {
           override def convertToGpu(lhs: Expression, rhs: Expression): GpuExpression =
             GpuLessThanOrEqual(lhs, rhs)
+        }),
+      GpuOverrides.expr[Alias](
```
jlowe, line 316: This should leverage the new `TypeSig` interval type shims added in #5020 to avoid copying a lot of code just to override the accepted types for Spark 3.3

jlowe (commented):
The temporary changes needed to compile on Spark 3.3 can be removed now that #5124 has been merged.

res-life:
Waiting https://github.com/NVIDIA/spark-rapids/issues/5150 to merge, has some conflict on Spark33XShims.

res-life:
Waiting https://github.com/NVIDIA/spark-rapids/issues/5150 to merge, has some conflict on Spark33XShims.

res-life:
build

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/Spark33XShims.scala
res-life: Updated

res-life:
Rewrited the division code by leveraging `decimal_128` type to reduce overflow check time,  got about 2x speedup.

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
```diff
@@ -184,6 +186,113 @@ object IntervalUtils extends Arm {
...
+
+  def getDouble(s: Scalar): Double = {
+    s.getType match {
+      case DType.INT8 => s.getByte.toDouble
+      case DType.INT16 => s.getShort.toDouble
+      case DType.INT32 => s.getInt.toDouble
+      case DType.INT64 => s.getLong.toDouble
+      case DType.FLOAT32 => s.getFloat.toDouble
+      case DType.FLOAT64 => s.getDouble
+      case _ => throw new IllegalArgumentException()
```
jlowe, line 198: While we don't expect to ever hit this, if we do then it will be annoying that the exception didn't bother to state what the unexpected type was.
```suggestion
      case t =>
        throw new IllegalArgumentException(s"Unexpected type $t")
```
Comment applies to the other instances of this below.
Comment applies to the

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
```diff
@@ -184,6 +186,113 @@ object IntervalUtils extends Arm {
...
+          case (lS: Scalar, rS: Scalar) =>
+            getLong(lS) == min && getLong(rS) == -1L
+        }
+      }
+    }
+
+    // 3. round(p.asDecimal / q)
+    val dT = DType.create(DType.DTypeEnum.DECIMAL128, -1)
+    val leftDecimal = p match {
+      case pCv: ColumnVector => pCv.castTo(dT)
```
jlowe, line 289: Is pre-casting really necessary, or will this happen naturally when we specify the desired output type in the `.div` call below?  If cudf is going to pre-cast for us during the div operation, it would be nice to avoid manifesting the intermediate result explicitly.

res-life: Cudf does not do this for us. If so, it will improve the performance.

```
    val dT = DType.create(DType.DTypeEnum.DECIMAL128, -1)
    val cv1 = ColumnVector.fromLongs(5L)
    val cv2 = ColumnVector.fromLongs(3L)
    println(cv1.div(cv2, dT).getScalarElement(0))
```
get Scalar{type=DECIMAL128 scale:-1 **value=1.0**}

But for the `long_cv.mul(long_cv, DECIMAL128(scale = 0))`, we can do it, the only difference is the `scale = 0` instead of `scale = 
 1`

I think we can file a `feature request`  to cuDF, what do you think?

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
res-life: done

res-life:
build

sql-plugin/src/main/330+/scala/org/apache/spark/sql/rapids/shims/intervalExpressions.scala
jlowe: Yeah I think cudf will only upcast to the "natural" type of the two operands and the operator then cast at the end to the desired type.  Since C++ won't naturally upcast to __int128_t for this operator, I don't think it's reasonable to ask cudf to add this feature.  I'm sure they would just push back and say we need to upcast if we need that behavior.  Otherwise it would add quite a bit of complexity to all their binary operators for arguably a fairly specialized type of behavior.

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
```diff
@@ -0,0 +1,287 @@
...
+      e => e.getMessage.contains("ArithmeticException"),
+      spark => {
+        val data = Seq(Row(Period.ofMonths(months)))
+        val schema = StructType(Seq(
+          StructField("c1", YearMonthIntervalType())))
+        spark.createDataFrame(spark.sparkContext.parallelize(data), schema)
+      }
+    ) {
+      df =>
+        if (num.isInstanceOf[Byte]) {
```
jlowe, line 132: Why not use pattern matching (i.e.: `num match { case _: Byte => ...`) here instead of a series of explicit instance checks?  Applies to the other places where it's doing a series of type checks.

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
```diff
@@ -0,0 +1,287 @@
...
+          } else if (num.equals(Double.PositiveInfinity)) {
+            df.selectExpr("c1 / cast('Infinity' as double)")
+          }
+          else if (num.equals(Double.NegativeInfinity)) {
+            df.selectExpr("c1 / cast('-Infinity' as double)")
+          } else {
+            df.selectExpr(s"c1 / ${num}d")
+          }
+        } else {
+          throw new IllegalStateException("")
```
jlowe, line 163: Nit: Should print a more useful message here in case it is hit, otherwise one has to line up the line number in the exception stacktrace with the source to understand what went wrong.  Also recommend using `fail` here since this is a test.

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
```diff
@@ -0,0 +1,287 @@
...
+                  StructField("c2", FloatType)
+                )))
+          } else if (num.isInstanceOf[Double]) {
+            (Seq(Row(Period.ofMonths(months), num.asInstanceOf[Double])),
+                StructType(Seq(
+                  StructField("c1", YearMonthIntervalType()),
+                  StructField("c2", DoubleType)
+                )))
+          } else {
+            throw new IllegalStateException("")
```
jlowe, line 209: Nit: use `fail` with a useful error message.

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
```diff
@@ -0,0 +1,287 @@
...
+                StructType(Seq(
+                  StructField("c1", FloatType)
+                )))
+          } else if (num.isInstanceOf[Double]) {
+            (Seq(Row(num)),
+                StructType(Seq(
+                  StructField("c1", DoubleType)
+                )))
+          } else {
+            throw new IllegalStateException("")
```
jlowe, line 253: Nit: use `fail` with a useful error message.

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
```diff
@@ -0,0 +1,287 @@
...
+      }
+    ) {
+      df => df.selectExpr(s"c1 / c2")
+    }
+
+    testBothCpuGpuExpectedException[SparkException](testCaseName + ", scalar / cv",
+      e => e.getMessage.contains("ArithmeticException"),
+      spark => {
+        val (data, schema) =
+          if (num.isInstanceOf[Byte]) {
```
jlowe, line 222: Nit: pattern matching would be more natural

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
```diff
@@ -0,0 +1,287 @@
...
+      spark => {
+        // Period is the external type of year-month type
+        val (data, schema) =
+          if (num.isInstanceOf[Byte]) {
+            (Seq(Row(Period.ofMonths(months), num.asInstanceOf[Byte])),
+                StructType(Seq(
+                  StructField("c1", YearMonthIntervalType()),
+                  StructField("c2", ByteType)
+                )))
+          } else if (num.isInstanceOf[Short]) {
```
jlowe, line 178: Pattern matching makes more sense here, as it's more concise to write and read and removes the need to cast, e.g.:
```suggestion
        val (data, schema) = num match {
          case n: Byte =>
            (Seq(Row(Period.ofMonths(months), n)),
                StructType(Seq(
                  StructField("c1", YearMonthIntervalType()),
                  StructField("c2", ByteType)
                )))
          case n: Short =>
            ...
```

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
res-life: done

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
res-life: done

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
res-life: done

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
res-life: done

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
res-life: done

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
res-life: done

res-life:
build

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
```diff
@@ -0,0 +1,232 @@
...
+      spark => {
+        val data = Seq(Row(Period.ofMonths(months)))
+        val schema = StructType(Seq(
+          StructField("c1", YearMonthIntervalType())))
+        spark.createDataFrame(spark.sparkContext.parallelize(data), schema)
+      }
+    ) {
+      df =>
+        num match {
+          case Byte => df.selectExpr(s"c1 / cast('$num' as Byte)")
```
jlowe, line 133: This does not match properly, as it's checking an instance against a class which will never match.  It needs to be written something like this instead:
```suggestion
          case _: Byte => df.selectExpr(s"c1 / cast('$num' as Byte)")
```

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
```diff
@@ -0,0 +1,232 @@
...
+          case _ => fail(s"Unsupported num type ${num.getClass}")
+        }
+    }
+
+    testBothCpuGpuExpectedException[SparkException](testCaseName + ", cv / cv",
+      e => e.getMessage.contains("ArithmeticException"),
+      spark => {
+        // Period is the external type of year-month type
+        val (data, schema) = num match {
+          case Byte => (Seq(Row(Period.ofMonths(months), num)), StructType(Seq(
```
jlowe, line 154: These cases will never match, needs to be something like this instead:
```suggestion
          case n: Byte => (Seq(Row(Period.ofMonths(months), n)), StructType(Seq(
```

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
```diff
@@ -0,0 +1,232 @@
...
+      }
+    ) {
+      df => df.selectExpr(s"c1 / c2")
+    }
+
+    testBothCpuGpuExpectedException[SparkException](testCaseName + ", scalar / cv",
+      e => e.getMessage.contains("ArithmeticException"),
+      spark => {
+        val (data, schema) = num match {
+          case Byte => (Seq(Row(num)), StructType(Seq(StructField("c1", ByteType))))
```
jlowe, line 193: Same case matching issue

jlowe (commented):
Tests are no longer passing after the recent changes, which implies they were not rerun after the changes were made.  Note that since this test is Spark 3.3.0+, it's not normally run as part of premerge (that may soon change once Spark 3.3 finally ships).  In the meantime, it's important to run it manually or we'll end up finding it doesn't pass on the nightly 3.3.0 runs after it already has been merged.

res-life:
build

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
res-life: Sorry for this mistake. 
Updated.

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
res-life: Updated.

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
res-life: Updated.

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
```diff
@@ -0,0 +1,236 @@
...
+        spark.createDataFrame(spark.sparkContext.parallelize(data), schema)
+      }
+    ) {
+      df =>
+        num match {
+          case b: Byte => df.selectExpr(s"c1 / cast('$b' as Byte)")
+          case s: Short => df.selectExpr(s"c1 / cast('$s' as Short)")
+          case i: Int => df.selectExpr(s"c1 / $i")
+          case l: Long => df.selectExpr(s"c1 / ${l}L")
+          case f: Float if f.equals(Float.NaN) => df.selectExpr("c1 / cast('NaN' as float)")
```
jlowe, line 137:
```suggestion
          case f: Float if f.isNaN => df.selectExpr("c1 / cast('NaN' as float)")
```

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
```diff
@@ -0,0 +1,236 @@
...
+          case f: Float if f.equals(Float.PositiveInfinity) =>
+            df.selectExpr("c1 / cast('Infinity' as float)")
+          case f: Float if f.equals(Float.NegativeInfinity) =>
+            df.selectExpr("c1 / cast('-Infinity' as float)")
+          case f: Float => df.selectExpr(s"c1 / ${f}f")
+          case d: Double if d.equals(Double.NaN) => df.selectExpr("c1 / cast('NaN' as double)")
```
jlowe, line 143:
```suggestion
          case d: Double if d.isNaN => df.selectExpr("c1 / cast('NaN' as double)")
```

jlowe (commented):
Minor nit on checking for NaN values using the dedicated predicate method, but otherwise lgtm.

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
res-life: done

tests/src/test/330/scala/com/nvidia/spark/rapids/IntervalDivisionSuite.scala
res-life: done

res-life:
build

jlowe (approved):

jlowe:
CI failure is unrelated, appears to be the (transient?) Cloudera dependency issue.  Kicking the build again.

jlowe:
build

PR #5136, opened 2022-04-02
Fix GpuRegExExtract about inconsistent to Spark 

### LINKED ISSUE #5135: [BUG] GpuRegExExtract is not align with RegExExtract
**Describe the bug**

1. In Spark, the pattern of `regexp_extract` isn't required to fully match the input string. Just as what extract means, regexp_extract finds any substring which matches the pattern and extracts them. However, GpuRegExExtract requires a full match on pattern.

```scala
val df =Seq("1a", "2a", "3a", "4a", "5a", "6a", "7a", "8a", "9a", "10a").toDF("c")
df.coalesce(1).select(regexp_extract(col("c"), "(a)", 1)).collect()
```
GPU result: `Array([], [], [], [], [], [], [], [], [], [])`
CPU result: `Array([a], [a], [a], [a], [a], [a], [a], [a], [a], [a])`

2. When group index is 0, GpuRegExExtract doesn't behave correctly.

CPU run: `regexp_extract('123abcEfg', '([0-9]+)[a-z]+([A-Z])', 0)` => `123abcE`
CPU run: `regexp_extract('123abcEfg', '([0-9]+)[a-z]+([A-Z])', 0)` => `123abcEfg`

BTW, the issue is originated from #5088.

sperlingxx:
Closes #5135

This PR fixes two inconsistent to Spark problems of GpuRegExExtract, which are raised in #5135.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
@@ -1079,7 +1079,20 @@ case class GpuRegExpExtract(
       regexp: GpuScalar,
       idx: GpuScalar): ColumnVector = {
 
-    val groupIndex = idx.getValue.asInstanceOf[Int]
+    // When group index is 0, it means extract the entire pattern including those parts which
+    // don't belong to any group. For instance,
+    // regexp_extract('123abcEfg', '([0-9]+)[a-z]+([A-Z])', 0) => 123abcE
+    // regexp_extract('123abcEfg', '([0-9]+)[a-z]+([A-Z])', 1) => 123
+    // regexp_extract('123abcEfg', '([0-9]+)[a-z]+([A-Z])', 2) => E
+    //
+    // To support the full match (group index 0), we wrap a pair of parentheses on the original
+    // cudfRegexPattern.
+    val (extractPattern, groupIndex) = idx.getValue match {
+      case i: Int if i == 0 =>
+        ("(" + cudfRegexPattern + ")", 0)
+      case i =>
+        (cudfRegexPattern, i.asInstanceOf[Int] - 1)
+    }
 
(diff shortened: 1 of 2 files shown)


sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1079,7 +1079,20 @@ case class GpuRegExpExtract(
...
+    // don't belong to any group. For instance,
+    // regexp_extract('123abcEfg', '([0-9]+)[a-z]+([A-Z])', 0) => 123abcE
+    // regexp_extract('123abcEfg', '([0-9]+)[a-z]+([A-Z])', 1) => 123
+    // regexp_extract('123abcEfg', '([0-9]+)[a-z]+([A-Z])', 2) => E
+    //
+    // To support the full match (group index 0), we wrap a pair of parentheses on the original
+    // cudfRegexPattern.
+    val (extractPattern, groupIndex) = idx.getValue match {
+      case i: Int if i == 0 =>
+        ("(" + cudfRegexPattern + ")", 0)
```
ttnghia, line 1092: What if `i != 0`?

andygrove: > What if `i != 0`?

Then it will fall through to the following `case i` statement

andygrove (approved):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
andygrove: This is Scala's "Pattern Guard" feature - https://docs.scala-lang.org/tour/pattern-matching.html

jlowe (approved):

PR #5137, opened 2022-04-03
Fix RapidsErrorUtils compile error for recent Spark 3.3.0 changes

### LINKED ISSUE #5133: [BUG] Compile error for Spark330 because of Spark changed the method signature: QueryExecutionErrors.mapKeyNotExistError

res-life:
Fixes #5133 

Fix RapidsErrorUtils compile error for recent Spark 3.3.0 changes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -464,6 +464,7 @@
                                         <source>${project.basedir}/src/main/320+-nondb/scala</source>
                                         <source>${project.basedir}/src/main/320until330-all/scala</source>
                                         <source>${project.basedir}/src/main/321+/scala</source>
+                                        <source>${project.basedir}/src/main/321until330-all/scala</source>
                                         <source>${project.basedir}/src/main/post320-treenode/scala</source>
                                     </sources>
                                 </configuration>
@@ -524,8 +525,9 @@
                                         <source>${project.basedir}/src/main/311until330-nondb/scala</source>
                                         <source>${project.basedir}/src/main/320+/scala</source>
                                         <source>${project.basedir}/src/main/320+-nondb/scala</source>
-                                        <source>${project.basedir}/src/main/321+/scala</source>
                                         <source>${project.basedir}/src/main/320until330-all/scala</source>
+                                        <source>${project.basedir}/src/main/321+/scala</source>
+                                        <source>${project.basedir}/src/main/321until330-all/scala</source>
                                         <source>${project.basedir}/src/main/post320-treenode/scala</source>
                                     </sources>
                                 </configuration>
@@ -597,8 +599,9 @@
                                         <source>${project.basedir}/src/main/311until330-all/scala</source>
                                         <source>${project.basedir}/src/main/311+-db/scala</source>
                                         <source>${project.basedir}/src/main/320+/scala</source>
-                                        <source>${project.basedir}/src/main/321+/scala</source>
                                         <source>${project.basedir}/src/main/320until330-all/scala</source>
+                                        <source>${project.basedir}/src/main/321+/scala</source>
+                                        <source>${project.basedir}/src/main/321until330-all/scala</source>
                                         <source>${project.basedir}/src/main/post320-treenode/scala</source>
                                     </sources>
                                 </configuration>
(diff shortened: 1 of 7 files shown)


res-life:
build

res-life:
@jlowe , please review

This fixes another 330 build issue, and it merged your PR #5124 to pass the compile.
About this issue, Spark added a new parameter `context` in `QueryExecutionErrors.mapKeyNotExistError`.
It makes a better error msg when `ElementAt` and `GetMapValue` occur `not found error`

`context`  is a SQL string with line and position info, like:

```
== SQL(line 1, position 0) ==
element_at(col, 5)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The `context` is generated from `TreeNode.origin.context`,
because of `origin` is thread local variable while CPU processing data, we can't catch it on GPU.
Let's only pass an empty string. This will no effect processing on GPU, only lack above msg when occurs `not found error`

e.g.:

**CPU error info**
```
Caused by: org.apache.spark.SparkNoSuchElementException: Key 5 does not exist. To return NULL instead, use 'try_element_at'. If necessary set spark.sql.ansi.enabled to false to bypass this error.
== SQL(line 1, position 0) ==
element_at(col, 5)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

**GPU error info**
```
Caused by: org.apache.spark.SparkNoSuchElementException: Key 5 does not exist. To return NULL instead, use 'try_element_at'. If necessary set spark.sql.ansi.enabled to false to bypass this error.
```

jlowe:
> because of origin is thread local variable while CPU processing data, we can't catch it on GPU.

I don't think this is correct.  IIUC the context is relative to the original SQL query, i.e.: how the current node in the plan maps back to the original SQL query text.  That applies to GPU processing just as much as CPU processing.

The CI failed because Spark added _yet another_ change to VectorizedColumnReader.  I've updated #5124 to fix that along with this RapidsErrorUtils change by passing the origin so the context can be retrieved when running on Spark 3.3+

jlowe:
Superceded by #5124.

res-life:
I tested the latest code, `origin`  is empty.
Still need to investigate in my opinion.

```

spark.conf.set("spark.sql.ansi.enabled",True)

schema = StructType([StructField("map_col", MapType(ByteType(), ByteType()))])

data = [({1:1},)]
df = spark.createDataFrame(data, schema)
df.show()

# test GPU
df.selectExpr("element_at(map_col, 5)").show()

# test CPU
spark.conf.set("spark.rapids.sql.enabled", False)
df.selectExpr("element_at(map_col, 5)").show()

```

Outputs of GPU and CPU are:
```
# GPU
Caused by: org.apache.spark.SparkNoSuchElementException: Key 5 does not exist. To return NULL instead, use 'try_element_at'. If necessary set spark.sql.ansi.enabled to false to bypass this error.

# CPU
Caused by: org.apache.spark.SparkNoSuchElementException: Key 5 does not exist. To return NULL instead, use 'try_element_at'. If necessary set spark.sql.ansi.enabled to false to bypass this error.
== SQL(line 1, position 0) ==
element_at(map_col, 5)
^^^^^^^^^^^^^^^^^^^^^^
```

jlowe:
Agree, seems using the origin is no more helpful than passing an empty string as was done in this PR.  I filed #5152 to investigate.

PR #5144, opened 2022-04-05
Adapt to storage-partitioned join additions in SPARK-37377 [databricks]

jlowe:
Fixes the build failures in Spark 3.3 caused by [SPARK-37377](https://issues.apache.org/jira/browse/SPARK-37377).  `DataSourceRDD` now expects a sequence-of-sequence-of-partitions and `BatchScanExec` similarly has changes that require both to be shimmed.  In the process of shimming these, I went ahead and pulled in the changes from [SPARK-35779](https://issues.apache.org/jira/browse/SPARK-35779) so we're better prepared to support runtime-filtering on scans, given we needed to shim `GpuBatchScanExec` anyway.


jlowe:
build

sql-plugin/src/main/320+/scala/org/apache/spark/sql/execution/datasources/rapids/DataSourceStrategyUtils.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021-2022, NVIDIA CORPORATION.
```
tgravescs, line 2: accidentally removed 2021?

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/Spark33XShims.scala
```diff
@@ -343,8 +343,8 @@ trait Spark33XShims extends Spark321PlusShims with Spark320PlusNonDBShims {
           override val childScans: scala.Seq[ScanMeta[_]] =
             Seq(GpuOverrides.wrapScan(p.scan, conf, Some(this)))
 
-          override def convertToGpu(): GpuExec =
-            GpuBatchScanExec(p.output, childScans.head.convertToGpu())
+          override def convertToGpu(): GpuExec = GpuBatchScanExec(p.output,
```
tgravescs, line 346: so this doesn't look like for 33X shims we tag properly to not run on gpu when the runtimeFilters are present? I'm guessing when we override the Spark320PlusShims we dropped that accidentally 

```
 override def tagPlanForGpu(): Unit = {
            if (!p.runtimeFilters.isEmpty) {
              willNotWorkOnGpu("Runtime filtering (DPP) on datasource V2 is not supported")
            }
          }
```

jlowe: Nice catch!  I updated this to check for both runtime filtering and key grouped partitions.

jlowe:
build

sql-plugin/src/main/320+/scala/org/apache/spark/sql/execution/datasources/rapids/DataSourceStrategyUtils.scala
jlowe: This is a case of git mistakenly thinking this is a file rename.  The files have very little in common other than a few imports and the ASL2.0 license boilerplate.  Otherwise the two files are not related.

tgravescs (approved):

PR #5157, opened 2022-04-07
Update the log warning of UDF compiler

### LINKED ISSUE #5104: [BUG] Unconditional warning in UDF Plugin "The compiler is disabled by default"
**Describe the bug**
Driver log warns about disabled compiler even if the user has `spark.rapids.sql.udfCompiler.enabled=true` in the launch config. 
```
22/03/31 09:07:59 WARN Plugin: Installing rapids UDF compiler extensions to Spark. The compiler is disabled by default. To enable it, set `spark.rapids.sql.udfCompiler.enabled` to true
```
Launch spark-shell with the Plugin and spark.rapids.sql.udfCompiler.enabled=true as part of the conf.

**Expected behavior**
Either don't print how to enable the UDF compiler if it's already enabled or say it's enabled and how to *disable* it. 

 - Environment location: any, local dev
 - Spark configuration settings related to the issue: `spark.rapids.sql.udfCompiler.enabled`

**Additional context**
N/A

HaoYang670:

Close #5104.
Update the log warning. Delete the content of how to enable udf compiler.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
@@ -90,6 +90,14 @@ object RapidsPluginUtils extends Logging {
       logWarning("RAPIDS Accelerator is disabled, to enable GPU " +
         s"support set `${RapidsConf.SQL_ENABLED}` to true.")
     }
+
+    if (conf.isUdfCompilerEnabled) {
+      logWarning("RAPIDS UDF compiler is enabled, to disable it " +
+      s"set `${RapidsConf.UDF_COMPILER_ENABLED}` to false.")
+    } else {
+      logWarning("RAPIDS UDF compiler is disabled, to enable it " +
+      s"set `${RapidsConf.UDF_COMPILER_ENABLED}` to true.")
+    }
   }
 
   def fixupConfigs(conf: SparkConf): Unit = {
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
@@ -92,11 +92,10 @@ object RapidsPluginUtils extends Logging {
     }
 
     if (conf.isUdfCompilerEnabled) {
-      logWarning("RAPIDS UDF compiler is enabled, to disable it " +
-      s"set `${RapidsConf.UDF_COMPILER_ENABLED}` to false.")
-    } else {
-      logWarning("RAPIDS UDF compiler is disabled, to enable it " +
-      s"set `${RapidsConf.UDF_COMPILER_ENABLED}` to true.")
+      logWarning("Experimental RAPIDS UDF compiler is enabled, in case of related failures " +
+      s"disable it by setting `${RapidsConf.UDF_COMPILER_ENABLED}` to false. " +
+      "More information is available at https://nvidia.github.io/spark-rapids/docs/FAQ.html#" +
+      "automatic-translation-of-scala-udfs-to-apache-spark-operations" )
     }
   }
 
--- a/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
+++ b/udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2022, NVIDIA CORPORATION.
  *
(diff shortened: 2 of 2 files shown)


udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
```diff
@@ -26,7 +26,7 @@ import org.apache.spark.sql.catalyst.rules.Rule
 class Plugin extends (SparkSessionExtensions => Unit) with Logging {
   override def apply(extensions: SparkSessionExtensions): Unit = {
     logWarning("Installing rapids UDF compiler extensions to Spark. The compiler is disabled" +
```
jlowe, line 28: IMO if the UDF compiler is disabled, we shouldn't print any message at all about it on startup.  A startup message about the UDF compiler should appear iff the UDF compiler is enabled.

abellina: +1 and I would leave the config in the log line also, but turn this around to be:

> To disable it, set `spark.rapids.sql.udfCompiler.enabled` to false

HaoYang670: I tried to get the `RapidsConf.isUdfCompilerEnabled`, but failed because I could not get an active Spark Session
```scala
  override def apply(extensions: SparkSessionExtensions): Unit = {
    SparkSession.getActiveSession match {
      case None => // no active session
      case Some(session) => // log warning based on `session.sparkContext.getConf.isUdfCompilerEnabled`
    }
    extensions.injectResolutionRule(logicalPlanRules)
  }
```

jlowe: There's no Spark session this early in startup.  We'll either need to move the logging message to the code that is injecting the plugin (see `RapidsPluginUtils.fixupConfigs`) which does have access to the configs, or the configs can be fetched from `SparkEnv.get.conf`.  I slightly prefer the former since the configs are right there.

HaoYang670:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -90,6 +90,14 @@ object RapidsPluginUtils extends Logging {
       logWarning("RAPIDS Accelerator is disabled, to enable GPU " +
         s"support set `${RapidsConf.SQL_ENABLED}` to true.")
     }
+
+    if (conf.isUdfCompilerEnabled) {
+      logWarning("RAPIDS UDF compiler is enabled, to disable it " +
```
gerashegalov, line 95: We might want to link to our FAQ https://nvidia.github.io/spark-rapids/docs/FAQ.html#automatic-translation-of-scala-udfs-to-apache-spark-operations . Without such context it's not clear for the user why she is warned about this setting. What makes it different from other switches we don't draw attention to? Looks like we consider it bleeding edge (experimental, unstable, evolving), and the wording for the warning should make it clear. E.g. 

> Experimental RAPIDS UDF compiler is enabled, in case of related failures disable it by ...

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -90,6 +90,14 @@ object RapidsPluginUtils extends Logging {
...
+      s"set `${RapidsConf.UDF_COMPILER_ENABLED}` to false.")
+    } else {
+      logWarning("RAPIDS UDF compiler is disabled, to enable it " +
```
gerashegalov, line 98: +1 to @jlowe's comment https://github.com/NVIDIA/spark-rapids/pull/5157#discussion_r845182574 .

udf-compiler/src/main/scala/com/nvidia/spark/udf/Plugin.scala
HaoYang670: Done!

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
HaoYang670: Done!

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
HaoYang670: Done!

HaoYang670:
build

jlowe (approved):

gerashegalov (approved):
LGTM

abellina (approved):

Issue #5154, opened 2022-04-07
[BUG] build failed in databricks 10.4 runtime (updated recently)


pxLi:
**Describe the bug**
image of databricks 10.4 runtime seems get updated today.

which failed find some default jar dependencies.
```
[2022-04-07T01:11:37.173Z] [INFO] --- maven-install-plugin:2.4:install-file (default-cli) @ rapids-4-spark-parent ---

[2022-04-07T01:11:37.428Z] [ERROR] The specified file '/databricks/jars/----workspace_spark_3_2--maven-trees--hive-2.3__hadoop-3.2--org.apache.parquet--parquet-hadoop--org.apache.parquet__parquet-hadoop__1.12.0-databricks-0003.jar' not exists
```
![image](https://user-images.githubusercontent.com/8086184/162120858-28398f3b-c975-4a97-bdf5-c4efe246c2d3.png)

tried (https://github.com/NVIDIA/spark-rapids/pull/5153) bump up some versions, and then it failed due to some func signature changes,
```
[2022-04-07T04:19:32.450Z] [INFO] Compiling 273 Scala sources and 29 Java sources to /home/ubuntu/spark-rapids/sql-plugin/target/spark321db/classes...
[2022-04-07T04:19:34.954Z] [ERROR] /home/ubuntu/spark-rapids/sql-plugin/src/main/311+-db/scala/com/nvidia/spark/rapids/shims/AQEUtils.scala:28: not enough arguments for method apply: (id: Int, plan: org.apache.spark.sql.execution.SparkPlan, originalPlan: org.apache.spark.sql.execution.SparkPlan, isSparkExchange: Boolean)org.apache.spark.sql.execution.adaptive.ShuffleQueryStageExec in object ShuffleQueryStageExec.
[2022-04-07T04:19:34.955Z] Unspecified value parameter isSparkExchange.
[2022-04-07T04:19:34.955Z] [ERROR]     ShuffleQueryStageExec(sqse.id, reusedExchange, sqse.originalPlan)
[2022-04-07T04:19:34.955Z] [ERROR]                          ^
[...]

PR #5165, opened 2022-04-07
Update Databricks 10.4 for changes to the QueryStageExec and ClusteredDistribution [databricks]

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/5154

Looks like 2 different changes.

- pulled in https://issues.apache.org/jira/browse/SPARK-38237 which was changed in spark 3.3 back. These changes are in our Spark 3.3 shim so it was just copying them back to 321db.
- adds an extra parameter to *QueryStageExec called isSparkExchange.

I had to copy some files around to make it 321db specific.


tgravescs:
build

jlowe (approved):

tgravescs:
markdown failures are because we pushed docs changes to official 22.04 jars but they haven't been published yet so can ignore those failures

PR #5169, opened 2022-04-08
Terminate test earlier if pytest ENV issue [skip ci]

pxLi:

We saw get cases/tests by tag funcs could hide errors in pipefails
Instead of deal w/ PIPESTATUS and adjust string manipulation, we simply add a pytest ENV check (~20s) before the actual test run which will also expose all stdout&stderr to console. e.g. SPARK ENV issue, cuda ENV issue, pytest dep issue, etc

also remove redundant 311 ver check since we support spark311+ only

tested internally.


GaryShen2008 (dismissed):

pxLi:
build

pxLi:
retargeted to 22.04

GaryShen2008 (approved):

pxLi:
build

PR #5171, opened 2022-04-08
Fix the bug COALESCING reading does not work for v2 parquet/orc datasource

wbo4958:
COALESCING reading is not working for v2 datasource, which means ORC and Parquet will auto-fall back to multi-threaded reading even without the input_xxx expressions or ignoreCorruptFiles since the flag `queryUsesInputFile` is always true. Finally

``` scala
  override val canUseCoalesceFilesReader: Boolean =
    rapidsConf.isParquetCoalesceFileReadEnabled && !(queryUsesInputFile || ignoreCorruptFiles)
```

canUseCoalesceFilesReader will always be false.

Thx @firestarman for the issue reporting.

To close https://github.com/NVIDIA/spark-rapids/issues/5215


wbo4958:
build

tgravescs (commented):
this might be hard to integration test but it might be nice to have unit test for this.

tgravescs:
was this purely discovered internally by us testing or customer reported?  Please file an issue as well.

wbo4958:
build

wbo4958:
build

wbo4958:
@tgravescs could you help to review it?

jlowe (approved):

PR #5172, opened 2022-04-08
Fix the bug COALESCING reading does not work for v2 parquet/orc datasource

wbo4958:
Backport https://github.com/NVIDIA/spark-rapids/pull/5171 to 22.04


wbo4958:
build

pxLi:
if decide to target 22.04, plz close #5171 as auto-merge will take care of the commit for 22.06

wbo4958:
I think we should merge it in 22.04 @sameerz what's your opinion?

tgravescs:
the datasource v2 readers are off by default so its definitely less likely this is an issue.  Do we know any customers explicitly using it?  Otherwise I would say leave out of 22.40

GaryShen2008:
Agree, let's leave it out of 22.04. It'll use multithread, not block the feature, only impact the performance.

Issue #5182, opened 2022-04-08
[BUG] Spark 3.3 integration tests arithmetic_ops_test.py::test_div_overflow_exception_when_ansi failures


tgravescs:
**Describe the bug**
latest spark 3.3 integration test build failed with:

13:05:19  FAILED ../../src/main/python/arithmetic_ops_test.py::test_div_overflow_exception_when_ansi[true-CAST(-9223372036854775808L as LONG) DIV -1]
13:05:19  FAILED ../../src/main/python/arithmetic_ops_test.py::test_div_overflow_exception_when_ansi[true-a DIV CAST(-1 AS INT)]
13:05:19  FAILED ../../src/main/python/arithmetic_ops_test.py::test_div_overflow_exception_when_ansi[true-a DIV b]

andygrove:
The issue appears to be caused by https://github.com/apache/spark/commit/c2ed15de94f611972c463ff8bba9268c1e93fd30 which changes the exception that Spark throws. The tests expect `java.lang.ArithmeticException` but Spark is now throwing `org.apache.spark.SparkArithmeticException`.

PR #5185, opened 2022-04-08
Fix test failure with Spark 3.3 by looking for less specific error message

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/5182

Spark 3.3 changed which exception they throw for a divide that causes an overflow in https://github.com/apache/spark/commit/c2ed15de94f611972c463ff8bba9268c1e93fd30, causing our tests to fail.

This PR changes the affected test to just look for `ArithmeticException` rather than specifically `java.lang.ArithmeticException`. This is the same approach used in the other tests.

### DIFF
--- a/integration_tests/src/main/python/arithmetic_ops_test.py
+++ b/integration_tests/src/main/python/arithmetic_ops_test.py
@@ -19,8 +19,7 @@ from data_gen import *
 from marks import ignore_order, incompat, approximate_float, allow_non_gpu
 from pyspark.sql.types import *
 from pyspark.sql.types import IntegralType
-from spark_session import with_cpu_session, with_gpu_session, with_spark_session, is_before_spark_320, \
-    is_before_spark_330, is_databricks91_or_later, is_spark_330_or_later
+from spark_session import with_cpu_session, with_gpu_session, with_spark_session, is_before_spark_320, is_before_spark_330, is_databricks91_or_later
 import pyspark.sql.functions as f
 from datetime import timedelta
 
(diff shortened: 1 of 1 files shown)


andygrove:
build

tgravescs (approved):

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -788,7 +788,7 @@ def test_div_overflow_exception_when_ansi(expr, ansi_enabled):
         assert_gpu_and_cpu_error(
             df_fun=lambda spark: _get_div_overflow_df(spark, expr).collect(),
             conf=ansi_conf,
-            error_message='java.lang.ArithmeticException: Overflow in integral divide')
+            error_message='ArithmeticException: Overflow in integral divide')
```
jlowe, line 791: Isn't the point to ensure we're throwing the same exception?  As such I'd rather see this test check for the correct exception class based on the Spark version -- if Spark 3.3 throws a different exception for this, then our plugin should throw the same class for the same type of exception on Spark 3.3.

I'm OK if we want to do this as a followup to refine the test if we aren't already doing the correct behavior now.  If we are, then this should choose a different message to check based on which Spark we're using as is done in other tests checking exception types.

andygrove: Yes, that's a good point. I filed https://github.com/NVIDIA/spark-rapids/issues/5196 as a follow-on.

jlowe (approved):

Issue #5099, opened 2022-03-30
[BUG]Profiling tool should not sum gettingResultTime


tgravescs:
**Describe the bug**
someone asked me a question about the type of gettingResultTime_sum field in the profiling tool because it was huge.
it looks like we sum this but its actually a wallclock time in spark.

```
 /**
   * The time when the task started remotely getting the result. Will not be set if the
   * task result was sent immediately when the task finished (as opposed to sending an
   * IndirectTaskResult and later fetching the result from the block manager).
   */
  var gettingResultTime: Long = 0
```

so we should not be summing this across tasks for stages/jobs aggregation.

PR #5202, opened 2022-04-11
Profiling tool: Remove gettingResultTime from stages & jobs aggregation

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/5099 

As mentioned in the issue, gettingResultTime from TaskInfo is `The time when the task started remotely getting the result` which is not a value can be summed. Removed it from `JobStageAggTaskMetricsProfileResult` and `SQLTaskAggMetricsProfileResult` case classes and updated the docs and test results.


nartal1:
build

tgravescs:
markdown failure is just 22.04 jar link that isn't published yet.

PR #5214, opened 2022-04-12
Add implicit `safeFree` for `RapidsBuffer`

### LINKED ISSUE #5098: [BUG] Harden calls to `RapidsBuffer.free`
As @gerashegalov mentioned in https://github.com/NVIDIA/spark-rapids/pull/5093#pullrequestreview-926548173, there are various places in the code where `free` is called on a `RapidsBuffer` instance with the assumption that it does not throw.  We either should have something like a `safeFree` implicit utility function or appropriate `try...finally` blocks to ensure attempts to release other resources at the same time are not interrupted.

HaoYang670:

Close #5098 

Add implicit classes to safely free `RapidsBuffer`s


HaoYang670:
Not add tests yet.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/implicits.scala
```diff
@@ -59,6 +59,30 @@ object RapidsPluginImplicits {
     }
   }
 
+  implicit class RapidsBufferColumn[A <: RapidsBuffer](rapidsBuffer: RapidsBuffer) {
```
firestarman, line 62:
```suggestion
  implicit class RapidsBufferColumn[A <: RapidsBuffer](rapidsBuffer: A) {
```
??

sql-plugin/src/main/scala/com/nvidia/spark/rapids/implicits.scala
```diff
@@ -59,6 +59,30 @@ object RapidsPluginImplicits {
...
+    def safeFree(e: Throwable = null): Unit = {
+      if (rapidsBuffer != null) {
+        if (e != null) {
+          try {
+            rapidsBuffer.free()
+          } catch {
+            case suppressed: Throwable => e.addSuppressed(suppressed)
+          }
+        } else {
+          rapidsBuffer.free()
```
sperlingxx, line 80: Why do we free a null ptr?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/implicits.scala
```diff
@@ -59,6 +59,30 @@ object RapidsPluginImplicits {
...
   }
 
+  implicit class RapidsBufferColumn[A <: RapidsBuffer](rapidsBuffer: RapidsBuffer) {
+
+    /**
+     * safeFree: Is an implicit on RapidsBuffer class that tries to free the resource, if an
+     * Exception was thrown prior to this free, it adds the new exception to the suppressed
+     * exceptions, otherwise just throws
+     *
+     * @param e Exception which we don't want to suppress
```
sperlingxx, line 69: Does the comment align with the implementation ?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferCatalog.scala
```diff
@@ -145,7 +146,7 @@ class RapidsBufferCatalog extends Logging {
   def removeBuffer(id: RapidsBufferId): Unit = {
     val buffers = bufferMap.remove(id)
     if (buffers != null) {
```
abellina, line 148: do we need this null check? It seems the implicit checks that the sequence is not null, so we should be really safe to just do:
```
val buffers = bufferMap.remove(id)
buffers.safeFree()
```

HaoYang670:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/implicits.scala
```diff
@@ -87,11 +111,44 @@ object RapidsPluginImplicits {
...
   }
 
+  implicit class RapidsBufferSeq[A <: RapidsBuffer](val in: SeqLike[A, _]) {
+    /**
+     * safeFree: Is an implicit on a sequence of RapidsBuffer classes that tries to free each
+     * element of the sequence, even if prior free calls fail. In case of failure in any of the
+     * free calls, an Exception is thrown containing the suppressed exceptions (getSuppressed),
+     * if any.
+     */
+    def safeFree(error: Throwable = null): Unit = if (in != null) {
```
gerashegalov, line 121: looks like significant duplication between safeFree and safeCLose? Can you try to extract common logic into a common base class, and maybe use a lambda `fn: A => Unit` to implement the difference between close and free?

HaoYang670:
build

HaoYang670:
Close this PR as something weird happened when merging the branch. I will open a new PR.

HaoYang670:
open #5471

Issue #5219, opened 2022-04-12
[FEA] Enable reading floating-point types from CSV and JSON by default


andygrove:
**Is your feature request related to a problem? Please describe.**
Reading floating-point numbers from CSV and JSON is a common operation and we would like it to be enabled by default. It was disabled by default in 22.04 due to cuDF having different precision to Java and therefore results were not the same as Spark. Now that this issue is resolved in https://github.com/rapidsai/cudf/pull/10622 we should be able to enable these types by default.

**Describe the solution you'd like**
Enable reading float and double from CSV and JSON by default and remove the associated configuration options.

**Describe alternatives you've considered**
None

**Additional context**
None

andygrove:
This is now covered by https://github.com/NVIDIA/spark-rapids/pull/5441

Issue #5211, opened 2022-04-12
[BUG] csv_test:test_basic_csv_read FAILED 


NvTimLiu:
**Describe the bug**

csv_test:test_basic_csv_read FAILED on branch-22.06 with `assert_gpu_and_cpu_are_equal_collect` error

```
[2022-04-12T04:54:29.825Z] =========================== short test summary info ============================
[2022-04-12T04:54:29.825Z] FAILED ../../src/main/python/csv_test.py::test_basic_csv_read[true--read_csv_df-simple_float_values.csv-StructType(List(StructField(number,DoubleType,true)))-{'header': 'true'}][APPROXIMATE_FLOAT]
[2022-04-12T04:54:29.825Z] FAILED ../../src/main/python/csv_test.py::test_basic_csv_read[true--read_csv_sql-simple_float_values.csv-StructType(List(StructField(number,DoubleType,true)))-{'header': 'true'}][APPROXIMATE_FLOAT]
[2022-04-12T04:54:29.825Z] FAILED ../../src/main/python/csv_test.py::test_basic_csv_read[true-csv-read_csv_df-simple_float_values.csv-StructType(List(StructField(number,DoubleType,true)))-{'header': 'true'}][APPROXIMATE_FLOAT]
[2022-04-12T04:54:29.825Z] FAILED ../../src/main/python/csv_test.py::test_basic_csv_read[true-csv-read_csv_sql-simple_float_values.csv-StructType(List(StructField(number,DoubleType,true)))-{'header': 'true'}][APPROXIMATE_FLOAT]
[2022-04-12T04:54:29.826Z] FAILED ../../src/main/python/csv_test.py::test_basic_csv_read[false--read_csv_df-simple_float_values.csv-StructType(List(StructField(number,DoubleType,true)))-{'header': 'true'}][APPROXIMATE_FLOAT]
[2022-04-12T04:54:29.826Z] FAILED ../../src/main/python/csv_test.py::test_basic_csv_read[false--read_csv_sql-simple_float_values.csv-StructType(List(StructField(number,DoubleType,true)))-{'header': 'true'}][APPROXIMATE_FLOAT]
[2022-04-12T04:54:29.826Z] FAILED ../../src/main/python/csv_test.py::test_basic_csv_read[false-csv-read_csv_df-simple_float_values.csv-StructType(List(StructField(number,DoubleType,true)))-{'header': 'true'}][APPROXIMATE_FLOAT]
[...]

PR #5220, opened 2022-04-12
WIP: Enable reading floating-point types from CSV and JSON by default

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/5219

Enable reading floating-point types from CSV and JSON by default now that https://github.com/rapidsai/cudf/pull/10622 is merged.

WIP because:

- This does not enable decimal types and we may want to include that as well?
- Need to fix https://github.com/NVIDIA/spark-rapids/issues/5211 first


andygrove:
Closing this in favor of https://github.com/NVIDIA/spark-rapids/pull/5441

PR #5230, opened 2022-04-12
Update csv float tests to reflect changes in precision in cuDF

andygrove:

Closes https://github.com/NVIDIA/spark-rapids/issues/5211

This PR removes one test value that was overflowing to `Inf` as a result of a recent change in precision in cuDF. It also adds exact tests for reading small floating-point values from CSV.

We already have documentation explaining that numbers in certain ranges, such as `1.7976931348623158E308 <= x < 1.7976931348623159E308` will have different behavior between CPU and GPU and the number removed from the test (`1.7976931348623158E308`) was within this range so I don't think any documentation update is required.


andygrove:
build

jlowe (approved):

Issue #5183, opened 2022-04-08
[BUG] UCX EGX integration test array_test.py::test_array_exists failures


tgravescs:
**Describe the bug**
The latest integration tests on EGX with UCX had the following failures:
```
12:48:34  FAILED ../../src/main/python/array_test.py::test_array_exists[3VL:off-data_gen0]
12:48:34  FAILED ../../src/main/python/array_test.py::test_array_exists[3VL:off-data_gen1]

```

`12:48:34  E           AssertionError: GPU and CPU boolean values are different at [1682, 'exists_longer_than_5']`

and aopther one with different values

abellina:
I was able to reproduce this in the same environment, but I didn't run with UCX on.

```
../../src/main/python/array_test.py::test_array_exists[3VL:off-data_gen0][IGNORE_ORDER({'local': True})] ^[[31mFAILED^[[0m^[[31m [ 25%]^[[0m
../../src/main/python/array_test.py::test_array_exists[3VL:off-data_gen1][IGNORE_ORDER({'local': True})] ^[[31mFAILED^[[0m^[[31m [ 50%]^[[0m
../../src/main/python/array_test.py::test_array_exists[3VL:on-data_gen0][IGNORE_ORDER({'local': True})] ^[[32mPASSED^[[0m^[[31m [ 75%]^[[0m
../../src/main/python/array_test.py::test_array_exists[3VL:on-data_gen1][IGNORE_ORDER({'local': True})] ^[[32mPASSED^[[0m^[[31m [100%]^[[0m
```

I digged at the differences a bit more:
```
CPU:
Row(a=[332248246, 545395431, -495984289], exists_even=True, exists_negative=True

GPU:
Row(a=[332248246, 545395431, -495984289], exists_even=True, exists_negative=False
```

```
CPU:
Row(a=[-655713209, 0, None, -132615933, -808871756, None], exists_even=True, exists_negative=True, exists_non_negative=True

GPU:
Row(a=[-655713209, 0, None, -132615933, -808871756, None], exists_even=True, exists_negative=True, exists_non_negative=False
```

It seems that either an array of string or an array of integer fails, with 3VL off (that said 3VL on seems to be ok?)

tgravescs:
I tried this against 22.04 and the test passes there.

tgravescs:
note I also tested with 22.06 from April 1st and it passed the tests, but trying April 5th build it fails.

abellina:
[...]

PR #5232, opened 2022-04-13
Fix issue in GpuArrayExists where a parent view outlived the child

abellina:

Closes https://github.com/NVIDIA/spark-rapids/issues/5183.

This fixes a subtle bug with `GpuArrayExists.legacyExists` that manifests itself with the CUDA async allocator, but is not caused by it.

In the old code, `noNullsChildView` was closed when `replaceChildNullsByFalseView` returned, and it was assumed to be a valid child in the calling code (`legacyExists`). I changed it so this child view is alive and well while we call `existsReduce`.

Note this seems to me like a bug that exists in 22.04 also.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/higherOrderFunctions.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/higherOrderFunctions.scala
@@ -349,12 +349,10 @@ case class GpuArrayExists(
       DType.BOOL8)
   }
 
-  private def replaceChildNullsByFalseView(cv: cudf.ColumnView): cudf.ColumnView = {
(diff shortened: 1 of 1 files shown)


abellina:
@tgravescs @sameerz @jlowe fyi it wasn't a cuDF bug as I originally thought. Should we backport this to 22.04?

abellina:
build

jlowe (approved):

jlowe:
build

gerashegalov (approved):
The change LGTM. 

As for 22.04 it is not a high priority because the call path is off by default on Spark 3.x. But so is the fix is also very localized to this call path, thus it's a strict improvement without risking to break anything that was not broken before.

Issue #5242, opened 2022-04-13
[BUG] Executor falls back to cudaMalloc if the pool can't be initialized


abellina:
While debugging: https://github.com/NVIDIA/spark-rapids/issues/5183, I was confused because some of our runs with 22.06 would "work" before @jlowe [added a fix to cuDF](https://github.com/rapidsai/cudf/pull/10585) to specify that RMM use the statically linked cudart.

The reason why our runs work boiled down to us using a different allocator than what we thought we were using (ASYNC would fail to start) with:

```
ai.rapids.cudf.CudfException: RMM failure at: /home/jenkins/agent/workspace/jenkins-cudf_nightly-dev-github-669-cuda11/cpp/build/_deps/rmm-src/include/rmm/mr/device/cuda_async_memory_resource.hpp:67: cudaMallocAsync not supported with this CUDA driver/runtime version
        at ai.rapids.cudf.Rmm.initializeInternal(Native Method)
```

This caused everything to start, and RMM to create a cuda memory resource by default, which it used for the duration of the test (and the test passed).

We should let this exception tear down the executor, instead of catching-and-log pattern we have now:  https://github.com/NVIDIA/spark-rapids/blob/branch-22.06/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala#L304.

abellina:
If I just remove the catch, I get the executor to bail, but I believe we should log _something_ from our side saying we are exiting now.

e.g.

```
22/04/13 14:39:05 ERROR RapidsExecutorPlugin: Exception in the executor plugin
ai.rapids.cudf.CudfException: RMM failure at: /home/jenkins/agent/workspace/jenkins-cudf-for-dev-32-cuda11/cpp/build/_deps/rmm-src/include/rmm/mr/device/cuda_async_memory_resource.hpp:67: cudaMallocAsync not supported with this CUDA driver/runtime version 
  at ai.rapids.cudf.Rmm.initializeInternal(Native Method)
  at ai.rapids.cudf.Rmm.initialize(Rmm.java:119)
  at com.nvidia.spark.rapids.GpuDeviceManager$.initializeRmm(GpuDeviceManager.scala:300)
  at com.nvidia.spark.rapids.GpuDeviceManager$.initializeMemory(GpuDeviceManager.scala:330)
  at com.nvidia.spark.rapids.GpuDeviceManager$.initializeGpuAndMemory(GpuDeviceManager.scala:137)
[...]

PR #5243, opened 2022-04-13
Throw again after logging that RMM could not intialize

abellina:

Closes: https://github.com/NVIDIA/spark-rapids/issues/5242.

This is a (small) proposed diff to make failure to initialize RMM (which throws a `CudfException`) fatal, instead of silently continuing with a default initialized raw cuda memory resource.

The executor logs will look like this:

```
22/04/13 14:56:06 ERROR GpuDeviceManager: Could not initialize RMM, exiting!
 ai.rapids.cudf.CudfException: RMM failure at: /home/jenkins/agent/workspace/jenkins-cudf-for-dev-32-cuda11/cpp/build/_deps/rmm-src/include/rmm/mr/device/    cuda_async_memory_resource.hpp:67: cudaMallocAsync not supported with this CUDA driver/runtime version
  at ai.rapids.cudf.Rmm.initializeInternal(Native Method)
```

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala
@@ -296,15 +296,9 @@ object GpuDeviceManager extends Logging {
         logInfo("Using legacy default stream")
       }
 
-      try {
-        Cuda.setDevice(gpuId)
-        Rmm.initialize(init, logConf, poolAllocation)
-        RapidsBufferCatalog.init(conf)
-      } catch {
-        case e: CudfException =>
-          logError("Could not initialize RMM, exiting!", e)
-          throw e
-      }
+      Cuda.setDevice(gpuId)
+      Rmm.initialize(init, logConf, poolAllocation)
+      RapidsBufferCatalog.init(conf)
     }
   }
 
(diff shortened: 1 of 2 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDeviceManager.scala
```diff
@@ -301,7 +301,9 @@ object GpuDeviceManager extends Logging {
         Rmm.initialize(init, logConf, poolAllocation)
         RapidsBufferCatalog.init(conf)
       } catch {
-        case e: Exception => logError("Could not initialize RMM", e)
+        case e: CudfException =>
+          logError("Could not initialize RMM, exiting!", e)
```
revans2, line 305: This is typically an anti-pattern

https://stackoverflow.com/questions/6639963/why-is-log-and-throw-considered-an-anti-pattern

Is there a reason we need the log statement? Should we just lat the exception continue up?

revans2: I would rather have the code that exits say we are exiting, otherwise there is coupling between the code that exits and here for no good reason.

abellina: @revans2 handled here: https://github.com/NVIDIA/spark-rapids/pull/5243/commits/f6545d334ecf1dd7e496e12b5212bbc7afe1fc5f

abellina:
build

jlowe (approved):

PR #5249, opened 2022-04-13
Use a bundled spark-rapids-jni dependency instead of external cudf dependency

### LINKED ISSUE #3196: Custom native code support for plugin
Increasingly there has been a desire to have Spark-specific behavior in cudf kernels, but cudf is not a library specific to the RAPIDS Accelerator and therefore should minimize Spark-specific behaviors.  However this does not diminish the need on the RAPIDS Accelerator side, and we have often implemented the required behaviors by hacking together a series of kernels which often uses much more memory and performs worse.
 
We need the ability to provide native GPU code with Spark-specific behaviors.  Like cudf, I think we will always have a CUDA classifier in the jar containing native code.  With that in mind, I'd like to see the RAPIDS Accelerator become a single jar, rather than the plugin+cudf combo jars we have today, that contains everything needed to run the plugin with Spark, i.e.
- Scala and Java code for the plugin
- Plugin-specific native code and CUDA kernels
- corresponding cudf jar contents
 
We could keep cudf as a separate, user-visible dependency as we do today and have the RAPIDS Accelerator jar add a cuda classifier.  However once the RAPIDS Accelerator jar also has a classifier like cudf, I do not see any reason to continue to keep these jars separate and make mismatched plugin+cudf jar combinations a possibility.

jlowe:
Fixes #3196.

This changes the RAPIDS Accelerator to be a self-contained jar, including the spark-rapids-jni and cudf code within the jar so a separate RAPIDS cudf jar is no longer necessary to be deployed.  This enables the project to use Spark-specific logic in custom kernels implemented in the [spark-rapids-jni project](https://github.com/NVIDIA/spark-rapids) which is in turn built on top of the [RAPIDS cudf project](https://github.com/rapidsai/cudf).

Note that because the RAPIDS Accelerator jar now contains native CUDA code, a CUDA version classifier has been added to the jar artifact.

Documentation has been updated to reflect the new model for using the RAPIDS Accelerator.

I left the check for cudf versions in place.  Arguably it's no longer needed because we're now bundling cudf via the bundled spark-rapids-jni jar, but it could prove useful if the user drops an older cudf jar in front of the RAPIDS Accelerator jar in the classpath.

This also fixes the multiple copies of the jucx.so library that were included in the dist jar, one copy per supported Spark version.


jlowe:
build

pom.xml
```diff
@@ -792,7 +792,7 @@
         <spark.test.version>${spark.version}</spark.test.version>
         <spark.version.classifier>spark${buildver}</spark.version.classifier>
         <cuda.version>cuda11</cuda.version>
-        <cudf.version>22.06.0-SNAPSHOT</cudf.version>
+        <spark-rapids-jni.version>22.06.0-SNAPSHOT</spark-rapids-jni.version>
```
tgravescs, line 795: we need to update the deploy script to push the new jar with classifier. I assume we have to push both classifier and non-classifier version.  @pxLi  perhaps has more insight.

pxLi: Yeah, currently we have both in our nightly deployment
https://oss.sonatype.org/content/repositories/snapshots/com/nvidia/spark-rapids-jni/22.06.0-SNAPSHOT/

pxLi:
@jlowe can we also update the 

> overall looks good. I'm assuming we will have IT builds to update to stop fetching the cudf jar as followups.

I will take care of that after this one got merged for both github and internal repos https://github.com/NVIDIA/spark-rapids/issues/5258

pxLi:
hi @jlowe 

Do you think if we need to link the empty jar to some cuda classifier one (like cuda11) by default? Or just remove the empty jar after build to avoid potential confusion? thx~

image
image

pom.xml
pxLi: also I saw json has already update default jar to classifier one, so we may just need to confirm it works w/ all our deploy scripts https://github.com/NVIDIA/spark-rapids/pull/5249/files#diff-2681b51a40a6a26d3a6f7f839cd810422226193ef273ea02d1b0b93c3a87cfa6R220-R222

jlowe:
Hmm, I'm not sure how the empty jar is being built.  I don't see it when build locally with either mvn or buildall.

> Do you think if we need to link the empty jar to some cuda classifier one (like cuda11) by default? Or just remove the empty jar after build to avoid potential confusion for developers?

We should do the same thing we do for cudf, i.e.: post the cuda11 classifier as the no-classifier build.  IIRC I believe we have to have a no-classifier build to publish based on the repository policy, @GaryShen2008 may know the details there.  If we don't have to publish a no-classifier version that's preferable, but if we do then we should have it match the base version classifier (i.e.: cuda11 for now).

pxLi (approved):

NvTimLiu:
>  IIRC I believe we have to have a no-classifier build to publish based on the repository policy

Correct, we have to have a no-classifier build to publish, otherwise, OSS package validation would FAILED on : 

```
Event: Failed: Javadoc Validation

failureMessage  Missing: no main jar artifact found in folder '/com/nvidia/rapids-4-spark_2.12/22.06.0'
```

Issue #5148, opened 2022-04-06
Add the MULTI-THREADED reading support for avro


firestarman:
The base reading for avro is ready, now it is time to support the MULTI-THREADED reading to try to get better performance when reading from cloud, just like what is done in GPU parquet reading.

PR #5255, opened 2022-04-14
Support multi-threaded reading for avro[databricks]

firestarman:
This PR is to enable the multi-threaded reading for avro.

It has mainly
- added 4 relevant configs.
- created the classes for the multi-threaded reading factory and reader, along with some utils.
- done some small refactor to reduce some duplicated code.
- updated the avro tests.

closes https://github.com/NVIDIA/spark-rapids/issues/5148


firestarman:
build

tgravescs:
can you file an issue for this?  By description if this is multi-threaded, are we doing coalescing as well in the future - clarify in issue please.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -941,6 +936,44 @@ object RapidsConf {
     .booleanConf
     .createWithDefault(false)
 
+  val AVRO_READER_TYPE = conf("spark.rapids.sql.format.avro.reader.type")
+    .doc("Sets the avro reader type. We support different types that are optimized for " +
+      "different environments. The original Spark style reader can be selected by setting this " +
+      "to PERFILE which individually reads and copies files to the GPU. Loading many small files " +
+      "individually has high overhead, and using MULTITHREADED is recommended instead." +
```
jlowe, line 943:
```suggestion
      "individually has high overhead, and using MULTITHREADED is recommended instead. " +
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMultiFileReader.scala
```diff
@@ -114,6 +114,24 @@ object MultiFileThreadPoolUtil {
...
 }
 
+/** A factory class to get a thread pool */
+class MultiFileThreadPoolFactory {
+  private var threadPool: Option[ThreadPoolExecutor] = None
+
+  private def initThreadPool(
+      threadTag: String,
+      numThreads: Int): ThreadPoolExecutor = synchronized {
+    if (threadPool.isEmpty) {
```
jlowe, line 124: Nit: Why is this checking for an initialized threadPool?  It's a private method called only in a `getOrElse` context, which means we know it won't be called unless threadPool is empty.  Either that or we should just ditch `initThreadPool` and have its implementation be the code for `getThreadPool`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMultiFileReader.scala
```diff
@@ -114,6 +114,24 @@ object MultiFileThreadPoolUtil {
...
+      threadPool = Some(MultiFileThreadPoolUtil.createThreadPool(threadTag, numThreads))
+    }
+    threadPool.get
+  }
+
+  def getThreadPool(threadTag: String, numThreads: Int): ThreadPoolExecutor = {
```
jlowe, line 130: What happens if someone calls `getThreadPool("foo", 1)` and then later someone calls `getThreadPool("bar", 2)`?  Seems like we're just ignoring the parameters completely if someone else beat us to the call which is a surprising result to the caller who requested a certain number of threads and/or a specific tag name.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/ExternalSource.scala
```diff
@@ -41,17 +43,15 @@ object ExternalSource {
   }
 
   def tagSupportForGpuFileSourceScanExec(meta: SparkPlanMeta[FileSourceScanExec]): Unit = {
+    val format = meta.wrapped.relation.fileFormat
     if (hasSparkAvroJar) {
-      meta.wrapped.relation.fileFormat match {
+       format match {
```
jlowe, line 48:
```suggestion
      format match {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuAvroScan.scala
```diff
@@ -151,92 +151,276 @@ case class GpuAvroPartitionReaderFactory(
...
+  protected def sendToGpu(
+      hostBuf: HostMemoryBuffer,
+      bufSize: Long,
+      splits: Array[PartitionedFile]): Option[ColumnarBatch] = {
+    withResource(hostBuf) { _ =>
+      if (bufSize == 0) {
+        None
+      } else {
+        // Dump buffer for debugging when required
+        dumpDataToFile(hostBuf, bufSize, splits, Option(debugDumpPrefix), Some("avro"))
```
jlowe, line 244: Nit: Rather than create a new Option object every time this is called, why not create it once when the RapidsConf value is read and pass that along to be reused?  Similarly we're creating an "avro" Option object here that is almost never used, would be good to cache this and reuse to avoid unnecessary garbage creation.

wbo4958:
> I need to look a bit more at core logic still.
> 
> What testing was done on this? Did you manually verify the multi-reader is working, did we do any perf test to show its helping? I know we didn't do it for other readers but it would be nice if we had some test to verify we are picking up the right reader and the reader is doing what we expect.

@firestarman, please figure out to add some unit tests for choosing the right reader for 'avro' in `GpuReaderSuites.scala`

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuAvroScan.scala
```diff
@@ -151,92 +151,276 @@ case class GpuAvroPartitionReaderFactory(
...
-  def filterBlocks(partFile: PartitionedFile): AvroBlockMeta = {
+  private val debugDumpPrefix = rapidsConf.avroDebugDumpPrefix
+  private val ignoreMissingFiles = sqlConf.ignoreMissingFiles
+  private val ignoreCorruptFiles = sqlConf.ignoreCorruptFiles
+
+  private val numThreads = rapidsConf.avroMultiThreadReadNumThreads
+  private val maxNumFileProcessed = rapidsConf.maxNumAvroFilesParallel
+
+  // Disable coalescing reading until it is supported.
+  override val canUseCoalesceFilesReader: Boolean = false
```
wbo4958, line 186: Could we enable this and force the reader to multithreaded and add the unit tests in GpuReaderSuite.scala

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuAvroScan.scala
```diff
@@ -273,200 +457,224 @@ class AvroPartitionReader(
...
-    try {
-      if (dataSize == 0) {
-        None
+}
+
+/**
+ * A PartitionReader that can read multiple AVRO files in parallel.
+ * This is most efficient running in a cloud environment where the I/O of reading is slow.
+ */
+class GpuMultiFileCloudAvroPartitionReader(
```
wbo4958, line 477: please add the parameters description

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuAvroScan.scala
```diff
@@ -151,92 +151,276 @@ case class GpuAvroPartitionReaderFactory(
...
-/**
- * Avro block meta info
- *
- * @param header the header of avro file
- * @param blocks the total block info of avro file
- */
-case class AvroBlockMeta(header: Header, blocks: Seq[BlockInfo])
+  /** Estimate the total size from the given block meta */
+  private def estimateOutputSize(blockMeta: AvroBlockMeta): Long = {
+    // For simplicity, we just copy the whole header of AVRO
```
wbo4958, line 341: For now, it's ok to copy the whole header, but from a long-term point of view, We may need to figure out how to just copy the useful information.

wbo4958 (commented):
Overall, LGTM

firestarman:
> can you file an issue for this? By description if this is multi-threaded, are we doing coalescing as well in the future - clarify in issue please.

Nice finding, We already have one but forgot to link to it. Updated.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMultiFileReader.scala
firestarman: This piece of code is not written by me (I just lifted it to a parent class to avoid duplicate code in the 3 format readers), however I think this is a Double-Checked Locking which could get a better perf if `getThreadPool` is called quite much.

docs/configs.md
```diff
@@ -75,7 +75,10 @@ Name | Description | Default Value
...
+<a name="sql.format.avro.reader.type"></a>spark.rapids.sql.format.avro.reader.type|Sets the avro reader type. We support different types that are optimized for different environments. The original Spark style reader can be selected by setting this to PERFILE which individually reads and copies files to the GPU. Loading many small files individually has high overhead, and using MULTITHREADED is recommended instead.MULTITHREADED is good for cloud environments where you are reading from a blobstore that is totally separate and likely has a higher I/O read cost. Many times the cloud environments also get better throughput when you have multiple readers in parallel. This reader uses multiple threads to read each file in parallel and each file is sent to the GPU separately. This allows the CPU to keep reading while GPU is also doing work. See spark.rapids.sql.format.avro.multiThreadedRead.numThreads and spark.rapids.sql.format.avro.multiThreadedRead.maxNumFilesParallel to control the number of threads and amount of memory used. By default this is set to AUTO so we select the reader we think is best.|AUTO
```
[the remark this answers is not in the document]
firestarman: >What happens if user specifies coalescing?

The `checkValues` should fail for this case.

```
.checkValues((RapidsReaderType.values - RapidsReaderType.COALESCING).map(_.toString))
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuAvroScan.scala
firestarman: Yes, we may do it in the later PR.

firestarman:
> @firestarman, please figure out to add some unit tests for choosing the right reader for 'avro' in `GpuReaderSuites.scala`

Added it, thanks for the information.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuAvroScan.scala
firestarman: done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuAvroScan.scala
firestarman: Instead I changed the tests a little to cover the cases for the avro reader type check.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuAvroScan.scala
firestarman: Done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/ExternalSource.scala
firestarman: Done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMultiFileReader.scala
firestarman: Updated the comment and the method name to try to match the actual behavior.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
firestarman: done.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileSourceScanExec.scala
```diff
@@ -579,9 +579,12 @@ case class GpuFileSourceScanExec(
           rapidsConf,
           allMetrics,
           queryUsesInputFile)
-      case _ =>
-        // never reach here
-        throw new RuntimeException(s"File format ${relation.fileFormat} is not supported yet")
+      case f =>
```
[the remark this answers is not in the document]
firestarman: updated.

firestarman:
>What testing was done on this? Did you manually verify the multi-reader is working, did we do any perf test to show its helping?
>I know we didn't do it for other readers but it would be nice if we had some test to verify we are picking up the right reader and the reader is doing what we expect.

Now I have added unit tests in `GpuReaderSuites` to veirfy the reader type being picked up and integration tests for the multi-threaded reading functionality. 
No perf test for now, but I plan to do it after supporting the coalescing. If perf test is required for this PR, I can priority it.

firestarman:
build

docs/configs.md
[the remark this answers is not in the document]
firestarman: Updated the doc.

jlowe (approved):

tgravescs:
> No perf test for now, but I plan to do it after supporting the coalescing. If perf test is required for this PR, I can priority it.

Personally I find it odd to put in something that is supposed to be for performance without showing it will help.  I guess in this case we are pretty confident based on other readers already using it and I as long as we are doing it before 22.06 ships I guess I"m ok, but prefer not to see it in the future.

firestarman:
Waiting for benchmark numbers, and @HaoYang670 is helping on this.

firestarman:
According to the tests we ran, the perf is quite bad for cloud envs. We have a solution and it will take much time, so close this first.

tgravescs:
@firestarman  can you provide some details about the issue and then proposed solution if its going to take time?

tgravescs:
looks like described in https://github.com/NVIDIA/spark-rapids/issues/5304

GaryShen2008:
> looks like described in #5304

Yes, @tgravescs could you please review the proposal?

PR #5265, opened 2022-04-17
Remove unused generic types

### LINKED ISSUE #5264: [BUG] Delete unused generic type.
**Describe the bug**
```scala
  implicit class ReallyAGpuExpression[A <: Expression](exp: Expression) {
    def columnarEval(batch: ColumnarBatch): Any = {
      exp.asInstanceOf[GpuExpression].columnarEval(batch)
    }
  }

  implicit class AutoCloseableColumn[A <: AutoCloseable](autoCloseable: AutoCloseable) {

    /**
     * safeClose: Is an implicit on AutoCloseable class that tries to close the resource, if an
     * Exception was thrown prior to this close, it adds the new exception to the suppressed
     * exceptions, otherwise just throws
     *
     * @param e Exception which we don't want to suppress
     */
    def safeClose(e: Throwable = null): Unit = {
      if (autoCloseable != null) {
        if (e != null) {
          try {
            autoCloseable.close()
          } catch {
            case suppressed: Throwable => e.addSuppressed(suppressed)
          }
        } else {
          autoCloseable.close()
        }
      }
    }
  }
  ```
The generic type `A` is not used in the class body. We should remove them.

### PR DESCRIPTION
**Pull request by HaoYang670**
Close #5264 

Just to remove some unused type annotations in the implicit classes.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/implicits.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/implicits.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by HaoYang670** [comment]
build

**Review by jlowe — APPROVED** [review]

File: sql-plugin/src/main/scala/com/nvidia/spark/rapids/implicits.scala [hunk 1]
```diff
@@ -29,13 +29,13 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
  */
 object RapidsPluginImplicits {
 
-  implicit class ReallyAGpuExpression[A <: Expression](exp: Expression) {
+  implicit class ReallyAGpuExpression(exp: Expression) {
```
gerashegalov, line 32: I think this was meant to be future-proof 
```
implicit class ReallyAGpuExpression[E <: Expression](exp: E) {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/implicits.scala
```diff
@@ -29,13 +29,13 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
...
     def columnarEval(batch: ColumnarBatch): Any = {
       exp.asInstanceOf[GpuExpression].columnarEval(batch)
     }
   }
 
-  implicit class AutoCloseableColumn[A <: AutoCloseable](autoCloseable: AutoCloseable) {
+  implicit class AutoCloseableColumn(autoCloseable: AutoCloseable) {
```
gerashegalov, line 38: similarly
```
implicit class AutoCloseableColumn[A <: AutoCloseable](autoCloseable: A) {
```

PR #5268, opened 2022-04-18
[BUG] When ASYNC is enabled GDS needs to handle cudaMalloced bounce buffers

abellina:
Given the ASYNC allocator is enabled, the UCX bounce buffers are allocated directly, bypassing the ASYNC pool (because memory from the async pool can't be mapped for purposes of GPUDirectRDMA).

This PR fixes GDS copies where it assumed the bounce buffer was a `DeviceMemoryBuffer` when it was really a `CudaMemoryBuffer` (e.g. straight from cudaMalloc). It also fixes a couple of leaks where `.slice` was used, but the sliced buffer was not closed in the stack.


jlowe (approved):

jlowe:
build

PR #5276, opened 2022-04-19
fix markdown check issue[skip ci]

nvliyuan:
fix markdown checker permissions issue and reference links issue

### DIFF
--- a/.github/workflows/markdown-links-check.yml
+++ b/.github/workflows/markdown-links-check.yml
@@ -24,7 +24,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@master
-    - uses: gaurav-nelson/github-action-markdown-link-check@v1
+#    - uses: gaurav-nelson/github-action-markdown-link-check@v1
+    - uses: nvliyuan/github-action-markdown-link-check@fix-markdown-check-fail
       with:
         max-depth: -1
         use-verbose-mode: 'yes'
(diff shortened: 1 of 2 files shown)


nvliyuan:
this pr is to fix the GitHub permissions issue, but I believe there are some other issues in the link check library, because there are some reference links marked as dead, but actually, they can be accessed, please see the [log](https://github.com/NVIDIA/spark-rapids/runs/6071383537?check_suite_focus=true) for more detail
image
and this link is not dead
image
I can not find a perfect workaround  or fix for now

sameerz:
I have a concern that the updated link check is referring to a personal github repository.  I notice other repos that use gaurav-nelson/github-action-markdown-link-check have fixed this issue without creating forks.  Is it possible for us to explore a bit further? 

Also, some links are legitimately broken, and without this check we are missing those.  For example, the links on https://github.com/NVIDIA/spark-rapids/blob/branch-22.06/docs/dev/shims.md for the following text are broken: 
1. [ShimExpression For 3.0.x and 3.1.x](https://github.com/NVIDIA/spark-rapids/blob/main/sql-plugin/src/main/post320-treenode/scala/com/nvidia/spark/rapids/shims/v2/TreeNode.scala#L23)
2. [ShimExpression For 3.2.x](https://github.com/NVIDIA/spark-rapids/blob/main/sql-plugin/src/main/pre320-treenode/scala/com/nvidia/spark/rapids/shims/v2/TreeNode.scala#L23)

nvliyuan:
add a safe directory before the actions/checkout step to work around the permission issue.

jlowe (approved):

jlowe:
build

PR #5289, opened 2022-04-21
Re-enable dollar ($) line anchor in regular expressions in find mode 

### LINKED ISSUE #4533: [FEA] Re-enable support for `$` in regular expressions
PR https://github.com/NVIDIA/spark-rapids/pull/4530 will remove support for `$` due to correctness issues. We need to re-implement it.

The challenge is that $ has different semantics between Java and cuDF.

Given the input `a\n` and the pattern `a$`, Java will find a match but cuDF will not.

There was a previous effort to transpile `$` to `[\r]?[\n]?$` and this worked for some cases but not all.

Here is a test that can be used to test an implementation:

```scala
test("line anchor $ - find") {
  val patterns = Seq("$\r", "a$", "\r$", "\n$", "\r\n$", "[\r\n]?$")
  val inputs = Seq("a", "a\n", "a\r", "a\r\n", "\r", "\n", "\r\n", "\n\r")
}
```

See above.

None

**Additional context**
This is related to cuDF issue https://github.com/rapidsai/cudf/issues/9620

NVnavkumar:
Fixes #4533.

This re-enables support for the end of line anchor ($) in regular expressions.  A couple of caveats:

1) This only enables support in `RegexFindMode`. A separate issue will need to filed for Replace and Split modes (if necessary).

2) Apache Spark actually only uses regular expressions in standard mode (not multiline). Which means that we only support the `$` as defined in that mode. This will not match the `$` in multiline mode.

3) This code handles how the line terminators are managed around the line anchor. The $ has different matching characteristics when used with line terminator characters in the regular expression. Particularly here are a couple of examples:
  * using the $ anchor with characters other than a line terminator sequence, the dollar then needs to include the optional matching of possible line terminator sequences (defined [here](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) in the **Line terminators** section). 
  * using the $ anchor with line terminator characters changes the behavior of the regular expression to potentially force the matching of a specific line terminator, there are 4\ cases to call out here:
     * `\r$` - this means to only match the CR before the end of the string, so no need to transpile in this case, just match the end of the string using `$` in cuDF, and the strings are equivalent
     * `[any other line terminator character including \n]$` - this means match that line terminator character plus optionally any other valid line terminator character before the end of the string `$`
     * `$\n` - this means to only match the LF (newline) before the end of the string, and no other line termination sequence. This requires the underlying cuDF to support negative lookahead groups, so this case will fall back to the CPU. (this is because `\r\n` is a valid line terminator sequence, and this means to forcefully not support that sequence which can only be handled by a negative lookahead group -- See [cudf#3100](https://github.com/rapidsai/cudf/issues/3100) on lookaheads)
     * `[any other line character including \r]$` - this means match that line terminator character plus optionally any other valid line terminator character before the end of the string `$`
  * multiple `$$` in a row are handled by Java by just reducing to 1 `$`


andygrove:
This is looking good. Should we also include form-feed `\f` in the tests?

NVnavkumar:
build

NVnavkumar:
build

andygrove (approved):

PR #5290, opened 2022-04-21
Add documentation about reporting security issues [skip ci]

sameerz:
Added documentation about reporting security issues to NVIDIA in the code repository and gh-pages documentation.


SECURITY.md
```diff
@@ -0,0 +1,17 @@
+## Security
+
+NVIDIA is dedicated to the security and trust of our software products and services, including all
```
jlowe, line 3: Should we simply have this document point readers to docs/security.md for details so we don't have to keep both copies in sync?  Oddly they already are not quite in sync, at least wrt. line wrapping.

jlowe (approved):

SECURITY.md
sameerz: I understand the desire not to duplicate.  It just seems odd to have to click on a link to go deeper when looking for information related to reporting a security issue, rather than having the information immediately available.  Would a symlink from SECURITY.md to docs/security.md work? 

I fixed the spacing mismatch in the meantime.

jlowe: I'm fine with leaving it duplicated for now.  If we find this gets updated often we can reconsider.

sameerz:
build

PR #5293, opened 2022-04-21
Support array_repeat

### LINKED ISSUE #5226: [FEA]Support array_repeat
I wish we can support array_repeat.

eg:
```
from pyspark.sql.functions import *
df = spark.createDataFrame([(["a", "b", "a"], ["b", "c"]), (["a","a"], ["b", "c"]), (["aa"], ["b", "c"])    ], ['x', 'y'])
df = spark.read.parquet("/tmp/testparquet")
df.select(array_repeat(df.x, 3).alias("repeat")).collect()
```

```
    ! <ArrayRepeat> array_repeat(x#72, 3) cannot run on GPU because GPU does not currently support the operator class org.apache.spark.sql.catalyst.expressions.ArrayRepeat
```

sperlingxx:

Closes #5226

This PR is to support array_repeat, which relies on unmerged cuDF PR(https://github.com/rapidsai/cudf/pull/10683).
The primary issue of `array_repeat` is to workaround the null and negative count values, which are supported by Spark but invalid in cuDF repeat.


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -508,6 +508,91 @@ case class GpuArrayMax(child: Expression) extends GpuUnaryExpression with Implic
...
+case class GpuArrayRepeat(left: Expression, right: Expression) extends GpuBinaryExpression {
+
+  override def dataType: DataType = ArrayType(left.dataType, left.nullable)
+
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector = {
+    // The primary issue of array_repeat is to workaround the null and negative count.
+    // Spark returns a null (list) when encountering a null count, while cudf::repeat simply
+    // throws an exception.
+    // Spark returns a empty list when encountering a negative count, while cudf::repeat simply
+    // throws an exception.
```
ttnghia, line 520: @jlowe Is this an example that we can completely workaround by re-implementing `cudf::repeat` in `spark-rapids-jni`?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -508,6 +508,91 @@ case class GpuArrayMax(child: Expression) extends GpuUnaryExpression with Implic
...
+      withResource(rhs.getBase.replaceNulls(one)) { notNull =>
+        withResource(GpuScalar.from(0, DataTypes.IntegerType)) { zero =>
+          withResource(notNull.lessThan(zero)) { lessThanZero =>
+            lessThanZero.ifElse(zero, notNull)
+          }
+        }
+      }
+    }
+    // Step 2. perform cuDF repeat
+    val repeated = closeOnExcept(refinedCount) { cnt =>
```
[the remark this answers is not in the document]
jlowe: Any time we call cudf code we need to anticipate a potential exception.  For example, the `Table` constructor could throw, or the `table.repeat` call could throw.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -508,6 +508,91 @@ case class GpuArrayMax(child: Expression) extends GpuUnaryExpression with Implic
...
+      withResource(new Table(lhs.getBase)) { table =>
+        table.repeat(cnt, true).getColumn(0)
+      }
+    }
+    // Step 3. generate list offsets from refined counts
+    val offsets = withResource(refinedCount) { cnt =>
```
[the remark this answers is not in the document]
sperlingxx: Yes, I added the missing the closeOnEx wrapper for `repeated`

integration_tests/src/main/python/array_test.py
```diff
@@ -284,6 +284,33 @@ def test_array_max(data_gen):
...
+
+@pytest.mark.parametrize('data_gen', orderable_gens + nested_gens_sample, ids=idfn)
+def test_array_repeat_with_count_scalar(data_gen):
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: unary_op_df(spark, data_gen).selectExpr(
+            'array_repeat(a, 3)',
+            'array_repeat(a, 1)',
+            'array_repeat(a, 0)',
+            'array_repeat(a, -2)',
+            'array_repeat("abc", 2)',
```
revans2, line 309: Because both parameters are scalars the call is probably replaced with a single scalar array at compile time. Have you checked that the test plan is covering what you want/expect?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -508,6 +508,91 @@ case class GpuArrayMax(child: Expression) extends GpuUnaryExpression with Implic
...
+
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector = {
+    // The primary issue of array_repeat is to workaround the null and negative count.
+    // Spark returns a null (list) when encountering a null count, while cudf::repeat simply
+    // throws an exception.
+    // Spark returns a empty list when encountering a negative count, while cudf::repeat simply
+    // throws an exception.
+
+    // Step 1. replace invalid counts
+    //  null -> 1
```
revans2, line 523: Why 1? a null array typically should look like an empty array but with validity set for it to be null.

integration_tests/src/main/python/array_test.py
sperlingxx:
```
GpuColumnarToRow false
+- GpuProject [gpuarrayrepeat(a#16, 3) AS array_repeat(a, 3)#18, gpuarrayrepeat(a#16, 1) AS array_repeat(a, 1)#19, gpuarrayrepeat(a#16, 0) AS array_repeat(a, 0)#20, gpuarrayrepeat(a#16, -2) AS array_repeat(a, -2)#21, [abc,abc] AS array_repeat(abc, 2)#22, [] AS array_repeat(abc, 0)#23, [] AS array_repeat(abc, -1)#24]
   +- GpuRowToColumnar targetsize(104857600)
      +- *(1) Scan ExistingRDD[a#16]
```
From the dumped plan, the optimizing of constant exprs doesn't seem to happen.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
sperlingxx: It should be 0. I fixed this problem.

sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -508,6 +508,91 @@ case class GpuArrayMax(child: Expression) extends GpuUnaryExpression with Implic
...
+    // The primary issue of array_repeat is to workaround the null and negative count.
+    // Spark returns a null (list) when encountering a null count, while cudf::repeat simply
+    // throws an exception.
+    // Spark returns a empty list when encountering a negative count, while cudf::repeat simply
+    // throws an exception.
+
+    // Step 1. replace invalid counts
+    //  null -> 0
+    //  negative values -> 0
+    val refinedCount = withResource(GpuScalar.from(0, DataTypes.IntegerType)) { zero =>
```
revans2, line 525: Question. Did you intend to use refinedCount with `table.repeat`? If not then this is not used until after we do the repeat, so it might be cleaner to just switch steps one and two so we calculate refinedCount closer to where it is used?  This is just a nit by the way. The code looks fine.

revans2 (approved):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
sperlingxx: Yes, I think we should use refinedCount to run `table.repeat`, because `table.repeat` can not handle count values of negative and null .

sperlingxx:
build

sperlingxx:
build

PR #5297, opened 2022-04-21
GpuStringSplit now honors the`spark.rapids.sql.regexp.enabled` configuration option

### LINKED ISSUE #5130: [BUG] string_split does not respect spark.rapids.sql.regexp.enabled config
**Describe the bug**
`string_split` does not respect the `spark.rapids.sql.regexp.enabled` configuration and will execute regular expressions on the GPU when this config is set to false.

Note that we should continue to ignore the config flag in the case where the delimiter can be transpiled to a simple string.

``` scala
scala> spark.conf.set("spark.rapids.sql.regexp.enabled", "false")

scala> val df = Seq("hello", "goodbyte").toDF("a").repartition(2)

scala> df.createTempView("t")

scala> spark.sql("SELECT split(a, '[eh]') FROM t").show
22/04/01 23:03:54 WARN GpuOverrides: 
!Exec <CollectLimitExec> cannot run on GPU because the Exec CollectLimitExec has been disabled, and is disabled by default because Collect Limit replacement can be slower on the GPU, if huge number of rows in a batch it could help by limiting the number of rows transferred from GPU to CPU. Set spark.rapids.sql.exec.CollectLimitExec to true if you wish to enable it
  @Partitioning <SinglePartition$> could run on GPU
    *Expression <Alias> cast(split(a#4, [eh], -1) as string) AS split(a, [eh], -1)#44 will run on GPU
      *Expression <Cast> cast(split(a#4, [eh], -1) as string) will run on GPU
        *Expression <StringSplit> split(a#4, [eh], -1) will run on GPU
    *Exec <ShuffleExchangeExec> will run on GPU
      *Partitioning <RoundRobinPartitioning> will run on GPU
      ! <LocalTableScanExec> cannot run on GPU because GPU does not currently support the operator class org.apache.spark.sql.execution.LocalTableScanExec
        @Expression <AttributeReference> a#4 could run on GPU

+------------------+
|split(a, [eh], -1)|
+------------------+
|         [, , llo]|
|       [goodbyt, ]|
+------------------+

```

**Expected behavior**
Should fall back to CPU if delimiter is a regular expression and when regexp is disabled.

N/A

**Additional context**
None

NVnavkumar:
Fixes #5130.

This allows GpuStringSplit to honor the `spark.rapids.sql.regexp.enabled` configuration flag. Desired behavior is to fallback to CPU when using a valid regular expression, but still use the GPU when using a simple string or when the pattern can be converted into a simple string.  This is done by using the existing `checkRegExp` method that already handles the logic when the transpilation to a simple string will still work, and also when the parameter is not an actual regular expression.  

Here's some PySpark code to test this branch:

```
from pyspark.sql.functions import expr
from pyspark.sql.types import StructType,StructField, StringType, IntegerType

spark = SparkSession.builder \
      .master("local[*]") \
      .appName("regex_bracket") \
      .config("spark.rapids.sql.explain","ALL") \
      .config("spark.rapids.sql.regexp.enabled", "false") \
      .config("spark.rapids.sql.castStringToTimestamp.enabled","true") \
      .config("spark.rapids.sql.exec.CollectLimitExec", "true") \
      .config("spark.rapids.sql.castFloatToString.enabled", "true") \
      .config("spark.rapids.sql.hasExtendedYearValues", "false") \
      .getOrCreate()  


data2 = [("abc|123abc|123",""),
    ("xyz|123xyz|123","Rose")
  ]

schema = StructType([ \
    StructField("firstname",StringType(),True), \
    StructField("middlename",StringType(),True) \
  ])
 
df = spark.createDataFrame(data=data2,schema=schema)
df.printSchema()
df.show(truncate=False)

regex = r"[a-z]\\|[0-9]"

# This code will fallback to the CPU
df2 = df.withColumn('newcol1', expr(f"""   split(firstname, "{regex}", 2)    """))
df2.explain()
df2.show(truncate=False)

transpilable = r"\\|"

# This code will continue to run on the GPU (Note that | is an escaped regular expression meta character but here is used as a simple delimeter)
df3 = df.withColumn('newcol2', expr(f"""   split(firstname, "{transpilable}", 3)    """))
df3.explain()
df3.show(truncate=False)


simple = "1"

# This code will continue to run on the GPU (this is a simple string)
df4 = df.withColumn('newcol2', expr(f"""   split(firstname, "{simple}", 3)    """))
df4.explain()
df4.show(truncate=False)
```

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
@@ -1419,7 +1419,9 @@ class GpuStringSplitMeta(
 
   override def tagExprForGpu(): Unit = {
     checkRegExp(expr.regex) match {
-      case Some((p, isRe)) => pattern = p; isRegExp = isRe
+      case Some((p, isRe)) =>
+        pattern = p
+        isRegExp = isRe
       case _ => throwUncheckedDelimiterException()
     }
     
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
@@ -1414,10 +1414,20 @@ class GpuStringSplitMeta(
     extends StringSplitRegExpMeta[StringSplit](expr, conf, parent, rule) {
   import GpuOverrides._
 
-  private var delimInfo: Option[(String, Boolean)] = None
+  private var pattern = ""
+  private var isRegExp = false
 
   override def tagExprForGpu(): Unit = {
-    delimInfo = checkRegExp(expr.regex)
+    checkRegExp(expr.regex) match {
+      case Some((p, isRe)) => pattern = p; isRegExp = isRe
+      case _ => throwUncheckedDelimiterException()
+    }
+    
+    // if this is a valid regular expression, then we should check the configuration
+    // whether to run this on the GPU
+    if (isRegExp) {
+      GpuRegExpUtils.tagForRegExpEnabled(this)
+    }
 
     extractLit(expr.limit) match {
       case Some(Literal(n: Int, _)) =>
@@ -1434,8 +1444,7 @@ class GpuStringSplitMeta(
       str: Expression,
       regexp: Expression,
       limit: Expression): GpuExpression = {
-    val delim: (String, Boolean) = delimInfo.getOrElse(throwUncheckedDelimiterException())
-    GpuStringSplit(str, regexp, limit, delim._1, delim._2)
+    GpuStringSplit(str, regexp, limit, pattern, isRegExp)
   }
 }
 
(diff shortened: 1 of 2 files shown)


NVnavkumar:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -1414,10 +1414,20 @@ class GpuStringSplitMeta(
...
   import GpuOverrides._
 
-  private var delimInfo: Option[(String, Boolean)] = None
+  private var pattern = ""
+  private var isRegExp = false
 
   override def tagExprForGpu(): Unit = {
-    delimInfo = checkRegExp(expr.regex)
+    checkRegExp(expr.regex) match {
+      case Some((p, isRe)) => pattern = p; isRegExp = isRe
```
andygrove, line 1422: nit: I'm not used to seeing semicolons in Scala and I think the code would be easier to read without that

```suggestion
      case Some((p, isRe)) =>
        pattern = p
        isRegExp = isRe
```

andygrove:
Could we add an integration test to confirm we are falling back to CPU when regex is disabled?

andygrove (dismissed):

NVnavkumar:
build

andygrove (approved):

NVnavkumar:
build

PR #5301, opened 2022-04-22
Support existence join type for broadcast nested loop join

### LINKED ISSUE #5034: [FEA] Implement ExistenceJoin for BroadcastNestedLoopJoin Exec
When the existential subquery is correlated but does not use equality, BNL may be chosen as an underlying physical Join. This is a contrived example:
  
```python
spark.createDataFrame([
    [1, 101, 2500],
    [2, 102, 1110], 
    [3, 103, 500], 
    [4, 104, 400], 
    [5, 105, 150], 
    [6, 106, 450]],
    'paymentId byte, cId int, amount long'
).createOrReplaceTempView('payments')

spark.createDataFrame([
    [101, 'Jon'], 
    [102, 'Aron'],
    [103, 'Sam'],
    [107, 'Jack']],
    'cId int, name string'
).createOrReplaceTempView('customers')

query_2="""
SELECT *
  FROM payments p
 WHERE EXISTS
         (SELECT c.cId
            FROM customers c    
           WHERE c.cId < p.cId)
    OR p.amount > 500
"""
sql(query_2).explain(True)
```
yielding the plan:
```
...
== Analyzed Logical Plan ==
paymentId: tinyint, cId: int, amount: bigint
Project [paymentId#128, cId#129, amount#130L]
+- Filter (exists#510 [cId#129] OR (amount#130L > cast(500 as bigint)))
   :  +- Project [cId#134]
   :     +- Filter (cId#134 < outer(cId#129))
   :        +- SubqueryAlias c
   :           +- SubqueryAlias customers
   :              +- View (`customers`, [cId#134,name#135])
   :                 +- LogicalRDD [cId#134, name#135], false
   +- SubqueryAlias p
      +- SubqueryAlias payments
         +- View (`payments`, [paymentId#128,cId#129,amount#130L])
            +- LogicalRDD [paymentId#128, cId#129, amount#130L], false

== Optimized Logical Plan ==
Project [paymentId#128, cId#129, amount#130L]
+- Filter (exists#514 OR (amount#130L > 500))
   +- Join ExistenceJoin(exists#514), (cId#134 < cId#129)
      :- LogicalRDD [paymentId#128, cId#129, amount#130L], false
      +- Project [cId#134]
         +- LogicalRDD [cId#134, name#135], false

== Physical Plan ==
GpuColumnarToRow false
+- GpuProject [paymentId#128, cId#129, amount#130L]
   +- GpuCoalesceBatches targetsize(2147483647)
      +- GpuFilter (exists#514 OR (amount#130L > 500)), true
            +- BroadcastNestedLoopJoin BuildRight, ExistenceJoin(exists#514), (cId#134 < cId#129)
               :- *(1) Scan ExistingRDD[paymentId#128,cId#129,amount#130L]
               +- BroadcastExchange IdentityBroadcastMode, [id=#6765]
                  +- GpuColumnarToRow false
                     +- GpuProject [cId#134]
                           +- *(2) Scan ExistingRDD[cId#134,name#135]
```

falls back on CPU when it comes to ExistenceJoin:

```
22/03/24 00:23:59 WARN GpuOverrides: 
  *Exec <FilterExec> will run on GPU
    *Expression <Or> (exists#15 OR (amount#3L > 500)) will run on GPU
      *Expression <GreaterThan> (amount#3L > 500) will run on GPU
    !Exec <BroadcastNestedLoopJoinExec> cannot run on GPU because ExistenceJoin(exists#15) currently is not supported
      @Expression <LessThan> (cId#7 < cId#2) could run on GPU
        @Expression <AttributeReference> cId#7 could run on GPU
        @Expression <AttributeReference> cId#2 could run on GPU
      ! <RDDScanExec> cannot run on GPU because GPU does not currently support the operator class org.apache.spark.sql.execution.RDDScanExec
        @Expression <AttributeReference> paymentId#1 could run on GPU
        @Expression <AttributeReference> cId#2 could run on GPU
        @Expression <AttributeReference> amount#3L could run on GPU
      !Exec <BroadcastExchangeExec> cannot run on GPU because the BroadcastNestedLoopJoin this feeds is not on the GPU
          ! <RDDScanExec> cannot run on GPU because GPU does not currently support the operator class org.apache.spark.sql.execution.RDDScanExec
            @Expression <AttributeReference> cId#7 could run on GPU
            @Expression <AttributeReference> name#8 could run on GPU
```

Execute ExistenceJoin BNL on GPU
 
N/A

res-life:
Closes #5034 
Support existence join type for broadcast nested loop join


res-life:
build

res-life:
JoinGatherer for existence join
Existence join generates an `exists` boolean column with `true` or `false` in it,
then appends it to the `output` columns. It does not shrink or expand left table.
Just provides the `exists` column for the next following operator(usually filter).

```
e.g.:
select * from left_table where
  left_table.column_0 >= 3
  or
  exists (select * from right_table where left_table.column_1 < right_table.column_1)

Explanation of this sql is:

Filter(left_table.column_0 >= 3 or `exists`)
  Existence_join // generate `exists` column, do not shrink or expand the rows of left table
    left_table
    right_table
```

This is to add a new JoinGatherer to handle this existence join.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
```diff
@@ -670,4 +675,133 @@ case class MultiJoinGather(left: JoinGatherer, right: JoinGatherer) extends Join
...
+        }
+      }
+    }
+
+    val ret = withResource(subTableCbTmp) { subTableCb =>
+      val subExistsCvTmp = withResource(gatherMap.toColumnView(0, gatherMap.getRowCount.toInt)) {
+        gatherView =>
+          // `exists.numRows` == `batch.numRows`,
+          //  with true or false in it indicating if the row is gathered
+          withResource(genExistsColumn(gatherView, batch.numRows())) { exists =>
```
firestarman, line 749: Personally I perfer to generate the column of existence according to the `subTableCbTmp`, not the whole batch. Then there is no need to split it next. Instead you can append the generated 'existence' column to the sub table directly.

firestarman (commented):
Took a quick review, and got some comments.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
sperlingxx: Just as @firestarman mentioned, this may lead to a lot of duplicated runs of full-scale scatter. IIUC, if we split the gather map as left table, we need to transform the global offset to the local offset  before scattering. Perhaps there is another approach, just cache the exist column for reuse over mini batches.

res-life:
build

integration_tests/src/main/python/join_test.py
```diff
@@ -841,3 +841,28 @@ def do_join(spark):
             "spark.sql.adaptive.enabled": aqeEnabled,
             "spark.sql.autoBroadcastJoinThreshold": bhjThreshold
         })
+
+
+@ignore_order
+@pytest.mark.parametrize('aqeEnabled', [True, False], ids=['AQE_ON', 'AQE_OFF'])
+def test_existence_join_in_broadcast_nested_loop_join(spark_tmp_table_factory, aqeEnabled):
```
jlowe, line 848: This is unlikely to test the handling of splitting the join output due to batch size limits.  There should be a test, either as another dimension of this test or a separate test, that will cause output batches to be split.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
```diff
@@ -670,4 +677,137 @@ case class MultiJoinGather(left: JoinGatherer, right: JoinGatherer) extends Join
...
+ *     left_table
+ *     right_table
+ * </code>
+ */
+class ExistenceJoinGathererImpl(
+    private val gatherMap: LazySpillableGatherMap,
+    private val data: LazySpillableColumnarBatch,
+    boundsCheckPolicy: OutOfBoundsPolicy,
+    spillCallback: Option[SpillCallback])
+    extends JoinGathererImpl(gatherMap, data, boundsCheckPolicy) {
```
jlowe, line 708: Was this intended to override `JoinGatherer`?  If not, it may make sense to have some of the members of `JoinGathererImpl` marked `protected` so we don't have to copy the code here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
```diff
@@ -670,4 +677,137 @@ case class MultiJoinGather(left: JoinGatherer, right: JoinGatherer) extends Join
...
+ *
+ * Filter(left_table.column_0 >= 3 or `exists`)
+ *   Existence_join // generate `exists` column, do not shrink or expand the rows of left table
+ *     left_table
+ *     right_table
+ * </code>
+ */
+class ExistenceJoinGathererImpl(
+    private val gatherMap: LazySpillableGatherMap,
+    private val data: LazySpillableColumnarBatch,
```
jlowe, line 705: IMO it makes no sense to mark these as `val` when they become "useless" immediately upon construction due to the creation of `lazyLeftTableWithExists`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
```diff
@@ -670,4 +677,137 @@ case class MultiJoinGather(left: JoinGatherer, right: JoinGatherer) extends Join
...
+              LazySpillableColumnarBatch(cb, spillCallback.get, "existence_join_batch")
+            }
+          }
+        }
+      }
+
+      lazyCb.allowSpilling()
+      lazyCb
+    }
+  }
```
jlowe, line 752: Maybe I'm missing something, but this this looks like we're making a spillable form of the complete answer to the join batch, i.e.: the left table with the existence column.  But _we instantiated the entire result here_.  What's the point of trying to split it up afterwards?  We've already generated the entire answer here and thus already taken the memory hit, so why not emit it directly?  This is one of the reasons we did not use the `JoinGatherer` interface when implementing existence join for hash-based joins.  I'm not sure it makes a lot of sense to do so.  The gather map is not going to explode, and the column we're adding as part of the join is minimal in size (i.e.: a boolean column).

If this is going to stick with the `JoinGatherer` approach (which oddly does no such gathering in the existence join case), it seems it should not manifest the entire output up-front but only the chunks that are requested at the time they are requested.  Otherwise we can just skip the spilling overhead and just emit the answer directly (as we did in the hash-join existence case).

integration_tests/src/main/python/join_test.py
res-life: Emit the batch directly as the `exists` boolean column is small.
No need to test the splitting.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
res-life: Updated

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
res-life: Updated

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
res-life: Emit the batch directly.

res-life:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -510,8 +510,7 @@ class ConditionalHashJoinIterator(
   }
 }
 
-
-class ExistenceJoinIterator(
+class MixedExistenceJoinIterator(
```
jlowe, line 513: This class is not only used with mixed joins, it's also used for any unconditional hash join.  I suggest calling this something like `HashedExistenceJoinIterator`

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExec.scala
```diff
@@ -637,3 +646,25 @@ case class GpuBroadcastNestedLoopJoinExec(
     }
   }
 }
+
+class ConditionalExistenceJoinIterator(
```
jlowe, line 650: Similarly this is not the iterator to use for any conditional existence join but specifically a nested loop join.  I suggest calling this `ConditionalNestedLoopExistenceJoinIterator` or something similar.

jlowe (commented):
Minor nits on the class names used but otherwise looking good.

res-life:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
res-life: done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExec.scala
res-life: done

jlowe (dismissed):

integration_tests/src/main/python/join_test.py
```diff
@@ -841,3 +841,28 @@ def do_join(spark):
             "spark.sql.adaptive.enabled": aqeEnabled,
             "spark.sql.autoBroadcastJoinThreshold": bhjThreshold
         })
+
+
+@ignore_order
+@pytest.mark.parametrize('aqeEnabled', [pytest.param(True, id='aqe:on'), pytest.param(False, id='aqe:off')])
```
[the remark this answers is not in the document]
gerashegalov: We should have a consistent id scheme across the pytest code base regardless of the choice made

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -601,7 +572,7 @@ class ExistenceJoinIterator(
    * the value "false", scattering "true" into column FC will produce the "exists"
    * column of ExistenceJoin
    */
-  private def existsScatterMap(leftColumnarBatch: ColumnarBatch): GatherMap = {
+  def existsScatterMap(leftColumnarBatch: ColumnarBatch): GatherMap = {
```
gerashegalov, line 575: nit:

```suggestion
  override def existsScatterMap(leftColumnarBatch: ColumnarBatch): GatherMap = {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExec.scala
```diff
@@ -637,3 +647,25 @@ case class GpuBroadcastNestedLoopJoinExec(
...
+    spillableBuiltBatch: LazySpillableColumnarBatch,
+    lazyStream: Iterator[LazySpillableColumnarBatch],
+    condition: CompiledExpression,
+    opTime: GpuMetric,
+    joinTime: GpuMetric
+) extends ExistenceJoinIterator(spillableBuiltBatch, lazyStream, opTime, joinTime) {
+
+  use(condition)
+
+  def existsScatterMap(leftColumnarBatch: ColumnarBatch): GatherMap = {
```
gerashegalov, line 661: nit:
```suggestion
  override def existsScatterMap(leftColumnarBatch: ColumnarBatch): GatherMap = {
```

gerashegalov, line 661: nit:
```suggestion
  override def existsScatterMap(leftColumnarBatch: ColumnarBatch): GatherMap = {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
```diff
@@ -601,7 +572,7 @@ class ExistenceJoinIterator(
    * the value "false", scattering "true" into column FC will produce the "exists"
    * column of ExistenceJoin
    */
```
gerashegalov, line 574: the comment should be moved to the abstract parent class method declaration

gerashegalov, line 574: the comment should be moved to the abstract parent class method declaration

gerashegalov (commented):
LGTM, some nits

res-life:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
res-life: Emit the batch directly as the `exists` boolean column is relatively small.
It's the same as `GpuHashJoin` does, no need to split now.

integration_tests/src/main/python/join_test.py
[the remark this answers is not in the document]
res-life: done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
```diff
@@ -137,8 +137,13 @@ trait JoinGatherer extends LazySpillable with Arm {
 object JoinGatherer extends Arm {
   def apply(gatherMap: LazySpillableGatherMap,
       inputData: LazySpillableColumnarBatch,
-      outOfBoundsPolicy: OutOfBoundsPolicy): JoinGatherer =
-    new JoinGathererImpl(gatherMap, inputData, outOfBoundsPolicy)
+      outOfBoundsPolicy: OutOfBoundsPolicy, isExistenceJoin: Boolean = false): JoinGatherer = {
+    if (!isExistenceJoin) {
```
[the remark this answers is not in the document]
res-life: done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/JoinGatherer.scala
```diff
@@ -670,4 +675,133 @@ case class MultiJoinGather(left: JoinGatherer, right: JoinGatherer) extends Join
...
+ *
+ * Explanation of this sql is:
+ *
+ * Filter(left_table.column_0 >= 3 or `exists`)
+ *   Existence_join // generate `exists` column, do not shrink or expand the rows of left table
+ *     left_table
+ *     right_table
+ * </code>
+ */
+class JoinGathererForExistenceJoin(
```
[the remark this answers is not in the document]
res-life: No need now.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
res-life: done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuBroadcastNestedLoopJoinExec.scala
res-life: done

res-life: done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/execution/GpuHashJoin.scala
res-life: done

res-life:
build

gerashegalov (approved):

Issue #4060, opened 2021-11-09
[BUG] test_hash_groupby_approx_percentile_long_repeated_keys failed intermittently


pxLi:
**Describe the bug**
related to https://github.com/NVIDIA/spark-rapids/pull/3770

rapids_databricks_nightly-dev-github build ID 212

```
[2021-11-09T03:30:21.638Z] ../../src/main/python/hash_aggregate_test.py:1163: 
[2021-11-09T03:30:21.638Z] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[2021-11-09T03:30:21.638Z] 
[2021-11-09T03:30:21.638Z] df_fun = <function test_hash_groupby_approx_percentile_long_repeated_keys.<locals>.<lambda> at 0x7f11f7f65320>
[2021-11-09T03:30:21.638Z] percentiles = [0.05, 0.25, 0.5, 0.75, 0.95]
[2021-11-09T03:30:21.638Z] conf = {'spark.rapids.sql.expression.ApproximatePercentile': 'true', 'spark.sql.adaptive.enabled': 'false'}
[2021-11-09T03:30:21.638Z] 
[2021-11-09T03:30:21.638Z]     def compare_percentile_approx(df_fun, percentiles, conf):
[2021-11-09T03:30:21.638Z]     
[2021-11-09T03:30:21.638Z]         # create SQL statements for exact and approx percentiles
[2021-11-09T03:30:21.638Z]         p_exact_sql = create_percentile_sql("percentile", percentiles)
[2021-11-09T03:30:21.638Z]         p_approx_sql = create_percentile_sql("approx_percentile", percentiles)
[2021-11-09T03:30:21.638Z]     
[2021-11-09T03:30:21.638Z]         def run_exact(spark):
[2021-11-09T03:30:21.638Z]             df = df_fun(spark)
[2021-11-09T03:30:21.638Z]             df.createOrReplaceTempView("t")
[2021-11-09T03:30:21.638Z]             return spark.sql(p_exact_sql)
[2021-11-09T03:30:21.638Z]     
[2021-11-09T03:30:21.638Z]         def run_approx(spark):
[2021-11-09T03:30:21.638Z]             df = df_fun(spark)
[2021-11-09T03:30:21.638Z]             df.createOrReplaceTempView("t")
[...]

PR #5307, opened 2022-04-25
Remove compatibility guide reference to issue #4060

andygrove:
Issue #4060 was resolved some time ago but we did not remove the reference to it from the compatibility guide.

This PR removes that reference.


sameerz:
build

PR #5310, opened 2022-04-25
Use C++ to parse and filter parquet footers.

revans2:
This still needs tests and documentation, but I wanted to get a PR up sooner than later.

This depends on https://github.com/NVIDIA/spark-rapids-jni/pull/199


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -334,6 +340,160 @@ private case class ParquetFileInfoWithBlockMeta(filePath: Path, blocks: Seq[Bloc
     partValues: InternalRow, schema: MessageType, isCorrectedInt96RebaseMode: Boolean,
     isCorrectedRebaseMode: Boolean, hasInt96Timestamps: Boolean)
 
+class HMBSeekableInputStream(hmb: HostMemoryBuffer, hmbLength: Long) extends SeekableInputStream {
```
jlowe, line 343: Seems like this belongs in `HostMemoryStreams.scala` and could reuse code from `HostMemoryInputStream` (or maybe `HostMemoryInputStream` should just add the `SeekableInputStream` functionality).

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1639,6 +1657,13 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
 
   lazy val isParquetInt96WriteEnabled: Boolean = get(ENABLE_PARQUET_INT96_WRITE)
 
+  lazy val parquetReaderFooterType: ParquetFooterReaderType.Value = {
+    get(PARQUET_READER_FOOTER_TYPE) match {
+      case "NATIVE" => ParquetFooterReaderType.NATIVE
+      case _ => ParquetFooterReaderType.JAVA
```
jlowe, line 1663: I think this should fail, rather than assume anything not "NATIVE" means Java.  For example, if the user specified "NATIV" it will silently assume they meant "JAVA".

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -358,21 +518,168 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
...
-    "msg=constructor ParquetFileReader in class ParquetFileReader is deprecated"
-  )
+  private def addNamesAndCount(names: ArrayBuffer[String], children: ArrayBuffer[Int],
+      name: String, num_children: Int): Unit = {
+    names += name
+    children += num_children
+  }
+
+  // TODO this needs to move into spark.sql package of some kind to get the right type
+  // checks, for now it is a bit hacky
```
jlowe, line 528: Needs a followup issue?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -358,21 +518,168 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
...
+
+  def readAndFilterFooter(
+      file: PartitionedFile,
+      conf : Configuration,
+      readDataSchema: StructType,
+      filePath: Path): ParquetFooter = {
+    val (names, children) = depthFirstNames(readDataSchema, !isCaseSensitive)
+    val fs = filePath.getFileSystem(conf)
+    val stat = fs.getFileStatus(filePath)
+    // TODO should we do some checks here???
```
jlowe, line 581: Would be good to elaborate on the type of checks desired (if not outright implement them).  Some minimal sanity checking is done below (i.e.: file length).

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -358,21 +518,172 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
...
       readDataSchema: StructType): ParquetFileInfoWithBlockMeta = {
     withResource(new NvtxRange("filterBlocks", NvtxColor.PURPLE)) { _ =>
       val filePath = new Path(new URI(file.filePath))
-      //noinspection ScalaDeprecation
-      val footer = withResource(new NvtxRange("readFooter", NvtxColor.YELLOW)) { _ =>
-        ParquetFileReader.readFooter(conf, filePath,
-          ParquetMetadataConverter.range(file.start, file.start + file.length))
+      val footer = footerReader match {
+        case ParquetFooterReaderType.NATIVE =>
+          System.err.println("NATIVE FOOTER READER...")
```
gerashegalov, line 658: Switch to real logging once done with WIP

abellina: agree with @gerashegalov

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -334,6 +340,160 @@ private case class ParquetFileInfoWithBlockMeta(filePath: Path, blocks: Seq[Bloc
...
+    val amountRead = read(buffer, offset, length)
+    val remaining = length - amountRead
+    if (remaining > 0) {
+      throw new EOFException("Reached the end of stream with " + remaining + " bytes left to read")
+    }
+  }
+
+  @throws[IOException]
+  override def read(buf: ByteBuffer): Int =
+    if (buf.hasArray) {
```
abellina, line 416: maybe a small comment somewhere on why we need two implementations one for direct and one for JVM byte buffers.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -358,21 +518,172 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
     }
   }
 
-  @scala.annotation.nowarn(
-    "msg=constructor ParquetFileReader in class ParquetFileReader is deprecated"
-  )
+  private def addNamesAndCount(names: ArrayBuffer[String], children: ArrayBuffer[Int],
+      name: String, num_children: Int): Unit = {
```
abellina, line 522:
```suggestion
      name: String, numChildren: Int): Unit = {
```

revans2:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
revans2: The reason I did this was because SeekableInputStream is a Parquet specific class. I wasn't sure I should be mixing those two. If you still want me to I am happy to do it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
revans2: I did it this way because the config should be enforcing some of this, but I will be more defensive because we know that the config is not 100% perfect in how it checks for things.

revans2:
build

revans2:
@jlowe @abellina @gerashegalov could you please take another look? I think I have addressed all of the review comments.

jlowe (approved):
All my comments are minor, mostly about potentially simplifying/optimizing the copies leveraging HostMemoryBuffer's ability to alias as a DirectByteBuffer.  Nothing is must-fix on my end.

revans2:
build

revans2:
Thanks for the review @jlowe I think I have addressed everything.

abellina (approved):

jlowe (dismissed):

revans2:
Converted to draft because I hit a bug with the latest code that I need to debug.

revans2:
build

revans2:
I reverted the last set of changes to the StreamReader.  They were causing issues when reading large files. I don't know why ByteBuffers are only used for large files but they are. Because the comments were nits I decided to revert this and I will file a follow on issue to see if we can make it work properly, along with some investigation into why tests were not taking that code path.

revans2:
I filed https://github.com/NVIDIA/spark-rapids/issues/5452 as the follow on issue.

jlowe (approved):

abellina (approved):

revans2:
build

PR #5315, opened 2022-04-26
Simplify conditional catalyst expressions generated by udf-compiler

### LINKED ISSUE #3985: [QST] UDF-Compiler: Predicate pushdown is not working in this case
**What is your question?**
Hi, It is a great open-source project, thanks for sharing. We have been evaluating the udf-compiler, it works great as advertised, but we run into a problem when we try to test a filter pushdown. We wrote a simple UDF as below, but the PushedFilters is empty with udf-compiler:
```
    spark.udf.register("isOlderThan20", (i: Int) => {i>20})
    val udfResult = spark.sql("SELECT * FROM people_with_schema WHERE isOlderThan20(age)")
    
   +- FileScan json [name#0,age#1] Batched: false, DataFilters: [if ((age#1 > 20)) true else false], ..... PushedFilters: [], 
```
What we expected is to have a physical plan like this:
```
    val udfResult = spark.sql("SELECT * FROM people_with_schema WHERE age > 20")
   
    !+- Filter (isnotnull(age#1) AND (age#1 > 20))                                                                                                                                                                                                                                                     
    +- FileScan ...PushedFilters: [IsNotNull(age), GreaterThan(age,20)], 
```
By turning on trace, it looks like that the udf-compiler translated this udf as "if-else", instead of just "age#1 >20":
`Filter if ((age#1 > 20)) true else false`

Wonder if this is by design? If not, wonder if you can suggest some ideas to us on how to fix this?

Thank you very much!
-Yong Wang

seanprime7:
if (c) true else false => c
if (c) false else true => !c

Signed-off-by: Sean Lee <selee@nvidia.com>

Resolves #3985

### DIFF
--- a/udf-compiler/src/main/scala/com/nvidia/spark/udf/CatalystExpressionBuilder.scala
+++ b/udf-compiler/src/main/scala/com/nvidia/spark/udf/CatalystExpressionBuilder.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -254,7 +254,7 @@ object CatalystExpressionBuilder extends Logging {
       val res = expr match {
         case And(Literal.TrueLiteral, c) => simplifyExpr(c)
         case And(c, Literal.TrueLiteral) => simplifyExpr(c)
-        case And(Literal.FalseLiteral, _) => Literal.FalseLiteral
+        case and: And if and.containsChild(Literal.FalseLiteral) => Literal.FalseLiteral
         case And(c1@LessThan(s1, Literal(v1, t1)),
         c2@LessThan(s2, Literal(v2, t2))) if s1 == s2 && t1 == t2 => {
           t1 match {
@@ -346,7 +346,7 @@ object CatalystExpressionBuilder extends Logging {
           }
         }
         case And(c1, c2) => And(simplifyExpr(c1), simplifyExpr(c2))
-        case Or(Literal.TrueLiteral, _) => Literal.TrueLiteral
+        case or: Or if or.containsChild(Literal.TrueLiteral) => Literal.TrueLiteral
         case Or(Literal.FalseLiteral, c) => simplifyExpr(c)
         case Or(c, Literal.FalseLiteral) => simplifyExpr(c)
         case Or(c1@GreaterThan(s1, Literal(v1, t1)),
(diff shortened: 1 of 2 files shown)


udf-compiler/src/main/scala/com/nvidia/spark/udf/CatalystExpressionBuilder.scala
```diff
@@ -255,6 +255,7 @@ object CatalystExpressionBuilder extends Logging {
         case And(Literal.TrueLiteral, c) => simplifyExpr(c)
```
jlowe, line 255: 2022 copyrights

seanprime7: Updated the copyright years

jlowe (dismissed):

udf-compiler/src/main/scala/com/nvidia/spark/udf/CatalystExpressionBuilder.scala
```diff
@@ -255,6 +255,7 @@ object CatalystExpressionBuilder extends Logging {
...
         case And(c, Literal.TrueLiteral) => simplifyExpr(c)
         case And(Literal.FalseLiteral, _) => Literal.FalseLiteral
+        case And(_, Literal.FalseLiteral) => Literal.FalseLiteral
```
gerashegalov, line 258: Can be collapsed to a single line
```suggestion
        case and: And if and.containsChild(Literal.FalseLiteral) => Literal.FalseLiteral
```

abellina (dismissed):
LGTM other than @gerashegalov's suggestions.

udf-compiler/src/main/scala/com/nvidia/spark/udf/CatalystExpressionBuilder.scala
seanprime7: done

udf-compiler/src/main/scala/com/nvidia/spark/udf/CatalystExpressionBuilder.scala
```diff
@@ -347,6 +348,7 @@ object CatalystExpressionBuilder extends Logging {
         }
         case And(c1, c2) => And(simplifyExpr(c1), simplifyExpr(c2))
         case Or(Literal.TrueLiteral, _) => Literal.TrueLiteral
+        case Or(_, Literal.TrueLiteral) => Literal.TrueLiteral
```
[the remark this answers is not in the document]
seanprime7: done

gerashegalov (approved):
LGTM

jlowe:
@seanprime7 please do not force-push once the PR is being reviewed.  It often orphans the reviewer comments and makes re-reviews more difficult.  Commits are already squashed when merged, so PRs don't need to optimize their commits before merging.

seanprime7:
copy that

jlowe (approved):

jlowe:
build

Issue #5119, opened 2022-03-31
[FEA] Support ANSI mode for SQL functions/operators


nartal1:
**Is your feature request related to a problem? Please describe.**
Below is the list of SQL functions/operators which are currently supported by the plugin but don't support ANSI mode. We have to add ANSI mode support for these functions to have same behavior as Spark. We may have to check the Spark versions in which ANSI support for these functions were added.

Please note that there are other functions which support ANSI mode but not listed here as we don't support those SQL functions in RAPIDS Accelerator for Apache Spark.

## SQL functions/operators.
- [x] Pmod
- [x] Remainder
- [x] ToUnixTimestamp
- [x] UnixTimestamp
- [x] GetTimestamp
- [x] DateAddInterval

res-life:
Finished.

Issue #1556, opened 2021-01-20
[FEA] Implement ANSI mode tests for string to timestamp functions


andygrove:
**Is your feature request related to a problem? Please describe.**
We need to implement ANSI mode tests for string to timestamp functions such as `to_date`, `to_timestamp`, `unix_timestamp`, and `to_unix_timestamp` to ensure we have the correct behavior for invalid inputs.

PR #5316, opened 2022-04-26
Support ANSI mode for `ToUnixTimestamp, UnixTimestamp, GetTimestamp, DateAddInterval`

res-life:
Contributes to #5119 

Supports ANSI mode for:
- ToUnixTimestamp
- UnixTimestamp
- GetTimestamp
- DateAddInterval

### DIFF
--- a/integration_tests/src/main/python/data_gen.py
+++ b/integration_tests/src/main/python/data_gen.py
@@ -975,6 +975,7 @@ nested_gens_sample = array_gens_sample + struct_gens_sample_with_decimal128 + ma
 
 ansi_enabled_conf = {'spark.sql.ansi.enabled': 'true'}
 no_nans_conf = {'spark.rapids.sql.hasNans': 'false'}
+legacy_interval_enabled_conf = {'spark.sql.legacy.interval.enabled': 'true'}
 
 def copy_and_update(conf, *more_confs):
     local_conf = conf.copy()
@@ -1030,10 +1031,3 @@ def append_unique_int_col_to_df(spark, dataframe):
     new_rows = append_unique_to_rows(collected)
     new_schema = StructType(existing_schema.fields + [StructField("uniq_int", IntegerType(), False)])
     return spark.createDataFrame(new_rows, new_schema)
-
-# add multiple confs
-def add_conf(*confs):
-    all_items = []
-    for conf in confs:
-        all_items += list(conf.items())
-    return dict(all_items)
--- a/integration_tests/src/main/python/data_gen.py
+++ b/integration_tests/src/main/python/data_gen.py
@@ -1030,3 +1030,10 @@ def append_unique_int_col_to_df(spark, dataframe):
     new_rows = append_unique_to_rows(collected)
     new_schema = StructType(existing_schema.fields + [StructField("uniq_int", IntegerType(), False)])
     return spark.createDataFrame(new_rows, new_schema)
+
+# add multiple confs
+def add_conf(*confs):
+    all_items = []
+    for conf in confs:
+        all_items += list(conf.items())
+    return dict(all_items)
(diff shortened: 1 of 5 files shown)


res-life:
build

integration_tests/src/main/python/date_time_test.py
```diff
@@ -13,14 +13,17 @@
...
-from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_fallback_collect
+from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_fallback_collect, assert_gpu_and_cpu_error
 from data_gen import *
 from datetime import date, datetime, timezone
 from marks import incompat, allow_non_gpu
 from pyspark.sql.types import *
 from spark_session import with_spark_session, is_before_spark_330
 import pyspark.sql.functions as f
 
+ANSI_ENABLE_CONF = {'spark.sql.ansi.enabled': 'true'}
```
jlowe, line 24: This is a duplicate of `ansi_enabled_conf` later in this file.

integration_tests/src/main/python/data_gen.py
```diff
@@ -1030,3 +1030,10 @@ def append_unique_to_rows(rows):
     new_rows = append_unique_to_rows(collected)
     new_schema = StructType(existing_schema.fields + [StructField("uniq_int", IntegerType(), False)])
     return spark.createDataFrame(new_rows, new_schema)
+
+# add multiple confs
+def add_conf(*confs):
```
jlowe, line 1035: Do we really need this?  We already have `copy_and_update` whichcan be used to accomplish the same thing and is used in many other places, e.g.:
```
combined_confs = copy_and_update(first_conf, second_conf, third_conf ...)
```

integration_tests/src/main/python/date_time_test.py
```diff
@@ -13,14 +13,17 @@
...
+LEGACY_INTERVAL_CONF = {'spark.sql.legacy.interval.enabled': 'true'}
```
jlowe, line 25: Nit: This should be moved next to the other predefined confs in this file

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -499,10 +504,17 @@ object GpuToTimestamp extends Arm {
...
-      dtype: DType): ColumnVector = {
+      dtype: DType,
+      failOnError: Boolean): ColumnVector = {
 
     // `tsVector` will be closed in replaceSpecialDates
     val tsVector = withResource(isTimestamp(lhs.getBase, sparkFormat, strfFormat)) { isTs =>
+      if(failOnError && !BoolUtils.isAllValidTrue(isTs)) {
+        // ANSI mode and has invalid value.
+        // CPU may throw `DateTimeParseException`, `DateTimeException` or `ParseException`
+        throw new IllegalArgumentException("Exception occurred when parsing timestamp in ANSI mode")
```
jlowe, line 515: Do we need a followup issue that we're not throwing a matching exception type here?

res-life: I will add more tests when fixing this PR: https://github.com/NVIDIA/spark-rapids/issues/1556.
Do we need to distinguish these exceptions?  If so, let's do it in this PR https://github.com/NVIDIA/spark-rapids/issues/1556

res-life:
build

integration_tests/src/main/python/date_time_test.py
res-life: done

integration_tests/src/main/python/date_time_test.py
res-life: Done.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
jlowe: > Do we need to distinguish these exceptions? 

I don't think it is high priority to fix, and I could see adding some more tests while still leaving the exception types unaddressed.  So if #1556 does not address the exception types, we should file a low priority issue to track after that.

jlowe (approved):

nartal1 (approved):

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
res-life: Filed an issue to track: https://github.com/NVIDIA/spark-rapids/issues/5355

PR #5317, opened 2022-04-26
Support arrays_zip

### LINKED ISSUE #5229: [FEA] Support arrays_zip
I wish we can support arrays_zip.

Eg:
```
from pyspark.sql.functions import *
df = spark.createDataFrame([(["a", "b", "a"], ["b", "c"]), (["a","a"], ["b", "c"]), (["aa"], ["b", "c"])    ], ['x', 'y'])
df = spark.read.parquet("/tmp/testparquet")
df.select(arrays_zip(df.x,df.y).alias("zip")).collect()
```

```
    ! <ArraysZip> arrays_zip(x#72, y#73, x, y) cannot run on GPU because GPU does not currently support the operator class org.apache.spark.sql.catalyst.expressions.ArraysZip
```

sperlingxx:

Closes #5229

This PR is to support arrays_zip. The main idea is leveraging `cudf::segmented_gather` to align all input arrays to the same length through padding non-longest arrays with nulls.


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2905,6 +2905,23 @@ object GpuOverrides extends Logging {
...
+        TypeSig.commonCudfTypes + TypeSig.DECIMAL_128 + TypeSig.NULL +
+          TypeSig.ARRAY + TypeSig.STRUCT + TypeSig.MAP),
+        TypeSig.ARRAY.nested(TypeSig.all),
+        repeatingParamCheck = Some(RepeatingParamCheck("children",
+          TypeSig.ARRAY.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL_128 + TypeSig.NULL +
+            TypeSig.ARRAY + TypeSig.STRUCT + TypeSig.MAP),
+          TypeSig.ARRAY.nested(TypeSig.all)))),
+      (in, conf, p, r) => new ExprMeta[ArraysZip](in, conf, p, r) {
+        override def convertToGpu(): GpuExpression = {
+          GpuArraysZip(childExprs.map(_.convertToGpu()))
```
revans2, line 2921: https://issues.apache.org/jira/browse/SPARK-35876 updated the signature of Arrays.zip to avoid some issues in the optimizer. For the versions that we currently support it went into 3.2.0 and 3.1.3. It would be good to document this here, and have a follow on issue to pay off the tech debt once we no longer support 3.1.2 or below.

integration_tests/src/main/python/array_test.py
```diff
@@ -331,4 +331,31 @@ def do_it(spark):
...
+
+@pytest.mark.parametrize('data_gen', array_zips_gen, ids=idfn)
+def test_arrays_zip(data_gen):
+    gen = StructGen(
+        [('a', data_gen), ('b', data_gen), ('c', data_gen), ('d', data_gen)], nullable=False)
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: gen_df(spark, gen).selectExpr(
+            'arrays_zip(a, b, c, d)',
+            'arrays_zip(a, b, c)',
+            'arrays_zip(a, a)',
```
revans2, line 348: Can we have some tests that mix in scalar arrays too?  Not null but things like

`array()` for an empty array or `array(9, 8, 7, 6)` for a scalar array with the given values?  That would also let us test that when we don't have a named expression that we generate the proper index name of the struct field.  Everything that you have here are named expressions.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -507,6 +509,136 @@ case class GpuArrayMax(child: Expression) extends GpuUnaryExpression with Implic
...
+      return GpuScalar(new GenericArrayData(Array.empty[Any]), dataType)
+    }
+
+    // Prepare input columns
+    closeOnExcept(workBuffers) { _ =>
+      children.foreach { expr =>
+        workBuffers += GpuExpressionsUtils.columnarEvalToColumn(expr, batch).getBase
+      }
+    }
+    val inputs = workBuffers.toArray
```
revans2, line 550: nit: This is very, very minor but technically this is what safeMap was written for.
```
val inputs = children.safeMap{ expr =>
  GpuExpressionsUtils.columnarEvalToColumn(expr, batch).getBase
}.toArray
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -507,6 +509,136 @@ case class GpuArrayMax(child: Expression) extends GpuUnaryExpression with Implic
...
+    workBuffers.clear()
+
+    withResource(inputs) { inputs =>
+      // Compute array sizes of input arrays
+      closeOnExcept(workBuffers) { _ =>
```
revans2, line 555: Again `safeMap` means we don't have to reuse `workBuffers` and the code is smaller.

```
val counts = inputs.safeMap(in => in.countElements())
withResource(counts) { _ =>
   ...
}
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -507,6 +509,136 @@ case class GpuArrayMax(child: Expression) extends GpuUnaryExpression with Implic
...
+      // Pick max array size of each row.
+      // Replace with zero if there exists null among input values.
+      // [1, 3, 5, null, 4]
+      // [2, 4, 3,    8, 0]   => [4, 4, 5, 0, 7]
+      // [4, 2, 3,   10, 7]
+      val maxSize = closeOnExcept(workBuffers) { _ =>
+        val max = withResource(ColumnVector.makeList(workBuffers: _*)) { sizes =>
+          workBuffers.foreach(_.close())
+          workBuffers.clear()
+          sizes.listReduce(SegmentedReductionAggregation.max())
```
revans2, line 570: Is it faster to create a list and then do a `listReduce`? or would it be faster to just keep a running NULL_MAX of the columns as you go?  The running max will use less memory, so I am inclined to go with that over this implementation, but it would definitely be more complicated to write, so I was curious what others thought about this.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -507,6 +509,136 @@ case class GpuArrayMax(child: Expression) extends GpuUnaryExpression with Implic
...
+    children.map(_.dataType.asInstanceOf[ArrayType].elementType)
+
+  override def withNewChildrenInternal(newChildren: IndexedSeq[Expression]): Expression =
+    legacyWithNewChildren(newChildren)
+
+  override def columnarEval(batch: ColumnarBatch): Any = {
+    val numRows = batch.numRows()
+    val numCols = children.size
+    // A container collects all column vectors in use, to ease the release on exceptions.
+    val workBuffers = mutable.ArrayBuffer[ColumnVector]()
```
revans2, line 538: I am not a fan of how workBuffers is reused all over the place. I personally would prefer to see variables that are specific to what we are doing, and what it is holding.  I would also like to see this broken down into smaller functions that are clear on what each thing is doing.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -507,6 +509,136 @@ case class GpuArrayMax(child: Expression) extends GpuUnaryExpression with Implic
...
+        withResource(max) { _ =>
+          withResource(GpuScalar.from(0, IntegerType)) { zero =>
+            max.replaceNulls(zero)
+          }
+        }
+      }
+
+      // Generate offsets from max sizes
+      // [4, 3, 5, 2] => [0, 4, 7, 12, 14]
+      workBuffers += closeOnExcept(workBuffers) { _ =>
```
revans2, line 581: This is so confusing to me. Why close workBuffers if there is an exception if we cleared it out on lines 568 and 569? But we are also adding something new into workBuffers, at the end, but it is just a single value? What is happening here?

integration_tests/src/main/python/array_test.py
```diff
@@ -331,4 +331,31 @@ def do_it(spark):
...
+            'arrays_zip(a)')
+    )
+
+
+def test_arrays_zip_corner_cases():
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: unary_op_df(spark, ArrayGen(int_gen), length=100).selectExpr(
+            'arrays_zip()',
+            'arrays_zip(null)',
+            'arrays_zip(null, null)',
```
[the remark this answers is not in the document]
sperlingxx:
```
GpuColumnarToRow false
+- GpuProject [gpuarrayszip() AS arrays_zip()#18, gpuarrayszip(null) AS arrays_zip(NULL)#19, gpuarrayszip(null, null) AS arrays_zip(NULL, NULL)#20, gpuarrayszip(null, a#16) AS arrays_zip(NULL, a)#21, gpuarrayszip(a#16, []) AS arrays_zip(a, array())#22, gpuarrayszip(a#16, [], [1,2]) AS arrays_zip(a, array(), array(1, 2))#23, gpuarrayszip(a#16, [1,2,4,3], [5]) AS arrays_zip(a, array(1, 2, 4, 3), array(5))#24]
   +- GpuRowToColumnar targetsize(104857600)
      +- *(1) Scan ExistingRDD[a#16]
```
According to the dumped plan,  it indeedly runs GpuArraysZop over these liternal values If I am not wrong.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
sperlingxx: Yes, I didn't realize I can use `safeMap` to simplify all these buffer tricks. I reorgnized all the codes: replacing workBuffers with `safeMap` and breaking the whole snippet into several functiions.

sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
sperlingxx: Filed an issue #5327, and dropped the link here as a notice

integration_tests/src/main/python/array_test.py
sperlingxx: Added cases for scalar arrays.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
sperlingxx: I am not sure which approach is faster. I chose the `listReduce` over max iterations because it invokes less times of native kernel calls. Perhaps we need a mini benchmark to provide some empirical evidence ?

revans2: I think this is fine as is. This is working, so there is less of a reason to change it.

revans2 (approved):
Generally it looks good.

PR #5319, opened 2022-04-26
Re-enable support for `\Z` in regular expressions on the GPU

### LINKED ISSUE #4532: [FEA] Re-enable support for `\Z` in regular expressions
PR https://github.com/NVIDIA/spark-rapids/pull/4530 will remove support for `\Z` due to correctness issues. We need to re-implement it.

See above.

None

**Additional context**
None

NVnavkumar:
Fixes #4532.

This re-enables support for the `\Z` string anchor in regular expressions on the GPU. In the default mode of regular expressions in Spark (Java), `\Z` is actually a synonym for `$`. The behavior differs only from `$` when the `MULTILINE` or `UNIX_LINES` flags are enabled in Java (in that `\Z` maintains the same behavior while `$` changes to matching ends of lines and not the end of the string). 

This branch also includes a fix for an edge case that came up during fuzz testing for `\Z` that involves matching groups at the end of the string. In this case, it can now capture the line terminator at the end of the string. 

Also, a documentation fix is included that was not updated earlier when `$` was re-enabled.


NVnavkumar:
build

NVnavkumar:
build

andygrove (approved):

PR #5349, opened 2022-04-28
Enable some repetitions for `\A` and `\Z`

### LINKED ISSUE #4800: [FEA] Enable support for more regular expressions with \A and \Z
Currently, there are certain escape sequences in regular expressions that are unsupported in the plugin because the behavior is either inconsistent with Spark or crashes the underlying `libcudf`.

`\A*`, `\A+` and `\Z` can be transpiled to equivalents: 

* `\A` in repetitions can be transpiled to single `\A` or `\A?` depending on whether there is a `*` or `+` or `{0,}` or `{1,}`, and etc.
* `\Z` can be transpiled to `[\n\r]*$`, and in repetition we could at least support `+` by essentially dropping it

NVnavkumar:
Fixes #4800 (which was also partially fixed by PR#5319 when `\Z` was finally re-enabled). 

This enables using `\A` and `\Z` in some repetition sequences to add more full support for those escape sequences in regular expressions on the GPU.  This enables:
* `+` near `\A` or `\Z`
* `{n}` or `{n,}` or `{n,m}` where `n > 0`

NOTE:
* `\A*` and `\A{...}` can be transpiled to `\A?`; however cuDF does not yet support `\A?`, so this `*` and `?` and  `{0,}` will all still fallback to CPU.


NVnavkumar:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -858,15 +858,67 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
 
-        case (RegexGroup(_, term), SimpleQuantifier(ch))
+        case (RegexGroup(capture, term), SimpleQuantifier(ch))
             if "+*".contains(ch) && !isSupportedRepetitionBase(term) =>
-          throw new RegexUnsupportedException(nothingToRepeat)
-        case (RegexGroup(_, term), QuantifierVariableLength(_, None))
+          (term, ch) match {
+            // \Z is not supported in groups
+            case (RegexEscaped('A'), '+') |
+                (RegexSequence(ListBuffer(RegexEscaped('A'))), '+') =>
```
andygrove, line 866: This pattern is repeated a few times. I wonder if it is worth introducing a utility function that can simply expressions to remove redundant list buffers?

andygrove (approved):
LGTM. I left one suggestion but not critical for this PR.

PR #5350, opened 2022-04-28
Halt Spark executor when encountering unrecoverable CUDA errors

### LINKED ISSUE #5029: [FEA] Stop running task attempts on executors that encounter "sticky" CUDA errors
Certain CUDA errors, like illegal memory access, are "sticky," meaning that all CUDA operations to the GPU after the error will continue to return the same error over and over.  No GPU operations will succeed after that point.

The RAPIDS Accelerator should take measures to prevent further task execution on the executor once these "sticky" exceptions are detected.  Tearing down the executor process is probably the best option, at least in the short-term.  Without an external shuffle handler we will lose the shuffle of tasks that have completed, but this is probably a better way to "fail fast" then allow the executor to keep accepting new tasks only to have them fail the first time they touch the GPU.

sperlingxx:

Closes #5029 

Following #5118, this PR detects unrecoverable (fatal) CUDA errors through the cuDF utility, which applys a more comprehensive way to determine whether a CUDA error is fatal or not.


sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -328,6 +324,23 @@ object RapidsExecutorPlugin {
       zipped.forall { case (e, a) => e <= a }
     }
   }
+
+  private val CUDA_EXCEPTION: String = classOf[ai.rapids.cudf.CudaException].getName
+  private val CUDA_FATAL_EXCEPTION: String = classOf[ai.rapids.cudf.CudaFatalException].getName
+  private val CUDF_EXCEPTION: String = classOf[ai.rapids.cudf.CudfException].getName
+
+  def isCudaException(ef: ExceptionFailure): Boolean = {
+    ef.description.contains(CUDA_EXCEPTION) || ef.toErrorString.contains(CUDA_EXCEPTION)
```
tgravescs, line 333: are the cudf exceptions not serializable?  Otherwise it looks like we should be able to use exception() function call to look at the actual exception type.   I guess we can keep the string parsing if the exception is None.  @jlowe for opinion

jlowe: Agree we should pursue a solution that does not rely on scraping description messages if possible.

sperlingxx: Done.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -328,6 +324,34 @@ object RapidsExecutorPlugin {
       zipped.forall { case (e, a) => e <= a }
     }
   }
+
+  private val CUDA_EXCEPTION: String = classOf[ai.rapids.cudf.CudaException].getName
```
jlowe, line 328: Curious, why not import the classes rather than fully qualify them everywhere?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -328,6 +324,34 @@ object RapidsExecutorPlugin {
...
+  private val CUDA_FATAL_EXCEPTION: String = classOf[ai.rapids.cudf.CudaFatalException].getName
+  private val CUDF_EXCEPTION: String = classOf[ai.rapids.cudf.CudfException].getName
+
+  def isCudaException(ef: ExceptionFailure): Boolean = ef.exception match {
+    case Some(ex) =>
+      ex.isInstanceOf[ai.rapids.cudf.CudaException]
+    case None =>
+      ef.description.contains(CUDA_EXCEPTION) ||
+        ef.toErrorString.contains(CUDA_EXCEPTION)
```
jlowe, line 337: Is it necessary to scrape the description message now that we're catching it by the class type?  Has the code been tested to verify we are not coming through the message-scraping code path?

sperlingxx: Hi @jlowe, I encountered a problem: I couldn't find out an easy way to trigger CUDA errors via spark APIs.

sperlingxx: Hi @jlowe, I tried the above approach, it did creating a fatal CUDA error. And I found that cuDF JNI couldn't indentify (fatal) CUDA errors due to RMM calls. Because CUDA API calls in rmm are wrapped with `RMM_CUDA_TRY_ALLOC` instead of `CUDF_CUDA_TRY`:
```c++
#define RMM_CUDA_TRY_ALLOC(_call)                                                                  \
  do {                                                                                             \
    cudaError_t const error = (_call);                                                             \
    if (cudaSuccess != error) {                                                                    \
      cudaGetLastError();                                                                          \
      auto const msg = std::string{"CUDA error at: "} + __FILE__ + ":" + RMM_STRINGIFY(__LINE__) + \
                       ": " + cudaGetErrorName(error) + " " + cudaGetErrorString(error);           \
      if (cudaErrorMemoryAllocation == error) {                                                    \
        throw rmm::out_of_memory{msg};                                                             \
      } else {                                                                                     \
        throw rmm::bad_alloc{msg};                                                                 \
      }                                                                                            \
    }                                                                                              \
  } while (0)
```

jlowe: > Shall we catch allrmm::bad_alloc as CudaFatalException in cuDF JNI?

rmm::bad_alloc simply means "something else went wrong other than an allocation failure."  It does not necessarily mean a CUDA fatal exception occurred.

However this demonstrates that the CUDA errors are asynchronous, and we're not guaranteed that libcudf is the one processing the error directly (and thus translating it into a CUDA fatal exception).  That means we really have no choice but to fallback to scraping the messages, looking for something that looks like a fatal CUDA error (i.e.: something like Tom's original change).  We can still leverage the cudf exception types, but I don't see a way to avoid scraping the messages in the worst case. 😞

sperlingxx: Yes, it seems no other way but scraping the error messages, in order to make sure that we catch as many fatal errors as possible. I added back Tom's original approach as a double check over CUDA errors thrown by cuDF/rmm, since libcudf may not be able to catch all fatal errors.

sperlingxx:
build

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
sperlingxx: Hi @jlowe, I filed a cuDF PR ([10884](https://github.com/rapidsai/cudf/pull/10884)) to do it.

tgravescs:
Please retarget for 22.08

sperlingxx:
Hi @tgravescs, I retargeted to branch-22.08 and I removed the message scraping codes for double-check, since cuDF can fully take care of the fatal error detection with https://github.com/rapidsai/cudf/pull/10884.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -287,20 +288,16 @@ class RapidsExecutorPlugin extends ExecutorPlugin with Logging {
...
-          System.exit(20)
-        }
+      case ef: ExceptionFailure if RapidsExecutorPlugin.isCudaFatalException(ef) =>
+        logError("Stopping the Executor based on exception being a fatal CUDA error: " +
+          s"${ef.toErrorString}")
+        System.exit(20)
+      case ef: ExceptionFailure if RapidsExecutorPlugin.isCudaException(ef) =>
+        logDebug(s"Executor onTaskFailed because of a non-fatal CUDA error: ${ef.toErrorString}")
+      case ef: ExceptionFailure if RapidsExecutorPlugin.isCudfException(ef) =>
+        logDebug(s"Executor onTaskFailed because of a CUDF error: ${ef.toErrorString}")
```
jlowe, line 298: Nit: I think the separate predicate functions are a bit overkill, and this can be simplified a bit by removing the redundant ExceptionFailure checking for each case, e.g.:
```suggestion
      case ef: ExceptionFailure =>
        ef.exception match {
          case _: CudaFatalException =>
            logError("Stopping the Executor based on exception being a fatal CUDA error: " +
              s"${ef.toErrorString}")
            System.exit(20)
          case _: CudaException =>
            logDebug(s"Executor onTaskFailed because of a non-fatal CUDA error: ${ef.toErrorString}")
          case _: CudfException =>
            logDebug(s"Executor onTaskFailed because of a CUDF error: ${ef.toErrorString}")
          case _ =>
            logDebug(s"Executor onTaskFailed: ${ef.toErrorString}")
        }
```

jlowe (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -288,19 +289,21 @@ class RapidsExecutorPlugin extends ExecutorPlugin with Logging {
...
+        ef.exception match {
+          case _: CudaFatalException =>
+            logError("Stopping the Executor based on exception being a fatal CUDA error: " +
+              s"${ef.toErrorString}")
+            System.exit(20)
+          case _: CudaException =>
+            logDebug(s"Executor onTaskFailed because of a non-fatal CUDA error: " +
+              s"${ef.toErrorString}")
+          case _: CudfException =>
+            logDebug(s"Executor onTaskFailed because of a CUDF error: ${ef.toErrorString}")
```
jlowe, line 301: My bad, ef.exception is an option so this needs to unpack it, e.g:
```suggestion
          case Some(_: CudaFatalException) =>
            logError("Stopping the Executor based on exception being a fatal CUDA error: " +
              s"${ef.toErrorString}")
            System.exit(20)
          case Some(_: CudaException) =>
            logDebug(s"Executor onTaskFailed because of a non-fatal CUDA error: " +
              s"${ef.toErrorString}")
          case Some(_: CudfException) =>
            logDebug(s"Executor onTaskFailed because of a CUDF error: ${ef.toErrorString}")
```

jlowe (approved):

Issue #5113, opened 2022-03-31
[FEA] ANSI mode: Support CAST between types


nartal1:
**Is your feature request related to a problem? Please describe.**
Apache Spark is adding cast support to make it ANSI compliant. This issue is to track the remaining cast's that we need to Support in RAPIDS Accelerator that are already in Spark code. Below list is not complete. We need to audit casting and make sure we cover all edge cases as well between types.

- [x] https://github.com/NVIDIA/spark-rapids/issues/1556
- [x] https://github.com/NVIDIA/spark-rapids/issues/1557
- [x] https://github.com/NVIDIA/spark-rapids/issues/5111
- [x] https://github.com/NVIDIA/spark-rapids/issues/5112

PR #5352, opened 2022-04-28
Fix spark330 build due to mapKeyNotExistError changed

res-life:
Closes #5351 

Signed-off-by: Chong Gao <res_life@163.com>

PR #5353, opened 2022-04-28
Supports casting between ANSI interval types and integral types

res-life:
Contributes #5113


res-life:
build

res-life:
Depends on this: https://github.com/NVIDIA/spark-rapids/pull/5352.
Cherry-picked the #5352 PR

res-life:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -561,6 +561,39 @@ object GpuCast extends Arm {
         GpuIntervalUtils.castStringToDayTimeIntervalWithThrow(
           input.asInstanceOf[ColumnVector], dayTime)
 
+      // cast(`day time interval` as integral)
+      case (dt: DataType, _: LongType) if GpuTypeShims.isSupportedDayTimeType(dt) =>
+        GpuIntervalUtils.dayTimeIntervalToLong(input.asInstanceOf[ColumnVector], dt)
```
revans2, line 566: But input is not guaranteed to be a `ColumnVector` for nested types. Like casting an array of DayTimeInterval to an Array of longs. These need to be ColumnView, but you should be able to treat it exactly the same as a ColumnVector.

sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/GpuTypeShims.scala
```diff
@@ -202,7 +202,16 @@ object GpuTypeShims {
...
+  def typesDayTimeCanCastTo: TypeSig = TypeSig.DAYTIME + TypeSig.STRING + TypeSig.integral
+
+  def typesYearMonthCanCastTo: TypeSig = TypeSig.integral
+
+  def typesDayTimeCanCastToOnSpark: TypeSig = TypeSig.DAYTIME + TypeSig.STRING + TypeSig.integral
+
+  def typesYearMonthCanCastToOnSpark: TypeSig = TypeSig.YEARMONTH + TypeSig.STRING +
+      TypeSig.integral
+
+  def additionalTypesIntegralCanCastTo: TypeSig = TypeSig.YEARMONTH + TypeSig.DAYTIME
```
revans2, line 215: We need something in the shim layer for hasSideEffects on GpuCast so that we do the right thing

res-life: Done

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
res-life: Done

res-life:
No need to update `AnsiCast` I think, please double check. @revans2 
Spark does not allow insertint `interval` column into `integral` column.

```
spark.sql("create table t2(c1 interval year, c2 long) using parquet")
spark.sql("insert into t2 values (interval 1 year, 9223372036854775807)")
spark.sql("insert into t2 select c2, c1 from t2")
pyspark.sql.utils.AnalysisException: Cannot write incompatible data to table '`default`.`t2`':
- Cannot safely cast 'c1': bigint to interval year
- Cannot safely cast 'c2': interval year to bigint
```

res-life:
@revans2  Help review, thanks.

revans2 (approved):

Issue #3851, opened 2021-10-18
[BUG] ShimLoader.updateSparkClassLoader fails with openjdk Java11


gerashegalov:
**Describe the bug**
Originally reported by @abellina 

```
21/10/18 23:18:41 INFO ShimLoader: Updating spark classloader jdk.internal.loader.ClassLoaders$AppClassLoader@5ffd2b27 with the URLs:
 jar:file:/home/gshegalov/gits/NVIDIA/spark-rapids/dist/target/rapids-4-spark_2.12-21.12.0-SNAPSHOT.jar!/spark3xx-common/, jar:file:/
home/gshegalov/gits/NVIDIA/spark-rapids/dist/target/rapids-4-spark_2.12-21.12.0-SNAPSHOT.jar!/spark320/
21/10/18 23:18:41 ERROR CoarseGrainedExecutorBackend: Executor self-exiting due to : Unable to create executor due to No such method:
 addURL() on object: jdk.internal.loader.ClassLoaders$AppClassLoader
java.lang.NoSuchMethodException: No such method: addURL() on object: jdk.internal.loader.ClassLoaders$AppClassLoader
        at org.apache.commons.lang3.reflect.MethodUtils.invokeMethod(MethodUtils.java:231)
        at org.apache.commons.lang3.reflect.MethodUtils.invokeMethod(MethodUtils.java:184)
        at com.nvidia.spark.rapids.ShimLoader$.$anonfun$updateSparkClassLoader$3(ShimLoader.scala:216)
```

**Steps/Code to reproduce bug**

Set `JAVA_HOME` environment  to an openjdk java11 location and run code with the default `spark.rapids.force.caller.classloader=true`. 

Note this issue does not reproduce with the Microsoft Build of OpenJDK 11

**Expected behavior**
It should work.

- short term just add another [method name](https://github.com/openjdk/jdk/blob/6765f902505fbdd02f25b599f942437cd805cad1/src/java.base/share/classes/jdk/internal/loader/ClassLoaders.java#L203) to try in the reflection calls, although JDK module encapsulation [will not make it easy ](https://stackoverflow.com/questions/41265266/how-to-solve-inaccessibleobjectexception-unable-to-make-member-accessible-m) 
- long term potentially #3803

**Environment details (please complete the following information)**
local dev environment 

**Additional context**
N/A

fcomte:
[...]

PR #5356, opened 2022-04-28
[WIP] ShimLoader.updateSparkClassLoader fails with openjdk Java11

ekrivokonmapr:
This PR is opened to fix one of the JDK11 support issue: https://github.com/NVIDIA/spark-rapids/issues/3851

java.lang.NoSuchMethodException: No such method: addURL() on object: jdk.internal.loader.ClassLoaders$AppClassLoader
        at org.apache.commons.lang3.reflect.MethodUtils.invokeMethod(MethodUtils.java:231)
        at org.apache.commons.lang3.reflect.MethodUtils.invokeMethod(MethodUtils.java:184)
        at com.nvidia.spark.rapids.ShimLoader$.$anonfun$updateSparkClassLoader$3(ShimLoader.scala:216)
        
Also, deprecated new Integer(_) was replaced.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
@@ -29,7 +29,7 @@ import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution.{ColumnarRule, SparkPlan}
-import org.apache.spark.util.MutableURLClassLoader
+import org.apache.spark.util.{MutableURLClassLoader, ParentClassLoader}
 
 /*
     Plugin jar uses non-standard class file layout. It consists of three types of areas,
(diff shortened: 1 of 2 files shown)


jlowe:
Thanks for the contribution, @ekrivokonmapr!  Couple of high-level things that need to be addressed:
- The PR needs to target the default branch, `branch-22.06`, as that is where we are actively developing for the next release,  22.06.
- The PR needs a signoff on at least one commit per the [contributing guide](https://github.com/NVIDIA/spark-rapids/blob/branch-22.06/CONTRIBUTING.md#sign-your-work)

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
```diff
@@ -209,24 +209,18 @@ object ShimLoader extends Logging {
   private def updateSparkClassLoader(): Unit = {
     // TODO propose a proper addClassPathURL API to Spark similar to addJar but
     //  accepting non-file-based URI
-    serializerClassloader().foreach { urlAddable =>
```
gerashegalov, line 212: This will break existing deployments for which this logic was critical. 

Yes it's not supported on vanilla JDK11 JVM. We can look at this deeper, but In the past experience `spark.rapids.force.caller.classloader=true` was really necessary only for advanced REPLs such as some multi-session notebook servers. We proposed it as the default value because it was safe on JDK8 and covered all the cases we have encountered. 

I would suggest to check if setting `spark.rapids.force.caller.classloader=false` works for your use cases https://github.com/NVIDIA/spark-rapids/issues/3851#issuecomment-992235387. 

If yes we should set it in code for JDK versions 11+.

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleTestHelper.scala
```diff
@@ -165,7 +165,7 @@ object RapidsShuffleTestHelper extends MockitoSugar with Arm {
   }
 
   def withMockContiguousTable[T](numRows: Long)(body: ContiguousTable => T): T = {
-    val rows: Seq[Integer] = (0 until numRows.toInt).map(new Integer(_))
+    val rows: Seq[Integer] = (0 until numRows.toInt).map(Integer.valueOf)
```
gerashegalov, line 168: This should be a separate PR because it's orthogonal #3851. It's a good change but it is already fixed by #4338

gerashegalov (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
ekrivokonmapr: Thank you, will check it using workaround.

tests/src/test/scala/com/nvidia/spark/rapids/shuffle/RapidsShuffleTestHelper.scala
ekrivokonmapr: Thanks for the review.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/ShimLoader.scala
ekrivokonmapr: Workaround works for me, will close this PR.

PR #5407, opened 2022-04-29
[QualUI] Support UI in Qualification Report

### LINKED ISSUE #5176: [FEA] Qualification tool UI
We want a UI for the qualification tool to make it easier to use.

We need to discuss and come up with mockup what we want that to look like exactly.

For first iteration we could just generate the html/js static content and they could load in browser.
Alternatively if we want to get more fancy we would need a server, but that also means user has to start a server at least until it can be nicely integrated into platform offerings.

amahussein:
fixes #5176 

The code changes generate html/js static content which the user can load into their browser.

**How it works:**

- static HTML/JS files are added to the repo in `tools/src/main/resources/ui`
- After analyzing all the application in the eventLogs, the `Qualification.qualifyApps` qould launch the UI report generator.
- First, the static files are going to be copied to the report output folder
- Second, the case classes will be serialized to JSON and written as JS variables.
- When the user loads the HTML/index.html file, the JS will be read and rendered in the viewer.

**How to use the feature**

- There is no requirement to enable the feature since it is enabled by default.
- Run the qualificationTool in the CLI, then load `output-folder/ui-report/html/index.html` in your folder

**Tools used:**

- Bootstrap4
- Datatables plugin
- JQuery and Mustache.

**Changes done to the scala code to support this feature:**

- added new flag UIEnabled set to true by default to enable/disable UI reporting.
- Added new fields to the `QualificationSummaryInfo`
    ```
    longestSqlDuration: Long,
    nonSqlTaskDurationAndOverhead: Long,
    estimatedDuration: Double,
    unsupportedDuration: Long,
    speedupDuration: Long,
    speedupFactor: Double,
    totalSpeedup: Double,
    speedupBucket: String)
   ```

- Modified QualOutputWriter to the CSV files
- Added methods in QualificationAppInfo to calculate more information for each app.
- Modified unit tests.

**Things Left on purpose:**

- Unit test for the added fields in QualificationSummaryInfo. Those fields are mocks for now and it would be useless to write unit tests for unstable code. (some of those fields are randomly generated)
- Updating the docs: The feature is not stable. Field names, and description is likely going to change over the next couple of weeks.

**Known issues to be improved later**

1. separate provider of the information so that it will be feasible to test the UI-report without the need to run the entire tool.
2. Create unit tests for stable parts of the feature. Currently, it won't be feasible to do that as the requirements change frequently.
3. refactor JS/HTML code
4. Maybe move the static code out of the repo and add it as a build dependency. UI code has more tendency to change and it will cause lots of overhead to do PRs on the main repo.


tools/src/main/scala/org/apache/spark/sql/rapids/tool/ui/QualificationReportGenerator.scala
```diff
@@ -0,0 +1,119 @@
...
+  def createQualReportGenerator(
+      provider: Qualification,
+      propFilePath: String = null): Unit = {
+    loadDefaultSparkProperties(conf, propFilePath)
+    val generator = new QualificationReportGenerator(conf, provider)
+    generator.launch()
+  }
+
+  def loadDefaultSparkProperties(conf: SparkConf, filePath: String = null): String = {
+    val propPath = Option(filePath).getOrElse(getDefaultUIProperties())
```
nartal1, line 106: I tried running this locally on my machine and got this Exception while reading the default properties file. Looks like the path(which is in String) is incorrect. Please let me know if I am missing something.

Run command: `java -cp rapids-4-spark-tools_2.12-22.06.0-SNAPSHOT.jar:$SPARK_HOME/jars/*  com.nvidia.spark.rapids.tool.qualification.QualificationMain /home/nartal/spark-rapids/spark-rapids/tools/src/test/resources/spark-events-qualification/*`

Exception:
```
INFO ToolTextFileWriter: Summary report output location: rapids_4_spark_qualification_output/rapids_4_spark_qualification_output.log
Exception in thread "main" java.lang.IllegalArgumentException: requirement failed: Properties file file:/home/nartal/spark-rapids/spark-rapids/tools/target/spark311/rapids-4-spark-tools_2.12-22.06.0-SNAPSHOT.jar!/ui/conf/qual-ui-report-defaults.conf does not exist
	at scala.Predef$.require(Predef.scala:281)
	at org.apache.spark.util.Utils$.getPropertiesFromFile(Utils.scala:2145)
	at org.apache.spark.sql.rapids.tool.ui.QualificationReportGenerator$.loadDefaultSparkProperties(QualificationReportGenerator.scala:110)
	at org.apache.spark.sql.rapids.tool.ui.QualificationReportGenerator$.createQualReportGenerator(QualificationReportGenerator.scala:103)
	at com.nvidia.spark.rapids.tool.qualification.Qualification.launchUIReportGenerator(Qualification.scala:155)
	at com.nvidia.spark.rapids.tool.qualification.Qualification.qualifyApps(Qualification.scala:89)
	at com.nvidia.spark.rapids.tool.qualification.QualificationMain$.mainInternal(QualificationMain.scala:92)
	at com.nvidia.spark.rapids.tool.qualification.QualificationMain$.main(QualificationMain.scala:33)
	at com.nvidia.spark.rapids.tool.qualification.QualificationMain.main(QualificationMain.scala)
```

amahussein: My bad! I was running the command line adding the classes to the cp.
I will work on fixing the resource loading.
Meanwhile, the `cp` needs to include the classes folder.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/ui/QualificationReportGenerator.scala
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.spark.internal.Logging
+import org.apache.spark.util.Utils.getPropertiesFromFile
+
+class QualificationReportGenerator(
+    conf: SparkConf,
+    provider: Qualification) {
+
+  import QualificationReportGenerator._
+
+  // TODO: use HDFS API to create the logoutput if necessary
```
tgravescs, line 37: please file na issue with all the followups we know of.

If this doesn't work with hdfs, what does it do?  this might cause our nightly tests to fail, in which case need to turn off UI by default for now.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/ui/QualificationReportGenerator.scala
```diff
@@ -0,0 +1,119 @@
...
+  val assetsFolder = new File(getAssetsPath)
+  val outputWorkDir = new File(provider.getReportOutputPath, "ui-report")
+  val destinationFolder = {
+    val timestamp = new SimpleDateFormat("yyyyMMddHHmm").format(new Date())
+    new File(s"${outputWorkDir}-${timestamp}")
```
tgravescs, line 42: why are we doing a timestamp in output path?

tools/src/main/scala/org/apache/spark/sql/rapids/tool/ui/QualificationReportGenerator.scala
```diff
@@ -0,0 +1,119 @@
...
+         |\t${appInfoRecs};
+         |let qualificationRecords =
+         |\t${infoSummary};
+         |let dataSourceInfoRecords =
+         |\t${dataSourceInfo};
+       """.stripMargin
+    val outputFile = new File(destinationFolder, s"js/mock-data.js")
+    val pw = new PrintWriter(outputFile)
+    pw.write(qualInfoSummaryContent)
+    pw.close
```
tgravescs, line 84: try finally to make sure this gets closed

tools/src/test/scala/com/nvidia/spark/rapids/tool/qualification/QualificationSuite.scala
```diff
@@ -67,6 +67,7 @@ class QualificationSuite extends FunSuite with BeforeAndAfterEach with Logging {
     .add("Unsupported Write Data Format", StringType, true)
     .add("Complex Types", StringType, true)
     .add("Unsupported Nested Complex Types", StringType, true)
+    // TODO - Add new fields (created for the UI Milestone) to the schema
```
tgravescs, line 70: we should update this with this pr

tools/src/main/resources/ui/assets/js/spur.js
```diff
@@ -0,0 +1,39 @@
+/*!
+ * spur-template - An admin template based on Bootstrap 4
+ * Version v1.1.0
```
tgravescs, line 3: did we modify this one?

amahussein: No, Haven't changed that yet.

tools/src/main/scala/org/apache/spark/sql/rapids/tool/ui/QualificationReportGenerator.scala
nartal1: This comment is not valid anymore since `qual-ui-report-defaults.conf` is removed in the latest commit.

amahussein:
replaced by #5407

PR #5418, opened 2022-05-03
Add GpuEqualToNoNans and update GpuPivotFirst to use to handle PivotFirst with NaN support enabled on GPU

### LINKED ISSUE #5322: [FEA] remove hasNans for Pivot

Spark is inconsistent in that for pivot NaN != NaN, but for our code it is. This is because we are using a GpuEqualTo internally to check if the value matches the pivot column, and GpuEqualTo has fixes in it to make NaN == NaN. We can fix this with an alternative GpuEqualTo that removes the NaN == NaN fixes.

NVnavkumar:
Fixes #5322.

This adds the GpuEqualToNoNans operation to support EqualTo where NaN != NaN. This is then used in GpuPivotFirst to allow GPU and CPU behavior to be consistent when the plugin knows that floats can be NaN. The check for the hasNans configuration is now removed so the operation will continue to run on the GPU even if the input data has NaN values.  The integration tests have also been updated to account for `hasNans` and support NaN float inputs in `pivot(...)` on the GPU.

### DIFF
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -309,6 +309,11 @@ _init_list_no_nans_with_decimalbig = _init_list_no_nans + [
     _grpkey_short_big_decimals, _grpkey_short_very_big_decimals, 
     _grpkey_short_very_big_neg_scale_decimals]
 
+_init_list_with_nans_and_no_nans_with_decimalbig = _init_list_with_nans_and_no_nans + [
+    _grpkey_small_decimals, _grpkey_big_decimals, _grpkey_short_mid_decimals,
+    _grpkey_short_big_decimals, _grpkey_short_very_big_decimals, 
+    _grpkey_short_very_big_neg_scale_decimals]
+
 
 _init_list_full_decimal = [_grpkey_short_full_decimals, 
     _grpkey_short_full_neg_scale_decimals]
@@ -445,8 +450,8 @@ def test_exceptAll(data_gen):
 @approximate_float
 @ignore_order(local=True)
 @incompat
-@pytest.mark.parametrize('data_gen', _init_list_no_nans_with_decimalbig, ids=idfn)
-@pytest.mark.parametrize('conf', get_params(_confs, params_markers_for_confs), ids=idfn)
+@pytest.mark.parametrize('data_gen', _init_list_with_nans_and_no_nans_with_decimalbig, ids=idfn)
+@pytest.mark.parametrize('conf', get_params(_confs_with_nans, params_markers_for_confs_nans), ids=idfn)
 def test_hash_grpby_pivot(data_gen, conf):
     assert_gpu_and_cpu_are_equal_collect(
         lambda spark: gen_df(spark, data_gen, length=100)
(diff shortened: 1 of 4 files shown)


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -316,6 +316,79 @@ case class GpuEqualNullSafe(left: Expression, right: Expression) extends CudfBin
...
 }
 
+case class GpuEqualToNoNans(left: Expression, right: Expression) extends CudfBinaryComparison
+    with NullIntolerant {
+  override def symbol: String = "="
+  override def outputTypeOverride: DType = DType.BOOL8
+  override def binaryOp: BinaryOp = BinaryOp.EQUAL
+
+  override def doColumnar(lhs: GpuColumnVector, rhs: GpuColumnVector): ColumnVector = {
+    val result = super.doColumnar(lhs, rhs)
```
revans2, line 326: Does CUDF compare NaNs as equal? I thought it didn't because C++ does not, because it follow the ieee standard.

>    The comparisons EQ, GT, GE, LT, and LE, when either or both operands is NaN returns FALSE.
>
>    The comparison NE, when either or both operands is NaN returns TRUE.

I htink we can deleted all of the overrides for doColumnar and just rely on the default implementations.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -316,6 +316,79 @@ case class GpuEqualNullSafe(left: Expression, right: Expression) extends CudfBin
...
+            }
+          }
+        }
+      }
+    } else {
+      result
+    }
+  }
+
+  override def convertToAst(numFirstTableColumns: Int): ast.AstExpression = {
```
revans2, line 382: I don't think we have a way to reach this code currently. We don't support AST in aggregation projections like that, so I think it would be cleaner to just delete this method and let the default implementation throw an exception if it is called.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
```diff
@@ -316,6 +316,79 @@ case class GpuEqualNullSafe(left: Expression, right: Expression) extends CudfBin
   }
 }
 
+case class GpuEqualToNoNans(left: Expression, right: Expression) extends CudfBinaryComparison
```
revans2, line 319: nit: Can we add a few scaladocs to this to explain why we need this?

revans2 (commented):
Looks good, I just would like to see if we can make it smaller.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/predicates.scala
NVnavkumar: removed doColumnar overrides in last commit

NVnavkumar:
build

integration_tests/src/main/python/hash_aggregate_test.py
```diff
@@ -510,41 +488,47 @@ def test_hash_multiple_grpby_pivot(data_gen, conf):
...
         conf=conf)
 
+@approximate_float
+@ignore_order(local=True)
+@allow_non_gpu('HashAggregateExec', 'PivotFirst', 'AggregateExpression', 'Alias', 'GetArrayItem',
+        'Literal', 'ShuffleExchangeExec', 'HashPartitioning', 'KnownFloatingPointNormalized',
+        'NormalizeNaNAndZero')
+@incompat
+@pytest.mark.parametrize('data_gen', [_grpkey_floats_with_nulls_and_nans], ids=idfn)
+def test_hash_pivot_groupby_duplicates_fallback(data_gen):
```
nartal1, line 522: Thanks for adding this test. I had documented in GpuOverrides but forgot to add the test when pivot values has duplicates.

revans2 (approved):

PR #5424, opened 2022-05-04
Exclude all unicode line terminator characters from matching dot

### LINKED ISSUE #5415: [BUG] Regular Expressions: matching the dot `.` doesn't fully exclude all unicode line terminator characters
**Describe the bug**
When using the wildcard `.`, the plugin will transpile it to `[^\r\n]`, because `.` in Java excludes line terminator characters when using the `.` wildcard, while cuDF does not (matches any character). However, this not exclude other unicode line terminator characters described [here](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#lt) in the section **Line terminators**. We should also exclude the next-line (`\u0085`), line-separator (`\u2028`), and paragraph-separator (`\u2029`) characters as well.

**Expected behavior**
The wildcard `.` should exclude these extra unicode line terminator characters in addition to carriage-return and newline.

**Additional context**
None

anthony-chang:

Fixes #5415

The wildcard `. ` matching should exclude all line terminators [defined in Java](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#lt) (Line terminator section). This PR adds exclusions for next-line (\u0085), line-separator (\u2028), and paragraph-separator (\u2029) characters.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
@@ -565,7 +565,8 @@ class CudfRegexTranspiler(mode: RegexMode) {
       case RegexChar(ch) => ch match {
         case '.' =>
           // workaround for https://github.com/rapidsai/cudf/issues/9619
-          RegexCharacterClass(negated = true, ListBuffer(RegexChar('\r'), RegexChar('\n')))
+          val terminatorChars = ListBuffer('\r', '\n', '\u0085', '\u2028', '\u2029')
+          RegexCharacterClass(negated = true, terminatorChars.map(RegexChar))
         case '$' if mode == RegexSplitMode || mode == RegexReplaceMode =>
           // see https://github.com/NVIDIA/spark-rapids/issues/4533
           throw new RegexUnsupportedException("line anchor $ is not supported in split or replace")
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
@@ -565,8 +565,9 @@ class CudfRegexTranspiler(mode: RegexMode) {
       case RegexChar(ch) => ch match {
         case '.' =>
           // workaround for https://github.com/rapidsai/cudf/issues/9619
-          val terminatorChars = ListBuffer('\r', '\n', '\u0085', '\u2028', '\u2029')
-          RegexCharacterClass(negated = true, terminatorChars.map(RegexChar))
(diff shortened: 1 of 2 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -565,7 +565,8 @@ class CudfRegexTranspiler(mode: RegexMode) {
       case RegexChar(ch) => ch match {
         case '.' =>
           // workaround for https://github.com/rapidsai/cudf/issues/9619
-          RegexCharacterClass(negated = true, ListBuffer(RegexChar('\r'), RegexChar('\n')))
+          val terminatorChars = ListBuffer('\r', '\n', '\u0085', '\u2028', '\u2029')
+          RegexCharacterClass(negated = true, terminatorChars.map(RegexChar))
```
NVnavkumar, line 569: `terminatorChars` is unnecessary here, there is already a local private val called `lineTerminatorChars` which has the same values in a Seq.

```suggestion
          RegexCharacterClass(negated = true, lineTerminatorChars.map(RegexChar))
```

NVnavkumar (changes requested):
Can we add more unit tests to test these cases? In particular add cases where these characters are in the input string, and the regular expression should match in CPU and GPU?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
anthony-chang: I had that at first but it would change the ordering of the line terminators. It would convert `. ` to `[^\n\r ... ]` instead of `[^\r\n ... ]`. It shouldn't make a difference but because of the way some of the tests check, tests would fail.

NVnavkumar: I think it would only change a few of the transpilation tests, but I don't think it would be too hard to change. The order within the character class also doesn't matter for the purposes of the regular expression, so you could do this and minimize how many tests to update.

NVnavkumar:
Changes look good. One more thing, can you try updating the variable here in `RegularExpressionTranspilerSuite.scala`, add the 3 unicode line terminators to it, and re-run the suite? This should enable the fuzz testing to generate more examples with line terminators to confirm that there aren't any edge cases missed.

```
  private val REGEXP_LIMITED_CHARS_COMMON = "|()[]{},.^$*+?abc123x\\ \t\r\n\f\u000bBsdwSDWzZ"
```

NVnavkumar (approved):

anthony-chang:
build

Issue #5361, opened 2022-04-28
Qualification tool implement Plan Parsing Execs group1 


tgravescs:
**Is your feature request related to a problem? Please describe.**

Modify qualification tool to parse the event logs for the following execs.   Get the duration, score,  is supported, etc..  should be based on changes in https://github.com/NVIDIA/spark-rapids/issues/5358

- [x] CoalesceExec
- [x] CollectLimitExec
- [ ] GlobalLimitExec
- [ ] LocalLimitExec
- [x] TakeOrderedAndProjectExec
- [x] ExpandExec
- [ ] GenerateExec
- [x] RangeExec
- [x] SampleExec
- [ ] SortExec
- [x] UnionExec
- [ ] CartesianProductExec

PR #5420, opened 2022-05-04
Qualification tool: Parsing Execs to get the ExecInfo #1 

nartal1:
This partially addresses https://github.com/NVIDIA/spark-rapids/issues/5361. Out of 12 execs in the issue, 7 are supported in this PR. I will create a follow on PR(s) for the remaining ones.
Execs supported in this PR:
- CoalesceExec 
- CollectLimitExec
- ExpandExec
- RangeExec 
- SampleExec
- TakeOrderedAndProjectExec 
- UnionExec

Added unit test which creates event logs on the fly and verifies if the Execs are supported and the scores associated with those.

PR #5426, opened 2022-05-05
Qualification tool: Parsing Execs to get the ExecInfo #2

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/5361. It is a follow on PR of https://github.com/NVIDIA/spark-rapids/pull/5420.

Execs supported in this PR: 
- GlobalLimitExec
- LocalLimitExec
- GenerateExec
- CartesianProductExec
- SortExec

Have added eventlog for testing GlobalLimit and LocalLimitExec. These execs are not present in final physical plan if `collect` is called. So the eventlog is generated using `show` (df.limit(10).show).  For rest of the execs, we create eventlogs on thr fly in the test itself.


tools/src/test/scala/com/nvidia/spark/rapids/tool/planparser/SqlPlanParserSuite.scala
```diff
@@ -82,6 +85,11 @@ class SQLPlanParserSuite extends FunSuite with BeforeAndAfterEach with Logging {
           assert(projects.forall(_.isSupported == true))
           assert(projects.forall(_.children.isEmpty))
           assert(projects.forall(_.duration.isEmpty))
+          val sorts = allChildren.filter(_.exec == "SortExec")
```
tgravescs, line 88: is it called SortExec or Sort?  The tests above for FilterExec and SortExec actually have a bug in them.
Add a check to make sure the sorts.size is 1

tgravescs (commented):
why do we have a global_local_limit_eventlog, I thought those execs were already in the nds_q86_test eventlog?

tools/src/test/scala/com/nvidia/spark/rapids/tool/planparser/SqlPlanParserSuite.scala
nartal1: Thanks for catching this. Updated the test. There are 3 Sorts in the the plan.

nartal1:
build

nartal1:
build

Issue #5312, opened 2022-04-26
[BUG]The coalesced AVRO file may contain different sync markers if the sync marker varies in the avro files being coalesced.


firestarman:
**Describe the bug**
The AVRO coalescing reading, introduced in https://github.com/NVIDIA/spark-rapids/pull/5306, will produce a big AVRO file with different sync markers when the sync marker varies in the original files. But one AVRO file should have the same sync marker in each sync block.

This bug exists because the reader simply concatenates the blocks from different files. We choose to do this is the implementation is quite simple for the quick start, and the cuDF AVRO reader ignores the sync markers for now.

**Expected behavior**
We should write the same sync marker into the coalesced AVRO file, no matter whether the original files have different sync markers.

tgravescs:
so to clarify this isn't really a problem right now because CUDF ignores the sync markers right now, but we should clean it up in case they do, correct?

PR #5428, opened 2022-05-05
Make the sync marker uniform for the Avro coalescing reader

firestarman:
This PR is to fix the issue https://github.com/NVIDIA/spark-rapids/issues/5312 by appending the same sync marker to every block that being coalesced into a single Avro file.

It has mainly

- introduced a new class named `BatchContext`, and a new method `createBatchContext` in the parent class of the coalescing reader. This new method returns a `BatchContext` who lives during the whole stage of building a coalesced memory file. By this design, the merged Avro header can be shared with the steps of estimating the output size, writing file header and building the block data for creating a batch. 
- updated the `copyBlocksData` and GPU Avro coalescing reader to support writing a given sync marker.
- added the related tests.

### Performance on Local
 - CPU 12 cores, and one GPU (Titan V, with 12GB memory)
- Non-partitioned 2000 avro files, 4.4GB in total in **LOCAL** storage

  ||CPU|PERFILE|COALESCING (non-uniform-sync)|COALESCING (uniform-sync)|
  |--|--|--|--|--|
  |time(sec)|27.844|24.758|16.005|15.713|

The numbers above show this change will not lead to any perf regression for the coalescing reader.

closes https://github.com/NVIDIA/spark-rapids/issues/5312

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMultiFileReader.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMultiFileReader.scala
@@ -562,6 +562,17 @@ trait SingleDataBlockInfo {
 }
 
 /**
+ * A context lives during the whole process of reading partitioned files
+ * to a batch buffer (aka HostMemoryBuffer) to build a memory file.
+ * Children can extend this to add more necessary fields.
+ * @param origChunkedBlocks a map with file as the key, and its blocks as the value
+ * @param schema shema info
+ */
+class BatchContext(
+  val origChunkedBlocks: LinkedHashMap[Path, ArrayBuffer[DataBlockBase]],
+  val schema: SchemaBase) {}
+
+/**
  * The abstracted multi-file coalescing reading class, which tries to coalesce small
  * ColumnarBatch into a bigger ColumnarBatch according to maxReadBatchSizeRows,
  * maxReadBatchSizeBytes and the [[checkIfNeedToSplitDataBlock]].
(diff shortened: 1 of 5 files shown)


firestarman:
build

jlowe:
I'm confused on the reported performance results.  Shouldn't this be reporting the performance of the GPU coalescing reader before/after this change rather than comparing the performance against the CPU?  Seems like we're changing way too many variables between the two setups to isolate the performance impact of any single change (i.e.: changing CPU to GPU, non-coalescing to coalescing, uniform sync vs. non-uniform sync, etc.).

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMultiFileReader.scala
```diff
@@ -733,11 +744,30 @@ abstract class MultiFileCoalescingPartitionReaderBase(
...
+   * Return a batch context which will be shared during the process of building a memory file,
+   * aka with the following APIs.
+   *   - calculateEstimatedBlocksOutputSize
+   *   - writeFileHeader
+   *   - getBatchRunner
+   *   - calculateFinalBlocksOutputSize
+   *   - writeFileFooter
+   * It is useful when something is needed by some or all of the above APIs.
+   * Children can override this to return a customized batch context.
+   * @param chunkedBlocks a map with file as the key, and its stripes as the value
```
jlowe, line 763:
```suggestion
   * @param chunkedBlocks mapping of file path to data blocks
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMultiFileReader.scala
```diff
@@ -733,11 +744,30 @@ abstract class MultiFileCoalescingPartitionReaderBase(
...
+   * @param clippedSchema shema info
```
jlowe, line 764:
```suggestion
   * @param clippedSchema schema info
```

firestarman:
> I'm confused on the reported performance results. Shouldn't this be reporting the performance of the GPU coalescing reader before/after this change rather than comparing the performance against the CPU? Seems like we're changing way too many variables between the two setups to isolate the performance impact of any single change (i.e.: changing CPU to GPU, non-coalescing to coalescing, uniform sync vs. non-uniform sync, etc.).

I thought this perf can show us even with this change, we still get better perf than CPU, will update it accordingly.

firestarman:
build

jlowe (approved):

PR #5432, opened 2022-05-06
Test case insensitive reading for Parquet and CSV

### LINKED ISSUE #5256: [FEA] Can Jenkins show test summary for IT tests
It would be nice if jenkins ui could show a summary of the test pass/fail like junit where it has number passed and failed, and could easily see which tests failed without having to look through log output.

One thing I remember seeing is junitxml=TEST-integration_test_result.xml.
Could se use that and integrate with  jenkins to display?

### LINKED ISSUE #5262: [BUG] case insensitive column reading might now work for parquet/CSV
**Describe the bug**
I don't have a repro case for this right now, but looking at the code it is clear that parquet, orc, and CSV look at spark.sql.caseSensitive to decide if the column names should be normalized or not. For parquet/orc we look at the names and normalize them to lower case when re-writing the files, but from what I can see we do not change the names of the columns in the files to be normalized, and CUDF does not support matching names in a case-insensitive way. The only place that I find this config even set is in an ORC test. At a minimum we need to test this for CSV and Parquet too.

HaoYang670:
Close #5262.

# Changes
Add 2 tests to confirm that when `spark.sql.caseSensitive = False`, GPU can normalize column names to lower case when reading the data frame.

No API changes.

### DIFF
--- a/integration_tests/src/main/python/csv_test.py
+++ b/integration_tests/src/main/python/csv_test.py
@@ -528,3 +528,15 @@ def test_round_trip_for_interval(spark_tmp_path, v1_enabled_list):
     assert_gpu_and_cpu_are_equal_collect(
         lambda spark: spark.read.schema(schema).csv(data_path),
         conf=updated_conf)
+
+@allow_non_gpu(any = True)
+def test_csv_read_case_insensitivity(spark_tmp_path):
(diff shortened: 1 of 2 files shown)


HaoYang670:
build

jlowe (approved):

PR #5434, opened 2022-05-07
Check schema compatibility when building parquet readers

### LINKED ISSUE #5200: [BUG] More detailed logs to show which parquet file and which data type has mismatch.
If we query a Hive Partition table whose DDL is using Bigint, however if the underline parquet data is using int, it will fail due to column type mismatch.

GPU Spark errors:
```
ai.rapids.cudf.CudfException: cuDF failure at: ../src/join/hash_join.cu:391: Mismatch in joining column data types
```

CPU Spark errors:
```
Caused by: org.apache.spark.sql.execution.QueryExecutionException: Parquet column cannot be converted in file file:///home/xxx/data/hive/testbigint/2022/part-00010-3d360f88-2fd7-4a22-b72e-ac055bb2c955-c000.snappy.parquet. Column: [b], Expected: bigint, Found: INT32
```

**How to repro:**
1. Hive CLI:
```
drop table testbigint;
create external table testbigint(a string, b bigint) PARTITIONED BY (k string)
STORED AS PARQUET LOCATION '/home/xxx/data/hive/testbigint';

create external table testbigint_dim(b string)
STORED AS PARQUET LOCATION '/home/xxx/data/hive/testbigint_dim';
```

2. Spark shell:
```
import spark.implicits._
import org.apache.spark.sql.Row

val data = Seq(
    Row("Adam",1000),
    Row("Bob",2000),
    Row("Cathy",9999)
)

val schema = StructType( Array(
                 StructField("a", StringType,true),
                 StructField("b", IntegerType,true)
             ))

val df = spark.createDataFrame(spark.sparkContext.parallelize(data),schema)

df.write.format("parquet").mode("overwrite").save("/home/xxx/data/hive/testbigint/2022")


val data2 = Seq(
    Row("Adam",1000L),
    Row("Bob",2000L),
    Row("Cathy",9999L)
)

val schema2 = StructType( Array(
                 StructField("a", StringType,true),
                 StructField("b", LongType,true)
             ))
val df2 = spark.createDataFrame(spark.sparkContext.parallelize(data2),schema2)
df2.write.format("parquet").mode("overwrite").save("/home/xxx/data/hive/testbigint/2021")


val data3 = Seq(Row("1000"),Row("2000"))
val schema3 = StructType( Array(StructField("b", StringType,true)))
val df3 = spark.createDataFrame(spark.sparkContext.parallelize(data3),schema3)
df3.write.format("parquet").mode("overwrite").save("/home/xxx/data/hive/testbigint_dim/")
```

3. Hive CLI
```
ALTER TABLE testbigint ADD PARTITION (k='2022') LOCATION '/home/xxx/data/hive/testbigint/2022';
ALTER TABLE testbigint ADD PARTITION (k='2021') LOCATION '/home/xxx/data/hive/testbigint/2021';
```

4. Spark shell
```
spark.conf.set("spark.rapids.sql.enabled",true)
spark.sql("""select count(*) from testbigint f, testbigint_dim d where  f.b=d.b  """).show

spark.conf.set("spark.rapids.sql.enabled",false)
spark.sql("""select count(*) from testbigint f, testbigint_dim d where  f.b=d.b  """).show
```

sperlingxx:
Fixes #5200 #5445

Currently, there exists no check of schema compatibility between file schema (parquet type) and read schema (spark type) when building GPU parquet readers. Thus, reading parquet data with incompatilble spark types is actually allowed on the GPU (such as: reading int as long), which leads to a lot of underfined behaviors. 

The purpose of this PR is to add the schema check, referring the checking process of Spark. Converters downcasting INT32 to Byte/Short/Date are added in this PR as well.

Note: This PR uses some deprecated API of parquet-mr, in order to accommodate Spark 3.1.


wbo4958:
LGTM

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -428,6 +431,177 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
...
+   * of an older version.
+   */
+  @scala.annotation.nowarn("msg=method getDecimalMetadata in class PrimitiveType is deprecated")
+  private def checkPrimitiveCompat(pt: PrimitiveType,
+                                   dt: DataType,
+                                   errorCallback: () => Unit): Unit = {
+    pt.getPrimitiveTypeName match {
+      case PrimitiveTypeName.BOOLEAN if dt == DataTypes.BooleanType =>
+      case PrimitiveTypeName.BOOLEAN =>
+        errorCallback()
```
jlowe, line 510: Nit: Any reason to not collect all the `errorCallback` cases at the end with a simple `case _ => errorCallback()` as the original Spark code essentially does?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -428,6 +431,177 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
...
+
+      // TODO: After we deprecate Spark 3.1, add YearMonthIntervalType
+      case PrimitiveTypeName.INT32 if dt == DataTypes.IntegerType =>
+      case PrimitiveTypeName.INT32 if dt == DataTypes.ByteType =>
+      case PrimitiveTypeName.INT32 if dt == DataTypes.ShortType =>
```
jlowe, line 515: Is this going to do the right thing or do we need an explicit downcast after cudf performs the load?  Similar casting concerns in other scenarios where Spark is casting as part of their updater selection (e.g.: Parquet uint32 to int64, Parquet uint64 -> Decimal(20,0),  int64->decimal32 etc.).  Ideally there should be some test cases to exercise these implicit conversion code paths to make sure we're doing the right thing.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -428,6 +431,177 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
...
+      case PrimitiveTypeName.INT32 if dt == DataTypes.DateType =>
+      // TODO: After we deprecate Spark 3.1, replace OriginalType with LogicalTypeAnnotation
+      case PrimitiveTypeName.INT32 if dt == DataTypes.LongType &&
+        pt.getOriginalType == OriginalType.UINT_32 =>
```
jlowe, line 519: I don't see Spark's updater code checking the original type, so curious why we're doing so here?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -428,6 +431,177 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
...
+      case dt =>
+        checkPrimitiveCompat(fileType.asPrimitiveType(),
+          dt,
+          () => errorCallback(rootFileType.get, rootReadType.get))
+    }
+  }
+
+  /**
+   * Check the compatibility over primitive types. This function refers to the `getUpdater` method
+   * of [[org.apache.spark.sql.execution.datasources.parquet.ParquetVectorUpdaterFactory]].
```
jlowe, line 496: I think this code would benefit if it were structured more like the Spark code so it's easier to follow along with the original and track changes that are needed.  IMHO translating the original code to a complicated match statement did not always result in improved readability.

sperlingxx: I fully agree on keeping the original code structure as much as we can. I rewrote the primitive check to follow Spark's style.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
sperlingxx: Something wrong will happen after cudf performs the load, we should conduct explicit downcasts. I filed a seperate issue for it: #5445.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
sperlingxx: Fixed.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -428,6 +431,177 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
...
+   * @param rootReadType read type of each root column
+   */
+  private def checkSchemaCompat(fileType: Type,
+                                readType: DataType,
+                                errorCallback: (Type, DataType) => Unit,
+                                rootFileType: Option[Type] = None,
+                                rootReadType: Option[DataType] = None): Unit = {
+    readType match {
+      case struct: StructType =>
+        // case insensitive field matching
```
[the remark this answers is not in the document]
sperlingxx: Yes, I made the change.

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
sperlingxx: This is an alterative of `IntLogicalAnnotation`. Spark 3.1 depends on parquet-mr 10.1, which doesn't have that feature.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -428,6 +432,191 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
...
+                                readType: DataType,
+                                errorCallback: (Type, DataType) => Unit,
+                                isCaseSensitive: Boolean,
+                                rootFileType: Option[Type] = None,
+                                rootReadType: Option[DataType] = None): Unit = {
+    readType match {
+      case struct: StructType =>
+        val fileFieldMap = fileType.asGroupType().getFields.asScala
+          .map { f =>
+            (if (isCaseSensitive) f.getName else f.getName.toLowerCase) -> f
```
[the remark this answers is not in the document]
sperlingxx: Done.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -521,11 +522,11 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
...
           return
         }
-        // TODO: enable below matches after we support downcast INT32 during parquet reading
-        //  https://github.com/NVIDIA/spark-rapids/issues/5445
-        // if (dt == DataTypes.ByteType || dt == DataTypes.ShortType || dt == DataTypes.DateType) {
-        //   return
-        // }
+        // TODO: Add below converters for INT32. Converters work when evolving schema over cuDF
+        //  table read from Parquet file. https://github.com/NVIDIA/spark-rapids/issues/5445
+         if (dt == DataTypes.ByteType || dt == DataTypes.ShortType || dt == DataTypes.DateType) {
```
sperlingxx, line 527: I uncommented this combinations because the UT cases of parquet writing fails if it is disabled. I will fix #5445 ASAP.

sperlingxx:
build

sperlingxx:
build

jlowe (approved):

sperlingxx:
build

sperlingxx:
build

firestarman (approved):
restore the approval

PR #4630, opened 2022-01-26
Fix the incomplete capture of SubqueryBroadcast 

sperlingxx:
Signed-off-by: sperlingxx <lovedreamf@gmail.com>

Fixes #4625 

This PR is to fix the logic on searching/matching  potential `SubqueryBroadcastExec` from partition filters of scans. In previous, we missed a possible condition: As a subquery, `SubqueryBroadcastExec` may be reused.  And this missing will lead to a crash when AQE is on.

PR #5465, opened 2022-05-11
Fix query planning issue when Iceberg is used with DPP and AQE

### LINKED ISSUE #5464: [BUG] Query failure with java.lang.AssertionError when using partitioned Iceberg tables
**Describe the bug**
When running the NDS 2.0 Power Run locally using partitioned Iceberg tables with Parquet data, the following exception is encountered when query 7 is run:

```
Caused by: java.lang.AssertionError: assertion failed
	at scala.Predef$.assert(Predef.scala:208)
	at org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec.$anonfun$doExecuteBroadcast$1(AdaptiveSparkPlanExec.scala:359)
	at org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec.withFinalPlanUpdate(AdaptiveSparkPlanExec.scala:366)
	at org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec.doExecuteBroadcast(AdaptiveSparkPlanExec.scala:358)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeBroadcast$1(SparkPlan.scala:197)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeQuery$1(SparkPlan.scala:222)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
	at org.apache.spark.sql.execution.SparkPlan.executeQuery(SparkPlan.scala:219)
	at org.apache.spark.sql.execution.SparkPlan.executeBroadcast(SparkPlan.scala:193)
	at org.apache.spark.sql.execution.SubqueryBroadcastExec.$anonfun$relationFuture$2(SubqueryBroadcastExec.scala:81)
	at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withExecutionId$1(SQLExecution.scala:139)
	at org.apache.spark.sql.execution.SQLExecution$.withSQLConfPropagated(SQLExecution.scala:163)
	at org.apache.spark.sql.execution.SQLExecution$.withExecutionId(SQLExecution.scala:137)
	at org.apache.spark.sql.execution.SubqueryBroadcastExec.$anonfun$relationFuture$1(SubqueryBroadcastExec.scala:78)
	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:659)
	at scala.util.Success.$anonfun$map$1(Try.scala:255)
	at scala.util.Success.map(Try.scala:213)
	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
	at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
```

Create a partitioned table using Iceberg parquet data and run NDS query 7 against it using the Spark RAPIDS plugin (latest 22.06 version).

**Expected behavior**
Expectation is that query will succeed and not fail with an exception.

Environment is local Linux desktop with RTX A5000 GPU.  Spark version is 3.2.1 with standard benchmark configuration.

**Additional context**
N/A

jlowe:
Fixes #5464.

#4630 added code to fixup the case where the subquery produces GPU-formatted data but we cannot replace the FileSourceScanExec.  Iceberg uses DataSource V2, so it appears as BatchScanExec rather than FileSourceScanExec.  Since we currently do not replace Iceberg reads, we end up with the same problem that the changes in #4630 was intended to solve, where the subquery is GPU-formatted but we don't replace the exec that is examining the subquery results.  Since the GPU data is columnar, Spark automatically inserts a ColumnarToRow between the AdaptiveSparkPlanExec and the BroadcastQueryStageExec which triggers the assert in Spark.

This adds similar handling for BatchScanExec as #4630 added for FileSourceScanExec.  In Spark 3.2+ BatchScanExec can have a dynamic runtimeFilter subquery which needs to be handled separately from the BatchScanExec in case we can translate one without translating the other.

### DIFF
--- a/integration_tests/src/main/python/iceberg_test.py
+++ b/integration_tests/src/main/python/iceberg_test.py
@@ -12,9 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import pytest
+
 from asserts import assert_gpu_and_cpu_are_equal_collect
-from marks import allow_non_gpu, iceberg
-from spark_session import with_cpu_session
+from data_gen import *
+from marks import allow_non_gpu, iceberg, ignore_order
+from spark_session import is_before_spark_320, is_databricks_runtime, with_cpu_session
 
 @allow_non_gpu('BatchScanExec')
 @iceberg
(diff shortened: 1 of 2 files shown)


jlowe:
build

tgravescs (approved):

PR #5477, opened 2022-05-12
Add support for `\h`, `\H`, `\v`, `\V`, and `\R` character classes

### LINKED ISSUE #4605: [FEA] Add regular expression support for new character classes introduced in Java 8
Java 8 added support for the character classes `\h`, `\H`, `\v\`, `\V`, and `\R`

https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html

We should add support for these character classes instead of falling back to CPU. 

None

**Additional context**
None

anthony-chang:
Closes #4605

Java 8 added support for the character classes `\h`, `\H` (horizontal whitespace), `\v`, `\V` (vertical whitespace), and `\R` (unicode linebreak sequence). This adds support for these character classes on GPU by transpiling to equivalent character classes, which can be found here: https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html. 

EDIT:
A couple updates to this PR:
- I've moved the transpiling of the character classes from `rewrite` to the `parse` section to handle the case where cuDF and Java treats newlines for negative character matches differently (which is dealt with in the `RegexCharacterClass` case in `rewrite`). By expanding the character classes earlier in `parse`, we can then make any modifications to the AST such as this case in `rewrite`
- I've disabled support for repetitions of `\R` due to an inconsistency within Java. The Java docs says `\R` is equivalent to `\r\n|[\n\u000B\u000C\r\u0085\u2028\u2029]` but if we run the pattern `\R{2}` against the input `a\r\nb`, Java finds no matches, whereas cuDF finds 1 match (which is correct).


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -675,6 +675,28 @@ class CudfRegexTranspiler(mode: RegexMode) {
             RegexChar(' '), RegexChar('\u000b'))
           chars ++= Seq('n', 't', 'r', 'f').map(RegexEscaped)
           RegexCharacterClass(negated = ch.isUpper, characters = chars)
+        case 'h' | 'H' =>
```
revans2, line 678: nit: Where did you get these unicode character lists from?  Might be nice to include a comment so others can reproduce the work or update it if needed.

revans2 (approved):

anthony-chang:
build

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -258,6 +258,24 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
       Seq("\u001eTEST"))
   }
 
+  test("horizonal whitespace - find") {
```
NVnavkumar, line 254: I think now that we have the Python integration tests, we might not need these Scala tests anymore.

anthony-chang:
build

integration_tests/src/main/python/string_test.py
```diff
@@ -867,6 +867,40 @@ def test_regexp_whitespace():
...
+        '''\xA0\u1680\u180e[abcd]\t\n{1,3} [0-9]\n {1,3}\x0b\t[abcd]\r\f[0-9]{0,10}
+            [\u2001-\u200a]{1,3}\u202f\u205f\u3000\x85\u2028\u2029
+        ''')
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark: unary_op_df(spark, gen).selectExpr(
+                'rlike(a, "\\h{2}")',
+                'rlike(a, "\\v{3}")',
+                'rlike(a, "[abcd]+\\h+[0-9]+")',
+                'rlike(a, "[abcd]+\\v+[0-9]+")',
+                'rlike(a, "\\H{3}")',
```
revans2, line 881: I think this may have the same problems as in https://github.com/NVIDIA/spark-rapids/pull/5479/files#r872460659

We might want to go back and verify that all of the regexp tests in here are doing what we expect them to do.

anthony-chang:
build

Issue #5478, opened 2022-05-12
[BUG] Inconsistent handling of word boundaries `\b` and `\B` with StringSplit for regular expressions


anthony-chang:
**Describe the bug**
We currently fallback to CPU for regular expressions that include `\b` or `\B` in `split` mode since the behaviour is not consistent with Spark. 

**Steps/Code to reproduce bug**
If we enable word boundaries in RegexSplitMode, the test `"string split fuzz"` in `RegularExpressionTranspilerSuite` produces this output:
```
string split fuzz *** FAILED ***
  string_split pattern=((\b)) isRegex=true data=xb\nB limit=-2 
  CPU [4]: xb, \n, B,  
  GPU [4]: , xb, \n, B 
```

**Expected behavior**
The CPU and GPU should be consistent. 

**Environment details (please complete the following information)**
None

**Additional context**
None

anthony-chang:
Depends on https://github.com/rapidsai/cudf/issues/11102

anthony-chang:
There are still a number of inconsistencies, particularly when the pattern consists entirely of word boundaries. cuDF/python has an extra empty token in the array returned by string split which Spark does not. Examples:

```
- \b *** FAILED ***
  string_split pattern=\b isRegex=true data=a limit=-1 
  CPU [2]: a,  
  GPU [3]: , a, (RegularExpressionTranspilerSuite.scala:764)

- \b\b *** FAILED ***
  string_split pattern=\b\b isRegex=true data=a limit=-1 
  CPU [2]: a,  
  GPU [3]: , a, (RegularExpressionTranspilerSuite.scala:764)

- \B\B *** FAILED ***
  string_split pattern=\B\B isRegex=true data=-+ limit=-1 
  CPU [3]: -, +,  
  GPU [4]: , -, +, (RegularExpressionTranspilerSuite.scala:764)
```

It is possible to use a workaround for this though. We can check if the pattern consists of entirely word or non-word boundaries:
```scala
def isEntirely(pattern: RegexAST, component: RegexAST): Boolean = {
  pattern match {
    case RegexSequence(parts) => parts.forall(isEntirely(_, component))
    case RegexGroup(_, term) => isEntirely(term, component)
    case RegexChoice(l, r) => isEntirely(l, component) || isEntirely(r, component)
    case `component` => true
    case _ => false
  }
}

[...]

PR #5479, opened 2022-05-12
Add support for word boundaries `\b` and `\B`

### LINKED ISSUE #4289: [FEA] Regexp: Add support for word and non-word boundaries in regexp pattern
We currently fall back to CPU for regexp patterns containing word (`\b`) and non-word (`\B`) boundaries in regexp patterns.

Here is one example of a difference between CPU and GPU for regexp_replace.

The test is effectively running `regexp_replace("A\nB", "\b", "_REPLACE")`.

CPU output:

```
_REPLACE_A_REPLACE_\n_REPLACE_B_REPLACE_
```

GPU output:

```
_REPLACE__REPLACE__REPLACE_A\nB
```

Support on GPU consistently with Spark.

None

**Additional context**
None

### LINKED ISSUE #4517: [FEA] Add support for word boundaries `\b` and `\B` in regular expressions
We currently fall back to CPU for regular expressions that include `\b` or `\B`.

We should support this on the GPU.

None

**Additional context**
None

anthony-chang:
Closes #4517, closes #4289

cuDF now supports word boundaries so we no longer need to fallback to CPU for them. However, there are consistencies from cuDF  (see #5478) so word boundaries are still disabled for GpuStringSplit. 


NVnavkumar (changes requested):
A different situation here; can you add a unit test for the fallback to CPU when in split mode?

anthony-chang:
build

integration_tests/src/main/python/string_test.py
```diff
@@ -836,6 +836,20 @@ def test_regexp_extract_idx_0():
                 'regexp_extract(a, "^([a-d]*)[0-9]*([a-d]*)\\z", 0)'),
         conf=_regexp_conf)
 
+def test_word_boundaries():
+    gen = mk_str_gen('([abcd]{1,3}[\n\t\r\f ]{0,2}[efgh]){1,5}')
+    assert_gpu_and_cpu_are_equal_collect(
+            lambda spark: unary_op_df(spark, gen).selectExpr(
+                'rlike(a, "\\b")',
```
revans2, line 843: I am a little confused by this. I manually tested rlike, regexp_extract and regexp_replace with strings that match the string gen pattern given and none of them match. rlike is always false, regexp_extract always returns an empty string and regexp_replace does not change the input string at all.  This is the same on both the CPU and the GPU.

My main concern is that we are not really testing these because we don't have anything where they match. Perhaps these patterns just never match on Spark in these situations. If so, then we should document it. If not then I really would like a test that causes them to match.

revans2: Actually ti turned out to be user error on my part in scala. I needed to double escape the \b and \B, i.e. \\B. Now I really just want to make sure that you have verified that it is hitting the code path you want. :)

revans2: Yes I had to change all of these to be double escaped to make the patterns work.

```
>>> spark.createDataFrame([("a e", "b a")], schema='a string,b string').selectExpr('regexp_replace(a, "\\\\b", "*") as works', 'regexp_replace(a, "\\b", "*") as not').show()

+-------+---+
|  works|not|
+-------+---+
|*a* *e*|a e|
+-------+---+
```

Also replacing boundaries which are not a character with an empty string is a NOOP. You really should replace them with something like '*' or anything that would make them appear.

anthony-chang: In that case, it looks like we need to change all character classes/string anchors supported by cuDF to use the double escape `\\\\` for every integration test here, correct?

revans2: I think so, but I am not 100% sure. I was running in the python repl so there could be some odd differences there. I would suggest that everywhere you suspect there may be some issues that you insert in a `data_gen.debug_df` to check that the output is exactly what you would expect and then if what I found is true, update the tests accordingly.

To keep the scope limited I would just update test new tests added for this PR and then you can file a follow on BUG and/or PR to fix the other tests.

NVnavkumar: I agree with changing all the integration tests. In my testing of other issues, I have confirmed the necessity to double escape any regular expression specific escape character in order to do the test properly (I *think* this is a Python specific behavior). Let's handle this in a follow-on issue.

NVnavkumar (approved):

integration_tests/src/main/python/string_test.py
NVnavkumar: @anthony-chang can you change at least this integration test to double-escape the `b` and `B` and add the `*` (or any other character replacement) that Bobby suggested?

NVnavkumar (changes requested):

revans2 (approved):

anthony-chang:
build

anthony-chang:
build

anthony-chang:
build

anthony-chang:
The tests fail with this case 
```
javaPattern=(?:w)\Z\B, cudfPattern=(?:w)([\n\r\u0085\u2028\u2029]|\r\n)?$\B, input='z11\u000b^+ww\r', cpu=z11\u000b^+ww\r, gpu=z11\u000b^+w_RE\PLACE_\r
```
I will need to look into whether we can support word boundaries surrounding string anchors. If not, we will need to fallback to CPU similar to in #5610

anthony-chang:
build

anthony-chang:
build

anthony-chang:
build

andygrove (changes requested):
This should be for 22.08 now

andygrove (approved):

andygrove:
build

sameerz:
build

NVnavkumar (approved):

PR #5485, opened 2022-05-13
Add cudaStreamSynchronize when a new device buffer is added to the spill framework

### LINKED ISSUE #4818: [BUG] ASYNC: the spill store needs to synchronize on spills against the allocating stream
When we register a buffer against the spill framework, the CPU may not be synchronized with the GPU stream that is adding the buffer. The GPU stream that is doing the freeing is also not synchronized. This is by design, and it is a good feature to have. We are tracking handles to GPU memory, but the state of the GPU streams is not something we are as concerned with at registration time.

At spill time that's a different story. Since spills happen from a task thread that ran OOM, the freeing CUDA stream may be (often is) different than the allocating stream (the stream that registered the buffer to begin with). This may mean we ask RMM or CUDA to free a pointer that isn't quite done.

In the recent move to ASYNC allocator it looks like it can be even worse: we could allocate a buffer asynchronously and turn around to spill it before it is done allocating (in the extreme).

Given PTDS, we can't reach at the allocating stream to introduce a synchronize, since the application only knows about a stream Id that is automatically managed by CUDA (it is the same ID for all threads from the application's point of view). A thread-local CUDA event could be used to manage this where, at buffer registration, an event is told to record the stream's actions up until that point. If the buffer needs to be spilled, if the spilling thread is different than the producing thread, we could then synchronize on the producing thread's CUDA event (which the `RapidsBuffer` needs to point to).

This is related to https://github.com/NVIDIA/spark-rapids/issues/4710.

abellina:
Closes #4818.

This is to fix the stream-ordered violation we could see currently in the spill framework: a buffer is allocated and acted on in stream A, and added to the store without any synchronization. Then stream B runs OOM and spills such buffer, not having waited on stream A.

The idea here is that if a device buffer is added to the store, we force in most cases a call to cudaStreamSynchronize, which should take care of the cases where we return from cuDF and have not stream synchronized there. There is a lot of stream synchronization in cuDF (via thrift usually) so it seems like this is an unlikely issue. That said, adding this synchronize, would be more robust as now the spill framework knows for a fact that the buffer can be freely copied (spilled).

Note that I added a flag to not synchronize in some cases, this is used for UCX since we synchronize on writes after compressing buffers, and also on reads after copying from the bounce buffer, so those two cases we can reliably say we have synchronized. I don't know of other cases where the synchronize in the spill framework can be skipped.

I have been measuring the impact of this in our benchmarks. I do not see any big differences so far in UCX or non-UCX runs. I am going to continue testing it today. Adding as WIP for comments.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsBufferStore.scala
@@ -138,8 +138,7 @@ abstract class RapidsBufferStore(
   def copyBuffer(buffer: RapidsBuffer, memoryBuffer: MemoryBuffer, stream: Cuda.Stream)
   : RapidsBufferBase = {
     freeOnExcept(createBuffer(buffer, memoryBuffer, stream)) { newBuffer =>
-      buffers.add(newBuffer)
-      catalog.registerNewBuffer(newBuffer)
+      addBuffer(newBuffer)
       newBuffer
     }
   }
@@ -208,6 +207,7 @@ abstract class RapidsBufferStore(
    * If the data transfer will be performed asynchronously, this method is responsible for
    * adding a reference to the existing buffer and later closing it when the transfer completes.
    * @note DO NOT close the buffer unless adding a reference!
+   * @note `createBuffer` impls should synchronize against `stream` before returning, if needed.
    * @param buffer data from another store
    * @param memoryBuffer memory buffer obtained from the specified Rapids buffer. The ownership
    *                     for `memoryBuffer` is transferred to this store. The store may close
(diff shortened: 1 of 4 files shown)


jlowe (dismissed):

abellina:
The fix https://github.com/NVIDIA/spark-rapids/pull/5485/commits/5ad4d43c68ad0a7cfc52dcbfb94caea5eae6d9f1 helps with some queries I saw an obvious regression (NDS q20 specifically). This makes it so that contig split is followed by a sync in the write path for the RapidsShuffleManager. I am still not done assessing the performance impacts of this, I should have numbers Monday.

abellina:
All tests I have executed with this PR show results in the noise, so I do not see blockers.

abellina:
build

jlowe (dismissed):

abellina:
ok I have related test failures, due to some mock object matchers I had used. I'll fix that soon

abellina:
build

jlowe (approved):

PR #5487, opened 2022-05-13
Add limited support for captured vars and athrow

seanprime7:
Only the primitive type variables captured from a method are supported.
athrow is supported only with SparkException and is converted to raise_error.

Following additional changes have also been made:
* A check to reject lambdas with void return type has been added.
* An operand stack bug fix for consturctor calls has been added.

This commit also simplifies the code that handles method calls.

Signed-off-by: Sean Lee <selee@nvidia.com>

### DIFF
--- a/udf-compiler/src/main/scala/com/nvidia/spark/udf/CFG.scala
+++ b/udf-compiler/src/main/scala/com/nvidia/spark/udf/CFG.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/udf-compiler/src/main/scala/com/nvidia/spark/udf/CFG.scala
+++ b/udf-compiler/src/main/scala/com/nvidia/spark/udf/CFG.scala
@@ -108,7 +108,7 @@ case class BB(instructionTable: SortedMap[Int, Instruction]) extends Logging {
         val defaultState = state.copy(cond = simplify(defaultCondition))
         newStates + (defaultSucc -> defaultState.merge(newStates.get(defaultSucc)))
       case Opcode.IRETURN | Opcode.LRETURN | Opcode.FRETURN | Opcode.DRETURN |
-           Opcode.ARETURN | Opcode.RETURN => states
+           Opcode.ARETURN | Opcode.RETURN | Opcode.ATHROW => states
       case _ =>
         val (0, successor) :: Nil = cfg.successor(this)
         // The condition, stack and locals from the current BB state need to be
--- a/udf-compiler/src/main/scala/com/nvidia/spark/udf/CatalystExpressionBuilder.scala
+++ b/udf-compiler/src/main/scala/com/nvidia/spark/udf/CatalystExpressionBuilder.scala
@@ -176,7 +176,36 @@ case class CatalystExpressionBuilder(private val function: AnyRef) extends Loggi
     // A basic block can have other branching instructions as the last instruction,
     // otherwise.
     if (basicBlock.lastInstruction.isReturn) {
-      newStates(basicBlock).expr
+      // When the compiled method throws exceptions, we combine the returned
+      // expression (expr) with exceptions (except0, except1, except2.., exceptn)
+      // based on exception conditions (cond0, cond1, cond2, .., condn) as follows:
+      //         If
+      //         |
+      //  +------+------+
+      //  |      |      |
+      // cond0 except0  If
+      //                |
+      //         +------+------+
+      //         |      |      |
+      //        cond1 except1  If
+      //                       |
+      //                +------+------+
+      //                |      |      |
+      //               cond2 except2  If
+      //                              |
+      //                              .
+      //                              .
+      //                              .
+      //                       +------+------+
+      //                       |      |      |
+      //                      condn exceptn  expr
+      newStates.foldRight(newStates(basicBlock).expr) { case ((bb, s), expr) =>
+        if (bb.lastInstruction.isThrow) {
+          Some(If(s.cond, s.stack.head, expr.get))
+        } else {
+          expr
+        }
+      }
     } else {
       // account for this block in visited
       val newVisited = visited + basicBlock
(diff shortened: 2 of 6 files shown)


udf-compiler/src/main/scala/com/nvidia/spark/udf/Instruction.scala
```diff
@@ -323,6 +349,17 @@ case class Instruction(opcode: Int, operand: Int, instructionStr: String) extend
...
 
+  private def athrow(state: State): State = {
+    val State(locals, top :: rest, cond, expr) = state
+    if (!top.isInstanceOf[Repr.SparkExcept]) {
+      throw new SparkException("Unsupported type for athrow")
+    }
+    // Empty the stack and convert the internal representation of
+    // org.apache.spark.SparkException object to RaiseError, then push it to the
+    // stack.
+    State(locals, List(RaiseError(top.asInstanceOf[Repr.SparkExcept].message)), cond, expr)
```
abellina, line 360: I believe one approach here would be to code up a new catalyst node `RaiseErrorV2` or `RaiseSpecificError` or something like this, where it has the CPU `eval` and `doGenCode`, as it would be very much like `RaiseError`, but it has an extra parameter which is the specific type of exception to throw. It would then instantiate this exception and throw the specific type.

For the gpu, we could replace this new node with the GPU specific type, and for the CPU it would work the way the caller intended. @jlowe @revans2 for some :eyes:

abellina (changes requested):
First pass thinking about the type of catalyst exception we should convert the exception to.

udf-compiler/src/main/scala/com/nvidia/spark/udf/CFG.scala
```diff
@@ -108,7 +108,7 @@ case class BB(instructionTable: SortedMap[Int, Instruction]) extends Logging {
         val defaultState = state.copy(cond = simplify(defaultCondition))
         newStates + (defaultSucc -> defaultState.merge(newStates.get(defaultSucc)))
       case Opcode.IRETURN | Opcode.LRETURN | Opcode.FRETURN | Opcode.DRETURN |
-           Opcode.ARETURN | Opcode.RETURN => states
+           Opcode.ARETURN | Opcode.RETURN | Opcode.ATHROW => states
```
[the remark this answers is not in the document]
seanprime7: Done

udf-compiler/src/test/scala/com/nvidia/spark/OpcodeSuite.scala
```diff
@@ -2350,6 +2351,44 @@ class OpcodeSuite extends FunSuite {
...
+      val dataset = List(2, 20).toDS()
+      val result = dataset.withColumn("new", u('value))
+      val ref = dataset.withColumn("new", lit(true))
+      checkEquiv(result, ref)
+    }
+    run(30)
+    try {
+      run(20)
+    } catch {
+      case e: Throwable =>
```
abellina, line 2387: @seanprime7 could this catch the `RuntimeException` specifically?

udf-compiler/src/main/scala/com/nvidia/spark/udf/CatalystExpressionBuilder.scala
```diff
@@ -176,7 +176,36 @@ case class CatalystExpressionBuilder(private val function: AnyRef) extends Loggi
     // A basic block can have other branching instructions as the last instruction,
     // otherwise.
     if (basicBlock.lastInstruction.isReturn) {
-      newStates(basicBlock).expr
+      // When the compiled method throws exceptions, we combine the returned
```
abellina, line 179: correct me if I am wrong, the test we have doesn't show this collapse behavior. I am hoping we can get a test that shows this.

udf-compiler/src/test/scala/com/nvidia/spark/OpcodeSuite.scala
seanprime7: Done.

udf-compiler/src/main/scala/com/nvidia/spark/udf/CatalystExpressionBuilder.scala
seanprime7: actually, `test("throw a SparkException object")` tests it.
The following analyzed logical plan is produced from `run(30)`:

```
== Analyzed Logical Plan ==
value: int, new: boolean
Project [value#4297, if (((value#4297 < 0) OR (value#4297 >= 30))) cast(raise_error(concat(concat(concat(concat(concat(, Fold number must be in range [0, ), cast(30 as string)), ), but
got ), cast(value#4297 as string)), .), NullType) as boolean) else true AS new#4300]
+- LocalRelation [value#4297]
```

udf-compiler/src/main/scala/com/nvidia/spark/udf/Instruction.scala
seanprime7: Updated `udf-to-catalyst-expressions.md`

udf-compiler/src/main/scala/com/nvidia/spark/udf/CatalystExpressionBuilder.scala
abellina: Ok, I also tried a more complex example:
```
      val myudf: (Int) => Boolean = i => {
        if (i < 0 || i >= x) {
          throw new SparkException(s"Fold number must be in range [0, $x), but got $i.")
        } else if (x % 2 != 0) {
          throw new SparkException(s"Invalid fold range! [0, $x). End of the range must be an even number.")
        }
        true
      }
``` 

That produced (with `run(30)`):
```
Project [value#5, if (((value#5 < 0) OR (value#5 >= 30))) cast(raise_error(concat(concat(concat(concat(concat(, Fold number must be in range [0, ), cast(30 as string)), ), but got ), cast(value#5 as string)), .), NullType) as boolean) else if ((((value#5 >= 0) AND (value#5 < 30)) AND NOT ((30 % 2) = 0))) cast(raise_error(concat(concat(concat(, Invalid fold range! [0, ), cast(30 as string)), ). End of the range must be an even number.), NullType) as boolean) else true AS my_fn(value)#9]
```

Which seems correct, though it does check the negated boolean expression of the if branch.

abellina:
@seanprime7 ok to take this PR off draft mode?

seanprime7:
took this PR off draft mode.

abellina:
Thanks @seanprime7

Issue #5246, opened 2022-04-13
Verify Parquet columnar encryption is handled safely


jlowe:
Spark 3.2 added support for columnar encryption in Parquet.  We probably cannot detect this at query planning time, but minimally we need to make sure we are not corrupting the data during reads on encrypted Parquet files and instead throwing an exception.  Long-term we either need to directly support encrypted data or dynamic, post-planning fallback to the CPU on an encrypted Parquet read.

HaoYang670:
I'd like to have a try!

HaoYang670:
Verified on Spark3.3.  It seems like our plugin could decrypt correctly.
```scala
scala> val df = Seq((1, 2, 3)).toDF("one", "tWo", "THREE")
df: org.apache.spark.sql.DataFrame = [one: int, tWo: int ... 1 more field]

scala> sc.hadoopConfiguration.set("parquet.encryption.kms.client.class" ,
     |                            "org.apache.parquet.crypto.keytools.mocks.InMemoryKMS")

scala> sc.hadoopConfiguration.set("parquet.encryption.key.list" ,
     |                    "keyA:AAECAwQFBgcICQoLDA0ODw== ,  keyB:AAECAAECAAECAAECAAECAA==")

scala> sc.hadoopConfiguration.set("parquet.crypto.factory.class" ,
     |                    "org.apache.parquet.crypto.keytools.PropertiesDrivenCryptoFactory")

scala> df.write.
     |    option("parquet.encryption.column.keys" , "keyA:one").
     |    option("parquet.encryption.footer.key" , "keyB").
     | parquet("./tmp/parquet")
22/05/07 20:08:10 WARN GpuOverrides: 
*Exec <DataWritingCommandExec> will run on GPU
  *Output <InsertIntoHadoopFsRelationCommand> will run on GPU
  ! <LocalTableScanExec> cannot run on GPU because GPU does not currently support the operator class org.apache.spark.sql.execution.LocalTableScanExec
    @Expression <AttributeReference> one#10 could run on GPU
    @Expression <AttributeReference> tWo#11 could run on GPU
    @Expression <AttributeReference> THREE#12 could run on GPU


scala> spark.read.parquet("./tmp/parquet").show
22/05/07 20:08:27 WARN GpuOverrides: 
[...]

PR #5489, opened 2022-05-13
Fallback to the CPU if Parquet encryption keys are set [databricks]

### LINKED ISSUE #5439: [BUG] Encrypted Parquet writes are being replaced with a GPU unencrypted write
As noted in #5246, plans to write encrypted Parquet files are being replaced with a GPU Parquet write that does not encrypt.  Until the GPU supports encrypted writes, the RAPIDS Accelerator should not replace Parquet writes that have either of the following Parquet configuration options specified:
- `parquet.encryption.column.keys`
- `parquet.encryption.footer.key`

razajafri:
This PR checks for the encryption keys and if set fallback to the CPU 

fixes #5439 


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -49,6 +49,15 @@ object GpuParquetFileFormat {
 
     val parquetOptions = new ParquetOptions(options, sqlConf)
 
+    val columnEncryption = options.getOrElse("parquet.encryption.column.keys", "")
+    val footerEncryption = options.getOrElse("parquet.encryption.footer.key", "")
+
+    if (!columnEncryption.isEmpty || !footerEncryption.isEmpty) {
+      meta.willNotWorkOnGpu("Encryption is not yet supported on GPU. To enable unset the " +
```
jlowe, line 56: "To enable" is misleading, as this will not enable encryption.  I suggest wording it as "If encrypted Parquet reads or writes are not required..." to be more explicit.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -49,6 +49,15 @@ object GpuParquetFileFormat {
...
+          "\"parquet.encryption.column.keys\" and \"parquet.encryption.footer.key\" in " +
+          "parquet options")
```
jlowe, line 58:
```suggestion
          "Parquet options")
```

integration_tests/src/main/python/parquet_write_test.py
```diff
@@ -279,6 +279,19 @@ def test_parquet_write_legacy_fallback(spark_tmp_path, ts_write, ts_rebase, spar
             'DataWritingCommandExec',
             conf=all_confs)
 
+@allow_non_gpu('DataWritingCommandExec')
+def test_parquet_write_encryption_fallback(spark_tmp_path, spark_tmp_table_factory):
+    gen = IntegerGen()
+    data_path = spark_tmp_path + '/PARQUET_DATA'
+    assert_gpu_fallback_write(
```
jlowe, line 286: There needs to be a test for falling back on Parquet reads, not just writes, when either of the encrypted keys are set.  For the read test, the CPU should perform the encrypted write.

razajafri: The issue #5439 specifically talks about writes. Should we tackle the reads as part of a separate PR which will completely fix #5246 or should we expand #5439 to include reads?

jlowe: We can handle reads in a separate PR.

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -49,6 +49,15 @@ object GpuParquetFileFormat {
 
     val parquetOptions = new ParquetOptions(options, sqlConf)
 
+    val columnEncryption = options.getOrElse("parquet.encryption.column.keys", "")
+    val footerEncryption = options.getOrElse("parquet.encryption.footer.key", "")
+
+    if (!columnEncryption.isEmpty || !footerEncryption.isEmpty) {
+      meta.willNotWorkOnGpu("Encryption is not yet supported on GPU. If encrypted Parquet" +
+          "reads or writes are not required unset the \"parquet.encryption.column.keys\" and " +
```
jlowe, line 57: This code is specific to the write path, so we should not mention reads here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetFileFormat.scala
```diff
@@ -49,6 +49,15 @@ object GpuParquetFileFormat {
 
     val parquetOptions = new ParquetOptions(options, sqlConf)
 
+    val columnEncryption = options.getOrElse("parquet.encryption.column.keys", "")
+    val footerEncryption = options.getOrElse("parquet.encryption.footer.key", "")
+
+    if (!columnEncryption.isEmpty || !footerEncryption.isEmpty) {
+      meta.willNotWorkOnGpu("Encryption is not yet supported on GPU. If encrypted Parquet" +
```
jlowe, line 56:
```suggestion
      meta.willNotWorkOnGpu("Encryption is not yet supported on GPU. If encrypted Parquet " +
```

integration_tests/src/main/python/parquet_write_test.py
```diff
@@ -279,6 +279,19 @@ def test_parquet_write_legacy_fallback(spark_tmp_path, ts_write, ts_rebase, spar
...
+        lambda spark, path: unary_op_df(spark, gen).coalesce(1).write
+            .option("parquet.encryption.footer.key" , "k1")
+            .option("parquet.encryption.column.keys" , "k2:a")
```
jlowe, line 289: These options should be tested independently to verify that only one of them is required for the fallback, not both together.

integration_tests/src/main/python/parquet_write_test.py
```diff
@@ -279,6 +279,24 @@ def test_parquet_write_legacy_fallback(spark_tmp_path, ts_write, ts_rebase, spar
...
 
+@allow_non_gpu('DataWritingCommandExec')
+@pytest.mark.parametrize('write_options', [{"parquet.encryption.footer.key": "k1"},
+                                           {"parquet.encryption.column.keys": "k2:a"},
+                                           {"parquet.encryption.footer.key": "k1", "parquet.encryption.column.keys": "k2:a"}])
+def test_parquet_write_encryption_fallback(spark_tmp_path, spark_tmp_table_factory, write_options):
+    def write_func(spark, path):
+        writer = unary_op_df(spark, gen).coalesce(1).write
+        for key in write_options:
+            writer.option(key , write_options[key])
```
jlowe, line 290:
```suggestion
        for k, v in write_options.items():
            writer.option(k , v)
```

jlowe (approved):
Minor suggestion but otherwise lgtm.

razajafri:
build

razajafri:
build failed due to bad connection

razajafri:
build

razajafri:
This time it failed with `Couldn't find any revision to build. Verify the repository and branch configuration for this job.`

razajafri:
build

razajafri:
build

PR #5491, opened 2022-05-14
Enables `$`, `\z`, and `\Z` in `REGEXP_REPLACE` on the GPU

### LINKED ISSUE #4425: [FEA] Support line anchor `$` and string anchors `\z` and `\Z` in regexp_replace
The string anchor `\Z` has different behavior between cuDF and Java/Spark when used in `regex_replace` / `replace_re`.

- Java definition: The end of the input but for the final terminator, if any
- cuDF definition: Matches at the end of the string

The string anchor `\z` also has different behavior when applied to empty strings.

Once https://github.com/NVIDIA/spark-rapids/pull/4424 is merged, we at least fall back to CPU if regexp_replace is called with a pattern containing `\z` or `\Z` but we should try and support this on the GPU.

Support `\z` and `\Z` fully on the GPU.

None.

**Additional context**
None.

NVnavkumar:
Fixes #4425.

This enables the end-of-line anchor `$`, end-of-input anchor `\z`, and end-of-input-before-line-termination anchor `\Z` to be used in regular expressions in `REGEXP_REPLACE(...)` on the GPU. This does this by doing minor manipulation on the replacement string in addition to regular expression transpiling. The idea is to potentially match a line termination in a capture group and use that as an available back-reference in the replacement string to preserve it in the output to match how Spark does this on the CPU.


NVnavkumar:
build

integration_tests/src/main/python/string_test.py
```diff
@@ -561,6 +561,28 @@ def test_re_replace_backrefs():
...
+    gen = mk_str_gen('.{0,5}TEST[\ud720 A]{0,5}TEST[\r\n\u0085\u2028\u2029]?')
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: unary_op_df(spark, gen).selectExpr(
+            'REGEXP_REPLACE(a, "TEST$", "")',
+            'REGEXP_REPLACE(a, "TEST$", "PROD")',
+            'REGEXP_REPLACE(a, "\ud720[A-Z]+$", "PROD")',
+            'REGEXP_REPLACE(a, "(\ud720[A-Z]+)$", "PROD")',
+            'REGEXP_REPLACE(a, "(TEST)$", "$1")',
+            'REGEXP_REPLACE(a, "^(TEST)$", "$1")',
+            'REGEXP_REPLACE(a, "\\\\ATEST\\\\Z", "PROD")',
```
revans2, line 574: There are a huge number of possible patterns for the string gen, but only a very few of them would match this, and many of the patterns below it. Can we confirm that positive cases show up in the input data, restrict the `.` at the beginning to be smaller so there are less choices, or add in some special cases to make sure that things like "TEST\n" show up in the output?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -568,9 +675,8 @@ class CudfRegexTranspiler(mode: RegexMode) {
           val terminatorChars = new ListBuffer[RegexCharacterClassComponent]()
           terminatorChars ++= lineTerminatorChars.map(RegexChar)
           RegexCharacterClass(negated = true, terminatorChars)
-        case '$' if mode == RegexSplitMode || mode == RegexReplaceMode =>
-          // see https://github.com/NVIDIA/spark-rapids/issues/4533
-          throw new RegexUnsupportedException("line anchor $ is not supported in split or replace")
+        case '$' if mode == RegexSplitMode =>
+          throw new RegexUnsupportedException("line anchor $ is not supported in split")
```
revans2, line 679: Is there a separate follow on issue that we need for Split?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -583,6 +689,10 @@ class CudfRegexTranspiler(mode: RegexMode) {
               // Java, it's treated as a single (b$ and b$$ are synonymous), so we create 
               // an empty RegexAST that outputs to empty string
               RegexEmpty()
+            case Some(RegexChar(ch)) if mode == RegexReplaceMode
+                && lineTerminatorChars.contains(ch) =>
+                throw new RegexUnsupportedException("Regex sequences with a line terminator "
```
revans2, line 694: Can we add a fallback test to verify that this is working as expected?  Does this conflict with the test just below it? What happens if we have something like `ABC$A\n` is this going to do the right thing?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -648,17 +772,17 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
               "string anchor \\Z is not supported in split or replace mode")
         case 'z' if mode == RegexSplitMode =>
           throw new RegexUnsupportedException("string anchor \\z is not supported in split mode")
         case 'z' =>
-          if (mode == RegexReplaceMode) {
-            // see https://github.com/NVIDIA/spark-rapids/issues/4425
-            throw new RegexUnsupportedException(
-              "string anchor \\z is not supported in replace mode")
-          }
+          // if (mode == RegexReplaceMode) {
```
revans2, line 781: Should this be deleted?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -648,17 +772,17 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
+          //   // see https://github.com/NVIDIA/spark-rapids/issues/4425
+          //   throw new RegexUnsupportedException(
+          //     "string anchor \\z is not supported in replace mode")
+          // }
           // cuDF does not support "\z" but supports "$", which is equivalent
```
revans2, line 786: Again do we need a follow on issue for split?

integration_tests/src/main/python/string_test.py
NVnavkumar: updated integration test in latest commit

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
NVnavkumar: removed in latest commit

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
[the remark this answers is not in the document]
NVnavkumar: So this is a strange edge case where the behavior in Java is a little strange. Here is an example:

Input: `AB\r\r\r`
Regex: `\r$`
Replacement: "_REPLACE_"

In Java, this becomes `AB\r_REPLACE__REPLACE_` (so it randomly behaves in a multiline mode), in cuDF, this becomes `AB\r\r_REPLACE_` (which is the same in Python). Since these are anchors, the underlying regex engine needs to be modified to handle this case somewhat properly, and from a cuDF perspective, this is not really defined behavior, so it wouldn't make any sense to fix it on that side.

revans2 (approved):
Still have a few nits that would be nice to have fixed, but they are not blockers.

NVnavkumar:
build

NVnavkumar:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRegExpReplaceMeta.scala
```diff
@@ -34,13 +34,29 @@ class GpuRegExpReplaceMeta(
...
+            pattern = pat
+            repl match {
+              case Some(rep) =>
+                GpuRegExpUtils.backrefConversion(rep) match {
+                  case (hasBackref, convertedRep) =>
+                    containsBackref = hasBackref
+                    replacement = Some(GpuRegExpUtils.unescapeReplaceString(convertedRep))
+                }
+              case _ =>
+            }
```
andygrove, line 59:
```suggestion
            repl.map(GpuRegExpUtils.backrefConversion).foreach {
                case (hasBackref, convertedRep) =>
                  containsBackref = hasBackref
                  replacement = Some(GpuRegExpUtils.unescapeReplaceString(convertedRep))
            }
```

andygrove, line 59:
```suggestion
            repl.map(GpuRegExpUtils.backrefConversion).foreach {
                case (hasBackref, convertedRep) =>
                  containsBackref = hasBackref
                  replacement = Some(GpuRegExpUtils.unescapeReplaceString(convertedRep))
            }
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -457,20 +531,49 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
+    // if we have a replacement, parse the replacement string using the regex parser to account
+    // for backrefs
+    val replacement = repl match {
+      case Some(s) => Some(new RegexParser(s).parseReplacement(countCaptureGroups(regex)))
+      case None => None
+    }
+    // repl match {
+    //   case Some(_) => System.out.println(replacement)
+    //   case _ =>
+    // }
```
andygrove, line 565: remove commented code

```suggestion
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -51,6 +51,31 @@ class RegexParser(pattern: String) {
...
+        case RegexSequence(parts) =>
+          sequence.parts ++= parts
+        case other =>
+          sequence.parts += other
+      }
+    }
+    sequence
+  }
+
+  def parseReplacementBase(until: () => Boolean): RegexAST = {
```
andygrove, line 67: remove unused param:

```suggestion
  def parseReplacementBase(): RegexAST = {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -51,6 +51,31 @@ class RegexParser(pattern: String) {
     ast
   }
 
+  def parseReplacement(numCaptureGroups: Int): RegexReplacement = {
+    val sequence = RegexReplacement(new ListBuffer(), numCaptureGroups)
+    while (!eof()) {
+      parseReplacementBase(() => eof()) match {
```
andygrove, line 57: remove unused param:

```suggestion
      parseReplacementBase() match {
```

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/stringFunctions.scala
```diff
@@ -853,7 +853,7 @@ class GpuRLikeMeta(
         case Literal(str: UTF8String, DataTypes.StringType) if str != null =>
           try {
             // verify that we support this regex and can transpile it to cuDF format
-            pattern = Some(new CudfRegexTranspiler(RegexFindMode).transpile(str.toString))
+            pattern = (new CudfRegexTranspiler(RegexFindMode).transpile(str.toString, None))._1
```
andygrove, line 856: nit: it isn't obvious when reading this code what `_1` represents - I will follow up with a separate PR to make this clearer

andygrove:
build

andygrove (approved):
LGTM overall. I will follow up with a PR to fix outstanding feedback from this PR.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -457,20 +531,45 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
+  }
+
   /**
    * Parse Java regular expression and translate into cuDF regular expression.
    *
    * @param pattern Regular expression that is valid in Java's engine
    * @return Regular expression in cuDF format
    */
-  def transpile(pattern: String): String = {
+  def transpile(pattern: String, repl: Option[String]): (Option[String], Option[String]) = {
```
andygrove, line 553: I don't think we need the first item in the tuple to be optional since we always return a value

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
andygrove: I added a test for this in the follow-up PR https://github.com/NVIDIA/spark-rapids/pull/5519

This case statement is guarded on the condition `mode == RegexReplaceMode` and the one below it is not so there is no conflict between the two. The `ch` being tested here is prior to the `$` so would not match the `ABC$A\n` case.

Issue #5294, opened 2022-04-21
Dist jar build should not need to unshim and dedup external dependencies


jlowe:
**Is your feature request related to a problem? Please describe.**
Currently external dependencies such as jucx and spark-rapids-jni are pulled in for every aggregator jar, and then the dist jar build ends up needing to ensure all the important parts of those jars are unshim'd and dedup'd to handle these dependencies properly.  This can cause problems when the aggregator jars pull in slightly different versions of those dependencies (e.g.: spark-rapids-jni snapshot changed during the build).

**Describe the solution you'd like**
We shouldn't pull in external dependencies into the aggregator jar unpack process.  Those should be handled separately.  Wondering if we can mark them as `<scope>provided</scope>` in the aggregator poms and directly depend on them in the `dist` pom, using the assembly plugin or something similar to bundle those dependencies in the dist jar directly rather than needing to manually ensure the parallel worlds build doesn't accidentally munge or needlessly duplicate their contents.

pxLi:
One potential issue of above solution could be that different spark shims could still use inconsistent jni libs to compile/build aggregator, but at last we just pick one version in dist pkg for all shims.

Probably we could specify a jni version at the beginning of the build, and use it for all shims building?

jlowe:
> Probably we could also specify a jni version at the beginning of the build, and use it for all shims building?

Agreed.  For any dependency, cudf or otherwise, we should try to avoid pulling in dynamic versions that can change across multiple builds like SNAPSHOT.

NvTimLiu:
Let me check this issue.

NvTimLiu:
mvn `-Dspark-rapids-jni.version=22.06.0-snap-id` seems not work well, as the e.g. `22.06.0-20220505.003730-22` will be transform to `22.06.0-SNAPSHOT` instead. Then mvn will download the latest jar instead of the one specified by  `-Dspark-rapids-jni.version=22.06.0-snap-id`

CLI of mvn tree

 **mvn -X  dependency:tree  -f sql_plugin**

```
[...]

PR #5501, opened 2022-05-16
Build against specified spark-rapids-jni snapshot jar [skip ci]

NvTimLiu:
part of https://github.com/NVIDIA/spark-rapids/issues/5294

Build rapids against the specified spark-rapids-jni snapshot jar, to avoid pulling different versions of
spark-rapids-jni dependency jars during the compile.

### DIFF
--- a/tests/src/test/scala/com/nvidia/spark/rapids/RapidsExecutorPluginSuite.scala
+++ b/tests/src/test/scala/com/nvidia/spark/rapids/RapidsExecutorPluginSuite.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, NVIDIA CORPORATION.
+ * Copyright (c) 2021-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


jenkins/spark-nightly-build.sh
```diff
@@ -22,6 +22,9 @@ set -ex
 ## export 'M2DIR' so that shims can get the correct Spark dependency info
 export M2DIR="$WORKSPACE/.m2"
 
+## maven options for building, e.g. '-Dspark-rapids-jni.version=xxx' to specify spark-rapids-jni dependency's version.
+MVN_OPT=${MVN_OPT:-''}
```
NvTimLiu, line 26: The build options need to be applied to every mvn cmd

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -313,11 +313,18 @@ object RapidsExecutorPlugin {
...
    * version 7.1.1.
+   * If the expected cudf version is a specified 'timestamp-seq' one, then it is satisfied by
+   * the SNAPSHOT version.
+   * For example, version 7.1-yyyymmdd.hhmmss-seq is satisfied by version 7.1-SNAPSHOT.
    */
   def cudfVersionSatisfied(expected: String, actual: String): Boolean = {
     val expHyphen = if (expected.indexOf('-') >= 0) expected.indexOf('-') else expected.length
     val actHyphen = if (actual.indexOf('-') >= 0) actual.indexOf('-') else actual.length
-    if (actual.substring(actHyphen) != expected.substring(expHyphen)) return false
+    if (actual.substring(actHyphen) != expected.substring(expHyphen) &&
```
NvTimLiu, line 323: After specify a special spark-rapids-jni snapshot version jar. e.g. `-Dspark-rapids-jni.version=22.06.0-20220420.050427-10`, we would got below build info from property files

In file `rapids4spark-version-info.properties`, jni version: cudf_version=`22.06.0-20220420.050427-10`

In file `spark-rapids-jni-version-info.properties`: version=`22.06.0-SNAPSHOT`

```
root@d03af8a1b975:/spark-rapids/dist/target/tmp# cat rapids4spark-version-info.properties
version=22.06.0-SNAPSHOT
cudf_version=22.06.0-20220420.050427-10
user=
revision=aa3dbab35832bb8a70b3daa13cbbdc1171825180
branch=branch-22.06
date=2022-05-10T07:20:48Z
url=https://github.com/NVIDIA/spark-rapids.git
root@d03af8a1b975:/spark-rapids/dist/target/tmp#
root@d03af8a1b975:/spark-rapids/dist/target/tmp# cat spark-rapids-jni-version-info.properties
version=22.06.0-SNAPSHOT
user=
revision=854d2dd1619ddcc8514c80081c656dff166ff4d7
branch=HEAD
date=2022-04-20T04:58:01Z
url=https://github.com/NVIDIA/spark-rapids-jni.git
root@d03af8a1b975:/spark-rapids/dist/target/tmp#

```
Though they refer to the same spark-rapids-jni jar dependency, running integration tests still failed with cudf expected/actual version MISMATCH

```

22/05/10 06:25:28.740 ScalaTest-main-running-MortgageSparkSuite INFO RapidsExecutorPlugin: RAPIDS Accelerator build: {version=22.06.0-SNAPSHOT, user=, url=https://github.com/NVIDIA/spark-rapids.git, date=2022-05-10T06:25:13Z, revision=aa3dbab35832bb8a70b3daa13cbbdc1171825180, cudf_version=22.06.0-20220420.050427-10, branch=branch-22.06}

22/05/10 06:25:28.741 ScalaTest-main-running-MortgageSparkSuite INFO RapidsExecutorPlugin: cudf build: {version=22.06.0-SNAPSHOT, user=, url=https://github.com/rapidsai/cudf.git, date=2022-04-20T04:58:01Z, revision=65b1cbdeda9cab57243d0a98e646c860ef86039e, branch=HEAD}
22/05/10 06:25:28.743 ScalaTest-main-running-MortgageSparkSuite ERROR RapidsExecutorPlugin: Exception in the exec 
```

jlowe (commented):
Allowing the spark-rapids-jni version to be manually anchored is nice, but this does not solve the issue described in #5294.  Even after this change, each Spark version will unpack and attempt to dedup the spark-rapids-jni jar and require manual sync of the unshim text files with the contents of that jar (and any other non-Spark-specific jar, like jucx).

Also shouldn't the `buildall` script be updated with similar changes?  The nightly build should not be the only build that has the fix.

NvTimLiu:
> 
Still dedup the spark-rapids-jni jar in aggregator spark-shims --> 

@jlowe  Sorry I may not understand this issue clearly before.
 
As aggregator depends on sql_plugin, and sql_plugin compile spark-rapids-jni into its jar.
 So do you mean we set scope <provided> for sql_plugin? 
Then add JNI<compile> scope in dist to avoid aggregate diff spark versions of shims dedupe spark-rapids-jni?

Will  update buildAll script, and follow the similar way for other non-Spark-specific jar, like jucx,

NvTimLiu:
> > sql_plugin compile spark-rapids-jni into its jar
> 
> sql_plugin needs spark-rapids-jni to compile, but it does _not_ pull spark-rapids-jni into the sql_plugin jar artifact. The spark-rapids-jni jar gets pulled into the dist jar by the custom dist jar build process (see dist/pom.xml and supporting build-parallel-worlds.xml and binary-dedupe.sh for details).
> 
> The problem is with the way non-Spark-specific dependencies are handled. They're essentially assumed to be Spark-specific dependencies, which requires them to go through the parallel-world, binary-dedupe, and unshim processing. Ideally we should only do this for dependencies that are Spark-specific (i.e.: the aggregator jars but not all their dependencies). One potential solution is to manually exclude the dependencies we know are not Spark specific (like spark-rapids-jni, jucx, possibly others) and treat them just like a normal dependency that is included into the dist artifact.

@jlowe  Thanks a lot for your explanation.

Yes the duplicated class files of non-Spark-specific dependencies are mainly spark-rapids-jni and jucx related. We can NOT shade them into Spark-specific aggregator jars to avoid multiple binary-dedupe.

As we would build sql_plugin multi times when building with diff spark shim versions, and spark-rapids-jni dependency jar may change during building, e.g., new spark-rapids-jni jars are pushed to maven repo.

 Suppose we can first patch for pulling the specified spark-rapids-jni dependency jar?  

Will follow up removing non-Spark-specific dependencies from aggregator jars in other pull requests. 
Also here is a issue, suppose we need to update our build scripts/pom.xml files, as there are some modules running tests against the aggregator jar. Simply remove non-Spark-specific dependencies from aggregator jars may causing tests FAILED.

https://github.com/NVIDIA/spark-rapids/blob/branch-22.06/tests/pom.xml#L56

https://github.com/NVIDIA/spark-rapids/blob/branch-22.06/api_validation/pom.xml#L148

jenkins/databricks/build.sh
```diff
@@ -20,9 +20,9 @@ set -ex
 SPARKSRCTGZ=$1
 # version of Apache Spark we are building against
 BASE_SPARK_VERSION=$2
-BUILD_PROFILES=$3
+MVN_OPT=$3
```
NvTimLiu, line 23: Additional options for mvn build on Databricks, e.g. MVN_OPT=`-Dspark-rapids-jnv.version=20220516.081920.30`

jlowe:
> Suppose we can first patch for pulling the specified spark-rapids-jni dependency jar?

Absolutely.  I just wanted to call out that this PR should not state that it resolves the issue, since it alone does not.

> Simply remove non-Spark-specific dependencies from aggregator jars may causing tests FAILED.

Yes, that is expected.  We cannot simply remove these dependencies, they are needed at compile time and runtime.  We need to process these dependencies differently when constructing the jars, not remove them entirely.

build/buildall
```diff
@@ -47,8 +47,12 @@ function print_usage() {
   echo "        Build in parallel, N (4 by default) is passed via -P to xargs"
   echo "   --install"
   echo "        Intall the resulting jar instead of just building it"
+  echo "   -o=MVN_OPT, --option=MVN_OPT"
+  echo "        use this option to build project with maven. E.g., --option='-Dcudf.version=cuda11'"
```
jlowe, line 51: Rather than needing to specify `-o` for every Maven option, why not simply assume that any argument encountered that isn't recognized as an option is the start of the arguments to pass through to Maven?

jenkins/spark-nightly-build.sh
jlowe: One way to help ensure this is to use a variable for the Maven command that includes the desired options, e.g.:
```shell
# use $MVN instead of "mvn" for every invocation of Maven
MVN="mvn ${MVN_OPT}"
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
jlowe: > Could you please help to check if this change what we expect fix for the issue https://github.com/NVIDIA/spark-rapids/issues/5294?

It does not per the prior discussion.  This is a good step towards the fix, but this does not address the deduplication and manual unshimming required for non-Spark dependencies.

> Maybe need we also to add some test cases in the file?

Yes, test cases should be added for the new behavior for SNAPSHOT matching.

build/buildall
NvTimLiu: Do you mean pass through args like `mvn ... $*`?  We have below block to mark `Unknow arg`,  
```
*)
  echo >&2 "Unknown arg: $1"
  print_usage
  exit 1
  ;;
```

`-o` seems more friendly than gathering all the unknow options as mvn opts.

and it can pass all options by `-o="-Dabc=1 -Pdef=2 ... `

NvTimLiu:
```
# use $MVN instead of "mvn" for every invocation of Maven
MVN="mvn ${MVN_OPT}"
```
> Good suggestion, updated

Yes, test cases should be added for the new behavior for SNAPSHOT matching.

> Updated

build/buildall
jlowe: Requiring the user to prefix every mvn option with `-o=` is annoying.  And as far as unknown arg, Maven will either process it as a useful argument or complain itself.  Given the whole point of this buildall script is to wrap invocations of Maven, what other use would there be for extra command-line arguments than pass them to Maven?

I'm not going to block the PR over this, but I find it odd saying it's "more friendly" for the user to jump through extra hoops here.

tests/src/test/scala/com/nvidia/spark/rapids/RapidsExecutorPluginSuite.scala
```diff
@@ -46,5 +46,7 @@ class RapidsExecutorPluginSuite extends FunSuite {
     assert(RapidsExecutorPlugin.cudfVersionSatisfied("7.0.1-special", "7.0.2-special"))
```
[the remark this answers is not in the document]
NvTimLiu: Good catch, fixed

jlowe (approved):

NvTimLiu:
build

NvTimLiu:
build

Issue #5325, opened 2022-04-27
[FEA] Support spark.sql.mapKeyDedupPolicy=LAST_WIN for `TransformKeys`


viadea:
I wish we can support spark.sql.mapKeyDedupPolicy=LAST_WIN to do deduplicate of the map keys.

For example:
```
from pyspark.sql.types import *
from pyspark.sql.functions import *
jsonString="""{"Zipcode":704,"ZipCodeType":"STANDARD","City":"PARC PARQUE","State":"PR","Zipcode":999}"""
df=spark.createDataFrame([(1, jsonString)],["id","value"])
df.show(truncate=False)
df2=df.withColumn("value",from_json(df.value,MapType(StringType(),StringType())))
df2.collect()

spark.conf.set("spark.sql.mapKeyDedupPolicy","LAST_WIN")
df2.withColumn('newcol', expr("""transform_keys(value, (k, v) -> split(k, "code")[0])""")).collect()
```

Not-supported-messages:
```
!Expression <TransformKeys> transform_keys(value#160, lambdafunction(split(lambda k#254, code, 2)[0], lambda k#254, lambda v#255, false)) cannot run on GPU because LAST_WIN is not supported for config setting spark.sql.mapKeyDedupPolicy
```

andygrove:
I updated the title to make it clear that this is asking for support for `LAST_WIN` for `transform_keys`. We actually already support `LAST_WIN` for `create_map` so we should have everything we need from cuDF.

For `create_map` we call `dropListDuplicatesWithKeysValues` to remove duplicates and this already supports the same semantics as `LAST_WIN`:

```
    // Apache Spark desires to keep the last duplicate element.
    auto [out_keys, out_vals] =
        cudf::lists::drop_list_duplicates(keys, vals, cudf::duplicate_keep_option::KEEP_LAST);
```

PR #5505, opened 2022-05-16
Support spark.sql.mapKeyDedupPolicy=LAST_WIN for TransformKeys

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/5325

Adds support for `spark.sql.mapKeyDedupPolicy=LAST_WIN` for `TransformKeys`.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/higherOrderFunctions.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/higherOrderFunctions.scala
@@ -508,7 +508,7 @@ case class GpuTransformKeys(
                   }
                 }
               }
-              GpuColumnVector.from(updatedMapView.copyToColumnVector(), dataType)
+              GpuColumnVector.from(deduped.incRefCount(), dataType)
             }
           }
         }
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/higherOrderFunctions.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/higherOrderFunctions.scala
@@ -474,7 +474,7 @@ case class GpuTransformKeys(
 
   override def prettyName: String = "transform_keys"
 
-  private val exceptionOnDupKeys = SQLConf.get.getConf(SQLConf.MAP_KEY_DEDUP_POLICY) ==
+  private def exceptionOnDupKeys = SQLConf.get.getConf(SQLConf.MAP_KEY_DEDUP_POLICY) ==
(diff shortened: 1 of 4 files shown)


andygrove:
build

nartal1 (dismissed):

andygrove:
Moving back to draft. I need to make sure `hasSideEffects` is implemented correctly as per https://github.com/NVIDIA/spark-rapids/pull/5546

revans2:
> Moving back to draft. I need to make sure `hasSideEffects` is implemented correctly as per #5546

Actually that is an existing bug in this. So you can fix it if you want here, but it does not have to be. This is a little more complex than #5546 because we can add an arbitrary new key. Before we could assume that there were no null keys, but here we cannot. So if you want to be exact the it should look something like.

```
  override lazy val hasSideEffects: Boolean = true
```

Because if a null is ever returned, then we have to throw an exception. A DataType in Spark itself has no indications of nullable vs not nullable. Only when it wrapped in another nested type do those start to show up. So there is no way for us to not know if the side effect is not possible with the information we have.

revans2:
Actually I am wrong. The expression itself is nullable. I feel dumb. It should be something like

```
 override lazy val hasSideEffects: Boolean =
    function.nullable || GpuCreateMap.exceptionOnDupKeys || super.hasSideEffects
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/higherOrderFunctions.scala
```diff
@@ -487,10 +494,23 @@ case class GpuTransformKeys(
...
+                // if there were duplicates.
+                withResource(deduped.getChildColumnView(0)) { a =>
+                  withResource(updatedMapView.getChildColumnView(0)) { b =>
+                    if (a.getRowCount != b.getRowCount) {
+                      throw GpuMapUtils.duplicateMapKeyFoundError
+                    }
+                  }
+                }
+              }
+              GpuColumnVector.from(updatedMapView.copyToColumnVector(), dataType)
```
revans2, line 511: Is there a reason we are using the original input and not the deduped input here? We use deduped when creating a map. This code should still work because I think the read picks the last matching key, but I am not 100% sure on that.

andygrove: I think this was an oversight. I will change this and see what other tests can be added to make show that we really are using LAST_WIN.

revans2 (approved):

andygrove:
build

Issue #5492, opened 2022-05-14
[BUG] com.nvidia.spark.rapids.RegexCharacterClass cannot be cast to com.nvidia.spark.rapids.RegexCharacterClassComponent


viadea:
When running below query, it failed 
```
: java.lang.ClassCastException: com.nvidia.spark.rapids.RegexCharacterClass cannot be cast to com.nvidia.spark.rapids.RegexCharacterClassComponent
```
Env:
22.06 snapshot jar

Minimum repro in pyspark:
```
from pyspark.sql.types import StructType,StructField, StringType, IntegerType
import pyspark.sql.functions as F
from pyspark.sql.functions import regexp_replace, col

data2 = [("a","","Smith","1111","M",2222),
    ("b","c","","3333","M",4444),
    ("d","","e","5555","M",6666),
    ("f","g","h","6666","F",4000),
    ("i","j","k","","F",-1)
  ]
 
schema = StructType([ \
    StructField("firstname",StringType(),True), \
    StructField("middlename",StringType(),True), \
    StructField("lastname",StringType(),True), \
    StructField("id", StringType(), True), \
    StructField("gender", StringType(), True), \
    StructField("salary", IntegerType(), True) \
  ])


df = spark.createDataFrame(data=data2,schema=schema)
df.printSchema()
df.show(truncate=False)


df.withColumn("firstname", regexp_replace(F.col("firstname"), '[^\sa-zA-Z0-9]', '')).collect()
```

viadea:
Full stacktrace is:
```
py4j.protocol.Py4JJavaError: An error occurred while calling o109.collectToPython.
: java.lang.ClassCastException: com.nvidia.spark.rapids.RegexCharacterClass cannot be cast to com.nvidia.spark.rapids.RegexCharacterClassComponent
	at com.nvidia.spark.rapids.CudfRegexTranspiler.$anonfun$rewrite$2(RegexParser.scala:706)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
	at scala.collection.immutable.List.foreach(List.scala:431)
	at scala.collection.generic.TraversableForwarder.foreach(TraversableForwarder.scala:38)
	at scala.collection.generic.TraversableForwarder.foreach$(TraversableForwarder.scala:38)
	at scala.collection.mutable.ListBuffer.foreach(ListBuffer.scala:47)
	at scala.collection.TraversableLike.map(TraversableLike.scala:286)
[...]

PR #5506, opened 2022-05-16
Fix ClassCastException in regular expression transpiler

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/5492

When transpiling a character class, the code assumed that each character would be transpiled to a valid character class component but in some cases, it could get transpiled to another character class and we cannot support that case. This PR makes us fall back to CPU instead of failing with a ClassCastException.


revans2 (dismissed):

andygrove:
build

andygrove:
build

andygrove:
@revans2 I dismissed your review before. Could you take another look?

revans2 (approved):

PR #5519, opened 2022-05-17
Address feedback in "string anchors regexp replace" PR

andygrove:
This is a follow-on PR to https://github.com/NVIDIA/spark-rapids/pull/5491 to address some review feedback items.

Changes in this PR:

- Change `transpile` return type so that first item in tuple is no longer an option
- Replace many but not all uses of `._1` when processing result of calling `transpile`
- Add an additional test


andygrove:
@revans2 This PR addresses some of the issues you raised in https://github.com/NVIDIA/spark-rapids/pull/5491

revans2 (approved):

andygrove:
build

PR #5522, opened 2022-05-17
Fix whitespace character class tests

anthony-chang:
- Previously, the python tests for whitespace `\s` and `\S` were not testing the intended code path. This PR fixes it by using double backslashes
- The fuzz tests caught a case we were not handling when a predefined character class is used immediately after a string anchor (eg. the pattern `a\Z\V`) which caused the transpilation to be incorrect. To fix this, I moved the rewriting of character classes `\s, \S, \v, \V, \h, \H` to the `rewrite()` stage and factored out the logic handle negated character classes

### DIFF
--- a/integration_tests/src/main/python/string_test.py
+++ b/integration_tests/src/main/python/string_test.py
@@ -853,7 +853,7 @@ def test_regexp_hexadecimal_digits():
 def test_regexp_whitespace():
     gen = mk_str_gen('\u001e[abcd]\t\n{1,3} [0-9]\n {1,3}\x0b\t[abcd]\r\f[0-9]{0,10}')
     assert_gpu_and_cpu_are_equal_collect(
-            lambda spark: unary_op_df(spark, gen).selectExpr('a',
+            lambda spark: unary_op_df(spark, gen).selectExpr(
                 'rlike(a, "\\\\s")',
                 'rlike(a, "\\\\s{3}")',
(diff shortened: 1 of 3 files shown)


anthony-chang:
build

integration_tests/src/main/python/string_test.py
```diff
@@ -854,16 +854,16 @@ def test_regexp_whitespace():
     gen = mk_str_gen('\u001e[abcd]\t\n{1,3} [0-9]\n {1,3}\x0b\t[abcd]\r\f[0-9]{0,10}')
     assert_gpu_and_cpu_are_equal_collect(
             lambda spark: unary_op_df(spark, gen).selectExpr(
-                'rlike(a, "\\s{2}")',
```
andygrove, line 857: In this line, the qualifier `{2}` is also removed. Is this intentional?

sameerz:
I think there are two separate changes in this PR, one for tests (related to whitespace) and one for a bug ( not handling when a predefined character class is used immediately after a string anchor ).  The changes ought to be in separate PRs.

revans2 (dismissed):

integration_tests/src/main/python/string_test.py
andygrove: I guess we have the repetition case covered in the next line, so this is fine

andygrove (approved):

anthony-chang:
Will split into two separate PRs

Issue #5511, opened 2022-05-17
[BUG] orc_test failures on 321cdh


tgravescs:
**Describe the bug**

orc_test.py (6 failed, 782 passed, 10 skipped)

`[2022-05-13T03:09:02.816Z] ../../src/main/python/orc_test.py::test_orc_forced_position[native-true-{'spark.rapids.sql.format.orc.reader.type': 'PERFILE'}] 22/05/13 03:09:03 WARN scheduler.TaskSetManager: Lost task 0.0 in stage 1040.0 (TID 2285) (rl-r7525-d32-u05.raplab.nvidia.com executor 2): ai.rapids.cudf.CudfException: cuDF failure at: /home/jenkins/agent/workspace/jenkins-spark-rapids-jni_nightly-dev-github-106-cuda11/thirdparty/cudf/cpp/src/io/orc/reader_impl.cu:1042: Expected streams data within stripe`

tgravescs:
this passed on 22.02, we did make changes so orc used the version with the spark distribution, not sure if its related at all, just need to investigate further.

PR #5523, opened 2022-05-18
Enable forcePositionEvolution for 321cdh

firestarman:
Get the value of `forcePositionEvolution` from the configuration instead of the hardcode `false` for 321cdh shim, since 321cdh also supports `forcePositionEvolution` and GPU should follow its behavior.

The test failed because when `forcePositionEvolution` is set to `true`, GPU still returns `false`, then CPU and GPU run into different pathes in method `requestedColumnIds`, producing different output.

fixes https://github.com/NVIDIA/spark-rapids/issues/5511


firestarman:
build

jlowe (approved):

PR #5527, opened 2022-05-18
Fix bug with character class immediately following a string anchor

anthony-chang:
The fuzz tests caught a case we were not handling when a predefined character class is used immediately after a string anchor (eg. the pattern `a\Z\V`) which caused the transpilation to be incorrect. To fix this, I moved the rewriting of character classes `\s, \S, \v, \V, \h, \H` to the rewrite() stage and factored out the logic handle negated character classes

(This PR contains the second half of the changes from #5522)


anthony-chang:
build

PR #5532, opened 2022-05-19
Support hex digits in character classes and escaped characters in character class ranges

### LINKED ISSUE #4505: [BUG] RegExp parse fails to parse character ranges containing escaped characters
**Describe the bug**
The regular expression `[\a-z]` does not get parsed correctly because we currently only support character ranges with non-escaped characters. The transpiler then converts this to `[\a\-z]` when passing it to cuDF, resulting in incorrect results.

This was found by the new test code in https://github.com/NVIDIA/spark-rapids/pull/4504

**Expected behavior**
This expression should produce the same results on CPU and GPU.

**Additional context**
None

### LINKED ISSUE #4518: [FEA] Add support for escaped unicode hex in regular expressions
We currently fall back to CPU for regular expressions that include escaped hex digits in the format such as `\\u0061` and fail with the error `invalid or unsupported escape character 'u' near index 0`. It would be better to fail consistently with current hex support (`\\x0061` would fail with `cuDF does not support hex digits consistently with Spark`). It would be even better if we can support this on GPU once https://github.com/NVIDIA/spark-rapids/issues/4486 is resolved.

See above.

None

**Additional context**
None

### LINKED ISSUE #4865: [BUG] Add support for regular expressions containing hexadecimal digits inside character classes, eg `[\x7f]`
**Describe the bug**
We support hexadecimal digits but with limitations. Right now, hexadecimal digits cannot be used inside character classes on the GPU.

Regular expressions with hexadecimal digits inside character classes do not work. Examples include `[\x02]`, `[\x7f]`, `[\x77]`, etc.

**Expected behavior**
We should support hexadecimal digits inside character classes.

anthony-chang:
Fixes #4865, fixes #4505, closes #4518

This PR allows hexadecimal digits to be used in character classes and ranges in character classes. To support this, I have also generalized this to support all valid escaped characters in character class ranges. 

A few notes:
- I define a list of supported metacharacters in a character class: `\, ^, -, ], +`. These are the characters that may be escaped in a character class
- In addition to the above list, I define `\n, \r, \t, \f, \a, \b, \e` as all the literals that use a backslash (https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html under "Characters")
- When a pattern uses the hexadecimal representation of a metacharacter in a character class (eg. using `\x5e` instead of `^`), Java treats this as an escaped metacharacter. Because of this, we cannot directly transpile hex digits to their unicode representation--we need to first check if it is a metacharacter. 
Example: `[\x5ea]` should transpile to `[\^a]`


anthony-chang:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -165,6 +165,41 @@ class RegexParser(pattern: String) {
   }
 
   private def parseCharacterClass(): RegexCharacterClass = {
+    val supportedMetaCharacters = "\\^-]+"
```
andygrove, line 168: :+1:

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -183,6 +173,20 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
     assertCpuGpuMatchesRegexpFind(patterns, Seq("", "\u0007", "a\u0007b", 
         "\u0007\u003f\u007f", "\u0080", "a\u00fe\u00ffb", "ab\ueeeecd"))
   }
+
+  test("hex digit character classes") {
+    val patterns = Seq(raw"[\x02]", raw"[\x2c]", raw"[\x7f]", raw"[\x80]", raw"[\x01-\xff]",
+      raw"[a-\xff]", raw"[\x20-z]")
+    val inputs = Seq("", "\u007f", "a\u007fb", "\u007f\u003f\u007f", "\u0080", "a\u00fe\u00ffb", 
+      "\u007f2", "abcd", "\u0000\u007f\u00ff\u0123\u0abc")
+    assertCpuGpuMatchesRegexpFind(patterns, inputs)
```
andygrove, line 182: can we also test with replace

andygrove (approved):
LGTM but would also be good to add tests for "replace" operations

anthony-chang:
build

andygrove (approved):

andygrove (approved):

Issue #3682, opened 2021-09-27
[FEA] Use conventional jar layout in dist jar if there is only one input shim


gerashegalov:
**Is your feature request related to a problem? Please describe.**
In a targeted cloud / vendor environment there is typically no need to have multi-spark-version jar artifact allowing to switch between Spark versions back and forth. 

We expect that the dist jar for such a targeted environment is likely to be built using the `individual` profile. In this case it's unnecessary to use non-standard Parallel World  jar layout https://github.com/NVIDIA/spark-rapids/blob/a5963e5a7ddf2b6db976e83ee8654d59e0c5f1eb/dist/pom.xml#L1674

**Describe the solution you'd like**
We need an option for dist jar based on a single `individual` buildver classified output to use  the standard jar layout without `/spark3*` directories at the root.

**Describe alternatives you've considered**
keep using PW layout even for single shim builds which elevates the risk of classloading issues

**Additional context**
#3381

revans2:
I personally disagree with this for exactly the same reasons pointed out here.

> keep using PW layout even for single shim builds which **elevates the risk of classloading issues**

If what we ship and expect most people to use is the uber/parallel world jar, then I want everyone to be testing that situation. I personally build and test regularly for a single version because I would rather wait 3 mins to build a single version than 10 mins for buildall to finish. But if I am testing things in a very different way I don't want that.  If someone has a targeted cloud environment where they are building, testing, and deploying for a single version of Spark, then they should make any changes needed to the build scripts. The common code should stay as is.

gerashegalov:
I think the intent of my issue and the wording diverged in that it sounds generally applicable to `individual` profile.  Instead it should read Parallel World layout should be opt-out. Note that this should not incur source code modification, just a packaging option.

I too advocate that we should definitely do most of our testing with the muti-shim jar, preferably minimumFeatureVersionMix profile. 

What I argue for vendor environments has little to do with the multi-shim build overhead, What we have there is an environment we have no control over, whereas we are pushing a jar that provides a solution that this type of environment most often doesn't have. E.g., if we deploy to Databricks for a specific runtime version known apriori there is absolutely no need for a complex mult-shim paralllel-world layout. 

Therefore, I think having an option to turn off non-standard jar layout creation at build time when needed is reasonable. Again, I agree that it should not be the default behavior.

revans2:
[...]

PR #5537, opened 2022-05-19
[WIP] Auto-detect single shim jar

gerashegalov:
This PR contributes to #3682.  Whenever the rapids-4-spark dist jar contains only one SparkShimServiceProvider, and it is loadable from the root jar path and not in a /spark3xx parallel world, parallel world classloading is disabled.

TODO: provide a build flag that would simply copy an aggregator jar as a dist jar instead of building parallel worlds.


gerashegalov:
closed in favor of #5528

Issue #5507, opened 2022-05-16
[FEA] Support GpuRaiseError


abellina:
This PR https://github.com/NVIDIA/spark-rapids/pull/5487 adds the ability to convert a UDF that can throw `SparkException` into a catalyst expression with `RaiseError`.

We'd like to have `GpuRaiseError` so that we can prevent a columnar-to-row transition when an error should be raised.

abellina:
For this issue we would need to be careful not to copy to host a huge output since `RaiseError` just takes an `Expression`, as it could evaluate to a column. Since we are throwing no matter what, we'd want to first slice this column to 1 row and bring that top result only.

PR #5540, opened 2022-05-19
Support RaiseError [databricks]

wbo4958:
This PR adds GpuRaiseError to replace RaiseError Expression.  It is to fix https://github.com/NVIDIA/spark-rapids/issues/5507.

### DIFF
--- a/integration_tests/src/main/python/misc_expr_test.py
+++ b/integration_tests/src/main/python/misc_expr_test.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2021, NVIDIA CORPORATION.
+# Copyright (c) 2020-2022, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,7 +14,7 @@
 
 import pytest
 
-from asserts import assert_gpu_and_cpu_are_equal_collect
+from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_and_cpu_error
 from data_gen import *
 from marks import incompat, approximate_float
 from pyspark.sql.types import *
@@ -31,3 +31,10 @@ def test_part_id():
             lambda spark : unary_op_df(spark, short_gen, num_slices=8).select(
                 f.col('a'),
                 f.spark_partition_id()))
+
+def test_raise_error():
+    assert_gpu_and_cpu_error(
+        lambda spark : unary_op_df(spark, short_gen, num_slices=8).select(
+                f.raise_error(f.col('a'))).collect(),
+        conf={},
+        error_message="java.lang.RuntimeException")
--- a/integration_tests/src/main/python/misc_expr_test.py
+++ b/integration_tests/src/main/python/misc_expr_test.py
@@ -35,11 +35,17 @@ def test_part_id():
 def test_raise_error():
     data_gen = ShortGen(nullable=False, min_val=0, max_val=20, special_cases=[])
     assert_gpu_and_cpu_are_equal_collect(
-        lambda spark : unary_op_df(spark, data_gen, num_slices=2).select(
+        lambda spark: unary_op_df(spark, data_gen, num_slices=2).select(
             f.when(f.col('a') > 30, f.raise_error("unexpected"))))
 
     assert_gpu_and_cpu_error(
-        lambda spark : unary_op_df(spark, short_gen, num_slices=2).select(
+        lambda spark: unary_op_df(spark, null_gen, length=2, num_slices=1).select(
                 f.raise_error(f.col('a'))).collect(),
         conf={},
         error_message="java.lang.RuntimeException")
+
+    assert_gpu_and_cpu_error(
+        lambda spark: unary_op_df(spark, short_gen, length=2, num_slices=1).select(
+                f.raise_error(f.lit("unexpected"))).collect(),
+        conf={},
+        error_message="java.lang.RuntimeException: unexpected")
--- a/integration_tests/src/main/python/misc_expr_test.py
+++ b/integration_tests/src/main/python/misc_expr_test.py
@@ -38,6 +38,9 @@ def test_raise_error():
         lambda spark: unary_op_df(spark, data_gen, num_slices=2).select(
             f.when(f.col('a') > 30, f.raise_error("unexpected"))))
 
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: spark.range(0).select(f.raise_error(f.col("id"))))
+
     assert_gpu_and_cpu_error(
         lambda spark: unary_op_df(spark, null_gen, length=2, num_slices=1).select(
                 f.raise_error(f.col('a'))).collect(),
--- a/integration_tests/src/main/python/misc_expr_test.py
+++ b/integration_tests/src/main/python/misc_expr_test.py
@@ -33,8 +33,13 @@ def test_part_id():
                 f.spark_partition_id()))
 
 def test_raise_error():
+    data_gen = ShortGen(nullable=False, min_val=0, max_val=20, special_cases=[])
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark : unary_op_df(spark, data_gen, num_slices=2).select(
+            f.when(f.col('a') > 30, f.raise_error("unexpected"))))
+
     assert_gpu_and_cpu_error(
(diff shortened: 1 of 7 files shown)


sql-plugin/src/main/scala/org/apache/spark/sql/rapids/misc.scala
```diff
@@ -0,0 +1,44 @@
...
+  /** Could evaluating this expression cause side-effects, such as throwing an exception? */
+  override def hasSideEffects: Boolean = true
+
+  override protected def doColumnar(input: GpuColumnVector): ColumnVector = {
+    if (input == null || input.getRowCount <= 0) {
+      throw new RuntimeException()
+    }
+
+    // Take the first one as the error message
+    val msg = input.copyToHost().getUTF8String(0).toString
```
wbo4958, line 40: Is there any necessary to only copy the first row to get the error msg instead of copying the whole column vector?

wbo4958:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/misc.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  override def dataType: DataType = NullType
+  override def inputTypes: Seq[AbstractDataType] = Seq(StringType)
+  override def toString: String = s"raise_error($child)"
+
+  /** Could evaluating this expression cause side-effects, such as throwing an exception? */
+  override def hasSideEffects: Boolean = true
+
+  override protected def doColumnar(input: GpuColumnVector): ColumnVector = {
+    if (input == null || input.getRowCount <= 0) {
```
revans2, line 35: `input` should never be null. If it is that is an internal error. I am okay with what you are doing but it would be good to know something unexpected happened.

If input has no rows in it, then I don't want to throw an exception. Just return an empty ColumnVector.  This I can see actually happening if you have an IF/ELSE to check for error cases. I don't know if there are any corner cases when nothing matched and we got an empty ColumnVector, but I can see it happening.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/misc.scala
revans2: This is on an error case. The entire job is going to fail. I am not too concerned with failing faster than the CPU. Yes it would be nice to not copy everything. You can do that with `getScalarElement`, which should keep the code small and clean.

```
withResource(input.getScalarElement(0)) { scalarMsg =>
  if (!scalarMsg.isValid()) {
    throw new RuntimeException()
  } else {
    throw new RuntimeException(scalarMsg.getJavaString())
  }
}
```

wbo4958: Good suggestion. Done.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/misc.scala
wbo4958: Yeah, I just changed this and added the related tests. Thx

wbo4958:
build

wbo4958:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/misc.scala
```diff
@@ -0,0 +1,51 @@
...
+  override def dataType: DataType = NullType
+  override def inputTypes: Seq[AbstractDataType] = Seq(StringType)
+  override def toString: String = s"raise_error($child)"
+
+  /** Could evaluating this expression cause side-effects, such as throwing an exception? */
+  override def hasSideEffects: Boolean = true
+
+  override protected def doColumnar(input: GpuColumnVector): ColumnVector = {
+    if (input.getRowCount <= 0) {
+      // For the case: when(condition, raise_error())
```
abellina, line 37: it would be nice to cover this specific case in the python tests, it doesn't seem like we are.

abellina: Actually, according to spark, I don't think this `raise_error()` (no args) is possible:
```
pyspark.sql.utils.AnalysisException: Invalid number of arguments for function raise_error. Expected: 1; Found: 0; line 1 pos 7
```

abellina (approved):
LGTM

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3477,7 +3477,15 @@ object GpuOverrides extends Logging {
         TypeSig.ARRAY.nested(TypeSig.commonCudfTypesWithNested),
         TypeSig.ARRAY.nested(TypeSig.all)),
       (e, conf, p, r) => new GpuGetArrayStructFieldsMeta(e, conf, p, r)
-    )
+    ),
+    expr[RaiseError](
+      "throw exception",
```
jlowe, line 3482: Nit: All other descriptions start with a capital letter and are a bit more descriptive, as seen in the generated configs.md docs.
```suggestion
      "Throws an exception",
```

integration_tests/src/main/python/misc_expr_test.py
```diff
@@ -31,3 +31,15 @@ def test_part_id():
...
+    data_gen = ShortGen(nullable=False, min_val=0, max_val=20, special_cases=[])
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark : unary_op_df(spark, data_gen, num_slices=2).select(
+            f.when(f.col('a') > 30, f.raise_error("unexpected"))))
+
+    assert_gpu_and_cpu_error(
+        lambda spark : unary_op_df(spark, short_gen, num_slices=2).select(
+                f.raise_error(f.col('a'))).collect(),
+        conf={},
+        error_message="java.lang.RuntimeException")
```
jlowe, line 45: The test should verify we are properly conveying the specified error message into the exception rather than just checking for the same exception type, checking both the null first element and non-null first element scenarios.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/misc.scala
jlowe: This is possible and should be tested, e.g.:
```
>>> import pyspark.sql.functions as f
>>> df = spark.range(0)
>>> df.count()
0
>>> df.select(f.raise_error(f.col("id"))).explain()
== Physical Plan ==
*(1) Project [raise_error(cast(id#12L as string), NullType) AS raise_error(id)#20]
+- *(1) Range (0, 0, step=1, splits=12)


>>> df.select(f.raise_error(f.col("id"))).collect()
[]
```

wbo4958:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
wbo4958: Thx, Done

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/misc.scala
wbo4958: Yeah, raise_error needs to accept the parameter and I just updated the comment.


> Actually, according to spark, I don't think this `raise_error()` (no args) is possible:
> 
> ```
> pyspark.sql.utils.AnalysisException: Invalid number of arguments for function raise_error. Expected: 1; Found: 0; line 1 pos 7
> ```

Yeah, raise_error needs to accept the parameter and I just updated the comment.

wbo4958: > This is possible and should be tested, e.g.:
> 
> ```
> >>> import pyspark.sql.functions as f
> >>> df = spark.range(0)
> >>> df.count()
> 0
> >>> df.select(f.raise_error(f.col("id"))).explain()
> == Physical Plan ==
> *(1) Project [raise_error(cast(id#12L as string), NullType) AS raise_error(id)#20]
> +- *(1) Range (0, 0, step=1, splits=12)
> 
> 
> >>> df.select(f.raise_error(f.col("id"))).collect()
> []
> ```

Done

integration_tests/src/main/python/misc_expr_test.py
wbo4958: Done

wbo4958:
build

revans2 (approved):

abellina (approved):

PR #5546, opened 2022-05-19
Set hasSideEffects correctly for GpuCreateMap

### LINKED ISSUE #5544: [BUG] GpuCreateMap needs to set hasSideEffects in some cases
**Describe the bug**
https://github.com/NVIDIA/spark-rapids/pull/5533#discussion_r877246723

if 
```
(SQLConf.get.getConf(SQLConf.MAP_KEY_DEDUP_POLICY) ==
        SQLConf.MapKeyDedupPolicy.EXCEPTION.toString
```

Then `GpuCreateMap` can throw an exception, which means it has side effects so we need to set that properly. The above is true by default.

just a quick guess but 

```
spark.range(10).selectExpr("id", "if(id == 0, null, map(id, id, id DIV 2, id)) as m").show()
```

**Expected behavior**
No exception is thrown when creating a map under an `if` statement with duplicate keys when the `if` statement would have filtered out that operation.

revans2:
This fixes #5544

### DIFF
--- a/integration_tests/src/main/python/map_test.py
+++ b/integration_tests/src/main/python/map_test.py
@@ -135,6 +135,12 @@ def test_get_map_value_timestamp_keys(data_gen):
             'a[timestamp "2022-01-01"]',
             'a[null]'))
 
+def test_map_side_effects():
+    assert_gpu_and_cpu_are_equal_collect(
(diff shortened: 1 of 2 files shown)


revans2:
build

jlowe (approved):

Issue #5512, opened 2022-05-17
[FEA] Qualification tool, hook up final output and output execs table


tgravescs:
**Is your feature request related to a problem? Please describe.**
we have all the pieces in place, hook up the final output and change to new proposed fields. Remove unneeded ones.

Also output the execs information into csv file.

Issue #5364, opened 2022-04-28
Qualification tool figure out duration calculation when cpu plan doesn't contain duration for each op


tgravescs:
**Is your feature request related to a problem? Please describe.**
So likely the plan parser doesn't have close to the right duration because not all operators have a duration.
So we can try to estimate based on the SQL task times, which is really estimated  with the stages task times added up. Also here we should probably look for overhead between the jobs and add that in.

To do this, we line up the SQLID to stages and then which operators were in each Stage based on the accumulators. Note not all operators have a mapping to stages.  Then if the stage task duration > the duration from parse plan then take difference * (speedup factor averaged for execs without duration)

Issue #3792, opened 2021-10-11
Qualification tool improvements


tgravescs:
**Is your feature request related to a problem? Please describe.**
We want to improve the qualification output to give user higher level of confidence a query is a good candidate for GPU.  In this case we only want to rely on the eventlog initially as anything else like explain API is something else the user has to run.  I don't really know how to improve it more without parsing the specific SQL operators and either see if we think they would run on the GPU and/or use some heuristics about the metrics and operator types.

We can try to provide more stats like:
- was there a particular stage in the SQL query that was the largest.  If so can we determine the sql operators in that stage (we can only sometimes do this from event log). Of those operators which one took longest.  Is that operator something we might do on the GPU.  Challenge here is we do not know the types of the input to all the operators. We have some of the schema from file input.  If do this definitely need asterisk by data types saying make sure we support it.
- Put together input/output sizes and shuffle sizes to see if its too small data
- skew data checks
- Look at particular operators we know generally do well - window operator, transcoding.  I don't think there is a way for us to decide if its High cardinality processing​ (joins, sorts, aggregations)

We want to put as much information we can into the ranking of the applications and then also be able to display what information we used to the user so they can understand why we chose that ranking.

We could also analyze the NDS queries in detail and how the qualification tool scores them to see if we can find other heuristics to add to make the qualification tool score better.  Including things we could reports for potential problems, etc.

Can we determine env to see if possible I/O bottle necks?  Like could we have them fill in information about network/disks, etc... not sure on this one.

Also investigate adjust the scores based on failed tasks/stages/etc..

Can we be more fine grain for RDD/Dataset that still have some dataframe operations. Right now we exclude the entire SQL ops when we see that, perhaps we can determine which parts better.

Other issues that go along with this.

- [x] https://github.com/NVIDIA/spark-rapids/issues/5176
- [x] https://github.com/NVIDIA/spark-rapids/issues/5263
- [x] https://github.com/NVIDIA/spark-rapids/issues/5269
- [x] https://github.com/NVIDIA/spark-rapids/issues/5358
- [x] https://github.com/NVIDIA/spark-rapids/issues/5361
- [x] https://github.com/NVIDIA/spark-rapids/issues/5362
- [x] https://github.com/NVIDIA/spark-rapids/issues/5363
- [x] https://github.com/NVIDIA/spark-rapids/issues/5364
- [x] https://github.com/NVIDIA/spark-rapids/issues/5512
[...]

PR #5550, opened 2022-05-19
Qual stage metrics

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/5512 and fixes https://github.com/NVIDIA/spark-rapids/issues/5364

Lots of changes here.  
So we can try to estimate based on the SQL task times, which is really estimated with the stages task times added up, and we use job overhead To do this, we line up the SQLID to stages and then which operators were in each Stage based on the accumulators. Note not all operators have a mapping to stages. 
We take the task time calculations and get a ratio and then try to apply them to the wall clock times to display to the user.

change some of the UDF based operators to have score of 1.2 as we only accelerate the data transfer and udfs still run on cpu.

Each exec checks for udfs and dataset individually to say those aren't supported vs doing the entire SQL operation.

We have 4 levels of recommendation, strongly, recommended, not and not applicable. not applicable is when we have stage or job failures.  Need to followup and update UI with good way to display this.

Updated the output to be what we want. Focus on wall clock for the user readable side, the csv file has the task level details.  I think we will want to put more in the csv file but for now this works.  Outputs the exec level in csv file and outputs the stages used in SQL operations in a csv file.

Need to do more testing on this but would like to get this in and be able to get others to test. A few tests were commented out to update later.

There is also more code cleanup and refactoring I would like to do to make this more readable but that can wait.


tgravescs:
build

tools/src/main/scala/com/nvidia/spark/rapids/tool/planparser/SQLPlanParser.scala
```diff
@@ -23,19 +23,19 @@ import com.nvidia.spark.rapids.tool.qualification.PluginTypeChecker
...
-    exec: String,
-    expr: String,
-    speedupFactor: Double,
-    duration: Option[Long],
-    nodeId: Long,
-    isSupported: Boolean,
-    children: Option[Seq[ExecInfo]], // only one level deep
-    stages: Seq[Int] = Seq.empty,
-    shouldRemove: Boolean = false) {
+class ExecInfo(
```
nartal1, line 28: Just for my understanding - Why was this changed to `class` from `case class` ?

tgravescs: Spark gave circular dependency error  when used in Spark itself on it since it uses ExecInfo for children. No problems when run with the tool

amahussein (approved):
I checked the UI works fine with this PR.
I approve this PR as long as we can followup with PRs addressing the issues listed in #3792

tgravescs:
build

nartal1 (approved):
LGTM.  The calculation of speedup can be improved in follow on PR's.

tgravescs:
build

tgravescs:
not sure why a test was hanging due to this, but fixed.

tgravescs:
build

nartal1 (approved):

Issue #5305, opened 2022-04-25
[FEA] Qualification tool: Operator mapping, check if execs/expressions off by default


nartal1:
**Is your feature request related to a problem? Please describe.**
This is follow on from the discussion https://github.com/NVIDIA/spark-rapids/pull/5298#discussion_r856214693 . We need to figure out a good way to get the configs that are set for certain execs.

nartal1:
Synced with @tgravescs on this. We are checking expressions and Execs which are OFF by default in GpuOverrides from the `disabledMsg` . Other than these, we don't see any important configs OFF by default in https://nvidia.github.io/spark-rapids/docs/configs.html .  
We will parse more information from the eventlog and update individual Execs parser in https://github.com/NVIDIA/spark-rapids/issues/6064. i.e for BHJ, we will use the tagging logic to check if a certain `joinType` is supported.
Closing this issue.

PR #5551, opened 2022-05-20
Qualification tool: Add support to json and avro formats

nartal1:
This contributes to https://github.com/NVIDIA/spark-rapids/issues/5305 .

JSON and AVRO file formats are mentioned as "Configured Off" by default in the qualification tool. At present, while calculating the speedup factor, it's always `0` as the file format's score is calculated as `0`. In this PR, we are assising the score so that speed up factor is non zero which in turn helps in improving the qualification of applications.


nartal1:
build

PR #5571, opened 2022-05-20
Update array_min and array_max to use new cudf operations

revans2:
CUDF finished work on segmented min and segmented max for strings and decimal. This lets us delete a lot of code and hopefully thing run even faster now too.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
@@ -395,56 +395,6 @@ case class GpuSortArray(base: Expression, ascendingOrder: Expression)
   }
 }
 
-// TODO switch over to array aggregations once
-//  https://github.com/rapidsai/cudf/issues/10417 is done
-object SlowGpuArrayAgg extends Arm{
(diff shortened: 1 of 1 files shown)


revans2:
build

jlowe (commented):
Nice!  Has unused imports that failed the verify build, but otherwise lgtm.

revans2:
build

revans2:
Our build system is too complicated. I do a build/buildall but it does not update the docs or warn me about style. So I have to do a buildall and a regular build for a single version to get everything I need to check before pushing something. This is really frustrating.

jlowe (approved):

gerashegalov (approved):
LGTM

gerashegalov:
> Our build system is too complicated. I do a build/buildall but it does not update the docs or warn me about style. So I have to do a buildall and a regular build for a single version to get everything I need to check before pushing something. This is really frustrating.

One can obviously add whatever is missing to buildall . It does scalastyle on the lowest shim involved.  

At least now we receive this automated feedback in a couple of minutes instead of waiting for hours:  htps://github.com/NVIDIA/spark-rapids/runs/6528205143?check_suite_focus=true

revans2:
@gerashegalov yes the few mins is so very nice compared to before.

PR #5578, opened 2022-05-20
Fix unfiltered partitions being used to create GpuBatchScanExec RDD

jlowe:
Fixes a typo in the GpuBatchScanExec for Spark 3.2.x where the unfiltered partitions were used to create the RDD rather than the filtered partitions.  This had the effect of disabling runtime filtering for scans under GpuBatchScanExec.


jlowe:
build

jlowe:
build

tgravescs (approved):

PR #5591, opened 2022-05-23
WIP: Enable regular expressions by default

andygrove:

Part of https://github.com/NVIDIA/spark-rapids/issues/4509

We want to enable regular expressions by default in 22.06. This does not mean we are guaranteeing 100% compatibility with Spark though so we need to document the edge cases that can produce different results.

- [x] Change configuration default to `true`
- [ ] Document all known edge cases in the compatibility guide


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1037,12 +1037,12 @@ object RapidsConf {
     .createWithDefault(true)
 
   val ENABLE_REGEXP = conf("spark.rapids.sql.regexp.enabled")
-    .doc("Specifies whether regular expressions should be evaluated on GPU. Complex expressions " +
-      "can cause out of memory issues so this is disabled by default. Setting this config to " +
-      "true will make supported regular expressions run on the GPU. See the compatibility " +
-      "guide for more information about which regular expressions are supported on the GPU.")
+    .doc("Specifies whether supported regular expressions should be evaluated on GPU. There " +
+      "are some known edge cases that can produce incorrect results and these are documented in " +
```
gerashegalov, line 1041: should this say that known edge cases are executed on CPU, at least this is the impression I have from reading compatibility guide.

gerashegalov, line 1041: should this say that known edge cases are executed on CPU, at least this is the impression I have from reading compatibility guide.

sameerz:
I agree with @gerashegalov 's comments, and am ok with the rest of the text wording.

andygrove:
I had been holding off updating this in the hope that https://github.com/NVIDIA/spark-rapids/pull/5610 and https://github.com/NVIDIA/spark-rapids/pull/5662 would be merged and then we would not need to list any known issues in this PR. However, I am blocked from merging those at the moment due to the need to have CI specify `LANG=en_US.UTF-8` before invoking the Scala unit tests so I will update this PR listing the issues that would have been solved by the other PRs so we can get this merged,

andygrove:
build

gerashegalov (approved):

andygrove:
build

PR #5603, opened 2022-05-24
Skip unshim and dedup of external spark-rapids-jni and jucx [databricks]

### LINKED ISSUE #5294: Dist jar build should not need to unshim and dedup external dependencies

res-life:
Closes #5294 

**Changes:**
- Stop pulling external dependencies such as `jucx` and `spark-rapids-jni` into every aggregator jar by marking them as `provided`  dependency scope. Now the aggregator jar is less than 10M.
```
ls -lh aggregator/target/spark330/
... 7.2M  ... rapids-4-spark-aggregator_2.12-22.06.0-SNAPSHOT.jar
```

- Copy one and only one version of them into `dist` jar after the deduping aggregator jar is done.

**Note:**
- Related PR: #5501
This PR supports building against specified spark-rapids-jni snapshot jar by specify `'-Dspark-rapids-jni.version=xx'`
- If depends on spark-rapids-jni snapshot version, the building may download multiple spark-rapids-jni snapthot jars for different Shims, but in the end, the packaging only picks one version of them. This is the potential risk, but I think we accept this risk for the snapshot build.

### DIFF
--- a/aggregator/pom.xml
+++ b/aggregator/pom.xml
@@ -55,18 +55,6 @@
             <version>${project.version}</version>
             <classifier>${spark.version.classifier}</classifier>
         </dependency>
-        <!-- set provided scope for JNI and ucx to avoid pulling into aggregator jar  -->
-        <dependency>
-            <groupId>org.openucx</groupId>
-            <artifactId>jucx</artifactId>
-            <scope>provided</scope>
-        </dependency>
-        <dependency>
-            <groupId>com.nvidia</groupId>
-            <artifactId>spark-rapids-jni</artifactId>
-            <classifier>${cuda.version}</classifier>
-            <scope>provided</scope>
-        </dependency>
     </dependencies>
     <build>
         <plugins>
@@ -92,7 +80,12 @@
                 <artifactId>maven-shade-plugin</artifactId>
                 <configuration>
                     <artifactSet>
-                        <excludes>org.slf4j:*</excludes>
+                        <excludes>
+                            <exclude>org.slf4j:*</exclude>
+                            <!-- excludes spark-rapids-jni and jucx, will package them in the dist submodule -->
+                            <exclude>com.nvidia:spark-rapids-jni:*</exclude>
+                            <exclude>org.openucx:jucx:*</exclude>
+                        </excludes>
                     </artifactSet>
                     <transformers>
                         <transformer
--- a/aggregator/pom.xml
+++ b/aggregator/pom.xml
@@ -59,13 +59,11 @@
         <dependency>
             <groupId>org.openucx</groupId>
             <artifactId>jucx</artifactId>
-            <version>${ucx.version}</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>com.nvidia</groupId>
             <artifactId>spark-rapids-jni</artifactId>
-            <version>${spark-rapids-jni.version}</version>
             <classifier>${cuda.version}</classifier>
             <scope>provided</scope>
         </dependency>
--- a/aggregator/pom.xml
+++ b/aggregator/pom.xml
@@ -55,6 +55,20 @@
             <version>${project.version}</version>
             <classifier>${spark.version.classifier}</classifier>
         </dependency>
+        <!-- set provided scope for JNI and ucx to avoid pulling into aggregator jar  -->
+        <dependency>
+            <groupId>org.openucx</groupId>
+            <artifactId>jucx</artifactId>
+            <version>${ucx.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.nvidia</groupId>
+            <artifactId>spark-rapids-jni</artifactId>
+            <version>${spark-rapids-jni.version}</version>
+            <classifier>${cuda.version}</classifier>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -301,7 +301,7 @@
                         </configuration>
                     </execution>
                     <execution>
-                        <phase>generate-resources</phase>
+                        <phase>generate-sources</phase>
                         <goals>
                             <goal>run</goal>
                         </goals>
@@ -408,6 +408,42 @@
                         <exclude>*.md</exclude>
                     </excludes>
                 </configuration>
+            </plugin>
+            <plugin>
+                <!-- The dependency scopes of spark-rapids-jni and jucx are provided in sql-plugin, shuffle-plugin and udf-compiler submodules, -->
+                <!-- this will skip the dedupe processing in `create-parallel-world` -->
+                <!-- Unpack spark-rapids-jni and jucx jars to parallel-world directory after `create-parallel-world`, then `parallel-world` will be packaged -->
+                <!-- Note all Shims use the same version of spark-rapids-jni and jucx -->
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>unpack-spark-rapids-jni-and-ucx</id>
+                        <!-- after the generate-sources phase of `create-parallel-world` -->
+                        <phase>generate-resources</phase>
+                        <goals>
+                            <goal>unpack</goal>
+                        </goals>
+                        <configuration>
+                            <artifactItems>
+                                <artifactItem>
+                                    <groupId>com.nvidia</groupId>
+                                    <artifactId>spark-rapids-jni</artifactId>
+                                    <version>${spark-rapids-jni.version}</version>
+                                    <excludes>META-INF</excludes>
+                                    <outputDirectory>${project.build.directory}/parallel-world</outputDirectory>
+                                </artifactItem>
+                                <artifactItem>
+                                    <groupId>org.openucx</groupId>
+                                    <artifactId>jucx</artifactId>
+                                    <version>${ucx.version}</version>
+                                    <excludes>META-INF</excludes>
+                                    <outputDirectory>${project.build.directory}/parallel-world</outputDirectory>
+                                </artifactItem>
+                            </artifactItems>
+                        </configuration>
+                    </execution>
+                </executions>
             </plugin>
         </plugins>
     </build>
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -430,14 +430,13 @@
                                 <artifactItem>
                                     <groupId>com.nvidia</groupId>
                                     <artifactId>spark-rapids-jni</artifactId>
-                                    <version>${spark-rapids-jni.version}</version>
+                                    <classifier>${cuda.version}</classifier>
                                     <excludes>META-INF</excludes>
                                     <outputDirectory>${project.build.directory}/parallel-world</outputDirectory>
                                 </artifactItem>
                                 <artifactItem>
                                     <groupId>org.openucx</groupId>
                                     <artifactId>jucx</artifactId>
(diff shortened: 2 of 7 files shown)


res-life:
build

dist/pom.xml
```diff
@@ -301,7 +301,7 @@
                         </configuration>
                     </execution>
                     <execution>
-                        <phase>generate-resources</phase>
+                        <phase>generate-sources</phase>
```
firestarman, line 304: Is it safe to change the execution stage ?
Maybe no need to change it. Instead inserting your `dependency` plugin behind this one would be enough.
Maven will execute them in sequence.

dist/pom.xml
```diff
@@ -409,6 +409,42 @@
                     </excludes>
                 </configuration>
             </plugin>
+            <plugin>
+                <!-- The dependency scopes of spark-rapids-jni and jucx are provided in sql-plugin, shuffle-plugin and udf-compiler submodules, -->
```
firestarman, line 413: NIT: 
Multiple lines comment can be achieved by 
```
<!--
spark-rapids-jni and jucx are ignored during the dedup process, because they are
 marked as "provided" dependencies and no longer exist in aggregator jars.

Now this "unpack" goal from the dependency plugin will pull out spark-rapids-jni and 
jucx directly from maven local or remote repo, and unpack them into our "parallel-world". 
Then they will be included into the dist jar automatically with packing the
 "parallel-world" folder.

Note all shims will use the same version of spark-rapids-jni and jucx.
-->
```

dist/pom.xml
```diff
@@ -409,6 +409,42 @@
                     </excludes>
                 </configuration>
             </plugin>
+            <plugin>
```
NvTimLiu, line 412: Dist.jar will be created during every buildver=spark-shim, should we unpack spark-rapids-jni/jucx only when building base-spark shim?

res-life: For build `mvn package -Dbuildver=330` which is base-spark shim, also needs the unpack.

res-life:
build

res-life:
build

dist/pom.xml
res-life: done

res-life:
Should target branch-22.08

res-life:
build

aggregator/pom.xml
```diff
@@ -55,6 +55,18 @@
             <version>${project.version}</version>
             <classifier>${spark.version.classifier}</classifier>
         </dependency>
+        <!-- set provided scope for JNI and ucx to avoid pulling into aggregator jar  -->
```
jlowe, line 58: If we cannot find a way to automatically pull in immediate provided dependencies, there needs to be a comment here explaining that any changes here likely impact the corresponding dependency list maintained in dist/pom.xml.

dist/pom.xml
```diff
@@ -409,6 +409,42 @@
...
+                 Note all Shims use the same version of spark-rapids-jni and jucx -->
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>unpack-spark-rapids-jni-and-ucx</id>
+                        <!-- run after `create-parallel-world`, the phase equals to the phase of the `create-parallel-world` -->
+                        <phase>generate-resources</phase>
+                        <goals>
+                            <goal>unpack</goal>
```
jlowe, line 426: Rather than manually maintaining a dependency list here that needs to be sync'd with the same list in the aggregator pom, would it be possible to automatically pull in the immediate provided dependences using the `unpack-dependencies` goal with configs like the following?
```
  <excludeTransitive>true</excludeTransitive>
  <includeScope>provided</includeScope>
```

aggregator/pom.xml
```diff
@@ -55,6 +55,20 @@
...
+        <dependency>
+            <groupId>org.openucx</groupId>
+            <artifactId>jucx</artifactId>
+            <version>${ucx.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.nvidia</groupId>
+            <artifactId>spark-rapids-jni</artifactId>
+            <version>${spark-rapids-jni.version}</version>
```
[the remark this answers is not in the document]
res-life: Done

res-life:
> Has this been tested with the -DallowConventionalDistJar feature added in https://github.com/NVIDIA/spark-rapids/pull/5528?

Tested, and it works. 
```
mvn clean install -DskipTests -DallowConventionalDistJar
```

Also tested `buildall`
```
mvn clean
./build/buildall -p=320,330
# run 330 IT
./integration_tests/run_pyspark_from_build.sh -k test_if_else
```

dist/pom.xml
res-life: `excludeTransitive` should be false because we want the transitive provided jars for the `dist` submodule.

Tested
```
        <goals>
            <goal>unpack-dependencies</goal>
        </goals>
        <configuration>
            <excludeTransitive>false</excludeTransitive>
            <includeScope>provided</includeScope>
        </configuration>
```

It did not work.

```
cd dist
mvn dependency:tree

[INFO] com.nvidia:rapids-4-spark_2.12:jar:22.06.0-SNAPSHOT
[INFO] \- com.nvidia:rapids-4-spark-aggregator_2.12:jar:spark311:22.06.0-SNAPSHOT:compile
[INFO]    +- com.nvidia:rapids-4-spark-sql_2.12:jar:spark311:22.06.0-SNAPSHOT:compile
[INFO]    |  +- com.nvidia:spark-rapids-jni:jar:cuda11:22.06.0-SNAPSHOT:compile
[INFO]    |  |  \- org.slf4j:slf4j-api:jar:1.7.30:compile
[INFO]    |  +- com.nvidia:rapids-4-spark-common_2.12:jar:22.06.0-SNAPSHOT:compile
[INFO]    |  \- com.google.flatbuffers:flatbuffers-java:jar:1.11.0:compile
[INFO]    +- com.nvidia:rapids-4-spark-shuffle_2.12:jar:spark311:22.06.0-SNAPSHOT:compile
[INFO]    |  \- org.openucx:jucx:jar:1.12.1:compile
[INFO]    \- com.nvidia:rapids-4-spark-udf_2.12:jar:spark311:22.06.0-SNAPSHOT:compile
[INFO] ------------------------------------------------------------------------

cd aggregator
mvn dependency:tree

[INFO] com.nvidia:rapids-4-spark-aggregator_2.12:jar:22.06.0-SNAPSHOT
[INFO] +- com.nvidia:rapids-4-spark-sql_2.12:jar:spark311:22.06.0-SNAPSHOT:compile
[INFO] |  +- com.nvidia:rapids-4-spark-common_2.12:jar:22.06.0-SNAPSHOT:compile
[INFO] |  \- com.google.flatbuffers:flatbuffers-java:jar:1.11.0:compile
[INFO] +- com.nvidia:rapids-4-spark-shuffle_2.12:jar:spark311:22.06.0-SNAPSHOT:compile
[INFO] +- com.nvidia:rapids-4-spark-udf_2.12:jar:spark311:22.06.0-SNAPSHOT:compile
[INFO] +- org.openucx:jucx:jar:1.12.1:provided
[INFO] \- com.nvidia:spark-rapids-jni:jar:cuda11:22.06.0-SNAPSHOT:provided
[INFO]    \- org.slf4j:slf4j-api:jar:1.7.30:provided
[INFO] ------------------------------------------------------------------------


```

The output of `dependency:tree`  in `dist` is:
[INFO]    |  \- org.openucx:jucx:jar:1.12.1:compile

While the output of ``  in `aggregator` is:
[INFO] +- org.openucx:jucx:jar:1.12.1:provided

The `jucx` is provided in aggregator.xml.
Seems weird.

Currently, we only have 2 external jars to handle.
IMO we now can explicitly specify the external jars, although there are 2 places to maintain.

aggregator/pom.xml
gerashegalov, line 58: Alternatively we could simply add these artifacts to the exclude list of the shade artifactSet
```
                    <artifactSet>
                        <excludes>
                            <exclude>org.slf4j:*</exclude>
                            <exclude>com.nvidia:spark-rapids-jni:*</exclude>
                            <exclude>org.openucx:jucx:*</exclude>
                        </excludes>
                    </artifactSet>
```

dist/pom.xml
```diff
@@ -409,6 +409,42 @@
...
+                        </goals>
+                        <configuration>
```
gerashegalov, line 428: To have a more obvious way of reasoning about the order of the dependency processing I would move this unpacks inside ant `build-parallel-worlds.xml`  to `copy-dependencies` target either before or after the for loop.

res-life:
build

aggregator/pom.xml
res-life: Yes, this one is more natural. Done.

dist/pom.xml
res-life: Not found a way to invoke the `unpack` goal of `maven-dependency-plugin` in the `maven-antrun-plugin`.
Just coping the jars in the local `~/.m2` path is not so good,  because we can specify the local repository path when executing `mvn` commands.  The `maven-dependency-plugin` is responsible to copy the right jars.

Now the `unpack` works and the order is right, let's keep it?

jlowe (approved):

dist/pom.xml
gerashegalov: it can be done similar to the dependency:get code there. 
```
+    <target name="unpack-ucx-and-jni">
+        <resources id="unpackArray">
+            <string>com.nvidia:spark-rapids-jni:${project.version}:jar:${cuda.version}</string>
+            <string>org.openucx:jucx:${ucx.version}</string>
+        </resources>
+        <pathconvert property="unpackList" refid="unpackArray" pathsep=","/>
+        <ac:for list="${unpackList}" param="unpackArtifact" trim="true">
+            <sequential>
+                <echo level="info">dependency:unpack @{unpackArtifact}</echo>
+                <exec executable="${maven.home}/bin/mvn" failonerror="true">
+                    <arg value="dependency:unpack"/>
+                    <arg value="-s" if:set="env.URM_URL"/>
+                    <arg value="${project.basedir}/../jenkins/settings.xml" if:set="env.URM_URL"/>
+                    <arg value="-B"/>
+                    <arg value="-Dmaven.repo.local=${maven.repo.local}" if:set="maven.repo.local"/>
+                    <arg value="-Dartifact=@{unpackArtifact}"/>
+                    <arg value="-DoutputDirectory=${project.build.directory}/parallel-world"/>
+                </exec>
+            </sequential>
+        </ac:for>
+    </target>
``` 

Your suggestion is fine with me.

gerashegalov (approved):
LGTM

res-life:
build

gerashegalov:
build

sperlingxx:
LGTM

sperlingxx (approved):
LGTM

res-life:
Re-target 22.08

gerashegalov (approved):

GaryShen2008 (approved):

PR #5604, opened 2022-05-24
ORC encrypted write should fallback to CPU [databricks]

### LINKED ISSUE #5463: [FEA] Check on ORC encryption for writes
We have had some issues with parquet write encryption not falling back to the CPU. We need to check on ORC and potentially fix it as well.

razajafri:
This PR falls back to the CPU if any of the ORC encryption configurations are set. 

The configs that we are checking for are 
```
hadoop.security.key.provider.path
orc.key.provider
orc.encrypt
orc.mask
```

fixes #5463 

### DIFF
--- a/integration_tests/src/main/python/orc_write_test.py
+++ b/integration_tests/src/main/python/orc_write_test.py
@@ -15,6 +15,7 @@
 import pytest
 
 from asserts import assert_gpu_and_cpu_writes_are_equal_collect, assert_gpu_fallback_write
+from spark_session import is_databricks104_or_later
 from datetime import date, datetime, timezone
 from data_gen import *
 from marks import *
@@ -172,6 +173,7 @@ def test_write_empty_orc_round_trip(spark_tmp_path, orc_gens):
 @pytest.mark.parametrize("provider", ["", "hadoop"])
 @pytest.mark.parametrize("encrypt", ["", "pii:a"])
 @pytest.mark.parametrize("mask", ["", "sha256:a"])
+@pytest.mark.skipif(is_databricks104_or_later(), reason="The test will fail on Databricks10.4 because `HadoopShimsPre2_3$NullKeyProvider` is loaded")
 def test_orc_write_encryption_fallback(spark_tmp_path, spark_tmp_table_factory, path, provider, encrypt, mask):
     def write_func(spark, write_path):
         writer = unary_op_df(spark, gen).coalesce(1).write
--- a/integration_tests/src/main/python/orc_write_test.py
+++ b/integration_tests/src/main/python/orc_write_test.py
@@ -166,3 +166,29 @@ def test_write_empty_orc_round_trip(spark_tmp_path, orc_gens):
         lambda spark, path: spark.read.orc(path),
         data_path,
         conf={'spark.rapids.sql.format.orc.write.enabled': True})
+
+@allow_non_gpu('DataWritingCommandExec')
+@pytest.mark.parametrize("path", ["", "kms://http@localhost:9600/kms"])
+@pytest.mark.parametrize("provider", ["", "hadoop"])
+@pytest.mark.parametrize("encrypt", ["", "pii:a"])
+@pytest.mark.parametrize("mask", ["", "sha256:a"])
+def test_orc_write_encryption_fallback(spark_tmp_path, spark_tmp_table_factory, path, provider, encrypt, mask):
+    def write_func(spark, write_path):
+        writer = unary_op_df(spark, gen).coalesce(1).write
+        if path != "":
+            writer.option("hadoop.security.key.provider.path", path)
+        if provider != "":
+            writer.option("orc.key.provider", provider)
+        if encrypt != "":
+            writer.option("orc.encrypt", encrypt)
+        if mask != "":
+            writer.option("orc.mask", mask)
+        writer.format("orc").mode('overwrite').option("path", write_path).saveAsTable(spark_tmp_table_factory.get())
+    if path == "" and provider == "" and encrypt == "" and mask == "":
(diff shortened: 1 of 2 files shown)


integration_tests/src/main/python/orc_write_test.py
```diff
@@ -166,3 +166,29 @@ def create_empty_df(spark, path):
...
+            writer.option("hadoop.security.key.provider.path", path)
+        if provider != "":
+            writer.option("orc.key.provider", provider)
+        if encrypt != "":
+            writer.option("orc.encrypt", encrypt)
+        if mask != "":
+            writer.option("orc.mask", mask)
+        writer.format("orc").mode('overwrite').option("path", write_path).saveAsTable(spark_tmp_table_factory.get())
+    if path == "" and provider == "" and encrypt == "" and mask == "":
+        pytest.xfail("Expected to fail as non-encrypted write will not fallback to CPU")
```
revans2, line 188: I don't like xfail here. If we expect it to NOT fall back to the CPU, then we should write a test that verifies that it did that.

razajafri: will a skip be better?

revans2:
But the configs overlap is some cases with reading encrypted data. Do we want to fall back to the CPU for a write when the read is the only thing that is encrypted?

razajafri:
> But the configs overlap is some cases with reading encrypted data. Do we want to fall back to the CPU for a write when the read is the only thing that is encrypted?

That's a good point but I haven't found a way to set the options in a global sense. The only two ways that I found we can set the options are while creating a table like 

```
CREATE TABLE encrypted (
  ssn STRING,
  email STRING,
  name STRING
)
USING ORC
OPTIONS (
  hadoop.security.key.provider.path "kms://http@localhost:9600/kms",
  orc.key.provider "hadoop",
  orc.encrypt "pii:ssn,email",
  orc.mask "nullify:ssn;sha256:email"
)
```
Further details of this can be found [here](https://spark.apache.org/docs/latest/sql-data-sources-orc.html#columnar-encryption) and the other one is while reading or writing the individual file like `spark.read.option(...).orc(...)` or `spark.write.option(...).orc(...)`

In the first case, the table write and read are both encrypted so I don't think there is a way to have one and not the other, in the second case, it's up to the user whether they want to encrypt the read or write. 

I wasn't able to find the case you are suggesting which will provide a global option to be used for writing ORC. If this case exists, then yes this could be a problem.

revans2 (dismissed):

razajafri:
build

razajafri:
So, it seems like Databricks 3.2.1 is failing this test because `HadoopShimsPre2_3$NullKeyProvider` is loaded

razajafri:
build

razajafri:
@revans2 can you take another look please

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuOrcFileFormat.scala
```diff
@@ -58,6 +58,17 @@ object GpuOrcFileFormat extends Logging {
         s"${RapidsConf.ENABLE_ORC_WRITE} to true")
     }
 
+    val keyProviderPath= options.getOrElse("hadoop.security.key.provider.path", "")
```
tgravescs, line 61: nit space before =

integration_tests/src/main/python/orc_write_test.py
```diff
@@ -166,3 +167,30 @@ def create_empty_df(spark, path):
         lambda spark, path: spark.read.orc(path),
         data_path,
         conf={'spark.rapids.sql.format.orc.write.enabled': True})
+
+@allow_non_gpu('DataWritingCommandExec')
+@pytest.mark.parametrize("path", ["", "kms://http@localhost:9600/kms"])
+@pytest.mark.parametrize("provider", ["", "hadoop"])
+@pytest.mark.parametrize("encrypt", ["", "pii:a"])
+@pytest.mark.parametrize("mask", ["", "sha256:a"])
+@pytest.mark.skipif(is_databricks104_or_later(), reason="The test will fail on Databricks10.4 because `HadoopShimsPre2_3$NullKeyProvider` is loaded")
```
tgravescs, line 176: does it make sense to try to gset the provider to something else to be able to test behavior on DB?

razajafri: I tried setting it to `test:///` which is what spark sets in their test code, but that doesn't work.

I am following the Spark teams lead on just skipping these tests although their tests are in Scala so they have control over skipping it only if the provider doesn't have test keys. I am just skipping them for DB10.4 only.

razajafri: Other combinations of the conf have resulted in similar errors

razajafri: So, I am not saying DB doesn't support encryption but when I set the provider to `unknown` I get errors saying that the provider is required. When I set it to `hadoop` I get the following error
```
Cause: org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 1.0 failed 1 times, most recent failure: Lost task 0.0 in stage 1.0 (TID 1) (localhost executor driver): java.lang.IllegalArgumentException: Unknown key pii
```
If I set it to `aws` I get the error that the provider url is unreachable or something along those lines. I imagine an encryption server needs to be setup before encryption can be actually tested on DB.

tgravescs: ok lets leave it as skip for 22.06, can you file a follow on issue to see if we can find a way to test on db 10.4.  We can discuss on the issue to see if we think its needed.

razajafri: Here is the follow-on. Please add to the issue or comment on it if something needs to be added 

#5722

PR #5605, opened 2022-05-24
Enable merge schema reading on ORC

### LINKED ISSUE #135: [BUG] mergeSchema on ORC reads does not work
**Describe the bug**
This is a lot like #60 but for ORC files.  If you try to use `mergeSchema` or provide your own reader schema that has more columns than the orc file does it results in an error.

an integration test is being added for this.

### LINKED ISSUE #5562: [BUG] read ORC file with various file schemas
**Describe the bug**
When I tried to remove the restriction of mergeSchema on ORC reading, GPU ORC reader returned me some weird outputs. With some investigation, I found that the problem occured when we read ORC file with various schemas. Interestingly, the problem occurs occasionally. And it occurs under all 3 read patterns.

```python
def test_merge_schema_read_1(spark_tmp_path):
    first_data_path = spark_tmp_path + '/ORC_DATA/key=0'
    with_cpu_session(
            lambda spark : gen_df(spark, [('a', int_gen)], length=50).write.orc(first_data_path))
    second_data_path = spark_tmp_path + '/ORC_DATA/key=1'
    with_cpu_session(
            lambda spark : gen_df(spark, [('b', int_gen), ('a', int_gen)], length=50).write.orc(second_data_path))
    data_path = spark_tmp_path + '/ORC_DATA'
            lambda spark : spark.read.orc(data_path),
            conf={'spark.rapids.sql.format.orc.reader.type': 'PERFILE',
                  'spark.sql.sources.useV1SourceList': ''})
```

sperlingxx:
Fixes #135 
Fixes #5562

This is to enable merge schema reading on ORC format. Meanwhile, it fixes a bug which fails reading ORC data from multiple file schemas, such as: reading merged schemas of different partitions.  The root cause of the bug is multi-thread modification of a thread unsafe variable (broadcasted Hadoop Conf).

### DIFF
--- a/integration_tests/src/main/python/orc_test.py
+++ b/integration_tests/src/main/python/orc_test.py
@@ -274,7 +274,6 @@ def test_partitioned_read_just_partitions(spark_tmp_path, v1_enabled_list, reade
             lambda spark : spark.read.orc(data_path).select("key"),
             conf=all_confs)
 
-@pytest.mark.xfail(reason='https://github.com/NVIDIA/spark-rapids/issues/135')
 @pytest.mark.parametrize('v1_enabled_list', ["", "orc"])
 @pytest.mark.parametrize('reader_confs', reader_opt_confs, ids=idfn)
 def test_merge_schema_read(spark_tmp_path, v1_enabled_list, reader_confs):
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala
@@ -125,9 +125,6 @@ object GpuOrcScan {
   def tagSupport(scanMeta: ScanMeta[OrcScan]): Unit = {
     val scan = scanMeta.wrapped
     val schema = StructType(scan.readDataSchema ++ scan.readPartitionSchema)
-    if (scan.options.getBoolean("mergeSchema", false)) {
-      scanMeta.willNotWorkOnGpu("mergeSchema and schema evolution is not supported yet")
-    }
     tagSupport(scan.sparkSession, schema, scanMeta)
   }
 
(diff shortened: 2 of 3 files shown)


sperlingxx:
build

wbo4958 (approved):

revans2:
How are we handling the schema evolution that ORC supports on java? I documented it [here](https://github.com/rapidsai/cudf/issues/5447) for CUDF, but it turned out to be java specific.

revans2:
https://github.com/apache/orc/blob/acbf0205e0919d16f7daf239c367fdd6dfb22ecc/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java#L2122-L2249

shows what is supported and before we can call this fixed we will need tests to verify that all of these work out of the box.

revans2:
Sorry I should add that it is for the types that we support. UNION is not supported so we can ignore it.

revans2 (changes requested):
Need to do what I described above. At least adding in tests for the schema evolution described above.

sperlingxx:
@revans2, with some tests, I found spark-rapids can not handle schema evolution with rules ORC supports on java at all. And If I am not wrong, it is a definate non-trivial work to support all schema evolution rules of ORC.

revans2:
@sperlingxx Yes it is non-trivial, but it should be doable. We have almost all of that working with cast. There are likely to be minor differences, but we should be able to make ti all work.

The part that scares me most is that we can still trigger schema evolution without mergeSchema. If I as a user pass in a schema myself when reading an ORC file I can run into similar situations.  If it is too large to do at once here, then we should split the work up into smaller issues along with an epic and work through them.

sameerz:
@sperlingxx can we retarget this to 22.08?

sperlingxx:
Close this PR since it doesn't address the issue.

sperlingxx:
Hi @revans2, is this PR left a data consistency issue unsolved as well ? If any, could you elaborate the issue ? I have little clue on it. Thank you.

revans2:
@sperlingxx sorry if I am not clear about this PR.

1.  This PR says it fixes #5562, but it does not.
2. This PR says it fixes  #135, but it just turns off checking if it was enabled. It does not fix any of the underlying issues with schema evolution.
Yes the check for merge schema is a hack and is not perfect when dealing with schema evolution. This is why we need to fix the underlying issues. Just deleting the check does not "fix" any of it. If you want to merge in this PR it increases the likelihood that we will hit these bugs, but it does not eliminate them.

Issue #5525, opened 2022-05-18
[BUG] Investigate more edge cases in regexp support


andygrove:
**Describe the bug**
We have fuzz tests that are passing but through some experimentation I was able to make the fuzz tests generate previously unseen patterns and data and this exposes more edge cases where we are not consistent between CPU and GPU.

## regexp_find examples

```
javaPattern=$x*\r, cudfPattern=(?:[\n\r\u0085\u2028\u2029]|\r\n)?$x*\r, input='\u000b.\r', cpu=true, gpu=false
javaPattern=\w[\r,B]\Z, cudfPattern=\w[\r,B](?:[\n\r\u0085\u2028\u2029]|\r\n)?$, input='D31b\r\n', cpu=false, gpu=true
```

## regexp_replace examples

```
javaPattern=^$\s, cudfPattern=^([\n\r\u0085\u2028\u2029]|\r\n)?$[ \u000b\n\t\r\f], input='\n', cpu=_RE\PLACE_, gpu=\n
javaPattern=\s\Z\Z, cudfPattern=[ \u000b\n\t\r\f]([\n\r\u0085\u2028\u2029]|\r\n)?$, input='w\f\n', cpu=w_RE\PLACE__RE\PLACE_, gpu=w_RE\PLACE_\n
```

**Steps/Code to reproduce bug**
Add one or more extra character to `RegularExpressionTranspilerSuiteREGEXP_LIMITED_CHARS_COMMON` and run the existing tests.

**Expected behavior**
No failures.

**Environment details (please complete the following information)**
N/A

**Additional context**
None

PR #5610, opened 2022-05-24
Fall back to CPU for unsupported regular expression edge cases with end of line/string anchors and newlines

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/5525

Through expanded fuzz testing, we recently discovered some edge cases that produce different results between CPU and GPU for regular expressions patterns with an end of line anchor `$` immediately next to a newline, begin-of-line anchor `^` or repetition that could produce empty results.

New code is added in this PR that checks for these unsupported patterns so that we can fall back to CPU for these cases. 

- [x] Implement checks
- [x] Update tests

Note: The checks for these edge cases are very broad and result in some regression and false positives which are outlined here #5659


tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -238,21 +244,27 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
...
   }
 
   test("line anchor $ - find") {
-    val patterns = Seq("$\r", "a$", "\r$", "\f$", "$\f", "\u0085$", "\u2028$", "\u2029$", "\n$",
-        "\r\n$", "[\r\n]?$", "\\00*[D$3]$", "a$b")
+    // are these special cases that we can allow? what is the rule?
+    // maybe these are ok because the $ is at the end of the pattern?
+    // "\r$", "\f$",  "\u0085$", "\u2028$", "\u2029$", "\n$", "\r\n$", "[\r\n]?$"
+    // not sure about these ...
+    // "$\r", "$\f", "\\00*[D$3]$"
```
anthony-chang, line 256: @NVnavkumar handled the `$\r` case in #5289, which transpiles to `\r(?:[\n\u0085\u2028\u2029])?$`, and similarly for `\u0085`, `\u2028`, `\u2029`. The other two don't have line terminators so I don't think there is a problem

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -338,6 +338,10 @@ class RegexParser(pattern: String) {
         throw new RegexUnsupportedException("escape at end of string", Some(pos))
       case Some(ch) =>
         ch match {
+          case 'r' | 'n' | 'f' =>
```
anthony-chang, line 341: Why do we need this? Wouldn't it treat `\\n` as the character literal `\n`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -696,6 +700,99 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
+        }
+      }
+//      println("everything is fine")
+    }
+
+    def checkEndAnchorNearNewline(regex: RegexAST): Unit = {
+      regex match {
+        case RegexSequence(parts) =>
+          parts.indices.foreach { i =>
+            if (i > 0) {
```
anthony-chang, line 774: Is this case necessary since we are already checking both `r1` and `r2` for if either are line anchors?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
andygrove: I ran into an exception something like "unsupported escape character 'r'". If you comment out this code and run the tests you should be able to reproduce it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
andygrove: Good point

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -383,6 +397,18 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
     assertCpuGpuMatchesRegexpFind(patterns, inputs)
   }
 
+  test("fall back to CPU for newline next to line or string anchor") {
+    // these patterns were discovered during fuzz testing and resulted in different
+    // results between CPU and GPU
+    val patterns = Seq(raw"\w[\r,B]\Z", raw"\s\Z\Z", "^$\\s", "$x*\\r", "$\\r")
```
anthony-chang, line 403: `\\r` is an escaped 'r' which isn't a valid regexp. `\r` is the carriage return 
```suggestion
    val patterns = Seq(raw"\w[\r,B]\Z", raw"\s\Z\Z", "^$\\s", "$x*\r", "$\r")
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
anthony-chang: See [my comment](https://github.com/NVIDIA/spark-rapids/pull/5610#discussion_r880892707). This case should be removed since it would incorrectly convert escaped characters to literals and throw the wrong error message.

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
andygrove: These patterns were generated by the fuzz tests and resulted in different behavior between CPU and GPU so I think we need to make sure we fall back to CPU for these patterns.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
andygrove: Agreed. I am working on removing this now.

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
andygrove: On second thought I think you are right. I think I ended up with an extra `\` in here from copy/paste.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -696,6 +696,101 @@ class CudfRegexTranspiler(mode: RegexMode) {
     }
   }
 
+  private def transpile(regex: RegexAST, replacement: Option[RegexReplacement],
+      previous: Option[RegexAST]): RegexAST = {
+
+    def isMaybeAnchor(regex: RegexAST): Boolean = {
+      contains(regex, {
+        case RegexChar('^') | RegexChar('$') |
+             RegexEscaped('A') | RegexEscaped('z') | RegexEscaped('Z') => true
```
anthony-chang, line 705: Correct me if I'm wrong, but I don't believe we ran into any cases where the beginning-of-line anchor was inconsistent? If so then we could enable them here.

andygrove: You are correct, we had not seen that before, but the new fuzz tests exposed a case. If you remove the `^` here and run the tests you will see:

```
javaPattern[1664]=$^\r, cudfPattern=(?:[\n\r\u0085\u2028\u2029]|\r\n)?$^\r, input='\r', cpu=true, gpu=false
```

anthony-chang: Hmm I suspect that has more to do with the `$` anchor being problematic since the patterns `^\r` and `\r^` both pass. Maybe we can split `isMaybeAnchor` to `isMaybeBeginningAnchor` and `isMaybeEndingAnchor` and throw an exception if the beginning-of-line anchor immediately precedes or succeeds an end-of-line anchor to limit our false positives?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -696,6 +713,90 @@ class CudfRegexTranspiler(mode: RegexMode) {
     }
   }
 
+  private def transpile(regex: RegexAST, replacement: Option[RegexReplacement],
+      previous: Option[RegexAST]): RegexAST = {
```
anthony-chang, line 717: nit: `replacement` and `previous` is never used here, just passed on to `rewrite`. It may be more descriptive to remove these, rename this method to `checkUnsupported` (take the code out of the closure), and then call this and `rewrite` inside `transpile`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -696,6 +713,90 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
+          case SimpleQuantifier('*') | SimpleQuantifier('?') => true
+          case QuantifierFixedLength(0) => true
+          case QuantifierVariableLength(0, _) => true
+          case _ => false
+        }
+        case _ => false
+      })
+    }
+
+    def checkPair(r1: RegexAST, r2: RegexAST): Unit = {
```
jlowe, line 761: nit: would be nice for this to have a more descriptive name, like "checkEndAnchorContext", so it's more obvious to the reader what this is checking.

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -308,7 +319,10 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
     assert(transpiled === expected)
   }
 
-  test("transpile complex regex 2") {
+  // TODO
+  // this was a test for transpiling but we did not ever try to run the
+  // resulting regex to see if it produced valid results
```
jlowe, line 324: Is there a followup issue for this?  Would be nice to link it here.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -696,6 +713,89 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
+          (containsNewline(r1) || containsEmpty(r1) || containsBeginAnchor(r1)))) {
+        throw new RegexUnsupportedException(
+          s"End of line/string anchor is not supported in this context: " +
+            s"${toReadableString(r1.toRegexString)}" +
+            s"${toReadableString(r2.toRegexString)}")
+      }
+    }
+
+    def checkUnsupported(regex: RegexAST): Unit = {
+      regex match {
```
anthony-chang, line 775: Maybe we can only run this if `containsEndAnchor(regex) && containsNewline(regex)` is true to support patterns like `^([a-z]*)([0-9]*)([a-z]*)$`

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -696,6 +713,90 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
+    def containsBeginAnchor(regex: RegexAST): Boolean = {
+      contains(regex, {
+        case RegexChar('^') | RegexEscaped('A') => true
+        case _ => false
+      })
+    }
+
+    def containsEndAnchor(regex: RegexAST): Boolean = {
+      contains(regex, {
+        case RegexChar('$') | RegexEscaped('z') | RegexEscaped('Z') => true
```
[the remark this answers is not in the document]
anthony-chang: Opened #5659 to track this

andygrove:
build

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
andygrove: I filed https://github.com/NVIDIA/spark-rapids/issues/5656 and linked to it

andygrove:
I believe that the CI failures are related to recent changes in the container around UTF-8. I can reproduce them locally if I unset `LANG`.

andygrove:
build

andygrove:
build

andygrove:
build

andygrove:
build

andygrove:
build

andygrove:
build

andygrove:
build

andygrove:
build

andygrove:
@anthony-chang this is passing now but we'll need a follow-on issue to add the characters that I removed from the fuzzer to make this pass in CI

anthony-chang (approved):
> @anthony-chang this is passing now but we'll need a follow-on issue to add the characters that I removed from the fuzzer to make this pass in CI

Created follow-up issue: #5711

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -745,24 +762,83 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
+      contains(regex, {
+        case RegexChar('\r') | RegexEscaped('r') => true
+        case RegexChar('\n') | RegexEscaped('n') => true
+        case RegexChar('\u0085') | RegexChar('\u2028') | RegexChar('\u2029') => true
+        case RegexEscaped('s') | RegexEscaped('v') | RegexEscaped('R') => true
+        case RegexEscaped('W') | RegexEscaped('D') | RegexEscaped('S') | RegexEscaped('V') =>
+          // these would get transpiled to negated character classes
+          // that include newlines
+          true
+        case RegexCharacterClass(true, _) => true
```
NVnavkumar, line 789: nit: there is an edge case here that can prevent this from being a newline, and that being a negated character class that includes all new line characters. I see no need to handle this at the moment, but should be noted in case this potentially comes up in some testing scenarios.

NVnavkumar (approved):
LGTM

andygrove:
@NVnavkumar @anthony-chang I will retarget this to 22.08

andygrove:
build

Issue #5570, opened 2022-05-20
Qualification tool investigate SQL Df value reported


tgravescs:
**Is your feature request related to a problem? Please describe.**
Currently with https://github.com/NVIDIA/spark-rapids/pull/5550/files we report the sql dataframe duration based on the stages we associated with supported operations. This could be short if execs don't have stages associated with it.  See what the diff is and perhaps use the spark reported sql df time and report the difference.

PR #5612, opened 2022-05-24
Qualification tool: update SQL Df value used and look at jobs in SQL

tgravescs:
Fixes https://github.com/NVIDIA/spark-rapids/issues/5570

Ok this has a couple of changes. 

1) Change to use the spark reported SQL DF duration instead of the stage duration since we might not have all stages.
2) I found that the sql mapping to stages can miss quite a few stages, especially in the case an entire SQL query doesn't have any mappings.  In these cases look at the SQL to Job mapping and get the stages from those and just average everything in that SQL query across those times.
3) Fixed a couple bugs.
4) removed a decimal check we should have removed a while.

This change makes the DF times much closer then what they were in previous version of the tool.


tgravescs:
build

tgravescs:
build

tools/src/main/scala/org/apache/spark/sql/rapids/tool/qualification/QualificationAppInfo.scala
```diff
@@ -303,18 +355,32 @@ class QualificationAppInfo(
...
+        info.duration.getOrElse(0L)
+      }
+      val sparkSQLDFWallClockDuration = allSQLDurations.sum
+      val longestSQLDuration = if (allSQLDurations.size > 0) {
+        allSQLDurations.max
+      } else {
+        0L
+      }
+
+      val sqlDFSupportedWallClockDurationBasedStages = perSqlStageSummary.map { s =>
```
nartal1, line 371: This variable is not used in any further calculation. Should it be removed OR should we take max of this and `sparkSQLDFWallClockDuration` while passing  in `QualificationAppInfo.calculateEstimatedInfoSummary` method ?

tgravescs:
build

nartal1 (approved):

PR #5623, opened 2022-05-24
Handle duplicates in negated character classes

anthony-chang:
The fuzz tests in #5610 caught an edge case where we have duplicates within a negated character class. 
```
javaPattern[4898]=[^\r\r], cudfPattern=[^\r\r], input='\n', cpu=true, gpu=false
```

This PR fixes this case. 


anthony-chang:
build

andygrove (approved):

PR #5643, opened 2022-05-25
Fix missing GpuBatchScanExec metrics in SQL UI

### LINKED ISSUE #5615: [BUG] GpuBatchScanExec only reports output row metrics
**Describe the bug**
GpuBatchScanExec is supposed to show the same types of GPU-related metrics as FileSourceScanExec (e.g.: buffer time, GPU decode time, output batches, etc.) but output rows are the only metrics shown even with `spark.rapids.sql.metrics.level=DEBUG`.

- Set `spark.rapids.sql.metrics.level` to `DEBUG`
- Set `spark.sql.sources.useV1SourceList` to `""`
- Run a simple query to scan a Parquet file (e.g.: `spark.read.parquet("tests/src/test/resources/timestamp-date-test.parquet").collect()`
- Examine the SQL UI and note that only output rows are shown as a metric for the GPU batch scan, e.g.:

image

**Expected behavior**
All of the metrics setup in `GpuBatchScanExecMetrics` should appear in the SQL UI for the `GpuBatchScan` node rather than just output rows.

Replicated the problem with Spark 3.1.2 and Spark 3.2.1

jlowe:
Fixes #5615.  SQLMetric instances were being created on the executor separately from the metrics on the driver, causing the metrics to get disconnected.  The eventlog would contain the metric data for the missing metrics, but the metric IDs would not match those predeclared in the SQL execution start event.  That explains why the SQL UI wouldn't show them, as it was looking for metric IDs that never received any updates in any of the task completion events.

### DIFF
--- a/sql-plugin/src/main/311until320-all/scala/com/nvidia/spark/rapids/shims/GpuBatchScanExec.scala
+++ b/sql-plugin/src/main/311until320-all/scala/com/nvidia/spark/rapids/shims/GpuBatchScanExec.scala
@@ -30,16 +30,16 @@ case class GpuBatchScanExec(
     @transient scan: Scan) extends DataSourceV2ScanExecBase with GpuBatchScanExecMetrics {
   @transient lazy val batch: Batch = scan.toBatch
 
-  scan match {
-    case s: ScanWithMetrics => s.metrics = allMetrics ++ additionalMetrics
-    case _ =>
-  }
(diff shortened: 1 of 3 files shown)


jlowe:
build

revans2 (approved):

tgravescs (approved):

PR #5645, opened 2022-05-25
Add support for octal digits in character classes

### LINKED ISSUE #4862: [FEA] Add support for regular expressions containing octal digits inside character classes , eg`[\0177]`
**Describe the bug**
We support octal digits but with limitations. Right now, octal digits cannot be used inside character classes on the GPU.

Regular expressions with octal digits inside character classes do not work. Examples include `[\02]`, `[\0177]`, `[\077]`, etc.

**Expected behavior**
We should support octal digits inside character classes.

anthony-chang:
Closes #4862

This PR adds support for octal digits within regular expression character classes. 

### DIFF
--- a/integration_tests/src/main/python/string_test.py
+++ b/integration_tests/src/main/python/string_test.py
@@ -881,6 +881,7 @@ def test_character_classes():
                 'rlike(a, "[\n-\\]")',
                 'rlike(a, "[+--]")',
                 'regexp_extract(a, "[123]", 0)',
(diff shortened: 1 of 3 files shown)


anthony-chang:
build

andygrove:
> This looks good based on the tests but how is this supported now? Are we transpiling the character range in this case to something that cuDF supports?

Ah, never mind. I see now that we simply parse the octal digit as a RegexChar now.

andygrove (approved):

PR #5676, opened 2022-05-27
Fix ORC reading over different schemas

### LINKED ISSUE #5562: [BUG] read ORC file with various file schemas

sperlingxx:
Fixes #5562

This PR is to fix the bug which fails reading ORC data from multiple file schemas. The root cause of the bug is multi-thread modification on a thread unsafe variable (broadcasted Hadoop Conf).

### DIFF
--- a/integration_tests/src/main/python/orc_test.py
+++ b/integration_tests/src/main/python/orc_test.py
@@ -299,6 +299,26 @@ def test_merge_schema_read(spark_tmp_path, v1_enabled_list, reader_confs):
 
 @pytest.mark.parametrize('v1_enabled_list', ["", "orc"])
 @pytest.mark.parametrize('reader_confs', reader_opt_confs, ids=idfn)
+def test_orc_read_multiple_schema(spark_tmp_path, v1_enabled_list, reader_confs):
+    first_gen_list = [('a', int_gen), ('b', int_gen)]
+    first_data_path = spark_tmp_path + '/ORC_DATA/key=0'
+    with_cpu_session(
+        lambda spark: gen_df(spark, first_gen_list, num_slices=10).write.orc(first_data_path))
+    second_gen_list = [('a', int_gen), ('b', int_gen), ('c', int_gen)]
+    second_data_path = spark_tmp_path + '/ORC_DATA/key=1'
+    with_cpu_session(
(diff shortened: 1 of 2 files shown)


sperlingxx:
build

jlowe (commented):
I find it interesting that this PR claims it fixes #5562 but does not use the same repro test from that issue as part of its test case.  The test in #5562 seems subtly different than the one used here, as the order of columns in the dataframe is intentionally different, whereas here the test preserves the same order and the extra one is on the end.  Have we verified the same test from #5562 now passes with this change?

sperlingxx:
> I find it interesting that this PR claims it fixes #5562 but does not use the same repro test from that issue as part of its test case. The test in #5562 seems subtly different than the one used here, as the order of columns in the dataframe is intentionally different, whereas here the test preserves the same order and the extra one is on the end. Have we verified the same test from #5562 now passes with this change?

Yes, the case in #5562 can be addressed by this PR.  I decided to use current case rather than the original case, because it looks stronger and more representative. In terms of the column order, I shuffled the order of columns to include the factor into the case.

sperlingxx:
build

tgravescs (approved):

Issue #5559, opened 2022-05-20
[FEA] Add `GpuMapConcat` support for nested (array, struct, map) types.


HaoYang670:
**Is your feature request related to a problem? Please describe.**
This is a follow on issue of https://github.com/NVIDIA/spark-rapids/pull/5533.

**Describe the solution you'd like**
`GpuMapConcat` should support arbitrary nested types for values once https://github.com/rapidsai/cudf/pull/10890 is merged.  For keys we need https://github.com/rapidsai/cudf/issues/11093 assuming that it also updates drop_list_duplicates.  These do not have to be done at the same time. We can split them up into separate PRs for keys and values:
- [x] #5686 
- [x] #6290. (Blocked by https://github.com/rapidsai/cudf/issues/11093 )

revans2:
I removed `GpuConcat` because it is covered by #5542

HaoYang670:
>We can split them up into separate PRs for keys and values.

Hi @revans2, I am a little confused about this comment.
Do you mean that we could also support nested keys? 
But are nested types (such as array, struct and map) hashable?

GaryShen2008:
Hi @HaoYang670 , seems https://github.com/rapidsai/cudf/issues/11093 has been closed. Can we support the nested type keys now?

HaoYang670:
Close this issue as we have completed the tasks in the checklist.

PR #5686, opened 2022-05-29
Add `GpuMapConcat` support for nested-type values

HaoYang670:
Re #5559.

Changes in this PR:
1. Support concatenating map columns with nested-type values (including Array, Struct, and Map)
2. Simplify the `GpuMapConcat` function by concatenating the map columns directly.
3. Add nested-type data gen into `map_gens_sample`


HaoYang670:
build

integration_tests/src/main/python/data_gen.py
```diff
@@ -969,7 +969,10 @@ def gen_scalars_for_sql(data_gen, count, seed=0, force_no_nulls=False):
 # Some map gens, but not all because of nesting
 map_gens_sample = all_basic_map_gens + [MapGen(StringGen(pattern='key_[0-9]', nullable=False), ArrayGen(string_gen), max_length=10),
         MapGen(RepeatSeqGen(IntegerGen(nullable=False), 10), long_gen, max_length=10),
-        MapGen(StringGen(pattern='key_[0-9]', nullable=False), simple_string_to_string_map_gen)]
+        MapGen(StringGen(pattern='key_[0-9]', nullable=False), simple_string_to_string_map_gen),
+        MapGen(IntegerGen(False), ArrayGen(int_gen)),
+        MapGen(BooleanGen(False), StructGen([['child0', byte_gen], ['child1', double_gen]]), max_length=10),
```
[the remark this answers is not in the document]
HaoYang670: Updated!

integration_tests/src/main/python/data_gen.py
```diff
@@ -969,7 +969,10 @@ def gen_scalars_for_sql(data_gen, count, seed=0, force_no_nulls=False):
 # Some map gens, but not all because of nesting
 map_gens_sample = all_basic_map_gens + [MapGen(StringGen(pattern='key_[0-9]', nullable=False), ArrayGen(string_gen), max_length=10),
         MapGen(RepeatSeqGen(IntegerGen(nullable=False), 10), long_gen, max_length=10),
-        MapGen(StringGen(pattern='key_[0-9]', nullable=False), simple_string_to_string_map_gen)]
+        MapGen(StringGen(pattern='key_[0-9]', nullable=False), simple_string_to_string_map_gen),
+        MapGen(IntegerGen(False), ArrayGen(int_gen)),
```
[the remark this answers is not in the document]
HaoYang670: > It would be good to have some limits on the total number of map entries and/or the length of the array
Updated!

HaoYang670:
build

HaoYang670:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -96,30 +94,9 @@ case class GpuMapConcat(children: Seq[Expression]) extends GpuComplexTypeMerging
...
-        withResource(Seq(keys.getChildColumnView(0), values.getChildColumnView(0))) { 
-          case Seq(k_child, v_chlid) =>
-            withResource(ColumnView.makeStructView(k_child, v_chlid)) {structs =>
-              withResource(keys.replaceListChild(structs)) { struct_list =>
-                GpuCreateMap.createMapFromKeysValuesAsStructs(dt, struct_list)
-              }
-            }
+      withResource(children.safeMap(columnarEvalToColumn(_, batch).getBase())) {cols =>
+        withResource(cudf.ColumnVector.listConcatenateByRow(cols: _*)) {structs =>
+          GpuCreateMap.createMapFromKeysValuesAsStructs(dataType, structs)
```
ttnghia, line 99: I don't think this PR can address "support nested types". As I know, this API internally calls `dropListDuplicatesWithKeysValues` which doesn't fully support nested types except for STRUCT of basic types.

ttnghia: Ah never mind, I missed the "values" word in the title. So support nested types only for values :smile:

jlowe (approved):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3104,11 +3104,11 @@ object GpuOverrides extends Logging {
     expr[MapConcat](
       "Returns the union of all the given maps",
       ExprChecks.projectOnly(TypeSig.MAP.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL_128 +
-          TypeSig.NULL),
+          TypeSig.NULL + TypeSig.ARRAY + TypeSig.STRUCT + TypeSig.MAP),
         TypeSig.MAP.nested(TypeSig.all),
         repeatingParamCheck = Some(RepeatingParamCheck("input",
           TypeSig.MAP.nested(TypeSig.commonCudfTypes + TypeSig.DECIMAL_128 +
-          TypeSig.NULL),
+          TypeSig.NULL + TypeSig.ARRAY + TypeSig.STRUCT + TypeSig.MAP),
```
revans2, line 3111: The problem here is that the checks don't distinguish between keys and values in a map. This check is adding nested support for both keys and values, but the code can only support nested types for values, not keys.

revans2 (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
HaoYang670: Sorry, I don't know how to check type for `key` and `value` individually. Seems like `ExprChecks.projectOnly` doesn't support this.

HaoYang670: I don't know how could we test the **nested key type** cases, as `MapGen` does not support `ArrayGen` as the key gen. (List in Python is unhashable)

revans2 (dismissed):
Just a nit for documentation

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
HaoYang670: Updated! And I have added 2 cases to test the falling back.

HaoYang670:
build

HaoYang670:
There is a weird failure: https://github.com/NVIDIA/spark-rapids/runs/6897078626?check_suite_focus=true#step:2:28678
```
 pyspark.sql.utils.AnalysisException: java.lang.RuntimeException: The root scratch dir: /tmp/hive on HDFS should be writable. Current permissions are: rwxr-xr-x
```

revans2:
build

revans2:
That is a really odd error. I think it has to be a transient error, but I am not sure on it.

HaoYang670:
@revans2 @jlowe @ttnghia Could you please help to review? Thanks a lot!.

revans2 (approved):
nit: would be nice to have a partial support tag on the param checks, but that is minor.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3103,14 +3103,24 @@ object GpuOverrides extends Logging {
       }),
     expr[MapConcat](
       "Returns the union of all the given maps",
+      // Currently, GpuMapConcat supports nested values but not nested keys.
+      // We will add the nested key support after
+      // https://github.com/rapidsai/cudf/pull/9452 is merged.
```
ttnghia, line 3108: I don't think this is the relevant cudf PR. The dependency should be https://github.com/rapidsai/cudf/issues/11093.

HaoYang670: Updated!

HaoYang670:
build

ttnghia (approved):

HaoYang670:
> nit: would be nice to have a partial support tag on the param checks, but that is minor.

Hi @revans2, sorry I am a little confused about what the `partial support tag` is. Is it the function `tagExprForGpu`?

jlowe:
> I am a little confused about what the partial support tag is.

This is the `withPsNote` method of `TypeSig` where we can add a note about something being only partially supported.  See callers of this method for some examples.

Issue #5670, opened 2022-05-27
[BUG] Job failed when parsing "java.lang.reflect.InvocationTargetException: org.apache.spark.sql.catalyst.parser.ParseException:" 


viadea:
I tried to rewrite the Mortgage ETL using a single raw CSV file downloaded directly from the https://capitalmarkets.fanniemae.com/credit-risk-transfer/single-family-credit-risk-transfer/fannie-mae-single-family-loan-performance-data website.

I can make CPU spark work, however when running in GPU mode, it failed when parsing:
```
Caused by: java.lang.reflect.InvocationTargetException: org.apache.spark.sql.catalyst.parser.ParseException:
```

Complete steps and error stacktrace is in https://github.com/viadea/spark-rapids-examples/blob/new_mortgage_etl/examples/XGBoost-Examples/mortgage/notebooks/scala/mortgage-ETL-new.ipynb .

The input data is a single CSV file : 2000Q1.csv

nvliyuan:
I just verified that the issue is specific to spark3.2.* version.
it works well with:
spark3.1.2  +  v2206-SNAPSHOT
spark3.1.2  +  v2204-GA
it failed with:
spark3.2.0 +  v2204-GA
spark3.2.0  + v2206-SNAPSHOT
The issue should be raised because UTC time data, because if I don't add config
```
  --conf spark.driver.extraJavaOptions=\"-ea -Duser.timezone=UTC\" \
  --conf spark.executor.extraJavaOptions=\"-ea -Duser.timezone=UTC\" \
```
GetTimestamp would fallback to cpu which would not raise issue
```
!Expression <GetTimestamp> gettimestamp(first_pay_date#1686, MMyyyy, TimestampType, Some(Asia/Shanghai), false) cannot run on GPU because Only UTC zone id is supported. Actual zone id: Asia/Shanghai; expression GetTimestamp gettimestamp(first_pay_date#1686, MMyyyy, TimestampType, Some(Asia/Shanghai), false) produces an unsupported type TimestampType
```

jlowe:
I was not able to reproduce this unless I also set `spark.rapids.sql.incompatibleDateFormats.enabled=true`

viadea:
Yes the reason why we need to enable `spark.rapids.sql.incompatibleDateFormats.enabled=true` is because the raw CSV data has the format as "MMyyyy" otherwise it falls back to CPU:
```
[...]

PR #5713, opened 2022-06-01
Fix GpuCast losing the timezoneId during canonialization

tgravescs:
Fixes https://github.com/NVIDIA/spark-rapids/issues/5670

The timezoneid setting was being lost during canonicalization on GpuCast which ultimately caused a parse error to occur because during plan canonicalization we end up trying to access a canoncialized Alias which will end up trying to do a parse if it is not resolved.  The GpuCast wasn't resolved because it didn't have the timezone id.

The fix is straight forward and we had a comment there to add in the needsTimeZone check.  Basically we were intentionally nulling it out when we shouldn't have been because we weren't checking needsTimeZone

I want to try to add a test case to the integration tests but need to find minimal repro cast, putting up here for early review and run tests and get in asap.  I can always followup with test.


tgravescs:
build

jbrennan333 (approved):
+1 this looks good to me.

Issue #5569, opened 2022-05-20
Qualification tool add more tests for stage failures


tgravescs:
**Is your feature request related to a problem? Please describe.**
In the qualification tool we added looking at stage failures. If stage failures happens we rank it as not applicable so user knows it had failures which would not be optimal timings

PR #5728, opened 2022-06-02
Qualification tool: Add test for stage failures

nartal1:
This closes https://github.com/NVIDIA/spark-rapids/issues/5569 .
Added a test which reads eventlog which has jobs/stage failures. For such eventlogs, the recommendation is "Not Applicable" as there isn't sufficient data. Test asserts expected behavior.
Other than this, I manually verified the UI when either job or stage failed. It recommends as "Not applicable" for both cases.


nartal1:
build

tgravescs (approved):

Issue #5451, opened 2022-05-10
[FEA] Update Spark2 explain code for 22.06


tgravescs:
**Is your feature request related to a problem? Please describe.**
We need to update all the Spark2 code for changes in 22.06.

PR #5738, opened 2022-06-03
Update Spark2 explain code for 22.06

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/5451.

Update the spark 2.x explain api code to pull in all the various changes. Run through the diff script and udpated everything I saw.  I manually ran this against tpcds queries and looked at output, only diff from 22.04 was dealing with a cast (Casting from floating point types to integral types) is now supported because we enabled it by default in 22.06.


tgravescs:
build

HaoYang670:
Build failure is due to #5744

tgravescs:
build

nartal1 (approved):
LGTM. I ran diff for some of the files locally and got the same  results as above.
@tgravescs Can this be merged now OR should we wait for pending PR's for 22.06 ?

tgravescs:
as far as I know everything pending isn't messing with the metadata but I'll keep an eye out.  merging it now.

Issue #5739, opened 2022-06-03
[BUG]. Spark 3.3 build failure - QueryExecutionErrors package scope changed


tgravescs:
**Describe the bug**
Build the spark rapids plugin against Spark 3.3

`sql-plugin/src/main/330+/scala/com/nvidia/spark/rapids/shims/RapidsErrorUtils.scala:27: object QueryExecutionErrors in package errors cannot be accessed in package org.apache.spark.sql.errors`

Looks like scope was changed in https://github.com/apache/spark/commit/61d22b6f313c20de1b65a595e88b6f5bd9595299 so we need to trampline it.

PR #5742, opened 2022-06-03
Spark-3.3 build fix - Move QueryExecutionErrors to sql package

nartal1:
This fixes https://github.com/NVIDIA/spark-rapids/issues/5739
As mentioned in the issue QueryExecutionErrors is changed to private scope now. Updated the code to fix build errors.
Calling QueryExecutionErrors methods from org.apache.spark.sql scope.


nartal1:
build

gerashegalov (approved):
LGTM

alternatively could have moved RapidsErrrorUtils to an org.apache.spark.sql namespace package

PR #5745, opened 2022-06-05
Move `RapidsErrorUtils` to `org.apache.spark.sql.shims` package

### LINKED ISSUE #5744: [BUG] Compile failure with Spark 3.2.2
**Describe the bug**
Failure message:
https://github.com/NVIDIA/spark-rapids/runs/6734105319?check_suite_focus=true#step:2:12045

Spark has made `object QueryExecutionErrors` to be `private[sql]`. (https://issues.apache.org/jira/browse/SPARK-39367)
We cannot access it now.

Affected versions: Spark 3.2.2

HaoYang670:
Close #5744 
The `QueryExecutionErrors` has been declared as `private[sql]` in Spark 330+ and Spark 321. Update the code to fix build error.


HaoYang670:
build

sql-plugin/src/main/311until320-nondb/scala/org/apache/spark/sql/rapids/shims/RapidsErrorUtils.scala
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.nvidia.spark.rapids.shims
+package org.apache.spark.sql.rapids.shims
```
tgravescs, line 17: we wouldn't need to move this one, correct?  They only changed in 3.2.1 and later?

tgravescs (approved):
going to merge to fix builds we can discuss if we want to move the < 3.2.1 back after

nartal1:
> going to merge to fix builds we can discuss if we want to move the < 3.2.1 back after

Curious to know why this was merged to 22.08 and not 22.06.

sql-plugin/src/main/311until320-nondb/scala/org/apache/spark/sql/rapids/shims/RapidsErrorUtils.scala
tgravescs: please file a followup issue

sameerz:
> > going to merge to fix builds we can discuss if we want to move the < 3.2.1 back after
> 
> Curious to know why this was merged to 22.08 and not 22.06.

The fix is for Spark-3.2.2 which is as yet unreleased, so would not be targeted for 22.06.

PR #5761, opened 2022-06-07
Throw an exception when attempting to read columnar encrypted Parquet files on the GPU [databricks]

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/5246
Since Spark 3.2, columnar encryption is supported for Parquet tables with Apache Parquet 1.12+

CUDF doesn't support reading columnar encrypted parquet files, the files are only read on the executor side in the tasks so we can't easily fallback to the CPU. 
To work around this, we check for the parquet encryption configs and if we see them we null them out and let the parquet footer filter checks throw. (Caused by: org.apache.parquet.crypto.ParquetCryptoRuntimeException: Trying to read file with encrypted footer. No keys available). The NATIVE reader is a bit different and we can check for the encrypted magic number PARE.   We attempt to catch those failures and then throw our own more useful message.

Message thrown looks like:
`Caused by: java.lang.RuntimeException: The GPU does not support reading encrypted Parquet files. To read encrypted or columnar encrypted files, disable the GPU Parquet reader via spark.rapids.sql.format.parquet.read.enabled.`

I manually tested v1/v2 readers, Native and java footer types and then perfile, coalescing and multithreaded readers.  see https://spark.apache.org/docs/latest/sql-data-sources-parquet.html#columnar-encryption for information used on testing. I wrote a columnar encrypted file using the CPU and then tested reading via GPU and made sure it threw.

Added automated test that relies on the parquet hadoop tests jar.  So we download it and rely on it being there for those versions that have parquet > 1.12.


tgravescs:
build

tgravescs:
build

tgravescs:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuParquetScan.scala
```diff
@@ -616,32 +625,47 @@ private case class GpuParquetFileFilterHandler(@transient sqlConf: SQLConf) exte
...
-              tableFooter.serializeThriftFile()
-          }
-          withResource(serialized) { serialized =>
-            withResource(new NvtxRange("readFilteredFooter", NvtxColor.YELLOW)) { _ =>
-              val inputFile = new HMBInputFile(serialized)
+      // Make sure we aren't trying to read encrypted files. For now, remove the related
+      // parquet confs from the hadoop configuration and try to catch the resulting
+      // exception and print a useful message
+      PARQUET_ENCRYPTION_CONFS.foreach { encryptConf =>
+        if (conf.get(encryptConf) != null) {
```
gerashegalov, line 632: nit: this check is not necessary

gerashegalov (dismissed):
LGTM

tgravescs:
Ok, added a test.  It will run in our regular nightly tests but I filed a follow to support in our integration test runs taht will need to download and ship the parquet hadoop tests jar.  I had to add a version check for CDH because it ships with older parquest version.

tgravescs:
build

jlowe (approved):

tgravescs:
thats is weird the build for this passed in jenkins... kicking again

tgravescs:
build

PR #5766, opened 2022-06-07
Fix the overflow of container type when casting floats to decimal

### LINKED ISSUE #5765: [BUG] Container decimal overflow when casting float/double to decimal
**Describe the bug**
The code of casting float/double to decimal creates a container decimal for the HALF_UP round: [link](https://github.com/NVIDIA/spark-rapids/blob/branch-22.08/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala#L1470). However, the precision of the container decimal should have increased along with the scale. The sole increase of scale may lead to decimal overflow if the current precision already hits the max precision of cuDF decimal (9 for Decimal32; 18 for Decimal 64). 
```scala
      val casted = if (targetType.getDecimalMaxPrecision == dt.scale) {
        checked.castTo(targetType)
      } else {
        val containerType = DecimalUtils.createDecimalType(dt.precision, dt.scale + 1)
        withResource(checked.castTo(containerType)) { container =>
          container.round(dt.scale, cudf.RoundMode.HALF_UP)
        }
      }
```

```scala
Seq[Double](459.288333).toDF("a").write.parquet("tmp123")
spark.read.parquet("tmp123").createOrReplaceTempView("tbl")
spark.sql("SELECT CAST(a AS decimal(9, 6)) FROM tbl") // returns -214.748364
spark.sql("SELECT CAST(CAST(a AS decimal(10, 6)) AS decimal(9, 6))  FROM tbl") // returns 459.288333
```

sperlingxx:
Fixes #5765 

This PR is to fix the potential overflow when casting float/double to decimal. The overflow occurs on the container decimal for HALF_UP round. The precision of the container decimal should increase along with the scale in case of cornor cases like #5765. 
Notice: This change will lead to a tiny precision loss of casting to decimal with precision 38 (max precison of DECIMAL 128).


sperlingxx:
build

ttnghia:
I assume that this should go to 22.06 release.

andygrove (approved):
The change looks reasonable to me but I have not worked on the decimal implementation and am not an expert in this area.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
```diff
@@ -1464,12 +1464,18 @@ object GpuCast extends Arm {
...
-        val containerType = DecimalUtils.createDecimalType(dt.precision, dt.scale + 1)
+        // Increase precision by one along with scale in case of overflow, which may lead to
+        // the upcast of cuDF decimal type. If precision already hits the max precision, it is safe
+        // to increase the scale solely because we have checked and replaced out of range values.
+        val containerType = DecimalUtils.createDecimalType(
+          dt.precision + 1 min DType.DECIMAL128_MAX_PRECISION, dt.scale + 1)
         withResource(checked.castTo(containerType)) { container =>
-          container.round(dt.scale, cudf.RoundMode.HALF_UP)
+          withResource(container.round(dt.scale, cudf.RoundMode.HALF_UP)) { rd =>
+            rd.castTo(targetType)
```
jlowe, line 1477: Nit; Doesn't this cast end up creating a redundant copy of the column in many (all?) cases?  Wondering if we should check if the resulting type from `round` is already the desired cudf type.  I would expect cudf's `round` to deterministically produce a result type based on the input cast type which makes me wonder why the cast is necessary here.

sperlingxx: @jlowe The cast here is for cases that underlying cuDF decimal type got promoted as precision + 1 (from Decimal32 to Decimal64). For case like that, I think we need to convert back to the original cuDF decimal type, to keep align with the precision.

sperlingxx:
Conducted a force push as rebasing to 22.06

sperlingxx:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
jlowe: We are pre-casting to the promoted precision but asking cudf to round to the originally desired precision.  Do we know for certain that cudf does not already put the result in the appropriate type as part of the round implementation?  Even if it does not, it is not always the case that the the promotion cast actually did anything (e.g.: it was DECIMAL32 before and after, because precision+1 didn't actually need DECIMAL64).

This isn't must-fix for proper functionality, more of a performance optimization for what may be a common case.

jlowe (approved):
I'm OK with it as-is, but it would be good to have a followup for what appears to be an unnecessary post-round cast in some (all?) cases.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuCast.scala
ttnghia: > Do we know for certain that cudf does not already put the result in the appropriate type as part of the round implementation?

I'm not sure if I understand your question correctly. In cudf, floating-point is just shifted (mul/div) by the `scale` value then rounded (towards zero) to integer.
https://github.com/rapidsai/cudf/blob/fae22213537a003c74e2da9feac1bda38f562f6f/cpp/include/cudf/fixed_point/fixed_point.hpp#L230

ttnghia: Filed: https://github.com/rapidsai/cudf/issues/11079

ttnghia:
Regarding to the reason behind this bug, I repost an explain from my cudf issue (https://github.com/rapidsai/cudf/issues/11079#issuecomment-1150345234):
```
(reposting this message from slack)

There is a long history behind this and truncation is definitely the desired and intentional behaviour.

Here are the main motivations for truncation:

Consistency with CNL. CNL is being proposed for the standard library and consistency with standard types is a goal of libcudf. Note that (IIRC) every other C++ fixed-point library has made the same choice.
Relying on the behaviour of integer types in C++ greatly simplifies certain binary operations like divison for fixed-point
We can (and have) provided rounding functionality through the cudf::round API so on the Spark side of things you are able to get whatever behaviour you want
Note that at one point, we actually did have rounding functionality baked into floating point construction for fixed-point. However, it wasn't actually to address the desire to be more "accurate" as @ttnghia is pointing out, it was to try and address inherent issues in floating point (such as 1.001 not being representable by floating point, meaning if you construct a fixed-point with scale -3, you end up with 1 , missing the .001). @harrism and I had many discussions about this and both agreed that "ideally" fixed-point should be able to avoid this at the end of the day. However, it ended up presenting too many issues / complications and not actually comprehensively fixing all cases. And note, trying to be "better" was not consistent with what CNL and other C++ fixed-point libraries do. So we decided at a certain point to just accept the inherent flaws of floating point and follow CNL. This was done in the following small PR: https://github.com/rapidsai/cudf/pull/6544 If you take a look at the unit tests, it is very clear how the behaviour changed by looking at the before and after. Furthermore, there was a bunch of "opposition research" done at one point in to what other C++ fixed-point libraries do. Our goal is not to be similar to Julia or Go or Python, but to be a generic C++ backend that provides all the tools to do whatever you want in the front end language (like cuDF or Spark-RAPIDS).
```

Issue #5678, opened 2022-05-27
[BUG] Inconsistency between the time zone in the fallback reason and the actual time zone checked in RapidsMeta.checkTImeZoneId


gerashegalov:
**Describe the bug**
`RapidsMeta.checkTImeZoneId` compares the default time zone of the JVM to UTC in order to fallback on CPU, but quotes the session time zone as the offending actual

**Steps/Code to reproduce bug**
1. Start a Spark REPL with an unsupported i.e., non UTC default time zone such as `GMT-8`:
```bash
pyspark --driver-java-options -Duser.timezone=GMT-8 \
  --packages com.nvidia:rapids-4-spark_2.12:22.04.0 \
  --conf spark.rapids.sql.enabled=true \
  --conf spark.plugins=com.nvidia.spark.SQLPlugin \
  --conf spark.rapids.sql.explain=ALL
```
2. Set session's time zone to UTC
```python
>>> spark.conf.set('spark.sql.session.timeZone', 'UTC')
```

3. Execute a query with a TimestampType
And observe a confusing message `Only UTC zone id is supported. Actual zone id: UTC`
```python
>>> spark.createDataFrame([ ['2022-06-01 07:45'] ], 'a string').selectExpr('hour(a)').collect()
22/05/27 08:23:58 WARN GpuOverrides: 
!Exec <ProjectExec> cannot run on GPU because not all expressions can be replaced
  @Expression <Alias> hour(cast(a#4 as timestamp), Some(UTC)) AS hour(a)#6 could run on GPU
    !Expression <Hour> hour(cast(a#4 as timestamp), Some(UTC)) cannot run on GPU because input expression Cast cast(a#4 as timestamp) (TimestampType is not supported when the JVM system timezone is set to GMT-08:00. Set the timezone to UTC to enable TimestampType support); Only UTC zone id is supported. Actual zone id: UTC
      !Expression <Cast> cast(a#4 as timestamp) cannot run on GPU because the GPU only supports a subset of formats when casting strings to timestamps. Refer to the CAST documentation for more details. To enable this operation on the GPU, set spark.rapids.sql.castStringToTimestamp.enabled to true.; Cast from StringType to TimestampType is not supported; Parsing the full rage of supported years is not supported. If your years are limited to 4 positive digits set spark.rapids.sql.hasExtendedYearValues to false.
        @Expression <AttributeReference> a#4 could run on GPU
[...]

PR #5767, opened 2022-06-07
Fix the time zone check issue

firestarman:
Fixes https://github.com/NVIDIA/spark-rapids/issues/5678

There are two settings for time zone, the Spark session local time zone and JVM's default time zone. And they may be different from each other according to issue #5678. 

With this PR, plugin now requires both of the two settings being equal to UTC to support timestamp type.


firestarman:
For early review

tgravescs:
did you mean this to be against 22.06?

firestarman:
> did you mean this to be against 22.06? Unless specifically requested this should go against 22.08

It is a bug fix, so I thought it should go into 06.

jlowe:
build

gerashegalov:
The change looks good as a narrow fix. 

However, the logic for timezone error handling in the plugin initialization is not quire clear  
- first we stash away the initial value of the default  timezone on the driver  https://github.com/NVIDIA/spark-rapids/blob/0b2eeb1589409a2f3813ac8a80a3ea24d57cdc0f/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala#L141 . 
- We don't throw an exception on the driver, if it's not a supported time zone, which I think is OK because it can still be changed for the JVM using java.util.TimeZone.setDefault
- Then on the executor side plugin init we throw an exception on discrepancy between the driver and executor default timezones but only if the driver timezone is UTC https://github.com/NVIDIA/spark-rapids/blob/aba390e2909fc9e753a128095d97bc46e774e8ec/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala#L221  This does not make sense to me. Either always throw when timezones are different. Or adjust the executor timezone with TimeZone.setDefault(TimeZone.getTimeZone(driverTimeZone)) unless it is explicitly overridden with  user.timezone in executor Java options

We need to file the issue to clarify this if it's beyond the current PR.

gerashegalov (approved):
LGTM with caveats regarding Plugin init checks

firestarman:
Personally the current time zone checks in Plugin init are OK. (Please correct me if I am wrong)

At the init stage on the driver, Plugin does not know whether the time zone will be used. So it is not necessary to check the time zone. And this check will be executed during the related operators overriding later. 

On executors, the time zone is required to be equal to the driver's only for the time related GPU operators, I think. If time zone is not supported, no time related GPU operators exists (all fall back to CPU), then no need to check the difference of the time zones.

firestarman:
I am going to merge this, if any issue, we can follow up in new issues.

gerashegalov:
> > We need to file the issue to clarify this if it's beyond the current PR.
> 
> Personally the current time zone checks in Plugin init are OK. (Please correct me if I am wrong)
> 
> At the init stage on the driver, Plugin does not know whether the time zone will be used. So it is not necessary to check the time zone here. And this check will be executed during the time related operators' overriding process later.
> 

> On executors, the time zone is required to be equal to the driver's only for the time related GPU operators, I think. If time zone is not supported, no time related GPU operators exist (all fall back to CPU), then no need to check the difference of the time zones.

So looks like you are arguing for dropping the check https://github.com/NVIDIA/spark-rapids/blob/aba390e2909fc9e753a128095d97bc46e774e8ec/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala#L221

because it's an init-time check.

Also again the default timezone of the driver as of init time may be different at the query time. I'll file an issue to clarify these limitations.

PR #5768, opened 2022-06-07
Support MMyyyy date/timestamp format

### LINKED ISSUE #5679: [FEA] Support MMyyyy date/timestamp format
I wish we can support MMyyyy date/timestamp format without the need to enable `spark.rapids.sql.incompatibleDateFormats.enabled=true`.

The reason is we may want to directly use Mortgage raw data as the input and inside it, the date related columns are all in MMyyyy format. 
This format should also be common as well.

res-life:
Closes #5679

- Support MMyyyy date/timestamp format
- Update documents according to the `CORRECTED_COMPATIBLE_FORMATS`

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
@@ -403,51 +403,53 @@ object GpuToTimestamp extends Arm {
   // or EXCEPTION. It is possible that other formats may be supported but these are the only
   // ones that we have tests for.
   val CORRECTED_COMPATIBLE_FORMATS = Map(
-    "yyyy-MM-dd" -> ParseFormatMeta('-', isTimestamp = false,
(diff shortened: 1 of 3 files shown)


res-life:
build

res-life:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala
```diff
@@ -555,12 +557,12 @@ object GpuToTimestamp extends Arm {
...
     // regex rules are written with '-' so we need to replace '-' with '/' here as necessary
     val rulesWithSeparator = format.separator match {
-      case '/' =>
+      case Some('/') =>
         regexReplaceRules.map {
           case RegexReplace(pattern, backref) =>
             RegexReplace(pattern.replace('-', '/'), backref.replace('-', '/'))
         }
-      case '-' =>
+      case Some('-') | Some(_) | None =>
```
andygrove, line 565: It looks like this is an exhaustive check for all other cases so could be simplified:

```suggestion
      case _ =>
```

andygrove (approved):

PR #5776, opened 2022-06-07
Enable regular expression support based on whether UTF-8 is in the current locale

### LINKED ISSUE #5629: [BUG] regexp unicode tests require LANG=en_US.UTF-8 to pass
This is a followup from a [discussion in #5627](https://github.com/NVIDIA/spark-rapids/pull/5627#discussion_r881662266).  Currently we're explicitly setting `LANG=en_US.UTF-8` in the CI environment to get some regexp tests to pass.  Users who may not have that same environment setup will also encounter these test failures, and we need to consider what to do in those situations.

NVnavkumar:
Fixes #5629.

This relaxes the requirement of needing the `LANG=en_US.UTF-8` environment variable to be set by switching the requirement to checking the default Charset that the JVM is using (which can be set via the `LANG` environment indirectly anyways. This allows users to have different languages in their locale but using the UTF-8 character set, which is much easier requirement to fulfill for most users. 

This also splits off the regular expression Python integration tests into a separate test file that checks for unicode support, and another set of regular expression Python integration tests that should fall back to the CPU when UTF-8 is not detected in the current locale. Also, a couple of fuzz tests for full unicode input are provided as well.


NVnavkumar:
build

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -993,16 +1022,21 @@ class FuzzRegExp(suggestedChars: String, skipKnownIssues: Boolean = true) {
...
-      () => RegexEscaped('B'),
       () => RegexEscaped('A'),
       () => RegexEscaped('G'),
       () => RegexEscaped('Z'),
       () => RegexEscaped('z')
     )
+    val generators = if (skipKnownIssues) {
+      baseGenerators
+    } else {
+      baseGenerators ++ Seq[() => RegexAST](
```
andygrove, line 1036: Could you add a link to the known issue here

tests/src/test/scala/com/nvidia/spark/rapids/RegularExpressionTranspilerSuite.scala
```diff
@@ -580,6 +589,22 @@ class RegularExpressionTranspilerSuite extends FunSuite with Arm {
       RegexReplaceMode)
   }
 
+  test("AST fuzz test - regexp_find - full unicode input") {
+    assume(isUnicodeEnabled())
```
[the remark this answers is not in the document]
NVnavkumar: We have 2 sets of integration tests `regexp_test.py` and `regexp_no_unicode_test.py`. The second file checks for fallback when UTF-8 is not in the locale.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -1025,6 +1025,14 @@ class CudfRegexTranspiler(mode: RegexMode) {
         case 'b' | 'B' if mode == RegexSplitMode =>
           // see https://github.com/NVIDIA/spark-rapids/issues/5478
           throw new RegexUnsupportedException("word boundaries are not supported in split mode")
+        case 'b' | 'B' =>
+          previous match {
+            case Some(RegexEscaped(ch)) if "DWS".contains(ch) =>
```
anthony-chang, line 1030: Do you know if this has to do word boundaries around negated character classes, or is it specifically `\D`, `\S`, `\W`?

NVnavkumar: Good catch, pushing another fix

NVnavkumar:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
anthony-chang: Oh I just remembered, in that case this should also include `\H` and `\V`

NVnavkumar:
build

NVnavkumar:
build

sameerz:
build

andygrove (commented):
LGTM.

NVnavkumar:
build

anthony-chang (approved):

jenkins/spark-nightly-build.sh
```diff
@@ -21,6 +21,8 @@ set -ex
 
 ## export 'M2DIR' so that shims can get the correct Spark dependency info
 export M2DIR=${M2DIR:-"$WORKSPACE/.m2"}
+## export 'LC_ALL' to set locale with UTF-8 so regular expressions are enabled
+export LC_ALL="en_US.UTF-8"
```
pxLi, line 25: this config is for all tests (nightly in this case, and also the update in premerge script)

one question is,
are we going to enable UTF-8 for all our integrations tests?
if yes, we should doc this somewhere (like https://github.com/NVIDIA/spark-rapids/tree/branch-22.08/integration_tests#prerequisities)to tell all our developers that our IT require UTF-8 as the context.
if no and only regex text need it, this should only be set as specific test-level context (inside the test file) instead of for all tests in script IMHO

jenkins/spark-premerge-build.sh
```diff
@@ -114,6 +116,8 @@ rapids_shuffle_smoke_test() {
 
 ci_2() {
     echo "Run premerge ci 2 testings..."
+    # export 'LC_ALL' to set locale with UTF-8 so regular expressions are enabled
+    export LC_ALL="en_US.UTF-8"
```
pxLi, line 120: same as above

NVnavkumar: As far I know, to do this in `pytest.ini`, we would need this plugin for pytest installed in the CI env  (https://pypi.org/project/pytest-env/). Do we have this plugin available to us in CI?

NVnavkumar: The underlying issue here is setting the locale, which is controlled by the shell. It's simple to override as an environment variable passed into the tests, but I don't know how to set that from Python, and use a mark for those tests.

jenkins/spark-nightly-build.sh
NVnavkumar: Hmm, I would think (if not now then eventually) string integration tests (non-regular expression tests) might require UTF-8 support as well. For now, they don't seem to have this depedency.

I have noticed this pattern in the premerge script, but did not see an equivalent in the nightly script:

```
INCLUDE_SPARK_AVRO_JAR=true TEST='avro_test.py' ./integration_tests/run_pyspark_from_build.sh
```

Regular expressions have 2 sets of tests, unit tests (Scalatest) and integration tests (pytest).  The Unicode-specific unit tests are skipped when UTF-8 is not in the environment (and therefore would run and pass), but the functionality itself would only be enabled when UTF-8 is in the environment (so these tests are testing an invalid environment in that case).  The integration tests in `regexp_test.py` could handled similarly to the above pattern (which is also used for for `regexp_no_unicode_test.py`). 

```
LC_ALL="en_US.UTF-8" TEST="regexp_test.py" ./integration_tests/run_pyspark_from_build.sh
```

This case though is only done in the premerge script. How would this be handled in the nightly script? And how would we handle the scala tests?

pxLi: > How would this be handled in the nightly script? And how would we handle the scala tests?

We could have multiple ways to do so like separate the CI in different stages/jobs, like use ENV/Arg to control only run specific tests. But I would like to know more inputs if below is already the plan cc @jlowe @revans2 @tgravescs 
```
Hmm, I would think (if not now then eventually) string integration tests (non-regular expression tests) might require UTF-8 support as well.
```

jenkins/spark-premerge-build.sh
pxLi: pytest.ini is supported in pytest by default as far as I know. We are using it in our pytest for a long time.

> override as an environment variable passed into the tests

This is good to me

NVnavkumar:
build

NVnavkumar:
build

NVnavkumar:
build

jenkins/spark-premerge-build.sh
```diff
@@ -124,6 +126,10 @@ ci_2() {
     TEST='not conditionals_test and not window_function_test and not struct_test and not time_window_test' \
       ./integration_tests/run_pyspark_from_build.sh
     INCLUDE_SPARK_AVRO_JAR=true TEST='avro_test.py' ./integration_tests/run_pyspark_from_build.sh
+    # export 'LC_ALL' to set locale with UTF-8 so regular expressions are enabled
+    LC_ALL="en_US.UTF-8" TEST="regexp_test.py" ./integration_tests/run_pyspark_from_build.sh
+    # export 'LC_ALL' to set locale without UTF-8 so regular expressions are disabled to test fallback
+    LC_ALL="en_US.iso88591" TEST="regexp_no_unicode_test.py" ./integration_tests/run_pyspark_from_build.sh
```
pxLi, line 132: looks like this one should already be covered in normal tests run w/o explicitly setting LC_ALL?

jenkins/spark-nightly-build.sh
```diff
@@ -95,6 +95,12 @@ for buildver in "${SPARK_SHIM_VERSIONS[@]:1}"; do
     $MVN -U -B clean install -pl '!tools' $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR \
         -Dcuda.version=$CUDA_CLASSIFIER \
         -Dbuildver="${buildver}"
+    # enable UTF-8 and run regular expression tests
+    env LC_ALL="en_US.UTF-8" $MVN verify -pl '!tools' $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR \
```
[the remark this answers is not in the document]
NVnavkumar: I tried this with `mvn test` but it would never run the integration test that are also necessary to run. This way it runs the 3 scalatests and the integration test as well.

NVnavkumar:
build

jenkins/spark-nightly-build.sh
```diff
@@ -111,6 +117,13 @@ $MVN -B clean install -pl '!tools' \
     -Dmaven.repo.local=$M2DIR \
     -Dcuda.version=$CUDA_CLASSIFIER
 
+# enable UTF-8 and run regular expression tests
+env LC_ALL="en_US.UTF-8" $MVN verify -pl '!tools' $MVN_URM_MIRROR -Dmaven.repo.local=$M2DIR \
```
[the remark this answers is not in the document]
NVnavkumar: addressed `wildcardSuites` issue in last commit

jenkins/spark-premerge-build.sh
NVnavkumar: removed this redundant last line in last commit (it was artifact of the previous code organization)

NVnavkumar:
build

NVnavkumar:
build

jenkins/spark-nightly-build.sh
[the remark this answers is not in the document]
pxLi: Actually we run integration test in https://github.com/NVIDIA/spark-rapids/blob/branch-22.08/jenkins/spark-tests.sh as downstream testing of our nightly build to cover all test scenarios (spark releases+OS, over 10 pipelines)

NVnavkumar:
build

pxLi (approved):
LGTM. Let's have a followup for nightly build and tests later

pxLi:
all premerge CI is blocked by https://github.com/NVIDIA/spark-rapids/issues/6003

pxLi:
build

jlowe (approved):

anthony-chang (approved):

Issue #5196, opened 2022-04-11
[FEA] Plugin should throw same arithmetic exceptions as Spark


andygrove:
**Is your feature request related to a problem? Please describe.**
This is a follow-on to https://github.com/NVIDIA/spark-rapids/issues/5182 where we made some exception checks less specific due to changes in Spark 3.3 causing test failures.

We really should throw the same arithmetic exceptions as Spark and these vary by Spark version.

**Describe the solution you'd like**
Update `arithmetic_ops_test.py` and `logic_test.py` to check for specific exceptions `java.lang.ArithmeticException` vs `org.apache.spark.SparkArithmeticException`.

**Describe alternatives you've considered**
None

**Additional context**
This is a checklist of all the tests we need to update.
- [x] `test_mod_pmod_by_zero` (#5354)
- [x] `test_unary_minus_ansi_overflow` (#5783)
- [x] `test_abs_ansi_overflow` (#5783)
- [x] `test_floor_ceil_overflow` (#5915)
- [x] `test_div_by_zero_ansi` (#5354)
- [x] `test_div_overflow_exception_when_ansi` (#5354)
- [x] `test_unary_minus_ansi_overflow_day_time_interval`(#5783)
- [x] `test_abs_ansi_overflow_day_time_interval`(#5783)
- [x] `test_add_overflow_with_ansi_enabled_day_time_interval`(#5949)
- [x] `test_subtraction_overflow_with_ansi_enabled_day_time_interval`(#5949)
- [x] `test_day_time_interval_division_overflow`(#6019)
- [x] `test_logical_with_side_effect`(#5949)
- [x] `test_add_overflow_with_ansi_enabled` (#6071)
- [x] `test_subtraction_overflow_with_ansi_enabled`(#6071)

HaoYang670:
Our plugin currently throw `java.lang.ArithmeticException`. Should we update it to throw `SparkArithmeticException`?
```scala
  def divByZeroError(): Nothing = {
    throw new ArithmeticException("divide by zero")
  }

  def divOverflowError(): Nothing = {
    throw new ArithmeticException("Overflow in integral divide.")
  }
```
https://github.com/NVIDIA/spark-rapids/blob/branch-22.06/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/arithmetic.scala#L704-L710


Maybe we should put these functions into the shim layers.

jlowe:
> Should we update it to throw SparkArithmeticException?

[...]

PR #5783, opened 2022-06-08
`assertMinValueOverflow` should throw same type of exception as Spark

HaoYang670:
Related to #5196 

## Changes in the PR
1. Throw same type of exception as Spark in `assertMinValueOverflow`. For 
For Spark 3.1 and 3.2, `java.lang.ArithmeticException` is thrown.
For Spark 3.3 and later versions, `SparkArithmeticException` is thrown: 
https://github.com/apache/spark/blob/branch-3.3/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/MathUtils.scala#L71-L76
https://github.com/apache/spark/blob/branch-3.3/sql/catalyst/src/main/scala/org/apache/spark/sql/errors/QueryExecutionErrors.scala#L492-L503
2. Update `test_unary_minus_ansi_overflow`, `test_unary_minus_ansi_overflow_day_time_interval`, `test_abs_ansi_overflow_day_time_interval` and `test_abs_ansi_overflow` to check the type of exceptions.

### DIFF
--- a/integration_tests/src/main/python/arithmetic_ops_test.py
+++ b/integration_tests/src/main/python/arithmetic_ops_test.py
@@ -341,10 +341,16 @@ def test_unary_minus_ansi_no_overflow(data_gen):
     (ShortType(), SHORT_MIN),
     (ByteType(), BYTE_MIN)], ids=idfn)
 def test_unary_minus_ansi_overflow(data_type, value):
+    """
+    We don't check the error messages because they are different for CPU and GPU
+    CPU: {name of the data type} overflow
+    GPU: One or more rows overflow for {name of the operation} operation
+    """
     assert_gpu_and_cpu_error(
             df_fun=lambda spark: _get_overflow_df(spark, [value], data_type, '-a').collect(),
             conf=ansi_enabled_conf,
-            error_message='ArithmeticException')
+            error_message='java.lang.ArithmeticException' if is_before_spark_330() else \
+                          'org.apache.spark.SparkArithmeticException')
 
 # This just ends up being a pass through.  There is no good way to force
 # a unary positive into a plan, because it gets optimized out, but this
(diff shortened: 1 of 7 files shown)


HaoYang670:
build

HaoYang670:
build

HaoYang670:
Met an odd error: https://github.com/NVIDIA/spark-rapids/runs/6914487613?check_suite_focus=true#step:2:36294, which is not related to this PR.

HaoYang670:
build

HaoYang670:
build

integration_tests/src/main/python/arithmetic_ops_test.py
```diff
@@ -378,7 +384,8 @@ def test_abs_ansi_overflow(data_type, value):
     assert_gpu_and_cpu_error(
             df_fun=lambda spark: _get_overflow_df(spark, [value], data_type, 'abs(a)').collect(),
             conf=ansi_enabled_conf,
-            error_message='ArithmeticException')
```
revans2, line 381: Do we need/want a similar "We don't check the error message" comment on all of these?

revans2 (approved):
Just a nit about comments. What would it take to actually output an error message that matches what Spark does? Performance is not really that big of a deal because we already know that we are going to fail so doing some more computation at that point is probably not that big of a deal.

integration_tests/src/main/python/arithmetic_ops_test.py
HaoYang670: Updated!

HaoYang670:
build

HaoYang670:
> What would it take to actually output an error message that matches what Spark does?

No much work I guess. But personally, I prefer the error message that our plugin gives, because it tells users which operation is error.

revans2 (approved):

Issue #5769, opened 2022-06-07
[BUG] arithmetic ops tests failing on Spark 3.3.0


jlowe:
Many arithmetic ops tests failed in the recent Spark 3.3.0 nightly tests against 22.06:
```
10:24:14  =========================== short test summary info ============================
10:24:14  FAILED ../../src/main/python/arithmetic_ops_test.py::test_mod_pmod_by_zero[pmod(a, cast(0 as {}))-Decimal(7,3)]
10:24:14  FAILED ../../src/main/python/arithmetic_ops_test.py::test_mod_pmod_by_zero[pmod(a, cast(0 as {}))-Decimal(12,2)]
10:24:14  FAILED ../../src/main/python/arithmetic_ops_test.py::test_mod_pmod_by_zero[pmod(a, cast(0 as {}))-Decimal(18,0)]
10:24:14  FAILED ../../src/main/python/arithmetic_ops_test.py::test_mod_pmod_by_zero[pmod(a, cast(0 as {}))-Decimal(20,2)]
10:24:14  FAILED ../../src/main/python/arithmetic_ops_test.py::test_mod_pmod_by_zero[pmod(a, cast(0 as {}))-Decimal(30,2)]
10:24:14  FAILED ../../src/main/python/arithmetic_ops_test.py::test_mod_pmod_by_zero[pmod(a, cast(0 as {}))-Decimal(36,5)]
10:24:14  FAILED ../../src/main/python/arithmetic_ops_test.py::test_mod_pmod_by_zero[pmod(a, cast(0 as {}))-Decimal(38,0)]
10:24:14  FAILED ../../src/main/python/arithmetic_ops_test.py::test_mod_pmod_by_zero[pmod(a, cast(0 as {}))-Decimal(38,10)]
10:24:14  FAILED ../../src/main/python/arithmetic_ops_test.py::test_mod_pmod_by_zero[pmod(cast(-12 as {}), cast(0 as {}))-Decimal(7,3)]
10:24:14  FAILED ../../src/main/python/arithmetic_ops_test.py::test_mod_pmod_by_zero[pmod(cast(-12 as {}), cast(0 as {}))-Decimal(12,2)]
10:24:14  FAILED ../../src/main/python/arithmetic_ops_test.py::test_mod_pmod_by_zero[pmod(cast(-12 as {}), cast(0 as {}))-Decimal(18,0)]
[...]

PR #5786, opened 2022-06-08
Fix rounds over decimal in Spark 330+

sperlingxx:
[SPARK-39226](https://github.com/apache/spark/commit/ee0aecca05af9b0cb256fd81a78430958a09d19f) changed the way to compute output precision and scale for round-like functions, which fails the round-like tests: #5769. 

This PR passes the datatype of round-like functions directly to GPU overrides, so as to adapt different Spark versions. The shim method approach doesn't work very well in this case because the computition of decimal precision and scale requiring member variables of rounds-like functions.
This PR also modifies the round logic over decimal type to adapt both original and new manner of datatype. In addition, some test cases are skipped in Spark 330+ since they are not supported by Spark any more.

### DIFF
--- a/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
+++ b/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
@@ -21,9 +21,8 @@ import java.io.Serializable
 import ai.rapids.cudf._
 import ai.rapids.cudf.ast.BinaryOperator
 import com.nvidia.spark.rapids._
-import com.nvidia.spark.rapids.RapidsPluginImplicits.ReallyAGpuExpression
 
-import org.apache.spark.sql.catalyst.expressions.{EmptyRow, Expression, ImplicitCastInputTypes}
+import org.apache.spark.sql.catalyst.expressions.{Expression, ImplicitCastInputTypes}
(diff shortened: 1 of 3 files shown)


sperlingxx:
build

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/mathExpressions.scala
```diff
@@ -590,9 +573,19 @@ abstract class GpuRoundBase(child: Expression, scale: Expression) extends GpuBin
...
+        // Only needs to perform round when required scale < input scale
+        val rounded = if (scaleVal < s) {
+          lhsValue.round(scaleVal, roundMode)
+        } else {
+          lhsValue.incRefCount()
+        }
+        withResource(rounded) { _ =>
+          // Fit the output datatype
+          rounded.castTo(
+            DecimalUtil.createCudfDecimal(dataType.asInstanceOf[DecimalType]))
```
jlowe, line 587: This cast seems unnecessary in most cases, may be worth checking if the desired cudf type is already the type computed to avoid an unnecessary copy.

jlowe (approved):
Like #5766, this has a post-round cast that probably can be avoided in most cases.  Not must-fix but would be good to have a followup issue to investigate if not addressed here.

nartal1 (approved):

PR #5795, opened 2022-06-08
Update FAQ to include local jar deployment via extraClassPath [skip ci]

razajafri:
As discussed in the meeting. We want to include the `--extraClassPath` conf in the doc to guide the users

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -520,9 +520,9 @@ If you are getting a warning `Avro library not found by the RAPIDS plugin.` or i
 Spark job by using the `--jars` or `--packages` option followed by the file path or maven path to 
 RAPIDS jar since that is the preferred way to run RAPIDS accelerator. 
 
-NOTE: You could set `spark.driver.extraClassPath` and `spark.executor.extraClassPath` followed by the path to Avro and 
-the RAPIDS Plugin jar but one important thing to keep in mind is that never mix the `jars/packages` 
-and extraClassPath confs
+Note, you can add locally installed jars for external packages such as Avro Data Sources and the RAPIDS Accelerator jars via  `spark.driver.extraClassPath` (--driver-class-path in the client mode) on the driver side, and `spark.executor.extraClassPath` on the executor side. However, you should not mix the deploy methods for either of the external modules.  Either deploy both Spark Avro and RAPIDS Accelerator jars as local jars via `extraClassPath` settings or use the `--jars` or `--packages` options.
+
+As a consequence, per  Issue #5796, if you also use the RAPIDS Shuffle Manager, your deployment option may be limited to the extraClassPath method.
 
 ### What is the default RMM pool allocator?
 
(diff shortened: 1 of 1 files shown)


razajafri:
@gerashegalov

docs/FAQ.md
```diff
@@ -520,6 +520,9 @@ If you are getting a warning `Avro library not found by the RAPIDS plugin.` or i
 Spark job by using the `--jars` or `--packages` option followed by the file path or maven path to 
 RAPIDS jar since that is the preferred way to run RAPIDS accelerator. 
 
+NOTE: You could use `--extraClassPath` followed by the path to Avro and the RAPIDS Plugin jar but one
```
NVnavkumar, line 523: Is this supposed to be `--extraClassPath` or `--conf spark.driver.extraClassPath=` or `--conf spark.executor.extraClassPath=`.  I've never been able to find a reference to `--extraClassPath` in the Spark online documentation anywhere, but `--jars` and `--packages` is definitely there.

gerashegalov: correct, the confs are 

spark.driver.extraClassPath
spark.executor.extraClassPath (it's [discouraged](https://spark.apache.org/docs/latest/configuration.html))  but we need it if we don't want to recommend patching SPARK_HOME/jars

In the client mode REPL the conf setting is too late, and therefore there exists `--driver-class-path`

razajafri: >  I've never been able to find a reference to --extraClassPath in the Spark online documentation anywhere

Sorry, that's my fault. I could swear I saw that somewhere but obviously not

docs/FAQ.md
```diff
@@ -520,6 +520,10 @@ If you are getting a warning `Avro library not found by the RAPIDS plugin.` or i
 Spark job by using the `--jars` or `--packages` option followed by the file path or maven path to 
 RAPIDS jar since that is the preferred way to run RAPIDS accelerator. 
 
+NOTE: You could set `spark.driver.extraClassPath` and `spark.executor.extraClassPath` followed by the path to Avro and 
+the RAPIDS Plugin jar but one important thing to keep in mind is that never mix the `jars/packages` 
+and extraClassPath confs
```
gerashegalov, line 525:
```suggestion
Note, you can add locally installed jars for external packages such as Avro Data Sources and the RAPIDS Accelerator jars via  `spark.driver.extraClassPath` (--driver-class-path in the client mode) on the driver side, and `spark.executor.extraClassPath` on the executor side. However, you should not mix the deploy methods for either of the external modules.  Either deploy both Spark Avro and RAPIDS Accelerator jars as local jars via `extraClassPath` settings or use the `--jars` or `--packages` options.

As a consequence, per  Issue #5796, if you also use the RAPIDS Shuffle Manager, your deployment option may be limited to the extraClassPath method.
```

gerashegalov (changes requested):

gerashegalov (approved):

gerashegalov:
build

Issue #5797, opened 2022-06-09
[FEA] Support RoundCeil and RoundFloor when scale is zero


nartal1:
**Is your feature request related to a problem? Please describe.**
We have currently disabled running RoundCeil and RoundFloor expressions on GPU and fall back to CPU for all inputs. We can enable it for scale=0 as it would call GpuCeil/GpuFloor internally. There is a bug where it produces incorrect results when the input type is `DecimalType(precision = 18, scale = 0)` .   
We should enable these integration tests to verify:
1. test_floor_scale_zero
2. test_ceil_scale_zero

PR #5798, opened 2022-06-09
Fall back to CPU for RoundCeil and RoundFloor expressions

nartal1:
This contributes to test failures in https://github.com/NVIDIA/spark-rapids/issues/5769 esp test_floor* and test_ceil* tests.
While debugging above failures found another bug where a simple ceil on Decimal type was failing.
```
scala> df.printSchema
root
 |-- value: decimal(18,1) (nullable = true)


scala> df.selectExpr("ceil(value)").collect
22/06/08 17:31:32 WARN GpuOverrides: 
  *Expression <Alias> CEIL(cast(value#35 as decimal(18,1))) AS CEIL(value)#45 will run on GPU
    *Expression <Ceil> CEIL(cast(value#35 as decimal(18,1))) will run on GPU
      *Expression <Cast> cast(value#35 as decimal(18,1)) will run on GPU
  !Exec <FileSourceScanExec> cannot run on GPU because CSV reading is not 100% compatible when reading doubles. To enable it please set spark.rapids.sql.csv.read.double.enabled to true.

res10: Array[org.apache.spark.sql.Row] = Array([68719476737]
```
Looks like it was introduced while working on RoundCeil where the `maxPrecision` was added. Above bug is fixed when the `maxPrecision` calculation is reverted. 
Currently we support RoundCeil/RoundFloor only when `scale=0`. We have a case where the tests fail when DecimalType(18,0) is given as input. So we are falling back to CPU whenever there is RoundCeil and RoundFloor expr in this PR.
Created a follow-on issue to debug above case and enable these expressions -  https://github.com/NVIDIA/spark-rapids/issues/5797


nartal1:
build

tgravescs (approved):

PR #5804, opened 2022-06-13
Support RoundCeil and RoundFloor when scale is zero

### LINKED ISSUE #5797: [FEA] Support RoundCeil and RoundFloor when scale is zero

res-life:
Closes #5797 

### problem
[SPARK-39226](https://github.com/apache/spark/commit/ee0aecca05af9b0cb256fd81a78430958a09d19f) changed the way to compute output precision and scale for `RoundCeil`,  `RoundFloor`  and other round-like operators.

After `SPARK-39226`, `RoundFloor(decimal(18,0) , 0)` and `RoundCeil(decimal(18,0) , 0)` return decimal(19, 0) instead of decimal(18, 0). See the code of Spark `RoundBase.dataType`

**// if `p` is 18, `s` is 0 and `_scale` is 0, return decimal(19, 0)**

```
  override lazy val dataType: DataType = child.dataType match {
    case DecimalType.Fixed(p, s) =>
      // After rounding we may need one more digit in the integral part,
      // e.g. `ceil(9.9, 0)` -> `10`, `ceil(99, -1)` -> `100`.
      val integralLeastNumDigits = p - s + 1
      if (_scale < 0) {
        // negative scale means we need to adjust `-scale` number of digits before the decimal
        // point, which means we need at lease `-scale + 1` digits (after rounding).
        val newPrecision = math.max(integralLeastNumDigits, -_scale + 1)
        // We have to accept the risk of overflow as we can't exceed the max precision.
        DecimalType(math.min(newPrecision, DecimalType.MAX_PRECISION), 0)
      } else {
        val newScale = math.min(s, _scale)
        // We have to accept the risk of overflow as we can't exceed the max precision.
        DecimalType(math.min(integralLeastNumDigits + newScale, 38), newScale)
      }
    case t => t
  }
```

### changes
- Revert commit `https://github.com/NVIDIA/spark-rapids/pull/5798` 
- Use Spark output types for `Floor`, `Ceil`, `RoundFloor` and `RoundCeil` operators.


res-life:
build

res-life:
build

tgravescs (approved):

Issue #5702, opened 2022-06-01
[FEA] set spark.rapids.sql.explain=NOT_ON_GPU by default


viadea:
I wish we can set spark.rapids.sql.explain=NOT_ON_GPU by default so that users do not need to turn it on manually.

This can help new users discover those not-supported-messages from driver log.

sameerz:
Discussed and the initial plan is to enable this feature by default.  We can reconsider in a future release.

PR #5819, opened 2022-06-14
Modify the default value of spark.rapids.sql.explain as NOT_ON_GPU

sinkinben:
Signed-off-by: sinkinben <sinkinben@outlook.com>

1. Set `NOT_ON_GPU` as the default value of "spark.rapids.sql.explain" in `RapidsConf.scala`.
2. See #5702 .


sinkinben:
build

firestarman (approved):

tgravescs (approved):
so the only other thing here is we may want to disable for our tests so the logs don't get much bigger.   thoughts @revans2

GaryShen2008:
@tgravescs Do we need to update the same code in spark2-sql-plugin? 
https://github.com/NVIDIA/spark-rapids/blob/branch-22.08/spark2-sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala#L1279

GaryShen2008:
I simply searched sql.explain in branch-22.08, I think all the places look good. Most of them mentioned to set ALL to get more detailed explain logs.

revans2 (dismissed):
Would like to see getting started updated to talk about it, but that is minor and we can do it if we start to see questions about it.

tgravescs (changes requested):
there is a section on debugging in the getting started docs that references it: https://nvidia.github.io/spark-rapids/docs/get-started/getting-started-on-prem.html#debugging
As far as adding more, it might be nice like Bobby said but I'm ok if we want to wait to see if users complain or its not clear.

Also I just realized that this will break the explain API logic in explainCatalystSQLPlan, where normally the explain API defaults to ALL unless its set in to something other than NONE in spark.rapids.sql.explain. So we need to update that to keep the default of ALL there.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1279,7 +1279,7 @@ object RapidsConf {
       "values are ALL: print everything, NONE: print nothing, NOT_ON_GPU: print only parts of " +
       "a query that did not go on the GPU")
     .stringConf
-    .createWithDefault("NONE")
+    .createWithDefault("NOT_ON_GPU")
```
gerashegalov, line 1282: As an engineer on the spark-rapids team I have the explain=ALL constantly. But as a user, I might be less inclined to see even explain=NOT_ON_GPU output without asking for it. 

I think we either should either keep it off, but maybe advertise it in the https://github.com/NVIDIA/spark-rapids/blob/9defb4cc08ae3853d546ac7e29744067b4c18bcd/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala#L84. Alternatively keep the default but advertise how it can be turned off

tgravescs: I think it was decided based on customer feedback that change to NOT_ON_GPU and advertise how to turn it off.  If this becomes an issues with other customers we could turn off again next release.

gerashegalov: In this case, I suggest to advertise it more prominently in the plugin init phase logPluginMode

GaryShen2008: Hi @gerashegalov, do you mean to print a clear log to tell what spark.rapids.sql.explain is in logPluginMode function?

sinkinben: Is it a good idea to advertise it in the docs `get-started/getting-started.md`?

gerashegalov: > Hi @gerashegalov, do you mean to print a clear log to tell what spark.rapids.sql.explain is in logPluginMode function?

yes, I mean we could print something like:

spark.rapids.sql.explain is set to the default value NOT_ON_GPU to print operations not executed on GPU. Set to ALL to print the complete query plan. Set to NONE to suppress the diagnostics about the query placement on the GPU.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala
```diff
@@ -100,6 +100,12 @@ object RapidsPluginUtils extends Logging {
       "More information is available at https://nvidia.github.io/spark-rapids/docs/FAQ.html#" +
       "automatic-translation-of-scala-udfs-to-apache-spark-operations" )
     }
+
+    logWarning("spark.rapids.sql.explain is set to the default value 'NOT_ON_GPU' to print " +
+        "operations not executed on GPU. Set to 'ALL' to print the complete query plan, " +
+        "display whether each operation is placed on the GPU. Set to 'NONE' to suppress the " +
+        "diagnostics about the query placement on the GPU.")
```
gerashegalov, line 107: I think we should move it inside the if branch `if (conf.isSqlEnabled && conf.isSqlExecuteOnGPU) {` on L85 so it prints only if GPU execution is enabled (default). 

And it should only be printed when the value for spark.rapids.sql.explain implies some output (this is an evolution from my prior review) to make it shorter. 

How about around L85:

```scala
if (conf.isSqlEnabled && conf.isSqlExecuteOnGPU) {
  logWarning("RAPIDS Accelerator is enabled, to disable GPU " +
    s"support set `${RapidsConf.SQL_ENABLED}` to false.")

  if (conf.explain != 'NONE') {
    logWarning("spark.rapids.sql.explain is set to ${conf.explain}. Set to 'NONE' to suppress the " +
        "diagnostics logging about the query placement on the GPU.")
  }
} else ...
```

sinkinben: I have updated the code here. Thanks.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -4142,7 +4142,11 @@ object GpuOverrides extends Logging {
    * GPUs.
    */
   private def explainCatalystSQLPlan(updatedPlan: SparkPlan, conf: RapidsConf): Unit = {
-    val explainSetting = if (conf.shouldExplain) {
+    // Since we set "NOT_ON_GPU" as the default value of spark.rapids.sql.explain, here we keep
+    // "ALL" as default value of "explainSetting", unless spark.rapids.sql.explain is changed
+    // by the user.
+    val explainSetting = if (conf.shouldExplain &&
+      conf.rapidsConfMap.containsKey(RapidsConf.EXPLAIN.key)) {
```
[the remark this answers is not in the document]
sinkinben: Thanks for your comments, I have updated the code. :-D

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -4136,6 +4136,13 @@ object GpuOverrides extends Logging {
     }
   }
 
+  /**
+   * To judge whether "spark.rapids.sql.explain" is explicitly set by the users.
+   */
+  private def isSQLExplainExplicitlySet(conf: RapidsConf) : Boolean = {
```
tgravescs, line 4142: sorry I may not have been clear enough, I meant add this to RapidsConf so that it could potentially be reused and you don't have to reach into rapidsConfMap.  I would make it generic so you pass can pass in any key.


Note, nit should be no space before : -> (conf: RapidsConf): Boolean

sinkinben: @tgravescs Thanks for you patience, hope that this time I got what you mean. I have submitted a new commit to fix this.

docs/get-started/getting-started.md
```diff
@@ -39,6 +39,11 @@ will vary depending on your cluster manager. Here are some example configs:
   - `--conf spark.task.resource.gpu.amount=1`
 - Specify a GPU discovery script (required on YARN and K8S):
   - `--conf spark.executor.resource.gpu.discoveryScript=./getGpusResources.sh`
+- Explain why some operations of a query were not placed on a GPU or not:
+  - `--conf spark.rapids.sql.explain=ALL` will display whether each operation is placed on GPU.
+  - `--conf spark.rapids.sql.explain=NONE` will disable the log of `rapids.sql.explain`.
```
tgravescs, line 44: switch the order of None and NOT_ON_GPU.. I think it would make sense to have them listed by how much we output.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -1965,4 +1965,11 @@ class RapidsConf(conf: Map[String, String]) extends Logging {
     // user-provided value takes precedence, then look in defaults map
     conf.get(key).orElse(optimizerDefaults.get(key)).map(toDouble(_, key))
   }
+
+  /**
+   * To judge whether "key" is explicitly set by the users.
+   */
+  def isConfExplicitlySet(key: String): Boolean = {
+    shouldExplain && conf.contains(key)
```
tgravescs, line 1973: remove shouldExplain out of here though since this is now a generic function you could call for any configs.    We don't really need the check in this case.

tgravescs:
build

sinkinben:
Hi, @tgravescs Is it fine to write commit messages like this?

```text
Modify the default value of spark.rapids.sql.explain as NOT_ON_GPU (#5819)


* Set the default value of `spark.rapids.sql.explain` to `NOT_ON_GPU`, both in spark2-sql-plugin and sql-plugin.
* Add description about `spark.rapids.sql.explain` in `getting-started.md`.
* Add log to notify the users that default value of `spark.rapids.sql.explain` is `NOT_ON_GPU` (in`RapidsPluginUtils`).
* Update the logic of `explainCatalystSQLPlan`

Signed-off-by: sinkinben <sinkinben@outlook.com>
```

Issue #4519, opened 2022-01-12
[FEA] Add regular expression support for Form Feed, Alert, and Escape control characters


andygrove:
**Is your feature request related to a problem? Please describe.**
I am filing this for sake of completeness and do not expect this to be a priority but we do not currently support the following items in regular expressions and we fall back to CPU. It should be simple to implement them.

```
\f | The form-feed character ('\u000C')
\a | The alert (bell) character ('\u0007')
\e | The escape character ('\u001B')
```

**Describe the solution you'd like**
Add support for these control characters.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #5829, opened 2022-06-14
Add regexp support for Alert, and Escape control characters

anthony-chang:
Closes https://github.com/NVIDIA/spark-rapids/issues/4519

Adds support for:
```
\a | The alert (bell) character ('\u0007')
\e | The escape character ('\u001B')
```

Since `\f` is already an escape sequence, we don't need to do anything special to support. 

I haven't added any additional unit tests since the fuzzer already generates these characters.


anthony-chang:
build

anthony-chang:
build

NVnavkumar (approved):
LGTM

Issue #5727, opened 2022-06-02
Add test for GpuCast canonicalization with timezone


tgravescs:
**Is your feature request related to a problem? Please describe.**
Need to try to find the minimal repro cast to be able to add an integration test for https://github.com/NVIDIA/spark-rapids/issues/5670

PR #5833, opened 2022-06-14
Add test for GpuCast canonicalization with timezone 

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/5727

This adds a test for https://github.com/NVIDIA/spark-rapids/issues/5670.  I verified that without the fix in https://github.com/NVIDIA/spark-rapids/issues/5670 the test throws an org.apache.spark.sql.catalyst.parser.ParseException.


tgravescs:
build

PR #5834, opened 2022-06-15
[WIP] Add support for null characters in regular expressions

anthony-chang:
Adds support for the null character `\u0000` in regular expressions. 

cuDF requires null characters to be represented as `\0` rather than the literal `\u0000`. But we can't simply transpile `\u0000` to `\0` since patterns such as `\u00002` (ie. null character followed by a 2) would transpile to `\02` (ie. the octal digit with codepoint=2) so we instead transpile `\u0000` to `(?:\0)`

I've also left null characters disabled in character classes due to a cuDF bug https://github.com/rapidsai/cudf/issues/11109

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
@@ -142,8 +142,7 @@ class RegexParser(pattern: String) {
       case '\\' =>
         parseEscapedCharacter()
       case '\u0000' =>
-        throw new RegexUnsupportedException(
-          "cuDF does not support null characters in regular expressions", Some(pos))
+        RegexGroup(false, RegexEscaped('0'))
       case '*' | '+' | '?' =>
         throw new RegexUnsupportedException(
           "base expression cannot start with quantifier", Some(pos))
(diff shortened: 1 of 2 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -1327,23 +1327,18 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
-          }
-        }
         if (endsWithLineAnchor(ll) || endsWithLineAnchor(rr)) {
           throw new RegexUnsupportedException(
             "cuDF does not support terms ending with line anchors on one side of a choice")
         }
 
+        // cuDF does not support terms ending with word boundaries on one side
+        // of a choice, such as "\\b|a"
+        if (endsWithWordBoundary(ll) || endsWithWordBoundary(rr)) {
```
anthony-chang, line 1337: Added in this PR because the updated fuzz tests caught this case

andygrove (dismissed):

anthony-chang:
build

anthony-chang:
build

anthony-chang:
build

NVnavkumar (commented):
Should update the predefined character classes that are supposed to include the null character as well.

NVnavkumar:
> > Should update the predefined character classes that are supposed to include the null character as well (such as `\p{ASCII}`, etc.
> 
> Null characters still don't work in character classes. We can do this after [rapidsai/cudf#11112](https://github.com/rapidsai/cudf/pull/11112) is merged

Can you file a tracking issue for this bug and note it in the code around those character classes?

anthony-chang:
build

NVnavkumar (approved):

PR #5840, opened 2022-06-15
[DOC] Updated RapidsConf to reflect the default value of `spark.rapids.sql.improvedFloatOps.enabled` [skip ci]

razajafri:
Since the `spark.rapids.sql.improvedFloatOps.enabled` is not disabled by default. 

This PR updates the RapidsConf to reflect that.


abellina (approved):
LGTM

tgravescs:
please put better description then updated docs

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -575,8 +575,7 @@ object RapidsConf {
   val IMPROVED_FLOAT_OPS = conf("spark.rapids.sql.improvedFloatOps.enabled")
     .doc("For some floating point operations spark uses one way to compute the value " +
       "and the underlying cudf implementation can use an improved algorithm. " +
-      "In some cases this can result in cudf producing an answer when spark overflows. " +
-      "Because this is not as compatible with spark, we have it disabled by default.")
+      "In some cases this can result in cudf producing an answer when spark overflows.")
     .booleanConf
     .createWithDefault(true)
```
tgravescs, line 580: we intentionally put the confs into our integration test with the defaults they had, if we change the default here those tests I believe should be updated as well...

integration_tests/src/main/python/spark_session.py:
    'spark.rapids.sql.improvedFloatOps.enabled': 'false',

Can you please file a 22.08 issue to update those tests. We may want to check other defualtas as well

razajafri:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
razajafri: created issue #5844

PR #5869, opened 2022-06-17
Qualification: fix sorting and add unit-tests script

### LINKED ISSUE #2949: Qualification tool: Document and automate generation of expected results files
since the qualification tool checks for certain unsupported features, if we ad support for those features the qualification tests can fail.  

We should document how to update those files and perhaps have a way to automatically generate to make it easier for devs. It should not automatically commit as dev should verify changes are ok.

see https://github.com/NVIDIA/spark-rapids/pull/2946#issuecomment-881536314 for an example

### LINKED ISSUE #5859: [BUG] Qualification tools csv order is not in sync
**Describe the bug**

I noticed that in `QualificationSuite`, the unit tests that check the order of the estimatedSummaryInfo  list are limited to the log file.
The CSV file showing the detailed report does not display the list in the same order.
The corner case is when multiple applications have equal speedup-factor. In that case the output will be non-deterministic depending on which app was added first.

An example of this corner case is `tools/src/test/resources/QualificationExpectations/qual_test_simple_expectation.csv` where we have three apps having same speed-ups.

**Expected behavior**
I suggest that the `writeDetailedReport` shows the details report in the same order the summary report does.

amahussein:

fixes #5859 and fixes #2949 

**Changes**

- **New**: replaced calls to html in the JS which used to cause false triggers of unsafe HTML.
- A small change in Qualification.scala
  - Instead of sorting EstimatedInfo, then QualInfoSummary separately. Now, the code sorts  QualInfoSummary first. Then the estimatedInfo array is extracted. This guarantees that the detailed report and the executive-summary have the same order when the default sorting is picked "Desc".
  - break the tie of order using the "endTime" as last resort. 
  - this required a fix in one of the expectations csv file.
- Added a script in `tools/src/test/resources/dev/run-qualification-tests.sh`
  - The new script can be used to regenerate the csv files.
  - example usage 
     ```
     ./run-qualification-tests.sh --cp=$JAVA_CP 
     ```

The above command will simply iterate on all the expected csv file and re-run the qualification tools. The generated output is then compared to the expected files.

```

"Run the Qualification tool to generate CSV files to update/replace Qualification expectations set."
"The script requires the following:"
"  1- A static var \$qualification_path_map <string, array<string>> that maps between"
"     unit-test name and the array of event logs to be processed for the test. This map"
"     needs to be updated as new test are added and/or log files of existing tests are"
"     modified."
"  2- \$qual_expectations_dir: A directory that has the expected csv files."
"     For each key K in \$qualification_path_map, a csv file should exist"
"     \${qual_expectations_dir}/\$k.csv"
"  3- \$qual_events_dir: A directory that has the event logs to be processed."

"The script iterates on the unit tests. For each key \"K\" in \$qualification_path_map:"
"  1- It runs the Qualification tool given the specified event logs."
"     The argument qual_out_dir=\${out-dir}/qualification-output/runs-output/\$K"
"     QualificationMain --output-directory \${qual_out_dir} \\"
"                        \${qualification_path_map[\$K]}"
"  2- It copies the rapids_4_spark_qualification_output.csv from \$qual_out_dir/"
"     to a new folder \${out-dir}/qualification-output/csv/\$K.csv"
"  3- Logs the result of \"diff\" between the two CSV files"
"     diff \${qual_expectations_dir}/\$K.csv \${out-dir}/qualification-output/csv/\$K.csv"

"Finally, the script logs the result of \"diff\" for two directories:"
"     diff \${qual_expectations_dir}/ \${out-dir}/qualification-output/csv/"
"The script returns a non-zero value if the two folders do not match."
log_msg
"Please make sure to update the CSV files for the failing unit tests."
log_msg
log_info "Optional Arguments:"
"  --rapids-jars-dir=<arg>       - Directory containing RAPIDS jars. By default the script sets it to the"
"                                  target directory of tools."
"                                  If the tools jar is loaded from default directory tools/target/**,"
"                                  then you may need to append the remaining RAPIDS jars to the \"--cp\""
"                                  argument."
"  --cp=<arg>                    - Classpath required as dependencies to run the qualification tool."
"                                  For example, if \$SPARK_HOME/jars are not in the directory <rapids-jars>,"
"                                  then pass it as --cp=\$SPARK_HOME/jars/*:\$CLASS_PATH"
"  --out-dir=<arg>               - Output directory passed to the qualification tool runs."
"                                  By default is is set to test/resources/dev/qualification-output"
"                                  The directory has two subdirectories: csv/ which has the CSV files; and"
"                                  runs-output/ which has the actual output of the tool."
"                                  target directory of tools."
"  --qual-expectations-dir=<arg> - Path of the reference CSV files."
"                                  By default is is set to test/resources/QualificationExpectations"
"  --qual-events-dir=<arg>       - By default, the scripts looks for log files inside the following two"
"                                  directories:"
"                                      - test/resources/spark-events-qualification; and"
"                                      - test/resources/spark-events-profiling"
"                                  When this argument is set in the CLI, it is the directory path of the Spark"
"                                  events logs used as input for the unit tests. So, the user needs to make"
"                                  sure that this directory has all the log files needed by the unit tests."
"  --prof-log-dir=<arg>          - Some the unit tests reads files located in another directory."
"                                  By default is is set to test/resources/spark-events-profiling"
"  --heap=<arg>                  - optional heap size. Default is 10g."
"  --help|-h                     - Shows Help."

"Example Usage:"
"  run-qualification-tests.sh --cp=\$CLASS_PATH --heap=5g"
"  This is equivalent to:"
"    java -Xmx5g \\"
"         -cp rapids-4-spark-tools_2.12-<version>-SNAPSHOT.jar:\$CLASS_PATH \\"
"         com.nvidia.spark.rapids.tool.qualification.QualificationMain \\"
"         --no-html-report --output-directory file:\$qual_out_dir \\"
"         \$LOGFILES"

"How to Add New Test:"
" 1- add a dummy csv file in \$qual-expectations-dir <new_test_name.csv>"
" 2- add the eventlog into the \$qual-events-dir <new_unit_log>"
" 3- update the definition of the hash in define_qualification_tests_map() to map between"
"    the expected file and the unit test name."
" 4- run the script. It is expected that the script fails because the dummy csv is incorrect."
" 5- copy the content of the generated output \$out-dir/csv/new_unit_test.csv"
"    into the expectation folders."

```


amahussein:
build

amahussein:
build

tools/src/test/resources/dev/run-qualification-tests.sh
```diff
@@ -0,0 +1,395 @@
...
+declare -A qualification_path_map
+
+show_help()
+{
+   # Display Help
+   log_msg
+   log_msg "Run the Qualification tool to generate CSV files to update/replace Qualification expectations set."
+   log_msg "The output folder of the CSV files is test/resources/dev/qualification-output/csv"
+   log_msg
+   log_msg "The script compares the CSV files. If pairs of files do not match or the directories are"
```
tgravescs, line 61: this sounds a bit weird, the whole point of the script is to update the results, so having it fail seems weird.  Perhaps we should say something like compares and reports on diffs. 
I assume user is responsible for looking at those and copying files if expected output changes

tools/src/test/resources/dev/run-qualification-tests.sh
```diff
@@ -0,0 +1,395 @@
...
+   log_msg "  --out-dir=<arg>               - Output directory passed to the qualification tool runs."
+   log_msg "                                  By default is is set to test/resources/dev/qualification-output"
+   log_msg "                                  The directory has two subdirectories: csv/ which has the CSV files; and"
+   log_msg "                                  runs-output/ which has the actual output of the tool."
+   log_msg "                                  target directory of tools."
+   log_msg "  --qual-expectations-dir=<arg> - Path of the reference CSV files."
+   log_msg "                                  By default is is set to test/resources/QualificationExpectations"
+   log_msg "  --qual-events-dir=<arg>       - Path of the Spark events logs used as input for the unit tests"
+   log_msg "                                  By default is is set to test/resources/spark-events-qualification"
+   log_msg "  --prof-log-dir=<arg>          - Some the unit tests reads files located in another directory."
```
tgravescs, line 81: nit grammar "Some the".    What happens if we just call this twice instead of having a separate option, once on qualification dir and once on files in profiling dir?  We don't use all the profiling once I thought so I assume user picks and chooses which ones to copy?

tools/src/test/resources/dev/run-qualification-tests.sh
```diff
@@ -0,0 +1,395 @@
...
+   log_msg "  This is equivalent to:"
+   log_msg "    java -Xmx5g \\"
+   log_msg "         -cp rapids-4-spark-tools_2.12-<version>-SNAPSHOT.jar:\$CLASS_PATH \\"
+   log_msg "         com.nvidia.spark.rapids.tool.qualification.QualificationMain \\"
+   log_msg "         --no-html-report --output-directory file:qualification-output/csv \\"
+   log_msg "         \$LOGFILES"
+   log_msg
+   log_msg "How to Add New Test:"
+   log_msg " 1- add a dummy csv file in qual-expectations-dir <new_unit_test.csv>"
+   log_msg " 2- add the eventlog into the prof-log-dir <new_unit_log>"
```
tgravescs, line 97: log file should generally be in qual-events-dir unless its common

tools/src/test/resources/dev/run-qualification-tests.sh
```diff
@@ -0,0 +1,395 @@
...
+   log_msg " 3- update the definition of the hash in define_qualification_tests_map() to map between"
+   log_msg "    the expected file and the unit test name."
+   log_msg " 4- run teh script. It is expected that the script fails because the dummy csv is incorrect."
```
tgravescs, line 100: the is misspelled

tools/src/test/resources/dev/run-qualification-tests.sh
```diff
@@ -0,0 +1,395 @@
...
+print_banner()
+{
+  printf '%80s\n' | tr ' ' -
+  log_info "$*"
+  printf '%80s\n' | tr ' ' -
+}
+
+set_rapids_jars_from_work_dir()
+{
+  rapids_tools_jar_file=( "$( find "${MODULE_PATH}" -type f \( -iname "*.jar" ! -iname "*tests.jar" ! -iname "original-rapids-4*.jar" \) )" )
```
tgravescs, line 134: seems like it would be safer to find rapids-4-spark-tools_*.jar

tools/src/test/resources/dev/run-qualification-tests.sh
```diff
@@ -0,0 +1,395 @@
...
+    # deleting user's data if the directory has other subdirectories.
+    BATCH_OUT_DIR="${script_out_dir}/qualification-output"
+    CSV_OUT_DIR="${BATCH_OUT_DIR}/csv"
+    RUNS_OUT_DIR="${BATCH_OUT_DIR}/runs-output"
+  else
+    log_msg "Script output is set as default: ${BATCH_OUT_DIR}"
+  fi
+
+  # cleanup output folder before running tests
+  if [ -d "$BATCH_OUT_DIR" ]; then
```
tgravescs, line 179: it would be safer to fail if directory exists unless users adds and overwrite flag

tools/src/test/resources/dev/run-qualification-tests.sh
```diff
@@ -0,0 +1,395 @@
...
+    PROF_LOG_DIR="${MODULE_PATH}/${RELATIVE_PROF_LOG_PATH}"
+  fi
+}
+
+initialize()
+{
+  arr=( ${WORK_DIR//"/${MODULE_NAME}/"/ } )
+  PROJECT_ROOT=${arr[0]}
+
+  process_events_and_expectations_paths
```
tgravescs, line 213: I would call this set_ since its not really processing them yet.

tools/src/test/resources/dev/run-qualification-tests.sh
```diff
@@ -0,0 +1,395 @@
...
+run_qualification_tool()
+{
+  for key in "${!qualification_path_map[@]}"; do
+    print_banner "Qualification tool:\t\t\t${key}"
+
+    output_dir="${RUNS_OUT_DIR}/${key}"
+
+    mkdir -p "${output_dir}"
+
+    ## set teh arguments
```
tgravescs, line 314: the misspelled

tools/src/test/resources/dev/run-qualification-tests.sh
```diff
@@ -0,0 +1,395 @@
...
+         ${qual_tool_args} \
+         ${qualification_path_map[$key]}
+
+    # check if the qualification tool failed
+    if [ $? -eq 0 ]
+    then
+    	log_info "OK: Running Qualification tool"
+    	process_qualification_output "$output_dir" "$key"
+    else
+      # we do not need to exit immediately because the two directories are compared any way
```
tgravescs, line 329: spacing off

tools/src/test/resources/dev/run-qualification-tests.sh
```diff
@@ -0,0 +1,395 @@
...
+  log_info "RAPIDS Jars Home                : ${rapids_jar_home}"
+  log_info "Classpath                       : ${java_cp}"
+}
+
+define_qualification_tests_map()
+{
+  qual_log_prefix="file:${QUAL_LOG_DIR}"
+  prof_log_prefix="file:${PROF_LOG_DIR}"
+
+  qualification_path_map[nds_q86_test_expectation]="${qual_log_prefix}/nds_q86_test ${qual_log_prefix}/malformed_json_eventlog.zstd"
```
tgravescs, line 233: @nartal1 just fyi, lets make sure we update this when adding new files.

tgravescs (commented):
I need to run the script myself still, this will be very nice to have.

tools/src/test/resources/dev/run-qualification-tests.sh
amahussein: I thought about that, but the problem with giving the user option not to overwrite is that a csv from previous run that was not generated by the most recent run would cause the final directory comparison to pass successfully.

tools/src/test/resources/dev/run-qualification-tests.sh
amahussein: Good catch. That's typo.

tools/src/test/resources/dev/run-qualification-tests.sh
amahussein: I will rephrase the comment to clarify.
The script will run all the tests and logging the result of the comparison of each unit test.
at the end, if the two folders do not match (all unit tests match), it will log the diff result of the two folders, and `exit 1`.
I used `exit 1` here in case someone wants to call the script programmatically. So instead of grepping the log to check against the output, the caller can just check the return of the script.

tools/src/test/resources/dev/run-qualification-tests.sh
amahussein: done!

amahussein:
I made some changes and added an update to the PR description.
The qualification JS code cause some false security checks. The fix put in this PR is to replace `.HTML()` with a method getter that does not trigger the XSS check.

tools/src/test/resources/dev/run-qualification-tests.sh
amahussein: I got rid of the second directory. Basically, if an input directory is passed to the script, then we should expect that it contains all the required log files.
I also updated the description to emphasize that.

tools/src/test/resources/dev/run-qualification-tests.sh
amahussein: I updated the description of the script.

tools/src/test/resources/dev/run-qualification-tests.sh
```diff
@@ -0,0 +1,425 @@
...
+   log_msg "  1- It runs the Qualification tool given the specified event logs."
+   log_msg "     The argument qual_out_dir=\${out-dir}/qualification-output/runs-output/\$K"
+   log_msg "     QualificationMain --output-directory \${qual_out_dir} \\"
+   log_msg "                        \${qualification_path_map[\$K]}"
+   log_msg "  2- It copies the rapids_4_spark_qualification_output.csv from \$qual_out_dir/"
+   log_msg "     to a new folder \${out-dir}/qualification-output/csv/\$K.csv"
+   log_msg "  3- Logs the result of \"diff\" between the two CSV files"
+   log_msg "     diff \${qual_expectations_dir}/\$K.csv \${out-dir}/qualification-output/csv/\$K.csv"
+   log_msg
+   log_msg "Finally, the script logs the result of \"diff\" for two directories:"
```
tgravescs, line 79: this is same statement as 3- above

tools/src/test/resources/dev/run-qualification-tests.sh
tgravescs: I'm not sure I understand, it would fail if any previous run is there and you try to run it again.  If you select overwrite then it removes it first.
I'm ok with leaving it this way, but please add comment in top level documentation.

tools/src/test/resources/dev/run-qualification-tests.sh
tgravescs: need to remove the option then. sorry if I was confusing, what does that mean then for this script?  we don't have all our eventlogs in one directory, so this script won't work with it?

tgravescs: Or you mean by default it works with our test setup, but if someone was using it for something else, they must all be in the same directory?

amahussein: > Or you mean by default it works with our test setup, but if someone was using it for something else, they must all be in the same directory?

Yes, exactly. this is what I meant.
By default, it will work fine for our test setup loading the log files from two directories.

tools/src/test/resources/dev/run-qualification-tests.sh
amahussein: I clarified the difference.
"3-" is diff between two CSV files.

The following step is to compare the entire two directories. I found that this step is necessary in case the Map is missing keys. Comparing the final directory is a second level to validate that the Map is not broken.

tools/src/test/resources/dev/run-qualification-tests.sh
amahussein: Good point. I added one more flag "-f|--force"

amahussein:
build

tgravescs (approved):

tools/src/test/resources/dev/run-qualification-tests.sh
```diff
@@ -0,0 +1,451 @@
+#!/usr/bin/env bash
```
tgravescs, line 1: sorry can we update the script name to something like generate-qualification-test-results, otherwise someone might think this is to run the tests

tgravescs:
so one thing we didn't address in https://github.com/NVIDIA/spark-rapids/issues/2949, is the documentation, I'm fine with  waiting and see if we really need it.

tgravescs:
build

amahussein:
build

amahussein:
build

tgravescs (approved):

PR #5870, opened 2022-06-17
Use a common thread pool across formats for multithreaded reads

jlowe:
Each file format is using a separate thread pool for multithreaded reads which is wasteful and adds mostly redundant configuration.  This consolidates all the separate thread pools into a single thread pool for multithreaded reads across all formats and deprecates the old per-format threadpool size configuration into a single configuration.  If there is a conflicting setting between the deprecated per-format configs or between the deprecated configs and the cross-format config, the largest value will be used.


jlowe:
build

revans2 (dismissed):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -720,6 +720,15 @@ object RapidsConf {
       .createWithDefault(false)
 
   // FILE FORMATS
+  val MULTITHREAD_READ_NUM_THREADS = conf("spark.rapids.sql.multiThreadedRead.numThreads")
+      .doc("The maximum number of threads on each executor to use for reading small " +
+        "files in parallel. This can not be changed at runtime after the executor has " +
+        "started. Used with COALESCING and MULTITHREADED readers, see " +
+        "spark.rapids.sql.format.parquet.reader.type for a discussion of reader types.")
```
tgravescs, line 727: this is referencing the parquet reader type still, probably want to add orc and avro or make it more generic somehow, that might help user know which reader types it works with as well

jlowe:
build

tgravescs (approved):

Issue #5813, opened 2022-06-14
[FEA] Set sql.json.read.double.enabled and sql.csv.read.double.enabled to `true` by default


sameerz:
**Is your feature request related to a problem? Please describe.**
In issue https://github.com/NVIDIA/spark-rapids/issues/5323 and PR https://github.com/NVIDIA/spark-rapids/pull/5441 we enabled floating point and some double operations by default.  We should also enable the following double configurations by default:

`spark.rapids.sql.csv.read.double.enabled`
`spark.rapids.sql.json.read.double.enabled`

**Describe the solution you'd like**
Update our configurations and documentation for the above settings to `true`.  

**Describe alternatives you've considered**
We've had these operations false by default due to slight variations between CPU and GPU for float and double calculations.  Setting these to true should not impact most use cases.  

**Additional context**
N/A

PR #5874, opened 2022-06-20
Change JSON and CSV read double config

GaryShen2008:
set spark.rapids.sql.json.read.double.enabled
and spakr.rapids.sql.csv.read.double.enabled true by default

Close https://github.com/NVIDIA/spark-rapids/issues/5813

Signed-off-by: gashen <gashen@nvidia.com>


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RapidsConf.scala
```diff
@@ -913,7 +913,7 @@ object RapidsConf {
   val ENABLE_READ_CSV_DOUBLES = conf("spark.rapids.sql.csv.read.double.enabled")
     .doc("CSV reading is not 100% compatible when reading doubles.")
     .booleanConf
-    .createWithDefault(false)
+    .createWithDefault(true)
```
tgravescs, line 916: do we need to update any tests?  ie tests assumed false so now we only test true

GaryShen2008: The only place to set it explicitly is https://github.com/NVIDIA/spark-rapids/blob/branch-22.08/integration_tests/src/main/python/csv_test.py#L171.
Every reading test related to double set spark.rapids.sql.csv.read.double.enabled to true explicitly.
I don't find any test case which requires it be false.

GaryShen2008: Same as json_test.py.

GaryShen2008: @tgravescs  Can you review it again?

sameerz:
Is this a duplicate of https://github.com/NVIDIA/spark-rapids/pull/5935 ? 

cc: @razajafri

razajafri:
> Is this a duplicate of #5935 ?
> 
> cc: @razajafri

Yes, looks like it is. You can decide which one to close

GaryShen2008:
Closed because duplicated

Issue #5867, opened 2022-06-17
[FEA] Add shim for Spark 3.3.1


tgravescs:
**Is your feature request related to a problem? Please describe.**
Now that Spark 3.3.0 is released we need a shim for 3.3.1 and we need to update our integration test builds

PR #5877, opened 2022-06-20
Add Apache Spark 3.3.1-SNAPSHOT Shims

anthony-chang:
Closes https://github.com/NVIDIA/spark-rapids/issues/5867

This PR adds shims for Spark version 3.3.1-SNAPSHOT, and a `330until340` shims folder. The shims should be identical to 3.3.0 except version number.


dist/pom.xml
```diff
@@ -51,7 +51,8 @@
         </noSnapshot.buildvers>
         <snapshot.buildvers>
             314,
-            322
+            322,
+            331
```
tgravescs, line 55: nit, put in order

sql-plugin/src/main/331/scala/com/nvidia/spark/rapids/shims/ParquetSchemaClipShims.scala
```diff
@@ -0,0 +1,91 @@
...
+import scala.collection.JavaConverters._
+
+import org.apache.parquet.schema.{MessageType, Type}
+
+import org.apache.spark.sql.execution.datasources.parquet.{ParquetReadSupport, ParquetUtils}
+import org.apache.spark.sql.execution.datasources.parquet.ParquetReadSupport.containsFieldIds
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.types.{StructField, StructType}
+
+object ParquetSchemaClipShims {
```
tgravescs, line 28: si this the same as 330?  If so perhaps we make a 330until340 directory

sql-plugin/src/main/331/scala/org/apache/spark/sql/execution/datasources/parquet/rapids/shims/ParquetCVShims.scala
```diff
@@ -0,0 +1,36 @@
...
+
+package org.apache.spark.sql.execution.datasources.parquet
+
+import org.apache.spark.memory.MemoryMode
+import org.apache.spark.sql.execution.vectorized.WritableColumnVector
+import org.apache.spark.sql.types.StructType
+
+object ParquetCVShims {
+
+  def newParquetCV(
```
tgravescs, line 25: same here, if this is same as 330 create a 330until340 directory that covers

tgravescs (commented):
In the future if you can add any details of things that changed or if everything the same just say 331 shim is same as 330 except version, that would help the reviewers out.

Is there a reason this was in draft?  Again if you can add comment that would be great.

dist/pom.xml
tgravescs: sorry looks like I misread, its already in order

tgravescs (approved):

PR #5878, opened 2022-06-21
Fix misleading documentation for `approx_percentile` and some other functions

andygrove:
The generated documentation stated that some expressions were disabled by default but this was not accurate now that `spark.rapids.sql.incompatibleOps.enabled` is true by default.

This PR changes the documentation generator to respect the default value of `spark.rapids.sql.incompatibleOps.enabled` when determining if expressions are enabled.

The `incompat` message for `ApproxPercentile` is also updated to remove the text about enabling `spark.rapids.sql.incompatibleOps.enabled`.

### DIFF
--- a/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
+++ b/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
@@ -3417,8 +3417,7 @@ object GpuOverrides extends Logging {
           val aggBuffer = c.aggBufferAttributes.head
           aggBuffer.copy(dataType = CudfTDigest.dataType)(aggBuffer.exprId, aggBuffer.qualifier)
         }
-      }).incompat("the GPU implementation of approx_percentile is not bit-for-bit " +
-          s"compatible with Apache Spark. To enable it, set ${RapidsConf.INCOMPATIBLE_OPS}"),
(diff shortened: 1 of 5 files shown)


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -3417,8 +3417,7 @@ object GpuOverrides extends Logging {
           val aggBuffer = c.aggBufferAttributes.head
           aggBuffer.copy(dataType = CudfTDigest.dataType)(aggBuffer.exprId, aggBuffer.qualifier)
         }
-      }).incompat("the GPU implementation of approx_percentile is not bit-for-bit " +
```
revans2, line 3420: But we are not 100% compatible with the Spark CPU version. We have to document this some where. Even if we as a group are okay with prioritizing performance over bit for bit identical results, we still need to document that this is different some how.

andygrove: @revans2 Could you take another look? Please see the updated PR description.

sameerz (approved):

revans2 (approved):

andygrove:
build

andygrove:
build

Issue #4837, opened 2022-02-22
[FEA] Optimize JSON reading of floating-point values


andygrove:
**Is your feature request related to a problem? Please describe.**
PR https://github.com/NVIDIA/spark-rapids/pull/4825 fixes some edge cases with parsing JSON numeric types but introduces an additional complex regular expression and also calls into `GpuCast.sanitizeStringToFloat` which also executes a complex regular expression which is mostly redundant for this path.

**Describe the solution you'd like**
The JSON reader should implement optimized custom code for casting string to float instead of re-using the GpuCast logic.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #5879, opened 2022-06-21
avoid duplicate sanitization step when reading JSON floats

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/4837

When reading floating-point values from JSON we use regexp to filter valid JSON floating-point numbers and then call GpuCast which does its own regexp filtering which is redundant in this case. This PR avoids the redundant check.

I ran performance tests and see an improvement with these changes

``` scala
// test data
val df = Range(0, 10000000).map(n => n * 3.141592).toDF("n")
df.write.json("/tmp/pies.json")

// test
spark.conf.set("spark.rapids.sql.json.read.double.enabled", true)
spark.conf.set("spark.rapids.sql.format.json.enabled", true)
spark.conf.set("spark.rapids.sql.format.json.read.enabled", true)
val schema = StructType(Seq(StructField("n", DataTypes.DoubleType, false)))
val df = spark.read.schema(schema).json("/tmp/pies.json")
df.createOrReplaceTempView("pi")
val df2 = spark.sql("SELECT avg(n) FROM pi")
spark.time(df2.collect)

// without optimization
Time taken: 4040 ms
Time taken: 3998 ms
Time taken: 3878 ms

// with optimization
Time taken: 3495 ms     
Time taken: 3685 ms
Time taken: 3728 ms
```


revans2 (approved):

andygrove:
build

PR #5883, opened 2022-06-22
Add support for `element_at` and `GetMapValue`

razajafri:
This PR adds support for `GetMapValue` and `element_at` to take a column of keys and return their values if the key exists otherwise null. 

This PR depends on https://github.com/rapidsai/cudf/pull/11128


sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2636,7 +2633,8 @@ object GpuOverrides extends Logging {
                   TypeSig.MAP.nested(TypeSig.commonCudfTypes + TypeSig.ARRAY + TypeSig.STRUCT +
                     TypeSig.NULL + TypeSig.DECIMAL_128 + TypeSig.MAP),
                   TypeSig.MAP.nested(TypeSig.all)),
-                ("key", TypeSig.commonCudfTypesLit() + TypeSig.lit(TypeEnum.DECIMAL), TypeSig.all))
+                ("key", TypeSig.commonCudfTypes + TypeSig.ARRAY + TypeSig.STRUCT +
```
revans2, line 2636: Why do we say that we support arrays, structs and maps as keys? We don't yet. They are not nested, so the only place that really becomes a problem, besides the docs, is if we try to use an empty structs as a key, but then we need a test for it.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -220,7 +220,21 @@ case class GpuElementAt(left: Expression, right: Expression, failOnError: Boolea
...
       case _: MapType =>
-        (_, _) => throw new UnsupportedOperationException("non-literal key is not supported")
+        (map, indices) => {
+          if (failOnError) {
+            withResource(map.getMapKeyExistence(indices)) { keyExists =>
+              withResource(keyExists.all()) { exist =>
+                if (!exist.isValid && exist.getBoolean) {
+                  map.getMapValue(indices)
+                } else {
+                  throw new NoSuchElementException("One of the keys doesn't exist")
```
revans2, line 230: nitL it would be nice to use the shim layer for this kind of error like what spark is doing just in case they decide to change it on us in the future. Also can we pick the first key that was missing? we already know we are going to throw an error, so performance is not critical.

```
  def mapKeyNotExistError(key: Any): NoSuchElementException = {
    new NoSuchElementException(s"Key $key does not exist.")
  }
```

integration_tests/src/main/python/map_test.py
```diff
@@ -338,6 +338,28 @@ def test_element_at_map_numeric_keys(data_gen):
...
+                         ids=idfn)
+def test_element_at_map_string_col_keys(data_gen):
+    keys = StringGen(pattern='key_[0-9]')
+    assert_gpu_and_cpu_are_equal_collect(
+        lambda spark: two_col_df(spark, data_gen, keys).selectExpr(
+            'element_at(a, b)', 'a[b]'),
+        conf={'spark.sql.ansi.enabled': False})
+
+@pytest.mark.parametrize('data_gen', [simple_string_to_string_map_gen], ids=idfn)
+def test_element_at_map_string_col_keys_ansi_fail(data_gen):
```
revans2, line 353: What about for get map value? Do we need a second test to verify that it works properly in ANSI mode?

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -220,7 +220,21 @@ case class GpuElementAt(left: Expression, right: Expression, failOnError: Boolea
           }
         }
       case _: MapType =>
-        (_, _) => throw new UnsupportedOperationException("non-literal key is not supported")
+        (map, indices) => {
+          if (failOnError) {
+            withResource(map.getMapKeyExistence(indices)) { keyExists =>
```
gerashegalov, line 225: in a conventional programming checking for key existence before doing the value lookup is considered a peroformance antipattern leading to an unnecessary second lookup. 

Is there a way to just build a logic with a single getMapValue?

razajafri: I thought about doing it without an extra lookup but there is no way for us to know if the map actually had a `null` value or was the key not found in the map.

razajafri:
build

razajafri:
CI is failing because of a dependency in cudf

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
ttnghia: This is very easy to achieve by re-implementing map lookup to return a pair of `{value, success}`. Otherwise, checking for existence then retrieving is doubly expensive.

ttnghia: I would prefer to go back and do that rather than accepting this double computation.

razajafri:
@revans2 I have addressed all your concerns I think

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2596,7 +2596,7 @@ object GpuOverrides extends Logging {
         TypeSig.all,
         ("map", TypeSig.MAP.nested(TypeSig.commonCudfTypes + TypeSig.ARRAY + TypeSig.STRUCT +
           TypeSig.NULL + TypeSig.DECIMAL_128 + TypeSig.MAP), TypeSig.MAP.nested(TypeSig.all)),
-        ("key", TypeSig.commonCudfTypesLit() + TypeSig.lit(TypeEnum.DECIMAL), TypeSig.all)),
+        ("key", TypeSig.commonCudfTypes + TypeSig.DECIMAL_128, TypeSig.all)),
```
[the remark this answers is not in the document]
revans2: What about it as a map key? We only test it as a value now.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala
```diff
@@ -2614,13 +2614,10 @@ object GpuOverrides extends Logging {
...
           TypeSig.ARRAY.nested(TypeSig.all) + TypeSig.MAP.nested(TypeSig.all)),
         ("index/key", (TypeSig.commonCudfTypes + TypeSig.DECIMAL_128)
-          .withPsNote(TypeEnum.INT, "Supported as array index. " +
-            "Only Literals supported as map keys.")
           .withPsNote(
             Seq(TypeEnum.BOOLEAN, TypeEnum.BYTE, TypeEnum.SHORT, TypeEnum.LONG,
               TypeEnum.FLOAT, TypeEnum.DOUBLE, TypeEnum.DATE, TypeEnum.TIMESTAMP,
-              TypeEnum.STRING, TypeEnum.DECIMAL),
-            "Unsupported as array index. Only Literals supported as map keys."),
+              TypeEnum.STRING, TypeEnum.DECIMAL), "Unsupported as array index."),
```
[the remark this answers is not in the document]
revans2: You are right. That is fine.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/ColumnVectorUtil.scala
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.rapids
+
+import ai.rapids.cudf.ColumnVector
+import com.nvidia.spark.rapids.{Arm, GpuScalar}
+
+object ColumnVectorUtil extends Arm {
```
revans2, line 22: nit: Could we work on naming this better? I don't really want a new `Util` that ends up being a dumping ground for random operations. This could go into GpuMapUtils if it is really specific to maps, or it could be renamed into GpuErrorUtils. If it is GpuErrorUtils, then I would prefer to see it pull in more and not just return a String, but actually throw the exception.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/collectionOperations.scala
```diff
@@ -220,7 +220,23 @@ case class GpuElementAt(left: Expression, right: Expression, failOnError: Boolea
           }
         }
       case _: MapType =>
-        (_, _) => throw new UnsupportedOperationException("non-literal key is not supported")
+        (map, indices) => {
+          if (failOnError) {
```
revans2, line 224: The code is still identical between the two implementations.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/ColumnVectorUtil.scala
razajafri: Putting it in `GpuMapUtils` makes more sense because this is map specific and we don't have to add an additional utility class

razajafri:
build

ttnghia (changes requested):
I believe that we can do much better to reduce the run time by half.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuMapUtils.scala
```diff
@@ -38,6 +41,41 @@ object GpuMapUtils extends Arm {
...
   }
 
+  def getMapValueOrThrow(
+      map: ColumnVector,
+      indices: ColumnVector,
+      dtype: DataType,
+      origin: Origin): ColumnVector = {
+    withResource(map.getMapKeyExistence(indices)) { keyExists =>
+      withResource(keyExists.all()) { exist =>
+        if (!exist.isValid && exist.getBoolean) {
```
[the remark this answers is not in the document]
razajafri: Great catch! I have added a test for this

razajafri:
build

razajafri:
merged https://github.com/rapidsai/cudf/pull/11147 and kicked off spark-rapids-jni nightly build

razajafri:
build

razajafri:
@revans2 @gerashegalov @ttnghia I think I have addressed your concerns. PTAL

revans2 (approved):

PR #5886, opened 2022-06-22
Avoid serializing plan in GpuCoalesceBatches, GpuHashAggregateExec, and GpuTopN

jlowe:
This fixes issues with the RDD closures in GpuCoalesceBatches, GpuHashAggregateExec, and GpuTopN to avoid serializing the plan object and all child nodes.  References to member variables were converted into local variables to sever the dependency between the closure and exec node object.


jlowe:
build

jlowe:
build

PR #5897, opened 2022-06-23
GpuBatchScanExec partitions should be marked transient

jlowe:
[SPARK-32168](https://issues.apache.org/jira/browse/SPARK-32168) marked the `partitions` member as `@transient`, but this change was not ported to the Spark 3.1.x version of `GpuBatchScanExec`.  This can cause a significant increase in serialized task binary size when the query plan is inadvertently serialized when the partition objects are relatively large, as is the case with Apache Iceberg partitions.


jlowe:
build

PR #5900, opened 2022-06-23
Optimize some common if/else cases

revans2:
`if(predicate, 0, 1)` or things like this are a very common operation in SQL. 

The current `GpuIf` code will try to optimize for the case when all of the values in the predicate are true, or when all of them are false. The only time this would work as an optimization is if either the cost of computing the true or false expression and the copy_if_else is very high. High enough to offset the cost of computing the allTrue/allFalse nearly every time.

We cannot totally know the probably of everything in the predicate for this batch being true or false. It does happen and depending on the use case it could happen a lot. But generally it is not that common. Also we can know, at least to some degree the cost of the `falseExpr`, `trueExpr`, and the copy_if_else. For cases where they are literal values this is all very very low.

Because the probably of the predicate being all true or all false is low I decided to just disable this optimization, except for the case when there are side effects and the cost of doing the if/else is much higher.

In a real world user query I saw this save over 13% for some specific profiling ranges. The actual total query time saved was about 300ms. A small, but measurable difference.


revans2:
build

abellina (approved):

jlowe (approved):

PR #5903, opened 2022-06-24
Avoid requiring single batch when using out-of-core sort

### LINKED ISSUE #5448: [BUG] partitioned writes require single batches and sorting, causing gpu OOM in some cases
We have seen cases where `GpuInsertIntoHadoopFsRelationCommand` runs OOM while trying to sort data for partitioned writes. The symptom is a task that is reading close to full GPU memory, and ends up having to spill the device store. It runs out of memory with most of the GPU in use.

Possible culprits:

This command [wants a single batch](https://github.com/NVIDIA/spark-rapids/blob/branch-22.06/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuInsertIntoHadoopFsRelationCommand.scala#L261) and the writer framework says it needs this [while out-of-core sort isn't implemented](https://github.com/NVIDIA/spark-rapids/blob/branch-22.06/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuDataWritingCommandExec.scala#L123) (it now is), so we should be able to stop requiring single batch (?). 

It is not clear to me whether the sort is absolutely necessary either. @jlowe pointed out it is likely to not hold open all the parquet writers as we read in batches, but that the sort is the likely issue here. We would need to confirm whether it is needed, or find other ways to handle partitions that are larger than GPU memory of a single executor.

res-life:
Closes #5448 

### Problem
The cause is that the input of the out-of-core sort is a single large batch.
The single large batch caused the OOM.

out-of-core sort should not require a single batch, it can pull all the input batches and then sort.
Before the out-of-core sort feature was added in, we do need the single batch for in-memory sorting.

### Solution
When executing partitioned writes which requires sorting,  we use out-of-core sort and do not require `single batch`.
Note: If specified `stable sort` configuration, still need `require a single batch` as before.


res-life:
build

res-life:
build

res-life:
I will file a follow-on issue to explore support for something like `DynamicPartitionDataConcurrentWriter`

res-life:
Thanks, @wjxiz1992 verified this PR against the corresponding NV bug: NDS 2.0 convert CSV to Parquet failed by OOM

res-life:
Filed a follow-on issue for `DynamicPartitionDataConcurrentWriter`: https://github.com/NVIDIA/spark-rapids/issues/5999

res-life:
@revans2  Help review

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuCreateDataSourceTableAsSelectCommand.scala
```diff
@@ -36,7 +36,8 @@ case class GpuCreateDataSourceTableAsSelectCommand(
     query: LogicalPlan,
     outputColumnNames: Seq[String],
     origProvider: Class[_],
-    gpuFileFormat: ColumnarFileFormat)
+    gpuFileFormat: ColumnarFileFormat,
+    useStableSort: Boolean)
```
revans2, line 40: My only nit is that we pass useStableSort around this code a lot, but in the final part when we do the sort we get it from a different location.

https://github.com/NVIDIA/spark-rapids/blob/4a313e7046d0aa34f284483a5ed643d9cb8b79f5/sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala#L207-L211

Could we please make it consistent? Either we pass it all the way down all the time, or we go off of the plan.conf all the time.

sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuFileFormatWriter.scala
```diff
@@ -204,7 +205,7 @@ object GpuFileFormatWriter extends Logging {
         val orderingExpr = GpuBindReferences.bindReferences(
           requiredOrdering
             .map(attr => SortOrder(attr, Ascending)), finalOutputSpec.outputColumns)
-        val sortType = if (RapidsConf.STABLE_SORT.get(plan.conf)) {
+        val sortType = if (useStableSort) {
```
res-life, line 208: Updated.

res-life:
`Premerge` is blocked by https://github.com/NVIDIA/spark-rapids/issues/6003

res-life:
build

revans2 (approved):

Issue #5904, opened 2022-06-24
[FEA] Add support for negated POSIX character classes in regular expressions


anthony-chang:
**Is your feature request related to a problem? Please describe.**
https://github.com/NVIDIA/spark-rapids/pull/5692 added support for the `\p{...}` syntax but there is also the uppercase  `\P{...}` syntax which are the negated versions of `\p{...}`

**Describe the solution you'd like**
Transpile `\P{...}` into the corresponding character class. 

**Describe alternatives you've considered**
None

**Additional context**
None

PR #5692, opened 2022-05-30
Add support for POSIX predefined character classes

anthony-chang:
Closes #4413

This PR adds support for `\p{...}` for these character classes defined [here](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) under "POSIX character classes (US-ASCII only)"

```
\p{Lower} | A lower-case alphabetic character: [a-z]
\p{Upper} | An upper-case alphabetic character:[A-Z]
\p{ASCII} | All ASCII:[\x00-\x7F]
\p{Alpha} | An alphabetic character:[\p{Lower}\p{Upper}]
\p{Digit} | A decimal digit: [0-9]
\p{Alnum} | An alphanumeric character:[\p{Alpha}\p{Digit}]
\p{Punct} | Punctuation: One of !"#$%&'()*+,-./:;<=>?@[\]^_`{\|}~
\p{Graph} | A visible character: [\p{Alnum}\p{Punct}]
\p{Print} | A printable character: [\p{Graph}\x20]
\p{Blank} | A space or a tab: [ \t]
\p{Cntrl} | A control character: [\x00-\x1F\x7F]
\p{XDigit} | A hexadecimal digit: [0-9a-fA-F]
\p{Space} | A whitespace character: [ \t\n\x0B\f\r]
```

Signed-off-by: Anthony Chang <antchang@nvidia.com>

PR #5905, opened 2022-06-24
Add support for negated POSIX character classes `\P`

anthony-chang:
Closes https://github.com/NVIDIA/spark-rapids/issues/5904

This PR adds support for the uppercase `\P{...}` syntax for negated POSIX predefined character classes. 

Related to: https://github.com/NVIDIA/spark-rapids/pull/5692, which added the `\p{...}` syntax


sql-plugin/src/main/scala/com/nvidia/spark/rapids/RegexParser.scala
```diff
@@ -804,6 +804,16 @@ class CudfRegexTranspiler(mode: RegexMode) {
...
+      case RegexCharacterRange(RegexChar(start), RegexChar(end)) =>
+        if (start <= '\n' && end >= '\r') {
+          Seq('\n', '\r')
+        } else if (start <= '\n' && end >= '\n') {
+          Seq('\n')
+        } else if (start <= '\r' && end >= '\r') {
+          Seq('\r')
+        } else {
+          Seq.empty
+        }
```
anthony-chang, line 816: Is there a more scala-idiomatic way to do this?

andygrove: This looks fine to me

andygrove (approved):

Issue #5624, opened 2022-05-24
[FEA] Let CPU handle Delta table's metadata related queries


viadea:
I wish we can just let CPU handle Delta table's metadata related queries.

The reason is there are some CPU fallbacks for Delta table's metadata queries such as the one reading _delta_log(Json files).
If the _delta_log is huge(say millions of rows), then the CPU fallback's performance penalty is not trivial. 

If we can just let CPU handle those metadata queries,  then at least the metadata queries' performance should be similar to CPU run.

andygrove:
Here is an example of some of the expensive transitions in a delta lake metadata query.

![delta-lake-meta](https://user-images.githubusercontent.com/934084/175554419-a55a835c-b088-4b78-bbaa-da69959e7851.png)

andygrove:
Setting `spark.rapids.sql.optimizer.enabled=true` removes some of the overhead here by avoiding moving to GPU for the final projection:

![delta-lake-opt](https://user-images.githubusercontent.com/934084/175555329-ebcee6ee-8aaa-428a-b369-4883b42d8d6b.png)

viadea:
@andygrove I checked the query plan and time before and after setting `spark.rapids.sql.optimizer.enabled=true` and the result is similar -- 30s or so.
The plan are different though.

PR #5912, opened 2022-06-24
Fall back to CPU for Delta Lake metadata queries [databricks]

andygrove:
Closes https://github.com/NVIDIA/spark-rapids/issues/5624

This PR adds a very specific "is this a Delta Lake metadata query?" check and then falls back to CPU in this case.

The integration test currently only runs on Databricks. I filed https://github.com/NVIDIA/spark-rapids/issues/5966 to expand Delta Lake testing to other Spark versions.

### DIFF
--- a/integration_tests/src/main/python/delta_lake_test.py
+++ b/integration_tests/src/main/python/delta_lake_test.py
@@ -1,0 +1,34 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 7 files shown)


abellina:
Should this fallback be under a config that can be disabled? In case someone does have a table named `_delta_log`, but it is not a delta lake table.

andygrove:
> Should this fallback be under a config that can be disabled? In case someone does have a table named `_delta_log`, but it is not a delta lake table.

Yes, that makes sense. I will add that.

andygrove:
build

integration_tests/src/main/python/delta_lake_test.py
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+from asserts import assert_gpu_fallback_collect
+from marks import allow_non_gpu, delta_lake
+from spark_session import with_cpu_session, is_databricks91_or_later
+
+_conf = {'spark.rapids.sql.explain': 'ALL'}
```
tgravescs, line 20: do we need this on for the test to work properly or just for user to double check output?  I'm assuming latter

tgravescs (approved):

integration_tests/src/main/python/delta_lake_test.py
andygrove: That's correct. This is not needed for the test just helpful to see what is happening.

Issue #5890, opened 2022-06-22
[FEA] Mount the alluxio buckets/paths on the fly when the query is being executed


viadea:
I wish we can mount the alluxio buckets/paths on the fly when the query is being executed.

For example, after the alluxio cluster is up and running, we do not need to mount all S3 buckets(or other sources) before running the query.
When the query is being executed, we mount all needed S3 buckets on the fly.

Issue #5872, opened 2022-06-18
[FEA] Set a default Alluxio Master IP


viadea:
Currently we need to get the Alluxio Master IP and then use below Spark RAPIDS config to convert the path:
```
spark.rapids.alluxio.pathsToReplace="s3://foo->alluxio://RM_hostname:19998/foo,gs://bar->alluxio://RM_hostname:19998/bar"
```

I wish we can set a default Alluxio Master IP. (Say Master Node's IP) so that users do not need to manually find the Alluxio Master IP.
Eg, `spark.rapids.alluxio.masterip=<Master Node's IP >`.

abellina:
I am not entirely sure what that IP should be. Is there a guarantee that the master process is over by the Spark master (I assume that's what you are looking to do).

The config you mention here is verbose, especially since `RM_hostname:19998` needs to be repeated for each mount. Could this be put into a different config so that the `pathsToReplace` one becomes: "s3://foo->/foo,gs://bar->bar" or something like this?

GaryShen2008:
Since we use this config to support auto mount feature by calling "alluxio fs mount", I don't think we can support the case that Alluxio master is not the same node of spark master. 
With this limitation, we can always get the alluxio master IP from alluxio's config file.
@viadea If you agree, I think it's no need to have this config. Instead, I'll have a new config like "spark.rapids.alluxio.automount.enabled" to enable the auto mount feature and get the master from alluxio's config.

viadea:
@GaryShen2008  I am thinking of some scenario:
If Alluxio master node is not one of the Spark nodes, then we could not find the alluxio's config. So how can we know where is the master?  Do you think we can have a config `spark.rapids.alluxio.masterip` which can override the value?

GaryShen2008:
> @GaryShen2008 I am thinking of some scenario: If Alluxio master node is not one of the Spark nodes, then we could not find the alluxio's config. So how can we know where is the master? Do you think we can have a config `spark.rapids.alluxio.masterip` which can override the value?

In that case, even the current auto mount feature won't work since we call the command line to mount the folder.
If we want to support such case, we'll have to investigate if it's possible to use Alluxio JDK to mount remotely.

PR #5925, opened 2022-06-28
Add Alluxio auto mount feature

GaryShen2008:
Mount the cloud bucket to Alluxio when driver converts FileSourceScanExec to GPU plan
The Alluxio master should be the same node as Spark driver node when using this feature
Introduce new configs:
    spark.rapids.alluxio.automount.enabled
    spark.rapids.alluxio.bucket.regex
    spark.rapids.alluxio.cmd

Signed-off-by: Gary Shen <gashen@nvidia.com>

Close https://github.com/NVIDIA/spark-rapids/issues/5872
Close https://github.com/NVIDIA/spark-rapids/issues/5890


GaryShen2008:
This change is to simplify the usage of Alluxio with Spark-Rapids when the Alluxio master node is the same node which runs Spark Driver app. E.g. When you run the notebook on Databricks, the spark driver will be ran on the master node and the Alluxio master is also installed on the master node.
The previous design for Alluxio usage, it requires users to run alluxio mount manually and provides a list of replacement rules like s3://bucket-foo/file -> alluxio://0.1.2.3:19998/bucket-foo/file. It's not easy to use.
This change introduces 3 new configs:
spark.rapids.alluxio.automount.enabled - enable the Alluxio auto mount feature
spark.rapids.alluxio.bucket.regex - the regex to decide which bucket to be mounted to Alluxio. Default value to match all the buckets starting with s3:// or s3a://.  (Optional)
spark.rapids.alluxio.mount.cmd - the customized command for a customized installation of Alluxio (Optional)

Prerequisites:
Alluxio master node is the same node which runs Spark Driver app. (So, it can call the command "alluxio fs mount" to mount buckets).

For a quick way to use this feature, you just need to set spark.rapids.alluxio.automount.enabled=true (and don't set spark.rapids.alluxio.pathsToReplace).
The workflow looks as below.
Normally, you'll set spark.hadoop.fs.s3a.access.key and spark.hadoop.fs.s3a.secret.key in spark config to read dataset on AWS S3. When you write df = spark.read.parquet("s3a://some-bucket/files"), we'll check the path of "s3a://some-bucket/files" with the spark.rapids.alluxio.bucket.regex. if it matches, we'll try to mount the bucket "some-bucket" to Alluxio's "/some-bucket" by calling a command line "alluxio fs mount --readonly --option s3a.accessKeyId=*** --option s3a.secretKey=*** /some-bucket s3a://some-bucket". 
We'll read the ALLUXIO_HOME environment variable to figure out where Alluxio installed and find out the Alluxio's master IP and port from ALLUXIO_HOME/conf/alluxio-site.properties.
We'll find the access key and secret key from spark config.
Finally, we'll replace the path in Spark plans from "s3a://some-bucket/files" to "alluxio://master_ip:port/some-bucket/files" then the spark job will read data through Alluxio to cache the data to local.

GaryShen2008:
Forgot to consider the case when driver context restarted, and the bucket has been mounted, it'll fail when mounting again.
I don't find a parameter to mount a mounted path, it'll return 255. Need to find a way to understand the mounted point.

Here we can still use command line, another option may use Alluxio java class but it'll depend on Alluxio jar.
I'll first try command line since it's simple.

GaryShen2008:
build

docs/configs.md
```diff
@@ -29,7 +29,10 @@ scala> spark.conf.set("spark.rapids.sql.incompatibleOps.enabled", true)
...
+<a name="alluxio.bucket.regex"></a>spark.rapids.alluxio.bucket.regex|A regex to decide which bucket should be auto-mounted to Alluxio. E.g. when setting as "^s3://bucket.*", the bucket which starts with "s3://bucket" will be mounted to Alluxio and the path "s3://bucket-foo/a.csv" will be replaced to "alluxio://0.1.2.3:19998/bucket-foo/a.csv". It's only valid when setting spark.rapids.alluxio.automount.enabled=true. The default value matches all the buckets in "s3://" or "s3a://" scheme.|^s3a{0,1}://.*
+<a name="alluxio.cmd"></a>spark.rapids.alluxio.cmd|Provide the Alluxio command, which is used to mount or get information. E.g. "su,ubuntu,-c,/opt/alluxio-2.8.0/bin/alluxio", it means: run Process(Seq("su", "ubuntu", "-c", "/opt/alluxio-2.8.0/bin/alluxio fs mount --readonly /bucket-foo s3://bucket-foo")), to mount s3://bucket-foo to /bucket-foo. the delimiter "," is used to convert to Seq[String] when you need to use a special user to run the mount command.|None
```
GaryShen2008, line 34: It allows to define a customized command. We may remove it for a security concern.

wbo4958:
> We'll find the access key and secret key from spark config

Will these configurations display on the Spark UI? and if yes, is it ok for that?

docs/configs.md
wbo4958, line 34: could we just use "su ubuntu -c /opt/alluxio-2.8.0/bin/alluxio" instead of separating them by ","

wbo4958: BTW, does "su ubuntu" require the password?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+          for (line <- buffered_source.getLines) {
+            if (line.startsWith("alluxio.master.hostname")) {
+              alluxio_master = line.split('=')(1).trim
+            } else if (line.startsWith("alluxio.master.rpc.port")) {
+              alluxio_port = line.split('=')(1).trim
+            }
+          }
+        } catch {
+          case e: FileNotFoundException =>
+            throw new RuntimeException(s"Not found Alluxio config in " +
```
wbo4958, line 64: seems FileNotFoundException should be better here

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+      alluxioCmd = conf.getAlluxioCmd.getOrElse(
+        Seq("su", "ubuntu", "-c", s"$alluxio_home/bin/alluxio"))
+
+      if (!isInit) {
+        // Default to read from /opt/alluxio-2.8.0 if not setting ALLUXIO_HOME
+        var alluxio_port: String = "19998"
+        var alluxio_master: String = null
+        var buffered_source: Source = null
+        try {
+          buffered_source = Source.fromFile(alluxio_home + "/conf/alluxio-site.properties")
```
wbo4958, line 54: here, maybe we can use the withResources in Arm to wrap a closable object

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+          for (line <- buffered_source.getLines) {
+            if (line.startsWith("alluxio.master.hostname")) {
```
wbo4958, line 56: what if there are some "space" prefixed?

eg.

``` console
        alluxio.master.hostname=xxxxx
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+              alluxio_master = line.split('=')(1).trim
+            } else if (line.startsWith("alluxio.master.rpc.port")) {
```
wbo4958, line 58: same as above

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+  var alluxio_home: String = "/opt/alluxio-2.8.0"
+  var isInit: Boolean = false
+
+  // Read out alluxio.master.hostname, alluxio.master.rpc.port
+  // from Alluxio's conf alluxio-site.properties
+  // We require an environment variable "ALLUXIO_HOME"
+  // This function will only read once from ALLUXIO/conf.
+  private def initAlluxioInfo(conf: RapidsConf): Unit = {
+    this.synchronized {
+      alluxio_home = scala.util.Properties.envOrElse("ALLUXIO_HOME", "/opt/alluxio-2.8.0")
```
wbo4958, line 44: is there any reason why not to put the alluxio_home/alluxio_Cmd into the `if (!isInit)` block?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+        }
+
+        if (alluxio_master == null) {
+          throw new RuntimeException(
+            s"Can't find alluxio.master.hostname from $alluxio_home/conf/alluxio-site.properties.")
+        }
+        alluxioMasterHost = Some(alluxio_master + ":" + alluxio_port)
+        // load mounted point by call Alluxio mount command.
+        // We also can get from REST API http://alluxio_master:alluxio_web_port/api/v1/master/info.
+        val (ret, output) = runAlluxioCmd(" fs mount")
```
wbo4958, line 78: I saw some exceptions will be thrown when failed to detect alluxio configuration or something else. But here if the "fs mount" failed to run, do we need to throw the exception also?

docs/configs.md
GaryShen2008: It can also use split to generate the seq unless there's another whitespace in the path.
To use the sequence way can fix the whitespace problem.
As my testing on Databricks, the notebook is ran as root user. So, "su ubuntu" doesn't require the password. But yes, this feature has some limitation like this if we want to use it on other environment.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
GaryShen2008: Just consider for the case that user changed the spark.rapids.alluxio.cmd at runtime.
So, it can be read in and use the new command. Mainly easier for debug case.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
GaryShen2008: Just want to print the file path $alluxio_home/conf/alluxio-site.properties and suggest to set ALLUXIO_HOME.
If FileNotFoundException can also print such message, it should be ok to change.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
GaryShen2008: Here I want to get the mounted path from alluxio command, but if failed to get, I regard it as no mounted path, and the exception will be threw when mounting a new path later.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
       relation: HadoopFsRelation,
       partitionFilters: Seq[Expression],
       dataFilters: Seq[Expression]): FileIndex = {
 
     val alluxioPathsReplace: Option[Seq[String]] = conf.getAlluxioPathsToReplace
+    val alluxioAutoMountEnabled = conf.getAlluxioAutoMountEnabled
+    val alluxioBucketRegex: String = conf.getAlluxioBucketRegex
 
-    if (alluxioPathsReplace.isDefined) {
+    val replaceFunc = if (alluxioPathsReplace.isDefined) {
```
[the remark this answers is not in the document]
GaryShen2008: Can you give me an example? The function requires different arguments like alluxioPathsReplace, I think.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
GaryShen2008: Use java.util.Properties to load the alluxio-site.properties to avoid space issue.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+            s"--option s3a.secretKey=${secret_key.get} /$bucket $remote_path"
+        }
+        val (output, _) = runAlluxioCmd(parameter)
+        if (output != 0) {
+          throw new RuntimeException(s"Mount bucket $remote_path to /$bucket failed $output")
+        }
+        logInfo(s"Mounted bucket $remote_path to /$bucket in Alluxio $output")
+        mountedBuckets(bucket) = remote_path
+      } else if (mountedBuckets(bucket).equals(remote_path)) {
+        logInfo(s"Already mounted bucket $remote_path to /$bucket in Alluxio")
```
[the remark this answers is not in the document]
GaryShen2008: This log means the s3 bucket has been mounted already. It'll happen when you run a new spark application.
So, it's a normal case that the bucket has been mounted.
Will the warning log be too fearful?

docs/configs.md
```diff
@@ -29,7 +29,10 @@ scala> spark.conf.set("spark.rapids.sql.incompatibleOps.enabled", true)
...
+<a name="alluxio.automount.enabled"></a>spark.rapids.alluxio.automount.enabled|Enable the feature of auto mounting the cloud storage to Alluxio. It requires the Alluxio master is the same node of Spark driver node. When it's true, it requires an environment variable ALLUXIO_HOME be set properly. The Alluxio master's host and port will be read from alluxio.master.hostname and alluxio.master.rpc.port(default: 19998) from ALLUXIO_HOME/conf/alluxio-site.properties, then replace a cloud path which matches spark.rapids.alluxio.bucket.regex like "s3://bar/b.csv" to "alluxio://0.1.2.3:19998/bar/b.csv", and the bucket "s3://bar" will be mounted to "/bar" in Alluxio automatically.|false
+<a name="alluxio.bucket.regex"></a>spark.rapids.alluxio.bucket.regex|A regex to decide which bucket should be auto-mounted to Alluxio. E.g. when setting as "^s3://bucket.*", the bucket which starts with "s3://bucket" will be mounted to Alluxio and the path "s3://bucket-foo/a.csv" will be replaced to "alluxio://0.1.2.3:19998/bucket-foo/a.csv". It's only valid when setting spark.rapids.alluxio.automount.enabled=true. The default value matches all the buckets in "s3://" or "s3a://" scheme.|^s3a{0,1}://.*
```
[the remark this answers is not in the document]
GaryShen2008: yes, basically, it can support more cloud providers. I just haven't tested the others. There might be some difference in the mount command like secret related parameters. The current code only supports s3.secretkey.

docs/get-started/getting-started-alluxio.md
```diff
@@ -312,6 +312,43 @@ There are two ways to leverage Alluxio in RAPIDS.
       --conf spark.executor.extraJavaOptions="-Dalluxio.conf.dir=${ALLUXIO_HOME}/conf" \
    ```
 
+## Alluxio auto mount for AWS S3 buckets
+
+There's a more user-friendly way to use Alluxio with RAPIDS when accessing S3 buckets.
```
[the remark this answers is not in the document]
GaryShen2008: ok, I'll update the doc.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
GaryShen2008: The alluxio command seq is like ("su", "ubuntu", "-c", "/opt/alluxio-2.8.0/bin/alluxio"), 
The string of " fs mount" is supposed to append to the last item in the seq to generate like ("su", "ubuntu", "-c", "/opt/alluxio-2.8.0/bin/alluxio fs mount").  
The original command should be  su ubuntu -c "/opt/alluxio-2.8.0/bin/alluxio fs mount".

wbo4958:
one question,

``` scala
val df = spark.read.parquet("s3:/bucket1/xxx")
df.show()


val df2 = spark.read.parquet("s3:/bucket2/xxx")
df2.show()
```

will this PR auto-mount s3:/bucket1 and s3:/bucket2 ?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+    if (i.length <= 1) {
+      throw new RuntimeException(s"path $path is not expected for Alluxio auto mount")
+    }
+    val bucket = i(1).split("/")(0)
+    (scheme, bucket)
+  }
+
+  private def runAlluxioCmd(param : String) : (Int,
+    scala.collection.mutable.ArrayBuffer[String]) = {
+    val params = alluxioCmd.tails.collect {
```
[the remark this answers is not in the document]
GaryShen2008: The seq should be like ("su", "ubuntu", "-c", "/opt/alluxio-2.8.0/bin/alluxio fs mount --readonly /$bucket $remote_path"), The final item in Seq is a full alluxio command, and it must be appended with the bucket name and secret keys if necessary.
If using alluxioCmd :+ param, the param will be regarded as the parameter of "su", as my understanding.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+        case Seq(first, _, _*) => first
+        case Seq(last) => last + param
+      }.toSeq
+    logInfo(s"Run command ${params.last}")
+    val out : scala.collection.mutable.ArrayBuffer[String] =
+      new scala.collection.mutable.ArrayBuffer[String](10)
+    val ret = Process(params).!(ProcessLogger(out += _, _ => Unit))
```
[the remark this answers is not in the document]
GaryShen2008: The Utils is private in org.apache.spark package.
If we want to use it, we have to define a wrapper class in org.apache.spark.rapids, I think.
Is it worth to create a wrapper class?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+  // Suppose the key doesn't exist when using like Databricks's instance profile
+  private def autoMountBucket(scheme: String, bucket: String,
+                      access_key: Option[String],
+                      secret_key: Option[String]): Unit = {
+    val remote_path = scheme + "//" + bucket
+    this.synchronized {
+      if (!mountedBuckets.contains(bucket)) {
+        // not mount yet, call mount command
+        val parameter = if (access_key.isEmpty) {
+          s" fs mount --readonly /$bucket $remote_path"
```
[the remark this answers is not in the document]
GaryShen2008: As explain in https://github.com/NVIDIA/spark-rapids/pull/5925#discussion_r916725991, to use seq here, it still requires to append the last item. 
Here the code has a flaw that it should check both access_key and secret_key. I plan to update it.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+  var isInit: Boolean = false
+
+  // Read out alluxio.master.hostname, alluxio.master.rpc.port
+  // from Alluxio's conf alluxio-site.properties
+  // We require an environment variable "ALLUXIO_HOME"
+  // This function will only read once from ALLUXIO/conf.
+  private def initAlluxioInfo(conf: RapidsConf): Unit = {
+    this.synchronized {
+      alluxio_home = scala.util.Properties.envOrElse("ALLUXIO_HOME", "/opt/alluxio-2.8.0")
+      alluxioCmd = conf.getAlluxioCmd.getOrElse(
```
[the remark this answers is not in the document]
GaryShen2008: yes, it's intentional. To support changing alluxio cmd at runtime, I put this code outside of the isInit section.
At least, easy for debug purpose.

docs/get-started/getting-started-alluxio.md
```diff
@@ -312,6 +312,43 @@ There are two ways to leverage Alluxio in RAPIDS.
...
+This feature requires the node running Spark driver has Alluxio installed,
+which means the node is also the master of Alluxio cluster. It will use `alluxio fs mount` command to
+mount the buckets in Alluxio.
+
+To enable the Alluxio auto mount feature, the simplest way is only to enable it by below config
+without setting `spark.rapids.alluxio.pathsToReplace`, which takes precedence over auto mount feature.
+``` shell
+--conf spark.rapids.alluxio.automount.enabled=true
+```
+If Alluxio is not installed in /opt/alluxio-2.8.0, you should set the environment variable `ALLUXIO_HOME`.
```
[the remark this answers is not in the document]
GaryShen2008: Just hope the user doesn't need to set it explicitly when alluxio is installed in /opt/alluxio-2.8.0.
Otherwise, the user must remember to set it when creating the DB cluster.
More user-friendly, I think.
@viadea What do you think?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+          // parse the output, E.g.
+          // s3a://bucket-foo/        on  /bucket-foo
+          // s3a://bucket-another/    on  /bucket-another
+          // /local_path              on  /
+          for (line <- output) {
+            val items = line.split(" +")
+            logInfo(line)
+            // We only support s3 remote path for now,
+            // need to change below if we want to support other type of cloud storage
+            if (items.length >= 3 && items(0).startsWith("s3") && !items(2).equals("/")) {
```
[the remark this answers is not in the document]
GaryShen2008: Changed to use java.net.URI to parse the output.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
 import org.apache.spark.sql.catalyst.expressions.{Expression, PlanExpression}
 import org.apache.spark.sql.execution.datasources.{FileIndex, HadoopFsRelation, InMemoryFileIndex}
 import org.apache.spark.sql.execution.datasources.rapids.GpuPartitioningUtils
 
-object AlluxioUtils {
+object AlluxioUtils extends Logging {
+  val mountedBuckets: scala.collection.mutable.Map[String, String] = scala.collection.mutable.Map()
+  var alluxioCmd: Seq[String] = null
+  var alluxioMasterHost: Option[String] = None
+  var alluxio_home: String = "/opt/alluxio-2.8.0"
```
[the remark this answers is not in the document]
GaryShen2008: Changed to use camel case

docs/get-started/getting-started-alluxio.md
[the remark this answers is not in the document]
GaryShen2008: Updated the doc.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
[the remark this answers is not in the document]
tgravescs: I might be remembering wrong but I thought the spark call allowed each parameter being separate, but if it’s a pain to use that though I’m ok leaving it this way

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+            // need to change below if we want to support other type of cloud storage
+            if (items.length >= 3 && items(0).startsWith("s3") && !items(2).equals("/")) {
+              val bucket = items(2).substring(1)
+              val remote_path = items(0).substring(0, items(0).length-1)
+              mountedBuckets(bucket) = remote_path
+              logInfo(s"Found mounted bucket $remote_path to /$bucket")
+            }
+          }
+        } else {
+          logWarning(s"Failed to run alluxio fs mount $ret")
```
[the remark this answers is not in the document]
GaryShen2008: The reason I just use logWarning is that I hope to postpone the failure when the user really uses a Alluxio path, which triggers the auto mount to run alluxio fs mount again.
If the user doesn't use any path which matches the regex, the job will keep working.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+        // We also can get from REST API http://alluxio_master:alluxio_web_port/api/v1/master/info.
+        val (ret, output) = runAlluxioCmd(" fs mount")
+        if (ret == 0) {
+          // parse the output, E.g.
+          // s3a://bucket-foo/        on  /bucket-foo
+          // s3a://bucket-another/    on  /bucket-another
+          // /local_path              on  /
+          for (line <- output) {
+            val items = line.split(" +")
+            logInfo(line)
```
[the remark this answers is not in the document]
GaryShen2008: Change it to logDebug.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+          }
+        } else {
+          logWarning(s"Failed to run alluxio fs mount $ret")
+        }
+        isInit = true
+      }
+    }
+  }
+
+  private def getSchemeAndBucketFromPath(path: String) : (String, String) = {
```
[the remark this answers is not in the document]
GaryShen2008: Changed to use URI to parse the scheme and bucket name.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,267 @@
...
+    val i = path.split("//")
+    val scheme = i(0)
```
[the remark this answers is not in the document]
GaryShen2008: Changed to use URI to parse and check the uri.getScheme direclty.

tgravescs:
is this all working and ready for another review?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
viadea: Couple of tests done in local workstation,here are my thoughts:
1. The error is like `java.lang.RuntimeException: Mount bucket s3a://mybucket/ to /mybucket failed 1`. Here the `1` is the `stdout` which does not give much debug info. I hope we can print stderr and stdout to show the reason why mount failed.
2. In my env, there is no need to "su" because the Alluxio cluster and Spark users are the same.
So if I tried to set:
```
spark.conf.set("spark.rapids.alluxio.cmd", "/home/xxx/alluxio-2.8.0/bin/alluxio")
```
Then it will fail with :
```
java.io.IOException: Cannot run program "/home/xxx/alluxio-2.8.0/bin/alluxio fs mount --readonly --option s3a.accessKeyId=xxx --option s3a.secretKey=yyy /mybucket s3a://mybucket/": error=2, No such file or directory
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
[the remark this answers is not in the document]
tgravescs: so I may have been wrong here, it looks like S3 allows spaces in the paths and URI doesn't handle that:

URISyntaxException: Illegal character in path at index 79: s3://..../dayhour=2022-06-10 03%3A00%3A00/....

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
GaryShen2008: I think the latest code has fixed above issue.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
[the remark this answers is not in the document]
GaryShen2008: Updated to define separate functions for different usage.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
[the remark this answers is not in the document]
GaryShen2008: Revert back to parse the scheme and bucket by myself. Not using URI.

GaryShen2008:
> is this all working and ready for another review?

Yes, I think so.

docs/configs.md
```diff
@@ -29,7 +29,10 @@ scala> spark.conf.set("spark.rapids.sql.incompatibleOps.enabled", true)
 
 Name | Description | Default Value
 -----|-------------|--------------
-<a name="alluxio.pathsToReplace"></a>spark.rapids.alluxio.pathsToReplace|List of paths to be replaced with corresponding alluxio scheme. Eg, when configureis set to "s3:/foo->alluxio://0.1.2.3:19998/foo,gcs:/bar->alluxio://0.1.2.3:19998/bar", which means:       s3:/foo/a.csv will be replaced to alluxio://0.1.2.3:19998/foo/a.csv and      gcs:/bar/b.csv will be replaced to alluxio://0.1.2.3:19998/bar/b.csv|None
+<a name="alluxio.automount.enabled"></a>spark.rapids.alluxio.automount.enabled|Enable the feature of auto mounting the cloud storage to Alluxio. It requires the Alluxio master is the same node of Spark driver node. When it's true, it requires an environment variable ALLUXIO_HOME be set properly. The Alluxio master's host and port will be read from alluxio.master.hostname and alluxio.master.rpc.port(default: 19998) from ALLUXIO_HOME/conf/alluxio-site.properties, then replace a cloud path which matches spark.rapids.alluxio.bucket.regex like "s3://bar/b.csv" to "alluxio://0.1.2.3:19998/bar/b.csv", and the bucket "s3://bar" will be mounted to "/bar" in Alluxio automatically.|false
```
tgravescs, line 32: how should one set ALLUXIO_HOME?  for instance does it need to be in the spark-env.sh or can be set on command line when launching spark-submit/spark-shell.  On yarn cluster mode might need to use spark.yarn.appMasterEnv.....   I'm fine with leaving these details about but woudl be nice to put in docs at some point later

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
tgravescs: I still think runAlluxioCmd should deal with putting a space in between it rather then having caller do it. we can do it later as followup though too

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,286 @@
...
+      Process(params).!(ProcessLogger(out += _, _ => Unit))
+    }
+    (ret, out)
+  }
+
+  // path is like "s3://foo/test...", it mounts bucket "foo" by calling the alluxio CLI
+  // And we'll append --option to set access_key and secret_key if existing.
+  // Suppose the key doesn't exist when using like Databricks's instance profile
+  private def autoMountBucket(scheme: String, bucket: String,
+                      access_key: Option[String],
```
tgravescs, line 136: indentation should be 4 spaces

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,286 @@
...
+    (scheme, bucket)
+  }
+
+  private def runAlluxioCmd(param : String) : (Int,
+    scala.collection.mutable.ArrayBuffer[String]) = {
+    val params = alluxioCmd.tails.collect {
+        case Seq(first, _, _*) => first
+        case Seq(last) => last + param
+      }.toSeq
+    logInfo(s"Run command ${params.last}")
```
tgravescs, line 118: I don't htink this is guaranteed to be the command.  I want to make sure we aren't printing the secret key out here by accident so I would prefer to see this removed.

GaryShen2008, line 118:
```suggestion
```

docs/configs.md
GaryShen2008: Yes, I think both can work for ALLUXIO_HOME. We read it in driver side. So, ALLUXIO_HOME should be set in driver environment. On Databricks, it's added into the environment variables under spark config. Let me update the doc for the suggested way to set it.

docs/get-started/getting-started-alluxio.md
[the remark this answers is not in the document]
GaryShen2008: Updated doc.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
GaryShen2008: Changed to add the space in runAlluxioCmd.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
GaryShen2008: Updated.

docs/configs.md
```diff
@@ -29,7 +29,10 @@ scala> spark.conf.set("spark.rapids.sql.incompatibleOps.enabled", true)
...
+<a name="alluxio.cmd"></a>spark.rapids.alluxio.cmd|Provide the Alluxio command, which is used to mount or get information. E.g. "su,ubuntu,-c,/opt/alluxio-2.8.0/bin/alluxio", it means: run Process(Seq("su", "ubuntu", "-c", "/opt/alluxio-2.8.0/bin/alluxio fs mount --readonly /bucket-foo s3://bucket-foo")), to mount s3://bucket-foo to /bucket-foo. the delimiter "," is used to convert to Seq[String] when you need to use a special user to run the mount command.|None
+<a name="alluxio.pathsToReplace"></a>spark.rapids.alluxio.pathsToReplace|List of paths to be replaced with corresponding alluxio scheme. E.g. when configure is set to "s3://foo->alluxio://0.1.2.3:19998/foo,gs://bar->alluxio://0.1.2.3:19998/bar", it means: "s3://foo/a.csv" will be replaced to "alluxio://0.1.2.3:19998/foo/a.csv" and "gs://bar/b.csv" will be replaced to "alluxio://0.1.2.3:19998/bar/b.csv". To use this config, you have to mount the buckets to Alluxio by yourself. If you set this config, spark.rapids.alluxio.automount.enabled won't be valid.|None
```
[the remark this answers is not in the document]
GaryShen2008: Fixed

docs/configs.md
GaryShen2008: Updated the doc to mention the default value. But the None is generated automatically if we use it as an optional config.

tgravescs: yeah optional with default seems weird. I would say make it not optional

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -16,92 +16,285 @@
...
+      Process(params).!(ProcessLogger(out += _, _ => Unit))
+    }
+    (ret, out)
+  }
+
+  // path is like "s3://foo/test...", it mounts bucket "foo" by calling the alluxio CLI
+  // And we'll append --option to set access_key and secret_key if existing.
+  // Suppose the key doesn't exist when using like Databricks's instance profile
+  private def autoMountBucket(scheme: String, bucket: String,
+                              access_key: Option[String],
```
tgravescs, line 135: spacing is still off, parameters wrapping needs to be indented 4 spaces

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
viadea: @GaryShen2008 I just confirmed that now the issue regarding `spark.rapids.alluxio.cmd` got fixed based on my test on my local workstation. Now it does not need to `su`

docs/configs.md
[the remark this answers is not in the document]
GaryShen2008: Filed https://github.com/NVIDIA/spark-rapids/issues/6069

docs/get-started/getting-started-alluxio.md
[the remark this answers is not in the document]
GaryShen2008: I removed this line, since now we only read the alluxio command path from the alluxio.cmd config.
The ALLUXIO_HOME is only used for reading Alluxio configurations.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
GaryShen2008: Fixed.

GaryShen2008:
build

docs/configs.md
GaryShen2008: Updated to use a default value.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -47,93 +50,287 @@ object AlluxioUtils extends Logging {
...
+      Process(params).!(ProcessLogger(out += _, _ => Unit))
+    }
+    (ret, out)
+  }
+
+  // path is like "s3://foo/test...", it mounts bucket "foo" by calling the alluxio CLI
+  // And we'll append --option to set access_key and secret_key if existing.
+  // Suppose the key doesn't exist when using like Databricks's instance profile
+  private def autoMountBucket(scheme: String, bucket: String,
+                                access_key: Option[String],
```
tgravescs, line 155: indentation still off here, shoudl be 4 spaces from left

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -47,93 +50,287 @@ object AlluxioUtils extends Logging {
...
+      } else {
+        val envAccessKey = scala.util.Properties.envOrNone("AWS_ACCESS_KEY_ID")
+        val envSecretKey = scala.util.Properties.envOrNone("AWS_ACCESS_SECRET_KEY")
+        (envAccessKey, envSecretKey)
+      }
+    }
+  }
+
+  private def genFuncForPathReplacement(replaceMapOption: Option[Map[String, String]]
+                                       ) : Option[Path => Path] = {
```
tgravescs, line 210: spacing off here as well

sql-plugin/src/main/scala/com/nvidia/spark/rapids/AlluxioUtils.scala
```diff
@@ -47,93 +50,287 @@ object AlluxioUtils extends Logging {
...
+          f
+        }
+      })
+    } else {
+      None
+    }
+  }
+
+  private def genFuncForAutoMountReplacement(conf: RapidsConf, relation: HadoopFsRelation,
+                                             alluxioBucketRegex: String) : Option[Path => Path] = {
```
tgravescs, line 232: s;acing off

tgravescs (approved):

PR #5927, opened 2022-06-28
[WIP] Enable Strings as a supported type for GpuColumnarToRow transitions

### LINKED ISSUE #5633: [FEA] Enable Strings as a supported type for GpuColumnarToRow transitions
To enable fast data transitions between host and device, we want to allow for strings to be converted from row to columnar format (and vice versa).  This specific enhancement involves the plugin enabling strings as a supported type for GpuColumnarToRow transitions.

Ref: https://github.com/rapidsai/cudf/issues/10033#issuecomment-1132121915

amahussein:

fixes #5633

**Changes:**

Addressing points in https://github.com/rapidsai/cudf/issues/10033#issuecomment-1132121915

1. Added new JCudfUtil. I hope that this Util class will hide all the details of how data types should be aligned in the unsafeRow. This will make code it easier to use the same code as we: estimate size, set offsets, build the unsafeRow
2. Added String types to the supported transitions.
3. Updated `AcceleratedColumnarToRowIterator` and how the packMap is constructed.
4. Added a new variable to CudfUnsafeRow that represents an initial rough estimate of the unsafeRow. This estimate can be later used to decide if the schema fits the CUDF optimized conversion.

**TODO:**

- Update [GeneratedInternalRowToCudfRowIterator](https://github.com/NVIDIA/spark-rapids/blob/67c34e2b40ec196b12d83667aadd85ba9d45c8b6/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuRowToColumnarExec.scala#L682).
- I think it can be better if all the code of updating offset is inside the new class JCudfUtil. In that case, alignment is hidden from the rest of the call sites.

### DIFF
--- a/sql-plugin/src/main/java/com/nvidia/spark/rapids/CudfUnsafeRow.java
+++ b/sql-plugin/src/main/java/com/nvidia/spark/rapids/CudfUnsafeRow.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package com.nvidia.spark.rapids;
 
+import ai.rapids.cudf.DType;
 import org.apache.spark.sql.catalyst.InternalRow;
 import org.apache.spark.sql.catalyst.expressions.Attribute;
 import org.apache.spark.sql.catalyst.expressions.SpecializedGettersReader;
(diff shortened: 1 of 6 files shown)


sql-plugin/src/main/java/com/nvidia/spark/rapids/CudfUnsafeRow.java
```diff
@@ -253,12 +310,12 @@ public Decimal getDecimal(int ordinal, int precision, int scale) {
 
   @Override
   public UTF8String getUTF8String(int ordinal) {
-//    if (isNullAt(ordinal)) return null;
-//    final long offsetAndSize = getLong(ordinal);
-//    final int offset = (int) (offsetAndSize >> 32);
-//    final int size = (int) offsetAndSize;
-//    return UTF8String.fromAddress(null, address + offset, size);
-    throw new IllegalArgumentException("NOT IMPLEMENTED YET");
+    if (isNullAt(ordinal)) return null;
```
revans2, line 313: nit: the style guide says we should not use one line if statements like this.

```
if (isNullAt(ordinal)) {
  return null;
}
```

Issue #5854, opened 2022-06-16
[BUG] Memory leaked in some test cases


res-life:
**Describe the bug**
Some test cases have leaked resources although they were passed, better to fix the tests.
There are many `WAS LEAKED ` error logs running test cases.
These error logs may mask the leaks in the product code.

```
All tests passed.
22/06/16 14:19:51.290 Thread-7 ERROR HostMemoryBuffer: A HOST BUFFER WAS LEAKED (ID: 1 7fde6ffff010)
22/06/16 14:19:51.300 Thread-7 ERROR MemoryCleaner: Leaked host buffer (ID: 1): 2022-06-16 14:19:42.0739 CST: INC
java.lang.Thread.getStackTrace(Thread.java:1559)
ai.rapids.cudf.MemoryCleaner$RefCountDebugItem.<init>(MemoryCleaner.java:301)
ai.rapids.cudf.MemoryCleaner$Cleaner.addRef(MemoryCleaner.java:82)
ai.rapids.cudf.MemoryBuffer.incRefCount(MemoryBuffer.java:232)
ai.rapids.cudf.MemoryBuffer.<init>(MemoryBuffer.java:98)
ai.rapids.cudf.HostMemoryBuffer.<init>(HostMemoryBuffer.java:196)
ai.rapids.cudf.HostMemoryBuffer.<init>(HostMemoryBuffer.java:192)
ai.rapids.cudf.HostMemoryBuffer.allocate(HostMemoryBuffer.java:144)
com.nvidia.spark.rapids.RapidsHostMemoryStore.<init>(RapidsHostMemoryStore.scala:38)
com.nvidia.spark.rapids.RapidsBufferCatalog$.init(RapidsBufferCatalog.scala:191)
com.nvidia.spark.rapids.GpuDeviceManager$.initializeRmm(GpuDeviceManager.scala:301)
com.nvidia.spark.rapids.GpuDeviceManager$.initializeMemory(GpuDeviceManager.scala:330)
com.nvidia.spark.rapids.GpuDeviceManager$.initializeGpuAndMemory(GpuDeviceManager.scala:137)
com.nvidia.spark.rapids.RapidsExecutorPlugin.init(Plugin.scala:232)
org.apache.spark.internal.plugin.ExecutorPluginContainer.$anonfun$executorPlugins$1(PluginContainer.scala:125)
scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:293)
[...]

PR #5938, opened 2022-06-30
Fix leaks in the test cases of CachedBatchWriterSuite

res-life:
contributes #5854

Fix leaks in the test cases of CachedBatchWriterSuite

### Changes 1
The following  `spyCol.split` is executed actually, ant it creates non-closed `cv` splits.

```
when(spyCol.split(any())).thenReturn(splitCols.toArray)
```

Should be 
```
when(spyCol).split()
```
See [mockito document](https://github.com/mockito/mockito/blob/v4.6.1/src/main/java/org/mockito/Mockito.java#L634toL644)

### Changes 2
Some refactor


res-life:
build

jlowe (approved):

Issue #106, opened 2020-06-02
[FEA] integration tests for disabeling input formats


revans2:
**Is your feature request related to a problem? Please describe.**
This is related to #61 
and depends on #92 

We should have integration tests that disable the various input formats while doing some GPU enabled computation and verify we produce a correct result.

Issue #95, opened 2020-06-02
[FEA] cast integration tests


revans2:
**Is your feature request related to a problem? Please describe.**
The existing cast and ansi-cast tests are very good.  We just need to make sure that they are pulled into an area/location that lets us run them as integration tests.  I think pyspark would be great for this, but if we want to leave them in scalatest that is fine too.

sameerz:
Deferring to Scala unit tests that already exist.

PR #5946, opened 2022-07-04
GpuGlobalLimitExec and GpuCollectLimitExec support offset

### LINKED ISSUE #5589: [FEA] `GpuGlobalLimitExec` and `GpuCollectLimitExec` support `offset`
A new parameter named `offset` is introduced in `GlobalLimitExec` and `CollectLimitExec` from Spark 340. Accordingly, our GPU versions should also support it.

```
case class GlobalLimitExec(limit: Int = -1, child: SparkPlan, offset: Int = 0)
  extends BaseLimitExec {
  assert(limit >= 0 || (limit == -1 && offset > 0))
  ...
}
```

Now they will fall back to CPU for non-zero offset.

sinkinben:
Closes #5589
Support the parameter `offset` in `GpuGlobalLimitExec` and `GpuCollectLimitExec`, which is introduced in spark-3.4.0.

Since spark-rapids are failing to build against spark-3.4.0, I verified this locally with some local changes.

### DIFF
--- a/integration_tests/src/main/python/limit_test.py
+++ b/integration_tests/src/main/python/limit_test.py
@@ -44,24 +44,23 @@ def test_non_zero_offset(offset):
     # 4096: offset > df.numRows
 
     offset_sql = "select * from tmp_table offset {}".format(offset)
-    data_gen = StructGen([('c1', int_gen),('c2', int_gen),('c3', int_gen)], nullable=False)
-    assert_gpu_and_cpu_are_equal_sql(lambda spark: gen_df(spark, data_gen, length=2048), 'tmp_table',
+    assert_gpu_and_cpu_are_equal_sql(lambda spark: unary_op_df(spark, int_gen, length=2048), 'tmp_table',
                                      offset_sql, conf=offset_test_conf)
 
 
-@pytest.mark.parametrize('limit, offset', [(0, 0), (1024, 500), (2048, 456), (3000, 111), (500, 500), (100, 600)])
+@pytest.mark.parametrize('limit, offset', [(0, 0), (0, 10), (1024, 500), (2048, 456), (3000, 111), (500, 500), (100, 600)])
 @pytest.mark.skipif(is_before_spark_340(), reason='offset is introduced from Spark 3.4.0')
 def test_non_zero_offset_with_limit(limit, offset):
     # In CPU version of spark, (limit, offset) can not be negative number.
     # Test case description:
     # (0, 0): Corner case: both limit and offset are 0
+    # (0, 10): Corner case: limit = 0, offset > 0
     # (1024, 500): offset < limit && limit < df.numRows
     # (2048, 456): offset < limit && limit = df.numRows
     # (3000, 111): offset < limit && limit > df.numRows
-    # (500, 500): limit = offset
-    # (100, 600): limit > offset
+    # (500, 500): offset = limit
+    # (100, 600): offset > limit
 
-    data_gen = StructGen([('c1', int_gen),('c2', int_gen),('c3', int_gen)], nullable=False)
     limit_offset_sql = "select * from tmp_table limit {} offset {}".format(limit, offset)
-    assert_gpu_and_cpu_are_equal_sql(lambda spark: gen_df(spark, data_gen, length=2048), 'tmp_table',
+    assert_gpu_and_cpu_are_equal_sql(lambda spark: unary_op_df(spark, int_gen, length=2048), 'tmp_table',
                                      limit_offset_sql, conf=offset_test_conf)
--- a/integration_tests/src/main/python/limit_test.py
+++ b/integration_tests/src/main/python/limit_test.py
@@ -30,19 +30,45 @@ def test_simple_limit(data_gen):
 
 @allow_non_gpu('CollectLimitExec', 'GlobalLimitExec', 'ShuffleExchangeExec')
 @pytest.mark.skipif(is_before_spark_340(), reason='offset is introduced from Spark 3.4.0')
-def test_non_zero_offset_fallback():
-    def test_fn(spark):
-        unary_op_df(spark, int_gen, num_slices=1).createOrReplaceTempView("offset_tmp")
-        # `offset` is not exposed to Pyspark, so use the sql string.
-        return spark.sql("select * from offset_tmp limit 10 offset 1")
+@allow_non_gpu('CollectLimitExec', 'GlobalLimitExec', 'ShuffleExchangeExec')
+@pytest.mark.skipif(is_before_spark_340(), reason='offset is introduced from Spark 3.4.0')
+def test_non_zero_offset_on_gpu():
+    conf = {
+        'spark.rapids.sql.exec.CollectLimitExec': 'true',
+        'spark.rapids.sql.exec.GlobalLimitExec': 'true'
+    }
 
-    assert_gpu_fallback_collect(
-        # We need some processing after the limit to avoid a CollectLimitExec
-        lambda spark: test_fn(spark).repartition(1),
-        'GlobalLimitExec',
-        conf = {'spark.sql.execution.sortBeforeRepartition': 'false'})
+    # num_slices is the number of partitions of data frame
+    def test_runner(sql, num_slices):
+        def test_instance(spark):
+            df = unary_op_df(spark, int_gen, num_slices=num_slices)
+            df.createOrReplaceTempView("tmp_table")
+            return spark.sql(sql)
+        return test_instance
 
-    assert_gpu_fallback_collect(
-        test_fn,
-        'CollectLimitExec',
-        conf = {'spark.sql.execution.sortBeforeRepartition': 'false'})
+    sql = "select * from tmp_table limit {} offset {}"
+    # Only one partition
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(0, 0), 1), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(10, 2), 1), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(2, 10), 1), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(1024, 1024), 1), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(2048, 2), 1), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(4096, 222), 1), conf=conf)
+
+    # More than one partition
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(4096, 222), 2), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(2048, 999), 3), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(213, 123), 4), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(10000, 123), 10), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(20481, 0), 10), conf=conf)
+
+    # With "sort by"
+    sql = "select * from tmp_table sort by a limit {} offset {}"
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(1123, 50), 2), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(200, 150), 3), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(100, 200), 2), conf=conf)
+
+    sql = "select * from tmp_table sort by a desc limit {} offset {}"
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(200, 50), 5), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(1230, 50), 5), conf=conf)
+
--- a/integration_tests/src/main/python/limit_test.py
+++ b/integration_tests/src/main/python/limit_test.py
@@ -30,7 +30,7 @@ def test_simple_limit(data_gen):
 
 @allow_non_gpu('CollectLimitExec', 'GlobalLimitExec', 'ShuffleExchangeExec')
 @pytest.mark.skipif(is_before_spark_340(), reason='offset is introduced from Spark 3.4.0')
-def test_non_zero_offset_on_gpu():
+def test_non_zero_offset():
     conf = {
         'spark.rapids.sql.exec.CollectLimitExec': 'true',
         'spark.rapids.sql.exec.GlobalLimitExec': 'true'
--- a/integration_tests/src/main/python/limit_test.py
+++ b/integration_tests/src/main/python/limit_test.py
@@ -14,9 +14,8 @@
 
 import pytest
 
-from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_fallback_collect
+from asserts import assert_gpu_and_cpu_are_equal_collect, assert_gpu_and_cpu_are_equal_sql
 from data_gen import *
-from marks import allow_non_gpu
 from spark_session import is_before_spark_340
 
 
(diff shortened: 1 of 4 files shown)


revans2:
> A temporary PR for code discussion.

What exactly do you want to discuss? For Spark 3.4.0+ we either need to fall back to the CPU if we see an offset in the GlobalLimitExec or we need to implement it to match what Spark is doing. I am fine with either solution.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -129,7 +129,82 @@ case class GpuLocalLimitExec(limit: Int, child: SparkPlan) extends GpuBaseLimitE
...
 case class GpuGlobalLimitExec(limit: Int, child: SparkPlan, offset: Int) extends GpuBaseLimitExec {
+  assert(limit >= 0 || (limit == -1 && offset > 0))
   override def requiredChildDistribution: List[Distribution] = AllTuples :: Nil
+
+  override def doExecuteColumnar(): RDD[ColumnarBatch]  = {
+    val opTime = gpuLongMetric(OP_TIME)
+    val childRdd = child.executeColumnar()
+    childRdd.mapPartitions {
+      iter => new Iterator[ColumnarBatch] {
+        println(iter.hashCode())
```
sinkinben, line 140: TODO: Remove this `println`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -129,7 +129,81 @@ case class GpuLocalLimitExec(limit: Int, child: SparkPlan) extends GpuBaseLimitE
...
+
+          // If remainingOffset >= numRows, then we should skip this batch, and return an empty one
+          if (remainingOffset >= batch.numRows()) {
+            remainingOffset -= batch.numRows()
+            batch.safeClose()
+            return new ColumnarBatch(new ArrayBuffer[GpuColumnVector](batch.numCols()).toArray, 0)
+          }
+          // remainingOffset < batch.numRow(), we need to get batch[remainingOffset:]
+          withResource(new NvtxWithMetrics("limit and offset", NvtxColor.ORANGE, opTime)) { _ =>
+            if (remainingOffset > 0) {
```
sinkinben, line 156: Maybe re-writing these into:
```scala
val result = if (batch.numRows() > remainingLimit) {
  sliceBatchWithOffset(batch, remainingOffset, remainingLimit)
} else {
  batch
}
remainingLimit -= result.numRows()
remainingOffset = 0
result
```
will be better.

integration_tests/src/main/python/limit_test.py
```diff
@@ -30,19 +30,45 @@ def test_simple_limit(data_gen):
...
-        # We need some processing after the limit to avoid a CollectLimitExec
-        lambda spark: test_fn(spark).repartition(1),
-        'GlobalLimitExec',
-        conf = {'spark.sql.execution.sortBeforeRepartition': 'false'})
-
-    assert_gpu_fallback_collect(
-        test_fn,
-        'CollectLimitExec',
-        conf = {'spark.sql.execution.sortBeforeRepartition': 'false'})
+@allow_non_gpu('CollectLimitExec', 'GlobalLimitExec', 'ShuffleExchangeExec')
```
firestarman, line 33: This line is a duplicate of line 31, and both should be removed after supporting `offset`.

integration_tests/src/main/python/limit_test.py
```diff
@@ -30,19 +30,45 @@ def test_simple_limit(data_gen):
...
+@pytest.mark.skipif(is_before_spark_340(), reason='offset is introduced from Spark 3.4.0')
```
firestarman, line 34: This line is a duplicate of line 32, better to remove it.

integration_tests/src/main/python/limit_test.py
```diff
@@ -30,19 +30,45 @@ def test_simple_limit(data_gen):
...
+def test_non_zero_offset_on_gpu():
```
firestarman, line 35:
```suggestion
def test_non_zero_offset():
```

integration_tests/src/main/python/limit_test.py
```diff
@@ -30,19 +30,45 @@ def test_simple_limit(data_gen):
...
+    conf = {
```
firestarman, line 36: Why does it need these two configs ?

integration_tests/src/main/python/limit_test.py
```diff
@@ -30,19 +30,45 @@ def test_simple_limit(data_gen):
...
+
+    # More than one partition
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(4096, 222), 2), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(2048, 999), 3), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(213, 123), 4), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(10000, 123), 10), conf=conf)
+    assert_gpu_and_cpu_are_equal_collect(test_runner(sql.format(20481, 0), 10), conf=conf)
+
+    # With "sort by"
+    sql = "select * from tmp_table sort by a limit {} offset {}"
```
firestarman, line 66: May I know why need cases for `sort` ?

integration_tests/src/main/python/limit_test.py
```diff
@@ -30,19 +30,45 @@ def test_simple_limit(data_gen):
...
+    # num_slices is the number of partitions of data frame
+    def test_runner(sql, num_slices):
+        def test_instance(spark):
+            df = unary_op_df(spark, int_gen, num_slices=num_slices)
+            df.createOrReplaceTempView("tmp_table")
+            return spark.sql(sql)
+        return test_instance
+
+    sql = "select * from tmp_table limit {} offset {}"
+    # Only one partition
```
firestarman, line 50: Could you add comment why we need these cases ?

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -129,7 +129,81 @@ case class GpuLocalLimitExec(limit: Int, child: SparkPlan) extends GpuBaseLimitE
  * Take the first `limit` elements of the child's single output partition.
  */
 case class GpuGlobalLimitExec(limit: Int, child: SparkPlan, offset: Int) extends GpuBaseLimitExec {
+  assert(limit >= 0 || (limit == -1 && offset > 0))
   override def requiredChildDistribution: List[Distribution] = AllTuples :: Nil
+
+  override def doExecuteColumnar(): RDD[ColumnarBatch]  = {
```
firestarman, line 135: Could we support offset by updating the code in `GpuBaseLimitExec`, instead of rewriting here ?
Then for `GpuGlobalLimitExec`, we pass through the offset parameter, but for `GpuLocalLimitExec`, force the offset being 0.

firestarman (changes requested):

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -129,7 +129,81 @@ case class GpuLocalLimitExec(limit: Int, child: SparkPlan) extends GpuBaseLimitE
...
+    val opTime = gpuLongMetric(OP_TIME)
+    val childRdd = child.executeColumnar()
+    childRdd.mapPartitions {
```
firestarman, line 138:
```
childRdd.mapPartitions { iter =>
  new Iterator[ColumnarBatch] {
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -129,7 +129,81 @@ case class GpuLocalLimitExec(limit: Int, child: SparkPlan) extends GpuBaseLimitE
...
+      iter => new Iterator[ColumnarBatch] {
+        private var remainingLimit = limit - offset
+        private var remainingOffset = offset
+
+        override def hasNext: Boolean = remainingLimit > 0 && iter.hasNext
+
+        override def next(): ColumnarBatch = {
+          val batch = iter.next()
+
+          // If remainingOffset >= numRows, then we should skip this batch, and return an empty one
```
firestarman, line 148: We should not return an empty one, but get next batch from children, until skipping `offset` rows in total.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -129,7 +129,81 @@ case class GpuLocalLimitExec(limit: Int, child: SparkPlan) extends GpuBaseLimitE
...
+  assert(limit >= 0 || (limit == -1 && offset > 0))
   override def requiredChildDistribution: List[Distribution] = AllTuples :: Nil
+
+  override def doExecuteColumnar(): RDD[ColumnarBatch]  = {
+    val opTime = gpuLongMetric(OP_TIME)
+    val childRdd = child.executeColumnar()
+    childRdd.mapPartitions {
+      iter => new Iterator[ColumnarBatch] {
+        private var remainingLimit = limit - offset
+        private var remainingOffset = offset
```
firestarman, line 141: No need to cache the offset value, but keep getting batches from children until skip `offset` rows. We return an empty batch iff there are not enough rows. e.g. offset is 100, but this partition has only 50 rows.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -129,9 +129,93 @@ case class GpuLocalLimitExec(limit: Int, child: SparkPlan) extends GpuBaseLimitE
...
+            }
+            batch.safeClose(exception)
+          }
+        }
+      }
+    }
+  }
 }
 
+
```
firestarman, line 218: Useless space line, better to remove.

firestarman, line 218: useless space line, better to remove.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -129,9 +129,93 @@ case class GpuLocalLimitExec(limit: Int, child: SparkPlan) extends GpuBaseLimitE
...
+          }
+          // If the last batch is null, then we have offset >= numRows in this partition.
+          // In such case, we should return an empty batch
+          if (batch == null) {
+            return new ColumnarBatch(new ArrayBuffer[GpuColumnVector](numCols).toArray, 0)
+          }
+
+          // Here remainingOffset < batch.numRow(), we need to get batch[remainingOffset:]
+          withResource(new NvtxWithMetrics("limit and offset", NvtxColor.ORANGE, opTime)) { _ =>
+            if (remainingOffset > 0) {
```
firestarman, line 165: The `if` `else` does not cover the case of ``remainingLimit` == -1` .

firestarman: You can refer to the CPU paths.
```
    if (offset > 0) {
      if (limit >= 0) {
        child.execute().mapPartitionsInternal(iter => iter.slice(offset, limit))
      } else {
        child.execute().mapPartitionsInternal(iter => iter.drop(offset))
      }
    } else {
      super.doExecute()
    }
  }
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -129,9 +129,93 @@ case class GpuLocalLimitExec(limit: Int, child: SparkPlan) extends GpuBaseLimitE
...
+
+        def sliceBatchWithOffset(batch: ColumnarBatch, offset: Int, limit: Int): ColumnarBatch = {
+          var exception: Throwable = null
+          var table: Table = null
+
+          val numCols = batch.numCols()
+          val result = new ArrayBuffer[GpuColumnVector](numCols)
+
+          val end = Math.min(offset + limit, batch.numRows())
+          try {
```
firestarman, line 190: This `try catch` can be replaced with `withResource` method.

sinkinben:
build

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -58,61 +58,75 @@ trait GpuBaseLimitExec extends LimitExec with GpuExec with ShimUnaryExecNode {
...
-                assert(subVector != null)
-                resultCVs.append(GpuColumnVector.from(subVector, attr.dataType))
-                assert(subVector.getRowCount == remainingLimit,
-                  s"result rowcount ${subVector.getRowCount} is not equal to the " +
-                    s"remainingLimit $remainingLimit")
+        def sliceBatchWithOffset(batch: ColumnarBatch, offset: Int, limit: Int): ColumnarBatch = {
+          val numCols = batch.numCols()
+          val end = Math.min(offset + limit, batch.numRows())
+          val result = new ArrayBuffer[GpuColumnVector](numCols)
+          withResource(batch) { batch =>
```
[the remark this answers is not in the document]
firestarman: Better to `withResource` the input `batch` ealier than `result`, e.g.
```
  withResource(batch) { _ =>
    withResource(new ArrayBuffer[GpuColumnVector](numCols)) { result =>
      withResource(GpuColumnVector.from(batch)) { table =>
          // xxxxxxx
      }
    }
  }
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -58,61 +58,75 @@ trait GpuBaseLimitExec extends LimitExec with GpuExec with ShimUnaryExecNode {
...
+    rdd.mapPartitions { iter =>
       new Iterator[ColumnarBatch] {
-        private var remainingLimit = limit
+        private var remainingLimit = limit - offset
+        private var remainingOffset = offset
 
-        override def hasNext: Boolean = remainingLimit > 0 && cbIter.hasNext
+        override def hasNext: Boolean = iter.hasNext
 
         override def next(): ColumnarBatch = {
```
firestarman, line 73: Double check.
```suggestion
        override def next(): ColumnarBatch = {
          if (!hasNext) throw NoSuchElementException("Next on empty iterator")
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -58,61 +58,75 @@ trait GpuBaseLimitExec extends LimitExec with GpuExec with ShimUnaryExecNode {
...
+          // If the last batch is null, then we have offset >= numRows in this partition.
+          // In such case, we should return an empty batch
+          if (batch == null) {
+            return new ColumnarBatch(new ArrayBuffer[GpuColumnVector](numCols).toArray, 0)
+          }
+
+          // Here remainingOffset < batch.numRow(), we need to get batch[remainingOffset:]
+          withResource(new NvtxWithMetrics("limit and offset", NvtxColor.ORANGE, opTime)) { _ =>
+            var result: ColumnarBatch = null
+            // If limit != -1, then remainingLimit >= 0 is valid, otherwise remainingLimit < 0,
```
firestarman, line 93:
```suggestion

```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -58,61 +58,75 @@ trait GpuBaseLimitExec extends LimitExec with GpuExec with ShimUnaryExecNode {
...
+            // we should take batch[remainingOffset:numRows].
```
firestarman, line 94:
```suggestion

```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -58,61 +58,75 @@ trait GpuBaseLimitExec extends LimitExec with GpuExec with ShimUnaryExecNode {
...
+                remainingOffset = 0
+              } else {
+                result = if (batch.numRows() > remainingLimit) {
+                  sliceBatchWithOffset(batch, 0, remainingLimit)
+                } else {
+                  batch
+                }
+              }
+              remainingLimit -= result.numRows()
             } else {
```
firestarman, line 107:
```suggestion
            } else {
              // when remainingLimit < 0, it can go here only when limit == -1, meaning no limitation.
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -58,61 +58,75 @@ trait GpuBaseLimitExec extends LimitExec with GpuExec with ShimUnaryExecNode {
...
+          if (batch == null) {
+            return new ColumnarBatch(new ArrayBuffer[GpuColumnVector](numCols).toArray, 0)
+          }
+
+          // Here remainingOffset < batch.numRow(), we need to get batch[remainingOffset:]
+          withResource(new NvtxWithMetrics("limit and offset", NvtxColor.ORANGE, opTime)) { _ =>
+            var result: ColumnarBatch = null
+            // If limit != -1, then remainingLimit >= 0 is valid, otherwise remainingLimit < 0,
+            // we should take batch[remainingOffset:numRows].
+            if (remainingLimit >= 0) {
```
firestarman, line 95: Per the updated `hasNext`, `remainingLimit` can not be equal to 0. 
Even if `remainingLimit == 0`, we should return an empty batch.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -58,61 +58,75 @@ trait GpuBaseLimitExec extends LimitExec with GpuExec with ShimUnaryExecNode {
...
+              if (remainingOffset > 0) {
+                result = sliceBatchWithOffset(batch, remainingOffset, remainingLimit)
```
firestarman, line 97: Is it possible `remainingLimit > batch.numRows()` here ?

firestarman: Code between line #95 to line #106 may be simplified as the below. You can try it.
```
  if (remainingOffset == 0 && (remainingLimit >= batch.numRows() || limit < 0)) {
    batch
  } else {
    // need to slice the batch,
    val length = if (remainingLimit >= batch.numRows() || limit < 0) {
      batch.numRows()
    } else {
      remainingLimit
    }
    remainingLimit -= length
    sliceBatchWithOffset(batch, remainingOffset, length)
  }
```

integration_tests/src/main/python/limit_test.py
```diff
@@ -28,21 +27,40 @@ def test_simple_limit(data_gen):
...
-        conf = {'spark.sql.execution.sortBeforeRepartition': 'false'})
+def test_non_zero_offset(offset):
+    # offset is used in the test cases having no limit
+    # 0: corner case
+    # 1024: offset < df.numRows
+    # 2048: offset = df.numRows
+    # 4096: offset > df.numRows
+
+    offset_sql = "select * from tmp_table offset {}".format(offset)
+    assert_gpu_and_cpu_are_equal_sql(lambda spark: unary_op_df(spark, int_gen, length=2048), 'tmp_table',
```
firestarman, line 47: More than 100 chars at line 47, you can try this
```
    assert_gpu_and_cpu_are_equal_sql(
        lambda spark: unary_op_df(spark, int_gen, length=2048),
        'tmp_table',
        offset_sql,
        conf=offset_test_conf)
```

integration_tests/src/main/python/limit_test.py
```diff
@@ -28,21 +27,40 @@ def test_simple_limit(data_gen):
...
+    # (0, 0): Corner case: both limit and offset are 0
+    # (0, 10): Corner case: limit = 0, offset > 0
+    # (1024, 500): offset < limit && limit < df.numRows
+    # (2048, 456): offset < limit && limit = df.numRows
+    # (3000, 111): offset < limit && limit > df.numRows
+    # (500, 500): offset = limit
+    # (100, 600): offset > limit
+
+    limit_offset_sql = "select * from tmp_table limit {} offset {}".format(limit, offset)
+    assert_gpu_and_cpu_are_equal_sql(lambda spark: unary_op_df(spark, int_gen, length=2048), 'tmp_table',
```
firestarman, line 65: More than 100 chars at line 47, you can try this

    assert_gpu_and_cpu_are_equal_sql(
        lambda spark: unary_op_df(spark, int_gen, length=2048),
        'tmp_table',
        limit_offset_sql,
        conf=offset_test_conf)

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -58,61 +58,85 @@ trait GpuBaseLimitExec extends LimitExec with GpuExec with ShimUnaryExecNode {
...
+              // And remainingOffset > 0 will be used only once, for the latter batches in this
+              // partition, set remainingOffset = 0
+              val length = if (remainingLimit >= batch.numRows() || limit < 0) {
+                batch.numRows()
+              } else {
+                remainingLimit
+              }
+              result = sliceBatchWithOffset(batch, remainingOffset, length)
+              remainingOffset = 0
+              remainingLimit -= result.numRows()
```
firestarman, line 115:
```suggestion
              remainingLimit -= length
```

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -58,61 +58,85 @@ trait GpuBaseLimitExec extends LimitExec with GpuExec with ShimUnaryExecNode {
...
+          }
+
+          // Here 0 <= remainingOffset < batch.numRow(), we need to get batch[remainingOffset:]
+          withResource(new NvtxWithMetrics("limit and offset", NvtxColor.ORANGE, opTime)) { _ =>
+            var result: ColumnarBatch = null
+            // limit < 0 (limit == -1) denotes there is no limitation, so when
+            // (remainingOffset == 0 && (remainingLimit >= batch.numRows() || limit < 0)) is true,
+            // we can take this batch completely
+            if (remainingOffset == 0 && (remainingLimit >= batch.numRows() || limit < 0)) {
+              result = batch
```
firestarman, line 103: Need to update the `remainingLimit ` in the `if` path.
```suggestion
              result = batch
              remainingLimit -= result.numRows()
```

firestarman: Or you can do this `remainingLimit -= result.numRows()` out of the `if else`.

sql-plugin/src/main/scala/com/nvidia/spark/rapids/limit.scala
```diff
@@ -58,61 +58,85 @@ trait GpuBaseLimitExec extends LimitExec with GpuExec with ShimUnaryExecNode {
...
 
-        override def hasNext: Boolean = remainingLimit > 0 && cbIter.hasNext
+        override def hasNext: Boolean = (limit == -1 || remainingLimit > 0) && iter.hasNext
 
         override def next(): ColumnarBatch = {
-          val batch = cbIter.next()
-          withResource(new NvtxWithMetrics("limit", NvtxColor.ORANGE, opTime)) { _ =>
-            val result = if (batch.numRows() > remainingLimit) {
-              sliceBatch(batch)
+          if (!iter.hasNext) {
```
firestarman, line 76:
```suggestion
          if (!hasNext) {
```

firestarman: Better to use this.hasNext, which also take care of the limit things.

revans2:
@sinkinben How are you testing this. From my understanding we cannot build against 3.4.0 yet. Even then when I try to run `select * from test_table offset 1` I get errors back from 3.4.0-SNAPSHOT on the CPU. If you cannot test this can we please mark this PR as draft and pull it out of the 22.08 release?

revans2:
Sorry I was using an out of date 3.4.0 :(. Will test again.

integration_tests/src/main/python/limit_test.py
```diff
@@ -28,21 +27,40 @@ def test_simple_limit(data_gen):
...
-        'CollectLimitExec',
-        conf = {'spark.sql.execution.sortBeforeRepartition': 'false'})
+def test_non_zero_offset(offset):
+    # offset is used in the test cases having no limit
+    # 0: corner case
+    # 1024: offset < df.numRows
+    # 2048: offset = df.numRows
+    # 4096: offset > df.numRows
+
+    offset_sql = "select * from tmp_table offset {}".format(offset)
```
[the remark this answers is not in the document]
revans2: I read through the code and GlobalLimitExec still has no guarantee on ordering. It is a race to see what is returned. So the test will need to be updated to take this into account. Most of the time in local mode the race always ends up the same, but there is no guarantee of this.

revans2:
Another question. When I run the tests on the CPU.

"select id from test_table limit 1 offset 80" does not insert in a GlobalLimitExec at all. It just does a CollectLimitExec. Did you verify that the test is hitting your change?

integration_tests/src/main/python/limit_test.py
```diff
@@ -28,21 +27,40 @@ def test_simple_limit(data_gen):
             conf = {'spark.sql.execution.sortBeforeRepartition': 'false'})
 
 
-@allow_non_gpu('CollectLimitExec', 'GlobalLimitExec', 'ShuffleExchangeExec')
+# This conf allows CollectLimitExec and GlobalLimitExec to run on GPU
+offset_test_conf = {
+    'spark.rapids.sql.exec.CollectLimitExec': 'true',
+    'spark.rapids.sql.exec.GlobalLimitExec': 'true'
```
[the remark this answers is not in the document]
firestarman: Nice catch, this has been updated.

firestarman:
Move to release 22.10, since it is a feature for Spark 3.4.0.
Need to retarget to 22.10 when it is branched out.

firestarman:
> Another question. When I run the tests on the CPU.
> 
> "select id from test_table limit 1 offset 80" does not insert in a GlobalLimitExec at all. It just does a CollectLimitExec. Did you verify that the test is hitting your change?

Yes, Hi @sinkinben , you can refer to the existing tests in `limit_test` to figure out how to hit `GlobalLimitExec`.

sinkinben:
@revans2 @firestarman

I have added a helper function to test both `GlobalLimit` and `CollectLimit`.

```python
def offset_test_wrapper(sql):
    conf = {'spark.rapids.sql.exec.CollectLimitExec': 'true'}

    # Create dataframe to test CollectLimit
    def spark_df(spark):
        unary_op_df(spark, int_gen, length=2048, num_slices=1).createOrReplaceTempView("tmp_table")
        return spark.sql(sql)
    assert_gpu_and_cpu_are_equal_collect(spark_df, conf)

    # Create dataframe to test GlobalLimit
    def spark_df_repartition(spark):
        return spark_df(spark).repartition(1)
    assert_gpu_and_cpu_are_equal_collect(spark_df_repartition, conf)
```

For `spark_df` above, it will generate override-plan like:
```text
*Exec <CollectLimitExec> will run on GPU
  *Partitioning <SinglePartition$> will run on GPU
  ! <RDDScanExec> cannot run on GPU because GPU does not currently support the operator class org.apache.spark.sql.execution.RDDScanExec
    @Expression <AttributeReference> a#207 could run on GPU
```

For `spark_df_repartition`, it will generate override-plan like:
```text
*Exec <ShuffleExchangeExec> will run on GPU
  *Partitioning <SinglePartition$> will run on GPU
  *Exec <GlobalLimitExec> will run on GPU
    *Exec <ShuffleExchangeExec> will run on GPU
      *Partitioning <SinglePartition$> will run on GPU
      ! <RDDScanExec> cannot run on GPU because GPU does not currently support the operator class org.apache.spark.sql.execution.RDDScanExec
        @Expression <AttributeReference> a#213 could run on GPU
```

Since `repartition(1)`, it will guarantee that there is no races.

firestarman:
LGTM

sameerz:
@sinkinben please retarget this to 22.10

sinkinben:
build

firestarman (approved):

sinkinben:
build

firestarman (approved):

sinkinben:
build

PR #5950, opened 2022-07-05
Fix GpuShuffleCoalesce op time metric doesn't include concat batch time

### LINKED ISSUE #5891: [BUG] GpuShuffleCoalesce op time metric doesn't include concat batch time
**Describe the bug**
I was looking at 2 applications, one had the debug metrics and one didn't. With debug metrics I noticed that in GpuShuffleCoalesce the op time metric doesn't include the concat batch time. For example:

```
GpuShuffleCoalesce
concat batch time total (min, med, max (stageId: taskId))
2.4 s (7 ms, 23 ms, 49 ms (stage 42.0: task 12338))
output columnar batches: 99
input rows: 2,569,770
GPU semaphore wait time total (min, med, max (stageId: taskId))
10.7 s (0 ms, 112 ms, 249 ms (stage 42.0: task 12375))
input columnar batches: 4,200
op time total (min, med, max (stageId: taskId))
55 ms (0 ms, 0 ms, 7 ms (stage 42.0: task 12383))
output rows: 2,569,770
```

I looked at the one without debug metrics enabled and it only has the op time metric so the user can't really see how much time is taken.

```
GpuShuffleCoalesce

op time total (min, med, max (stageId: taskId))
60 ms (0 ms, 0 ms, 2 ms (stage 42.0: task 12346))
```

We should fix GpuShuffleCoalesce to include the concat batch time in op time. If there is a reason we don't do this we need to make sure all metrics are available with default spark.rapids.sql.metrics.level

res-life:
Fixes #5891 

The problem in the code is that the optime metric does not cover the concatenation time performed in the next() method. 
Update the range of optime to include the next().

### DIFF
--- a/docs/tuning-guide.md
+++ b/docs/tuning-guide.md
@@ -303,7 +303,7 @@ Custom Spark SQL Metrics are available which can help identify performance bottl
 | buildTime         | build time                   | Time to load the build-side of a join.                                                                                                                                             |
 | collectTime       | collect time                 | For a broadcast the amount of time it took to collect the broadcast data back to the driver before broadcasting it back out.                                                       |
 | computeAggTime    | aggregation time             | Time computing an aggregation.                                                                                                                                                     |
-| concatTime        | concat batch time            | Time to concatenate batches.                                                                                                                                                       |
+| concatTime        | concat batch time            | Time to concatenate batches. Runs on CPU.                                                                                                                                                      |
 | copyBufferTime    | copy buffer time             | Time spent on copying upstreaming data into Rapids buffers.                                                                                                                        |
 | filterTime        | filter time                  | Time spent applying filters within other operators, such as joins.                                                                                                                 |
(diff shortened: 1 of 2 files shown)


res-life:
build

res-life:
The op time now also covers the `SEMAPHORE_WAIT_TIME` time.

revans2:
> The op time now also covers the `SEMAPHORE_WAIT_TIME` time.

We do not want this. The semaphore wait time is supposed to be a separate thing. Ideally op time is only the time on the GPU.

From https://github.com/NVIDIA/spark-rapids/blob/branch-22.08/docs/tuning-guide.md#metrics



opTime | op time | Time that an operator takes, exclusive of the time for executing or  fetching results from child operators, and typically outside of the time  it takes to acquire the GPU semaphore.
-- | -- | --

res-life:
build

res-life:
build

Issue #5896, opened 2022-06-23
[BUG] Profiling tool on taking a really long time for integration tests


tgravescs:
**Describe the bug**
Our profiling tool integration tests are taking really long time now, we need to investigate. 

One example is `Took 1005559ms to process `where as that similar log file used to take 3944ms.

We recently checked in https://github.com/NVIDIA/spark-rapids/pull/5868 so perhaps something unexpected is happening with that.

tgravescs:
I have a feeling the intersect we added here is very costly when number of accumulators gets large:
https://github.com/NVIDIA/spark-rapids/blob/branch-22.08/tools/src/main/scala/com/nvidia/spark/rapids/tool/planparser/SQLPlanParser.scala#L85

zhanga5:
temporarily disabled profiling tool test on CSPs as impacted by this issue

PR #5951, opened 2022-07-05
Fix issue with Profiling tool taking a long time due to finding stage ids that maps to sql nodes

tgravescs:
fixes https://github.com/NVIDIA/spark-rapids/issues/5896

the time it took to do the intersection of the node accumulators with stage accumulators on large application really added up. Change to use a hashmap to be able to look up quickly.  Change to use Set for the stages to keep things consistent and output the same order.

Tested with our integration test event logs that were taking a long time and confirmed times much smaller.  One example took 13 seconds now with this change instead of 16 minutes.


tgravescs:
build

tgravescs:
build

nartal1 (approved):
LGTM. I'm curious to see the reduction in time in profiling tool integration tests.

Issue #5617, opened 2022-05-24
Qualification tool support parsing expressions (part 1)


tgravescs:
**Is your feature request related to a problem? Please describe.**
Enhance the Qualification tool to take expressions into account. Currently we look at exec but we want to do expressions as the next level.

as we implement them we should look if any duration information we can get from it.  I don't think there will be.

Since each Exec will have different format for expressions, we will end up having different parsers for parsing expressions information in Execs. Note that some of the Execs will not take any expressions, so we can ignore those.

## Execs triaged which require parsers 
- [ ] WindowExec

## Completed Execs - we support exprs in these Execs
- [x] FilterExec - https://github.com/NVIDIA/spark-rapids/pull/5880
- [x] ProjectExec - https://github.com/NVIDIA/spark-rapids/pull/5952
- [x] HashAggregateExec - https://github.com/NVIDIA/spark-rapids/pull/6042
- [x] ObjectHashAggregateExec - https://github.com/NVIDIA/spark-rapids/pull/6042
- [x] SortAggregateExec - https://github.com/NVIDIA/spark-rapids/pull/6042
- [x] SortExec - https://github.com/NVIDIA/spark-rapids/pull/6042

## Execs that may take expressions but don't show up in eventlog
- [x] RangeExec
- [x] UnionExec
- [x] CollectLimitExec
- [x] GlobalLimitExec
- [x] LocalLimitExec

PR #5952, opened 2022-07-06
Qualification tool: Parse expressions in ProjectExec

nartal1:
This contributes to https://github.com/NVIDIA/spark-rapids/issues/5617 .
In this PR we parse expressions which are in ProjectExec. Project may have aliases so made sure we are taking that into account. Tested with several expressions locally but added only 1 test each for supporting and non-supporting case. Supporting case pretty much covers all cases where the ProjectExec string needs to be parsed.
There was a bug in FilterExec parser where we are not taking into account if the predicate was `+`. Updated that as well.

### DIFF
--- a/tools/src/main/scala/com/nvidia/spark/rapids/tool/planparser/SQLPlanParser.scala
+++ b/tools/src/main/scala/com/nvidia/spark/rapids/tool/planparser/SQLPlanParser.scala
@@ -268,6 +268,25 @@ object SQLPlanParser extends Logging {
     funcName
   }
 
+  def parseProjectExpressions(exprStr: String): Array[String] = {
+    val parsedExpressions = ArrayBuffer[String]()
+    // remove the alias names before parsing
+    val pattern = """(AS) ([(\w# )]+)""".r
+    // This is to split multiple column names in Project. Project may have a function on a column.
+    // This will contain array of columns names specified in ProjectExec.
+    val paranRemoved = pattern.replaceAllIn(exprStr.replace("),", "::"), "")
+        .split(",").map(_.trim).map(_.replaceAll("""^\[+""", "").replaceAll("""\]+$""", ""))
+    val functionPattern = """(\w+)\(.*\)""".r
+    paranRemoved.foreach { case expr =>
+      val functionName = getFunctionName(functionPattern, expr)
+      functionName match {
+        case Some(func) => parsedExpressions += func
+        case _ => // NO OP
+      }
+    }
+    parsedExpressions.toArray
+  }
+
   def parseFilterExpressions(exprStr: String): Array[String] = {
     val parsedExpressions = ArrayBuffer[String]()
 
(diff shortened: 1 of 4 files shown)


nartal1:
build

tgravescs:
If you haven't yet we should test against the NDS event logs

tools/src/main/scala/com/nvidia/spark/rapids/tool/planparser/SQLPlanParser.scala
```diff
@@ -268,6 +268,25 @@ object SQLPlanParser extends Logging {
     funcName
   }
 
+  def parseProjectExpressions(exprStr: String): Array[String] = {
+    val parsedExpressions = ArrayBuffer[String]()
+    // remove the alias names before parsing
```
tgravescs, line 273: can you add an example in the comment

tools/src/main/scala/com/nvidia/spark/rapids/tool/planparser/SQLPlanParser.scala
```diff
@@ -268,6 +268,25 @@ object SQLPlanParser extends Logging {
...
+    val pattern = """(AS) ([(\w# )]+)""".r
+    // This is to split multiple column names in Project. Project may have a function on a column.
+    // This will contain array of columns names specified in ProjectExec.
+    val paranRemoved = pattern.replaceAllIn(exprStr.replace("),", "::"), "")
```
tgravescs, line 277: lets start putting more description on these so when we come back to them we can understand.... put example as well.

tools/src/test/scala/com/nvidia/spark/rapids/tool/planparser/SqlPlanParserSuite.scala
```diff
@@ -556,4 +557,59 @@ class SQLPlanParserSuite extends FunSuite with BeforeAndAfterEach with Logging {
...
   }
+
+  test("Expressions supported in ProjectExec") {
+    TrampolineUtil.withTempDir { outputLoc =>
+      TrampolineUtil.withTempDir { eventLogDir =>
+        val (eventLog, _) = ToolTestUtils.generateEventLog(eventLogDir,
+          "ProjectExprsSupported") { spark =>
+          import spark.implicits._
+          val df1 = Seq(9.9, 10.2, 11.6, 12.5).toDF("value")
+          df1.write.parquet(s"$outputLoc/testtext")
```
tgravescs, line 568: put parquet in name of directory

tools/src/test/scala/com/nvidia/spark/rapids/tool/planparser/SqlPlanParserSuite.scala
```diff
@@ -556,4 +557,59 @@ class SQLPlanParserSuite extends FunSuite with BeforeAndAfterEach with Logging {
...
+  }
+
+  test("Expressions not supported in ProjectExec") {
+    TrampolineUtil.withTempDir { outputLoc =>
+      TrampolineUtil.withTempDir { eventLogDir =>
+        val (eventLog, _) = ToolTestUtils.generateEventLog(eventLogDir,
+          "ProjectExprsNotSupported") { spark =>
+          import spark.implicits._
+          val df1 = spark.sparkContext.parallelize(List(10, 20, 30, 40)).toDF
+          df1.write.parquet(s"$outputLoc/testtext")
```
tgravescs, line 596: same here put parquet inname for outputloc

nartal1:
> If you haven't yet we should test against the NDS event logs

I tested this against NDS event logs and surprisingly we don't have any unsupported exprs in any of the ProjectExec. I created some custom eventlogs with unsupported exprs to be sure that the testing is done correctly.

nartal1:
build

